PR #8629, opened 2018-02-23
unittests/tests-core: Fix tests in loop not being executed

ochristi:
Patch to address unreachable code in test-core-atomic.
The condition of the loop was incorrect due copypasta, this fixes it.


smlng:
but you need to change the commit message, it should start with the module you fixed, e.g. `unittests: fix loop in tests-core-atomic`, that way you also get around the "CI Error" cause you message started with `fixes` 😉

ochristi:
Ah, I was not sure why it was failing. I changed the message and rebased.

aabadie:
All green, go!

PR #8585, opened 2018-02-17
boards: stk3600: add support (v2)

basilfx:
### Contribution description

This PR add supports for the Silicon Labs STK3600. The files are very similar to the SLTB001a and STK3700 already supported.

For convenience, [this is the diff](https://gist.github.com/basilfx/4bcf978fbc5e558614f28ce90001bb72) between STK3700 and STK3600 board files.

### Issues/PRs references

#8520

PR #8520, opened 2018-02-05
boards: stk3700: add support (v2)

basilfx:
### Contribution description

This PR add supports for the Silicon Labs [STK3700](https://github.com/basilfx/EFM2RIOT/blob/master/dist/doc/STK3700.md). In this new version, common files have been moved to `boards/common/silabs`.

The files are very similar to the SLTB001a already supported.

### Issues/PRs references

Made possible by #8139, successor of #8297, with recommendations of #8295.

PR #8630, opened 2018-02-23
boards: slstk3401a: add support (v2)

basilfx:
### Contribution description

This PR add supports for the Silicon Labs SLSTK3401a. The files are similar to the SLTB001a and STK3600 and STK3700 already supported. The main difference with the STK3x00 is that is has a second series MCU and it has a memory LCD (supported by U8g2).

For convenience, [this is the diff](https://gist.github.com/basilfx/f0221b8470f571002a8720e9e925e61f) between STK3600 and SLSTK3401a board files.

### Issues/PRs references

#8585
#8520


boards/slstk3401a/Makefile.dep
```diff
@@ -0,0 +1,9 @@
+ifneq (,$(filter saul_default,$(USEMODULE)))
+    USEMODULE += saul_gpio
```
aabadie: indentation should be 2 spaces here

boards/slstk3401a/Makefile.features
```diff
@@ -0,0 +1,14 @@
...
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
+FEATURES_PROVIDED += periph_spi
+FEATURES_PROVIDED += periph_timer
+FEATURES_PROVIDED += periph_uart
+
+# The board MPU family (used for grouping by the CI system)
+FEATURES_MCU_GROUP = cortex_m4_2
+
+include $(RIOTCPU)/efm32/Makefile.features
```
aabadie: just comparing to nucleos and I think if a '-' at the beginning of the line is required

boards/slstk3401a/include/board.h
```diff
@@ -0,0 +1,120 @@
...
+#define DISP_EN_PIN         GPIO_PIN(PD, 15)
+/** @} */
+
+/**
+ * @name    Temperature sensor configuration
+ *
+ * Connection to the on-board temperature/humidity sensor (Si7021).
+ * @{
+ */
+#ifndef SI7021_ENABLED
```
aabadie: I don't think this is the way to go. Just bind this to the fact that the module is loaded or not, e.g. use `MODULE_SI7021`, or even `MODULE_SI70XX`

aabadie: So in fact this part is not needed.

boards/slstk3401a/board.c
```diff
@@ -0,0 +1,40 @@
...
+void board_init(void)
+{
+    /* initialize the CPU */
+    cpu_init();
+
+    /* perform common board initialization */
+    board_common_init();
+
+    /* initialize the Si7021 sensor */
+#if SI7021_ENABLED
```
aabadie: use:
```c
#ifdef MODULE_SI70XX
```
instead. This flag is set automatically when one adds USEMODULE += si7021 in the application makefile.

boards/slstk3401a/Makefile.features
basilfx: According to [this](https://stackoverflow.com/a/34395651) link, it should be prefixed if you don't care if the file exists. But I do :-)

I see that others have it, but I wonder why the `cpu` include is somewhat optional...

boards/slstk3401a/Makefile.dep
basilfx: Fixed

boards/slstk3401a/Makefile.features
basilfx: Fixed it though.

boards/slstk3401a/include/board.h
basilfx: Removed this part.

boards/slstk3401a/board.c
basilfx: Fixed.

basilfx:
Fixed comments and rebased PR.

boards/slstk3401a/Makefile.features
gebart: I do think the optional include in the other boards is a bug though.

jia200x:
Don't have this board to test, but code-wise looks fine to me

basilfx:
@haukepetersen Do you have access to this board, and willing to test this one?

basilfx:
> Don't have this board to test, but code-wise looks fine to me

@jia200x I can also provide you with remote access (via SSH), if you want to try?

jia200x:
>@jia200x I can also provide you with remote access (via SSH), if you want to try?

Thanks for the offer. Unfortunately I won't be in the office until next monday, so I think would be nice if someone else can test it before :)

basilfx:
@jia200x Are you still up for testing this one? So far nobody had the time :-)

jia200x:
Tested, works as expected.

jia200x:
let's wait for @aabadie and @kYc0o

boards/slstk3401a/board.c
```diff
@@ -0,0 +1,40 @@
...
+
+void board_init(void)
+{
+    /* initialize the CPU */
+    cpu_init();
+
+    /* perform common board initialization */
+    board_common_init();
+
+#ifdef MODULE_SI70XX
```
kYc0o: is it possible to not use this module physically on the board? I mean, that you can disable it somehow so you need this guard?

boards/slstk3401a/include/periph_conf.h
```diff
@@ -0,0 +1,214 @@
...
+ */
+
+#ifndef PERIPH_CONF_H
+#define PERIPH_CONF_H
+
+#include "cpu.h"
+
+#include "periph_cpu.h"
+
+#include "em_cmu.h"
```
kYc0o: Can you put all the includes together?

boards/slstk3401a/board.c
```diff
@@ -0,0 +1,40 @@
...
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Bas Stottelaar <basstottelaar@gmail.com>
+ *
+ * @}
+ */
+
+#include "board.h"
+#include "board_common.h"
+
+#include "periph/gpio.h"
```
kYc0o: I think the space in between is not necessary.

boards/slstk3401a/include/board.h
```diff
@@ -0,0 +1,117 @@
...
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Bas Stottelaar <basstottelaar@gmail.com>
+ */
+
+#ifndef BOARD_H
+#define BOARD_H
+
+#include "cpu.h"
+
+#include "periph_conf.h"
```
kYc0o: Here too.

kYc0o:
You can squash the changes so the PR is ready.

boards/slstk3401a/board.c
basilfx: Well, removing this condition will always enable the sensor, even if you don't use it. I would therefore not want to remove it.

boards/slstk3401a/board.c
kYc0o: I didn't see this comment. IMHO this is not needed at all, since the inclusion of the driver is not really optional, the device is physically attached to the board and cannot be detached or deactivated, thus it belongs to the board and should be initialised whenever it's used.

The iotlab-m3 board has also some sensors attached, maybe you can take a look there to see how it's implemented, but as far as I can tell, they are only initialised when they're used through auto_init.

Another option would be to add a params file so you can access the sensor through SAUL using the default example.

aabadie: >  since the inclusion of the driver is not really optional, the device is physically attached to the board

yes it is, if the application doesn't load the driver module, there's no need for initializing and setting the pin that control it on the board.

boards/slstk3401a/board.c
kYc0o: So what you mean is that this pin can turn on/off the sensor? If that's the case, then this is a kind of "software detachment".

IMHO it looks a bit strange to have it here, but I don't have in mind right now a better way to integrate this functionality.

boards/slstk3401a/board.c
kYc0o: I just got the purpose of this, made my comment in my review.

boards/slstk3401a/board.c
basilfx: Just to clarify: this pin is not directly connected to the enable line of the Si7021: it's connected to a board-specific IC that will toggle VDD for the Si7021.

So to use this Si7021, you have to enable this pin to supply it VDD. The reason to disable it, is to not take into account its consumption when profiling it.

basilfx:
Squashed and rebased.

boards/slstk3401a/board.c
kYc0o: @basilfx Yes is what I thought, a real physical software controlled switch. That's why I didn't have an idea on how to integrate it, but I guess we should leave it as it is here and if the case comes across again we find a better integration.

kYc0o:
ACK. Merge when Murdock agrees.

basilfx:
Thank you all!

PR #8632, opened 2018-02-23
drivers/sdcard_spi: use of ucrc16 by default

cyrilfr:
### Contribution description

We probably should use checksum algorithms provided in RIOT by default. Here the standard lightweight CCITT CRC16 is used to avoid code duplication.

The CRC7 isn't implemented yet. Maybe a candidate for a future refactoring?

### DIFF
--- a/drivers/sdcard_spi/sdcard_spi.c
+++ b/drivers/sdcard_spi/sdcard_spi.c
@@ -24,6 +24,7 @@
 #include "sdcard_spi_params.h"
 #include "periph/spi.h"
 #include "periph/gpio.h"
+#include "checksum/ucrc16.h"
 #include "xtimer.h"
 
 #include <stdio.h>
@@ -44,9 +45,6 @@ static sd_rw_response_t _write_data_packet(sdcard_spi_t *card, char token, const
 
 /* CRC-7 (polynomial: x^7 + x^3 + 1) LSB of CRC-7 in a 8-bit variable is always 1*/
 static char _crc_7(const char *data, int n);
-
-/* CRC-16 (CRC-CCITT) (polynomial: x^16 + x^12 + x^5 + x^1) */
-static uint16_t _crc_16(const char *data, size_t n);
 
 /* use this transfer method instead of _transfer_bytes to force the use of 0xFF as dummy bytes */
 static inline int _transfer_bytes(sdcard_spi_t *card, const char *out, char *in, unsigned int length);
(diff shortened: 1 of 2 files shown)


MichelRottleuthner:
Thank you for addressing this! I had that in mind for a while now, but never took the time to actually do it - shame on me! :). 
>The CRC7 isn't implemented yet. Maybe a candidate for a future refactoring?

Yes! there are also a few more things in this driver that could be refactored. For reference the change from char to uint8_t discussed in #7853, some improvements regarding timeout handling, performance tweaks and while at it - some code cleanup. But one thing at a time...

drivers/sdcard_spi/sdcard_spi.c
```diff
@@ -622,9 +606,9 @@ static sd_rw_response_t _read_data_packet(sdcard_spi_t *card, char token, char *
 
         char crc_bytes[2];
         if (_transfer_bytes(card, 0, crc_bytes, sizeof(crc_bytes)) == sizeof(crc_bytes)) {
-            uint16_t data__crc_16 = (crc_bytes[0] << 8) | crc_bytes[1];
+            uint16_t data_crc16 = (crc_bytes[0] << 8) | crc_bytes[1];
 
-            if (_crc_16(data, size) == data__crc_16) {
+            if (ucrc16_calc_be((const uint8_t *)data, size, UCRC16_CCITT_POLY_BE, 0) == data_crc16) {
```
MichelRottleuthner: `const` isn't really required here, right?

drivers/sdcard_spi/sdcard_spi.c
```diff
@@ -712,8 +696,8 @@ static sd_rw_response_t _write_data_packet(sdcard_spi_t *card, char token, const
 
     if (_transfer_bytes(card, data, 0, size) == size) {
 
-        uint16_t data__crc_16 = _crc_16(data, size);
-        char crc[sizeof(uint16_t)] = { data__crc_16 >> 8, data__crc_16 & 0xFF };
+        uint16_t data_crc16 = ucrc16_calc_be((const uint8_t *)data, size, UCRC16_CCITT_POLY_BE, 0);
```
MichelRottleuthner: same as above. It does no harm, but for readability I'd remove the 'const' here.

MichelRottleuthner:
looks like something went wrong while squashing. Did you follow the procedure described [here](https://github.com/RIOT-OS/RIOT/wiki/Git-cheatsheet)?

cyrilfr:
Yes, but it does not seem to work...

MichelRottleuthner:
Congrats to your first contribution to RIOT :tada:

PR #8639, opened 2018-02-26
pkg/semtech-loramac: add link check support

aabadie:
### Contribution description

This PR adds support for the LoRaWAN "LinkCheckReq" MAC command.

The feature adds the possibility to validate the connectivity with the network by retrieving 2 informations:
- the number of gateways reachable by the end device
- the demodulation margin indicates the link margin in dB of the last received command (from the specs:
"A value of 0 means that the frame was received at the demodulation floor (0 dB or no
margin) while a value of 20, for example, means that the frame reached the gateway 20 dB
above the demodulation floor"

The link check request will be sent with the next TX packet and the reply will be received by the nodes in the following RX window. This means that one has to schedule the request and check for the result after a TX.

This PR also adds the `link_check` subcommand to the `loramac` command of the `pkg_semtech-loramac` test application shell. Use it as follows:

```
# schedule the link check
> loramac link_check
Link check request scheduled
# then retrieve the link check information
> loramac tx test
TX done
Link check information:
- Demodulation margin: 38
- Number of gateways: 1
```

~As a side note, I'm planning to refactor the API of the semtech loramac package by introducing a structure that will handle a loramac instance. This structure will also contain usefull attributes (radio device descriptor, keys, rx_data, etc) and will simplify the netdev adaption (I think) and help fixing thread safety issues.~ see #8798 

### Issues/PRs references

Now based on ~#8798~

### DIFF
--- a/pkg/semtech-loramac/contrib/semtech_loramac.c
+++ b/pkg/semtech-loramac/contrib/semtech_loramac.c
@@ -241,6 +241,15 @@ static void mlme_confirm(MlmeConfirm_t *confirm)
             }
             break;
 
+        case MLME_LINK_CHECK:
+            if (confirm->Status == LORAMAC_EVENT_INFO_STATUS_OK) {
+                DEBUG("[semtech-loramac] link check received\n");
+                msg_t msg;
+                msg.type = MSG_TYPE_LORAMAC_LINK_CHECK;
+                msg.content.ptr = confirm;
+                msg_send(&msg, semtech_loramac_pid);
+            }
+
         default:
             break;
     }
@@ -313,6 +322,7 @@ void _init_loramac(semtech_loramac_t *mac,
     semtech_loramac_set_class(mac, LORAMAC_DEFAULT_DEVICE_CLASS);
     semtech_loramac_set_tx_port(mac, LORAMAC_DEFAULT_TX_PORT);
     semtech_loramac_set_tx_mode(mac, LORAMAC_DEFAULT_TX_MODE);
+    mac->link_chk.available = false;
 }
 
 static void _join_otaa(semtech_loramac_t *mac)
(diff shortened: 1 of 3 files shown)


pkg/semtech-loramac/contrib/semtech_loramac.c
```diff
@@ -580,6 +594,28 @@ uint8_t semtech_loramac_join(uint8_t type)
     return SEMTECH_LORAMAC_JOIN_SUCCEEDED;
 }
 
+void semtech_loramac_request_link_check(void)
+{
+    _semtech_loramac_link_check_received = false;
+    MlmeReq_t mlmeReq;
+    mlmeReq.Type = MLME_LINK_CHECK;
+    LoRaMacMlmeRequest(&mlmeReq);
```
jia200x: I would try to use here something similar to the [mlme_confirm](https://github.com/RIOT-OS/RIOT/blob/master/pkg/semtech-loramac/contrib/semtech_loramac.c#L242) function. That is, sending a message to the LoRaMAC thread to process this MLME request. Otherwise, this won't be thread safe.

jia200x:
>As a side note, I'm planning to refactor the API of the semtech loramac package by introducing a structure that will handle a loramac instance. This structure will also contain usefull attributes (radio device descriptor, keys, rx_data, etc) and will simplify the netdev adaption (I think) and help fixing thread safety issues.

That would be awesome

pkg/semtech-loramac/contrib/semtech_loramac.c
jia200x: (Since in this case, this function is called from main thread)

aabadie: I know that problem, it's the same story with the get/set functions. But for the moment, it's fine I think, until I finalize the refactoring I'm working on (as described in the initial comment).

jia200x: maybe [this branch](https://github.com/Inria-Chile/RIOT/blob/gnrc_lorawan/drivers/include/net/netdev2/lorawan.h) is helpful for you. It's an abandoned branch though

aabadie: Thanks, it helps certainly! That's indeed in the spirit of what I have in mind.

aabadie:
> That would be awesome

Here it is: #8798. I still need to adapt this one.

aabadie:
> I still need to adapt this one.

Done and tested.

aabadie:
@jia200x, next one is here ;) and it is now rebased with latest master.

jia200x:
@aabadie tested and works good.

The only issue I see is the fact the link_check command is not being show in the `usage`:
```
2018-04-19 14:41:30,338 - INFO #  loramac
2018-04-19 14:41:30,341 - INFO # Usage: loramac <get|set|join|tx>
```
But works as expected!

aabadie:
> The only issue I see is the fact the link_check command is not being show in the usage

Ah yes, good catch! It should be fixed now

aabadie:
@jia200x, do you still see blocking issues with this one ?

aabadie:
ping @jia200x

jia200x:
Looks good! It's fine to me

jia200x:
ACK & GO

PR #8641, opened 2018-02-26
crypto/cbc: add return values to documentation

pwillem:
### Contribution description
Add documentation for return values of `cipher_encrypt_cbc` and `cipher_decrypt_cbc`

### Issues/PRs references
none

### DIFF
--- a/sys/include/crypto/modes/cbc.h
+++ b/sys/include/crypto/modes/cbc.h
@@ -20,6 +20,7 @@
 #ifndef CRYPTO_MODES_CBC_H
 #define CRYPTO_MODES_CBC_H
 
+#include <stddef.h>
 #include "crypto/ciphers.h"
 
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


sys/include/crypto/modes/cbc.h
```diff
@@ -20,6 +20,7 @@
 #ifndef CRYPTO_MODES_CBC_H
 #define CRYPTO_MODES_CBC_H
 
+#include <stddef.h>
```
aabadie: Is this really needed ?

pwillem: I tought RIOT's header files should be stand alone. This header file only includes `ciphers.h` which only includes `stdint.h`.

More practical: I removed RIOT's `assert.h` in order to test with c11 `static_assert` and I got an compilation error in this header because `size_t` wasn't defined in my app.

aabadie: ok

sys/include/crypto/modes/cbc.h
```diff
@@ -36,9 +37,13 @@ extern "C" {
  * @param input_len  length of the input data
  * @param output     pointer to allocated memory for encrypted data. It has to
  *                   be of size data_len + BLOCK_SIZE - data_len % BLOCK_SIZE.
+ *
+ * @return            <0 on error
```
aabadie: Looking at the code, I would give the return code (e.g. CIPHER_ERR_INVALID_LENGTH and CIPHER_ERR_ENC_FAILED) returned to be more precise.

sys/include/crypto/modes/cbc.h
```diff
@@ -36,9 +37,13 @@ extern "C" {
...
+ * @return            0 on success
```
aabadie: Not sure this is needed, because 0 bytes are consumed on success so the next return would be enough

sys/include/crypto/modes/cbc.h
```diff
@@ -50,9 +55,12 @@ int cipher_encrypt_cbc(cipher_t* cipher, uint8_t iv[16], const uint8_t* input,
  * @param input_len  length of the input data
  * @param output     pointer to allocated memory for plaintext data. It has to
  *                   be of size input_len.
+ *
+ * @return            <0 on error
```
aabadie: Same comment about return codes

pwillem:
ok squashed

aabadie:
Thanks @pwillem. All green, go!

PR #8648, opened 2018-02-27
drivers: MH-Z19: initial support

bergzand:
### Contribution description

The MH-Z19 is a relative cheap (15 euro) CO2 ppm sensor. This PR adds basic support for reading the CO2 measurement over the UART interface. Saul integration is provided with this PR.

As UART is asynchronous, I had to implement a bit of synchronisation and timeout support in the driver itself. This is the reason two mutexes are used and the reason the const keyword had to be dropped in the saul interface.

### Issues/PRs references

None

### DIFF
--- a/drivers/include/mhz19.h
+++ b/drivers/include/mhz19.h
@@ -48,7 +48,7 @@ enum {
     MHZ19_OK            =  0,       /**< everything was fine */
     MHZ19_ERR_UART      = -1,       /**< error initializing the UART bus */
     MHZ19_ERR_TIMEOUT   = -2,       /**< UART timeout on retrieving sensor data */
-    MHZ19_ERR_CHECKSUM  = -3,       /**< Checksum failure on received data */
+    MHZ19_ERR_CHECKSUM  = -3,       /**< checksum failure on received data */
 };
 
 /**
@@ -62,9 +62,9 @@ typedef struct {
  * @brief   Device descriptor for a MH-Z19 device
  */
 typedef struct {
-    const mhz19_params_t *params;   /**< Device parameters */
+    const mhz19_params_t *params;   /**< device parameters */
     mutex_t mutex;                  /**< protect against simulaneous access */
-    mutex_t sync;                   /**< Transfer complete or timeout sync */
+    mutex_t sync;                   /**< transfer complete or timeout sync */
     uint8_t idx;                    /**< rx buffer index */
     uint8_t rxmem[MHZ19_BUF_SIZE];  /**< rx buffer */
 } mhz19_t;
--- a/drivers/include/mhz19.h
+++ b/drivers/include/mhz19.h
@@ -1,0 +1,105 @@
+/*
+ * Copyright (C) 2018 Koen Zandberg
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_mhz19 MH-Z19 CO2 sensor
+ * @ingroup     drivers_sensors
+ *
+ * @brief       MH-Z19 CO2 sensor driver
+ *
+ * ## Description
+ *
+ * The MH-Z19 is a C02 sensor. Measurements are provided in ppm over UART and
+ * PWM. Only the UART interface is implemented in this driver. PPM range from
+ * 0 (theoretically) to 2000 or 5000 depending on the sensor settings.
+ *
+ * @note The sensor requires considerable time before accurate measurements are
+ * provided
+ *
+ * @{
+ *
+ * @file
+ * @brief       Interface definition for the mh-z19 CO2 sensor driver.
+ *
+ * @author      Koen Zandberg <koen@bergzand.net>
+ */
+
+#ifndef MHZ19_H
+#define MHZ19_H
+
+#include "periph/uart.h"
+#include "saul.h"
+#include "mhz19_internals.h"
+#include "mutex.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Status and error return codes
+ */
+enum {
+    MHZ19_OK            =  0,       /**< everything was fine */
+    MHZ19_ERR_UART      = -1,       /**< error initializing the UART bus */
+    MHZ19_ERR_TIMEOUT   = -2,       /**< UART timeout on retrieving sensor data */
+    MHZ19_ERR_CHECKSUM  = -3,       /**< Checksum failure on received data */
+};
+
+/**
+ * @brief   Device initialization parameters
+ */
+typedef struct {
+    uart_t uart;        /**< UART device that sensor is connected to */
+} mhz19_params_t;
+
+/**
+ * @brief   Device descriptor for a MH-Z19 device
+ */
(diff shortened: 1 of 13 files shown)


sys/auto_init/saul/auto_init_mhz19.c
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (C) 2017 Koen Zandberg
```
basilfx: 2018

drivers/include/mhz19.h
```diff
@@ -0,0 +1,105 @@
...
+#endif
+
+/**
+ * @brief   Status and error return codes
+ */
+enum {
+    MHZ19_OK            =  0,       /**< everything was fine */
+    MHZ19_ERR_UART      = -1,       /**< error initializing the UART bus */
+    MHZ19_ERR_TIMEOUT   = -2,       /**< UART timeout on retrieving sensor data */
+    MHZ19_ERR_CHECKSUM  = -3,       /**< Checksum failure on received data */
```
basilfx: small caps checksum

drivers/mhz19/include/mhz19_internals.h
```diff
@@ -0,0 +1,92 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @name    MH-Z19 Baud rate
+ *
+ * Fixed at 9600 by design.
+ */
+#define MHZ19_UART_BAUDRATE             9600
```
basilfx: I think some whitespace can be removed. I dont think you have to align all defines in the file at the same level, but per block.

drivers/mhz19/include/mhz19_params.h
```diff
@@ -0,0 +1,77 @@
+/*
```
basilfx: Adapt to recent (SAUL) param unification.

sys/auto_init/saul/auto_init_mhz19.c
```diff
@@ -0,0 +1,71 @@
...
+{
+    for (unsigned i = 0; i < MHZ19_NUMOF; i++) {
+        LOG_DEBUG("[auto_init_saul] initializing mhz19 #%u\n", i);
+
+        if (mhz19_init(&mhz19_devs[i], &mhz19_params[i]) < 0) {
+            LOG_ERROR("[auto_init_saul] error initializing mhz19 #%u\n", i);
+            continue;
+        }
+
+        /* temperature */
```
basilfx: Not really temperature :-)

tests/driver_mhz19/README.md
```diff
@@ -0,0 +1,12 @@
+# MH-Z19 Driver Test
+
+## Introduction
+This test will test if the MH-Z19 C02 gas sensor is working.
```
basilfx: Mention MH-Z19B as well. It works as well

basilfx:
> The MH-Z19 is a relative cheap ...
Just like the Si70xx, would it make sense to rename this to MH-Z19x? I don't have a strong opinion on this.

> ... I had to implement a bit of synchronisation ..
I can only tell it works, but I don't have an expert opinion about this

It works, with a few changes (tested SAUL and driver test application using STK3600).

basilfx:
It's a nice device. I know its an initial implementation, but could you implement an additional command (e.g. autocalibrate, or set range), so that we have a basis for sending commands? It would not be a blocking thing, but nice to have this already in place :)

bergzand:
The mh-z19 combines reading and writing in the same command structure it shouldn't be to hard to modify it for write operations. I'd rather not implement the zero point and the span point commands. From what I've read it's rather easy to break the sensor via those commands.

For now I'd like to keep this PR focused on an initial implementation with only support for reading the sensor. This because the other commands are either not that useful or seem to require some black magic.

sys/auto_init/saul/auto_init_mhz19.c
bergzand: Fixed

drivers/include/mhz19.h
bergzand: Fixed

drivers/mhz19/include/mhz19_internals.h
bergzand: Removed some whitespace here and below this line

drivers/mhz19/include/mhz19_params.h
bergzand: Should be adapted now

sys/auto_init/saul/auto_init_mhz19.c
bergzand: Not really no, removed :)

tests/driver_mhz19/README.md
bergzand: Added a small explanation about the MH-Z19B also working

bergzand:
@basilfx I think I addressed all your comments, thanks for the review!

basilfx:
My comments have been addressed, and I am able to run the test example/saul out-of-the box on a STK3600.

I don't feel 100% comfortable to ACK for code style. Maybe someone else can take a quick look?

bergzand:
If someone else is going to review this, I'm mostly interested in whether my approach to tackle (sychronize) the asynchronous nature of the UART interface is acceptable :)

gebart:
Regarding the synchronization between isr and driver: I have a hard time following when and what is supposed to happen at each synchronization point. The general idea with locking a mutex twice and unlocking from isr is sound, but it is a bit difficult to follow the code when there is more than one synchronization point in the same function (get_ppm function).
@bergzand do you think you might be able to refactor it a little to make it more readable. It might be enough to add some comments and/or factor out common statements into a separate inline function to make it less confusing.

bergzand:
@gebart Thanks for the comments, I've refactored the code a bit to split the transmission and the PPM calculation, to hopefully clarify the locking around a single exchange a bit. Now with extra comments!

drivers/mhz19/mhz19.c
```diff
@@ -81,35 +81,59 @@ int mhz19_init(mhz19_t *dev, const mhz19_params_t *params)
     return MHZ19_OK;
 }
 
-int mhz19_get_ppm(mhz19_t *dev, int16_t *ppm)
+/*
```
basilfx: I think you need `/**`.

drivers/mhz19/mhz19.c
```diff
@@ -81,35 +81,59 @@ int mhz19_init(mhz19_t *dev, const mhz19_params_t *params)
...
+     */
     mutex_lock(&dev->sync);
 
-    /* Unlock mutex again to reset it */
+    /* Unlock mutex again to reset it to unlocked */
     mutex_unlock(&dev->sync);
-    /* Clean up timer */
+    /* Clean up timer in case it didn't fire yet */
     xtimer_remove(&timer);
+
```
basilfx: Remove blank line.

basilfx:
@bergzand, any updates on this? :-)

aabadie:
It would be sad to close a PR in such a good shape. @bergzand, can you provide some output of the test application ? If you could also squash, that would be perfect.

kaspar030:
Followup in https://github.com/RIOT-OS/RIOT/pull/13126.

basilfx:
#13126 is now merged.

PR #8552, opened 2018-02-13
sys/vfs: Exclude stdio file numbers from auto allocation

gebart:
### Contribution description

Do not auto-allocate the stdio file descriptor numbers to avoid conflicts between normal file system users and stdio drivers such as uart_stdio, rtt_stdio which need to be able to bind to these specific file descriptor numbers.
Adds a regression test for vfs_bind to check that it never returns stdio file descriptors (which fails without the other commit)

Tested on native and mulle.

### Issues/PRs references

Fixes #8309 
An alternative to #8310

PR #8310, opened 2017-12-21
sys/vfs: reserve file descriptors for stdin, stdout, and stderr

samkumar:
### Contribution description

Previously, the code to reserve file descriptors 0, 1, and 2 for stdin, stdout, and stderr was in the uart_stdio code. However, this is insufficient for platforms that use the rtt_stdio module to write to the screen. Therefore, I moved the code to the VFS code, where it will get executed no matter what mechanism is needed to write characters to the screen.

### Issues/PRs references

Fixes #8309.

Issue #8309, opened 2017-12-20
VFS Module does not reserve file descriptors 0, 1, and 2


samkumar:
#### Description

The VFS module does not reserve file descriptors 0, 1, and 2 for stdin, stdout, and stderr. As a result, the following bugs exist:
* Opening a file could be assigned to one of these descriptors.
* Utilities in newlib such as `printf` do not work. The output is written to file descriptor 1 and `uart_stdio_write` never gets called.

More generally, there seems to be no `vfs_init` procedure, so it seems that all the `_vfs_open_files` array is unintialized. In my opinion, the correct solution is to add a `vfs_init` procedure that initializes the `_vfs_open_files` array and reserves file descriptors 0, 1, and 2.

#### Steps to reproduce the issue

Method 1: Call printf().
Method 2: Open a socket with the posix_sockets module.

#### Expected results

Expected Results for Method 1: The output is written to the screen.
Expected Results for Method 2: The resulting file descriptor is greater than 2.

#### Actual results

Actual Results for Method 1: Nothing is written to the screen.
Actual Results for Method 2: The resulting file descriptor is 0.

#### Versions

I am building the code from a Linux computer running Ubuntu 16.04.
The platform I am targeting is HAMILTON (@immesys).

Output from `print_toolchain_versions.sh`:
```
./print_toolchain_versions.sh 
Installed toolchain versions
----------------------------
          native gcc: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609
          msp430-gcc: msp430-gcc (GCC) 4.6.3 20120301 (mspgcc LTS 20120406 unpatched)
             avr-gcc: missing/error
   arm-none-eabi-gcc: arm-none-eabi-gcc (15:4.9.3+svn231177-1) 4.9.3 20150529 (prerelease)
    mips-mti-elf-gcc: missing/error
               clang: clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)
arm-none-eabi-newlib: "2.2.0"
 mips-mti-elf-newlib: missing/error
            avr-libc: missing/error (missing/error)
            cppcheck: missing
          coccinelle: missing
```

immesys:
This would affect other platforms too though right?

samkumar:
Yes, but after further investigation I believe that the problem may be restricted to `rtt_stdio`. The `uart_stdio` module registers stdin, stdout, and stderr with VFS in its init function.

So there are two options that I have identified.

[...]

PR #8654, opened 2018-02-28
sys/rtt_stdio: reserve file descriptors for stdin, stdout, and stderr

samkumar:
### Contribution description

The uart_stdio module binds to file descriptors 0, 1, and 2 when VFS is enabled. This modifies rtt_stdio to do the same. Without this pull request, stdio does not work as expected when rtt_stdio and VFS are used simultaneously.

### Issues/PRs references

Fixes the second half of #8309. The first half of #8309 was fixed by #8552. See also Pull Request #8310.

### DIFF
--- a/sys/rtt_stdio/rtt_stdio.c
+++ b/sys/rtt_stdio/rtt_stdio.c
@@ -78,12 +78,21 @@
  */
 
 #include <stdio.h>
+#if MODULE_VFS
+#include <unistd.h>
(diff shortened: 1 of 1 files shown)


samkumar:
I think @gebart would be a good review, since @gebart also reviewed Issue #8309 and Pull Request #8310.

gebart:
@samkumar Thank you for contributing!

Issue #6577, opened 2017-02-08
periph/i2c: tracking the remodeling


haukepetersen:
The changes proposed in #6575 and #6576 need many efforts to be undertaken. This issue will track the progress and should help to synchronize the work efforts.

Modules in RIOT that need adaption:
CPUS:
- [x] atmega_common (@mali #9252 )
- [x] efm32 (@basilfx #9208)
- [x] nrf51 (@aabadie #9477)
- [x] nrf52 (@dylad #9334)
- [x] cc2538 (@smlng #9347 )
- [x] kinetis (@gebart #9262)
- [x] sam0_common (@dylad #9198)
- [x] stm32 (@aabadie #9202 )

DRIVERS AND TESTS:
- [x] adcxx1c (@vincent-d #9210)
- [x] ads101x (#9165)
- [x] adxl345 (@dylad)
- [x] at30tse75x (@aabadie #9229)
- [x] bh1750fvi (@MichelRottleuthner #9494)
- [x] bmx055 (@smlng #9466)
- [x] bmx280 (maybe @haukepetersen with #8383, @basilfx #9206 for the time being)
- [x] bmp180 (@aabadie #9228)
- [x] ds1307 (@MichelRottleuthner #9493)
- [x] fxos8700 (@gebart #9268)
- [x] hdc1000 (@Hyungsin #9455)
- [x] hih6130 (@gebart #9270)
- [x] hts221(@aabadie ) new one => #9298
- [x] ina220 (@gebart, #9269)
- [x] isl29020 (@aabadie, #8688 )
- [x] isl29125 (@MichelRottleuthner, #9392)
- [x] jc42 (@MichelRottleuthner #9492)
- [x] l3g4200d (@aabadie, #8698)
- [x] lc709203f (@MichelRottleuthner #9481)
- [x] lis3mdl (@aabadie #9276)
- [x] lps331ap (@aabadie, #8757 )
- [x] lsm303dlhc (@aabadie, #8699 )
- [x] lsm6dsl (@aabadie ) => new one #9299
- [x] mag3110 (@aabadie, #8758)
- [x] mma8x5x (@vincent-d #9265 )
- [x] mpl3115a2 (@aabadie #9275)
- [x] mpu9150 (@aabadie #9277)
- [x] pn532 (@MichelRottleuthner, ~~#9432~~ #9458)
- [x] si114x (@aabadie #9314)
- [x] si70xx (@basilfx #9205)
[...]

PR #8688, opened 2018-02-28
drivers/isl29020: apply unified params definition scheme

aabadie:
### Contribution description

This PR cleans up the isl29020 implementation to make more inline with other drivers.

### Issues/PRs references

#6577

### DIFF
--- a/drivers/isl29020/isl29020.c
+++ b/drivers/isl29020/isl29020.c
@@ -43,16 +43,14 @@ int isl29020_init(isl29020_t *dev, const isl29020_params_t *params)
 
     /* acquire exclusive access to the bus */
     i2c_acquire(DEV_I2C);
-    /* initialize the I2C bus */
-    i2c_init_master(DEV_I2C, I2C_SPEED_NORMAL);
 
     /* configure and enable the sensor */
     tmp = (ISL29020_CMD_EN | ISL29020_CMD_MODE |
            ISL29020_RES_INT_16 | DEV_RANGE | (DEV_MODE << 5));
-    res = i2c_write_reg(DEV_I2C, DEV_ADDR, ISL29020_REG_CMD, tmp);
+    res = i2c_write_reg(DEV_I2C, DEV_ADDR, ISL29020_REG_CMD, tmp, 0);
     /* release the bus for other threads */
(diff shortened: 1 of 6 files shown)


aabadie:
Just tested on iotlab-m3, works

drivers/isl29020/isl29020.c
```diff
@@ -59,16 +62,16 @@ int isl29020_read(const isl29020_t *dev)
...
-    i2c_release(dev->i2c);
+    ret = i2c_read_reg(DEV_I2C, DEV_ADDR, ISL29020_REG_LDATA, &low, 0);
+    ret += i2c_read_reg(DEV_I2C, DEV_ADDR, ISL29020_REG_HDATA, &high, 0);
+    i2c_release(DEV_I2C);
     if (ret < 2) {
         return -1;
     }
     res = (high << 8) | low;
-    DEBUG("ISL29020: Raw value: %i - high: %i, low: %i\n", res, high, low);
+    DEBUG("ISL29020: Raw value: %i - high: %i, low: %i\n", res, high, low, 0);
```
dylad: I guess there is some wrong copy&paste here

sys/auto_init/saul/auto_init_isl29020.c
```diff
@@ -49,13 +54,12 @@ extern saul_driver_t isl29020_saul_driver;
 
 void auto_init_isl29020(void)
 {
-    for (unsigned int i = 0; i < ISL29020_NUM; i++) {
-        const isl29020_params_t *p = &isl29020_params[i];
+    assert(ISL29020_NUM == ISL29020_INFO_NUM);
```
dylad: Just wondering why you used `==` and not `>= `?

aabadie: Because the loop below is accessing 2 different lists, one for the list of device descriptor and one for list device information: each list length must match.

drivers/isl29020/isl29020.c
aabadie: fixed

kYc0o:
Please don't forget to advertise the API change on the driver usage.

PR #8696, opened 2018-02-28
driver/adxl345: rename enums to avoid conflicts

Josar:
driver/adxl345: rename enums to avoid conflicts


PeterKietzmann:
ACK and go

PR #7067, opened 2017-05-16
drivers/adxl345: fixes & cleanup

dylad:
@haukepetersen As promised, this PR should address all your missing reviews from #6799 
I hope I didn't forget something.

PR #8697, opened 2018-02-28
drivers/lis3dh: apply unified driver params scheme + cleanup

aabadie:
### Contribution description

This PR update the params definitions scheme and makes some cleanup in the driver implementation.

### Issues/PRs references

### DIFF
--- a/drivers/include/lis3dh.h
+++ b/drivers/include/lis3dh.h
@@ -736,10 +736,8 @@ typedef struct {
  * @brief   Device descriptor for LIS3DH sensors
  */
 typedef struct {
-    spi_t spi;              /**< SPI device the sensor is connected to */
-    spi_clk_t clk;          /**< clock speed of the SPI bus */
-    gpio_t cs;              /**< Chip select pin */
-    int16_t scale;          /**< Current scale setting of the sensor */
+    lis3dh_params_t params; /**< Device initialization parameters */
+    uint16_t scale;         /**< Internal sensor scale */
 } lis3dh_t;
 
 /**
--- a/sys/auto_init/saul/auto_init_lis3dh.c
+++ b/sys/auto_init/saul/auto_init_lis3dh.c
@@ -29,7 +29,7 @@
 /**
  * @brief   Define the number of configured sensors
  */
-#define LIS3DH_NUM    (sizeof(lis3dh_params)/sizeof(lis3dh_params[0]))
+#define LIS3DH_NUM    (sizeof(lis3dh_params) / sizeof(lis3dh_params[0]))
 
 /**
  * @brief   Allocate memory for the device descriptors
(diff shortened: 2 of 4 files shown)


sys/auto_init/saul/auto_init_lis3dh.c
```diff
@@ -49,18 +54,17 @@ extern saul_driver_t lis3dh_saul_driver;
...
-        int res;
+    assert(LIS3DH_NUM == LIS3DH_INFO_NUM);
 
+    for (unsigned int i = 0; i < LIS3DH_NUM; i++) {
         LOG_DEBUG("[auto_init_saul] initializing lis3dh #%u\n", i);
 
-        res = lis3dh_init(&lis3dh_devs[i], &lis3dh_params[i]);
-        if (res < 0) {
+        if (lis3dh_init(&lis3dh_devs[i], &lis3dh_params[i]) < 0) {
             LOG_ERROR("[auto_init_saul] error initializing lis3dh #%u\n", i);
```
gebart: I would prefer printing res in the error message instead of removing the variable

PeterKietzmann:
@aabadie would you mind to address @gebart's comment so we can proceed here?

sys/auto_init/saul/auto_init_lis3dh.c
gebart: Alright

drivers/include/lis3dh.h
```diff
@@ -736,10 +736,8 @@ typedef struct {
  * @brief   Device descriptor for LIS3DH sensors
  */
 typedef struct {
-    spi_t spi;              /**< SPI device the sensor is connected to */
-    spi_clk_t clk;          /**< clock speed of the SPI bus */
-    gpio_t cs;              /**< Chip select pin */
-    int16_t scale;          /**< Current scale setting of the sensor */
+    lis3dh_params_t params; /**< Device initialization parameters */
```
gebart: We would save some ram if we place params in ROM and use a pointer. The extra indirection should not matter here, the extra few cpu cycles and ROM accesses will not be limiting the performance of the device

gebart: Some of the parameters are only used by the init function and do not need to be cached in RAM if you are not using a pointer.

aabadie: Agreed, but IMHO this is a more general discussion regarding initialization parameters and how to the handle them. There's also [a wiki page](https://github.com/RIOT-OS/RIOT/wiki/Guide:-Writing-a-device-driver-in-RIOT#device-descriptor-and-parameter-configuration) that explains how to write a device driver. This page doesn't mention the use of a pointer for params.
Other drivers are also not using the params as pointer. Do you think it would make sense to update this in follow-up PR(s) or is it really blocking for you here ?

gebart: I don't want to block this as there are more pressing issues to address in other parts of the system. However, there is a mention of pointers on the wiki page you linked, excerpt below:

>In many cases it makes sense, to copy the xxx_params data into the device descriptor during initialization. In some cases, it is however better to just link the params data via pointer and only copy selected data.

aabadie: Indeed, you are right!  But in the current state and regarding the discussion here and in #7067, this is confusing I think.

PR #8698, opened 2018-02-28
drivers/l3g4200d: apply unified driver params scheme + cleanup

aabadie:
### Contribution description

This PR updates the params definitions scheme and makes some cleanup in the driver implementation.

### Issues/PRs references

#6577


aabadie:
Just tested on iotlab-m3, works. But I realized that int1 and int2 pins are not used in the code. So the refactoring could go further by simply removing them.

aabadie:
> However, for future reference it might be better to stick to API adaptions and have the param scheme stuff in a followup PR.

The problem is that this PR was opened long time ago. I just rebased it on new_i2c_if and applied the I2C rework there for simplicity.

PR #8699, opened 2018-02-28
drivers/lsm303dlhc: apply unified driver params scheme + cleanup

aabadie:
### Contribution description

This PR updates the params definitions scheme and makes some cleanup in the internal driver implementation.

### Issues/PRs references


aabadie:
Just tested on iotlab-m3 (not on iotlab), works

dylad:
Can someone test it on a second CPU arch so we can push the big green button ?

PR #5757, opened 2016-08-17
make: fix linking

kaspar030:
Ever wondered why exactly files needed to be put into "UNDEF" in order to link correctly?

When considering archives, ld only considers objects in archives if there is at least one symbol in them needed elsewhere. Also, it doesn't seem to consider a symbol if it already found a weak symbol before.

In the current make system, this leads to:
- weak symbols not overridden by strong symbols unless the object is passed to the linker outside of an archive (in "UNDEF")
- weak symbols don't get overridden unless at least one symbol of that object is used somewhere (this is why overridden ISRs are only actually used if some other function of that file is used, e.g., isr_timer0 will only be used instead of the weak-default dummy handler if e.g., timer_init() is called)

Both feel hacky to me.

This PR encloses processing of module-archives in "--whole-archive", causing all objects to be always considered for linking. That way, the UNDEF-objects are not needed anymore.

Unfortunately this always overrides all defined isr defined in periph_*, as they are currently all always built. So to not get somewhat bloated binaries, this depends on ~~#5065~~ #7241.

~~Also, the config module breaks, so #5756 is also required.~~ (#5756 merged)

PR #8711, opened 2018-03-01
make: Use ld -r for intermediate linking of each module

gebart:
### Contribution description

As discussed in #5757, using `ld -r --unique` and creating intermediate object files can be preferable to creating archives (using `ar rcs`, current practice). This eliminates hidden issues such as a weak symbol may override a strong symbol even though both are included in the final link.
This means that each module will first be linked into a single object file `module.o` instead of a static archive `module.a`.

New build system environment variables are introduced:
 - `LD`: target low level linker, default `$(PREFIX)ld`, must be ld, gcc does not work here.
 - `ILINK`: command used for intermediate linking (incremental linking), default `$(LD)`.
 - `ILINKFLAGS`: flags for incremental linking, default `-r`. Certain platforms may need to specify the CPU type or other flags here, e.g. AVR, MIPS

This PR eliminates the variable `UNDEF` entirely because it is no longer needed.

### Issues/PRs references

Initial suggestion in #5757


gebart:
TODO: 
- [ ] verify that binaries are still correct.
- [ ] Check if any pkgs need updating

cladmi:
In #5757 PR, @kaspar030 needed to handle certain libraries without the '--whole-archive' and so with the lazy linking algorithm.

Even if some/(all?) of the packages could be fixed to produce an object file, there may be reasons for an external libraries or a module to need to be built as an archive, or also we could be lazy to migrate external build scripts all at once.
So starting with a mechanism that allows some code to be compiled and used as library could help migrate in a more smoother way. It could be adapted from the `EXTRAL_LIBS` of #5757 PR.

Also, if moving away from 'libraries' archives, would it make sense to rename `BASELIBS`?


I found some references to the '.a' output in `bindist`:
```
examples/bindist/Makefile:# and that the resulting .a should be saved when doing "make bindist"
examples/bindist/README.md:compiled and linked binary, bindist.a, abc.a and Makefiles.
makefiles/bindist.inc.mk:DIST_FILES += $(BIN_USEMODULE:%=bin/$(BOARD)/%.a)
```

And the `bindist` example does not build for the moment because it does not find `main`.

Makefile.base
```diff
@@ -65,10 +65,17 @@ DEP := $(OBJC:.o=.d) $(OBJCXX:.o=.d) $(ASSMOBJ:.o=.d)
...
 
-$(BINDIR)/$(MODULE).a $(OBJ): | $(BINDIR)/$(MODULE)/
+$(BINDIR)/$(MODULE).o $(OBJ): | $(BINDIR)/$(MODULE)/
 
-$(BINDIR)/$(MODULE).a: $(OBJ) | ${DIRS:%=ALL--%}
-	$(Q)$(AR) $(ARFLAGS) $@ $?
+# An empty object file is included in the link to avoid situations where modules
+# do not provide any code (e.g. sys, some boards), causing the build to fail
+# with a missing object file.
+EMPTYOBJ = $(BINDIR)/$(MODULE)/.empty.o
```
Kijewski: I am not sure that using a generic empty.o will work. If I'm not mistaken (I did not test it), the linker will reject an object file if its not the right architecture.

I guess something like that would be more robust:

```make
ifeq $(,$(OBJ))
  OBJ := $(BINDIR)/$(MODULE)/empty.o

  $(BINDIR)/$(MODULE)/empty.o: $(RIOTBASE)/empty.c
        $(CC) $(CFLAGS) -w -c -o $@ $<
ifeq
```

```c
static void __attribute__((unused)) _placeholder(void) {}
```

gebart: It works, try it. It is the correct architecture because of using the real CFLAGS when generating it.
Not sure the ifeq suggestion is robust because of lazy evaluation of recursive variables, does OBJ always have its final value when interpreting this makefile?

kaspar030:
I'm very looking forward to this one!

gebart:
Feel free to join in if you feel like working on this. Currently some pkgs have problems linking because of multiple inclusions of the same object file in the final link

cladmi:
I really like the concept of packages registering themselves as building an `.a` but I do not really like the name `USEARCHIVE`. For me it does not match what we have currently in `USEMODULE` `USEPKG` used in dependencies. Its more something like `MODULES_PRODUCING_ARCHIVES_INSTEAD_OF_OBJECTS` but should have a better name than this.

I will create sub pull requests to fix what is currently broken in master.

gebart:
rebased to fix conflicts

gebart:
@cladmi I take any suggestions for a name better than USEARCHIVE. I think it's a bit too cryptic right now.

kaspar030:
> I take any suggestions for a name better than USEARCHIVE. I think it's a bit too cryptic right now.

Traditionally statically linked libraries are linked as .a. Thus I'd suggest LINKLIBS or just LIBS.

cladmi:
I would like to have a precise name, even maybe a strange name so that it cannot be understood directly as something else, and I would prefer that people look it up in the documentation as it should be an exception when not possible to do otherwise.

`LIBSMODULES`, `ARCHIVEMODULES`, `ARCHIVELIBS` ?

As it's meant to be used in `modules.inc.mk` when setting `BASELIBS`, with the following I would understand it

    _OBJECTMODULES = $(filter-out $(ARCHIVEMODULES), $(REALMODULES))
    BASELIBS += $(_OBJECTMODULES:%=$(BINDIR)/%.o)
    BASELIBS += $(ARCHIVEMODULES:%=$(BINDIR)/%.a)

kaspar030:
> As it's meant to be used in modules.inc.mk when setting BASELIBS

The name BASELIBS is not a set-in-stone one either, more something from the past that can be re-thought when oit makes sense.

kaspar030:
> Traditionally statically linked libraries are linked as .a. Thus I'd suggest LINKLIBS or just LIBS.

Let's not discuss naming so much. Just go with LIBS, it has been standard for linking static libraries in archives for decades. It is also the shortest, and it is less confusing than "ARCHIVELIBS", "EXTRALIBS", "MODULELIBS", which all contain more confusion than they add information.

Everything unclear about "LIBS" can be clarified with a line or two in the documentation.

gebart:
I noticed that some tests gain a few bytes in both ROM and RAM usage. This could be because of identical section names in different objects being merged while incrementally linking, like what happened in the semtech_loramac region module, in https://github.com/RIOT-OS/RIOT/pull/8711/commits/acdfcbf87b598d3e37bb9e95ad0ce268682a1243 where each of the regions defined a static variable named Channels which is a buffer that is only used locally, but they all end up in the same section (.bss.Channels) and then when one of them is linked in, all of the others are also linked, even though there are no references to them. The linker can only garbage collect entire sections.

I tried experimenting with the [linker flag `--split-by-file`](https://linux.die.net/man/1/ld) but it seems to mess up the debugging info and linking fails with multiple definitions of debug symbols.. 

Any suggestions on how to address this?

cladmi:
As now you introduced `LIBS`, I think that having the `REALPKGS` and `REALMODULES` separation is not necessary.
So `REALPKGS` could be removed and all the changes from `USEMODULE` to `USEPKG` also.
It would simplify the review I think.

Standardizing this would be great but it was even decided to go in the other direction and keep `USEPKG` an internal thing only: https://github.com/RIOT-OS/RIOT/issues/6668

Makefile.include
```diff
@@ -355,13 +355,13 @@ endif # RIOTNOLINK
 $(ELFFILE): $(BASELIBS)
 	$(Q)$(_LINK) -o $@
 
-$(BINDIR)/$(APPLICATION_MODULE).a: $(RIOTBUILD_CONFIG_HEADER_C) $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS)
+$(BINDIR)/$(APPLICATION_MODULE).o: $(RIOTBUILD_CONFIG_HEADER_C) $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS)
```
cladmi: Here the `$(USEPKG:%=$(BINDIR)/%.a)` should I think also be with a `.o`.

Makefile.include
```diff
@@ -355,13 +355,13 @@ endif # RIOTNOLINK
...
 	$(Q)DIRS="$(DIRS)" "$(MAKE)" -C $(APPDIR) -f $(RIOTMAKE)/application.inc.mk
-$(BINDIR)/$(APPLICATION_MODULE).a: FORCE
+$(BINDIR)/$(APPLICATION_MODULE).o: FORCE
 
 # Other modules are built by application.inc.mk and packages building
-_SUBMAKE_LIBS = $(filter-out $(BINDIR)/$(APPLICATION_MODULE).a $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS), $(BASELIBS))
-$(_SUBMAKE_LIBS): $(BINDIR)/$(APPLICATION_MODULE).a $(USEPKG:%=$(BINDIR)/%.a)
+_SUBMAKE_LIBS = $(filter-out $(BINDIR)/$(APPLICATION_MODULE).o $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS), $(BASELIBS))
+$(_SUBMAKE_LIBS): $(BINDIR)/$(APPLICATION_MODULE).o $(USEPKG:%=$(BINDIR)/%.a)
```
cladmi: Here too  the `$(USEPKG:%=$(BINDIR)/%.a)` should I think also be with a `.o`.

gebart:
- Rebased
- Added --unique to ILINKFLAGS
- Added specific ILINKFLAGS for riscv

The use of --unique makes the linker keep multiple sections of the same name separate instead of merging them which lets the garbage collector remove more unused local symbols. Without this flag, some modules e.g. semtech_loramac_mac_region, would allocate more memory than they were actually using, because of the same variable name occurring as a static global in multiple files.

gebart:
Murdock passes now, except for the squashing check.
Ready for review and discussion @cladmi @kaspar030 @Kijewski

cladmi:
Sorry I missed your ping, I think you can squash before the review it will be easier as it is now stable. I put it on my next week review list. Ping me if I forget !

cladmi:
I will try to revert the changes to packages USEPKG to USEMODULES as this is part of a main PR resolution https://github.com/RIOT-OS/RIOT/pull/9307. It should not be necessary anymore as there is LIBS to handle the specific cases that can't be build as objects.

cladmi:
Question, what did you use to debug size issues or symbols and things ?

> because of the same variable name occurring as a static global in multiple files.

Is this not a bad practice ?

gebart:
@cladmi 
>Question, what did you use to debug size issues or symbols and things ?

I used objdump, nm and readelf and the map file generated during build to see what changed between builds. 

>>because of the same variable name occurring as a static global in multiple files.
>
>Is this not a bad practice ?

No, it's fine. These are file scope global variables (static keyword), so there's no name collision.

cladmi:
Update:

I tried diffing size between master and this branch:

* `gnrc_minimal` has same build sizes
* `nanocoap-server` without `sha256/hashes` has same size for wsn430. So I will try debugging this.

gebart:
some debugging for wsn430-v1_3b:
https://gist.github.com/gebart/2117e89d39dcd2ac08a7cfbb857ec535
use some diff tool to check the differences.
New symbols for ld-r branch are:

000012be l     O .bss	00000040 isr_ctx
00004b02 l     F .text	00000084 isr_handler
00004b86 g     F .text	00000064 isr_port1
00004bea g     F .text	00000064 isr_port2

These combined add up to the additional memory shown by the size tool, except for 8 bytes of ROM, which may be padding.

gebart:
It is possible that FEATURES_REQUIRED contain periph_gpio from somewhere, even though it is not used by the application

gebart:
This is likely the problem :
https://github.com/RIOT-OS/RIOT/blob/a6c7b0a61b34ca2a76cfe254a46d80f27d571e84/Makefile.dep#L788

Possible solution:
Let CPU periph build a static archive for now. Instead of an object file.

gebart:
@cladmi I tested locally to rebase this branch on top of #9845, and it fixes the size increase in wsn430.

cladmi:
> Possible solution:
Let CPU periph build a static archive for now. Instead of an object file.

I agree with this. There might also be some 'cpu_name_periph_common' that need it.
It would still be a required to fix them to be able to use XFA for auto_init/shell for these periphs but better introduce the new linking earlier anyway.

gebart:
@cladmi When creating #9845, I grepped through the sources to determine which modules uses gpio_init_int and add them to the appropriate Makefile.dep files. That is a more complete and future proof solution than to create a static archive from the periph drivers.

kaspar030:
> That is a more complete and future proof solution than to create a static archive from the periph drivers.

+1, let's fix this right instead of increasing debt. Maybe even as followup.

cladmi:
@gebart I missed some punctuation in my comment which made it ambiguous, I was saying, yes go with LIBS for periph :)
We will fix the rest later (or even later than later).

cladmi:
There are some other size difference that can be displayed with [https://cladmi.github.io/devpages/cmpmurdocksize.html?...](https://cladmi.github.io/devpages/cmpmurdocksize.html?origin=https%3A%2F%2Fci.riot-os.org%2FRIOT-OS%2FRIOT%2F9836%2F14074eeda106ded94f7f46ca46189993ba659683%2Fsizes.json&destination=https%3A%2F%2Fci.riot-os.org%2FRIOT-OS%2FRIOT%2F8711%2F05e9a2e1f1868c101838ed55f1af4d89bcc4643c%2Fsizes.json)
However, it requires a CORS enable add-on and the sort function is not working correctly…

gebart:
My counter proposal is to merge #9845 first.

@cladmi could you post a summary here of the significant differences? I.e. Anything bigger than 8-12 bytes.

cladmi:
The summary is "Total: 23648, Same: 569, Diff: 23079" and from a scrollbar evaluation more than 90% of the application with a difference have more than 8-12 bytes, but 80% are below 50 bytes.

I will rebase after #9845 and compare again with murdock and locally. It may be a size difference between different build paths in murdock.

gebart:
rebased after #9845 was merged

cladmi:
I forgot something, is that for a PR murdock is merging in the "current" master before testing. So it is hard to compare two PRs if they are not merged in the same reference:

At the beginning of murdock output there is:

```
-- github reports HEAD of master as 8ada7a84c517249e30076b0c3e387651abc380a6
--- creating merge commit ...
-- merging f5cab9b2e9d782b73ed1371f38085ac05d0de421 into 8ada7a84c517249e30076b0c3e387651abc380a6
--- done.
```
The "into" should be the same for comparing PRs.

kaspar030:
> I forgot something, is that for a PR murdock is merging in the "current" master before testing.

Sorry, I thought that was clear... the only way to get this is to trigger a build after each master merge and delay completion of PR's that are built before the master build is ready. We've already decided that that's totally acceptable for the added information.

cladmi:
It knew it, but I forgot to take it into account when testing and giving results.
But the result is not that different anyway…

### Comparing ld-r with #9836 (rebase PR)

[Diff #9836 #8711](https://cladmi.github.io/devpages/cmpmurdocksize.html?origin=https%3A%2F%2Fci.riot-os.org%2FRIOT-OS%2FRIOT%2F9836%2F36ad10a174bde4f2e28aaa2257f4c36518a39723%2Fsizes.json&destination=https%3A%2F%2Fci.riot-os.org%2FRIOT-OS%2FRIOT%2F8711%2Ff5cab9b2e9d782b73ed1371f38085ac05d0de421%2Fsizes.json)

Summary:

Total: 23646, Same: 1044, Diff: 22602

There are ~2000 that have a size increase `>100`, I only kept these ones listed.
https://gist.github.com/cladmi/6c92082bc97dbbe35f7d8e195ef14f7a#file-diff-md
The biggest changes are really board specific, which makes sense.

And ~290 that have a size decrease `>24` which goes to `-580`.
https://gist.github.com/cladmi/6c92082bc97dbbe35f7d8e195ef14f7a#file-diff_negative-md
Mainly `mips` board. It could be a sign of something missing, or something really not working before and solved.

I already know there are some false negative, build values with no data and `unittests` different.
I explain it below.

### Comparing two times the same commit

Before parsing the results, I tested with two time the same commit in two different PRs, they are tested on the same "master" commit and I get differences. https://github.com/RIOT-OS/RIOT/pull/9836 and https://github.com/RIOT-OS/RIOT/pull/9855
Many unittests have different size, from -196 to +292.
Some sizes are not always parsed:
* `tests/cortexm_common_ldscript` (because of [parallel output](https://ci.riot-os.org/RIOT-OS/RIOT/9855/36ad10a174bde4f2e28aaa2257f4c36518a39723/output/compile/tests/cortexm_common_ldscript/arduino-mkr1000:gnu.txt))
* `pic32` boards printing `objcopy:` before: [`tests/lwip_sock_tcp | pic32-wifire:gnu`](https://ci.riot-os.org/RIOT-OS/RIOT/9836/36ad10a174bde4f2e28aaa2257f4c36518a39723/output/compile/tests/lwip_sock_tcp/pic32-wifire:gnu.txt) and [`tests/gnrc_sock_udp | pic32-clicker:gnu`](https://ci.riot-os.org/RIOT-OS/RIOT/9855/36ad10a174bde4f2e28aaa2257f4c36518a39723/output/compile/tests/gnrc_sock_udp/pic32-clicker:gnu.txt).
[Diff #9836 #9855](https://cladmi.github.io/devpages/cmpmurdocksize.html?origin=https%3A%2F%2Fci.riot-os.org%2FRIOT-OS%2FRIOT%2F9836%2F36ad10a174bde4f2e28aaa2257f4c36518a39723%2Fsizes.json&destination=https%3A%2F%2Fci.riot-os.org%2FRIOT-OS%2FRIOT%2F9855%2F36ad10a174bde4f2e28aaa2257f4c36518a39723%2Fsizes.json)
I will take this into account while doing the diff with this PR.

cladmi:
I will try debugging some of the CPU issues this week. Thanks for the `objdump` example in https://gist.github.com/gebart/2117e89d39dcd2ac08a7cfbb857ec535

gebart:
if any non-msp430 CPUs have increased size it is probably also because of similarily unused GPIO ISRs being linked. It should be obvious in a symbol table comparison.

cladmi:
For many arm boards I have this difference when going to the `ld-r` PR.

``` diff
diff -u objdump.RIOT_test objdump.RIOT
--- objdump.RIOT_test   2018-09-04 17:43:06.764371566 +0200
+++ objdump.RIOT        2018-09-04 17:43:12.701016764 +0200
@@ -209,7 +208,7 @@
 l     O .text  00000006 new_addr.7216
 g     F .text  00000010 nmi_default
 g     F .text  00000060 od_hex_dump
- w    F .text  00000002 panic_arch
+g     F .text  0000001c panic_arch
 g     F .text  00000002 periph_init
 l     O .bss   000000c8 _pktbuf
 l     F .text  00000070 _pktbuf_alloc
```

Does not it mean that it was actually never used ? (maybe only on my computer gcc-8)

cladmi:
I am currently using this https://gist.github.com/cladmi/35b7816e7fb7bd75e8f5040dcac646db to test.
In a directory I have the script, `RIOT_test` which is the reference PR and `RIOT` with this one.

cladmi:
It also looks like `pthread_tls` is always included when `pthread` is used which was not the case before.
Maybe a separate module will be required here.

gebart:
@cladmi thanks for analyzing further. I will look into the pthread_tls issue right now. Will post back here in a few minutes

gebart:
@cladmi I updated the branch with a fix for the pthread_tls issue.
The size difference you are seeing for panic_arch is a bug in master where the weak default in core is used instead of the cpu specific version in cortexm_common etc.

https://github.com/RIOT-OS/RIOT/blob/master/core/panic.c#L44

https://github.com/RIOT-OS/RIOT/blob/master/cpu/cortexm_common/panic.c

sys/posix/pthread/Makefile
```diff
@@ -1 +1,6 @@
+# exclude pthread_tls if not selected explicitly
+ifeq (,$(filter pthread_tls,$(USEMODULE)))
+  SRC := $(filter-out pthread_tls.c,$(wildcard *.c))
```
cladmi: I do not really like this pattern with the filter-out, I think it would be clearer to just create a subdirectory with a normal module.

cladmi:
@gebart For `panic_arch` yes, I just find it strange that we only find it now.

kaspar030:
> @gebart For panic_arch yes, I just find it strange that we only find it now.

Maybe our code never panicked. ;)

sys/posix/pthread/Makefile
kaspar030: Or use submodules?

gebart: Need to rename the file pthread_tls.c -> tls.c or make the submodule name pthread_pthread_tls

cladmi:
Added a fix in the ref PR for `panic_arch` to compare.

cladmi:
We are moving forward: 

[Total: 24281, Same: 7819, Diff: 16462](https://cladmi.github.io/devpages/cmpmurdocksize.html?origin=https%3A%2F%2Fci.riot-os.org%2FRIOT-OS%2FRIOT%2F9836%2F6b029a74d3f5a2826d766c3a629511b4ec24a888%2Fsizes.json&destination=https%3A%2F%2Fci.riot-os.org%2FRIOT-OS%2FRIOT%2F8711%2Fa453bfccc0ea2a5a73b0bba764e34874cefa5205%2Fsizes.json)

cladmi:
But esp boards have been merged too, so new CPU support.

cladmi:
Boards with a lot of changes remaining:

* `native`
* `hifive1` I do not see that much difference on my computer…
    * examples/nanocoap_server | hifive1:gnu | Total diff: 2470
* `chronos`
* mips boards, but there are also big increase and decrease so not really sure… I am thinking it is related to `USEMODULE += periph_timer` for `gettimeofday` but strange that there are big decrease too.

Also packages examples do change a lot, will see what I can find there.

Makefile.include
cladmi: I did a PR to remove the named file for now has they are not needed and ambiguous in case a package could be a pseudomodule. https://github.com/RIOT-OS/RIOT/pull/9935

Only `BASELIBS` will list objects and archives.

cladmi:
@gebart Could you split the `mips` changes in a separate PR, I cannot really explain them myself.

I will try to split the current `periph_gpio_irq` PRs and also migrate the CPUs that are currently not following it.

cpu/mips32r2_common/newlib_syscalls_mips_uhi/syscalls.c
```diff
@@ -46,6 +46,8 @@ extern char _sheap; /* start of the heap */
 extern char _eheap; /* end of the heap */
 char *heap_top = &_sheap + 4;
 
+/* MIPS newlib crt implements _init,_fini and _exit and manages the heap */
+#ifndef __mips__
```
gebart: @cladmi do you mean that this should be forked off?

cladmi: If possible yes, also with the related changes in `Makefile.dep`.

I think it would would help comparing the size change between master and this one as it is my main way of debugging currently.

cladmi:
I will rebase the branch to fix the Conflicting files and have a murdock build run.

cladmi:
The rebased version diff is similar to the old one, only adaptation for `ROM_LEN` and `RAM_LEN`  in atmega cpus and that `cpu/atmega_common/periph/gpio.c` had some refactoring in a function.

I checked with:
```
# git commit-tree NEW_HEAD^{tree} -p riot/master -p PR_SHA1 -m message
git commit-tree 9900a9d0d15c3d1006559745038b9089d8657f1d^{tree} -p c7e6d1599082c953fa1f7b89348d757e283d8020 -p cf783b53207ca5de32943bcb7bfd3fbc7b28f63b -m "diff"
```

I push it to your branch.

cladmi:
I rebased again to compare the size with https://github.com/RIOT-OS/RIOT/pull/9836

I think we are also at a point where the branch would benefit from being squashed. What do you think ?
Maybe you want to do it as I reverted commits at the end that you may disagree with.
Or I can take care if you prefer.

cladmi:
> Maybe we should split the pthread_tls.c and cortexm panic arch into their own PRs?

It is planned but I did not manage to find time for it. But it is the reason why I put them in first too.

> I don't remember what was the reason for the reverted commits. Do they not make sense any more? Does it link correctly now without those pkg fixes?

I think it linked properly as I manage to compare size.

My reason to remove them is that it was needed at some point to not duplicate files in `BASELIBS` but I think it has been made useless by the commit adding `LIBS` as you made it https://github.com/RIOT-OS/RIOT/commit/87b1c6a3f5ff9b8a29d3e097c672b0670e33c0c5#diff-5bc412bce420f428058c9b896bedbc2d
I also still plan to replace all `USEPKG` by `USEMODULE` in #9307 and will try to split independent parts out of it (== the one not touching the dependencies location).

My original comment about it:
>>  I will try to revert the changes to packages USEPKG to USEMODULES as this is part of a main PR resolution #9307 . It should not be necessary anymore as there is LIBS to handle the specific cases that can't be build as objects.

gebart:
#10195 is an interesting feature and touches some of the same parts as this PR. I think we should merge #10195 first and then rebase this on top of that. It is possible that we might have to refactor this yet again, but in the end using thin archives in the intermediate steps can speed up the builds and reduce the disk usage. We still want to eliminate the need for the UNDEF variable and fix the problems with weak symbols that this PR addresses.

gebart:
Opened #10200 for the pthread_tls link fix

jcarrano:
@gebart Why is `ld -r` preferred over `--whole-archive` (maybe it is answered in this thread, but there are 100+ comments).

>  using thin archives in the intermediate steps can speed up the builds and reduce the disk usage

I remember reading something like that (using thin archives for intermediates) on a linux kernel mailing list.

gebart:
@jcarrano 
> @gebart Why is `ld -r` preferred over `--whole-archive` (maybe it is answered in this thread, but there are 100+ comments).

I can't remember now if there is a reason, only that the current linking process without --whole-archive does require workarounds and hacks for `__attribute__((used))` variables (#9105) and for overriding weak symbols (`UNDEF += syscalls.o`). In my opinion, the build would be more reliable if we can eliminate the need for UNDEF and that weak symbols are always overridden by strong symbols if there is one, even when nothing else in that module is being referenced.

> 
> > using thin archives in the intermediate steps can speed up the builds and reduce the disk usage
> 
> I remember reading something like that (using thin archives for intermediates) on a linux kernel mailing list.

I think you are referring to https://lkml.org/lkml/2017/6/2/320

jcarrano:
Open question: How can this feature interact with LTO?

cladmi:
I was the first to ask about "why not `ld -r`" because I saw the kernel was doing it, and because I never heard of `--whole-archive` so was wondering if we would be the only one to use it.
Another reason, is that creating one object for each module was going in the direction on starting to think about sending objects to update only one module like @kYc0o was doing with contiki `elfloader` a long time ago.
And having a big archive compared to having a big object file did not change anything.

But now if the `thin archive` can save a lot of space, creating a big object file would be a regression.
So maybe we should avoid it.
And until we find if re-linking is better or worst than archive, we could use `--whole-archive` as in the first PR.
And maybe later think again about having a rule to create relinked objects from the archive if some developers need them and maybe use it instead of archives if it has benefits.

Whatever the solution we take `--whole-archive` or `ld -r` it still requires fixing the same things in the build system anyway which is currently the blocking part.

@gebart If you have other commands to debug size difference than the `objdump -t` I would gladly take them :)

jcarrano:
> starting to think about sending objects to update only one module

Once we have XFA that kind of partial update will be tricky because the object will not necessarily be a contiguous piece of memory.

I think that `ld -r` can be an interesting target for people wanting to distribute binary-only libraries.

cladmi:
@jcarrano not 100%, it will mean that some of the operations will need to have a "register at runtime".
By using xfa we will still make all the default one in the same array at compile time, and the ones that must be registered later will be in a list. So more memory is saved for those at compile time.

It is already somehow the case for the shell for example, where some are added during `shell_init` it is still "compile time" but only registered when starting and all the other commands are put together in a single array.

kaspar030:
It is also possible to drop both ```ld -r``` and using archives altogether and just pass all object files in the final linking step.

jcarrano:
> just pass all object files in the final linking step.

It is convenient from the point view of the build system to have the target of a module be a single file. That way the code that is building the application does not have to know anything about the internals of a module, only that it produces a single archive or object with the same name as the module.

cladmi:
I start a rebase of the branch with history rewrites: The original one was https://github.com/RIOT-OS/RIOT/commit/ae1b3c327d0ccf5ddb0ce8a0d81fa7862d4a3269
* squash of the 'revert'
* autosquash of commits
* Squash of commits
* Rebase on master

cladmi:
The history re-order only rewrites have been done, https://github.com/RIOT-OS/RIOT/compare/https://github.com/RIOT-OS/RIOT/commit/ae1b3c327d0ccf5ddb0ce8a0d81fa7862d4a3269..ca2edb6ddb9df1a85974174e0c513194545d7659

I changed the order to define `LIBS` first before using it. Then start to define variables using it and making a difference between non libs and libs. I did not touch the files at all but added some TODOs I want to do in the commit message.

I will now rebase on `riot/master`. This will change the files content.

kaspar030:
Seems like "ld -r" is incompatible with LTO: https://lwn.net/Articles/744507/

jcarrano:
> Seems like "ld -r" is incompatible with LTO:

If that's the case, we may have to go with --whole-archive, which is the route I prefer, as it is less invasive.

PR #5429, opened 2016-05-10
periph: hwcrypto: initial draft for hardware crypto support

basilfx:
TL;DR; draft for hardware crypto driver interface, request for comments or close it :-)

Some MCUs contain hardware peripherals to speed-up cryptographic operations (ciphers + hashes). While RIOT-OS has support for several ciphers and hashes, there is no hardware support. The crypto API needs an overhaul (see #4798) and I think this PR would help to redesign it. Then a new crypto API could use this to replace software implementations if desired (similar to how xtimer covers the timer peripheral).

This PR proposes a new driver interface (`hwcrypto`, in line with `hwrng`) that should abstract the crypto peripherals across multiple MCUs. Some design considerations:
- Multiple ciphers and hashes may be supported, but defining a single encrypt/decrypt/update/finalize per cipher or hash seems overrated. A single encrypt/decrypt + update/finalize should cover them all.
- Peripherals may have state. For instance, using AES, one may load the key directly into the peripheral register instead of storing it in a context. Due to this, locking must be used to ensure mutual exclusive access.
- To save power, crypto peripherals can be turned off and turned on when needed (this could invalidate a context/state of a peripheral).
- An encrypt or decrypt method may support larger block sizes, so it can process larger blocks of data without CPU intervention.

I implemented the current specification for the EFM32 Gemstone MCU (unpolished code): https://gist.github.com/basilfx/bb59ce9b14a81e94e3836960373c6c8a

Commit 05511e0 contains the specification, 7b05094 is an example application comparing the software versus hardware implementations. On my STK3401A (EFM32), this gives me the following result (AES is ~4x faster, SHA256 ~3x faster):

```
main(): This is RIOT! (Version: 2016.07-devel-110-g4a133-MacBook-Pro-van-Bas.local)
Starting benchmarks.
CRYPTO 3DES: 2786 ops in 10 seconds.
CRYPTO AES128: 225474 ops in 10 seconds.
CRYPTO TWOFISH: 12332 ops in 10 seconds.
CRYPTO SHA256: 123354 ops in 10 seconds.
HWCRYPTO AES128: 1431591 ops in 10 seconds.
HWCRYPTO AES256: 1247614 ops in 10 seconds.
HWCRYPTO SHA1: 354406 ops in 10 seconds.
HWCRYPTO SHA256: 346080 ops in 10 seconds.
Benchmarks done!
```

I still have some design questions:
- Some peripherals need aligned buffers (also for performance). How to enforce?
- What about cipher modes, should software take care?
[...]

PR #8715, opened 2018-03-01
[RFC] fs: add CryptoFS

vincent-d:
### Contribution description

New episode in the show: "Let's play with file systems" :D

This is a PoC of a file-level encryption file system.
This is not a real file system as it relies on vfs to actually store, but it's more a vfs-enable pseudo file system, which uses AES128 to encrypt/decrypt files on the fly.

It's not meant to be merged as-is, it's probably not that secured, especially without a secured way to store the key, but that was fun to try.
Performances are far from good on real hardware, but I guess AES comes with a cost...

### Issues/PRs references

Based on #8559 for the example.

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -710,6 +710,12 @@ ifneq (,$(filter fatfs_vfs,$(USEMODULE)))
   USEMODULE += vfs
 endif
 
+ifneq (,$(filter cryptofs,$(USEMODULE)))
+  USEMODULE += vfs
+  USEMODULE += hashes
+  USEMODULE += crypto
+endif
+
 ifneq (,$(filter benchmark,$(USEMODULE)))
   USEMODULE += xtimer
 endif
--- a/examples/filesystem/Makefile
+++ b/examples/filesystem/Makefile
@@ -44,9 +44,12 @@ USEMODULE += littlefs
 # USEMODULE += fatfs
 USEMODULE += constfs
 # USEMODULE += devfs
+USEMODULE += cryptofs
 
 # Set file systems specific variables
-ifneq (,$(filter littlefs, $(USEMODULE)))
+ifneq (,$(filter cryptofs, $(USEMODULE)))
+    CFLAGS += -DVFS_FILE_BUFFER_SIZE=92 -DVFS_DIR_BUFFER_SIZE=44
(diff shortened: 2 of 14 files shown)


basilfx:
> Performances are far from good on real hardware, but I guess AES comes with a cost...

Another reason for me to fix/continue #5429 :-)

vincent-d:
Tested with #5429 and #9152 and the performances are not significantly better when encrypting. I could improve the overall performance by removing a lot of unused seeks, so it's quite usable now ;)

If someone thinks it's a good addition (and the design is good enough) (s)he can take a look and test it :)

tcschmidt:
What would be a use case of a (symmetrically encrypted) file system on a RIOT device? Where would this fs reside on? What attack surface shall this shield?
@vincent-d Can you elaborate a bit upon this?

vincent-d:
@tcschmidt my use case is I have a SPI NOR flash. On this flash I'll store files that I want to ensure they are not modified nor can be read by an attacker using the SPI directly. So encrypting them and using a hash is one way to solve this issue. We of course need an underlying file system to actually manage the flash (in my case it's littlefs). In my case the key can reside on a secure element and be only loaded in RAM, but storing it in the internal flash of the microcontroller might also be enough.

This is definitely not a bullet-proof encrypted file system and as stated before this is more a proof of concept.
We might at least improve the encryption mode (counter mode for instance) or use another cipher, but it could come in a v2.

tcschmidt:
Thanks @vincent-d - I'm wondering, if you use symmetric crypto with a key available at the node, why would you want to protect against an attacker that has hardware-access to the same node?

vincent-d:
Accessing an external SPI flash is easier than accessing the internal flash of the microcontroller. And in my case I can store the key in a secured way (secure element).

Using asymetric crypto woudn't add more guarantee anyway, as we would have to store the private key as we do for the symetric key.

tcschmidt:
> Using asymetric crypto woudn't add more guarantee anyway, as we would have to store the private key as we do for the symetric key.

This depends on the use case. Anyway, I wasn't asking for asymmetric crypto, just about the meaning of this PR. Maybe a clear use case description (and limitation) would be helpful to prevent unqualified deployment.

PR #8718, opened 2018-03-01
cpu/efm32/periph/timer: remove timer_set duplicate

cladmi:
timer_set is already provided by 'periph_common' so does not need to be
provided by cpu/periph.

The build system currently silently ignores this.

### Issues/PRs references

Its required for https://github.com/RIOT-OS/RIOT/pull/8711

It was also disabled by https://github.com/RIOT-OS/RIOT/pull/5757 in `periph_conf.h` but the function can be removed as its identical to `periph_common` https://github.com/RIOT-OS/RIOT/blob/2018.01/drivers/periph_common/timer.c#L24.

### DIFF
--- a/cpu/efm32/periph/timer.c
+++ b/cpu/efm32/periph/timer.c
@@ -98,11 +98,6 @@ int timer_init(tim_t dev, unsigned long freq, timer_cb_t callback, void *arg)
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


cladmi:
I have not tested that the board still works. Only compiled.

cladmi:
Thanks you.

PR #8720, opened 2018-03-01
cpu/mips32r2_common: use `periph_common` timer_set

cladmi:
Use 'timer_set' provided by 'periph_common' instead of re-implementing it.

It also fixes 'mips32r2_generic' that should have been defining
'PERIPH_TIMER_PROVIDES_SET' as 'mips32r2_common' uses 'periph_common' but is not
necessary if removing the function.

### Issues/PRs references

Its required for #8711

It was also disabled by #5757 in periph_conf.h but the function can be removed as it has the same behavior as periph_common https://github.com/RIOT-OS/RIOT/blob/2018.01/drivers/periph_common/timer.c#L24.

### DIFF
--- a/cpu/mips32r2_common/periph/timer.c
+++ b/cpu/mips32r2_common/periph/timer.c
@@ -121,23 +121,6 @@ int timer_init(tim_t dev, unsigned long freq, timer_cb_t cb, void *arg)
     return 0;
 }
 
-int timer_set(tim_t dev, int channel, unsigned int timeout)
-{
-    assert(dev == 0);
-    assert(channel < CHANNELS);
-
-    (void)dev;
-
-    timeout >>= TIMER_ACCURACY_SHIFT;
-    timeout <<= TIMER_ACCURACY_SHIFT;
-
-    uint32_t status = irq_disable();
-    compares[channel] = counter + timeout;
-    irq_restore(status);
-
-    return channel;
-}
-
 int timer_set_absolute(tim_t dev, int channel, unsigned int value)
 {
     assert(dev == 0);
--- a/cpu/mips32r2_common/periph/timer.c
+++ b/cpu/mips32r2_common/periph/timer.c
@@ -135,7 +135,7 @@ int timer_set_absolute(tim_t dev, int channel, unsigned int value)
     compares[channel] = value;
     irq_restore(status);
 
-    return channel;
+    return 1;
 }
 
 int timer_clear(tim_t dev, int channel)
@@ -149,7 +149,7 @@ int timer_clear(tim_t dev, int channel)
     compares[channel] = 0;
     irq_restore(status);
 
-    return channel;
+    return 1;
 }
 
 unsigned int timer_read(tim_t dev)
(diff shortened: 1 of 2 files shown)


cladmi:
I have not tested that the board still works. And I will let Murdock check the builds.

cladmi:
Another option to solve this could be to define `PERIPH_TIMER_PROVIDES_SET` for `mips32r2_generic`.

neiljay:
Is the generic timer_set always called with interrupts disabled ?

> int timer_set(tim_t dev, int channel, unsigned int timeout)
> {
>     return timer_set_absolute(dev, channel, timer_read(dev) + timeout);
> }

if you took a high latency interrupt (or even worse rescheduled) between the timer_read and the timer_set_absolute your timer is not going to be very accurate ?

cladmi:
@neiljay I did not think about that, I just copied what was done for other cpus. I could change it to defining PERIPH_TIMER_PROVIDES_SET in mips32r2_generic `periph_conf.h` if you prefer.

neiljay:
@cladmi maybe fix the generic implementation to disable interrupts as well first ?

int timer_set(tim_t dev, int channel, unsigned int timeout)
{
    uint32_t status = irq_disable();
    int ret = timer_set_absolute(dev, channel, timer_read(dev) + timeout);
    irq_restore(status);
    return ret 
}

gebart:
I agree, it's better to fix the generic implementation

cladmi:
The general fix should be done in another PR then. This one can wait until its done.
I could do something next week.

cladmi:
Thinking again about this. Would it really help to put this `irq_disable` ?
I think, it does not really change the final behavior.
Let me explain my thoughts with a `timer_set(5 seconds)` and a 3 seconds re-schedule (to give numbers).

Without `irq_disable`

* if re-scheduled between `timer_get` and `timer_set_absolute`, `timer` get is evaluated before re-scheduling, so the callback is triggered after 5 seconds since the function call.
* If re-scheduled before `timer_get`, `timer_get` will be done 3 seconds after and the callback is after 8 seconds in total.

With `irq_disable`

* In the `irq_disable` section, it cannot be re-scheduled so its a 5 seconds in total.
* If re-scheduled before `irq_disable`, `timer_get` will also be done 3 seconds after and the callback is after 8 seconds in total.

So adding `irq_disable` only protects a few cycles more. Your precision depend on how long can a high priority task take the cpu anyway.

neiljay:
If the de-sceduled / high latency period is bigger than the set period, the timer may never fire at all (or atleast until it wraps).

swapping you example so we want a timer to fire in 3s but he reschedule takes 5.

time(s)		fn
0		timer_get()
<reschedule>
5		time_set_absolute(0+3)

5 > 3 so time never fires until the timer wraps.

Neil

cladmi:
I proposed another PR that just fixes the double definition to allow merging it now.
This way it keeps the current behavior but disables defining `timer_set` in `periph_common`.

cladmi:
Closed in favor of #8817

PR #6995, opened 2017-05-02
cpu+board: Add support for NXP Kinetis KW41Z and FRDM-KW41Z development board

gebart:
Port RIOT to the recently released NXP FRDM-KW41Z development boards.

Currently missing support for ~~the radio and~~ the onboard NOR flash.
I2C needs a rewrite, this CPU has one i2c module clocked from the bus clock, and one clocked by the system clock. The I2C rewrite will be on hold until the I2C API is reworked, so for now, one of the two available I2C buses will be clocked at the wrong speed, this can be worked around in the application by requesting a different speed or modifying the configuration to fit.

A radio transceiver driver is available in #7107  

depends on ~~#7362~~ ~~#6994~~ ~~#6978~~ ~~#6916~~ ~~#6993~~ ~~#7434~~ ~~#7882~~ ~~#8560~~

PR #8724, opened 2018-03-02
kinetis: Set LPUART clock source during uart_lpuart_init

gebart:
### Contribution description

The Kinetis LPUART hardware module can choose from multiple clock sources, to use the module the user needs to select a clock source by setting a bitfield in the SIM_SOPT2 register. This PR moves this configuration into the periph uart driver, to reduce the code duplication in board_init between different boards and to make it easier to switch UART settings on an existing board.
The only existing board which has a LPUART module is the FRDM-K22F, but the module is not used in the default configuration, so this PR is mainly in preparation for the FRDM-KW41Z in #6995 which only has LPUART type UART hardware modules.

### Issues/PRs references

Split from #6995 
Required by #6995 for functional UART.


cpu/kinetis/include/cpu_conf_kinetis.h
```diff
@@ -72,6 +72,11 @@ extern "C"
 #if !defined(OSC0) && defined(OSC)
 #define OSC0 OSC
 #endif
+#if !defined(SIM_SOPT2_LPUART0SRC_MASK) && defined(SIM_SOPT2_LPUARTSRC_MASK)
+#define SIM_SOPT2_LPUART0SRC_MASK SIM_SOPT2_LPUARTSRC_MASK
+#define SIM_SOPT2_LPUART0SRC_SHIFT SIM_SOPT2_LPUARTSRC_SHIFT
+#define SIM_SOPT2_LPUART0SRC SIM_SOPT2_LPUARTSRC
+#endif
```
kYc0o: Can these defines be aligned? Just to see that they're actually defining a value.

cpu/kinetis/periph/uart.c
```diff
@@ -296,7 +305,19 @@ static inline void uart_init_lpuart(uart_t uart, uint32_t baudrate)
     LPUART_Type *dev = uart_config[uart].dev;
     uint32_t clk = uart_config[uart].freq;
 
-    /* Remember to select a module clock in board_init! (SIM->SOPT2[LPUART0SRC]) */
+    /* Set LPUART clock source */
+#ifdef SIM_SOPT2_LPUART0SRC
```
kYc0o: The value of this definition doesn't matter? It only needs to be defined to execute the code below? It's just because above there is:

```c
#define SIM_SOPT2_LPUART0SRC SIM_SOPT2_LPUARTSRC
```

thus the value might be needed or evaluated.

cpu/kinetis/periph/uart.c
```diff
@@ -63,6 +63,15 @@
 #define LPUART_OVERSAMPLING_RATE (16)
 #endif
 
+#ifndef LPUART_0_SRC
+/* Default LPUART clock setting to avoid compilation failures, define this in
+ * periph_conf.h to set board specific configuration if using the LPUART. */
```
kYc0o: Maybe this comment can be above the definition?

cpu/kinetis/periph/uart.c
```diff
@@ -296,7 +305,19 @@ static inline void uart_init_lpuart(uart_t uart, uint32_t baudrate)
...
 
-    /* Remember to select a module clock in board_init! (SIM->SOPT2[LPUART0SRC]) */
+    /* Set LPUART clock source */
+#ifdef SIM_SOPT2_LPUART0SRC
+    if (dev == LPUART0) {
+        SIM->SOPT2 = (SIM->SOPT2 & ~SIM_SOPT2_LPUART0SRC_MASK) |
+            SIM_SOPT2_LPUART0SRC(LPUART_0_SRC);
+    }
+#endif
+#ifdef SIM_SOPT2_LPUART1SRC
```
kYc0o: Though I see a define for `SIM_SOPT2_LPUART0SRC`, I can't find for `SIM_SOPT2_LPUART1SRC`. Is this intentional? Same for `SIM_SOPT2_LPUART1SRC_MASK` used below, since there is a definition for `SIM_SOPT2_LPUART0SRC_MASK`.

cpu/kinetis/include/cpu_conf_kinetis.h
gebart: Will fix

cpu/kinetis/periph/uart.c
gebart: They are defined by the vendor and put headers for the CPUs which have the appropriate hardware module. The definition in cpu_conf_kinetis.h is only a compatibility definition for some CPU headers which have named the bit field LPUARTSRC instead of LPUART0SRC. See the other definitions in the same area in cpu_conf_kinetis.h.

cpu/kinetis/periph/uart.c
gebart: LPUART1SRC only exist for CPUs which have two or more lpuart modules (LPUART0, LPUART1). Therefore we need the preprocessor conditional to avoid compilation errors on the CPUs which don't have these modules.

gebart:
@kYc0o aligned definitions in cpu_conf_kinetis.h

cpu/kinetis/periph/uart.c
gebart: addressed

cpu/kinetis/periph/uart.c
kYc0o: Ok I got it. Thanks!

cpu/kinetis/periph/uart.c
kYc0o: Ok I see, that's right for me.

kYc0o:
Perfect, I think everything is in order here, so you may squash and merge once everything is green.

gebart:
@kYc0o squashed, missing ACK

PR #8725, opened 2018-03-02
kinetis: Move errata workarounds to separate function

gebart:
### Contribution description

Clean up Kinetis implementation by moving hardware errata workarounds to a separate function which will be called during boot.

The RSIM errata fix was not applied correctly in all clock configurations, by moving it to the boot it will be applied always.

### Issues/PRs references

split from #6995


kYc0o:
Tested on frdm-k64f, mulle and pba-d-01-kw2x.

gebart:
@kYc0o Thanks for testing

PR #8345, opened 2018-01-10
cc2538/include/periph_cpu.h: Changed UART configuration format for cc2538.

danpetry:
### Contribution description

Changed the style of the UART configuration for different boards, from
a define based configuration to one based on an array of structs, one
struct for each UART, with the format of the struct defined in
cc2538/include/periph_cpu.h.

  - Defined the fields of the struct in periph_cpu.h
  - Removed the compilation includes that were in uart.c for each UART
  - Implemented a generic ISR subroutine for clarity
  - Implemented the configuration on openmote-cc2538 and cc2538dk boards.

Notes to reviewer:

  - Openmote has been tested, but cc2538dk has not, due to lack of the
    latter board.
  - Some information about UART1 on the cc2538dk has been deleted, because
    UART1 was not being used. This information would have gone in the
    second index of the config array - but presence of this second index
    would have enabled UART1, in the new style of configuration. The
    same applies for the flow control for UART0 on the cc2538dk; flow
    control is only available on UART1 for the cc2538 cpu.
  - The switch on the UART number has been deleted. The default statement
    offered protection on the "uart" variable being outside the range allowed
    by the cpu. This protection is no longer present. What effect might this
    have? Suggest that the uart_t type should be enumerated on a per-cpu
    basis, to restrict the range, but would this be worth the effort to change?

### Issues/PRs references

Ticks the UART checkbox for cc2538 in https://github.com/RIOT-OS/RIOT/issues/7941.

PR #8726, opened 2018-03-02
boards: add UART1 config for Zolertia Remotes

smlng:
### Contribution description

This PR adds UART1 to periph conf for all Zolertia remotes, see pinout in docu [here](https://github.com/Zolertia/Resources/wiki/RE-Mote#re-mote-revision-a-pin-out). It also adapts UART config to use port names instead of "meaningless" numbers.

### Issues/PRs references

#8345

### DIFF
--- a/boards/common/remote/include/periph_common.h
+++ b/boards/common/remote/include/periph_common.h
@@ -84,11 +84,19 @@ static const uart_conf_t uart_config[] = {
         .tx_pin   = GPIO_PIN(0, 1),
         .cts_pin  = GPIO_UNDEF,
         .rts_pin  = GPIO_UNDEF
+    },
+    {
+        .dev      = UART1_BASEADDR,
+        .rx_pin   = GPIO_PIN(2, 1),
+        .tx_pin   = GPIO_PIN(2, 0),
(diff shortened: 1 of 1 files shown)


boards/common/remote/include/periph_common.h
```diff
@@ -78,17 +78,26 @@ static const timer_conf_t timer_config[] = {
  * @{
  */
 static const uart_conf_t uart_config[] = {
+    /* UART0 is mapped to debug usb */
     {
         .dev      = UART0_BASEADDR,
-        .rx_pin   = GPIO_PIN(0, 0),
-        .tx_pin   = GPIO_PIN(0, 1),
+        .rx_pin   = GPIO_PIN(PORT_A, 0),
```
PeterKietzmann: In the past I learned that people get confused when they look at a `periph_conf` file and (CPU specific) expressions such as `PORT_A` are used which won't work in a generic way from the application code. Obviously, using literals is nicer in general. What's your opinion on that?

A-Paul: > [...] Obviously, using literals is nicer in general. [...]

I strongly object to this proposition.

PR #8727, opened 2018-03-02
tools/ci: enhance print_toolchain script

smlng:
### Contribution description

this PR enhances the tool script to print versions of installed compilers, devel libs and tools:

    - add generic command version printer
    - add cmake version
    - enhance output structure

### Issues/PRs references

None

### DIFF
--- a/dist/tools/ci/print_toolchain_versions.sh
+++ b/dist/tools/ci/print_toolchain_versions.sh
@@ -1,19 +1,20 @@
 #!/bin/sh
 
-gcc_version() {
-    local cc
+get_cmd_version() {
     if [ -z "$1" ]; then
-        cc=gcc
+        return
+    fi
+
+    local cmd="$1"
+    if command -v "$cmd" 2>&1 >/dev/null; then
+        ver=$("$cmd" --version 2> /dev/null | head -n 1)
+        if [ -z "$ver" ]; then
+            ver="error"
+        fi
     else
-        cc=$1-gcc
+        ver="missing"
     fi
-    local ver=
-    if command -v "$cc" 2>&1 >/dev/null; then
-        ver=$("$cc" --version | head -n 1)
-    fi
-    if [ -z "$ver" ]; then
-        ver=missing/error
-    fi
+
     printf "%s" "$ver"
 }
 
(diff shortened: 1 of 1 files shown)


dist/tools/ci/print_toolchain_versions.sh
```diff
@@ -1,19 +1,17 @@
...
-    local ver=
-    if command -v "$cc" 2>&1 >/dev/null; then
-        ver=$("$cc" --version | head -n 1)
-    fi
-    if [ -z "$ver" ]; then
-        ver=missing/error
+
+    local cmd="$1"
+    if command -v "$cmd" 2>&1 >/dev/null; then
+        ver=$("$cmd" --version | head -n 1)
```
gebart: Please add
```sh
if [ -z "$ver" ]; then
    ver = error
fi
```
to detect some other errors

smlng:
> newlib_version and avr_libc_version can be simplified in the same way 

not quite, at least how they use the `get_define` function differs.

gebart:
@smlng I meant change the argument to avr_libc_version into "avr-gcc" and newlib_version argument to "${p}-gcc"

smlng:
ah, now I get it ... yeah will do

dist/tools/ci/print_toolchain_versions.sh
```diff
@@ -24,83 +25,52 @@ get_define() {
...
     printf "%s" "$line"
 }
 
 newlib_version() {
     if [ -z "$1" ]; then
-        cc=gcc
-    else
-        cc=$1-gcc
+        printf "%s" "error"
     fi
```
gebart: Missing return above

dist/tools/ci/print_toolchain_versions.sh
```diff
@@ -24,83 +25,52 @@ get_define() {
...
-}
-
-git_version() {
-    local cmd="git"
-    if command -v "$cmd" 2>&1 >/dev/null; then
-        ver=$("$cmd" --version | head -n 1)
-    else
-        ver="missing"
+        printf "%s" "error"
     fi
```
gebart: Missing return above

smlng:
@gebart can you have another look, I amended changes as suggested and squashed.

PR #8732, opened 2018-03-03
cortexm_common: don't try to set MEMFAULTENA on ARMv6-M

girtsf:
Before this change, if one tried to build a Cortex-M0+ target that had
an MPU, compilation would fail due to missing
`SCB_SHCSR_MEMFAULTENA_Msk` in SCB structure. Cortex-M0+ is a ARMv6-M
arch (unlike most other targets that have MPU support). ARMv6-M has more
limited support for fault conditions, see ARMv6-M Architecture Reference
Manual, D3.6.2.

### DIFF
--- a/cpu/cortexm_common/mpu.c
+++ b/cpu/cortexm_common/mpu.c
@@ -36,8 +36,13 @@ int mpu_enable(void) {
 #if __MPU_PRESENT
     MPU->CTRL |= MPU_CTRL_PRIVDEFENA_Msk | MPU_CTRL_ENABLE_Msk;
 
-    /* Enable the memory fault exception */
+#ifdef SCB_SHCSR_MEMFAULTENA_Msk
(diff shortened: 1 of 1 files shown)


cpu/cortexm_common/mpu.c
```diff
@@ -36,8 +36,11 @@ int mpu_enable(void) {
 #if __MPU_PRESENT
     MPU->CTRL |= MPU_CTRL_PRIVDEFENA_Msk | MPU_CTRL_ENABLE_Msk;
 
-    /* Enable the memory fault exception */
+#if !defined(CPU_ARCH_CORTEX_M0PLUS)
```
gebart: I think it is simpler to just protect this with `#ifdef SCB_SHCSR_MEMFAULTENA_Msk`

girtsf:
Updated to use an `#ifdef`.

hexluthor:
Looks good to me, thanks. :+1:

PR #8739, opened 2018-03-05
cpu: stm32_common: always enable PWR module

kaspar030:
### Contribution description

STM32F[024] don't enable the clock domain necessary to set deep sleep options, basically disabling most low-power features.

~~This PR adds an optional pm_init() call in periph_init(), and uses that facility to activate PWR's clock.~~

This PR now just enables PWR on all stm32.

### DIFF
--- a/cpu/stm32_common/cpu_init.c
+++ b/cpu/stm32_common/cpu_init.c
@@ -30,12 +30,21 @@
 
 #include "cpu.h"
 #include "stmclk.h"
+#include "periph_cpu.h"
 #include "periph/init.h"
+
+#if defined (CPU_FAM_STM32L4)
+#define BIT_APB_PWREN       RCC_APB1ENR1_PWREN
+#else
+#define BIT_APB_PWREN       RCC_APB1ENR_PWREN
+#endif
 
 void cpu_init(void)
 {
     /* initialize the Cortex-M core */
     cortexm_init();
+    /* enable PWR module */
+    periph_clk_en(APB1, BIT_APB_PWREN);
     /* initialize the system clock as configured in the periph_conf.h */
     stmclk_init_sysclk();
     /* trigger static peripheral initialization */
--- a/cpu/stm32f1/periph/rtt.c
+++ b/cpu/stm32f1/periph/rtt.c
@@ -151,7 +151,7 @@ void rtt_clear_alarm(void)
 
 void rtt_poweron(void)
 {
-    periph_clk_en(APB1, (RCC_APB1ENR_BKPEN|RCC_APB1ENR_PWREN)); /* enable BKP and PWR, Clock */
+    periph_clk_en(APB1, RCC_APB1ENR_BKPEN);  /* enable BKP, Clock */
     /* RTC clock source configuration */
     PWR->CR |= PWR_CR_DBP;                   /* Allow access to BKP Domain */
     RCC->BDCR |= RCC_BDCR_LSEON;             /* Enable LSE OSC */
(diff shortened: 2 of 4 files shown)


cpu/stm32_common/periph/pm.c
```diff
@@ -86,4 +86,9 @@ void pm_off(void)
     irq_disable();
     pm_set(0);
 }
+
+void pm_init(void)
+{
+    periph_clk_en(APB1, RCC_APB1ENR_PWREN);
```
aabadie: Maybe add a comment saying why this is needed ?

vincent-d:
This looks legit, but I'm not sure this is as simple as that. 
In `stmclk_common.c`, there are these two functions:
```C
void stmclk_dbp_unlock(void)
{
    periph_clk_en(APB1, BIT_APB_PWREN);
    PWR->REG_PWR_CR |= BIT_CR_DBP;
}

void stmclk_dbp_lock(void)
{
    PWR->REG_PWR_CR &= ~(BIT_CR_DBP);
    periph_clk_dis(APB1, BIT_APB_PWREN);
}
```

These functions are called in `rtc_init()`, `rtc_poweron()` and `rtc_poweroff()`. So if one turns off the RTC, the clock will be disabled.

kaspar030:
> These functions are called in rtc_init(), rtc_poweron() and rtc_poweroff(). So if one turns off the RTC, the clock will be disabled.

True... And we don't keep track on which clocks are used by whom...

I see two ways here:

- enable the clock for PWR in every call to pm_set()
- guard the calls in stmclk_dbp_lock/unlock with ```#ifdef MODULE_PERIPH_PM```

what do you think?

vincent-d:
> And we don't keep track on which clocks are used by whom...

I agree with that.

I'm not sure why it's enabled/disabled in `stmclk_dbp_(un)lock`. So I would even go simpler and just not disable the clock in `stmclk_dbp_lock` and enable  it in both `stmclk_dbp_unlock` and `pm_init`.

kaspar030:
> So I would even go simpler and just not disable the clock in stmclk_dbp_lock and enable it in both stmclk_dbp_unlock and pm_init.

Now that there's a common cpu_init(), I changed the PR to just globally and always enable PWR's clock.

I tried measuring the impact. In a ```while(1);```, an f401 is using 12125uA with PWR disabled, and 12126uA with PWR enabled, with the last digit quite unstable, measured on nucleo-f401 with an eevblog 121gw.

cpu/stm32_common/cpu_init.c
```diff
@@ -32,10 +32,18 @@
...
+#else
+#define BIT_APB_PWREN       RCC_APB1ENR_PWREN
+#endif
+
 void cpu_init(void)
 {
     /* initialize the Cortex-M core */
     cortexm_init();
+    /* enable PWR module */
+    periph_clk_en(APB1, BIT_APB_PWREN);
```
vincent-d: Missing `periph_cpu.h` include

cpu/stm32f1/periph/rtt.c
```diff
@@ -151,7 +151,7 @@ void rtt_clear_alarm(void)
 
 void rtt_poweron(void)
 {
-    periph_clk_en(APB1, (RCC_APB1ENR_BKPEN|RCC_APB1ENR_PWREN)); /* enable BKP and PWR, Clock */
+    periph_clk_en(APB1, (RCC_APB1ENR_BKPEN)); /* enable BKP, Clock */
```
vincent-d: Nit-picky and I won't block for that, but you could have removed the parenthesis here and below

cpu/stm32_common/cpu_init.c
kaspar030: jup, fixed

cpu/stm32f1/periph/rtt.c
kaspar030: done

vincent-d:
And go!

PR #8743, opened 2018-03-05
cpu/saml21: fix RTC wrong condition for #if

dylad:
### Contribution description

This PR fixes a wrong #if condition for the RTC driver.

### DIFF
--- a/cpu/saml21/periph/rtc.c
+++ b/cpu/saml21/periph/rtc.c
@@ -22,6 +22,13 @@
 #include "periph/rtc.h"
 #include "periph_conf.h"
 
+/* SAML21 rev B needs an extra bit, which in rev A defaults to 1, but isn't
+ * visible. Thus define it here. */
+#ifndef RTC_MODE2_CTRLA_CLOCKSYNC
+#define RTC_MODE2_CTRLA_CLOCKSYNC_Pos   15
+#define RTC_MODE2_CTRLA_CLOCKSYNC       (0x1ul << RTC_MODE2_CTRLA_CLOCKSYNC_Pos)
+#endif
+
 typedef struct {
     rtc_alarm_cb_t cb;        /**< callback called from RTC interrupt */
     void *arg;                /**< argument passed to the callback */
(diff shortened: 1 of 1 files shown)


kaspar030:
How does this manifest? (How can I test?)

dylad:
This condition was written because SAML21 revision B must set this bit if you want to use `rtc_get_time` function. 
SAML21 revision A don't need this bit (probably because a sillicon bug between this two variants). This issue is descripted in this [document](http://ww1.microchip.com/downloads/en/AppNotes/Atmel-42512-Migrating-from-SAM-L21-Variant-A-to-Variant-B_ApplicationNote_AT13382.pdf)
Since SAML21-XPRO board can be equipped with either rev A (old boards not sold anymore but they are still in the wild) or rev B, this bit must be set for rev B (to use `rtc_get_time` function).
I don't know if we can set this bit with a SAML21 rev A so I decide to keep the #if but just change the condition because the previous one cannot occur as SAML21XXXB does not exist in RIOT.

So if you want to test it, you will need a one SAML21 rev B and  one rev A :)

cpu/saml21/periph/rtc.c
```diff
@@ -97,7 +97,7 @@ void rtc_init(void)
 
     /* RTC config with RTC_MODE2_CTRL_CLKREP = 0 (24h) */
     RTC->MODE2.CTRLA.reg = RTC_MODE2_CTRLA_PRESCALER_DIV1024 |  /* CLK_RTC_CNT = 1KHz / 1024 -> 1Hz */
-#if (SAML21XXXB) || (SAMR30)
+#if (CPU_MODEL_SAML21J18B)
```
kaspar030: How about changing this like in this patch:

https://gist.github.com/9842253687bd2698a9b344779dd83500

E.g., define the bit if not visible, and always set it.
I can confirm that setting the bit on SAML21 rev A doesn't hurt.

dylad: > I can confirm that setting the bit on SAML21 rev A doesn't hurt

Then I'll change it :)

dylad:
@kaspar030 I've pushed the requested changes.

kaspar030:
pls squash!

dylad:
Squashed

PR #8744, opened 2018-03-05
atmega: add last instruction print

Josar:
Add stackpointer print to cpu_print_last_instruction()

Signed-off-by: Josua Arndt  <josuaarndt@live.de>


kYc0o:
Did you tested this in some way? I know that I can intentionally trigger a fault to see it but maybe you have a more concrete example.

Josar:
I have no concrete example, i thought it woulf help me debug. As i got assertion errors.

Just set `CFLAGS += -DDEBUG_ASSERT_VERBOSE` and trigger a `assert(1!=0);`

kYc0o:
However, you need to change the PR description and commit message. Just remove the "...", it can be as simple as "atmega: add last instruction print" for instance.

PR #8747, opened 2018-03-06
cpu/nrf51: add PWM implementation

SemjonKerner:
This is an initial peripheral pwm implementation for nordic nrf51 cpu. This implementation provides all functions in pwm.h. It aims to create little cpu load by utilizing nordics programmable periphal interconnection (ppi) to connect timer events and gpio tasks. Since the cpu is part of the calliope-mini, it also adds defines to the boards peripheral config.

### DIFF
--- a/boards/calliope-mini/include/periph_conf.h
+++ b/boards/calliope-mini/include/periph_conf.h
@@ -57,18 +57,11 @@ static const timer_conf_t timer_config[] = {
         .channels = 3,
         .bitmode  = TIMER_BITMODE_BITMODE_16Bit,
         .irqn     = TIMER1_IRQn
-    },
-    {
-        .dev      = NRF_TIMER2,
-        .channels = 3,
-        .bitmode  = TIMER_BITMODE_BITMODE_16Bit,
-        .irqn     = TIMER2_IRQn
     }
 };
 
 #define TIMER_0_ISR         isr_timer0
 #define TIMER_1_ISR         isr_timer1
-#define TIMER_2_ISR         isr_timer2
 
 #define TIMER_NUMOF         (sizeof(timer_config) / sizeof(timer_config[0]))
 /** @} */
@@ -127,6 +120,15 @@ static const i2c_conf_t i2c_config[] = {
  * @{
  */
 #define RADIO_IRQ_PRIO      1
+/** @} */
+
+/**
+ * @name    PWM configuration
+ * @{
+ */
+#define PWM_NUMOF           (1U)
+#define PWM_TIMER           NRF_TIMER2
+#define PWM_PIN             (0U)
 /** @} */
 
 #ifdef __cplusplus
--- a/cpu/nrf51/periph/pwm.c
+++ b/cpu/nrf51/periph/pwm.c
@@ -1,0 +1,245 @@
+/*
+ * Copyright (C) 2018 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser General
+ * Public License v2.1. See the file LICENSE in the top level directory for more
+ * details.
+ */
+
+/**
+ * @ingroup     cpu_nrf51
+ * @ingroup     drivers_periph_pwm
+ * @{
+ *
+ * @file
+ * @brief       Low-level PWM driver implementation
+ *
+ * @author      Semjon Kerner <semjon.kerner@fu-berlin.de>
+ * @}
+ */
+
+#include <stdint.h>
+#include <string.h>
+
+#include "periph/gpio.h"
+#include "periph/pwm.h"
+
+#define ENABLE_DEBUG        (0)
+#include "debug.h"
+
+#define PWM_PS_MAX          (9U)
+#define PWM_PPI_CHANNELS    ((1 << PWM_PPI_A) | (1 << PWM_PPI_B))
+#ifndef PWM_PERCENT_VAL
+    #define PWM_PERCENT_VAL (1U)
+#endif
+
+static const uint32_t divtable[10] = {
+    (CLOCK_CORECLOCK >> 0),
+    (CLOCK_CORECLOCK >> 1),
+    (CLOCK_CORECLOCK >> 2),
+    (CLOCK_CORECLOCK >> 3),
+    (CLOCK_CORECLOCK >> 4),
+    (CLOCK_CORECLOCK >> 5),
+    (CLOCK_CORECLOCK >> 6),
+    (CLOCK_CORECLOCK >> 7),
+    (CLOCK_CORECLOCK >> 8),
+    (CLOCK_CORECLOCK >> 9),
+};
+
+static uint32_t init_data[2];
+
+uint32_t pwm_init(pwm_t dev, pwm_mode_t mode, uint32_t freq, uint16_t res)
+{
+    assert(dev == 0 && ((mode == PWM_LEFT) || (mode == PWM_RIGHT)));
+
+    /* reset and configure the timer */
+    PWM_TIMER->POWER = 1;
+    PWM_TIMER->TASKS_STOP = 1;
+    PWM_TIMER->BITMODE = TIMER_BITMODE_BITMODE_32Bit;
+    PWM_TIMER->MODE = TIMER_MODE_MODE_Timer;
+    PWM_TIMER->TASKS_CLEAR = 1;
+
+    /* calculate and set prescaler */
+    uint32_t timer_freq = freq * res;
+    uint32_t lower = (timer_freq - (PWM_PERCENT_VAL * (timer_freq / 100)));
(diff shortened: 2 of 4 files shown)


boards/calliope-mini/include/periph_conf.h
```diff
@@ -36,7 +36,7 @@ extern "C" {
 #define CLOCK_HFCLK         (16U)           /* set to  0: internal RC oscillator
                                                       16: 16MHz crystal
                                                       32: 32MHz crystal */
-#define CLOCK_LFCLK         (0)             /* set to  0: internal RC oscillator
+#define CLOCK_LFCLK         (0U)             /* set to  0: internal RC oscillator
```
haukepetersen: I would prefer to move the changes unrelated to the actual PR (marking numbers as unsigned) into at least their own commit, if not their own PR...

boards/calliope-mini/include/periph_conf.h
```diff
@@ -78,9 +71,9 @@ static const timer_conf_t timer_config[] = {
  * @{
  */
 #define RTT_NUMOF           (1U)
-#define RTT_DEV             (1)             /* NRF_RTC1 */
-#define RTT_MAX_VALUE       (0x00ffffff)
-#define RTT_FREQUENCY       (1024)
+#define RTT_DEV             (1U)             /* NRF_RTC1 */
+#define RTT_MAX_VALUE       (0x00ffffffU)
```
haukepetersen: not sure, but IMHO all hex numbers are implicitly `unsigned`, so no need for the `U` in this case.

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,224 @@
...
+
+#include <stdint.h>
+#include <string.h>
+#include <math.h>
+
+#include "periph/gpio.h"
+#include "periph/pwm.h"
+#include "vectors_cortexm.h"
+
+#define ENABLE_DEBUG                (1)
```
haukepetersen: should be disabled, right?!

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,224 @@
...
+#include "debug.h"
+#define DEBUG_ASSERT_VERBOSE        (1U)
```
haukepetersen: remove this line

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,224 @@
...
+
+#define NRF51_PWM_PRESCALER_MIN     (0U)
```
haukepetersen: as these are local defines, I usually don't prefix them with the full module names. IMHO this makes the code better to understand, as it is clear which defines are local and which ones come from a global context. But this might be subjective?!

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,224 @@
...
+#define NRF51_PWM_PRESCALER_MAX     (10U)
+#define NRF51_PWM_PPI_CHANNELS      (((1U) << PWM_PPI_A) | ((1U) << PWM_PPI_B))
+#ifndef NRF51_PWM_PERCENT_VAL
+    #define NRF51_PWM_PERCENT_VAL   (1U)
+#endif
+
+inline uint32_t apply_prescaler(uint8_t prescaler)
```
haukepetersen: use `static`.

And as I recently learned: no need to declare this `inline`, as the compiler will take care of this anyway.

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,224 @@
...
+{
+    return 16000000 >> prescaler;
+}
+
+/*
+ * compare requested frequency with available frequency
+ * has a margin of PERCENT_VAL
+ */
+bool is_close_to(uint32_t f_requ, uint32_t f_avail)
```
haukepetersen: `static`...

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,224 @@
...
+{
+    assert (dev == 0 && ((mode == PWM_LEFT) || (mode == PWM_RIGHT)));
+
+    /* reset and configure the timer */
+    PWM_TIMER->TASKS_STOP = 1;
+    PWM_TIMER->BITMODE = TIMER_BITMODE_BITMODE_32Bit;
+    PWM_TIMER->MODE = TIMER_MODE_MODE_Timer;
+    PWM_TIMER->TASKS_CLEAR = 1;
+
+    /* calculate and set prescaler */
```
haukepetersen: looking at this code and thinking about the problem once more, I think we can make this quite more efficient:
```c
// on top of this file define
static const uint32_t divtable[11] = {
    (CLOCK_CPUCLOCK >> 0),
    (CLOCK_CPUCLOCK >> 1),
    ...
    (CLOCK_CPUCLOCK >> 10),
};

// here:
uint32_t lower = (freq - (freq / 100 * NRF51_PWM_PERCENT_VAL));
uint32_t upper = (freq + (freq / 100 * NRF51_PWM_PERCENT_VAL));
for (uint32_t ps = 0; ps <= (PS_MAX + 1); ps++) {
    if (ps == (PW_MAX + 1)) {
        DEBUG("[pwm] init error: resolution or frequency not supported\n");
        return 0;
    }
    if ((divtable[ps] < upper) && (divtable[ps] > lower)) {
        PWM_TIMER->PRESCALER = ps;
        freq = divtable[ps];
        break;
    }
}
...
return freq;
```

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,224 @@
...
+{
+    assert(dev == 0);
+    return 1;
+}
+
+void pwm_poweron(pwm_t dev)
+{
+    assert(dev == 0);
+
+#if CPU_FAM_NRF51
```
haukepetersen: this PWM driver is written specifically for the `nrf51`, so no need for this ifdef...

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,224 @@
...
+        return true;
+    }
+    return false;
+}
+
+uint32_t pwm_init(pwm_t dev, pwm_mode_t mode, uint32_t freq, uint16_t res)
+{
+    assert (dev == 0 && ((mode == PWM_LEFT) || (mode == PWM_RIGHT)));
+
+    /* reset and configure the timer */
```
haukepetersen: to be sure the power should be turned on first (`PWM_TIMER->POWER = 1`)

cpu/nrf51/periph/pwm.c
SemjonKerner: Indeed this should be the faster solution, but it seems to me as if you confuse pwm frequency and clock frequency at some point. I'll have a look how to reuse your suggestion, thx.

boards/calliope-mini/include/periph_conf.h
SemjonKerner: This sounds right. I'll keep it in mind when squashing, if this is ok with you.

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,217 @@
...
+    PWM_TIMER->MODE = TIMER_MODE_MODE_Timer;
+    PWM_TIMER->TASKS_CLEAR = 1;
+
+    /* calculate and set prescaler */
+    uint32_t timer_freq = freq * res;
+    uint32_t lower = (timer_freq - (PWM_PERCENT_VAL * (timer_freq / 100)));
+    uint32_t upper = (timer_freq + (PWM_PERCENT_VAL * (timer_freq / 100)));
+    for (uint32_t ps = 0;
+         ps <= (PWM_PS_MAX + 1);
+         ps++) {
```
haukepetersen: this should fit in one line, right?!

cpu/nrf51/periph/pwm.c
```diff
@@ -0,0 +1,217 @@
...
+    (CLOCK_CORECLOCK >> 1),
+    (CLOCK_CORECLOCK >> 2),
+    (CLOCK_CORECLOCK >> 3),
+    (CLOCK_CORECLOCK >> 4),
+    (CLOCK_CORECLOCK >> 5),
+    (CLOCK_CORECLOCK >> 6),
+    (CLOCK_CORECLOCK >> 7),
+    (CLOCK_CORECLOCK >> 8),
+    (CLOCK_CORECLOCK >> 9),
+    (CLOCK_CORECLOCK >> 10),
```
haukepetersen: the maximum allowed prescaler value is `9`, right? So we have 10 possible values -> this array should have only 10 members, correct?

haukepetersen:
all green -> go

@SemjonKerner nice work!

PR #8751, opened 2018-03-07
tests/periph_pwm: add shell commands

SemjonKerner:
This PR adds shell commands to the pwm test similar to  eg. the gpio test. It aims to provide enhanced feedback and more hands on testing capabilities. With this shell commands all default functions of a pwm driver can be tested separately and more thorougly. The old testing capabilites were kept in an own function and thus remain intact and may still be used to give a quick overview over the drivers functions.

### DIFF
--- a/tests/periph_pwm/main.c
+++ b/tests/periph_pwm/main.c
@@ -20,42 +20,140 @@
  * every 1s on every channel.
  *
  * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Semjon Kerner <semjon.kerner@fu-berlin.de>
  *
  * @}
  */
 
 #include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <limits.h>
 
 #include "xtimer.h"
+#include "shell.h"
 #include "timex.h"
 #include "periph/pwm.h"
 
-#define INTERVAL    (10LU * US_PER_MS) /* 10 ms */
-#define STEP        (10)
+#define OSC_INTERVAL    (10LU * US_PER_MS) /* 10 ms */
+#define OSC_STEP        (10)
+#define OSC_MODE        PWM_LEFT
+#define OSC_FREQU       (1000U)
+#define OSC_STEPS       (1000U)
+#define PWR_SLEEP       (1U)
 
-#define MODE        PWM_LEFT
-#define FREQU       (1000U)
-#define STEPS       (1000U)
+static uint32_t initiated;
 
-
-int main(void)
+static unsigned _get_dev(const char *dev_str)
 {
+    unsigned dev = (unsigned)atoi(dev_str);
+    if (dev >= PWM_NUMOF) {
+        printf("Error: device PWM_DEV(%u) is unknown\n", dev);
+        return UINT_MAX;
+    }
+    return dev;
+}
+
+static int _init(int argc, char** argv)
+{
+    if (argc != 5) {
+        printf("usage: %s <dev> <mode> <frequency> <resolution>\n", argv[0]);
+        printf("\tdev: device by number between 0 and %u\n", PWM_NUMOF - 1);
+        puts("\tmode:\n");
+        puts("\t\t0: left aligned\n");
+        puts("\t\t1: right aligned\n");
+        puts("\t\t2: center aligned\n");
+        puts("\tfrequency: desired frequency in Hz\n");
+        printf("\tresolution: number between 2 and %" PRIu16 "\n", UINT16_MAX);
+        return 1;
+    }
+
+    unsigned dev = _get_dev(argv[1]);
+    if (dev == UINT_MAX) {
+        return 1;
+    }
+
+    pwm_mode_t pwm_mode;
+    switch(atoi(argv[2])) {
+        case(0):
+            pwm_mode = PWM_LEFT;
+            break;
+        case(1):
+            pwm_mode = PWM_RIGHT;
+            break;
+        case(2):
+            pwm_mode = PWM_CENTER;
+            break;
+        default:
+            printf("Error: mode %d is not supported.\n", atoi(argv[2]));
+            return 1;
+    }
+
+    uint32_t pwm_freq = pwm_init(PWM_DEV(dev), pwm_mode,
+                                 (uint32_t)atoi(argv[3]),
+                                 (uint16_t)atoi(argv[4]));
+    if (pwm_freq != 0) {
+        printf("The pwm frequency is set to %" PRIu32 "\n", pwm_freq);
+        initiated |= (1 << dev);
+        return 0;
+    }
+
+    puts("Error: device is not initiated");
+    return 1;
+}
+
+static int _set(int argc, char**argv)
+{
+    if (argc != 4) {
+        printf("usage: %s <dev> <ch> <val>\n", argv[0]);
+        printf("\tdev: device by number between 0 and %d\n", PWM_NUMOF - 1);
+        puts("\tch: channel of device\n");
+        puts("\tval: duty cycle\n");
+        return 1;
+    }
+
+    unsigned dev = _get_dev(argv[1]);
+    if (dev == UINT_MAX) {
+        return 1;
+    }
+
+    if ((initiated & (1 << dev)) == 0) {
+        puts("Error: pwm is not initiated.\n");
+        puts("Execute init function first.\n");
+        return 1;
+    }
+
+    uint8_t chan = (uint8_t)atoi(argv[2]);
+    if (chan >= pwm_channels(PWM_DEV(dev))) {
+        printf("Error: channel %d is unknown.\n", chan);
+        return 1;
+    }
+
+    pwm_set(PWM_DEV(dev), chan, (uint16_t)atoi(argv[3]));
+    return 0;
+}
+
+static int _oscillate(int argc, char** argv)
+{
+    (void)argc;
+    (void)argv;
+
     int state = 0;
-    int step = STEP;
+    int step = OSC_STEP;
     xtimer_ticks32_t last_wakeup = xtimer_now();
 
     puts("\nRIOT PWM test");
-    puts("Connect an LED or scope to PWM pins to see something\n");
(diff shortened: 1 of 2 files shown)


tests/periph_pwm/main.c
```diff
@@ -68,12 +167,94 @@ int main(void)
...
+    }
+    printf("[pwm] Error: power state %s not available.", argv[2]);
+    return 1;
+
+}
+
+int power_test(int argc, char** argv){
+    if (argc != 2) {
+        printf("usage: %s <dev>\n", argv[0]);
+        printf("\tdev: device by number between 0 and %d\n", PWM_NUMOF-1);
```
haukepetersen: add spaces around minus sign, as in `PWM_NUMOF - 1`

tests/periph_pwm/main.c
```diff
@@ -68,12 +167,94 @@ int main(void)
...
+
+    uint32_t dev = strtol(argv[1], NULL, 10);
+    if (dev >= PWM_NUMOF) {
+        printf("[pwm] Error: device %ld is unknown.\n", dev);
+        return 1;
+    }
+
+    printf("Powering down PWM device and sleeping for %d second(s)...\n",
+           PWR_SLEEP);
+    pwm_poweroff(dev);
```
haukepetersen: you should not use the numeric PWM identifier directly, but always use the 'access marcro' `PWM_DEV(dev)`

tests/periph_pwm/main.c
```diff
@@ -68,12 +167,94 @@ int main(void)
...
+        printf("usage: %s <dev> <state>\n", argv[0]);
+        printf("\tdev: device by number between 0 and %d\n", PWM_NUMOF-1);
+        printf("\tstate:\n");
+        printf("\t\t0: power off\n");
+        printf("\t\t1: power on\n");
+        return 1;
+    }
+
+    errno = 0;
+    uint32_t dev = strtol(argv[1], NULL, 10);
```
haukepetersen: here and in the rest of the file: any reason for parsing 'long int' (strtol) vs simply parsing 'int' (atoi)? Since the target range is fairly small, it seems to me that
```c
unsigned dev = (unsigned)atoi(argv[1]);
```
would be the more efficient choice...

tests/periph_pwm/main.c
```diff
@@ -68,12 +167,94 @@ int main(void)
...
     }
 
     return 0;
 }
+
+int power(int argc, char** argv){
+    if (argc != 3) {
+        printf("usage: %s <dev> <state>\n", argv[0]);
+        printf("\tdev: device by number between 0 and %d\n", PWM_NUMOF-1);
+        printf("\tstate:\n");
```
haukepetersen: you could/should use `puts()` for printing strings without formatting parameters (slightly more efficient...).

tests/periph_pwm/main.c
```diff
@@ -68,12 +167,94 @@ int main(void)
...
+            case (1):
+                printf("Powering up PWM device.\n");
+                pwm_poweron(dev);
+                return 0;
+            default:
+                break;
+        }
+    }
+    printf("[pwm] Error: power state %s not available.", argv[2]);
+    return 1;
```
haukepetersen: IMHO this block can be simplified to something like
```c
switch (atoi(argv[2])) {
    case 0:
        puts("Powering down PWM device");
        pwm_poweroff(PWM_DEV(dev));
        break;
    case 1:
        puts("Powering up PWM device");
        pwm_poweron(PWM_DEV(dev));
        break;
    default:
        puts("Unable to parse target state");+
        return 1;
}
return 0
```

tests/periph_pwm/main.c
```diff
@@ -68,12 +167,94 @@ int main(void)
...
+        printf("usage: %s <dev>\n", argv[0]);
+        printf("\tdev: device by number between 0 and %d\n", PWM_NUMOF-1);
+        return 1;
+    }
+
+    uint32_t dev = strtol(argv[1], NULL, 10);
+    if (dev >= PWM_NUMOF) {
+        printf("[pwm] Error: device %ld is unknown.\n", dev);
+        return 1;
+    }
```
haukepetersen: this code could be re-used:

```c
// on top of this file
#include <limits.h>
...
static unsigned _get_dev(const char *dev_str)
{
    unsigned dev = (unsigned)atoi(dev_str);
    if (dev >= PWM_NUMOF) {
        printf("error: device PWM_DEV(%u) is unknown\n", dev);
        return UINT_MAX;
    }
    return dev;
}

// and whenever you need to parse the device
unsigned dev = _get_dev(argv[xxx]);
if (dev == UINT_MAX) {
    return 1;
}
...
```

tests/periph_pwm/main.c
```diff
@@ -20,42 +20,141 @@
...
+#define OSC_STEP        (10)
+#define OSC_MODE        PWM_LEFT
+#define OSC_FREQU       (1000U)
+#define OSC_STEPS       (1000U)
+#define PWR_SLEEP       (1U)
 
-#define MODE        PWM_LEFT
-#define FREQU       (1000U)
-#define STEPS       (1000U)
+bool initiated;
```
haukepetersen: something does not quite fit: what happens if there a device supports multiple PWM devices? E.g.
```
init 0 ...
set 1 xx
```
This will work, as `initiated` is set to `true` while initializing PWM_DEV(0), but does not yield the wanted behavior...

So I would suggest making this variable a bitfield:
```c
static uint32_t initialized = 0;

....
// in init()
initialized |= (1 << dev);

// in set()
if (initialized & (1 << dev) {
    ...
}
```

haukepetersen: oh, and of course it needs to be declared `static` :-)

tests/periph_pwm/main.c
```diff
@@ -20,42 +20,141 @@
...
+            break;
+        case(2):
+            pwm_mode = PWM_CENTER;
+            break;
+        default:
+            printf("[pwm] Error: mode %d is not supported.\n", mode);
+            return 1;
+    }
+
+    if (pwm_init(dev, pwm_mode, freq, res) != 0) {
```
haukepetersen: One more thing I just noticed: I think it would make sense to print the actual set PWM frequency after calling init.

tests/periph_pwm/main.c
```diff
@@ -20,42 +20,135 @@
...
 {
+    unsigned dev = (unsigned)atoi(dev_str);
+    if (dev >= PWM_NUMOF) {
+        printf("[pwm] Error: device PWM_DEV(%u) is unknown\n", dev);
+        return UINT_MAX;
+    }
+    return dev;
+}
+
+int init(int argc, char** argv) {
```
haukepetersen: here and below: make all shell command functions `static`. Also it is a good idea to prefex their names with an underscore.

tests/periph_pwm/main.c
```diff
@@ -20,42 +20,135 @@
...
+            return 1;
+    }
+
+    uint32_t pwm_freq = pwm_init(PWM_DEV(dev), pwm_mode,
+                                 atoi(argv[3]), atoi(argv[4]));
+    if (pwm_freq != 0) {
+        printf("[pwm] The pwm frequency is set to %ld\n", pwm_freq);
+        initiated |= (1 << dev);
+        return 0;
+    }
```
haukepetersen: How about some error message in case the initialization failed?!

tests/periph_pwm/main.c
```diff
@@ -20,42 +20,135 @@
...
+
+    unsigned dev = _get_dev(argv[1]);
+    if (dev == UINT_MAX) {
+        return 1;
+    }
+
+    if ((initiated & (1 << dev)) == 0) {
+        puts("[pwm] Error: pwm is not initiated.\n");
+        puts("Execute init function first.\n");
```
haukepetersen: nit-picky: I would be consistent: either put the `[pwm] ` prefix in front of all the output, or (my preferred way) don't use that prefix in this code here, as the shell command's output does not really need to be differentiated in this way...

tests/periph_pwm/main.c
```diff
@@ -20,42 +20,135 @@
...
+        return 1;
+    }
+
+    if (atoi(argv[2]) >= pwm_channels(PWM_DEV(dev))) {
```
haukepetersen: better cast the `atoi` result to `unint8_t` here, so no compiler can complain about signed/unsigned comparison

haukepetersen: maybe event better: use a variable for the channel (`uint8_t chan = (uint8_t)atoi(argv[2]);`), as this variable is used more than once in the code below...

tests/periph_pwm/main.c
```diff
@@ -20,42 +20,135 @@
...
+        return 1;
+    }
+
+    pwm_set(PWM_DEV(dev), atoi(argv[2]), atoi(argv[3]));
+    return 0;
+}
+
+int oscillate(int argc, char** argv) {
+    if (argc > 1) {
+        printf("Parameters %s not needed.\n", argv[1]);
```
haukepetersen: I'd see drop this check, simply ignore all parameters (saves a little code size :-) ).

tests/periph_pwm/main.c
```diff
@@ -68,12 +161,88 @@ int main(void)
...
+        case (1):
+            puts("Powering up PWM device.\n");
+            pwm_poweron(PWM_DEV(dev));
+            break;
+        default:
+            puts("[pwm] Error: power state not available.\n");
+            return 1;
+    }
+    return 0;
+
```
haukepetersen: remove newline

tests/periph_pwm/main.c
```diff
@@ -20,42 +20,141 @@
...
+        printf("Error: device PWM_DEV(%u) is unknown\n", dev);
+        return UINT_MAX;
+    }
+    return dev;
+}
 
-int main(void)
+static int _init(int argc, char** argv)
+{
+
```
haukepetersen: remove newline

haukepetersen:
and feel free to squash once the last comment is addressed

haukepetersen:
All green, good to go. Thanks for all the work!

PR #8763, opened 2018-03-09
boards/nucleo-l152: increase xtimer backoff

### LINKED ISSUE #7347: xtimer_usleep stuck for small values
On `nucleo-l1` (L152RE) `xtimer_usleep` never returns if you hand it values <=11.

minimum example to reproduce the error:
```c
#include <stdio.h>
#include "xtimer.h"

int main(void)
{
    for (int i = 100; i > 0; i--){
        printf("going to sleep %d usecs...\n", i);
        xtimer_usleep(i);
    }
    return 0;
}
```
For me this always leads to the application getting stuck after printing the following line: 
```
going to sleep 11 usecs...
```
In xtimer.c the callback [`_callback_unlock_mutex`](https://github.com/RIOT-OS/RIOT/blob/master/sys/xtimer/xtimer.c#L42) that should be fired after [`_xtimer_set64(&timer, offset, long_offset)`](https://github.com/RIOT-OS/RIOT/blob/master/sys/xtimer/xtimer.c#L64) is never called.

I also did a quick test on `nucleo-l476` where the faulty behaviour could not be observed.

aabadie:
### Contribution description

This PR just increases the xtimer backoff value for the STM32L1 based boards. It has been tested on nucleo-l152re board and fixes 2 stucked xtimer based test applications: `xtimer_usleep_short` and `xtimer_periodic_wakeup`.

Other xtimer test applications are still passing.

### Issues/PRs references

Fixes #7347

### DIFF
--- a/boards/common/nucleo/include/board_nucleo.h
+++ b/boards/common/nucleo/include/board_nucleo.h
@@ -48,7 +48,7 @@ extern "C" {
 #endif
 
 #if defined(CPU_FAM_STM32L1)
-#define XTIMER_BACKOFF              (3)
+#define XTIMER_BACKOFF              (11)
 #define XTIMER_OVERHEAD             (6)
 #endif
 


aabadie:
Maybe @smlng and @kYc0o are interested by this one.

kYc0o:
Somehow I missed this one. Tested on nucleo-l152 and nz32-sc151 with success.

PR #7712, opened 2017-10-10
cpu: stm32l1: add flashpage writing support

kYc0o:
Just added flashpage write support for stm32l1 cpu's, which can be tested with tests/periph_flashpage.

PR #8768, opened 2018-03-12
cpu/stm32: unify support for flashpage/flashpage_raw with L0/L1/F0/F1 families

aabadie:
### Contribution description

This PR adapts the STM32 flashpage feature to make it work with the STM32L0 family. It has been tested with success on nucleo-l0xx and b-l072z-lrwan1 boards.

### Issues/PRs references

None

### DIFF
--- a/cpu/stm32_common/periph/flashpage.c
+++ b/cpu/stm32_common/periph/flashpage.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2016 Freie Universität Berlin
+ *               2018 Inria
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -15,11 +16,14 @@
  * @brief       Low-level flash page driver implementation
  *
  * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Francisco Acosta <francisco.acosta@inria.fr>
+ * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
  *
  * @}
  */
 
 #include "cpu.h"
+#include "stmclk.h"
 #include "assert.h"
 
 #define ENABLE_DEBUG        (0)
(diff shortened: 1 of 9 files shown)


cpu/stm32_common/periph/flashpage.c
```diff
@@ -27,12 +27,33 @@
...
+#define CNTRL_REG_LOCK         (FLASH_PECR_PELOCK)
+#define KEY_REG                (FLASH->PEKEYR)
+#define FLASH_CR_PER           (FLASH_PECR_ERASE | FLASH_PECR_PROG)
+#define FLASH_CR_PG            (FLASH_PECR_FPRG | FLASH_PECR_PROG)
+#define FLASHPAGE_LIMIT        (FLASHPAGE_SIZE / 4)
+#else
+#define CNTRL_REG              (FLASH->CR)
+#define CNTRL_REG_LOCK         (FLASH_CR_LOCK)
+#define KEY_REG                (FLASH->KEYR)
+#define FLASHPAGE_LIMIT        (FLASHPAGE_SIZE / 2)
```
kYc0o: This `FLASHPAGE_LIMIT` define is not documented and thus I don't see what is really about.

cpu/stm32_common/periph/flashpage.c
```diff
@@ -41,46 +62,65 @@ void flashpage_write(int page, void *data)
 
     /* unlock the flash module */
     DEBUG("[flashpage] unlocking the flash module\n");
-    if (FLASH->CR & FLASH_CR_LOCK) {
-        FLASH->KEYR = FLASH_KEY1;
-        FLASH->KEYR = FLASH_KEY2;
+    if (CNTRL_REG & CNTRL_REG_LOCK) {
+        KEY_REG = FLASH_KEY1;
+        KEY_REG = FLASH_KEY2;
```
kYc0o: You're redefining this values, but I don't see why you don't use them below, which makes me wonder what's the use of it after all.

cpu/stm32_common/periph/flashpage.c
```diff
@@ -41,46 +62,65 @@ void flashpage_write(int page, void *data)
...
     }
 
+#if defined(CPU_FAM_STM32L0)
+    DEBUG("[flashpage] unlocking the flash program memory\n");
+    if (!(CNTRL_REG & CNTRL_REG_LOCK)) {
+        if (FLASH->PECR & FLASH_PECR_PRGLOCK) {
```
kYc0o: This nested if's seem a bit weird to me... especially because both `CTRL_REG` and `FLASH->PECR` are defining the same registers.

cpu/stm32_common/periph/flashpage.c
aabadie: I'm sorry @kYc0o, but I'd like that you read this PR carefully before commenting (and also the datasheet). They are not reused below because they are not the same (FLASH->PEKEYR != FLASH->PRGKEYR).

vincent-d:
I'd tend to prefer this approach over #7712 but I also think `flashpage_write_raw` could be added easily, but it could also come in a follow up as it could be implemented for every supported stm32 cpu.

cpu/stm32_common/periph/flashpage.c
kYc0o: I know the datasheet and I know what is this about (my PR was there for about 6 months), I was referring to `CTRL_REG` and actually since you're defining `KEY_REG` you could also define `KEY_PREG` to be reused in other parts of the code, especially if your intention is to add the EEPROM driver.

kYc0o:
> I'd tend to prefer this approach over #7712 but I also think flashpage_write_raw could be added easily, but it could also come in a follow up as it could be implemented for every supported stm32 cpu.

I still think we can merge both approaches, no need to add more PRs or wait for them. As I said it's trivial and adds a lot of flexibility while writing into flash.

What I propose is to add my changes to this PR, if @vincent-d likes it more.

aabadie:
> What I propose is to add my changes to this PR

Indeed there are good ideas on both sides. I updated this PR to integrate them and also added the flashpage_raw feature to stm32f0 and stm32f1.

Tested in nucleo-f030, nucleo-l152 and b-l072z-lrwan1 boards and works. I'm not sure of the values for alignment and block size, so it may be worth checking that.

cpu/stm32_common/periph/flashpage.c
```diff
@@ -27,60 +30,150 @@
...
+
+static void _lock(void)
+{
+    DEBUG("[flashpage] locking the flash module\n");
+    CNTRL_REG |= CNTRL_REG_LOCK;
+}
+
+void flashpage_write_raw(void *target_addr, void *data, size_t len)
+{
+    /* The actual minimal block size for writing is 16B, thus we
```
vincent-d: is it really 16B ?

cpu/stm32_common/periph/flashpage.c
```diff
@@ -27,60 +30,150 @@
...
+    /* STM32L0 only supports word sizes */
+    uint32_t *page_addr = flashpage_addr(page);
+#else
+    /* Default is to support half-word sizes */
     uint16_t *page_addr = flashpage_addr(page);
-    uint16_t *data_addr = (uint16_t *)data;
+#endif
     uint32_t hsi_state = (RCC->CR & RCC_CR_HSION);
 
     /* the internal RC oscillator (HSI) must be enabled */
```
vincent-d: isn't it needed in `flashpage_write_raw` as well?

cpu/stm32_common/periph/flashpage.c
```diff
@@ -27,60 +30,150 @@
...
     RCC->CR |= (RCC_CR_HSION);
     while (!(RCC->CR & RCC_CR_HSIRDY)) {}
```
vincent-d: you could use `stmclk_enable_hsi` and `stmclk_disable_hsi` below

cpu/stm32_common/periph/flashpage.c
```diff
@@ -15,76 +16,177 @@
...
+
+    DEBUG("[flashpage_raw] unlocking the flash module\n");
+    _unlock();
+
+    DEBUG("[flashpage] write: now writing the data\n");
+#if !(defined(CPU_FAM_STM32L0) || defined(CPU_FAM_STM32L1))
+    /* set PG bit and program page to flash */
+    CNTRL_REG |= FLASH_CR_PG;
+#endif
+    for (size_t i = 0; i < (FLASHPAGE_SIZE / FLASHPAGE_DIV); i++) {
```
vincent-d: `len / FLASHPAGE_DIV` ?

cpu/stm32_common/periph/flashpage.c
```diff
@@ -15,76 +16,177 @@
...
-        /* clear program bit again */
-        FLASH->CR &= ~(FLASH_CR_PG);
-        DEBUG("[flashpage] write: done writing data\n");
+        flashpage_write_raw(page_addr, data, FLASHPAGE_SIZE);
     }
 
     /* finally, lock the flash module again */
-    DEBUG("flashpage] now locking the flash module again\n");
-    FLASH->CR |= FLASH_CR_LOCK;
+    _lock();
```
vincent-d: I'm wondering if it wouldn't be cleaner to lock right after erasing, as unlocking/locking are done again in `flashpage_write_raw` in case of writing

cpu/stm32_common/periph/flashpage.c
```diff
@@ -15,76 +16,177 @@
...
+    /* ensure the length doesn't exceed the actual flash size */
+    assert(((unsigned)target_addr + len) <
+           (CPU_FLASH_BASE + (FLASHPAGE_SIZE * FLASHPAGE_NUMOF)) + 1);
+
+#if defined(CPU_FAM_STM32L0) || defined(CPU_FAM_STM32L1)
+    uint32_t *dst = target_addr;
+    uint32_t *data_addr = (uint32_t *)data;
+#else
+    uint16_t *dst = target_addr;
+    uint16_t *data_addr = (uint16_t *)data;
```
vincent-d: There is something I'm not sure to understand here: this seems to allow half-word (i.e. 2 bytes) alignment, as `FLASHPAGE_DIV` seems to suggest too. But `FLASHPAGE_RAW_ALIGNMENT` and `FLASHPAGE_RAW_BLOCKSIZE` are 4 and then it's not allowed to write only 2 bytes.

Am I missing something?

cpu/stm32_common/periph/flashpage.c
kYc0o: During my tests, it was actually possible to write even 4B, but I left 16B as the smallest.

aabadie:
@vincent-d @kYc0o, I tested more in depth the flashpage_raw feature: I think it works (not really sure how the test application is supposed to be used) with L0 and L1 but it crashes with F0 (I have no F1, except iotlab-m3 but have problem flashing this one).
Do you have an idea what could be problem ? I seems to be an alignment issue but I don't understand why.

cpu/stm32_common/periph/flashpage.c
```diff
@@ -15,76 +16,176 @@
...
+           (CPU_FLASH_BASE + (FLASHPAGE_SIZE * FLASHPAGE_NUMOF)) + 1);
+
+#if defined(CPU_FAM_STM32L0) || defined(CPU_FAM_STM32L1)
+    uint32_t *dst = target_addr;
+    uint32_t *data_addr = (uint32_t *)data;
+#else
+    uint16_t *dst = (uint16_t *)target_addr;
+    uint16_t *data_addr = (uint16_t *)data;
+#endif
+
```
vincent-d: HSI needs to be enabled here and disabled after locking, otherwise it crashes on stm32f091

aabadie: > HSI needs to be enabled here and disabled after locking

Thanks, I'll try that

aabadie: Thanks a lot @vincent-d, that was it! I refactored a bit the code again and tested on nucleo-l073 and nucleo-f070, it works on both.

aabadie:
@vincent-d @kYc0o, I think we are nearly good here.

aabadie:
@kYc0o just missing your ACK, and we are good !

kYc0o:
It looks ok. I'll test and give my ACK.

kYc0o:
Tested ACK.

cpu/stm32f1/include/cpu_conf.h
```diff
@@ -65,6 +65,13 @@ extern "C" {
 #elif defined(CPU_MODEL_STM32F103RE) || defined(CPU_MODEL_STM32F103ZE)
 #define FLASHPAGE_NUMOF     (256U)
 #endif
+
+/* The minimum block size which can be written is 2B. However, the erase
+ * block is always FLASHPAGE_SIZE.
+ */
+#define FLASHPAGE_RAW_BLOCKSIZE    (2U)
+/* Writing should be always 4 bytes aligned */
+#define FLASHPAGE_RAW_ALIGNMENT    (4U)
```
aabadie: @cladmi, It could be one of those 2 lines, can you test with a value of 2 ?

cladmi: With `ENABLE_DEBUG (1)` in `stm32_common/periph/flashpage.c` it works so it must be something else.

cladmi:
Broken on `iotlab-m3`. Test procedure on IoT-LAB.
```
dump 255
1521050271.397776;m3-380;> dump 255
# 0xff everywhere
test 255
1521050275.361341;m3-380;> test 255
     # need to type enter to see the error
1521050278.902577;m3-380;error verifying the content of page 255: >  
dump 255
1521050290.161453;m3-380;> dump 255
# 0xff everywhere
```

With `ENABLE_DEBUG (1)`  its working properly
```
test 255
...
1521049756.371371;m3-380;[flashpage_raw] writing s to 0x807fffe
1521049756.372367;m3-380;[flashpage] write: done writing data
1521049756.373414;m3-380;flashpage_raw] now locking the flash module again
1521049756.373651;m3-380;[flashpage] locking the flash module
1521049756.375369;m3-380;wrote local page to flash page 255 at addr 0x807f800
dump 255
1521050605.346219;m3-380;> dump 255
# a b c d... as expected
```

PR #8331, opened 2018-01-07
arm bug minor cpu/stm32f1 Fixed wrong flash page sizes

Icaltary:
### Contribution description
This PR fixes a wrong FLASHPAGE_SIZE value for STM32F103x8 and STM32F103xB CPUs. As stated in the datasheet [1] the page size of a flash page is 1024 bytes for these controllers.
The old value (2048) is still correct for the STM32F103xC, STM32F103xD and STM32F103xE [2].

[1] https://www.st.com/content/ccc/resource/technical/document/datasheet/33/d4/6f/1d/df/0b/4c/6d/CD00161566.pdf/files/CD00161566.pdf/jcr:content/translations/en.CD00161566.pdf (p. 57)
[2] https://www.st.com/content/ccc/resource/technical/document/datasheet/59/f6/fa/84/20/4e/4c/59/CD00191185.pdf/files/CD00191185.pdf/jcr:content/translations/en.CD00191185.pdf (p. 64)

### Issues/PRs references
None

PR #7700, opened 2017-10-09
net/gcoap: Add detailed message init options

kb2ma:
This PR implements the fifth step in the confirmable messaging plan in #7192, and depends on #7548. Presently, the gcoap_xxx_init() functions accept parameters that can change from message to message, for example the message code (GET, PUT, etc). On the other hand, some parameters, like message type, expect a default value and are not included in the init functions. 

For control over _all_ useful parameters, this PR adds gcoap_req_init_opts() for requests and gcoap_obs_init_opts() for observe notifications. The init opts functions accept a gcoap_send_opts_t structure, which includes all of the available attributes. The documentation for each function describes which attributes are read. This infrastructure will be used in the remaining PRs for confirmable messaging -- synchronous messaging and separate ACK response PRs.

My approach with this PR has been to keep 'simple things easy' while allowing 'difficult things to be possible'. The existing (easy) ...init() functions are unchanged and still useful for many use cases, but when you need full control (difficult), all of the parameters are available to be set in ...init_opts(). The ...init() functions were refactored to use the ...init_opts() functions.

PR #8772, opened 2018-03-13
net/nanocoap: rework option handling

kaspar030:
### Contribution description

This PR reworks nanocoap's option handling. Instead of having a field in coap_pkt_t for each supported CoAP option, the parsing step now just notes down the options with option number and offset in an array. Furthermore, some layers of parsing/iterating functions are provided.

As gcoap is not yet adapted, some compatibility layer makes it still work.

As a first user for the new option handling, this PR also adds block1 support to nanocoap, with an example, blockwise-capable /sha256 enpoint for examples/nanocoap_server.

### DIFF
--- a/sys/include/net/nanocoap.h
+++ b/sys/include/net/nanocoap.h
@@ -47,9 +47,14 @@ extern "C" {
  * @name    Nanocoap specific maximum values
  * @{
  */
-#define NANOCOAP_URL_MAX        (64)
-#define NANOCOAP_QS_MAX         (64)
+#define NANOCOAP_NOPTS_MAX      (16)
+#define NANOCOAP_URI_MAX        (64)
 /** @} */
+
+#ifdef MODULE_GCOAP
+#define NANOCOAP_URL_MAX        NANOCOAP_URI_MAX
+#define NANOCOAP_QS_MAX         (64)
+#endif
 
 /**
  * @name    CoAP option numbers
@@ -220,14 +225,26 @@ typedef struct __attribute__((packed)) {
  * @brief   CoAP option array entry
  */
 typedef struct {
-    coap_hdr_t *hdr;                /**< pointer to raw packet              */
-    uint8_t url[NANOCOAP_URL_MAX];  /**< parsed request URL                 */
-    uint8_t qs[NANOCOAP_QS_MAX];    /**< parsed query string                */
-    uint8_t *token;                 /**< pointer to token                   */
-    uint8_t *payload;               /**< pointer to payload                 */
-    unsigned payload_len;           /**< length of payload                  */
-    uint16_t content_type;          /**< content type                       */
-    uint32_t observe_value;         /**< observe value                      */
+    uint16_t opt_num;           /**< full CoAP option number    */
+    uint16_t offset;            /**< offset in packet           */
+} coap_optpos_t;
+
+/**
+ * @brief   CoAP PDU parsing context structure
+ */
+typedef struct {
+    coap_hdr_t *hdr;                            /**< pointer to raw packet   */
+    uint8_t *token;                             /**< pointer to token        */
+    uint8_t *payload;                           /**< pointer to payload      */
+    uint16_t payload_len;                       /**< length of payload       */
+    uint16_t options_len;                       /**< length of options array */
+    coap_optpos_t options[NANOCOAP_NOPTS_MAX];  /**< option offset array     */
+#ifdef MODULE_GCOAP
+    uint8_t url[NANOCOAP_URI_MAX];              /**< parsed request URL      */
+    uint8_t qs[NANOCOAP_QS_MAX];                /**< parsed query string     */
+    uint16_t content_type;                      /**< content type            */
+    uint32_t observe_value;                     /**< observe value           */
+#endif
 } coap_pkt_t;
 
 /**
@@ -391,6 +408,33 @@ size_t coap_put_option_ct(uint8_t *buf, uint16_t lastonum, uint16_t content_type
  * @returns     amount of bytes written to @p buf
  */
 size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uint16_t optnum);
+
+
+/**
+ * @brief    Get content type from packet
+ *
+ * @param[in]   pkt     packet to work on
+ *
+ * @returns     the packet's content type value if included,
+ *              COAP_FORMAT_NONE otherwise
+ */
+unsigned coap_get_content_type(coap_pkt_t *pkt);
+
+/**
+ * @brief    Get the packet's request URI
+ *
+ * This function decodes the pkt's URI option into a "/"-seperated and
+ * NULL-terminated string.
+ *
+ * Caller must ensure @p target can hold at least NANOCOAP_URI_MAX bytes!
+ *
+ * @param[in]   pkt     pkt to work on
+ * @param[out]  target  buffer for target URI
+ *
+ * @returns     -ENOSPC     if URI option is larger than NANOCOAP_URI_MAX
+ * @returns     nr of bytes written to @p target (including '\0')
+ */
+int coap_get_uri(coap_pkt_t *pkt, uint8_t *target);
 
 /**
  * @brief   Get the CoAP version number
@@ -556,6 +600,7 @@ static inline unsigned coap_method2flag(unsigned code)
     return (1 << (code - 1));
 }
 
+#if defined(MODULE_GCOAP) || defined(DOXYGEN)
 /**
  * @brief   Identifies a packet containing an observe option
  *
@@ -590,6 +635,7 @@ static inline uint32_t coap_get_observe(coap_pkt_t *pkt)
 {
     return pkt->observe_value;
 }
+#endif
 
 /**
  * @brief   Reference to the default .well-known/core handler defined by the
--- a/sys/net/application_layer/nanocoap/nanocoap.c
+++ b/sys/net/application_layer/nanocoap/nanocoap.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016-17 Kaspar Schleiser <kaspar@schleiser.de>
+ * Copyright (C) 2016-18 Kaspar Schleiser <kaspar@schleiser.de>
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -29,7 +29,7 @@
 #include "debug.h"
 
 static int _decode_value(unsigned val, uint8_t **pkt_pos_ptr, uint8_t *pkt_end);
-static uint32_t _decode_uint(uint8_t *pkt_pos, unsigned nbytes);
+int coap_get_option_uint(coap_pkt_t *pkt, unsigned opt_num, uint32_t *target);
 
 /* http://tools.ietf.org/html/rfc7252#section-3
  *  0                   1                   2                   3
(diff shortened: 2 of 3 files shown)


sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -46,17 +47,14 @@ static uint32_t _decode_uint(uint8_t *pkt_pos, unsigned nbytes);
...
 {
-    uint8_t *urlpos = pkt->url;
     coap_hdr_t *hdr = (coap_hdr_t *)buf;
-
     pkt->hdr = hdr;
 
     uint8_t *pkt_pos = hdr->data;
     uint8_t *pkt_end = buf + len;
 
-    memset(pkt->url, '\0', NANOCOAP_URL_MAX);
```
pekkanikander: Is this change documented?  If not, may it cause nasty surprises to existing code?

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -67,9 +65,13 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
         pkt->token = NULL;
     }
 
+    coap_optpos_t *optpos = pkt->options;
+    unsigned option_count = 0;
```
pekkanikander: Is using just `unsigned` instead of the "more standard" `unsigned int` according to the used style guide?

sys/net/application_layer/nanocoap/nanocoap.c
pekkanikander: Or does the change on line #L119 override always the URL field?

kaspar030: ```coap->pkt``` does not exist anymore, unless gcoap is used. In that case, #L119 does overwrite the field.

Anyhow, I've added another null-termination for the case that the option is not present.

sys/net/application_layer/nanocoap/nanocoap.c
kaspar030: In RIOT just ```unsigned``` seems to be "more standard". Personally I prefer less typing and one-word type identifiers...

sys/net/application_layer/nanocoap/nanocoap.c
kaspar030: Ah, you mean whether removing the field might cause nasty surprises? Very probably. :)

kb2ma:
Thanks for this work, Kaspar! I have browsed through it, and just from a structural perspective:

- I would separate Block support into another PR. It seems too much to refactor coap_pkt_t and option parsing -- *and* to add support for a CoAP extension RFC in a single PR.
- With work at this level, I would expect to see at least some basic unit tests to validate reading options.

Naturally, I want to adapt gcoap to use this work ASAP. This will be a good way for me to get familiar with the implementation, and provide more detailed comments.

kaspar030:
> No time. :) Maybe sometimes extensive testing is enough?

Ok, that's no excuse. I've added a (basic) test for get_option_uri(), which tests parsing options into the array, iterating through multi options and finally also the get_option_uri().

kb2ma:
Excellent, thanks a lot for that rework. I understand the ambivalence about unit tests, and appreciate sinking the extra time into it. We can build on it as needed. 

Now it's time for me to get to work and provide some feedback. :-)

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -89,58 +89,142 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
...
 
-    DEBUG("coap pkt parsed. code=%u detail=%u payload_len=%u, 0x%02x\n",
+    pkt->options_len = option_count;
+    if (!pkt->payload) {
+        pkt->payload = pkt_pos;
+    }
+
+#ifdef MODULE_GCOAP
+    coap_get_uri(pkt, pkt->url);
+    pkt->observe_value = UINT32_MAX;
```
kb2ma: Must update this attribute if observe option is present.

kb2ma:
Shouldn't coap_find_option(), coap_get_content_type(), coap_get_len(), and coap_iterate_option() be declared in nanocoap.h?

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -89,58 +89,142 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
...
+
+    return content_type;
+}
+
+int coap_get_uri(coap_pkt_t *pkt, uint8_t *target)
+{
+    uint8_t *opt_pos = coap_find_option(pkt, COAP_OPT_URI_PATH);
+    if (!opt_pos) {
+        DEBUG("nanocoap: no COAP_OPT_URI_PATH option\n");
+        *target = '\0';
```
kb2ma: The absence of a Uri-Path option means the path is "/". See Item 7 in sec. 6.5 of 7252. This actually is a bug in the current implementation as well.

kb2ma:
I just added my thoughts on adaption of gcoap to this PR in #8831. No concerns beyond the issues I raised [above](https://github.com/RIOT-OS/RIOT/pull/8772#pullrequestreview-104772257).

sys/net/application_layer/nanocoap/nanocoap.c
kaspar030: fixed

sys/net/application_layer/nanocoap/nanocoap.c
kaspar030: > This idea points to a more general question -- should the get/set option functions be based on the type of data format rather than the specific option?

Yeah, very good idea. I've started in this direction with a ```coap_get_option_uint()``` for getting the observe value.

kaspar030: (that reply belongs to the getting/setting options comment below)

kaspar030:
> I see you have removed coap_hdr_get_ver(). gcoap does not need this function specifically, but it does need coap_hdr_ functions for message type and token length.

Removal of ```coap_hdr_get_ver()``` was an accident. Probably happened when we moved nanocoap into main while adding doxygen. I intend to keep the coap_hdr_* functions.

kaspar030:
> Shouldn't coap_find_option(), coap_get_content_type(), coap_get_len(), and coap_iterate_option() be declared in nanocoap.h?

They should... Can we postpone that? I hoped to clean up the naming consistency a little.
IMO, the prefixing is a little off, and there should be clear ```coap_pkt_get/set(), coap_hdr_get/set(), coap_opt_X_get/set() ...``` naming conventions. I think they are a off sometimes.

kb2ma:
> I hoped to clean up the naming consistency a little.

No worries. I'm definitely in favor of standardizing the function names to make it easier to use them. I can experiment with the declarations as needed to prototype the gcoap adaptation.

sys/include/net/nanocoap.h
```diff
@@ -392,6 +409,33 @@ size_t coap_put_option_ct(uint8_t *buf, uint16_t lastonum, uint16_t content_type
...
+ *
+ * This function decodes the pkt's URI option into a "/"-seperated and
+ * NULL-terminated string.
+ *
+ * Caller must ensure @p target can hold at least NANOCOAP_URI_MAX bytes!
+ *
+ * @param[in]   pkt     pkt to work on
+ * @param[out]  target  buffer for target URI
+ *
+ * @returns     -EBADMSG    if no URI option in packet
```
bergzand: With the [comment](https://github.com/RIOT-OS/RIOT/pull/8772#discussion_r176366289) from @kb2ma resolved, this return error code is not valid anymore for the specified situation.

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -89,58 +90,181 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
...
+    DEBUG("coap pkt parsed. code=%u detail=%u payload_len=%u, nopts=%u, 0x%02x\n",
           coap_get_code_class(pkt),
           coap_get_code_detail(pkt),
-          pkt->payload_len, hdr->code);
+          pkt->payload_len, option_count, hdr->code);
 
     return 0;
 }
 
+uint8_t *coap_find_option(coap_pkt_t *pkt, unsigned opt_num)
```
bergzand: Is there a reason for not marking the `coap_pkt_t *pkt` as `const` (`const coap_pkt_t *pkt`) here and in the other getters below?

sys/include/net/nanocoap.h
kaspar030: fixed. I also fixed the return value for the case where there's no URI option. It returned 1, but doc and other cases return 2, because the '\0' gets counted.

This makes me wonder - what's more appropriate from an API perspective? Return the number of bytes written to the output buffer, or return the length of the extracted path string? We can change that now, not so much later on...

kb2ma:
Just added a couple of inline comments. My Observe tests passed after I updated coap_get_option_uint(). When these comments are addressed, I'm ready to approve.

sys/include/net/nanocoap.h
```diff
@@ -392,6 +409,32 @@ size_t coap_put_option_ct(uint8_t *buf, uint16_t lastonum, uint16_t content_type
  */
 size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uint16_t optnum);
 
+
+/**
+ * @brief    Get content type from packet
+ *
+ * @param[in]   pkt     packet to work on
+ *
+ * @returns     the packet's content type value
```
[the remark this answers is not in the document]
kaspar030: good idea, done

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -89,58 +90,181 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
...
+int coap_get_option_uint(coap_pkt_t *pkt, unsigned opt_num, uint32_t *target)
+{
+    assert(target);
+
+    uint8_t *opt_pos = coap_find_option(pkt, opt_num);
+    if (opt_pos) {
+        uint16_t delta;
+        int option_len = 0;
+        uint8_t *pkt_pos = _parse_option(pkt, opt_pos, &delta, &option_len);
+        if (option_len) {
```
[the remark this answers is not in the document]
kaspar030: yes, fixed.

kaspar030:
@kb2ma addressed your comments

sys/net/application_layer/nanocoap/nanocoap.c
haukepetersen: This function should be exported in the header, right?!

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -89,58 +90,181 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
...
+        }
+        else {
+            DEBUG("nanocoap: discarding packet with invalid option length.\n");
+            return -EBADMSG;
+        }
+    }
+    return -1;
+}
+
+uint8_t *coap_iterate_option(coap_pkt_t *pkt, uint8_t **optpos, int *opt_len, int first)
```
haukepetersen: same here: should be exported, right?!

kb2ma:
@haukepetersen, see the comments above starting [here](https://github.com/RIOT-OS/RIOT/pull/8772#pullrequestreview-104772257), which match your comments. I agree on more generic functions. See my ideas for gcoap in #8331.

haukepetersen:
sorry, the comments above slipped through.

@kb2ma are you referring to #7700? #8331 seems to something quite different :-)

For further proceedings, I suggest we go with this PR as is - after another thorough round of testing of course. Then I could prepare a first step of generalization when adding the 'LOCATION_PATH' option needed for the RD client.

kb2ma:
> are you referring to #7700? #8331 seems to be something quite different :-)

Whoops, meant #8831, for gcoap adaptation to these changes. It includes similar thoughts on data-centric option functions.

> For further proceedings, I suggest we go with this PR as is.

I agree. I plan to test latest fixups today.

kaspar030:
- squashed

@bergzand @haukepetersen you good for now?

kaspar030:
Nice, thanks for the reviews!

haukepetersen:
Thanks for the actual work :-)

PR #8773, opened 2018-03-13
drivers/flashpage: use const pointers for write and verify

vincent-d:
### Contribution description

This update the periph_flashpage interface to use const pointer for writes and verify operations.

### Issues/PRs references

Based on #8768

### DIFF
--- a/cpu/efm32/periph/flashpage.c
+++ b/cpu/efm32/periph/flashpage.c
@@ -26,12 +26,12 @@
 
 #include "em_msc.h"
 
-void flashpage_write(int page, void *data)
+void flashpage_write(int page, const void *data)
 {
     assert(page < (int)FLASHPAGE_NUMOF);
 
     uint32_t *page_addr = (uint32_t *)flashpage_addr(page);
-    uint32_t *data_addr = (uint32_t *)data;
+    const uint32_t *data_addr = data;
 
     MSC_Init();
     MSC_ErasePage(page_addr);
--- a/cpu/msp430_common/periph/flashpage.c
+++ b/cpu/msp430_common/periph/flashpage.c
@@ -24,11 +24,11 @@
 #include "irq.h"
 #include "periph/flashpage.h"
 
-void flashpage_write(int page, void *data)
+void flashpage_write(int page, const void *data)
 {
     assert(page < FLASHPAGE_NUMOF);
 
-    uint8_t *src = (uint8_t *)data;
+    const uint8_t *src = data;
     uint8_t *dst = (uint8_t *)flashpage_addr(page);
     unsigned istate;
 
--- a/cpu/nrf5x_common/periph/flashpage.c
+++ b/cpu/nrf5x_common/periph/flashpage.c
@@ -23,12 +23,12 @@
 #include "assert.h"
 #include "periph/flashpage.h"
 
-void flashpage_write(int page, void *data)
+void flashpage_write(int page, const void *data)
 {
     assert(page < (int)FLASHPAGE_NUMOF);
 
     uint32_t *page_addr = (uint32_t *)flashpage_addr(page);
-    uint32_t *data_addr = (uint32_t *)data;
+    const uint32_t *data_addr = data;
 
     /* erase given page */
     NRF_NVMC->CONFIG = NVMC_CONFIG_WEN_Een;
--- a/cpu/sam0_common/periph/flashpage.c
+++ b/cpu/sam0_common/periph/flashpage.c
@@ -56,7 +56,7 @@ static void _lock(void)
 #endif
 }
 
-void flashpage_write_raw(void *target_addr, void *data, size_t len)
+void flashpage_write_raw(void *target_addr, const void *data, size_t len)
 {
     /* The actual minimal block size for writing is 16B, thus we
      * assert we write on multiples and no less of that length.
@@ -72,7 +72,7 @@ void flashpage_write_raw(void *target_addr, void *data, size_t len)
            (CPU_FLASH_BASE + (FLASHPAGE_SIZE * FLASHPAGE_NUMOF)));
 
     uint32_t *dst = (uint32_t *)target_addr;
-    uint32_t *data_addr = (uint32_t *)data;
+    const uint32_t *data_addr = data;
 
     /* write 4 bytes in one go */
     len /= 4;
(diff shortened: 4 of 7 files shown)


vincent-d:
Rebased

kYc0o:
@cladmi can you take a look? I remember we talked something about this but decided to not go this way for a reason... I can't recall it right now...

cladmi:
Going for `const` is good, as it only tells that functions do not modify the data, so I ACK this part. Its a better API with it.

However you removed the `(uint32_t *)` cast from the `void *data` instead of going for `(const uint32_t *)` and that should be reverted I think.

From what I remember, the goal of these lines was to somehow counter a lack in the api types saying that "data" should be 4 bytes aligned. Its a hack instead of a proper fix, but there is not need to remove it now.

cpu/efm32/periph/flashpage.c
```diff
@@ -26,18 +26,17 @@
 
 #include "em_msc.h"
 
-void flashpage_write(int page, void *data)
+void flashpage_write(int page, const void *data)
 {
     assert(page < (int)FLASHPAGE_NUMOF);
 
     uint32_t *page_addr = (uint32_t *)flashpage_addr(page);
-    uint32_t *data_addr = (uint32_t *)data;
```
cladmi: I think this should be kept as `(const uint32_t *)data` as it somehow says alignment constraints, even if its note in the API.

cladmi: I did not repeat but its also present in other files.

vincent-d: I don't understand why you want that.

In most cases, the implicit cast is still there anyway. And casting does not enforce any alignment. In this example, without casting, you could have a crash if the alignment is not met. If you use this temporary variable and you cast, you will have the right alignment, but you could write data that you do not expect to write. So I'm not sure which is worst, but I think a crash would be better to catch issues here.

cladmi: I would say it should not be changed in this PR that says to only change to `const`. Fixing this it is a different problem.

I agree that the current state is far from perfect. But random crash are not good either.
Having this line helped me debug a crash in someone else code with a buffer coming from the shell and there was no other mention that for this platform it should be aligned and having this in the code helped me find it.

I would prefer an API fix with a per platform type or alignment definition but it would say it should be another PR.

vincent-d: OK to re-add the temp variable but I would prefer keeping implicit cast, which is IMO more aligned with the rest of the code base.

cladmi: I re-checked all the changes and my bad, I generalized this line as if you did the same to the other files.

For the other ones you just removed the cast part, which is true does not help at all. Keeping only the assignement to an `uint32_t *` variable is enough I agree. So you can keep them as you changed now as murdock does not issue warnings.

Here however you also removed the `uint32_t *data_addr` instead of only defining it const.
And from the `MSC_WriteWordI` implementation, it also cast the pointer to `uint32_t` somewhere, so I would prefer keeping this `data_addr` intermediate variable.

cladmi: :D I was answering at the same time.

And yes, re-adding the temp variable here is enough for me.

vincent-d: OK, will do!

vincent-d: Done

cladmi:
I agree on the API change, not tested on mentioned hardware.
Lets see Murdock output.

kYc0o:
Before merging, I was testing a bit more thoroughly this PR and discovered that the stm32 driver is not working properly, at least on my desk.

Can someone test on the F variants that `test <page>` works using tests/periph_flashpage ?

vincent-d:
Works as expected for me om nucleo-f091 (stm32f091). What do you see @kYc0o ?

kYc0o:
I'm on a nucleo-f072 (stm32f072rb) and when I do `test 50` it hangs, without any error, it's just responseless... If I reset it still responseless (though I have serial output), but when I execute any command it doesn't answer. If I turn it off and on again I can again interact with it. I execute `dump 50` and nothing is written. I have the latest st-link firmware. The same behaviour (even worse because it makes my board unflashable afterwards) on my IoT-Lab boards.

Very strange... Do you have access to IoT-Lab?

kYc0o:
Ok so I just discovered that I have the right to write one command on each turn off/turn on round (another issue...). The pattern written by `test 50` it's there and works.

kYc0o:
However, I have no issues at all on nucleo-l152. It might be my board which has something strange. If we could have a valid test on iotlab-m3 it would be awesome. Yesterday I succeeded to test but today I have failed assertions with `flashpage_write_raw`.

vincent-d:
I gave another try on nucleo32-f042 and it works like a charm.

I don't have access to IoT-Lab

kYc0o:
This is my output on IoT-Lab

```console
test 100
1521046351.786173;m3-100;> test 100

help
1521046356.899114;m3-100;error verifying the content of page 100: > help
1521046356.901197;m3-100;Command              Description
1521046356.903274;m3-100;---------------------------------------
1521046356.903466;m3-100;info                 Show information about pages
1521046356.903601;m3-100;dump                 Dump the selected page to STDOUT
1521046356.907255;m3-100;dump_local           Dump the local page buffer to STDOUT
1521046356.907447;m3-100;read                 Read and output the given page
1521046356.907583;m3-100;write                Write (ASCII) data to the given page
1521046356.909116;m3-100;write_raw            Write (ASCII, max 64B) data to the given address
1521046356.909308;m3-100;erase                Erase the given page
1521046356.912688;m3-100;edit                 Write bytes to the local page
1521046356.912881;m3-100;test                 Write and verify test pattern
```

Somehow the serial is retained... I don't know exactly what can be related but I'm investigating.

aabadie:
> Somehow the serial is retained... I don't know exactly what can be related but I'm investigating.

This is a bug in the test application a `\n` is missing at the end of `error verifying the content of page 100:`, just press enter. But this means that the value are not written correctly.

cladmi:
It works on IoT-LAB with commits before https://github.com/RIOT-OS/RIOT/pull/8768

cladmi:
On `iotlab-m3` it works when there is `ENABLE_DEBUG` in `stm32_common/periph/flashpage.c` both before and after this PR.
It also works on `samr21-xpro`

So this PR can be merged as the problem comes from somewhere else.

kYc0o:
Ok, let's merge this then.

PR #8774, opened 2018-03-13
drivers: add mtd wrapper for periph_flashpage

vincent-d:
### Contribution description

This adds a basic mtd wrapper around perip_flashpage interface. periph_flashpage_raw is needed to write.

### Issues/PRs references

Based on #8773

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -342,14 +342,21 @@ ifneq (,$(filter mrf24j40,$(USEMODULE)))
   FEATURES_REQUIRED += periph_spi
 endif
 
-ifneq (,$(filter mtd_sdcard,$(USEMODULE)))
+ifneq (,$(filter mtd_%,$(USEMODULE)))
   USEMODULE += mtd
-  USEMODULE += sdcard_spi
-endif
 
-ifneq (,$(filter mtd_spi_nor,$(USEMODULE)))
-  USEMODULE += mtd
-  FEATURES_REQUIRED += periph_spi
+  ifneq (,$(filter mtd_sdcard,$(USEMODULE)))
+    USEMODULE += sdcard_spi
+  endif
+
+  ifneq (,$(filter mtd_spi_nor,$(USEMODULE)))
+    FEATURES_REQUIRED += periph_spi
+  endif
+
+  ifneq (,$(filter mtd_flashpage,$(USEMODULE)))
+    FEATURES_REQUIRED += periph_flashpage
+    FEATURES_REQUIRED += periph_flashpage_raw
+  endif
 endif
 
 ifneq (,$(filter my9221,$(USEMODULE)))
--- a/drivers/include/mtd_flashpage.h
+++ b/drivers/include/mtd_flashpage.h
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2018 OTA keys S.A.
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_mtd_flashpage   Flashpage MTD
+ * @ingroup     drivers_storage
+ * @brief       Driver for internal flash devices implementing flashpage interface
+ *
+ * @{
+ *
+ * @file
+ * @brief       Interface definition for the flashpage memory driver
+ *
+ * @author      Vincent Dupont <vincent@otakeys.com>
+ */
+
+#ifndef MTD_FLASHPAGE_H
+#define MTD_FLASHPAGE_H
+
+#include "mtd.h"
+#include "periph/flashpage.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/**
+ * @brief   Macro helper to initialize a mtd_t with flash-age driver
+ */
+#define MTD_FLASHPAGE_INIT_VAL(_pages_per_sector) { \
+    .driver = &mtd_flashpage_driver, \
+    .sector_count = FLASHPAGE_NUMOF, \
+    .pages_per_sector = _pages_per_sector,\
+    .page_size = FLASHPAGE_SIZE / _pages_per_sector,\
+}
+
+/**
+ * @brief   Flashpage MTD device operations table
+ */
+extern const mtd_desc_t mtd_flashpage_driver;
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* MTD_FLASHPAGE_H */
+/** @} */
--- a/drivers/mtd_flashpage/mtd_flashpage.c
+++ b/drivers/mtd_flashpage/mtd_flashpage.c
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C) 2018 OTA keys S.A.
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_mtd_flashpage
+ * @brief       Driver for internal flash devices implementing flashpage interface
+ *
+ * @{
+ *
+ * @file
+ * @brief       Implementation for the flashpage memory driver
+ *
+ * @author      Vincent Dupont <vincent@otakeys.com>
+ * @}
+ */
+
+#include <string.h>
+#include <errno.h>
+#include <assert.h>
+
+#include "cpu_conf.h"
+#include "mtd_flashpage.h"
+#include "periph/flashpage.h"
+
+#define MTD_FLASHPAGE_END_ADDR     CPU_FLASH_BASE + (FLASHPAGE_NUMOF * FLASHPAGE_SIZE)
+
+static int _init(mtd_dev_t *dev)
+{
+    (void)dev;
+    assert(dev->pages_per_sector * dev->page_size == FLASHPAGE_SIZE);
+    return 0;
+}
+
+static int _read(mtd_dev_t *dev, void *buf, uint32_t addr, uint32_t size)
+{
+    assert(addr < MTD_FLASHPAGE_END_ADDR);
+    (void)dev;
+
+    if (addr % FLASHPAGE_RAW_ALIGNMENT) {
+        return -EINVAL;
+    }
+
+    memcpy(buf, (void*)addr, size);
+
+    return size;
+}
+
+static int _write(mtd_dev_t *dev, const void *buf, uint32_t addr, uint32_t size)
+{
+    (void)dev;
+    if (addr % FLASHPAGE_RAW_ALIGNMENT) {
+        return -EINVAL;
+    }
+    if ((uintptr_t)buf % FLASHPAGE_RAW_ALIGNMENT) {
+        return -EINVAL;
+    }
+    if (size % FLASHPAGE_RAW_BLOCKSIZE) {
+        return -EOVERFLOW;
+    }
+    if (addr + size > MTD_FLASHPAGE_END_ADDR) {
+        return -EOVERFLOW;
+    }
+    flashpage_write_raw((void *)addr, buf, size);
+
+    return size;
+}
+
+int _erase(mtd_dev_t *dev, uint32_t addr, uint32_t size)
+{
+    size_t sector_size = dev->page_size * dev->pages_per_sector;
+
+    if (size % sector_size) {
+        return -EOVERFLOW;
+    }
+    if (addr + size > MTD_FLASHPAGE_END_ADDR) {
+        return - EOVERFLOW;
+    }
+    if (addr % sector_size) {
+        return - EOVERFLOW;
+    }
+    for (size_t i = 0; i < size; i += sector_size) {
+        flashpage_write(flashpage_page((void *)addr), NULL);
+    }
+
+    return 0;
+}
+
+
+const mtd_desc_t mtd_flashpage_driver = {
+    .init = _init,
+    .read = _read,
+    .write = _write,
+    .erase = _erase,
+};
--- a/drivers/include/mtd.h
+++ b/drivers/include/mtd.h
@@ -158,7 +158,7 @@ struct mtd_desc {
 int mtd_init(mtd_dev_t *mtd);
 
 /**
- * @brief   mtd_read Read data from a MTD device
+ * @brief   Read data from a MTD device
  *
  * No alignment is required on @p addr and @p count.
  *
@@ -177,10 +177,11 @@ int mtd_init(mtd_dev_t *mtd);
 int mtd_read(mtd_dev_t *mtd, void *dest, uint32_t addr, uint32_t count);
 
 /**
- * @brief   mtd_read write data to a MTD device
+ * @brief   Write data to a MTD device
  *
  * @p addr + @p count must be inside a page boundary. @p addr can be anywhere
- * but the buffer cannot overlap two pages.
+ * but the buffer cannot overlap two pages. Though some devices might enforce alignement
+ * on both @p addr and @p buf.
  *
  * @param      mtd   the device to write to
  * @param[in]  src   the buffer to write
@@ -194,11 +195,12 @@ int mtd_read(mtd_dev_t *mtd, void *dest, uint32_t addr, uint32_t count);
  * @return -EOVERFLOW if @p addr or @p count are not valid, i.e. outside memory,
  * or overlapping two pages
  * @return -EIO if I/O error occured
+ * @return -EINVAL if parameters are invalid (invalid alignment for instance)
  */
 int mtd_write(mtd_dev_t *mtd, const void *src, uint32_t addr, uint32_t count);
 
 /**
- * @brief   mtd_erase Erase sectors of a MTD device
+ * @brief   Erase sectors of a MTD device
  *
  * @p addr must be aligned on a sector boundary. @p count must be a multiple of a sector size.
  *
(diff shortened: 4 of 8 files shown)


drivers/Makefile.dep
```diff
@@ -227,16 +227,23 @@ ifneq (,$(filter mrf24j40,$(USEMODULE)))
   FEATURES_REQUIRED += periph_spi
 endif
 
-ifneq (,$(filter mtd_sdcard,$(USEMODULE)))
+ifneq (,$(filter mtd_%,$(USEMODULE)))
   USEMODULE += mtd
+endif
+
+ifneq (,$(filter mtd_sdcard,$(USEMODULE)))
```
aabadie: Could all mtd_ sub modules be nested in the generic block `mtd_%` ? This would avoid confusion I think.

vincent-d: done

drivers/include/mtd_flashpage.h
```diff
@@ -0,0 +1,50 @@
...
+extern "C"
+{
+#endif
+
+#define MTD_FLASHPAGE_INIT_VAL(_pages_per_sector) { \
+    .driver = &mtd_flashpage_driver, \
+    .sector_count = FLASHPAGE_NUMOF, \
+    .pages_per_sector = _pages_per_sector,\
+    .page_size = FLASHPAGE_SIZE / _pages_per_sector,\
+}
```
kYc0o: I'm not really sure if this is a good idea... What's the purpose of having this wrapper? The biggest problem I see is that you'd hardly know where you can write without precise knowledge about the non-filled flash space left when flashing a firmware... The main purpose of the flashpage driver is to write new program data, or extend the current one (new elf files, scripts of any kind, etc.), and, if I understand correctly, this driver aims to write user data...

vincent-d: The goal is to be able to use a file system on internal flash.

It's then completely up to the developer to be sure of what's he's doing and it might be dangerous, of course.

kYc0o: I see the point, but I have the issue of erasing. How do you cope with the small erases case? You need to always backup the whole block in RAM before adding something, then erase the whole block and write it from the buffer...

vincent-d: I don't understand your point here. This is the case with any flash, either internal or not.
A file system designed to run on a flash will not try to erase a few bytes, but always the entire sector. What is different here?

vincent-d: Most SPI flashes have a sector size of 4KB, or even 64KB. And a typical "page" size (i.e. maximum number of bytes you can write with a single instruction) is 256B on these flashes.

Then yes, we rely on the file system to manage erases and write cycles.

kYc0o: So, in overall (a bit out of scope now) using ANY file system over a flash-based MTD device will take *at least* as much RAM as the smallest erasable block is it? I just checked and for the N25Q128 series (the one attached to the iotlab-m3) has 4KB as eraseable sub-sector.

vincent-d: If you take the assumption that a sector will be modified, this will be true. But a good filesystem won't need to modify a byte in the middle of a sector. See littles or spiffs, they don't need a buffer of the size of a sector.

kYc0o:
Please squash so we have Murdock happy.

tests/unittests/tests-mtd_flashpage/tests-mtd_flashpage.h
```diff
@@ -0,0 +1,37 @@
...
+#define TESTS_MTD_FLASHPAGE_H
+
+#include "embUnit.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+    * @brief   The entry point of this test suite.
```
aabadie: Indentation is off in this doxygen block

drivers/include/mtd_flashpage.h
```diff
@@ -0,0 +1,50 @@
...
+#include "periph/flashpage.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#define MTD_FLASHPAGE_INIT_VAL(_pages_per_sector) { \
+    .driver = &mtd_flashpage_driver, \
+    .sector_count = FLASHPAGE_NUMOF, \
```
aabadie: So by default, IIUC the MTD can take the whole flash. Can we make this configurable at build time with a more reasonable default value (half of the flash) ?

aabadie: @vincent-d, do you think it would make sense to limit the default size used to half of the flash size ?

vincent-d: I took a new look to this and there is 2 options:
- we leave it as-is and rely on the configuration of the filesystem for the offset in the mtd (i.e. the mtd allows access to the whole internal flash and the filesystem is configured correctly to not write outside its area)
- we change the mtd_flashpage to let the filesystem access only a part of the internal flash, but we need to have a specific `mtd_flashpage_t` descriptor to store the offset

If we only change `sector_count` the filesystem will be able to access the number of sector passed through this field, but it will always start from address 0 (the first sector)

cladmi:
I was pointed here as I was writing a PR to limit the ROM used for a firmware: https://github.com/RIOT-OS/RIOT/pull/9351 It might be of your interest too.

tests/unittests/tests-mtd_flashpage/tests-mtd_flashpage.c
```diff
@@ -0,0 +1,128 @@
...
+    TEST_ASSERT_EQUAL_INT(0, ret);
+}
+
+static void test_mtd_erase(void)
+{
+    /* Erase last sector */
+    int ret = mtd_erase(dev, (FLASHPAGE_NUMOF - 1) * FLASHPAGE_SIZE, FLASHPAGE_SIZE);
+    TEST_ASSERT_EQUAL_INT(0, ret);
+
+    /* Erase with wrong size (les than sector size) */
```
aabadie: les**s**

drivers/mtd_flashpage/mtd_flashpage.c
```diff
@@ -0,0 +1,88 @@
...
+    (void)dev;
+    memcpy(buf, (void *)addr, size);
+
+    return size;
+}
+
+static int _write(mtd_dev_t *dev, const void *buf, uint32_t addr, uint32_t size)
+{
+    if (addr % FLASHPAGE_RAW_ALIGNMENT) {
+        return -EOVERFLOW;
```
aabadie: It doesn't look like an overflow, but rather like an invalid value.  Would returning -EINVAL make more sense here and in the check for `FLASHPAGE_RAW_ALIGNMENT` and `FLASHPAGE_RAW_BLOCKSIZE`.
The mtd API would need adaption though.

drivers/include/mtd_flashpage.h
aabadie: I'm not sure to understand everything clearly.

If the first option is kept, we'll have to trust the filesystem implementation and most of them come from external packages. My question in this case is how do you configure this external filesystem ? (I checked littlefs and I'm not sure to understand how to do that)

If the second option is kept, could the offset be automatically determined from the number of sectors defined (using offset =  flashsize - number_of_sector) ?

> he filesystem will be able to access the number of sector passed through this field, but it will always start from address 0 

Is this an issue ?

kYc0o:
Now that I think twice, I see that this PR worth the "risk" of letting the user use the internal flash at will.

I'd ACK it if @aabadie 's comments are addressed.

kYc0o:
@aabadie ping.

aabadie:
> @aabadie ping.

From what I see, I still have unaddressed comments. So it should be ping @vincent-d :)

tests/unittests/tests-mtd_flashpage/tests-mtd_flashpage.c
vincent-d: fixed

drivers/mtd_flashpage/mtd_flashpage.c
vincent-d: addressed

vincent-d:
@aabadie @kYc0o comments should be addressed

kYc0o:
Can we go on this one?

kYc0o:
@aabadie ping.

vincent-d:
squashed

aabadie:
Static tests are already reporting issues with a missing doxygen group

kYc0o:
I guess this PR needs rebase, it's failing on something about periph_flashpage driver internals.

kYc0o:
Murdock fails again... I think it might have something to do with dependencies, since the functions `_lock` and `_unlock` are factorised IIRC. Let me check to give more hints.

kYc0o:
I see now. `flashpage_write_raw` is not implemented for some stm32 devices, which are being tested by Murdock. Since for unittests some stm32 boards are in the whitelist, it gets compiled and fails.

I'd say that a dependency on `FEATURES_PROVIDED` would solve the problem.

vincent-d:
The problem is actually that on stm32, `periph_flash_common` is needed. But the dependency evaluation which pulls  `periph_flash_common` in is done in the `stm32_common/Makefile.include` which is included before the `drivers/Makefile.dep` evaluation.

So to summarize, `mtd_flashpage` requires `periph_flashpage` and `periph_flashpage_raw` from `drivers/Makefile.dep` but `periph_flashpage` requires `periph_flashpage_common` from `stm32_common/Makefile.include` before knowing `periph_flashpage` is required.

I tried a fix to pull `periph_flashpage_common` correctly from `stm32_common/Makefile.dep`

cpu/stm32_common/Makefile.dep
```diff
@@ -5,3 +5,9 @@ ifneq (,$(filter periph_i2c,$(USEMODULE)))
     USEMODULE += periph_i2c_2
   endif
 endif
+
+# flashpage and eeprom periph implementations share flash lock/unlock functions
+# in periph_flash_common
+ifneq (,$(filter periph_flashpage periph_eeprom,$(FEATURES_REQUIRED)))
+  USEMODULE += periph_flash_common
+endif
```
kYc0o: Well, this is perfect, but now it comes the annoying part... can you split this into a separated PR? BTW @cladmi just by chance aren't you dealing with this in one of your bunch of PRs?

vincent-d: See #10146

vincent-d:
Rebased since #10153 has been merged

kYc0o:
I'm testing this for nucleo-l152re without success. It fails while trying to erase the page in the unittest. I debugged a bit and found that the address where it wants to write is not valid, so the page. When trying to erase an invalid page the assert fails. @vincent-d can you check?

tests/unittests/tests-mtd_flashpage/tests-mtd_flashpage.c
```diff
@@ -0,0 +1,128 @@
...
+
+#include "mtd.h"
+#include "mtd_flashpage.h"
+
+static mtd_dev_t _dev = MTD_FLASHPAGE_INIT_VAL(8);
+static mtd_dev_t *dev = &_dev;
+
+static void setup_teardown(void)
+{
+    mtd_erase(dev, (FLASHPAGE_NUMOF - 1) * FLASHPAGE_SIZE, dev->pages_per_sector * dev->page_size);
```
kYc0o: I think the problem is here, this doesn't take into account `CPU_FLASH_BASE`, thus it tries to erase at an address with offset 0. This problem is here and below.

kYc0o: I think it should be `CPU_FLASH_BASE + ((FLASHPAGE_NUMOF - 1) * FLASHPAGE_SIZE)`. You may want to factor it out in a macro as it is extensively used.

kYc0o:
After some more debugging I think the problem is not that easy to solve, this PR would need a bit more work.

tests/unittests/tests-mtd_flashpage/tests-mtd_flashpage.c
```diff
@@ -0,0 +1,128 @@
...
+    TEST_ASSERT_EQUAL_INT(sizeof(buf_read), ret);
+    TEST_ASSERT_EQUAL_INT(0, memcmp(buf, buf_read, sizeof(buf)));
+    TEST_ASSERT_EQUAL_INT(0, memcmp(buf_empty, buf_read + sizeof(buf), sizeof(buf_empty)));
+
+    ret = mtd_erase(dev, (FLASHPAGE_NUMOF - 1) * FLASHPAGE_SIZE, dev->pages_per_sector * dev->page_size);
+    TEST_ASSERT_EQUAL_INT(0, ret);
+
+    /* Unaligned write / read */
+    ret = mtd_write(dev, buf, (FLASHPAGE_NUMOF - 1) * FLASHPAGE_SIZE + sizeof(buf_empty), sizeof(buf));
+    TEST_ASSERT_EQUAL_INT(sizeof(buf), ret);
```
kYc0o: Is this actually expected? I'm trying to fix this code since it doesn't work for several platforms but then I have some errors maybe related to the actual test.

I was expecting here a `-EINVAL`.

tests/unittests/tests-mtd_flashpage/tests-mtd_flashpage.c
```diff
@@ -0,0 +1,128 @@
...
+
+    ret = mtd_read(dev, buf_read, (FLASHPAGE_NUMOF - 1) * FLASHPAGE_SIZE, sizeof(buf_read));
+    TEST_ASSERT_EQUAL_INT(sizeof(buf_read), ret);
```
kYc0o: Here is more or less the same as above. However, in the read function there's no error nor check when the address is not aligned.

tests/unittests/tests-mtd_flashpage/tests-mtd_flashpage.c
```diff
@@ -0,0 +1,128 @@
...
+    TEST_ASSERT_EQUAL_INT(0, memcmp(buf, buf_read, sizeof(buf)));
+    TEST_ASSERT_EQUAL_INT(0, memcmp(buf_empty, buf_read + sizeof(buf), sizeof(buf_empty)));
+
+    ret = mtd_erase(dev, (FLASHPAGE_NUMOF - 1) * FLASHPAGE_SIZE, dev->pages_per_sector * dev->page_size);
+    TEST_ASSERT_EQUAL_INT(0, ret);
+
+    /* Unaligned write / read */
+    ret = mtd_write(dev, buf, (FLASHPAGE_NUMOF - 1) * FLASHPAGE_SIZE + sizeof(buf_empty), sizeof(buf));
```
kYc0o: What makes the write unaligned?

tests/unittests/tests-mtd_flashpage/tests-mtd_flashpage.c
kYc0o: After searching around, I found that ARM devices can read from unaligned addresses, thus this test might not make sense. However, I don't know if other architectures are supposed to be compatible with this driver, though it relies on the flashpage_raw module, which for now is only supported on ARM.

gebart: Cortex-M0 will hard fault when you try to word read unaligned memory locations. CM3/4 will not.

kYc0o: Oh, weird... a stm32f072rb was not hard faulting. Anyways, the code to send the error is missing, as well as the right platform definitions about align size. By the way, how those read addresses should be aligned?

kYc0o:
@vincent-d  I just pushed some fixups to your branch with some fixes. Let me know if you agree with them.

kYc0o:
@vincent-d ping.

vincent-d:
@kYc0o your changes look OK. Maybe someone else should take a look and test (I don't have hardware supporting this at the moment).

kYc0o:
Maybe @fjmolinas would like to test? The failures and differences are mostly related to stm32l152 devices, and other cortex-m series uC which I also don't have now.

aabadie:
I can test tomorrow on nucleo-l1 and some other boards.

aabadie:
It would also be great to rebase this PR on top of latest master as it contains other changes for unittests and nucleo-l1.

vincent-d:
I guess this one only needs squashing now, @aabadie ?

kYc0o:
@aabadie did you test on a stm21l152?

aabadie:
> did you test on a stm21l152?

Not the very last version (I admit I forgot...), but I tested it before (see https://github.com/RIOT-OS/RIOT/pull/8774#pullrequestreview-278405786) the test was cut out from the unittest and it was working.

kYc0o:
Cold you try again nonetheless? I really have doubts about the differences on those platforms.

drivers/include/mtd_flashpage.h
```diff
@@ -0,0 +1,53 @@
...
+#endif
+
+/**
+ * @brief   Macro helper to initialize a mtd_t with flash-age driver
+ */
+#define MTD_FLASHPAGE_INIT_VAL(_pages_per_sector) { \
+    .driver = &mtd_flashpage_driver, \
+    .sector_count = FLASHPAGE_NUMOF, \
+    .pages_per_sector = _pages_per_sector,\
+    .page_size = FLASHPAGE_SIZE / _pages_per_sector,\
```
benpicco, line 40: What is the reason to create 'virtual' sectors here?
Usually page is the max transfer unit and sector the min erase unit. For internal flash they tend to be the same. So why artificially make the erase granularity worse?

PR #8775, opened 2018-03-13
periph: hwcrypto: initial draft for hardware crypto support (version 2)

basilfx:
### Contribution description

This is my second attempt to define a peripheral interface for hardware crypto devices. In this version, hardware accelerated (block) ciphers and hashing algorithms are supported.

It is a complete rewrite of #5429:

* Dropped the context parameter (peripherals should define state if needed)
* Added support for multiple hardware crypto peripherals (EFM32 has support)
* Removed long list of defines
* Removed the weird include structure

I have also added a peripheral test and a working implementation for the EFM32.

### Issues/PRs references

#5429 (first version)

### DIFF
--- a/drivers/include/periph/hwcrypto.h
+++ b/drivers/include/periph/hwcrypto.h
@@ -1,0 +1,394 @@
+/*
+ * Copyright (C) 2016-2018 Bas Stottelaar <basstottelaar@gmail.com>
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_periph_hwcrypt Hardware Crypto
+ * @ingroup     drivers_periph
+ * @brief       Low-level hardware crypto peripheral driver
+ *
+ * Modern CPUs may be equipped with a hardware crypto peripheral to speed up
+ * cryptographical operations such as (as)symmetric ciphers and hashing
+ * algorithms. With no or little intervention, the CPU can typically execute
+ * more operations than the software alternative.
+ *
+ * This driver provides an abstract for the different ciphers and hashes that
+ * may be supported. Using HAVE_HWCRYPTO_xxx defines, software may choose to
+ * use the hardware crypto alternative, or fall back to a software variant if
+ * not defined.
+ *
+ * Current interface is optimized for (as)symmetric block ciphers and block
+ * hashes. It supports multiple hardware crypto peripherals, including limited
+ * support.
+ *
+ * All operations block until they are finished. Operations may change the
+ * internal peripheral state (e.g. load the key, keep the digest interally),
+ * therefore must be used to control mutual exclusive access.
+ *
+ * @{
+ * @file
+ * @brief       Low-level hardware crypto driver interface definitions
+ *
+ * @author      Bas Stottelaar <basstottelaar@gmail.com>
+ */
+
+#ifndef HWCRYPTO_H
+#define HWCRYPTO_H
+
+#include <stdint.h>
+
+#include "periph_cpu.h"
+#include "periph_conf.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef HAVE_HWCRYPTO_T
+/**
+ * @brief   Hardware crypto type identifier
+ */
+typedef unsigned int hwcrypto_t;
+#endif
+
+/**
+ * @brief   Supported ciphers.
+ *
+ * Define HAVE_HWCRYPTO_xxx in periph_cpu.h for each supported cipher.
+ *
+ * Extend at will, but also update drivers/periph_common/hwcrypto.c.
+ */
+typedef enum {
+    HWCRYPTO_AES128,                  /**< AES128 cipher */
+    HWCRYPTO_AES192,                  /**< AES192 cipher */
+    HWCRYPTO_AES256,                  /**< AES256 cipher */
+    HWCRYPTO_DES,                     /**< DES cipher */
+    HWCRYPTO_3DES,                    /**< 3DES cipher */
+    HWCRYPTO_TWOFISH,                 /**< Two-fish cipher */
+    HWCRYPTO_RSA512,                  /**< RSA512 cipher */
+    HWCRYPTO_RSA1024,                 /**< RSA1024 cipher */
+    HWCRYPTO_RSA2048,                 /**< RSA2048 cipher */
+    HWCRYPTO_RSA4096,                 /**< RSA4096 cipher */
+    HWCRYPTO_ECC128,                  /**< ECC128 cipher */
+    HWCRYPTO_ECC192,                  /**< ECC192 cipher */
+    HWCRYPTO_ECC256,                  /**< ECC256 cipher */
+    HWCRYPT_CIPHER_NUMOF              /**< number of ciphers available */
+} hwcrypto_cipher_t;
+
+/**
+ * @brief   Supported hashing algorithms.
+ *
+ * Define HAVE_HWCRYPTO_xxx in periph_cpu.h for each supported hashing
+ * algorithm.
+ *
+ * Extend at will, but also update drivers/periph_common/hwcrypto.c.
+ */
+typedef enum {
+    HWCRYPTO_MD4,                     /**< MD4 hashing algorithm */
+    HWCRYPTO_MD5,                     /**< MD5 hashing algorithm */
+    HWCRYPTO_SHA1,                    /**< SHA1 hashing algorithm */
+    HWCRYPTO_SHA3,                    /**< SHA3 hashing algorithm */
+    HWCRYPTO_SHA224,                  /**< SHA224 hashing algorithm */
+    HWCRYPTO_SHA256,                  /**< SHA256 hashing algorithm */
+    HWCRYPTO_SHA384,                  /**< SHA384 hashing algorithm */
+    HWCRYPTO_SHA512,                  /**< SHA512 hashing algorithm */
+    HWCRYPT_HASH_NUMOF                /**< number of hashes available */
+} hwcrypto_hash_t;
+
+/**
+ * @brief   Setup options for cipher and hash methods.
+ *
+ * Extend at will.
+ */
+typedef enum {
+    HWCRYPTO_OPT_KEY,
+    HWCRYPTO_OPT_IV,
+    HWCRYPTO_OPT_COUNTER,
+    HWCRYPTO_OPT_PADDING,
+} hwcrypto_opt_t;
+
+/**
+ * @brief   Supported cipher modes of operation, to be used with
+ *          @p HWCRYPTO_OPT_MODE.
+ *
+ * Extend at will.
+ */
+typedef enum {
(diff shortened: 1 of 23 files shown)


drivers/include/periph/hwcrypto.h
```diff
@@ -0,0 +1,394 @@
...
+ * @return                  -1 on error
+ */
+int hwcrypto_release(hwcrypto_t dev);
+
+/**
+ * @brief   Power on the hardware crypto peripheral.
+ *
+ * @param[in] dev           the device to initialize
+ */
+void hwcrypto_poweron(hwcrypto_t dev);
```
gebart: Do you need this function? Does the hwcrypto device ever need to run asynchronously or in the background? Otherwise, I suggest that turning on and off the power should be handled automatically by the other functions

vincent-d: Maybe do the power on/off operations in `hwcrypto_acquire`/`hwcrypto_release`?

drivers/include/periph/hwcrypto.h
```diff
@@ -0,0 +1,394 @@
...
+ *
+ * @return                  nonzero value if supported
+ * @return                  0 if unsupported
+ */
+int hwcrypto_hash_supported(hwcrypto_t dev, hwcrypto_hash_t hash);
+
+/**
+ * @brief   Initialize the hardware crypto peripheral.
+ *
+ * The peripheral should be turned on in this method.
```
vincent-d: Is it really required?

I have the impression this could be called from `periph_init` as spi or rtc can be initialized. So I would suggest the periph is not on after the call of this function, but setup to be ready for later usage. This is somehow the same discussion as for power on/off function, I think.

drivers/include/periph/hwcrypto.h
```diff
@@ -0,0 +1,394 @@
...
+ *
+ * @param[in] dev           the device to initialize
+ *
+ * @return                  0 on success
+ * @return                  -1 if an error occurs
```
vincent-d: An anonymous enum would look more consistant with other periph interfaces.

drivers/periph_common/hwcrypto.c
```diff
@@ -0,0 +1,114 @@
...
+ * @}
+ */
+
+#include "periph/hwcrypto.h"
+
+#ifndef PERIPH_HWCRYPTO_PROVIDES_CIPHER_SUPPORTED
+/**
+ * @brief Bitmap of supported ciphers.
+ */
+static uint32_t cipher_bitmap = (
```
vincent-d: `const`?

drivers/periph_common/hwcrypto.c
```diff
@@ -0,0 +1,114 @@
...
+
+    return cipher_bitmap & (1 << cipher);
+}
+#endif /* PERIPH_HWCRYPTO_PROVIDES_CIPHER_SUPPORTED */
+
+#ifndef PERIPH_HWCRYPTO_PROVIDES_HASH_SUPPORTED
+/**
+ * @brief Bitmap of supported hashes.
+ */
+static uint32_t hash_bitmap = (
```
vincent-d: `const`?

tests/periph_hwcrypto/README.md
```diff
@@ -0,0 +1,8 @@
+Expected result
```
basilfx: Note to self to fix this

drivers/include/periph/hwcrypto.h
basilfx: I will now move them to `hwcrypto_acquire`/`hwcrypto_release`.

If we find a use-case than it should be easy to implement them at a later stage.

drivers/include/periph/hwcrypto.h
basilfx: Let's move it to `periph_init`

drivers/include/periph/hwcrypto.h
basilfx: Fixed.

drivers/periph_common/hwcrypto.c
basilfx: Fixed.

drivers/periph_common/hwcrypto.c
basilfx: Fixed.

tests/periph_hwcrypto/README.md
basilfx: Fixed.

drivers/include/periph/hwcrypto.h
```diff
@@ -0,0 +1,401 @@
...
+ *
+ * In case the peripheral is busy, this function will block until the
+ * peripheral is available again.
+ *
+ * @param[in] dev           the device to initialize
+ *
+ * @return                  HWCRYPTO_OK on success
+ * @return                  HWCRYPTO_ERROR on error
+ */
+int hwcrypto_acquire(hwcrypto_t dev);
```
vincent-d: This could be `void hwcrypto_acquire(hwcrypto_t dev)`, same for release.

This would make it consistent with `spi_acquire/release`

vincent-d:
I'll tend to ACK this one, though I still have a minor comment.

Does someone have more comments? @gebart?

Needs rebase, btw.

bergzand:
Maybe a bit late, but I'd like to have a look at the API proposed here. I have some time this evening during the Hack'n'ack

benpicco:
@basilfx ping

I'd like to contribute an implementation for samd5x

tcschmidt:
@basilfx @benpicco @PeterKietzmann how are we going to proceed here? This seems stuck ...

PeterKietzmann:
Together with @Einhornhool and @leandrolanzieri we are currently working on implementing different hardware crypto peripherals (on the KW22 at first) against the collection of crypto APIs that already exist in RIOT (SHA-1, SHA-256, AES-128, ...), and we configure the software/hardware implementation using Kconfig.
I really appreciate @basilfx's work here but to honest, I don't see why we would need a centralized hwcrypto API for all kinds of primitives. If you want to calculate a SHA-1 hash, then you call the SHA-1 hash function. Dependant on the underlying hardware, it gives you the best implementation (hardware or software). Isn't it as easy as this or do I miss something?

drivers/periph_common/hwcrypto.c
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+#include "periph/hwcrypto.h"
+
+#ifndef PERIPH_HWCRYPTO_PROVIDES_CIPHER_SUPPORTED
+/**
+ * @brief Bitmap of supported ciphers.
+ */
+static const uint32_t cipher_bitmap = (
+#ifdef HAVE_HWCRYPTO_AES128
```
kaspar030, line 28: I propose using `|` instead of `+` for chaining these. The error case (duplicate entry or meaning) is less ugly...

kaspar030:
> Together with @Einhornhool and @leandrolanzieri we are currently working on implementing different hardware crypto peripherals (on the KW22 at first) against the collection of crypto APIs that already exist in RIOT (SHA-1, SHA-256, AES-128, ...), and we configure the software/hardware implementation using Kconfig.

> I really appreciate @basilfx's work here but to honest, I don't see why we would need a centralized hwcrypto API for all kinds of primitives.

So you started some alternative without raising concerns or commenting on previous work that is almost four years old? That's a bit odd.

> Isn't it as easy as this or do I miss something?

Well, I guess the way you're describing it, each implementation would be independent, not taking similarities of different hw implementations or capabilities into account. Concurrent access (of probably different modes), initialization, mapping to the actual APIs etc. would have to be done N times. That could be a nicer way, but my gut feeling is that the different crypto hardwares are actually quite similar in their capabilities and more important, restrictions. I see missed opportunities for code sharing. But I guess we have to judge that when you show some code. Which IMO you "Together with @Einhornhool and @leandrolanzieri" should do rather sooner than later, as with your announcement you've basically blocked @basilfx's contribution.

PeterKietzmann:
First of all, I don't understand the emotional reaction. We want to find a good solution to this problem that works well with current platforms and tools in RIOT. Furthermore, I did not intend to dismiss this PR unconditionally and it seems we are going into the discussion right now, which is good.

> So you started some alternative without raising concerns or commenting on previous work that is almost four years old? That's a bit odd.

- It is not just an alternative: (i) we need the hardware specific code anyway and (ii) for us it acts also as a proof of concept for using Kconfig.
- If I remember correctly it was you who motivated to have different implementations for new features to figure out the best of both worlds.
- The lessons learned and results of our work are needed to legitimate us commenting on the technical level that this PR is. And to be honest, we are not yet there.

> not taking similarities of different hw implementations or capabilities into account

Are you talking about different MCUs or different features on the same MCU. In the former case you might be right, although I simply presumed that peripheral drivers stay MCU specific. Furthermore, where does this PR share the code between different MCU drivers? In the latter case, there can always be a common component that is used by different peripheral drivers.

> Which IMO you "Together with @Einhornhool and @leandrolanzieri" should do rather sooner than later, 

I think this comment is misplaced. 

> as with your announcement you've basically blocked @basilfx's contribution.

This PR is two years old and its predecessor is four. There hasn't been much review and it has been quiet for a long time. I would argue that it doesn't hurt to wait a bit longer, since there hasn't been much drive. But in any way, I'm not blocking the PR as long as I haven't carefully reviewed and evaluated the details.

benpicco:
> I really appreciate @basilfx's work here but to honest, I don't see why we would need a centralized hwcrypto API for all kinds of primitives. If you want to calculate a SHA-1 hash, then you call the SHA-1 hash function. Dependant on the underlying hardware, it gives you the best implementation (hardware or software). Isn't it as easy as this or do I miss something?

Why not both?
You  have an internal hardware crypto API and on top a user facing crypto API.
If the hardware implements the primitive, you'd just statically inline call that, or do some massaging around it to make the user facing API nice / keep the low-level API low level.

Much like with the pagewise MTD API where the low-level driver only has to provide a `.write_page()` function and the user only has to call a `mtd_write()` function and the MTD API will take care of splitting the buffer into pages.

If the hardware does not implement the feature, as software implementation would be used instead (this can be done at compile time).

drivers/include/periph/hwcrypto.h
```diff
@@ -0,0 +1,401 @@
...
+ * @return                  HWCRYPTO_NOTSUP if option not supported or
+ *                          applicable
+ * @return                  HWCRYPTO_INVALID if value is not supported or
+ *                          applicable
+ * @return                  HWCRYPTO_ERROR if other error
+ */
+int hwcrypto_cipher_set(hwcrypto_t dev,
+                        hwcrypto_opt_t option,
+                        const void* value,
+                        uint32_t size);
```
maribu, line 252: Why not having separate functions for distinct operations, e.g.?

1. `int hwcrypto_cipher_key(hwcrypto_t dev, const void *key, const void *iv);`
2. `int hwcrypto_cipher_padding(hwcrypto_t dev, uint8_t padding);`
3. `int hwcrypto_cipher_counter(hwcrypto_t dev, uint64_t nonce, uint64_t counter);`

maribu (commented):
The API looks quite good to me. Some comments on it inline.

How about splitting this up, so that the API and the implementations can be reviewed independently?

PR #8777, opened 2018-03-14
tests/unittest/tests-tweetnacl: increase main stacksize

kaspar030:
### Contribution description

#8667 increased the unittests stack requirements, causing samr21-xpro to crash.
This PR increases the stacksize accordingly.

### Issues/PRs references

#8776


bergzand:
Just to be sure, is this going to cause problems with the smaller (memory-wise) boards?

kaspar030:
> is this going to cause problems with the smaller (memory-wise) boards?

As the main stack is defined statically, that *should* be caught at compile time...

PR #8784, opened 2018-03-15
sys/shell: switch stdout to unbuffered mode

girtsf:
Without this, if there is enough memory, newlib will enable
line-buffered mode by default and allocate 1024 byte buffer. Any
characters echoed by the shell would keep accumulating in the buffer and
only get "echoed" back when ENTER is pressed or command line overflows.

Tested: verified that behavior has not changed on a nucleo32-f031, and that echo now works properly on the board I'm working with.

### DIFF
--- a/sys/shell/shell.c
+++ b/sys/shell/shell.c
@@ -282,6 +282,13 @@ static inline void print_prompt(void)
 
 void shell_run(const shell_command_t *shell_commands, char *line_buf, int len)
 {
+#ifndef SHELL_NO_ECHO
+    /* If echo is enabled, switch stdout to unbuffered mode. Otherwise we'll
+     * buffer output characters and only echo them back after ENTER has been
+     * pressed. */
+    setvbuf(stdout, NULL, _IONBF, 0);
+#endif
+
     print_prompt();
 
     while (1) {


girtsf:
Let me know if you think there's a better way for this, or if I've missed something obvious in how to do this better (other than just setting this from my application).

gebart:
Does this apply only to newlib without -specs=nano.specs or does newlib-nano really allocate a 1024 byte buffer?

PeterKietzmann:
@girtsf 

> Does this apply only to newlib without -specs=nano.specs or does newlib-nano really allocate a 1024 byte buffer?

girtsf:
Apologies @gebart, I haven't played with RIOT in a while, so I'm afraid I won't be able to answer your question. Feel free to drop the PR.

PeterKietzmann:
@jcarrano IIRC you've been working with our shell internals recently. Would you mind to check if this PR is still useful or whether we should close it?

jcarrano:
@PeterKietzmann I will take a look. If it works, then I'd rather have it as the default for all of RIOT, not only the shell. The main reason for output line buffering is to minimize write() calls and thus system calls, which are expensive (the buffering is done in userspace until it is time to flush). Given that RIOT does not have the notion of kernel/userspace separation and no system calls, buffering is not necessary.*

(*) there is still quite a chain of function calls leading from write to the final destination in the uart buffer, but that's another story.

PR #8787, opened 2018-03-16
kinetis: Avoid reinitializing RTT if running

gebart:
### Contribution description

The RTC hardware is only reset on power on reset, this avoids
accidentally clearing the RTC counter if rtt_init is called when the RTC
is already running.

We check the Time Invalid Flag to see if the RTC hardware needs initialization

### Issues/PRs references

none


gebart:
Replaced by #8930

PR #8467, opened 2018-01-27
pkg: add cn-cbor CBOR implementation plus unit tests

lorenz9314:
### Contribution description

Adds the CBOR implementation cn-cbor to RIOT including unit tests. The test cases have been partially taken from cn-cbors original unit tests.

### Issues/PRs references

The package itself has been taken from #8025, commits 772e8778665aa1214996d7efe7f54b88d60925e0, fd7b6ab7343a3e91bd48bfc4d18e668c42572758 and ae2c180a0fd2282cd4419e66cea7d4b14892d3f6

PR #8791, opened 2018-03-16
pkg: cncbor: misc fixes

kaspar030:
### Contribution description

Testing in #8467 was limited to native. This PR fixes some problems on ARM:

1. cn-cbor uses unaligned ntoh\[sl\](), causing hard faults on ARM. -> added an alignment safe implementation

2. the unittests parse_hex() had faulty malloc+sscanf. replaced with fmt.

### Issues/PRs references

#8467. @

### DIFF
--- a/pkg/cn-cbor/Makefile.dep
+++ b/pkg/cn-cbor/Makefile.dep
@@ -1,0 +1,1 @@
+USEMODULE += posix
--- a/pkg/cn-cbor/patches/0001-implement-alignment-safe-ntoh16p-ntoh32p.patch
+++ b/pkg/cn-cbor/patches/0001-implement-alignment-safe-ntoh16p-ntoh32p.patch
@@ -1,0 +1,37 @@
+From a8ad896613571b16c53d9915f7b413fae7ea2879 Mon Sep 17 00:00:00 2001
+From: Kaspar Schleiser <kaspar@schleiser.de>
+Date: Fri, 16 Mar 2018 15:33:00 +0100
+Subject: [PATCH] implement alignment-safe ntoh16p() && ntoh32p()
+
+---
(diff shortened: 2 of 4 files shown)


kaspar030:
@lorenz9314 can you take a look at this?

nmeum:
Did you consider also submitting these patches upstream?

kaspar030:
> Did you consider also submitting these patches upstream?

The original code has a comment like "if you cannot do unaligned accesses, fix these:", so I assumed upstream is not interested. Then again, as is, noone ever used that code on ARM before...

gebart:
I see that upstream code comment as more of a "feel free to provide a PR if you need alignment", not a wontfix

kaspar030:
> I see that upstream code comment as more of a "feel free to provide a PR if you need alignment", not a wontfix

https://github.com/cabo/cn-cbor/pull/47

lorenz9314:
@kaspar030 Yep, sure. I pretty much took the parse_hex() function from the original cn-cbor tests. This however seems much more elegant (and functioning) anyway.

kYc0o:
Got the same problem while testing cbor related things.

Should we wait for upstream merge?

kYc0o:
AND, we didn't spot this because Murdock doesn't build unittests separately, and while building the whole unittests the module `posix` is added for other tests, so it gets magically imported.

I think we should provide a fix for that. I think @kaspar030 was working already on something like that while doing #8801 ?

kaspar030:
I suggest we merge this, upstreaming might take time. We can remove the patch easily later. Also, this PR includes some other fixes to the package.

kaspar030:
- squashed

kYc0o:
Then go!

PR #8047, opened 2017-11-15
boards: ikea-tradfri: add support

basilfx:
Last week I found out that the [IKEA TRÅDFRI](http://www.ikea.com/us/en/catalog/categories/departments/lighting/36812/) lighting products and accessories use the Silicon Labs EZR32MG1P microcontroller. Let it be a coincidence that the Thunderboard Sense (#7929) which was merged this week, uses the same CPU!

Now if we look into their line of products, it happens that they use the same CPU module in all(?) of their products: [here](https://www.heise.de/make/artikel/Ikea-Tradfri-Anleitung-fuer-ein-ESP8266-Lampen-Gateway-3598411.html), [here](http://tradfri.blogspot.nl/2017/04/teardown-of-ikea-tradfri-bulb-led1623g12.html) and [here](https://github.com/basilfx/TRADFRI-Hacking). Nonetheless, you can isolate them from the product, and use them for your own ideas, or leave them in and flash a custom firmware. 

The MCU is a ARM Cortex M4 with 256 kB of flash. It's quite packed (including hardware crypto), flexible pin mapping and has an integrated radio which supports Thread, ZigBee Light Link and more. 

It won't surprise me if people buy the cheapest light available (the GU-10 I used costed me 6,99 euro) and upgrade, hack or repurpose them. Let RIOT-OS be one of the first to support them :-)

This PR is a quick port to this new platform. I have more information on the module over [here](https://github.com/basilfx/TRADFRI-Hacking) and a demonstration video [here](https://www.youtube.com/watch?v=yi_Z2WtmdDU). For flashing you need to connect a JTAG/SWD compatible debugger, but I can flash it the exact same way as the Thunderboard Sense when I use SEGGER JLink. You do not even have to remove the module.

PR #8794, opened 2018-03-16
boards: efm32: adapt pr feedback for other boards

basilfx:
### Contribution description

During the review of #8630 some comments were raised that have slipped through. This PR adapts these comments for the:

* `ikea-tradfri`
* `sltb001a`
* `stk3600`
* `stk3700`

### Issues/PRs references

#7929
#8047  
#8585
#8520

### DIFF
--- a/boards/sltb001a/board.c
+++ b/boards/sltb001a/board.c
@@ -34,23 +34,23 @@ void board_init(void)
     board_common_init();
 
 #ifdef MODULE_SILABS_PIC
-    /* enable the CCS811 air quality/gas sensor */
 #if CCS811_ENABLED
+    /* enable the CCS811 air quality/gas sensor */
     pic_write(CCS811_PIC_ADDR, (1 << CCS811_PIC_EN_BIT) | (1 << CCS811_PIC_WAKE_BIT));
 #endif
 
-    /* enable the IMU sensor */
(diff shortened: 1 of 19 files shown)


basilfx:
For what its worth, this is the proof that it still runs after 4ff507a 🤓

SLTB001a:
schermafbeelding 2018-03-16 om 20 20 20

SLSTK3401a:
schermafbeelding 2018-03-16 om 20 23 44

boards/sltb001a/board.c
```diff
@@ -21,9 +21,8 @@
 
 #include "board.h"
 #include "board_common.h"
-#include "pic.h"
-
 #include "periph/gpio.h"
+#include "pic.h"
```
aabadie: Maybe this include could be protected with a `#ifdef MODULE_SILABS_PIC` ?

basilfx: I can add that, but it compiles fine without (if I remove `silabs_pic` from the `USEMODULES`).

Let me know what you prefer.

basilfx: On second thought: I'll add it. That's more similar to how we also include optional peripherals.

basilfx:
Added the conditional includes.

PR #8795, opened 2018-03-18
newlib: link without -lnosys

gebart:
### Contribution description

lnosys and nosys.specs is mainly intended to provide empty stubs for syscalls, which we do not want since we implement the syscalls ourselves.

### Issues/PRs references

none


kaspar030:
Alright, basic samr21-xpro tests still work. ACK.

PR #8798, opened 2018-03-19
pkg/semtech-loramac: refactor API and make it thread-safe

aabadie:
### Contribution description

This PR is a refactoring of the Semtech loramac package: the idea was initially to introduce a dedicated struct for handling the mac in a thread-safe way. The immediate consequence is a global API change of the package but thanks to this, any access to the internal mac can be protected using mutexes and data exchange is simplified between the upstream code and the RIOT adaption code (using ipc messages).
I also went a little further by refactoring other things:
- remove useless static structures => saves a lot of memory
- remove access from the thread handler function to static variables defined in the loramac upstream code
- decouple the send from the receive => in the long run, it should be possible to adapt this package to the netif interface/thread
- adapt the test application and the documentation

There might things I overlooked so comments and tests are very welcome.

cc @jia200x and @dylad who might be interested.

### Issues/PRs references

None

### DIFF
--- a/pkg/semtech-loramac/contrib/semtech_loramac.c
+++ b/pkg/semtech-loramac/contrib/semtech_loramac.c
@@ -28,6 +28,7 @@
 #include <string.h>
 
 #include "msg.h"
+#include "mutex.h"
 
 #include "net/netdev.h"
 #include "net/loramac.h"
@@ -63,30 +64,21 @@
 
 #define SEMTECH_LORAMAC_MSG_QUEUE                   (16U)
 #define SEMTECH_LORAMAC_LORAMAC_STACKSIZE           (THREAD_STACKSIZE_DEFAULT)
+static msg_t _semtech_loramac_msg_queue[SEMTECH_LORAMAC_MSG_QUEUE];
 static char _semtech_loramac_stack[SEMTECH_LORAMAC_LORAMAC_STACKSIZE];
 kernel_pid_t semtech_loramac_pid;
-kernel_pid_t semtech_loramac_handler_pid;
 
+sx127x_t sx127x;
 RadioEvents_t semtech_loramac_radio_events;
-uint8_t semtech_loramac_dev_eui[LORAMAC_DEVEUI_LEN];
-uint8_t semtech_loramac_app_eui[LORAMAC_APPEUI_LEN];
-uint8_t semtech_loramac_app_key[LORAMAC_APPKEY_LEN];
-uint8_t semtech_loramac_nwk_skey[LORAMAC_NWKSKEY_LEN];
-uint8_t semtech_loramac_app_skey[LORAMAC_APPSKEY_LEN];
-uint8_t semtech_loramac_dev_addr[LORAMAC_DEVADDR_LEN];
-
-static uint8_t _semtech_loramac_radio_payload[SX127X_RX_BUFFER_SIZE];
-static semtech_loramac_rx_data_t _semtech_loramac_rx_data;
+LoRaMacPrimitives_t semtech_loramac_primitives;
+LoRaMacCallback_t semtech_loramac_callbacks;
 
 typedef struct {
-    uint8_t port;
-    uint8_t cnf;
-    uint8_t dr;
     uint8_t *payload;
     uint8_t len;
 } loramac_send_params_t;
 
-typedef void (*semtech_loramac_func_t)(void *);
+typedef void (*semtech_loramac_func_t)(semtech_loramac_t *, void *);
 
 /**
  * @brief   Struct containing a semtech loramac function call
@@ -99,12 +91,13 @@ typedef struct {
 } semtech_loramac_call_t;
 
 /* Prepares the payload of the frame */
-static bool _semtech_loramac_send(uint8_t cnf, uint8_t port, uint8_t dr,
+static bool _semtech_loramac_send(semtech_loramac_t *mac,
                                   uint8_t *payload, uint8_t len)
 {
     DEBUG("[semtech-loramac] send frame %s\n", (char *)payload);
     McpsReq_t mcpsReq;
     LoRaMacTxInfo_t txInfo;
+    uint8_t dr = semtech_loramac_get_dr(mac);
 
     if (LoRaMacQueryTxPossible(len, &txInfo) != LORAMAC_STATUS_OK) {
         DEBUG("[semtech-loramac] empty frame in order to flush MAC commands\n");
@@ -115,10 +108,10 @@ static bool _semtech_loramac_send(uint8_t cnf, uint8_t port, uint8_t dr,
         mcpsReq.Req.Unconfirmed.Datarate = (int8_t)dr;
     }
     else {
-        if (cnf == LORAMAC_TX_UNCNF) {
+        if (mac->cnf == LORAMAC_TX_UNCNF) {
             DEBUG("[semtech-loramac] MCPS_UNCONFIRMED\n");
             mcpsReq.Type = MCPS_UNCONFIRMED;
-            mcpsReq.Req.Unconfirmed.fPort = port;
+            mcpsReq.Req.Unconfirmed.fPort = mac->port;
             mcpsReq.Req.Unconfirmed.fBuffer = payload;
             mcpsReq.Req.Unconfirmed.fBufferSize = len;
             mcpsReq.Req.Unconfirmed.Datarate = (int8_t)dr;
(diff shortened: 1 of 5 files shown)


jia200x:
Hi @aabadie 

Nice approach!
I only have some comments in the thread-safe part. I will do a quick review.

tests/pkg_semtech-loramac/main.c
```diff
@@ -76,49 +74,49 @@ static int _cmd_loramac(int argc, char **argv)
 
         if (strcmp("deveui", argv[2]) == 0) {
             uint8_t deveui[LORAMAC_DEVEUI_LEN];
-            semtech_loramac_get_deveui(deveui);
+            semtech_loramac_get_deveui(&loramac, deveui);
```
jia200x: I would have preferred here the usage of IPC rather than direct calls + mutex. IMO:
- If there's a single entry point for all commands, there shouldn't be needs for mutex (e.g [here](https://github.com/RIOT-OS/RIOT/blob/master/pkg/openthread/contrib/netdev/openthread_netdev.c#L122))
- In can be integrated to netapi, since most of these commands are network options.

What do you think?

jia200x: Basically, this means reusing MSG_TYPE_LORAMAC_CMD

pkg/semtech-loramac/contrib/semtech_loramac.c
```diff
@@ -63,30 +64,21 @@
 
 #define SEMTECH_LORAMAC_MSG_QUEUE                   (16U)
 #define SEMTECH_LORAMAC_LORAMAC_STACKSIZE           (THREAD_STACKSIZE_DEFAULT)
+static msg_t _semtech_loramac_msg_queue[SEMTECH_LORAMAC_MSG_QUEUE];
 static char _semtech_loramac_stack[SEMTECH_LORAMAC_LORAMAC_STACKSIZE];
 kernel_pid_t semtech_loramac_pid;
-kernel_pid_t semtech_loramac_handler_pid;
 
+sx127x_t sx127x;
```
jia200x: Do you think it could make sense to also isolate these kind of variables in a struct?

jia200x: (e.g this could come from the netdev_t pointer)

tests/pkg_semtech-loramac/main.c
aabadie: I understand what you mean and I avoided it on purpose. That would have required other internal get/set functions, with switch/case depending on the parameters, very similar to netdev.

The actual approach is simpler and I think it ends up to the same result.

If the semtech-loramac API is later integrated into netif, then it will be possible to call it in the netapi. I need to rework [my first attempt](https://github.com/aabadie/RIOT/tree/pkg_semtech-loramac-netif) to achieve this.

pkg/semtech-loramac/contrib/semtech_loramac.c
aabadie: The problem is that it is used by the [radio adaption code](https://github.com/RIOT-OS/RIOT/blob/master/pkg/semtech-loramac/contrib/semtech_loramac_radio.c#L36) as an `extern` variable. Using a get/set function for passing/retrieving the pointer as you initially did won't be better I think.
The best solution would be to pass it the all semtech-loramac function but then it will require massive patches in the upstream code.

pkg/semtech-loramac/contrib/semtech_loramac_getset.c
```diff
@@ -27,217 +29,303 @@
...
     MibRequestConfirm_t mibReq;
     mibReq.Type = MIB_RX2_DEFAULT_CHANNEL;
     LoRaMacMibGetRequestConfirm(&mibReq);
-    return mibReq.Param.Rx2DefaultChannel.Datarate;
+    datarate = mibReq.Param.Rx2DefaultChannel.Datarate;
+    mutex_unlock(&mac->lock);
+    return datarate;
+}
+
+void semtech_loramac_set_rx2_params(semtech_loramac_t *mac,
```
dylad: I have a compilation error here because `semtech_loramac_t *mac` argument is not used.

aabadie: Fixed

pkg/semtech-loramac/contrib/semtech_loramac_getset.c
```diff
@@ -27,217 +29,302 @@
...
     MibRequestConfirm_t mibReq;
     mibReq.Type = MIB_RX2_DEFAULT_CHANNEL;
     LoRaMacMibGetRequestConfirm(&mibReq);
-    return mibReq.Param.Rx2DefaultChannel.Datarate;
+    datarate = mibReq.Param.Rx2DefaultChannel.Datarate;
+    mutex_unlock(&mac->lock);
+    return datarate;
+}
+
+void semtech_loramac_set_rx2_params(semtech_loramac_channel_params_t params)
```
dylad: Can we declare it as static ? In the current state, this function is not threadsafe if user call it from its app.

aabadie: Good idea, and it saves a few bytes (24) in ROM. Changed

dylad:
I don't know if it is related to this PR but I'm encountering issue with the following command : 
```
> loramac set tx_power 14

> loramac get tx_power

TX power index: 0
> loramac set tx_power 10

> loramac get tx_power

TX power index: 0
```

aabadie:
> I'm encountering issue with the following command

It is definitely a bug, and not only on this branch, I can reproduce it on master. `ChannelsDefaultTxPower` should be used to set the TX power.

aabadie:
@dylad, see #8835

jia200x:
@dylad unfortunately I don't have hardware to test. We bought a gateway by January and haven't received any news about it... codewise looks OK to me though

jia200x:
@aabadie just tried it with a nucleo-l152 and SX1276, but I get:

```
> loramac tx asd
2018-03-26 13:36:25,201 - INFO #  loramac tx asd
2018-03-26 13:36:25,203 - INFO # Failed: mac is busy
```

Any clues?

jia200x:
I tried with the master branch and freezes. I'm checking the driver

aabadie:
> I tried with the master branch and freezes

Yes, there's an issue with nucleo-l152 but I didn't find the time to dig into. Can you try with another nucleo (f103 or something) ?

dylad:
I also retry, transmissions work fine with SAML21 / SX1276 and a TTN gateway in both directions.

jia200x:
@aabadie it also happens in #8639

aabadie:
> it also happens in #8639

This is normal since it's based on this one ;)

jia200x:
@aabadie the Mac busy msg vanished, but it hangs with ABP:
```
> loramac join abp
Join procedure succeeded!
> loramac tx asd
help
help
help
help
help
help
help
help
```

aabadie:
> I will investigate

I think I know what is the problem: there's no fallback when the mcps_confirm event is not "status ok" and no message is sent from the the MAC thread to the caller thread. I worked on that problem recently and have a branch that fixes this partially.

dylad:
But I did not encounter this issue on my hardware, isn't it weird ?

aabadie:
> But I did not encounter this issue on my hardware, isn't it weird ?

More or less. That can happen if the server reply is not in the receiving time frame of the end device. In general it works.
It would be great if @jia200x could provide more information on his setup (iotlab node used, DR, etc) but also try the other activation method (OTAA).

dylad:
Any news on this ?

jia200x:
I will try to find some time between today and tomorrow

aabadie:
@jia200x, I tested on iotlab and the problem you raised with ABP activation is also present in master. When using ABP activation, the TX confirmable mode doesn't work because no confirmation message is received by the end device. I know how to fix it (already worked on that in this [branch](https://github.com/aabadie/RIOT/tree/pr/pkg/semtech_loramac_mcps_fail))

aabadie:
There's problem with receiving data when using ABP activation (this is also the case in master).

dylad:
So now the question is : Should we fix it here ? or in another PR ?

aabadie:
> Should we fix it here ? or in another PR ?

I prefer in another PR: see #8982 (and mainly 99af4f464dac0b741874b26b7d6958ad331df580)

dylad:
>  prefer in another PR: see #8982 (and mainly 99af4f4)

Great, I'll give it a try this afternoon if possible.
@jia200x should we merge this one ?

jia200x:
@dylad I'm OK with merging this one since the bug is not related to this PR.

PR #7406, opened 2017-07-25
net: added CoAP RD client for simple node registration

haukepetersen:
~~based on #7402~~
also needs
- ~~https://github.com/kaspar030/sock/pull/16~~
- ~~https://github.com/kaspar030/sock/pull/17~~

This PR provides a CoAP resource directory (RD) client implementation, that uses the simplified client registration as described in section 5.3.2 in [draft-ietf-core-resource-directory-11](https://tools.ietf.org/pdf/draft-ietf-core-resource-directory-11.pdf). The client implementation can be used in two different ways:

1. standalone (`USEMODULE += rdcli_simple_standalone`): in this mode the module is running in its own thread, which is started automatically on system startup via `auto-init`. In this mode, nothing has to be done by the user.
2. 'manual' (`USEMODULE += rdcli_simple`): in this mode, the user has to take care of calling `rdcli_simple_register()` for initial registration and then periodically or on certain events to refresh the registration and to reset the lifetime counter in the resource directory.

The `rdcli_simple` client is statically configured through the `sys/net/include/rdcli_config.h` header.

PR #8800, opened 2018-03-19
net/gcoap: allow to not define a response handler

haukepetersen:
### Contribution description
Sometimes it makes sense to send out CoAP packages without needing/wanting to act on any responses. So this PRs allows to parse a `NULL` pointer as response handler.

### Issues/PRs references
Cut out and needed for #7428 and #7406


kaspar030:
Looks fine to me. @kb2ma all yours. ;)

PR #8801, opened 2018-03-19
murdock: initial hardware-in-the-loop support

kaspar030:
### Contribution description

This PR adds:

- the make target "test-murdock", which will run "make test" for the given board on murdock. (needs authentication to be set up)

- a per-application variable "TEST_WHITELIST", which is supposed to contain a list of boards on which CI should execute "make test".

- some code in ```.murdock``` that executes "make test" after successful compilation, on nightly builds or if the label "CI: run tests" is set.

Don't merge (yet), this needs a worker for every board in "TEST_WHITELIST", otherwise CI builds stall forever.

### DIFF
--- a/.murdock
+++ b/.murdock
@@ -1,13 +1,40 @@
 #!/bin/sh
 
+export TEST_BOARDS_AVAILABLE=${TEST_BOARDS_AVAILABLE:-"samr21-xpro"}
 export RIOT_CI_BUILD=1
 export STATIC_TESTS=${STATIC_TESTS:-1}
 export CFLAGS_DBG=""
 export DLCACHE_DIR=${DLCACHE_DIR:-~/.dlcache}
 
+NIGHTLY=${NIGHTLY:-0}
+RUN_TESTS=${RUN_TESTS:-${NIGHTLY}}
+
+check_label() {
+    local label="${1}"
+    [ -z "${CI_PULL_LABELS}" ] && return 1
+    echo "${CI_PULL_LABELS}" | grep -q "${label}"
+    return $?
+}
+
+[ "$RUN_TESTS" != "1" ] && {
+    check_label "CI: run tests" && RUN_TESTS=1
+}
+
 error() {
     echo "$@"
     exit 1
+}
+
+# true if word "$1" is in list of words "$2", false otherwise
+# uses grep -w, thus only alphanum and "_" count as word bounderies
+# (word "def" matches "abc-def")
+is_in_list() {
+    [ $# -ne 2 ] && return 1
+
+    local needle="$1"
+    local haystack="$2"
+
+    echo "$haystack" | grep -q -w "$needle"
 }
 
 _greplist() {
@@ -91,14 +118,23 @@ compile() {
     [ ! -d "$appdir" ] && error "$0: compile: error: application directory \"$appdir\" doesn't exist"
     [ ! -d "boards/$board" ] && error "$0: compile: error: board directory \"boards/$board\" doesn't exist"
 
+    # compile
     CCACHE_BASEDIR="$(pwd)" BOARD=$board RIOT_CI_BUILD=1 \
         make -C${appdir} clean all -j${JOBS:-4}
     RES=$?
 
+    # run tests
     if [ $RES -eq 0 ]; then
-        if [ "$board" = "native" -a "$appdir" = "tests/unittests" ]; then
-            make -C${appdir} test
-            RES=$?
+        if [ $RUN_TESTS -eq 1 -o "$board" = "native" ]; then
+            if [ -f "${BINDIR}/.test" ]; then
+                if [ "$board" = "native" ]; then
+                    BOARD=$board make -C${appdir} test
+                    RES=$?
+                elif is_in_list "$board" "$TEST_BOARDS_AVAILABLE"; then
+                    BOARD=$board make -C${appdir} test-murdock
+                    RES=$?
+                fi
+            fi
         fi
     fi
 
@@ -111,6 +147,32 @@ compile() {
     fi
 
     return $RES
+}
+
+test_job() {
+    local appdir=$1
+    local board=$2
+    local flashfile="$3"
+
+    [ ! -f "$flashfile" ] && {
+        echo "$0: _test(): flashfile \"$flashfile\" doesn't exist!"
+        return 1
+    }
+
+    dwqc \
+        ${DWQ_JOBID:+--subjob} \
+        --file $flashfile:$appdir/bin/${board}/$(basename $flashfile) \
+        --queue ${TEST_QUEUE:-$board} \
+        --maxfail 1 \
+        "./.murdock run_test $appdir $board"
+}
+
+run_test() {
+    local appdir=$1
+    local board=$2
+    print_worker
+    echo "-- executing tests for $appdir on $board:"
+    BOARD=$board make -C$appdir flash-only test
 }
 
 # execute static tests
--- a/makefiles/murdock.inc.mk
+++ b/makefiles/murdock.inc.mk
@@ -1,0 +1,6 @@
+test-murdock: link
+	cd $(RIOTBASE) && \
+		./.murdock make_test_job \
+		$$(realpath --relative-to $(RIOTBASE) $(APPDIR)) \
+		$(BOARD) \
+		$(HEXFILE)
--- a/makefiles/murdock.inc.mk
+++ b/makefiles/murdock.inc.mk
@@ -1,6 +1,36 @@
-test-murdock: link
+#
+# This file contains helper targets used by the CI.
+#
+
+# (HACK) get actual flash binary from FFLAGS.
+FLASHFILE:=$(filter $(HEXFILE) $(ELFFILE:.elf=.bin) $(ELFFILE),$(FFLAGS))
+
+#
+# This target will run "make test" on the CI cluster.
+#
+# In order to work, these requirements must be fulfilled:
+# - DWQ_REPO and DWQ_COMMIT are set correctly
+# - the user has set up autossh & proper authentication for connecting to the CI
+# (intended to be used by CI only for now)
+test-murdock:
 	cd $(RIOTBASE) && \
-		./.murdock make_test_job \
+		./.murdock test_job \
 		$$(realpath --relative-to $(RIOTBASE) $(APPDIR)) \
 		$(BOARD) \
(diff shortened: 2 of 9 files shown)


tests/unittests/Makefile
```diff
@@ -226,12 +226,18 @@ BASELIBS += $(UNIT_TESTS:%=$(BINDIR)/%.a)
...
+
 include $(RIOTBASE)/Makefile.include
 
 .PHONY: $(UNIT_TESTS)
 
 all:
 
+info-unittests:
+	@echo $(UNIT_TESTS)
+
```
kYc0o: It won't be useful to give info for any test? Why only for unittests (at least in the examples you're giving).

kaspar030: Only unittests are split into tests-*. This info target outputs that information. I want to use that to split up the unittests into seperate binaries for platforms where all tests don't fit.

kaspar030:
This is ready for review.

kaspar030:
I'll proactively squash now that I'm happy ;)

kYc0o:
I added more tests. Let's see what murdock says.

makefiles/murdock.inc.mk
```diff
@@ -0,0 +1,35 @@
...
+test-murdock:
+	cd $(RIOTBASE) && \
+		./.murdock test_job \
+		$$(realpath --relative-to $(RIOTBASE) $(APPDIR)) \
+		$(BOARD) \
+		$(FLASHFILE)
+
+# don't whitelist tests if there's no binary
+ifeq (1,$(RIOTNOLINK))
+  TEST_WHITELIST:=
```
aabadie: Maybe call this variable TEST_BOARD_WHITELIST to be more explicit ?

.murdock
```diff
@@ -91,14 +118,25 @@ compile() {
...
 
+    # run tests
     if [ $RES -eq 0 ]; then
-        if [ "$board" = "native" -a "$appdir" = "tests/unittests" ]; then
-            make -C${appdir} test
-            RES=$?
+        if [ $RUN_TESTS -eq 1 -o "$board" = "native" ]; then
+            if [ -f "${BINDIR}/.test" ]; then
+                if [ "$board" = "native" ]; then
+                    BOARD=$board make -C${appdir} test
```
cladmi: Why not run the native tests also with `test-murdock` this would allow separating their output in murdock result ?

makefiles/murdock.inc.mk
```diff
@@ -0,0 +1,35 @@
+#
+# This file contains helper targets used by the CI.
+#
+
+# (HACK) get actual flash binary from FFLAGS.
+FLASHFILE:=$(filter $(HEXFILE) $(ELFFILE:.elf=.bin) $(ELFFILE),$(FFLAGS))
```
cladmi: It will not work for boards using `jlink.sh` as flasher as it does not get the filename from command line but from an environment variable… so there is nothing in FFLAGS.

Also having a `FLASHFILE` variable is something I am interested to fix in general.
Move the FLASHFILE outside of FFLAGS and be called as `$(FLASHER) $(FFLAGS) $(FLASHFILE)`.
We currently mix `FLASHFILE` and `HEXFILE/ELFFILE/BINFILE` names right now, and some HEXFILE are  `.bin`.

Also, for `nucleo` boards we generate only an `hex` file when a `.bin` file is needed to flash over the mounted volume. And we will need to be able to generate both outputs to handle different flashing.

makefiles/murdock.inc.mk
kaspar030: how about TEST_ON_CI_WHITELIST?

aabadie: CI_BOARD_WHITELIST ?

.murdock
```diff
@@ -91,14 +118,25 @@ compile() {
     [ ! -d "$appdir" ] && error "$0: compile: error: application directory \"$appdir\" doesn't exist"
     [ ! -d "boards/$board" ] && error "$0: compile: error: board directory \"boards/$board\" doesn't exist"
 
+    # compile
     CCACHE_BASEDIR="$(pwd)" BOARD=$board RIOT_CI_BUILD=1 \
         make -C${appdir} clean all -j${JOBS:-4}
     RES=$?
 
+    # run tests
```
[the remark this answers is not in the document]
kaspar030: Tests are only actually run on native (see below).

On other boards, the compile function runs "make test-murdock", which in turn creates a "run_test" job. thus the test jobs only created by compile().
Without massive refactoring, it has to be done like this.

makefiles/murdock.inc.mk
```diff
@@ -0,0 +1,35 @@
...
+		$(BOARD) \
+		$(FLASHFILE)
+
+# don't whitelist tests if there's no binary
+ifeq (1,$(RIOTNOLINK))
+  TEST_WHITELIST:=
+endif
+
+# create $(BINDIR)/.test file only if BOARD is in $(TEST_WHITELIST)
+link: $(BINDIR)/.test
```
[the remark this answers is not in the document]
kaspar030: > Makefile remark:

will fix

.murdock
```diff
@@ -91,14 +118,25 @@ compile() {
...
-        if [ "$board" = "native" -a "$appdir" = "tests/unittests" ]; then
-            make -C${appdir} test
-            RES=$?
+        if [ $RUN_TESTS -eq 1 -o "$board" = "native" ]; then
+            if [ -f "${BINDIR}/.test" ]; then
+                if [ "$board" = "native" ]; then
+                    BOARD=$board make -C${appdir} test
+                    RES=$?
+                else
+                    if is_in_list "$board" "$TEST_BOARDS_AVAILABLE"; then
```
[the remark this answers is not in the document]
kaspar030: yup

.murdock
kaspar030: This is an optimization. Justs running the test saves the CI to create a job (with the attached .elf), copy job to the murdock master (over potentially slow network), then dispatch to a worker. That's an unnecessary network round trip.

The end result is that failed native tests end up as failed compile job, which is IMO cosmetic and (for now) worth the benefits of saving the network bandwidth.

makefiles/murdock.inc.mk
kaspar030: Yeah. I started fixing that for this PR, then decided that I'd prefer to support boards working with this FLASHFILE hack first and solve the problem afterwards. Ideally, once solved elsewhere, only the "FLASHFILE" variable has to be removed,and until then, we can run tests on all boards where the hack works.

makefiles/murdock.inc.mk
kaspar030: are boards in CI_BOARDS_WHITELIST compile-tested or not?

aabadie: > are boards in CI_BOARDS_WHITELIST compile-tested or not?

Good point. Since you need to build before running, there's no need for it indeed. The HIL test will also check the build.

kaspar030:
> I added more tests. Let's see what murdock says.

@kYc0o please open another PR for extending the enabled tests, so we split infrastructure and its application. Otherwise, this PR is unnecessarily blocked until all tests complete successfully.

makefiles/murdock.inc.mk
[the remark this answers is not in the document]
kaspar030: > The target could be set .PHONY to be re-run everytime.

I chose that option so if the Makefile changes, this removes a potentially existing .test.

kYc0o:
> @kYc0o please open another PR for extending the enabled tests, so we split infrastructure and its application. Otherwise, this PR is unnecessarily blocked until all tests complete successfully.

OK, I pushed here since it was already adding some tests, and I wasn't sure it would work also for a separated PR. I'll open another one then.

makefiles/murdock.inc.mk
kaspar030: So TEST_ON_CI_WHITELIST is better than TEST_WHITELIST? @aabadie @cladmi?

.murdock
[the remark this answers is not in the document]
kaspar030: -> wontfix

makefiles/murdock.inc.mk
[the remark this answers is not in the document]
kaspar030: > The target could be set .PHONY to be re-run everytime.

I chose that option so if the Makefile changes, this removes a potentially existing .test after Makefile changes.

.murdock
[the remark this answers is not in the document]
kaspar030: done

makefiles/murdock.inc.mk
cladmi: Yep, should be fixed in its proper PR.

makefiles/murdock.inc.mk
cladmi: I even missed it was a variable only for CI and not for tests in general. So yes having it explicit is good.

.murdock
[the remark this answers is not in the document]
cladmi: No problem to run everything at once.
I was just talking about renaming the function, which could then in bash, call two sub functions to split the function in two in the script. But its cosmetic and only used by murdock. So do as you wish with it.

kaspar030:
- changed "TEST_WHITELIST" to "TEST_ON_CI_WHITELIST".

IMO whe should get this in. @cladmi care to ACK?

.murdock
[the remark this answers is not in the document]
kaspar030: the murdock scripts use the actual function name to sort the results. Thus if "compile" gets changed into e.g., "dispatch_compile_and_test", murdock would call that job "dispatch_compile_and_test/<application>/<board>" and also put the job's output into that folder + /output.txt. Thus I'm reluctant to change too much for now.

kaspar030:
@smlng @cgundogan maybe you could take a look?

kaspar030:
@cladmi *ping* ACK now and I'll bring a random berlin bottled craft beer to the next pizza day!

cladmi:
Its good for me, I let you merge when you want if you want more feedback.

kaspar030:
> Its good for me, I let you merge when you want if you want more feedback.

Let's see how the nightly looks like.

kaspar030:
Ahja, and let's not tell everyone that on-hardware test can be triggered using the "CI: run tests" label, and test this a bit with selected PR's first. :smirk:

kaspar030:
> (which beer ? :p)

Surprise!

PR #8802, opened 2018-03-20
boards: slstk3402a: add support

basilfx:
### Contribution description

This PR add supports for the Silicon Labs SLSTK3402a. The files are similar to the SLTB001a and STK3600, STK3700 and SLSTK3401a already supported. The main difference with the STK3x00 is that is has a second series MCU and it has a memory LCD (supported by U8g2). This MCU also provides two crypto accelerators, and a hardware number generator.

I can provide remote access to the hardware, if needed.

### Issues/PRs references

#8585
#8520
#8630


basilfx:
Updated vendor headers per #8873.

basilfx:
Rebased.

boards/slstk3402a/include/board.h
```diff
@@ -0,0 +1,119 @@
...
+#endif
+
+/**
+ * @name    Xtimer configuration
+ *
+ * The timer runs at 250 KHz to increase accuracy.
+ * @{
+ */
+#define XTIMER_HZ           (250000UL)
+#define XTIMER_WIDTH        (16)
```
kYc0o: I don't know if all boards are configured like this, but it looks to me a bit weird. Isn't possible to run the timer at 1MHz? 250KHz is a scaled frequency so you might run in the troubles as in #9049. Can you verify?

basilfx: Other boards are the same. I changed to 250kHz because I did not have enough accuracy to play with. It uses two 16-bit timers in cascading mode, although this chip (generation) has support for a 32-bit timer that I will add support for.

kYc0o:
Is this conflicting with the I2C rework? Because I'd like to merge it but obviously the I2C would be a blocker.

basilfx:
There is already a PR to adapt the boards: #9208. Adding this one is not so hard. So when merged, I can adapt that one.

basilfx:
Alternatively, I could remove I2C support for the time being? What do you think, @aabadie/@kYc0o?

kYc0o:
I'd say yes, I trust you won't forget 😉

basilfx:
Removed I2C definitions for now. See 866f151.

kYc0o:
Did you remove the feature? I guess Murdock complains about it.

basilfx:
@kYc0o I forgot to remove that features, yes. Now the build is OK. Can I squash?

kYc0o:
Yes please squash.

basilfx:
Squashed and rebased.

PR #7277, opened 2017-06-29
cpu/stm32: unified ADC driver implementations

haukepetersen:
EDIT
~~This PR is still work in progress, just wanted to put it already out there to prevent multiple people working on this...~~

I have not test all of the STM CPUs, yet. ~~Especially I doubt the L1 and L4's to work, as it seems they have subtle differences in their ADC registers...~~

L1 works, L4 was not implemented before anyway...

PR #8804, opened 2018-03-20
cpu/stm32l4: add adc implementation

MichelRottleuthner:
### Contribution description
Adds implementation for ADC on stm32l4. Currently this only includes configuration for nucleo-l476, and only for the analog pins of the arduino header.

### Issues/PRs references
Could make sense to merge this with #7277.
I didn't base it on that PR as it doesn't seem working completely for now, and the ADCs on the L4 are actually quite a bit different to initialize.

### DIFF
--- a/cpu/stm32l4/periph/adc.c
+++ b/cpu/stm32l4/periph/adc.c
@@ -1,0 +1,214 @@
+/*
+ * Copyright (C) 2014-2016 Freie Universität Berlin
+ * Copyright (C) 2018 HAW-Hamburg
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser General
+ * Public License v2.1. See the file LICENSE in the top level directory for more
+ * details.
+ */
+
+/**
+ * @ingroup     cpu_stm32l4
+ * @ingroup     drivers_periph_adc
+ * @{
+ *
+ * @file
+ * @brief       Low-level ADC driver implementation
+ *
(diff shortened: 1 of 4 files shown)


cpu/stm32l4/periph/adc.c
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * @brief   Load the ADC configuration
+ */
+static const adc_conf_t adc_config[] = ADC_CONFIG;
+
+/**
+ * @brief   Allocate locks for all three available ADC devices
+ */
+static mutex_t locks[] = {
```
kaspar030: ```static mutex_t locks[ADC_DEVS];``` will have the same effect.

kaspar030:
Does anyone have an stm32l4 board for testing?

aabadie:
> Does anyone have an stm32l4 board for testing?

me, but I won't be able to test before next tuesday unfortunately

cpu/stm32l4/periph/adc.c
aabadie: I allowed myself to directly address your comment @kaspar030. Tested with nucleo-l476 and still works.

tcschmidt:
Test seems o.k.. Are the change requests satisfied @kaspar030 ? Can we proceed with this?

kaspar030:
needs rebase

MichelRottleuthner:
Murdock spits out some unrelated error regarding xtimer_now64_continuity test. @kaspar030: do you have an idea where this comes from?

PR #8807, opened 2018-03-20
nucleo-f4xx: increase xtimer backoff to 7 for F4 family

kYc0o:
### Contribution description
`XTIMER_BACKOFF` is defined with a very low value which fails at 7us for xtimer_sleep on the F4 nucleo family.

This PR increases it to 7 which makes all tests pass.

Tested on:
- [ ] f4vi1
- [ ] msbiot
- [x] nucleo-f401
- [x] nucleo-f446
- [x] nucleo-f303ze
- [x] nucleo-f410
- [ ] nucleo-f411
- [x] nucleo144-f412
- [ ] nucleo144-f413
- [x] nucleo144-f429
- [x] nucleo144-f446
- [x] stm32f4discovery

### Issues/PRs references
#7347

### DIFF
--- a/boards/common/nucleo/include/board_nucleo.h
+++ b/boards/common/nucleo/include/board_nucleo.h
@@ -53,7 +53,7 @@ extern "C" {
 #endif
 
 #if defined(CPU_FAM_STM32F4) || defined(CPU_MODEL_STM32F303ZE)
-#define XTIMER_BACKOFF              (5)
+#define XTIMER_BACKOFF              (8)
 #define XTIMER_OVERHEAD             (6)
 #endif
 /** @} */


kYc0o:
Please add on the first message of this PR the already tested boards to see if it doesn't break anything. Still, I'm not convinced this is a good approach to solve this problem, but it's what we have now.

Also, we could add to this PR other cases if needed. So please report other boards experiencing this behaviour by running tests/xtimer_usleep_short and if it fails, post at which time it failed.

aabadie:
Added nucleo-f303ze (e.g. nucleo144-f303) because it's also affected by the change and thus it should be checked as well.

bergzand:
Added nucleo-f401 to test, also affected and I'm curious whether it requires the same value of 7 as well. Board is on my desk so @kYc0o feel free to grab it tomorrow if I forgot about this. :)

kYc0o:
Tested on F401 and it needs a backoff of 8, changed in my last commit.

bergzand:
Grepping through the currently supported boards we have the following boards using a stm32f4 CPU:

- f4vi1
- msbiot
- nucleo-f401
- nucleo-f410
- nucleo-f411
- nucleo-f446
- nucleo144-f412
- nucleo144-f413
- nucleo144-f429
- nucleo144-f446
- stm32f4discovery

Preferably I would like to have them all checked for this problem. But I also don't see a huge problem with merging this and leaving the list for an issue. Only thing I'm afraid of then is that the issue will silently die

kYc0o:
@bergzand I added the boards you found to the main description.

aabadie:
@kYc0o, can you squash this one ? I think it can be merged.

kYc0o:
Just tested on stm32f4discovery, it works, I'll squash and this should be ready.

aabadie:
All green, so let's go

PR #8808, opened 2018-03-20
cpu/atmega_common: use __temp_reg__

Josar:
Using __temp_reg__ in assebler code to ensure proper working in future.

### DIFF
--- a/cpu/atmega_common/include/cpu.h
+++ b/cpu/atmega_common/include/cpu.h
@@ -74,25 +74,20 @@ void cpu_init(void);
 
 /**
  * @brief   Print the last instruction's address
- *
- * @todo:   Not supported
  */
-static inline void cpu_print_last_instruction(void)
+__attribute__((always_inline)) static inline void cpu_print_last_instruction(void)
 {
     uint8_t hi;
     uint8_t lo;
     uint16_t ptr;
 
-    __asm__ volatile( "in r0, __SP_H__; \n\t"
-                      "mov %0, r0       \n\t"
-                             : "=g"(hi)
-                             :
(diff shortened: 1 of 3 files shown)


basilfx:
Hi @Josar, thanks for your contribution. Can you adopt your commit message according to [this](https://github.com/RIOT-OS/RIOT/wiki/Development-procedures)?

Something like `cpu/atmega_common: use stackpointer with temp_reg`?

kaspar030:
Doesn't compile.

Josar:
I tried locally they all compile. Did you try to compile them locally?
So Question why don't they compile with Murdock?

My Setup:
gcc version 4.9.2 (AVR_8_bit_GNU_Toolchain_3.5.4_1709)
avr-libc 2.0.0

Murdock:
avr-gcc: avr-gcc (GCC) 4.9.2

Has anyone an idea why it does not compile?

Josar:
"It's a good idea to use tmp_reg and zero_reg instead of r0 or r1, just in case a new compiler version changes the register usage definitions."

https://www.nongnu.org/avr-libc/user-manual/inline_asm.html#asm_code

kaspar030:
> It's a good idea to use tmp_reg and zero_reg instead of r0 or r1, just in case a new compiler version changes the register usage definitions."

thanks!

So what was the issue, why did it not compile before?

Josar:
Is this PR good to go?

Josar:
@kaspar030 @basilfx some more things to do?

basilfx:
I don't have experience with ATmega, so I leave this one up to @kaspar030.

PR #8812, opened 2018-03-21
boards: stm32f4discovery: include: periph_conf.h: Change I2C_0_SDA_PIN

graznik:
The STM32F4DISCOVERY on-board I2C peripheral CS43L22 is connected
to PB6 (SCL) and PB9 (SDA). Both pins have external 4.7K pull-up
resistors and are ready for I2C communication by default. The
default RIOT configuration uses PB7 as SDA pin which makes first
I2C bus bring up unnecessarily complicated.

* Change #define for I2C_0_SDA_PIN from 7 to 9

### DIFF
--- a/boards/stm32f4discovery/include/periph_conf.h
+++ b/boards/stm32f4discovery/include/periph_conf.h
@@ -259,7 +259,7 @@ static const spi_conf_t spi_config[] = {
 #define I2C_0_SCL_AF        4
 #define I2C_0_SCL_CLKEN()   (periph_clk_en(AHB1, RCC_AHB1ENR_GPIOBEN))
 #define I2C_0_SDA_PORT      GPIOB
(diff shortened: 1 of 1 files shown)


kaspar030:
Change seems legit to me. Is there a chance that we break user applications with this change?

MrKevinWeiss:
Hmm, this will be blocked since the I2C API is undergoing a re-haul.  If you would like to propagate this to the new_i2c_if branch I keep an eye on it so it can get merged quickly.

@graznik it makes sense though!

MrKevinWeiss:
@graznik are you still interested in changing the default i2c sda pin?  If not I can take it over since it so it is up to date (it really doesn't make sense to only have one line connected to a bus).

smlng:
closing, because outdated due to new API and superseded by #10445

PR #8820, opened 2018-03-22
stm32/common : add STM32_PM_SLEEP define

Marc-Aurele:
The STM32_PM_SLEEP has been added to be used with
pm_unblock function.

Signed-off-by: Aurelien Fillau <aurelien.fillau@gmail.com>

### DIFF
--- a/cpu/stm32_common/include/periph_cpu_common.h
+++ b/cpu/stm32_common/include/periph_cpu_common.h
@@ -85,6 +85,7 @@ extern "C" {
  * @name    Power modes
  * @{
  */
+#define STM32_PM_SLEEP        (2U)
 #define STM32_PM_STOP         (1U)
 #define STM32_PM_STANDBY      (0U)
 /** @} */


kaspar030:
Sleep sets the cortex-m CPU idle mode, right? If yes, what's the use-case for blocking that?

kaspar030:
*ping*

Marc-Aurele:
Yes, sleep sets the cpu in idle mode. At sart-up all mode are blocked, i need to unblock at least the sleep mode by using pm_unblock(STM32_PM_SLEEP) so that the CPU can go in sleep, that's why i have created this macro but maybe i have missed something ?

kaspar030:
> At sart-up all mode are blocked, i need to unblock at least the sleep mode by using pm_unblock(STM32_PM_SLEEP) so that the CPU can go in sleep

That shouldn't be the case. STM32 has PM_NUM_MODES defined as 2. Those are modes "0" and "1". pm_layered will set mode "2" when all modes are blocked.

Can you confirm that your board doesn't go to idle (mode 2) with master?

To do so, set ```ENABLE_DEBUG=1``` in pm_layered.c, add ```CFLAGS += -DTHREAD_STACKSIZE_IDLE=THREAD_STACKSIZE_MAIN``` in to the hello-world Makefile and check it's output.

Marc-Aurele:
ok, i'll check that as soon as possible

Issue #8569, opened 2018-02-15
EFR32: track radio driver progress


basilfx:
This issue is to track the progress of the integrated radio drivers of the EFR32 family (supported via EFM32).

Currently, the following boards have support:

* [SLTB001a](https://github.com/RIOT-OS/RIOT/tree/master/boards/ikea-tradfri)
* [IKEA-tradfri](https://github.com/RIOT-OS/RIOT/tree/master/boards/sltb001a)

Current work:

* Initial driver based on RAIL 1.x by @kaibeckmann: [here](https://github.com/kaibeckmann/RIOT/tree/efr32mg12_radio_dev_currentRiot2)
* Non-working initial upgrade to RAIL 2.x by @basilfx: [here](https://github.com/basilfx/RIOT/tree/temp/kai_rail_lib_2.0),

TODO:

- [x] Port RAIL 1.x driver to 2.x (@kaibeckmann #9212)
  - [x] Gecko SDK package updates
- [ ] GNRC (@kaibeckmann #9212)
- [ ] ...

Caveats/comments/notes:

* RAIL 2.x supports multi-protocol (e.g. 802.15.4 and BLE). How to handle?
* RAIL 1.x/2.x is a blob with fixed ISR names (can be remapped, e.g. `arm-none-eabi-objcopy librail_efr32xg1_gcc_release.a --redefine-sym FRC_PRI_IRQHandler=isr_frc_pri`)

basilfx:
So far I know @kYc0o, @pyropeter, @kaibeckmann and I are interested.

My knowledge on radio drivers is very limited, but I can help to fix things on the EF{M, R}32 side though :-)

PR #8824, opened 2018-03-22
boards: slwstk6000b: add support

basilfx:
### Contribution description

This PR add supports for the Silicon Labs SLWSTK6000B development kit for wireless applications. Although the files are very similar to the other PRs I have submitted, this board is actually a base board that can host different CPU modules. All CPU modules have fixed pins for certain functions, including ones for the onboard LCD and Si7021 temperature sensor.

I came up with an idea similar to how I supported the EFM32 family, by using `modules.txt` as a 'database' for mapping the base board to a CPU module. I don't think duplicating this board for all the different modules would work out.

To compile, you actually have to override the `BOARD_MODULE` (defaults to `SLWBR4162a`) using something like this:

`BOARD=slwstk6000b BOARD_MODULE=slwrb4162a make -j8`

I currently have two CPU modules supported: `SLWBR4150A` and `SLWBR4162A`. However, there are [15+ more modules](https://www.silabs.com/products/development-tools/wireless/mesh-networking/mighty-gecko-starter-kit) that fit this base board.

### Issues/PRs references

#8585
#8520
#8630
#8569


basilfx:
Asking @kaibeckmann to test this PR, since he has access to the boards.

kaibeckmann:
Hey I found a couple of things.
For the two modules I tested, the cpu was wrong. Therefore the right header files were missing:
- efr32mg1p233f256gm48.h
- efr32mg12p332f1024gl125.h

I like the submodule system, but I think the new definition for the pins is not necessary. First the names are derived from the slwrb4162a, but the datesheets of the other modules have different names for the pins with the same GPIO. That is quite confusing. And than it's not necessary, because the GPIOs seem to be the same (at least for the 3 or 4 datasheets I checked). 

The main difference of the modules are the cpu and additional periphery, like SPI-Flash, Touch-Slider, radio transceiver and if there are separate GPIOs to enable sensors, dispay etc.

basilfx:
Thanks @kaibeckmann for testing.

The mapping of modules to CPU is based on the list on the page of the [SLWSTK6000B](https://www.silabs.com/products/development-tools/wireless/mesh-networking/mighty-gecko-starter-kit) itself.

However, looking at the reference manuals, I do see different MCU's being used. I'll ask Silicon Labs to update their list, somehow.

Regarding the pin numbers: it is needed. If you look at the reference manuals, you can see that some MCU pins map to different module pins. For instance, the `brd4150a` maps P0 to PA2, but it maps P0 to PD8 on the `brd4162a`. Then there are also a lot of pins that aren't connected on some boards.

The pin numbers are confusing at first, because they don't link to the reference manuals, where you can see the actual module pins. Here are some:

* `[brd4150a](https://www.silabs.com/documents/public/reference-manuals/brd4150a-rm.pdf)`
* `[brd4154a](https://www.silabs.com/documents/public/reference-manuals/brd4154a-rm.pdf)`
* `[brd4162a](https://www.silabs.com/documents/public/reference-manuals/brd4162a-rm.pdf)`
* `[brd4167a](https://www.silabs.com/documents/public/reference-manuals/brd4167a-rm.pdf)`

kaibeckmann:
Hmm I see your point, but the module pin names are confusing. It makes the understanding and using of the boards for someone new more complex.

Maybe a more elaborate name schema would help? Something like MODULE_PIN_P0 ?

If the MCU pin for the basic peripherals really can change, than the multiplexing part should properly be moved to the module header as well (UART, I2C, SPI, ...).

basilfx:
I can change it to `MODULE_PIN_P0`. That is a better name, and I will also put in the docs that it concerns the module pinout.

I haven't encountered a reason for moving the multiplexing part as well, but if it is necessary, I'll do it. I want to keep as much as possible in `board.h` and `periph_conf.h`.

basilfx:
I have fixed the board name to module mapping, based on the reference manuals and renamed the pin names.

boards/slwstk6000b/Makefile.include
```diff
@@ -0,0 +1,27 @@
...
+# set default port depending on operating system
+PORT_LINUX ?= /dev/ttyACM0
+PORT_DARWIN ?= $(firstword $(sort $(wildcard /dev/tty.usbmodem*)))
+
+# setup serial terminal
+include $(RIOTMAKE)/tools/serial.inc.mk
+
+# setup JLink for flashing
+export JLINK_DEVICE := $(CPU_MODEL)
+include $(RIOTMAKE)/tools/jlink.inc.mk
```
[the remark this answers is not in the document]
basilfx: I think this is a bug. I compared JLink with OpenOCD, and OpenOCD sets DBG=GDB, but JLink doesn't. Will look into this tonight.

boards/slwstk6000b/Makefile.include
```diff
@@ -0,0 +1,27 @@
...
+
+# set default port depending on operating system
+PORT_LINUX ?= /dev/ttyACM0
+PORT_DARWIN ?= $(firstword $(sort $(wildcard /dev/tty.usbmodem*)))
+
+# setup serial terminal
+include $(RIOTMAKE)/tools/serial.inc.mk
+
+# setup JLink for flashing
+export JLINK_DEVICE := $(CPU_MODEL)
```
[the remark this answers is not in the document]
basilfx: I've added additional data in `modules.txt`.

boards/slwstk6000b/include/periph_conf.h
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * @name    UART configuration
+ * @{
+ */
+static const uart_conf_t uart_config[] = {
+    {
+        .dev = USART0,
+        .rx_pin = F6,
+        .tx_pin = F7,
```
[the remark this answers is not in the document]
basilfx: Fixed.

boards/slwstk6000b/include/periph_conf.h
```diff
@@ -0,0 +1,194 @@
...
+        .scl_pin = P12,
+        .loc = I2C_ROUTELOC0_SDALOC_LOC16 |
+               I2C_ROUTELOC0_SCLLOC_LOC14,
+        .cmu = cmuClock_I2C0,
+        .irq = I2C0_IRQn
+    }
+};
+
+#define I2C_NUMOF           PERIPH_NUMOF(i2c_config)
+#define I2C_0_ISR           isr_i2c1
```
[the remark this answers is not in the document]
basilfx: Fixed.

boards/slwstk6000b/Makefile.include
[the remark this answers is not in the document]
basilfx: #8853 should address this.

basilfx:
@kaibeckmann do the latest adaptions work for you? The debug problem should now be addressed.

kaibeckmann:
OK I tested it for the slwrb4150a and slwrb4162a with the default app and the tests-core. It works fine.

basilfx:
Updated vendor headers per #8873.

kaibeckmann:
OK I run the tests again, everything is working as expected.

basilfx:
Ping @kYc0o or @jia200x.

boards/slwstk6000b/modules/slwrb4162a/include/board_module.h
```diff
@@ -0,0 +1,111 @@
...
+#define MODULE_PIN_P37      GPIO_PIN(PB, 10)
+#define MODULE_PIN_P38      GPIO_UNDEF
+#define MODULE_PIN_P39      GPIO_UNDEF
+#define MODULE_PIN_P40      GPIO_UNDEF
+#define MODULE_PIN_P41      GPIO_PIN(PF, 8)
+#define MODULE_PIN_P42      GPIO_PIN(PF, 9)
+#define MODULE_PIN_P43      GPIO_PIN(PF, 10)
+#define MODULE_PIN_P44      GPIO_PIN(PF, 11)
+#define MODULE_PIN_P45      GPIO_PIN(PF, 12)
+/** @} */
```
kYc0o: How different are these definitions to the ones for the other module?

basilfx: So far as I can tell, different enough. There are modules that only define half of these pins.

cpu/efm32/families/efr32mg12p/vectors.c
```diff
@@ -0,0 +1,135 @@
+/*
+ * Copyright (C) 2015-2018 Freie Universität Berlin
+ *
```
kYc0o: Freie Universität?

basilfx: Yes, file was copy-pasted from EZR32WG, by @haukepetersen.

boards/slwstk6000b/modules/slwrb4162a/include/board_module.h
kYc0o: I was just wondering if we could merge them in a single file (with `#ifdef`s), since I see that at least the beginning of the definitions are the same.

kYc0o: Ok, please squash.

basilfx:
Squashed and rebased. Murdock is happy.

basilfx:
@kYc0o @kaibeckmann Good to go?

basilfx:
Thanks!

PR #8828, opened 2018-03-23
sys/spin_random: CPU delay library

gebart:
### Contribution description

Provides random CPU runtime delays.
Useful to artificially add fuzziness to hardware timings in tests.

### Issues/PRs references

Used by #8531 for avoiding phase lock between timer under test and test driver

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -569,6 +569,11 @@ ifneq (,$(filter saul,$(USEMODULE)))
   USEMODULE += phydat
 endif
 
+ifneq (,$(filter spin_random,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_timer
+  USEMODULE += random
+endif
+
 ifneq (,$(filter phydat,$(USEMODULE)))
   USEMODULE += fmt
 endif
--- a/sys/include/spin_random.h
+++ b/sys/include/spin_random.h
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2018 Eistec AB
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    sys_spin_random Spin_random - random CPU delays
+ * @ingroup     sys
+ * @brief       CPU spinning routines for random runtime delays
+ *
+ * @{
+ * @file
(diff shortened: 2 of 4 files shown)


kaspar030:
I think a little generalization would be useful. Can we maybe change the calibration funtion so it returns a "spin factor" so that spin(factor*N) takes e.g., N microseconds or timer ticks?

The spin() function can be shortened to ```static inline void spin(volatile uint32_t n) { while (n--){}}```. The volatile parameter is enough to prevent optimization, no asm needed.

IMO not using a global variable for the random delay is more useful: ```void spin_random_delay(uint32_t spin_max) { spin(random_uint32_range(0, spin_max));}```
I'm actually not sure if it makes sense to keep this one-liner around, as it adds random as dependency.

With those changes, I also suggest renaming to just "spin".

(Sorry, I think I changed your idea quite a lot. Does this make sense to you?)

gebart:
I like the static spin_max variable in this instance because it simplifies the usage a lot. See how the spin_random_delay function is used in #8531. Eliminating the random dependency and making the spin duration into an argument would require passing that spin_max variable around the test and also chaining a call to the random module in every place we need the delay, making this module much less convenient to use.

gebart:
Do you have any use for a counted spin loop without a stop condition with an accuracy of +/- 20% outside of tests?
I don't see why you would not use xtimer for this in the real world applications.

kaspar030:
> Do you have any use for a counted spin loop without a stop condition with an accuracy of +/- 20% outside of tests?

Nope. I was trying to make it a place in sys/...

gebart:
Do you think it is not right to put it under sys? 
I figured that sys was the proper place for reusability since we don't have any libraries under tests, even though in my opinion this module is only useful in testing situations, but it is general enough to be shared by other tests than just the timer benchmark.

kaspar030:
> Do you think it is not right to put it under sys?
> I figured that sys was the proper place for reusability since we don't have any libraries under tests, even though in my opinion this module is only useful in testing situations, but it is general enough to be shared by other tests than just the timer benchmark.

Honestly I think ```spin_random_delay()``` plus its calibration function should stay with the test, I don't think anyone is going to re-use it. It seems too specialized.

I wouldn't mind a generic ```spin()``` in sys/, somehow I'm using such a delay function a lot. But that might be because I do a lot of low-level timer stuff...

PR #8516, opened 2018-02-05
doc: move wiki to Doxygen documentation (second attempt)

jia200x:
### Contribution description
This is a second attempt of moving Board specifications from Wiki to Doxygen. Replaces the work in #8508 .

Wiki entries were copied "as it" and referenced to the corresponding group, [with the same format as existing Doxygen entries like Arduino MKRFOX1200](http://doc.riot-os.org/group__boards__arduino-mkrfox1200.html)
As said in the previous PR, the motivation is to help developers keeping an up-to-date End User documentation for each board without adding too much overhead, and help the End User to find all the documentation in the same place. 

- [x] deduplicate `@defgroup` and `@brief`
- [x] remove tabs
- [x] split lines to fit 80 chars

NB tables doesn't work if splitted => NR > 80. Also, this PR doesn't intend to update information for each entry.

### Some other issues addressed by this PR
* Fix mulle `todo` block

### Entries with broken imgs (addressed in further PRs)
- arduino-mega2560
- samd21-xpro
- saml21-xpro
- stm32f3discovery
- z1

### Following steps (PRs coming soon)
* Add "Family" documentation (ARM, MSP430, etc).
* Unify the current board documentation structure, so all boards have the same content (Hardware Description, Pin Mapping, Toolchain, etc).
### Issues/PRs references
#8508

PR #8829, opened 2018-03-24
cpu/atmega1284p: support and boards/mega-xplained: support

ZetaR60:
This is a port of the Atmel Mega1284P Xplained board, and the ATmega1284P MCU. You can see documentation for the board here:
https://www.microchip.com/DevelopmentTools/ProductDetails.aspx?PartNO=ATMEGA1284P-XPLD

I have tested: UART, shell, button input (via SAUL), LED output (via SAUL), ADC input (via SAUL), IRQ (via tests/irq), threading (via tests/), and xtimer (via tests/). Everything seems to be working properly.


boards/mega-xplained/include/mega-xplained_pinmap.h
```diff
@@ -0,0 +1,90 @@
...
+
+#ifndef MEGA_XPLAINED_PINMAP_H
+#define MEGA_XPLAINED_PINMAP_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Mapping of MCU pins to Mega Xplaind pins
```
aabadie: The `@brief` should be en `@name` with a `@{` on the next line

boards/mega-xplained/include/board.h
```diff
@@ -0,0 +1,89 @@
...
+ *               2016 Laurent Navet <laurent.navet@gmail.com>
+ *               2018 Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/*
+ * @defgroup    boards_mega-xplained
```
aabadie: Better moving the `@defgroup` and `@ingroup` in a separate doc.txt file. See #8516 and arduino-mkrfox1200 for examples.

ZetaR60:
Oops. I did a rebase to try and fix CI simultaneous to what you did. Let me set it back (I messed up some details due to being new to git).

aabadie:
> I did a rebase to try and fix CI simultaneous to what you did. Let me set it back

I can re-push my local branch. Is it ok fo you ?

ZetaR60:
I have set the branch back to my backup before my attempted rebase. Unfortunately your squash was lost. I will wait for you to squash this time, because I seem to have a tendency to mess them up.

aabadie:
>  I seem to have a tendency to mess them up

That's ok, I recommend that you read the good git documentation links provided by GitHub: https://help.github.com/articles/git-and-github-learning-resources/

FYI, this is what I did locally:
```
# Retrieve your PR branch
$ git remote add ZetaR60 git@github.com:ZetaR60/RIOT.git
$ git fetch ZetaR60  # Then I can checkout your PR branch
$ git checkout RIOT_xplained

# Rebase to the latest master branch of RIOT (my RIOT fork remote is called riot)
$ git pull --rebase riot master

# Rebase in interactive mode: move the commit line at the right place, replace 'pick' with 'f' on the commits to squash, save and exit
$ git rebase riot/master -i

# Need to force push because the branches (remote and local) have diverged
$ git push ZetaR60 RIOT_xplained
```

ZetaR60:
Thanks. I will use those tutorials next time.

There seems to have been an issue with moving the LED pin definitions. It looked like it worked at first, but they are not lighting correctly.

boards/mega-xplained/include/gpio_params.h
```diff
@@ -0,0 +1,86 @@
...
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Matthew Blue <matthew.blue.neuro@gmail.com>
+ */
+
+#ifndef GPIO_PARAMS_H
+#define GPIO_PARAMS_H
+
+#include "board.h"
+#include "saul/periph.h"
+#include "mega-xplained_pinmap.h"
```
aabadie: This include is not needed.

ZetaR60:
I was using SAUL in examples/default to toggle the LEDs manually. I did check all of them prior to creating the PR. Even if I revert my code back to that point, they still don't work. So, I feel like I am going crazy: I clearly remember them lighting up on my final pre-submission test.

Also, I tried tests/leds just now, and it seems that it expects some definitions for LEDs that SAUL doesn't require. But even after adding those definitions the LEDs don't light.

You can manually light the LEDs using the buttons (since the button inputs and the LEDs are tied to the same pin), so I can verify that the LEDs themselves work. Also, reading the same pin (via the button input) works correctly. When writing the pin, SAUL reports success, but does not actually change the pin state.

boards/mega-xplained/include/board.h
```diff
@@ -0,0 +1,128 @@
...
+#define XTIMER_HZ                   (125000UL)
+#define XTIMER_BACKOFF              (40)
+/** @} */
+
+/**
+ * @name    LED pin configuration
+ * @{
+ */
+#define LED0_PIN     GPIO_PIN(PORT_B, 0)
+#define LED0_MODE    GPIO_OD
```
aabadie: Maybe try with `GPIO_OUT`

boards/mega-xplained/include/board.h
```diff
@@ -0,0 +1,128 @@
...
+#define XTIMER_WIDTH                (16)
+#define XTIMER_HZ                   (125000UL)
+#define XTIMER_BACKOFF              (40)
+/** @} */
+
+/**
+ * @name    LED pin configuration
+ * @{
+ */
+#define LED0_PIN     GPIO_PIN(PORT_B, 0)
```
aabadie: You could also add macros for controlling the LEDs: see [here](https://github.com/RIOT-OS/RIOT/blob/master/boards/common/arduino-atmega/include/board_common.h#L57). You will also have to initialize the LEDs pin in the `board_init` function (in board.c).
This will be a good test for checking the LED can be controlled correctly by RIOT.

ZetaR60:
Apparently I was crazy after all: GPIO_OUT works, and it seems that cpu/atmega_common does not support the GPIO_OD mode. I must have misremembered testing the open-drain configuration.

The reason that the outputs need to be open-drain is that the button inputs are on the same pin. So if the system is asserting the line (by pulling it low), it will turn the LED on. If the line is not asserted, it will be pulled high by the LED circuit if the button is not being pushed, or will be pulled low by the button. Thus, the state of the button input can be measured so long as the system is not asserting the line. This won't work with the current way that cpu/atmega_common is using the GPIOs. I can comment out some buttons and LEDs, leaving two of each that do not conflict, and make a note that open-drain support is required to enable all of them.

I was experimenting with the macros earlier. They also failed due to the lack of open-drain. I will add the macros for the LEDs that are enabled.

boards/mega-xplained/board.c
```diff
@@ -0,0 +1,81 @@
...
+    /* initialize stdio via USART_1 */
+    SystemInit();
+
+    /* initialize the CPU */
+    cpu_init();
+
+    irq_enable();
+}
+
+/**
```
kaspar030: some stuff in this file seems to be duplicate with ```common/arduino-atmega/board.c```, maybe re-use?

ZetaR60:
I have concerns about the clarity of creating a dependency on the Arduino code. Yes, much of the code is very similar, but it is not an Arduino. Wouldn't having a non-Arduino board depend on Arduino code be confusing to someone reading the code? Or someone changing the Arduino code with the expectation that it wouldn't affect non-Arduino code. I can have it use common/arduino-atmega if you want, but I feel wary of doing so.

It might be better to have a boards/common/atmega-minimal that both the Arduino boards and this board could depend on, but that would require more substantial changes.

boards/mega-xplained/board.c
aabadie: This was also my initial impression, see [first comment](https://github.com/RIOT-OS/RIOT/pull/8829#pullrequestreview-106701875). But as @ZetaR60 is pointing out, this board is not an arduino board.
To move faster, I proposed to fix that in a follow-up PR. This PR is already 99% ready IMHO (just missing the LED macros and initialization).

boards/mega-xplained/include/gpio_params.h
ZetaR60: Removed.

boards/mega-xplained/board.c
```diff
@@ -0,0 +1,104 @@
...
+
+#include "board.h"
+#include "cpu.h"
+#include "uart_stdio.h"
+#include "periph/gpio.h"
+
+void SystemInit(void);
+static int uart_putchar(char c, FILE *stream);
+static int uart_getchar(FILE *stream);
+void led_init(void);
```
aabadie: This function should be defined `static`

boards/mega-xplained/board.c
```diff
@@ -0,0 +1,104 @@
...
+    uart_stdio_read(&c, 1);
+    return (int)c;
+}
+
+/**
+ * @brief Initialize the on-board LEDs
+ */
+void led_init(void)
+{
+/* Not currently supported due to lack of GPIO_OD support */
```
aabadie: Maybe just remove if not supported. If you think you could add the GPIO_OD support in a follow-up PR (might be done in `cpu/atmega_common/periph/gpio.c` ?), then you could re-add it here in the meantime. What do you think?

ZetaR60:
The LEDs are working properly now. Both through SAUL, and using tests/leds

As a note: tests/buttons fails, but this is due to the lack of general PCINT support in cpu/atmega_common

boards/mega-xplained/board.c
ZetaR60: Done.

boards/mega-xplained/board.c
ZetaR60: I removed the commented code, but left the message about it being unsupported.

boards/mega-xplained/board.c
```diff
@@ -0,0 +1,99 @@
...
+    uart_stdio_read(&c, 1);
+    return (int)c;
+}
+
+/**
+ * @brief Initialize the on-board LEDs
+ */
+void led_init(void)
+{
+/* LED0,2 currently unsupported due to lack of GPIO_OD support */
```
aabadie: Please add a 4 spaces indent, here

ZetaR60:
Really only now waiting for the change requests to either clear or be updated.

Figuring out the limited things I need to do with Git has been one of the most frustrating software experiences I have had in a while.

kaspar030:
> @kaspar030 is your concerns about the lack of factorization really a blocker ?

The code duplication is the issue, and yes, that is a blocker. Adding features is easy, keeping a codebase with >100 target devices maintainable is not. Not having to work on multiple versions of the same code is key to that.

There are already three copies of the atmega stdio initialization code in the tree, and @aabadie, you reviewed the two newer ones, and never provided the follow-PR.

I'm preparing a PR now, please wait...

kaspar030:
> I'm preparing a PR now, please wait...

#8837. I think it is easier if we merge this now and I update the other PR accordingly.

@ZetaR60 thank you for a nice first PR!

ZetaR60:
@aabadie @kaspar30 Thank you very much for the help, and thanks for putting up with some newbieness. ;)

PR #8835, opened 2018-03-26
pkg/semtech-loramac: fix tx_power get/set functions

aabadie:
### Contribution description

This PR fixes the issue reported in this [comment](https://github.com/RIOT-OS/RIOT/pull/8798#issuecomment-376111162)

Using `MIB_CHANNELS_TX_POWER` has no effect on the MAC. The fix simply consists in using `MIB_CHANNELS_DEFAULT_TX_POWER` instead.

### Issues/PRs references

comment raised while reviewing #8798

### DIFF
--- a/pkg/semtech-loramac/contrib/semtech_loramac_getset.c
+++ b/pkg/semtech-loramac/contrib/semtech_loramac_getset.c
@@ -190,8 +190,8 @@ void semtech_loramac_set_tx_power(uint8_t power)
 {
     DEBUG("[semtech-loramac] set TX power %d\n", power);
     MibRequestConfirm_t mibReq;
-    mibReq.Type = MIB_CHANNELS_TX_POWER;
-    mibReq.Param.ChannelsTxPower = power;
+    mibReq.Type = MIB_CHANNELS_DEFAULT_TX_POWER;
+    mibReq.Param.ChannelsDefaultTxPower = power;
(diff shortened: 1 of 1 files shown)


aabadie:
> This PR doesn't solved the issue or maybe I'm doing something wrong

Weird, I tested it and it was working. I'll give another try.

aabadie:
Just tested it again and I think I understand your problem: in EU868 all tx power modes cannot be used see [this enum](https://github.com/RIOT-OS/RIOT/blob/master/sys/include/net/loramac.h#L527) => only tx power indexes from 0 to 7 are supported.

dylad:
Oh so index 0 stand for 20 dB if supported and 1 -> 14 dB right ?
I thought you have to set the value in dB unit so I was trying around 10 - 14 dB ...
I'll give it another shot

dylad:
I re test with and without, it works and it was working before
Sorry, I should have check the required values before reporting an non-existent issue.

aabadie:
> I should have check the required values before reporting an non-existent issue.

no problem, this shows that the documentation is not perfect.

aabadie:
> I re test with and without, it works and it was working before

Just understood this. And rechecked the loramac-node code: this PR is useless IMHO. I think we can close it. Do you agree @dylad ?

dylad:
@aabadie yes we can close this. Sorry about that...

PR #8840, opened 2018-03-27
makefiles: FIX boards generating a binary '.hex' file

cladmi:
### Contribution description

Some boards generate a '.hex' file that is a created with '-Obinary' which is wrong and is problematic to create rules by extension used in https://github.com/RIOT-OS/RIOT/pull/8838

I also include some cleanup for HEXFILE/ELFFILE/BINFILE variables in the boards Makefile. To use them and remove unused BINFILE.

### Issues/PRs references

Used in https://github.com/RIOT-OS/RIOT/pull/8838

### DIFF
--- a/boards/bluepill/Makefile.include
+++ b/boards/bluepill/Makefile.include
@@ -21,6 +21,7 @@ ifeq ($(PROGRAMMER),dfu-util)
   export RESET = # dfu-util has no support for resetting the device
 
   export OFLAGS = -O binary
+  HEXFILE = $(ELFFILE:.elf=.bin)
   export FFLAGS = -d 1d50:6017 -s 0x08002000:leave -D "$(HEXFILE)"
 else
 
--- a/boards/bluepill/Makefile.include
+++ b/boards/bluepill/Makefile.include
@@ -16,7 +16,6 @@ include $(RIOTMAKE)/tools/serial.inc.mk
 # (ground) GPIO B1.
 ifeq ($(PROGRAMMER),dfu-util)
   export ROM_OFFSET ?= 0x2000 # Skip the space needed by the embedded bootloader
-  export BINFILE = $(patsubst %.elf,%.bin,$(ELFFILE))
   export FLASHER = dfu-util
   export DEBUGGER = # no debugger
   export RESET = # dfu-util has no support for resetting the device
--- a/boards/calliope-mini/Makefile.include
+++ b/boards/calliope-mini/Makefile.include
@@ -13,7 +13,6 @@ include $(RIOTMAKE)/tools/serial.inc.mk
 PROGRAMMER ?= fscopy
 ifeq (fscopy,$(PROGRAMMER))
   export OFLAGS = -O ihex
-  export HEXFILE = $(ELFFILE:.elf=.hex)
   export FFLAGS =
   export DEBUGGER_FLAGS =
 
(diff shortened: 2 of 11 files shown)


boards/bluepill/Makefile.include
```diff
@@ -16,12 +16,12 @@ include $(RIOTMAKE)/tools/serial.inc.mk
 # (ground) GPIO B1.
 ifeq ($(PROGRAMMER),dfu-util)
   export ROM_OFFSET ?= 0x2000 # Skip the space needed by the embedded bootloader
-  export BINFILE = $(patsubst %.elf,%.bin,$(ELFFILE))
   export FLASHER = dfu-util
   export DEBUGGER = # no debugger
   export RESET = # dfu-util has no support for resetting the device
 
   export OFLAGS = -O binary
+  HEXFILE = $(ELFFILE:.elf=.bin)
```
kaspar030: I don't understand the improvement here.

cladmi: When I see  `hello-world.elf`, I expect it to be an `elf` executable.
When I see `hello-world.hex`, I expect it to bean `ihex` file, not a `binary` file.
Without this, for the boards I changed, you get a `binary` file called `hello-world.hex`.

The fix makes it generate a file named with `.bin` instead of `.hex`.

In master:
```
$ make  BOARD=bluepill PROGRAMMER=dfu-util
$ file bin/bluepill/hello-world.hex 
bin/bluepill/hello-world.hex: data

# which is different from an `ihex` file.
$ file bin/iotlab-m3/hello-world.hex
bin/iotlab-m3/hello-world.hex: ASCII text, with CRLF line terminators
```

Also, having the wrong format for the output is problematic for me, because in https://github.com/RIOT-OS/RIOT/pull/8838 I plan to add the following rules which would break if you do the opposite.

```
%.hex: %elf
    $(Q)$(OBJCOPY) $(OFLAGS) -Oihex $< $@
 %.bin: %.elf
    $(Q)$(OBJCOPY) $(OFLAGS) -Obinary $< $@
```

For fixing it, I used what is done in other files in RIOT
```
git grep  'ELFFILE:.elf=.bin'
boards/cc2538dk/Makefile.include:export HEXFILE = $(ELFFILE:.elf=.bin)
boards/common/remote/Makefile.include:export HEXFILE = $(ELFFILE:.elf=.bin)
boards/mbed_lpc1768/Makefile.include:export HEXFILE = $(ELFFILE:.elf=.bin)
boards/nrf6310/Makefile.include:export HEXFILE = $(ELFFILE:.elf=.bin)
boards/opencm904/Makefile.include:export HEXFILE = $(ELFFILE:.elf=.bin)
boards/openmote-cc2538/Makefile.include:  export HEXFILE = $(ELFFILE:.elf=.bin)
makefiles/tools/bossa.inc.mk:export HEXFILE = $(ELFFILE:.elf=.bin)
makefiles/tools/edbg.inc.mk:HEXFILE = $(ELFFILE:.elf=.bin)
makefiles/tools/jlink.inc.mk:export HEXFILE = $(ELFFILE:.elf=.bin)
```

Using `HEXFILE` variable for the output of `objcopy` is bad if it is not an `ihex` file but its what is currently done in RIOT. Changing this is addressed in the global PR by introducing `FLASHFILE` and this line will be replaced by `FLASHFILE = $(BINFILE)`.

kaspar030: ok, thanks!

kaspar030:
Do we need to test flashing for all affected boards?

kaspar030:
> Do we need to test flashing for all affected boards?

No.

PR #7604, opened 2017-09-14
Adds the atmega1284p CPU and the first version of the INGA board to RIOT

roberthartung:

Issue #6526, opened 2017-01-31
atmega based boards freeze when main thread is over


kYc0o:
While testing gnrc_minimal with arduino-mega2560 + XBee, the board seems to freeze when the main thread is over, as the examples behaves. However, if an infinite loop is used to avoid the thread to finish, the board doesn't freeze and it answers to pings.

I think it's related to the way threading works on this platform, maybe only task switching, but I didn't investigate further.

smlng:
@kYc0o I doubt this is a bug. I mean the `main` of `gnrc_minimal` states:

```

    /* main thread exits */
    return 0;
```

So I'd say this is not a problem in general, I understand (and checked) that other platforms do not stop answering pings after main thread completes. But then, whats the use of that? If main (and any other thread) are done, the node wouldn't do anything anymore.

OlegHahm:
Not sure I understand your point, but in GNRC the main thread is not responsible for responding to ICMP echo requests. Hence, it should reply with pongs whether the main thread terminates or not.

smlng:
I don't have an xbee shield available. However, I ran `tests/thread_msg` on board _arduino duemilanove_ and changed the loop to run 100 times (instead of 3). In this tests 3 threads are created by main and exchange messages for some time (i.e., 100x). The point is _main_ **exits** after creating the 3 threads, but the threads run - so it does **not** freeze. So the cause is not that main is over. 

Maybe if every thread is done it finally freezes, but then: whats the problem with that --other than its different to other platforms? However again: if the app (and all it threads) terminates, what use does it have to answer pings? Typically an IoT app will have some infinite loop driving the system, hence the problem should not arise?

smlng:
I also added a `xtimer_sleep(3)` in thread1 forcing the system to go into idle thread. I observed a problem with power management on cc2538 recently, which wasn't able to get out of idle after a sleep, I thought this might be a problem here, too -- however, the tests suggests it is not the case here.

smlng:
so basically that agrees with @OlegHahm, stating: 

> remember the main thread has no particular role or function in RIOT

So either there is no interrupt from the network device/driver or something blocks the system from handling such interrupt and switch to the network threads responsible to answer pings.

kYc0o:
However it doesn't explain why adding an infinite loop at the end keeps the board alive...

smlng:
[...]

PR #4572, opened 2016-01-04
kinetis/rtc: move rtt-to-rtc wrapper to drivers/periph_common

basilfx:
This should fix #4565.

There is no change in the code. I only moved the file, changed docs and added the defines (similar to the common methods for SPI).

I don't own a kinetis board. However, I have ensured that `tests/periph_rtc` would still compile (using `frdm-k64f`). Also, I tested this code on my local EFM32 port, which works too.

PR #8842, opened 2018-03-28
cpu/atmega_common: RTT and RTC support

ZetaR60:
This adds RTT and RTC support to the ATmegas. It is currently partially working on ATMega1284P using the mega-xplained board. So far as I can tell, the timer used with the real-time crystal is the same across all supported ATmegas, so if another board is added with a RTC crystal it should work right away. The crystal cannot be added to any of the Arduino boards without soldering, unfortunately.

There are a few problems with the code that I need to work out, but I need some advice on how to proceed:
- The use of RTC_NUMOF in rtt.c does not work, but I am not sure what to use to guard the RTC code.
- The use of thread_yield while waiting for busy conditions to clear lags messages in tests/periph_rtt a long time. But the wait can be up to 70ms, which seems really long to just spin if xtimer is not enabled.
- periph_rtt is required for the RTC to work, but I don't know how to add this dependency without adding it for a bunch of stuff that doesn't need it (e.g. it must be added to FEATURES_REQUIRED in tests/periph_rtc).
- ld does not seem to be able to find mktime and gmtime_r for rtc.c, leading to a 'undefined reference' error during linking.

The RTT code is based on some prior work by Robert Hartung in #7604

### DIFF
--- a/boards/mega-xplained/Makefile.features
+++ b/boards/mega-xplained/Makefile.features
@@ -2,6 +2,8 @@
 FEATURES_PROVIDED += periph_adc
 FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += periph_i2c
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
 FEATURES_PROVIDED += periph_spi
 FEATURES_PROVIDED += periph_timer
 FEATURES_PROVIDED += periph_uart
--- a/boards/mega-xplained/include/periph_conf.h
+++ b/boards/mega-xplained/include/periph_conf.h
@@ -42,6 +42,22 @@ extern "C" {
 #define CLOCK_CORECLOCK     (8000000UL)
 /** @} */
 
+/**
+ * @name RTC configuration
+ * @{
+ */
+#define RTC_NUMOF    (1U)
+/** @} */
+
+/**
+ * @name RTT configuration
+ * @{
+ */
+#define RTT_NUMOF        (1U)
+#define RTT_MAX_VALUE    (0x00FFFFFF)    /* 24-bit timer */
+#define RTT_FREQUENCY    (32U)
+/** @} */
+
 #ifdef __cplusplus
 }
 #endif
--- a/cpu/atmega_common/periph/rtc.c
+++ b/cpu/atmega_common/periph/rtc.c
@@ -1,0 +1,296 @@
+/*
+ * Copyright (C) 2018 Acutam Automation, LLC
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     cpu_atmega_common
+ * @ingroup     drivers_periph_rtc
+ *
+ * @{
+ *
+ * @file
+ * @brief       RTC interface wrapper for use with RTT modules.
+ *
+ * @author      Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * In order to safely sleep when using the RTT:
+ * 1. Disable interrupts
+ * 2. Write to one of the asynch registers (e.g. TCCR2A)
+ * 3. Wait for ASSR register's busy flags to clear
+ * 4. Re-enable interrupts
+ * 5. Sleep before interrupt re-enable takes effect
+ *
+ * @}
+ */
+
+#include <avr/interrupt.h>
+#include <time.h>
+
+#include "cpu.h"
+#include "irq.h"
+#include "periph/rtc.h"
+#include "periph/rtt.h"
+#include "periph_conf.h"
+#include "thread.h"
+
+#define ENABLE_DEBUG (0)
+#include "debug.h"
+
+/* guard file in case no RTC device is defined */
+#if RTC_NUMOF
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct {
+    time_t alarm;               /* alarm_cb when time == alarm */
+    rtc_alarm_cb_t alarm_cb;    /* callback called from RTC alarm */
+    void *alarm_arg;            /* argument passed to the callback */
+} rtc_state_t;
+
+static rtc_state_t rtc_state;
+
+/* RTC second counter. This is treated as-if it is a register as it is
+ * incremented by the overflow ISR. This MUST BE marked volatile for two
+ * reasons: to guarantee ordering with other register operations, and to
+ * force redundant access to occur (in case it was changed by the ISR). */
+static volatile time_t time_cnt;
+
+static inline void _asynch_wait(void)
+{
+    /* Wait until all busy flags clear. According to the datasheet,
+     * this can take up to 2 positive edges of TOSC1 (32kHz). */
+    while (ASSR & ((1 << TCN2UB) | (1 << OCR2AUB) | (1 << OCR2BUB)
+                   | (1 << TCR2AUB) | (1 << TCR2BUB))) {}
+}
+
+/* This safely reads time_cnt and TCNT2, since an atomic read cannot be provided.
+ * Note: still requires prior synch using _asynch_wait() */
+static inline time_t _safe_time_get(void)
+{
+    time_t time_tmp;
+    uint8_t irq_flag, cnt_tmp;
+
+    do {
+        /* Operations must occur in this order: */
+        irq_flag = TIFR2 & (1 << TOV2);
+        time_tmp = time_cnt;
+        cnt_tmp = TCNT2;
+
+    /* Detect rollover after time_tmp read. If condition is met, then rollover
+     * just occured, and it is safe to read again now. */
+    } while ((time_tmp != time_cnt) || (irq_flag != (TIFR2 & (1 << TOV2))));
+
+    if (irq_flag) {
+        time_tmp += 8;
+    }
+
+    return time_tmp + (cnt_tmp >> 5);
+}
+
+void rtc_init(void)
+{
+    /* RTC depends on RTT */
+    rtt_init();
+}
+
+int rtc_set_time(struct tm *time)
+{
+    uint8_t irq_flag;
+
+    /* Make sure it is safe to read TCNT2, in case we just woke up */
+    DEBUG("RTT sleeps until safe to read TCNT2\n");
+    TCCR2A = 0;
+    _asynch_wait();
+
+    /* Convert to seconds since the epoch */
+    time_t time_secs = mk_gmtime(time);
+
+    /* Make non-atomic writes atomic */
+    unsigned state = irq_disable();
+
+    do {
+        /* Operations must occur in this order: */
+        time_cnt = time_secs - (TCNT2 >> 5);
+        irq_flag = TIFR2 & (1 << TOV2);
+
+    /* Detect rollover after time_cnt is written. If condition is met, then
+     * rollover just occured, and it is safe to write now. */
+    } while (irq_flag != (TIFR2 & (1 << TOV2)));
+
+    irq_restore(state);
+
+    DEBUG("RTC set time: %" PRIu32 " seconds\n", time_cnt);
+
+    return 0;
+}
+
+int rtc_get_time(struct tm *time)
+{
+    /* Make sure it is safe to read TCNT2, in case we just woke up */
+    DEBUG("RTT sleeps until safe to read TCNT2\n");
+    TCCR2A = 0;
+    _asynch_wait();
+
+    time_t time_secs = _safe_time_get();
+
+    /* Convert from seconds since the epoch */
+    gmtime_r(&time_secs, time);
+
+    DEBUG("RTC get time: %" PRIu32 " seconds\n", time_secs);
+
+    return 0;
+}
+
+int rtc_set_alarm(struct tm *time, rtc_alarm_cb_t cb, void *arg)
+{
+    /* Disable alarm */
+    rtc_clear_alarm();
+
+    /* Make sure it is safe to read TCNT2, in case we just woke up, and */
+    /* safe to write OCR2B (in case it was busy) */
+    DEBUG("RTC sleeps until safe read TCNT2 and to write OCR2B\n");
+    TCCR2A = 0;
+    _asynch_wait();
+
+    time_t now = _safe_time_get();
+    time_t alarm_tmp = mk_gmtime(time);
+
+    if (alarm_tmp < now) {
+        DEBUG("RTC alarm set in the past. Time: %" PRIu32 " seconds, alarm: %"
+              PRIu32 "\n", now, rtc_state.alarm);
+        cb(arg);
+        return 0;
+    }
+
+    /* Calculate time to wait, based on TCNT2 = 0 */
+    time_t diff = rtc_state.alarm - now;
+
+    /* Make non-atomic writes atomic */
+    unsigned state = irq_disable();
+
+    /* Set alarm time */
+    rtc_state.alarm = alarm_tmp;
+
+    /* Prepare the counter for sub 8-second precision */
+    OCR2B = (uint8_t)(diff << 5);
+
+    rtc_state.alarm_arg = arg;
+    rtc_state.alarm_cb = cb;
+
+    irq_restore(state);
+
+    DEBUG("RTC set alarm: %" PRIu32 " seconds, OCR2B: %" PRIu8 "\n",
+          rtc_state.alarm, OCR2B);
+
+    /* Enable irq only if alarm is in the 8s period before it overflows */
+    if (diff < 8) {
+        /* Clear interrupt flag */
+        TIFR2 = (1 << OCF2B);
+
+        /* Enable interrupt */
+        TIMSK2 |= (1 << OCIE2B);
+
+        DEBUG("RTT alarm interrupt active\n");
+    }
+    else {
+        DEBUG("RTT alarm interrupt not active\n");
+    }
+
+    return 0;
+}
+
+int rtc_get_alarm(struct tm *time)
+{
+    /* Convert from seconds since the epoch */
+    gmtime_r(&rtc_state.alarm, time);
+
+    DEBUG("RTC get alarm: %" PRIu32 " seconds\n", alarm);
+
+    return 0;
+}
+
+void rtc_clear_alarm(void)
+{
+    /* Make non-atomic writes atomic */
+    unsigned state = irq_disable();
+
+    /* Disable alarm interrupt */
+    TIMSK2 &= ~(1 << OCIE2B);
+
+    rtc_state.alarm_cb = NULL;
+    rtc_state.alarm_arg = NULL;
+
+    irq_disable(state);
+}
+
+void rtc_poweron(void)
+{
+    rtt_poweron();
+}
+
+void rtc_poweroff(void)
+{
+    rtt_poweroff();
+}
+
+void atmega_rtc_incr(void)
+{
+    time_cnt += 8;
+
+    /* If alarm not set, nothing else to do */
+    if (rtc_state.alarm_cb == NULL) {
+        return;
+    }
+
+    /* Check to see if alarm was missed */
+    if (rtc_state.alarm <= time_cnt) {
+        /* Clear callback */
+        rtc_alarm_cb_t cb = rtc_state.alarm_cb;
+        rtc_state.alarm_cb = NULL;
+
+        /* Execute callback */
+        cb(rtc_state.alarm_arg);
+
+        return;
+    }
+
+    /* Enable irq only if alarm is in the 8s period before it overflows */
+    if ((rtc_state.alarm - time_cnt) < 8) {
+        /* Clear interrupt flag */
+        TIFR2 = (1 << OCF2B);
+
+        /* Enable interrupt */
+        TIMSK2 |= (1 << OCIE2B);
+    }
+}
+
+ISR(TIMER2_COMPB_vect)
+{
+    __enter_isr();
+    /* Disable alarm interrupt */
+    TIMSK2 &= ~(1 << OCIE2B);
+
+    /* Execute callback */
+    if (rtc_state.alarm_cb != NULL) {
+        /* Clear callback */
+        rtc_alarm_cb_t cb = rtc_state.alarm_cb;
+        rtc_state.alarm_cb = NULL;
+
+        /* Execute callback */
+        cb(rtc_state.alarm_arg);
+    }
+
+    __exit_isr();
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* RTC_NUMOF */
--- a/cpu/atmega_common/periph/rtt.c
+++ b/cpu/atmega_common/periph/rtt.c
@@ -1,0 +1,336 @@
+/*
+ * Copyright (C) 2018 Acutam Automation, LLC
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/*
+ * @ingroup cpu_atmega_common
+ * @ingroup drivers_periph_rtt
+ * @{
+ *
+ * @file
+ * @brief       Low-level ATmega RTT driver implementation
+ *
+ * @author      Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * In order to safely sleep when using the RTT:
+ * 1. Disable interrupts
+ * 2. Write to one of the asynch registers (e.g. TCCR2A)
+ * 3. Wait for ASSR register's busy flags to clear
+ * 4. Re-enable interrupts
+ * 5. Sleep before interrupt re-enable takes effect
+ *
+ * @}
+ */
+
+#include <avr/interrupt.h>
+
+#include "cpu.h"
+#include "irq.h"
+#include "periph/rtt.h"
+#include "periph_conf.h"
+#include "thread.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+/* guard file in case no RTT device is defined */
+#if RTT_NUMOF
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if MODULE_PERIPH_RTC
+extern void atmega_rtc_incr(void);
+#endif
+
+typedef struct {
+    uint16_t ext_comp;          /* Extend compare to 24-bits */
+    rtt_cb_t alarm_cb;          /* callback called from RTT alarm */
+    void *alarm_arg;            /* argument passed to the callback */
+    rtt_cb_t overflow_cb;       /* callback called when RTT overflows */
+    void *overflow_arg;         /* argument passed to the callback */
+} rtt_state_t;
+
+static rtt_state_t rtt_state;
+
+/* Extend TCNT2 to 24-bits. This is treated as-if it is a register as it is
+ * incremented by the overflow ISR. This MUST BE marked volatile for two
+ * reasons: to guarantee ordering with other register operations, and to
+ * force redundant access to occur (in case it was changed by the ISR). */
+static volatile uint16_t ext_cnt;
+
+static inline void _asynch_wait(void)
+{
+    /* Wait until all busy flags clear. According to the datasheet,
+     * this can take up to 2 positive edges of TOSC1 (32kHz). */
+    while (ASSR & ((1 << TCN2UB) | (1 << OCR2AUB) | (1 << OCR2BUB)
+                   | (1 << TCR2AUB) | (1 << TCR2BUB))) {}
+}
+
+/* This safely reads ext_cnt and TCNT2, since an atomic read cannot be provided.
+ * Note: still requires prior synch using _asynch_wait() */
+static inline uint32_t _safe_cnt_get(void)
+{
+    uint16_t ext_cnt_tmp;
+    uint8_t irq_flag, cnt_tmp;
+
+    do {
+        /* Operations must occur in this order: */
+        irq_flag = TIFR2 & (1 << TOV2);
+        ext_cnt_tmp = ext_cnt;
+        cnt_tmp = TCNT2;
+
+    /* Detect rollover after ext_cnt read. If condition is met, then rollover
+     * just occured, and it is safe to read again now. */
+    } while ((ext_cnt_tmp != ext_cnt) || (irq_flag != (TIFR2 & (1 << TOV2))));
+
+    if (irq_flag) {
+        ext_cnt_tmp++;
+    }
+
+    return (((uint32_t)ext_cnt_tmp << 8) | (uint32_t)cnt_tmp);
+}
+
+void rtt_init(void)
+{
+    DEBUG("Initializing RTT\n");
+
+    rtt_poweron();
+
+    /*
+     * From the datasheet section "Asynchronous Operation of Timer/Counter2"
+     * p148 for ATmega1284P.
+     * 1. Disable the Timer/Counter2 interrupts by clearing OCIE2x and TOIE2.
+     * 2. Select clock source by setting AS2 as appropriate.
+     * 3. Write new values to TCNT2, OCR2x, and TCCR2x.
+     * 4. To switch to asynchronous: Wait for TCN2UB, OCR2xUB, TCR2xUB.
+     * 5. Clear the Timer/Counter2 Interrupt Flags.
+     * 6. Enable interrupts, if needed
+     */
+
+    /* Disable all timer 2 interrupts */
+    TIMSK2 = 0;
+
+    /* Select asynchronous clock source */
+    ASSR = (1 << AS2);
+
+    /* Set counter to 0 */
+    TCNT2 = 0;
+
+    /* Reset compare values */
+    OCR2A = 0;
+    OCR2B = 0;
+
+    /* Reset timer control */
+    TCCR2A = 0;
+
+    /* 32768Hz / 1024 = 32 ticks per second */
+    TCCR2B = (1 << CS22) | (1 << CS21) | (1 << CS20);
+
+    /* Wait until not busy anymore */
+    DEBUG("RTT waits until ASSR not busy\n");
+    _asynch_wait();
+
+    /* Clear interrupt flags */
+    /* Oddly, this is done by writing ones; see datasheet */
+    TIFR2 = (1 << OCF2B) | (1 << OCF2A) | (1 << TOV2);
+
+    /* Enable 8-bit overflow interrupt */
+    TIMSK2 |= (1 << TOIE2);
+
+    DEBUG("RTT initialized\n");
+}
+
+void rtt_set_overflow_cb(rtt_cb_t cb, void *arg)
+{
+    /* Make non-atomic write to callback atomic */
+    unsigned state = irq_disable();
+
+    rtt_state.overflow_cb = cb;
+    rtt_state.overflow_arg = arg;
+
+    irq_restore(state);
+}
+
+void rtt_clear_overflow_cb(void)
+{
+    /* Make non-atomic write to callback atomic */
+    unsigned state = irq_disable();
+
+    rtt_state.overflow_cb = NULL;
+    rtt_state.overflow_arg = NULL;
+
+    irq_restore(state);
+}
+
+uint32_t rtt_get_counter(void)
+{
+    /* Make sure it is safe to read TCNT2, in case we just woke up */
+    DEBUG("RTT sleeps until safe to read TCNT2\n");
+    TCCR2A = 0;
+    _asynch_wait();
+
+    return _safe_cnt_get();
+}
+
+void rtt_set_counter(uint32_t counter)
+{
+    /* Wait until not busy anymore (should be immediate) */
+    DEBUG("RTT sleeps until safe to write TCNT2\n");
+    _asynch_wait();
+
+    /* Make non-atomic writes atomic (for concurrent access) */
+    unsigned state = irq_disable();
+
+    /* Prevent overflow flag from being set during update */
+    TCNT2 = 0;
+
+    /* Clear overflow flag */
+    /* Oddly, this is done by writing a one; see datasheet */
+    TIFR2 = 1 << TOV2;
+
+    ext_cnt = (uint16_t)(counter >> 8);
+    TCNT2 = (uint8_t)counter;
+
+    irq_restore(state);
+
+    DEBUG("RTT set counter TCNT2: %" PRIu8 ", ext_cnt: %" PRIu16 "\n",
+          TCNT2, ext_cnt);
+}
+
+void rtt_set_alarm(uint32_t alarm, rtt_cb_t cb, void *arg)
+{
+    /* Disable alarm */
+    rtt_clear_alarm();
+
+    /* Make sure it is safe to read TCNT2, in case we just woke up, and */
+    /* safe to write OCR2B (in case it was busy) */
+    DEBUG("RTC sleeps until safe read TCNT2 and to write OCR2B\n");
+    TCCR2A = 0;
+    _asynch_wait();
+
+    uint32_t now = _safe_cnt_get();
+
+    if (alarm < now) {
+        DEBUG("RTC alarm set in the past. Time: %" PRIu32 " seconds, alarm: %"
+              PRIu32 "\n", now, alarm);
+        cb(arg);
+        return;
+    }
+
+    /* Make non-atomic writes atomic */
+    unsigned state = irq_disable();
+
+    /* Set the alarm value. Atomic for concurrent access */
+    rtt_state.ext_comp = (uint16_t)(alarm >> 8);
+    OCR2A = (uint8_t)alarm;
+
+    rtt_state.alarm_cb = cb;
+    rtt_state.alarm_arg = arg;
+
+    irq_restore(state);
+
+    DEBUG("RTT set alarm TCNT2: %" PRIu8 ", OCR2A: %" PRIu8 "\n", TCNT2, OCR2A);
+
+    /* Enable alarm interrupt only if it will trigger before overflow */
+    if (rtt_state.ext_comp <= (uint16_t)(now >> 8)) {
+        /* Clear interrupt flag */
+        TIFR2 = (1 << OCF2A);
+
+        /* Enable interrupt */
+        TIMSK2 |= (1 << OCIE2A);
+
+        DEBUG("RTT alarm interrupt active\n");
+    }
+    else {
+        DEBUG("RTT alarm interrupt not active\n");
+    }
+}
+
+uint32_t rtt_get_alarm(void)
+{
+    return (((uint32_t)rtt_state.ext_comp << 8) | (uint32_t)OCR2A);
+}
+
+void rtt_clear_alarm(void)
+{
+    /* Make non-atomic writes atomic */
+    unsigned state = irq_disable();
+
+    /* Disable alarm interrupt */
+    TIMSK2 &= ~(1 << OCIE2A);
+
+    rtt_state.alarm_cb = NULL;
+    rtt_state.alarm_arg = NULL;
+
+    irq_restore(state);
+}
+
+void rtt_poweron(void)
+{
+    power_timer2_enable();
+}
+
+void rtt_poweroff(void)
+{
+    power_timer2_disable();
+}
+
+ISR(TIMER2_OVF_vect)
+{
+    __enter_isr();
+
+    ext_cnt++;
+
+    /* Enable RTT alarm if overflowed enough times */
+    if (rtt_state.ext_comp <= ext_cnt) {
+        /* Clear interrupt flag */
+        TIFR2 = (1 << OCF2A);
+
+        /* Enable interrupt */
+        TIMSK2 |= (1 << OCIE2A);
+    }
+
+#if MODULE_PERIPH_RTC
+    /* Increment RTC by 8 seconds */
+    atmega_rtc_incr();
+#endif
+
+    /* Virtual 24-bit timer overflowed */
+    if (ext_cnt == 0) {
+        /* Execute callback */
+        if (rtt_state.overflow_cb != NULL) {
+            rtt_state.overflow_cb(rtt_state.overflow_arg);
+        }
+    }
+    __exit_isr();
+}
+
+ISR(TIMER2_COMPA_vect)
+{
+    __enter_isr();
+    /* Disable alarm interrupt */
+    TIMSK2 &= ~(1 << OCIE2A);
+
+    if (rtt_state.alarm_cb != NULL) {
+        /* Clear callback */
+        rtt_cb_t cb = rtt_state.alarm_cb;
+        rtt_state.alarm_cb = NULL;
+
+        /* Execute callback */
+        cb(rtt_state.alarm_arg);
+    }
+
+    __exit_isr();
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* RTT_NUMOF */
--- a/boards/jiminy-mega256rfr2/Makefile.dep
+++ b/boards/jiminy-mega256rfr2/Makefile.dep
@@ -1,1 +1,3 @@
 USEMODULE += boards_common_atmega
+
+include $(RIOTCPU)/atmega_common/Makefile.dep
--- a/boards/jiminy-mega256rfr2/Makefile.features
+++ b/boards/jiminy-mega256rfr2/Makefile.features
@@ -4,5 +4,7 @@ include $(RIOTBOARD)/common/arduino-atmega/Makefile.features
 # Put defined MCU peripherals here (in alphabetical order)
 # Peripherals are defined in common/arduino-atmega/Makefile.features
 # Add only additional Peripherals
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
 
 -include $(RIOTCPU)/atmega256rfr2/Makefile.features
--- a/boards/jiminy-mega256rfr2/include/periph_conf.h
+++ b/boards/jiminy-mega256rfr2/include/periph_conf.h
@@ -33,6 +33,22 @@ extern "C" {
 #define CLOCK_CORECLOCK     (8000000UL)
 /** @} */
 
+/**
+ * @name RTC configuration
+ * @{
+ */
+#define RTC_NUMOF    (1U)
+/** @} */
+
+/**
+ * @name RTT configuration
+ * @{
+ */
+#define RTT_NUMOF        (1U)
+#define RTT_MAX_VALUE    (0x00FFFFFF)    /* 24-bit timer */
+#define RTT_FREQUENCY    (32U)
+/** @} */
+
 #ifdef __cplusplus
 }
 #endif
--- a/boards/mega-xplained/Makefile.dep
+++ b/boards/mega-xplained/Makefile.dep
@@ -4,3 +4,5 @@ ifneq (,$(filter saul_default,$(USEMODULE)))
   USEMODULE += saul_adc
   USEMODULE += saul_gpio
 endif
+
+include $(RIOTCPU)/atmega_common/Makefile.dep
--- a/cpu/atmega_common/Makefile.dep
+++ b/cpu/atmega_common/Makefile.dep
@@ -1,0 +1,3 @@
+ifneq (,$(filter periph_rtc,$(USEMODULE)))
+    USEMODULE += periph_rtt
+endif


cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,337 @@
...
+
+    /* Execute callback */
+    if (rtt_state.alarm_cb != NULL) {
+        rtt_state.alarm_cb(rtt_state.alarm_arg);
+    }
+
+    /* Wait until it is safe to re-enter power-save */
+    TCCR2A = TCCR2A;
+    while( ASSR & (1 << TCR2AUB) ) {
+        if (sched_context_switch_request) {
```
kaspar030: why is this in the loop?

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,190 @@
...
+    TIMSK2 &= ~(1 << OCIE2B);
+
+    /* Set callback */
+    rtc_state.alarm_cb = cb;
+    rtc_state.alarm_arg = arg;
+
+    /* Wait until not busy anymore (should be immediate) */
+    DEBUG("RTC sleeps until safe to write OCR2B\n");
+    while( ASSR & (1 << OCR2BUB) ) {
+#ifdef MODULE_XTIMER
```
kaspar030: I suggest factoring this ifdef into it's own static function (maybe with the umber of RTC cycles as parameter), it is used at least twice.

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,337 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if RTC_NUMOF
+extern void atmega_rtc_incr(void);
+#endif
+
+typedef struct {
+    uint8_t ext_cnt[3];      /* Counter to make 8-bit timer 32-bit */
```
kaspar030: How about just using uint32_t? IMO the two wasted bytes in the struct might be worth increasing readability throughut the file.

cpu/atmega_common/periph/rtt.c
ZetaR60: According to the datasheet: if the MCU is woken from sleep by the asynchronous timer (as we have here in the interrupt), and then sleeps again before 1 cycle of the timer has passed, it doesn't reset the interrupt logic and will fail to wake a second time when an interrupt should have been generated. I will add some comments to make the purpose of each wait clearer.

cpu/atmega_common/periph/rtt.c
ZetaR60: I started with uint32_t, but the fact that I was wasting two bytes (and increasing the memory usage by 20%) kept bothering me. It doesn't seem like a lot, but you could be using this on the ATmega328P, which only has 2kB of ram!

cpu/atmega_common/periph/rtt.c
kaspar030: Ok, so ```while( ASSR & (1 << TCR2AUB) ) {}``` does the waiting. But why is the ```if (sched_... ``` in that while loop? It could be right after? Isn't ```thread_yield()``` executed multiple times?

ZetaR60: Oh. That was just because I saw that it was done that way in other interrupt routines. I assumed that the effect would be to switch threads; the switching would unset the context request; and then it would spin without switching threads until the check passed. I would use xtimer, but the docs for it say not to use it in interrupts because it will just spin anyway.

ZetaR60: Removed.

cpu/atmega_common/periph/rtc.c
ZetaR60: Done. Also done for RTT.

ZetaR60:
I have found that the local copy of cpu/atmega_common/avr-libc-extra/time.h conflicts with the system avr-libc, and makes the time.h functions produce garbage output. Since this was actually used almost nowhere, I removed it. There was a minor struct that is not provided by avr-libc time.h, which I moved into cpu/atmega_common/include/sys/time.h. I can split this off into another PR, if you think it is necessary.

I have fixed all of the problems listed in my first comment, except for the lagged messages. There is one final tricky bug I am trying to track down: there is a race condition that is related to use of xtimer_usleep that I have not been able to fix. When xtimer is disabled, everything is fine; when xtimer is enabled and debug is enabled, everything is fine; when only xtimer is enabled, some checks hang and alarms never trigger.

cpu/atmega_common/periph/rtt.c
ZetaR60: These are now uint16_t, making the RTT a 24-bit timer. Both readable, and no wasted bytes.

ZetaR60:
time.h change is now #8857

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,270 @@
...
+    TIFR2 &= ~( (1 << OCF2B)  | (1 << OCF2A) | (1 << TOV2) );
+
+    /* Initialize callbacks */
+    rtt_state.alarm_cb = NULL;
+    rtt_state.overflow_cb = NULL;
+
+    /* Enable 8-bit overflow interrupt */
+    TIMSK2 |= (1 << TOIE2);
+
+    irq_enable();
```
gebart: this will force enable IRQs even if they were disabled before. This is a safer alternative: 
```c
unsigned mask = irq_disable();
...
irq_restore(mask);
```

gebart:
Regarding using RTC as a wrapper for RTT, see also: 
- https://github.com/RIOT-OS/RIOT/blob/master/cpu/kinetis/periph/rtc.c
- https://github.com/RIOT-OS/RIOT/issues/4565
- https://github.com/RIOT-OS/RIOT/pull/4572

ZetaR60:
Unlike a simple RTC wrapper like in cpu/kinetis, this implementation can be used concurrently to the RTT. This is because it keeps time independently of the RTT, and it utilizes a different counter channel on the timer.

ZetaR60:
I have found that the xtimer_usleep problem is an instance of #6526 and not directly related to the code here. The hang on tests/periph_rtc is solved by turning off xtimer, or adding a while(1) to the end of main. It seems to be the same problem every other interrupt has after exiting main.

I don't think there are any remaining outstanding issues. As soon as #8857 merges, it should be ready for CI build testing.

cpu/atmega_common/periph/rtt.c
ZetaR60: Removed.

While the RTT requires IRQs to be enabled, it never needs to disable them globally.

ZetaR60:
The xtimer_usleep problem is fixed by #8904 for tests/periph_rtt, and thus is probably an instance of #8896.

Unfortunately, there seems to be another race condition either added by #8904, or uncovered by it, that prevents tests/periph_rtc from working.

ZetaR60:
It seems that moving rtc_init conflicts with the organization of the init systems (see #8914). It is also placed early enough that xtimer cannot be used. Swapping threads is not going to be useful because other threads have not been started (it is before kernel_init). Is it acceptable in this case to spin for 30-60ms while waiting for the busy flags to clear, or is there another alternative? This is required by the hardware for safe access to the asynchronous registers.

gebart:
OK, I briefly skimmed the ATMega1284p data sheet to find out more about the hardware.
Assumptions:
- From the code, it seems like TOSC1 clock signal is assumed coming from a 32768 Hz RTC crystal.
- We want a relatively high prescaler on that clock source so that we don't need to service the overflow interrupt too often, since the timer/counter is only 8 bits wide.

From the data sheet:
>When writing to one of the registers TCNT2, OCR2x, or TCCR2x, the value is transferred to a
temporary register, and **latched after two positive edges on TOSC1**. The user should not write
a new value before the contents of the temporary register have been transferred to its
destination. Each of the five mentioned registers have their individual temporary register, which
means that e.g. writing to TCNT2 does not disturb an OCR2x write in progress. To detect that a
transfer to the destination register has taken place, the Asynchronous Status Register – ASSR
has been implemented.

My conclusion from that text is that the synchronization takes 1 < t <= 2 [TOSC1 ticks], which is running at 32768 Hz, meaning that the syncronization is between 30.5 µs and 61 µs. The nomenclature in the rest of that chapter is that TOSC1 is the input to the prescaler, and not the divided counter signal.

Please tell me if you have any indication that the synchronization is indeed using the divided clock signal instead of the unscaled one. To me it seems like a pretty bad hardware design choice to let the silicon take 1/16 second to read/write a value from a timer. AFAICT using TOSC1 directly instead of the divided signal would not introduce any asynchronous problems since the TOSC1 signal and the scaled clock signal are synchronous compared to each other, since they are derived from the same signal, albeit with some small, non-zero, phase shift because of propagation delays in the prescaler.

If my conclusions are true, you should drop the xtimer calls completely and just spin for the sync flags to clear, 60 µs is definitely OK to spin while setting a slow 32 Hz timer target in my opinion. It is worth mentioning this spin in the documentation or code comments somewhere so that developers can find out about this without studying the details of the data sheet. I suggest putting it in the documentation block for the `__asynch_wait` function.

Also, please consider any possible race conditions when using the RTC and RTT from two different threads. It may be necessary to mask interrupts in some parts of the low level hardware accesses to prevent clobbering each others settings, especially in read-modify-write situations.

ZetaR60:
@gebart Sorry for my confusion. I thought I had seen an entry indicating that the synchronization unit used the prescaler and not TOSC1 directly, but I was mistaken.

__asynch_wait now spins, and I put a comment on about how long that will take. The only remaining problem is solved by #8857. Let me know when you want me to squash.

There shouldn't be any race conditions with using RTT and RTC together since the RTC does not actually change any RTT settings. The alarm used by the RTC is handled by counter B, while the alarm used by the RTT is counter A. So long as nothing runs rtt_poweroff the worst that can happen is a 0-8 glitch in timing when running rtt_set_counter while the RTC is running, but this should not prevent any ISRs from firing properly (just delay them). I suppose I could add some code to compensate for the timing glitch, but that would require editing RTC variables in the RTT code.

gebart:
@ZetaR60 I don't think you need to handle that case here and now, I just wanted to make sure you had considered what the potential races are in this implementation.

I just realized something, you need to rename __async_wait, double underscores are forbidden in identifiers in C, outside of the compiler provided macros and support libs

ZetaR60:
Changed to _asynch_wait. I was a bit confused about single and double underscores, considering the usage in e.g. cpu/atmega_common/thread_arch.c. Am I correct in thinking that single underscores are preferred for naming static functions in RIOT libraries?

Also ran uncrustify.

ZetaR60:
Squashed with requested changes incorporated.

gebart:
@ZetaR60 it is easier to review the PR if you don't squash your changes until you get a request to squash by a maintainer. Otherwise the person reviewing will need to try to remember what the code was like before. Making reviewing easy helps getting your PR merged more quickly.

ZetaR60:
@gebart Sorry. I thought that the amount of time waited was enough. I managed to unsquash from a backup.

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,259 @@
...
+    /* Wait until all busy flags clear. According to the datasheet,
+     * this can take up to 2 positive edges of TOSC1 (32kHz). */
+    while (ASSR & ((1 << TCN2UB) | (1 << OCR2AUB) | (1 << OCR2BUB)
+                   | (1 << TCR2AUB) | (1 << TCR2BUB))) ;
+}
+
+ISR(TIMER2_OVF_vect) {
+    __enter_isr();
+    /* Enable RTT alarm if overflowed enough times */
+    if (rtt_state.ext_comp == rtt_state.ext_cnt) {
```
Josar:
```
if (rtt_state.ext_comp == rtt_state.ext_cnt+1) 
```

As it gets incremented below.

ZetaR60:
Thanks for the test! The issue should be fixed now. Also rebased on trunk to pick up the changes to context swap handling.

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,261 @@
...
+
+void _asynch_wait(void)
+{
+    /* Wait until all busy flags clear. According to the datasheet,
+     * this can take up to 2 positive edges of TOSC1 (32kHz). */
+    while (ASSR & ((1 << TCN2UB) | (1 << OCR2AUB) | (1 << OCR2BUB)
+                   | (1 << TCR2AUB) | (1 << TCR2BUB))) ;
+}
+
+ISR(TIMER2_OVF_vect) {
```
Josar: Could be reduced futher more as the 16 bit should automatically overflow. 
```
ISR(TIMER2_OVF_vect) {
    __enter_isr();

    /* Enable RTT alarm if overflowed enough times */
    if (++rtt_state.ext_cnt == rtt_state.ext_comp ) {
        TIMSK2 |= (1 << OCIE2A);
    }

#if MODULE_PERIPH_RTC
    /* Increment RTC by 8 seconds */
    atmega_rtc_incr();
#endif

    /* Virtual 24-bit timer overflow */
    if (rtt_state.ext_cnt == 0) {

        /* Execute callback */
        if (rtt_state.overflow_cb != NULL) {
            rtt_state.overflow_cb(rtt_state.overflow_arg);
        }
    }

    if (sched_context_switch_request) {
        thread_yield();
    }
    __exit_isr();
}
```

ZetaR60: "Should", yes, and would in almost all cases. However, the standard states that the result of an overflow is undefined.

Josar: This only accounts for signed values. unsigned values are well defined in the C standard.

```
§6.2.5/9
A computation involving unsigned operands can never overflow,
because a result that cannot be represented by the resulting unsigned integer type is
reduced modulo the number that is one greater than the largest value that can be
represented by the resulting type.
```

ZetaR60: My mistake. Removed the extra logic.

boards/mega-xplained/Makefile.features
```diff
@@ -2,6 +2,8 @@
 FEATURES_PROVIDED += periph_adc
 FEATURES_PROVIDED += periph_gpio
 FEATURES_PROVIDED += periph_i2c
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
```
kYc0o: For other atmega to work, you need to add these into it's Makefile.features.

boards/mega-xplained/include/periph_conf.h
```diff
@@ -65,6 +65,22 @@ extern "C" {
...
+#define RTC_NUMOF    (1U)
+/** @} */
+
+/**
+ * @name RTT configuration
+ * @{
+ */
+#define RTT_NUMOF        (1U)
+#define RTT_MAX_VALUE    (0x00FFFFFF)    /* 24-bit timer */
+#define RTT_FREQUENCY    (32U)
```
kYc0o: This also needs to go for all atmegas. Do you know if all these settings work for all?

ZetaR60:
@kYc0o Unfortunately, this PR does not add support for _all_ ATmegas, because doing so is impossible without physical modifications to the boards. For example, arduino-mega2560 doesn't use the pins for the RTC crystal for anything, but also doesn't expose them on the headers (they are just not connected). I don't think that this PR should add RTT/RTC support to the arduino-mega2560 because adding the crystal would require careful soldering to these NC pins. Other boards are similar.

The only other boards that use an RTC crystal are waspmote-pro and jiminy-mega256rfr2. waspmote-pro provide it via the DS3231SN and not the ATmega. I can add support in jiminy-mega256rfr2 though.

boards/mega-xplained/Makefile.features
ZetaR60: Added support to jiminy-mega256rfr2. See comment below review.

boards/mega-xplained/include/periph_conf.h
ZetaR60: Added support to jiminy-mega256rfr2. See comment below review.

The settings are based on how the driver works, so they are the same for all users of this driver.

ZetaR60:
Note that Murdock is expected to fail until #9405 is resolved.

kYc0o:
I see, then I'll try to test in mega-xplained and ACK it afterwards. Please add to the jiminy board if possible, so we have as much boards as possible already covered by your code.

kYc0o:
sorry, forgot to hit "comment", but you have already reacted. Thanks for the contribution!

@Josar can you test if it works correctly for jiminy?

ZetaR60:
@Josar can you re-test with the latest changes? Thanks.

Josar:
@ZetaR60 
One thing giving me a headache is this
avr8-gnu-toolchain
```
    struct tm {
        int8_t          tm_sec; /**< seconds after the minute - [ 0 to 59 ] */
        int8_t          tm_min; /**< minutes after the hour - [ 0 to 59 ] */
        int8_t          tm_hour; /**< hours since midnight - [ 0 to 23 ] */
        int8_t          tm_mday; /**< day of the month - [ 1 to 31 ] */
        int8_t          tm_wday; /**< days since Sunday - [ 0 to 6 ] */
        int8_t          tm_mon; /**< months since January - [ 0 to 11 ] */
        int16_t         tm_year; /**< years since 1900 */
        int16_t         tm_yday; /**< days since January 1 - [ 0 to 365 ] */
        int16_t         tm_isdst; /**< Daylight Saving Time flag */
    };
```
vs atmega_common/avr-libc-extra

```
struct tm {
        int tm_sec;
        int tm_min;
        int tm_hour;
        int tm_mday;
        int tm_wday;
        int tm_mon;
        int tm_year;
        int tm_yday;
        int tm_isdst;
    };
```    

This produces bad output values for `print_time`  in `test/periph_rtc` as avr defines integer as 16bit.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,218 @@
...
+        rtt_init();
+    }
+}
+
+int rtc_set_time(struct tm *time)
+{
+    uint8_t offset;
+
+    /* Make sure it is safe to read TCNT2, in case we just woke up */
+    DEBUG("RTT sleeps until safe to read TCNT2\n");
```
Josar: I would prefer a Debug msg printing the time. This would be of much more help.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,218 @@
...
+
+    return 0;
+}
+
+int rtc_get_time(struct tm *time)
+{
+    time_t time_secs;
+
+    /* Make sure it is safe to read TCNT2, in case we just woke up */
+    DEBUG("RTT sleeps until safe to read TCNT2\n");
```
Josar: See last comment.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,218 @@
...
+}
+
+int rtc_set_alarm(struct tm *time, rtc_alarm_cb_t cb, void *arg)
+{
+    /* Disable alarm interrupt */
+    TIMSK2 &= ~(1 << OCIE2B);
+    rtc_state.alarm_cb = NULL;
+
+    /* Wait until not busy anymore (should be immediate) */
+    DEBUG("RTC sleeps until safe to write OCR2B\n");
```
Josar: Also here msg with now and alarm would be of more help.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,218 @@
...
+}
+
+int rtc_set_time(struct tm *time)
+{
+    uint8_t offset;
+
+    /* Make sure it is safe to read TCNT2, in case we just woke up */
+    DEBUG("RTT sleeps until safe to read TCNT2\n");
+    TCCR2A = 0;
+    _asynch_wait();
```
Josar: _async_wait() is implemented in RTT and RTC, can this not be reduced to one?

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,218 @@
...
+
+    /* Prepare the counter for sub 8-second precision */
+    OCR2B = ((uint8_t)rtc_state.alarm & 0x07) << 5;
+
+    /* Interrupt safe order of assignment */
+    rtc_state.alarm_arg = arg;
+    rtc_state.alarm_cb = cb;
+
+    /* Enable alarm only if it will trigger in < 8 seconds */
+    if ((rtc_state.alarm & 0xFFFFFFF8) <= (rtc_state.time & 0xFFFFFFF8)) {
```
Josar: The comment could state
```
/* Enable Interrupt only if alarm is in this 8s counter period before it overflows*/
```

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    return 0;
+}
+
+int rtc_get_alarm(struct tm *time)
+{
+    /* Convert from seconds since the epoch */
+    /* Note: Cast tells the compiler to discard volatile */
+    gmtime_r((time_t *)&rtc_state.alarm, time);
```
Josar: Would this also work and keeping the reason for volatile
```    
    time_t alarm = rtc_state.alarm;
    gmtime_r(&alarm, time);
```

I think it is not good to discard the volatile.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,218 @@
...
+void rtc_poweroff(void)
+{
+    rtt_poweroff();
+}
+
+void atmega_rtc_incr(void)
+{
+    rtc_state.time += 8;
+
+    /* Enable alarm only if it will trigger in < 8 seconds */
```
Josar: see comment above

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,218 @@
...
+ISR(TIMER2_COMPB_vect) {
+    __enter_isr();
+    /* Disable alarm interrupt */
+    TIMSK2 &= ~(1 << OCIE2B);
+
+    if (rtc_state.alarm_cb != NULL) {
+        rtc_state.alarm_cb(rtc_state.alarm_arg);
+    }
+
+    if (sched_context_switch_request) {
```
Josar: Just keeping in mind that this has to be changed when rebasing with https://github.com/RIOT-OS/RIOT/pull/8988 merged

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,259 @@
...
+
+    /* Virtual 24-bit timer overflowed */
+    if (rtt_state.ext_cnt == 0) {
+        /* Execute callback */
+        if (rtt_state.overflow_cb != NULL) {
+            rtt_state.overflow_cb(rtt_state.overflow_arg);
+        }
+    }
+
+    if (sched_context_switch_request) {
```
Josar: previous comment

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,259 @@
...
+    __enter_isr();
+    /* Disable alarm until overflowed enough times */
+    TIMSK2 &= ~(1 << OCIE2A);
+
+    /* Execute callback */
+    if (rtt_state.alarm_cb != NULL) {
+        rtt_state.alarm_cb(rtt_state.alarm_arg);
+    }
+
+    if (sched_context_switch_request) {
```
Josar: previous comment

cpu/atmega_common/periph/rtc.c
ZetaR60: I added these to diagnose hang issues with _asynch_wait. I think it would be confusing to combine these particular messages with a time output. Which functions do you think would be particularly helpful to have the current time in? I can add separate debug messages for that.

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,249 @@
...
+    rtt_state.ext_comp = (uint16_t)(alarm >> 8);
+    OCR2A = (uint8_t)alarm;
+
+    /* Interrupt safe order of assignment */
+    rtt_state.alarm_arg = arg;
+    rtt_state.alarm_cb = cb;
+
+    /* Enable alarm interrupt only if it will trigger before overflow */
+    if (rtt_state.ext_comp <= rtt_state.ext_cnt) {
+        TIMSK2 |= (1 << OCIE2A);
```
Josar: before enabling clear
        /* Clear interrupt flags */
        TIFR2 = (1 << OCF2A);

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,249 @@
...
+}
+
+ISR(TIMER2_OVF_vect) {
+    __enter_isr();
+
+    rtt_state.ext_cnt++;
+
+    /* Enable RTT alarm if overflowed enough times */
+    if (rtt_state.ext_comp == rtt_state.ext_cnt) {
+        TIMSK2 |= (1 << OCIE2A);
```
Josar: Not clearing will execute the compare match isr imediatly as there is allready a set flag 
/* Clear interrupt flags */
        TIFR2 = (1 << OCF2A);

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,214 @@
...
+
+    return 0;
+}
+
+int rtc_get_alarm(struct tm *time)
+{
+    /* Convert from seconds since the epoch */
+    /* Note: assignment is to discard volatile */
+    time_t alarm = rtc_state.alarm;
+    gmtime_r(alarm, time);
```
Josar: gmtime_r(&alarm, time);

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,214 @@
...
+    rtc_alarm_cb_t alarm_cb;    /* callback called from RTC alarm */
+    void *alarm_arg;            /* argument passed to the callback */
+} rtc_state_t;
+
+static volatile rtc_state_t rtc_state;
+
+void rtc_init(void)
+{
+    /* Only initialize RTT once */
+    if (!(ASSR & (1 << AS2))) {
```
Josar: Doe not work for me. Running the test `periph_rtc` seems this condition is true but timer is not activated.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,214 @@
...
+    _asynch_wait();
+
+    offset = TCNT2;
+
+    /* Convert to seconds (highest 3 bits) */
+    offset = (offset & 0xE0) >> 5;
+
+    /* Convert to seconds since the epoch */
+    rtc_state.time = mk_gmtime(time) - offset;
+
```
Josar: Something like

```
printf("RTC set epoch %" PRIu32 "\n", rtc_state.time );
```

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,214 @@
...
+    time_secs = (time_t)TCNT2;
+
+    /* Convert to seconds (highest 3 bits of TCNT2) */
+    time_secs = (time_secs & 0x000000E0) >> 5;
+
+    time_secs += rtc_state.time;
+
+    /* Convert from seconds since the epoch */
+    gmtime_r(&time_secs, time);
+
```
Josar:
```
DEBUG("RTC get epoch %" PRIu32 "\n", time_secs);
```

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,214 @@
...
+    return 0;
+}
+
+int rtc_set_alarm(struct tm *time, rtc_alarm_cb_t cb, void *arg)
```
Josar: I rewrote this as the curretn implementation did not result in proper behaviour.
The Interrupt was not set at the right point in time. And it was not activated properly when near the overflow.

This resulted in a proper behaviour:

```
int rtc_set_alarm(struct tm *time, rtc_alarm_cb_t cb, void *arg)
{
    /* Disable alarm interrupt */
    TIMSK2 &= ~(1 << OCIE2B);
    rtc_state.alarm_cb = NULL;

    /* Wait until not busy anymore (should be immediate) */
    DEBUG("RTC sleeps until safe to write OCR2B\n");
    _asynch_wait();

    /* Set alarm time */
    rtc_state.alarm = mk_gmtime(time);

    if (rtc_state.alarm < (rtc_state.time + (TCNT2 >> 5))) {
        DEBUG("RTC set alarm allready passed.\n");
        /* Shoot alarm if allready passed.*/
        rtc_state.alarm_cb(rtc_state.alarm_arg);
        return 0;
    }

    /* Calculate time to wait, based on TCNT2=0 */
    uint32_t diff = rtc_state.alarm - rtc_state.time;
    /* Prepare the counter for sub 8-second precision */
    OCR2B = ((uint8_t)diff & 0x07) << 5;

    DEBUG("RTC alarm epoch %" PRIu32 " OCR2B %" PRIu8 " TCNT2 %" PRIu8 "\n",
          rtc_state.alarm, OCR2B, TCNT2 );

    /* Interrupt safe order of assignment */
    rtc_state.alarm_arg = arg;
    rtc_state.alarm_cb = cb;

    /* Enable irq only if alarm is in the 8s period before it overflows */
    if (diff < 8) {
        /* Clear interrupt flag */
        TIFR2 = (1 << OCF2B);
        TIMSK2 |= (1 << OCIE2B);
        DEBUG("RTC interrupt activated\n");
    }
    else {
        DEBUG("RTC interrupt NOT activated\n");
    }

    return 0;
}
```

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,214 @@
...
+{
+    rtt_poweroff();
+}
+
+void atmega_rtc_incr(void)
+{
+    rtc_state.time += 8;
+
+    /* Enable irq only if alarm is in the 8s period before it overflows */
+    if ((rtc_state.alarm & 0xFFFFFFF8) == (rtc_state.time & 0xFFFFFFF8)) {
```
Josar:
```
    /* Only if there is a alarm */
    if (rtc_state.alarm !=0) {
    /* Enable irq only if alarm is in the 8s period before it overflows */
        if ((rtc_state.alarm - rtc_state.time) < 8) {
            if (rtc_state.alarm <= rtc_state.time) {
                DEBUG("RTC incr alarm allready passed.\n");
                /* Prevent alarm offset if time is too soon */
                rtc_state.alarm =0;
                rtc_state.alarm_cb(rtc_state.alarm_arg);
            }
            else {
                DEBUG("RTC incr alarm activated.\n");
                /* Clear interrupt flag */
            }
        }
    }
```

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,214 @@
...
+                   | (1 << TCR2AUB) | (1 << TCR2BUB))) ;
+}
+
+ISR(TIMER2_COMPB_vect) {
+    __enter_isr();
+    /* Disable alarm interrupt */
+    TIMSK2 &= ~(1 << OCIE2B);
+
+    if (rtc_state.alarm_cb != NULL) {
+        rtc_state.alarm_cb(rtc_state.alarm_arg);
```
Josar:
```
        rtc_alarm_cb_t cb = rtc_state.alarm_cb;
        rtc_state.alarm_cb = NULL;
        cb(rtc_state.alarm_arg);
```

cpu/atmega_common/periph/rtt.c
ZetaR60: Fixed.

cpu/atmega_common/periph/rtc.c
ZetaR60: Fixed.

cpu/atmega_common/periph/rtc.c
ZetaR60: Strange. I have no explanation for this. I have removed the check. It shouldn't actually be harmful to run rtt_init twice, except that it resets a bunch of settings.

cpu/atmega_common/periph/rtc.c
ZetaR60: Is it expected for an alarm to clear itself like this? It shouldn't trigger again, unless you have a year 2038 problem.

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,249 @@
...
+    rtt_state.alarm_cb = NULL;
+
+    /* Wait until not busy anymore (should be immediate) */
+    DEBUG("RTT sleeps until safe to write OCR2A\n");
+    _asynch_wait();
+
+    /* Set the alarm value */
+    rtt_state.ext_comp = (uint16_t)(alarm >> 8);
+    OCR2A = (uint8_t)alarm;
+
```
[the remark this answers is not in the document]
ZetaR60: Added debug messages.

cpu/atmega_common/periph/rtc.c
ZetaR60: Nevermind; I see why you did this.

cpu/atmega_common/periph/rtc.c
ZetaR60: This part has been updated.

cpu/atmega_common/periph/rtc.c
ZetaR60: This part has been updated.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,241 @@
...
+
+    /* If alarm not set, nothing else to do */
+    if (rtc_state.alarm_cb == NULL) {
+        return;
+    }
+
+    /* Check to see if alarm was missed */
+    if (rtc_state.alarm <= rtc_state.time) {
+        rtc_state.alarm_cb(rtc_state.alarm_arg);
+        rtc_clear_alarm();
```
Josar: Do not clear after the callback, it might have set a new alarm.

ZetaR60: Fixed.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,247 @@
...
+              PRIu32 "\n", now, rtc_state.alarm);
+        cb(arg);
+        return 0;
+    }
+
+    /* Calculate time to wait, based on TCNT2 = 0 */
+    diff = rtc_state.alarm - rtc_state.time;
+
+    /* Prepare the counter for sub 8-second precision */
+    OCR2B = (uint8_t)(rtc_state.alarm << 5);
```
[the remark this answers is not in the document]
ZetaR60: Fixed. Mask is redundant.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,247 @@
...
+
+ISR(TIMER2_COMPB_vect) {
+    __enter_isr();
+    /* Disable alarm interrupt */
+    TIMSK2 &= ~(1 << OCIE2B);
+
+    /* Execute callback */
+    if (rtc_state.alarm_cb != NULL) {
+        /* Clear callback */
+        rtc_alarm_cb_t cb = rtc_state.alarm_cb;
```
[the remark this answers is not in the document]
ZetaR60: Sorry; thought I compile tested before pushing. Fixed.

shr70:
Can confirm everything running properly on the jiminy-mega256rfr2 now

ZetaR60:
#9405 has been resolved. Support for RTT/RTC on ATmega can now move forward!

ZetaR60:
Rebased due to age.

ZetaR60:
@kaspar030 @gebart @kYc0o @Josar Could you please update your reviews when you get a chance? I can squash to provide a clean list of commits if you prefer.

I have retested on ```mega-xplained``` with new docker image. ```tests/periph_rtt``` and ```tests/periph_rtc``` work as expected.

kYc0o:
Everything OK for me here. However I'd like to wait for @gebart and @kaspar030 reviews.

aabadie:
@gebart, @kaspar030, it seems like your comments were addressed here. Ok to squash ?

PeterKietzmann:
@kaspar030, @gebart please remove your change requests! @maribu interested in this PR?

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,288 @@
...
+    /* Enable RTT alarm if overflowed enough times */
+    if (rtt_state.ext_comp <= rtt_state.ext_cnt) {
+        /* Clear interrupt flag */
+        TIFR2 = (1 << OCF2A);
+
+        /* Enable interrupt */
+        TIMSK2 |= (1 << OCIE2A);
+    }
+
+#if MODULE_PERIPH_RTC
```
maribu: While this seems to work, I would prefer `#ifdef MODULE_PERIPH_RTC`. This would be consistent to the negation of the macro, as

``` C
#ifndef MODULE_PERIPH_RTC
```
will work but
```C
#if !MODULE_PERIPH_RTC
```
will not work. (See this as a suggestion, I will not block the PR on this.)

maribu:
Oh, I think I did cause the merge conflict here. Now all board periph configuration code resides in `boards/common/atmega/include/periph_conf_atmega_common.h` and the individual `periph_conf.h` just include this, while potentially overriding stuff of the common periph conf. I think this approach can be used for this PR as well.

I think the cleanest way would be to check in `periph_conf_atmega_common.h` if the `RTC_NUMOF` is not yet defined. If not defined, then define it to be `(1U)` and add the RTC config. And in the common `periph_conf.h` of the Arduino boards just add `#define RTC_NUMF (0U)` to disable the RTC for them.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,252 @@
...
+static inline void _asynch_wait(void);
+
+typedef struct {
+    time_t time;                /* seconds since the epoch */
+    time_t alarm;               /* alarm_cb when time == alarm */
+    rtc_alarm_cb_t alarm_cb;    /* callback called from RTC alarm */
+    void *alarm_arg;            /* argument passed to the callback */
+} rtc_state_t;
+
+static volatile rtc_state_t rtc_state;
```
maribu: `volatile` is a beast few people fully understand. (And I admit that I do not fully understand it either ;-).) Are you sure that it is right to be used here? If so, please add comments explaining why it is needed. Otherwise, please get rid of it :-)

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,252 @@
...
+    if ((rtc_state.alarm - rtc_state.time) < 8) {
+        /* Clear interrupt flag */
+        TIFR2 = (1 << OCF2B);
+
+        /* Enable interrupt */
+        TIMSK2 |= (1 << OCIE2B);
+    }
+}
+
+void _asynch_wait(void)
```
maribu: `static`

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,288 @@
...
+typedef struct {
+    uint16_t ext_cnt;           /* Counter to make 8-bit timer 24-bit */
+    uint16_t ext_comp;          /* Extend compare to 24-bits */
+    rtt_cb_t alarm_cb;          /* callback called from RTT alarm */
+    void *alarm_arg;            /* argument passed to the callback */
+    rtt_cb_t overflow_cb;       /* callback called when RTT overflows */
+    void *overflow_arg;         /* argument passed to the callback */
+} rtt_state_t;
+
+static volatile rtt_state_t rtt_state;
```
maribu: Is `volatile` correct here? If so, please add comments explaining it.

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,288 @@
...
+{
+    power_timer2_enable();
+}
+
+void rtt_poweroff(void)
+{
+    power_timer2_disable();
+}
+
+void _asynch_wait(void)
```
maribu: `static`

ZetaR60:
Volatile is actually pretty simple: the compiler may not optimize away volatile memory operations, and may not reorder volatile operations with respect to other volatile operations.

In this case, volatile is being used to prevent the compiler from reordering operations that access the state structs with respect to operations that change the timer interrupt mask. This allows for writing interrupt safe code without globally disabling interrupts. If volatile is removed then the code would no longer be interrupt safe as written. I have added a comment to clarify this.

cpu/atmega_common/periph/rtc.c
ZetaR60: Comment added. Also, see post below.

cpu/atmega_common/periph/rtc.c
ZetaR60: Declaration with static is provided near the beginning of the file.

cpu/atmega_common/periph/rtt.c
ZetaR60: Comment added. Also, see post below.

cpu/atmega_common/periph/rtt.c
ZetaR60: Declaration with static is provided near the beginning of the file.

maribu:
@ZetaR60: I think `volatile` is not used correctly here. I'll think about it tomorrow with a fresh mind to confirm this. Btw: The first answer to this question is a good read regarding use of `volatile`: https://stackoverflow.com/questions/4557979/when-to-use-volatile-with-multi-threading

@keestux seems to be an expert on `volatile`, maybe he can have also a look

maribu:
@ZetaR60: I'm sure that `volatile` is incorrect here. You should either use [C11 atomics](https://en.cppreference.com/w/c/atomic) (good), [Atomic Buildtins](https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Atomic-Builtins.html) (bad), or inline assembly (ugly) to guarantee atomic access to the variables. Or just disable (the relevant) interrupts during access.

ZetaR60:
@maribu I can see why people are confused about volatile. That post you linked to is highly misleading, at least when applied to C (not sure how C++ is different, if at all). Rather than look at random vague advice about purpose, intent, and domains of usefulness, we need to look at actual mechanism and specification.

The GCC documentation says specifically that volatile prevents reordering with other volatile variables:

> The minimum requirement is that at a sequence point all previous accesses to volatile objects have stabilized and no subsequent accesses have occurred. Thus an implementation is free to reorder and combine volatile accesses that occur between sequence points, but cannot do so for accesses across a sequence point.

That is from [https://gcc.gnu.org/onlinedocs/gcc/Volatiles.html](url)

My assertion is also supported by the C99 standard document, ISO/IEC 9899:1999

> An  object  that  has  volatile-qualified  type  may  be  modified  in  ways  unknown  to  the
> implementation  or  have  other  unknown  side  effects.   Therefore any expression referring
> to such an object shall be evaluated strictly according to the rules of the abstract machine,
> as described in 5.1.2.3.  Furthermore, at every sequence point the value last stored in the
> object shall agree with that prescribed by the abstract machine, except as modified by the
> unknown  factors  mentioned  previously.

That is from §6.7.3 or p109. [http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf](url)

A sequence point includes variable assignments. In other words, at every variable assignment the actual value assigned to a volatile must agree with the abstract definition of what the programmer wrote, i.e. the compiler cannot reorder variable assignments. This is exactly how I am using volatile, and it is exactly how it is intended to be used. In this case, the goal to be accomplished is to prevent accesses to the struct from being moved out from between writes to the timer interrupt mask.

```
    /* Disable alarm interrupt */
    TIMSK2 &= ~(1 << OCIE2A);

    [some rtt_state accesses]

    /* Enable interrupt */
    TIMSK2 |= (1 << OCIE2A);
```

If those rtt_state accesses are not marked volatile, then the compiler is free to optimize them by effectively moving them around. There is no assurance that they will not take place before the first write to TIMSK2, or after the second write to TIMSK2. By marking rtt_state volatile, the compiler is forced to place the accesses between the TIMSK2 writes. The documentation explicitly supports this exact use of volatile.

cpu/atmega_common/periph/rtt.c
```diff
@@ -0,0 +1,293 @@
...
+        DEBUG("RTT alarm interrupt active\n");
+    }
+    else {
+        DEBUG("RTT alarm interrupt not active\n");
+    }
+}
+
+uint32_t rtt_get_alarm(void)
+{
+    return (((uint32_t)rtt_state.ext_comp << 8) | (uint32_t)OCR2A);
```
maribu: Somehow my inline comments vanished.

Here you are accessing `rtt_state_ext_comp`, which is 16 bit of RAM. On an 8-bit CPU this access will be two instructions. Afterwards you'll access `OCR2A`, which will add more instructions. If between those instructions an interrupt kicks in and a different thread gets scheduled `rtt_state.ext_comp` may be changed. As a result the two 8 bit chunks of `rtt_state.ext_comp` accessed might belong to different states of `rtt_state.ext_comp` and might combined have a result that was neither correct before nor after the interrupt. This is racy and `volatile` is not helping here. Also C11 atomics will not help (but at least fail on compilation), as there is no way to read both `rtt_sate.ext_comp` and `OCR2A` in one instant.

maribu:
@ZetaR60  Keep also in byte that the hardware-mapped memory like `TIMSK2` is indeed (and correctly) `volatile`:

``` C
#define _MMIO_BYTE(mem_addr) (*(volatile uint8_t *)(mem_addr))
#define _SFR_MEM8(mem_addr) _MMIO_BYTE(mem_addr)
 #define TIMSK2 _SFR_MEM8(0x70)
```

So the access to `TIMSK2` will be "a fence" regarding the reordering of sequence points. There really is nothing to gain from declaring `rtt_state` to be `volatile`.

ZetaR60:
@maribu 
> So the access to `TIMSK2` will be "a fence" regarding the reordering of sequence points. There really is nothing to gain from declaring `rtt_state` to be `volatile`.

This is _incorrect_, volatile _does not_ provide any assurance that things will not be reordered _with respect to non-volatile access_. In order for `rtt_state` to be prevented from being reordered with respect to `TIMSK2` _they must both be volatile_. Please read the documentation I referred you to more carefully; the GCC one _specifically_ states this on the paragraph immediately after the one that I quoted from:

> Accesses to non-volatile objects are not ordered with respect to volatile accesses. You cannot use a volatile object as a memory barrier to order a sequence of writes to non-volatile memory.

Please do not make strong imperative claims of code change necessity when someone has _specifically cited_ exactly passages official documentation without ensuring that you understand the citation that was made.

Now, you have correctly pointed out in your inline comment that I have a few remaining issues with returning possible inconsistent results due to the timer interrupting in the middle of non-atomic operations. I will work on fixing these; however, the fix is to utilize the same approach of disabling the interrupt via a volatile register followed by access to the volatile state struct.

You may have an incorrect perception that my code is intended to provide general multi-thread access. Like the other periph systems, it does not. The use of volatile is specifically to prevent the timer events from creating an inconsistent state by disabling those events before accessing `rtt_state` / `rtc_state`. If an application requires multi-thread access to the API, it must globally disable interrupts like with every other periph system. It is only interrupt safe from events internal to the RTT/RTC system.

cpu/atmega_common/periph/rtt.c
ZetaR60: Out of scope for the reason why volatile is employed here. Consistency is only guaranteed for multiple threads if they are only reading. Multi-thread read/write consistency is not guaranteed unless global interrupts are disabled before making the API calls. However, multi-thread writes are in general undesirable for this API as you cannot e.g. set an alarm for each thread anyway.

Your critique applies to e.g. `rtt_get_counter` with respect to timer events, so I will work on fixing that.

maribu:
> An object that has volatile-qualified type may be modified in ways unknown to the implementation or have other unknown side effects.

In other words: Hardware that is somehow mapped into the memory has to be volatile. The state structures you're declaring resides in ordinary memory: It is not changed by the hardware, nor does changing it have any side effects like changing the voltage level at a GPIO pin. This is not a use case for `volatile`.

> Furthermore, at every sequence point the value last stored in the object shall agree with that prescribed by the abstract machine, except as modified by the unknown factors mentioned previously.

``` C
    TIMSK2 &= ~(1 << OCIE2A); /* <-- volatile access */

    [some rtt_state accesses] /* <-- sequence points */

    TIMSK2 |= (1 << OCIE2A); /* <-- volatile access */
```

This means specifically, that during the access to the `rtt_state`, which are sequence points, "the value last stored in the object [here `TIMSK2`] shall agree with that prescribed by the abstract machine". This means during those accesses, the volatile object `TIMSK2` must have the value "interrupts disabled".

---

But, as seeing is believing:
``` C
#include <stdint.h>
typedef union {
    uint32_t u32;
    uint8_t u8[4];
} foo_t;

extern foo_t a, b;
extern
#ifdef USE_VOLATILE
volatile
#endif
int interrupts_enabled;

void bar(void) {
    interrupts_enabled = 0;
    a.u8[0] = 0xa0;
    a.u8[1] = 0xa1;
    b.u8[0] = 0xb0;
    a.u8[2] = 0xa2;
    a.u8[3] = 0xa3;
    interrupts_enabled = 1;
    b.u8[1] = 0xb1;
    b.u8[2] = 0xb2;
    b.u8[3] = 0xb3;
}
```

(Partial) output of `gcc -O3 -o foo1.s -S foo.c`:
``` .s
bar:
.LFB0:
	.cfi_startproc
	movl	$-1549622880, a(%rip)
	movl	$1, interrupts_enabled(%rip)
	movl	$-1280134736, b(%rip)
	ret
	.cfi_endproc
```

Here the accesses get reordered and combined to a single 32-bit write. (And one access to `interrrupts_enabled` is optimized out, as it was deemed unneeded when not taking side effects into account.)

(Partial) output of `gcc -O3 -o foo2.s -S -DUSE_VOLATILE foo.c`:
``` .s
bar:
.LFB0:
	.cfi_startproc
	movl	$-19534, %eax
	movl	$0, interrupts_enabled(%rip)
	movb	$-80, b(%rip)
	movl	$-1549622880, a(%rip)
	movl	$1, interrupts_enabled(%rip)
	movb	$-79, 1+b(%rip)
	movw	%ax, 2+b(%rip)
	ret
	.cfi_endproc
```

Here the accesses to `a` get still optimized and reordered, as they all have been between the two accesses two `interrupts_enabled`. Those accesses to `b` that are supposed to happen while interrupts are disabled are indeed happening then, and those supposed to happen when interrupts are enabled again are happening there.

ZetaR60:
> > An object that has volatile-qualified type may be modified in ways unknown to the implementation or have other unknown side effects.
> 
> In other words: Hardware that is somehow mapped into the memory has to be volatile. The state structures you're declaring resides in ordinary memory: It is not changed by the hardware, nor does changing it have any side effects like changing the voltage level at a GPIO pin. This is not a use case for `volatile`.

You are drawing conclusions that do not follow from the narrow claim. Running an ISR is not a part of the original process state machine and is therefore "unknown" to it. This is exactly what volatile is intended for, and exactly what I am using it for.

> > Furthermore, at every sequence point the value last stored in the object shall agree with that prescribed by the abstract machine, except as modified by the unknown factors mentioned previously.
> 
> ```c
>     TIMSK2 &= ~(1 << OCIE2A); /* <-- volatile access */
> 
>     [some rtt_state accesses] /* <-- sequence points */
> 
>     TIMSK2 |= (1 << OCIE2A); /* <-- volatile access */
> ```
> 
> This means specifically, that during the access to the `rtt_state`, which are sequence points, "the value last stored in the object [here `TIMSK2`] shall agree with that prescribed by the abstract machine". This means during those accesses, the volatile object `TIMSK2` must have the value "interrupts disabled".

This is a section specifically on the nature of volatile objects. I have added some implied language to the quote in question:

> Furthermore, at every sequence point the value _of the volatile object_ last stored in the _volatile_ object shall agree with that prescribed by the abstract machine, except as modified by the unknown factors mentioned previously.

This does not guarantee the state of `rtt_state` during an expression with `TIMSK2`, unless that expression also refers to `rtt_state`.

> But, as seeing is believing:
> [snipped example code]
> 
> Here the accesses to `a` get still optimized and reordered, as they all have been between the two accesses two `interrupts_enabled`. Those accesses to `b` that are supposed to happen while interrupts are disabled are indeed happening then, and those supposed to happen when interrupts are enabled again are happening there.

Okay, so in that case GCC is evaluating the code in the manner you describe, but this is not guaranteed by the specification. The specification makes it clear that this is implementation defined and cannot be relied on.

§5.1.2.3 (which is referred to by the previous section I mentioned):
¶8
> EXAMPLE 1    An implementation might define a one-to-one correspondence between abstract and actual semantics: at every sequence point, the values of the actual objects would agree with those specified by the abstract semantics.  The keyword volatile would then be redundant.

¶9
> Alternatively,  an  implementation  might  perform  various  optimizations  within  each  translation  unit,  such that the actual semantics would agree with the abstract semantics only when making function calls across translation  unit  boundaries.   In  such  an  implementation,  at  the  time  of  each  function  entry  and  function return where the calling function and the called function are in different translation units, the values of all externally  linked  objects  and  of  all  objects  accessible  via  pointers  therein  would  agree  with  the  abstract semantics.  Furthermore, at the time of each such function entry the values of the parameters of the called function and of all objects accessible via pointers therein would agree with the abstract semantics.  In this type of implementation, _objects referred to by interrupt service routines_ activated by the signal function would  _require  explicit  specification  of volatile storage_,  as  well  as  other  implementation-defined restrictions.

Italics mine on paragraph 9.

ZetaR60:
I have made some alterations: volatile is reduced to what is necessary for proper integrity (`rtt_state.ext_cnt` and `rtt_state.time` only now). A new inline function `_safe_cnt_get()` uses volatile to safely get by re-checking. And, I brought in `irq_disable()` and `irq_restore()` to replace the requirement for a specific callback assignment order. I will work on testing this code tomorrow.

@maribu I re-read my earlier comments, and I feel that at least one comes off as being quite abrasive.  That was not my intention, and I apologize if you found it unpleasant. It is difficult to convey that sort of non-literal meaning in a purely textual medium. I do appreciate that you are taking time to review my code.

maribu:
@ZetaR60: If my interpretion of the standard is indeed wrong, you should go for [compiler memory barriers](https://en.m.wikipedia.org/wiki/Memory_ordering#Compiler_memory_barrier) instead of `volatile`, as those have less impact on performance.

```C
static inline void disable_timer_int(void) {
    asm volatile("" ::: "memory");
    TIMSK2 &= ~(1 << OCIE2A);
    asm volatile("" ::: "memory");
```

and the same for reenable. That will do exactly what is intended and not disable any more optimisation than needed.

maribu:
@ZetaR60: I did some more research regarding moving memory access to globals around accesses to volatiles:

[This write up by Prof John Regehr](https://blog.regehr.org/archives/28a) states that there two fractions:

> One school of thought is that compilers may not move accesses to global variables around accesses to volatile variables.  There seems to be a consistent reading of the standard that backs this up.

But he states most compilers (at least back in 2010) share exactly your view on that, so that they do move access to non-`volatile` globals around accesses to volatiles. So your claim that additional measures need to be taken to prevent that reordering is 100% correct. (And obviously it doesn't even matter whether or not my reading of the standard is correct, as in the end we will have to work with real world compiler and how they do things.)

I'd still suggest to use the compiler memory barrier instead of `volatile` to prevent the reordering. This will prevent optimisation just as much as needed, and also the intent of the code will be more readable.

Memo to my self: A PR adding memory barriers e.g. via `static inline void memory_barrier(void) { asm....}` would be nice.

ZetaR60:
> @ZetaR60: If my interpretion of the standard is indeed wrong, you should go for [compiler memory barriers](https://en.m.wikipedia.org/wiki/Memory_ordering#Compiler_memory_barrier) instead of `volatile`, as those have less impact on performance.
> 
> ```c
> static inline void disable_timer_int(void) {
>     asm volatile("" ::: "memory");
>     TIMSK2 &= ~(1 << OCIE2A);
>     asm volatile("" ::: "memory");
> ```
> 
> and the same for reenable. That will do exactly what is intended and not disable any more optimisation than needed.

No, I feel that this is bad advice and it won't do what you intend with respect to optimization.
1) This use of volatile here does exactly what is needed, with very little possible additional optimization.
2) Memory barriers should not be used as a replacement for volatile. Volatile prevents only reordering with other volatile accesses, but memory barriers prevent _all_ forms of optimization from passing the memory barrier.
3) This is a GCC specific idiom. While using something like this is sometimes okay, a standard construction (like volatile) should be strongly preferred.
4) `_safe_cnt_get()` would require half a dozen memory barriers, compared to a single volatile.


> @ZetaR60: I did some more research regarding moving memory access to globals around accesses to volatiles:
> 
> [This write up by Prof John Regehr](https://blog.regehr.org/archives/28a) states that there two fractions:

This is a broken link for me.


> > One school of thought is that compilers may not move accesses to global variables around accesses to volatile variables.  There seems to be a consistent reading of the standard that backs this up.
> 
> But he states most compilers (at least back in 2010) share exactly your view on that, so that they do move access to non-`volatile` globals around accesses to volatiles. So your claim that additional measures need to be taken to prevent that reordering is 100% correct. (And obviously it doesn't even matter whether or not my reading of the standard is correct, as in the end we will have to work with real world compiler and how they do things.)

A bit off topic then, but according to the standard doc, writes to global variables in the translation environment may be reordered, while writes to globals in the execution environment may not. The translation environment is basically everything from the source file that gets compiled (with includes). The execution environment is basically everything outside of that (i.e. the stuff calling the code). So, essentially, if you reference a global with `extern` it gets protection from reordering. This also applies to pointers that point to memory in the execution environment.


> I'd still suggest to use the compiler memory barrier instead of `volatile` to prevent the reordering. This will prevent optimisation just as much as needed, and also the intent of the code will be more readable.
> 
> Memo to my self: A PR adding memory barriers e.g. via `static inline void memory_barrier(void) { asm....}` would be nice.

Please don't go sprinkling things like this around. I do not mean to offend, but I do not feel that you currently have a deep enough understanding of how volatile and memory barriers work to adjudicate their useage. Adding new memory barriers improperly to working code has a high chance of degrading it, and a low chance of making it safer.

maribu:
@ZetaR60: I happily admit that my knowledge about the technical details of `volatile` are limited. But I am aware of the consensus among C developers regarding the usage of `volatile` and we should stick to that, unless we have a very, very good reason to not do so.

`volatile` is used by C developers for three things:

1. To force the compiler to actually perform the reads and writes to the variable that is hardware mapped.
2. To get atomic accesses
3. To enforce ordering of memory access

The conses is that 1. is totally valid, as this is exactly what `volatile` was introduced for historically. 2. does not work, as this has nothing to do with atomic accesses. This use of `volatile` is simply a bug. And the conses is to not use it for 3. for various reasons, which are:

- The degree by which ordering is guaranteed when using `volatile` accesses depends on who you ask. This is a very bad feature if compilers give different guarantees on that
- It is totally useless on systems that have out of order execution or multiple processor (cores). This is not yet relevant for RIOT, but it seems to be a matter of time until out of order processors hit the embedded market. Also RIOT is already running on the ESP32 (using only one core), so RIOT might gain support for running on more than one processor (core) soon.

The Linux kernel guys are explicitly forbidding the use of `volatile` variables except for memory directly linked to hardware (and some “stupid legacy” code)  https://www.kernel.org/doc/html/latest/process/volatile-considered-harmful.html Unlike me (as you pointed out) , those guys have the required expertise to draw the conclusion that `volatile` is considered harmful.

> Memory barriers should not be used as a replacement for volatile. Volatile prevents only reordering with other volatile accesses, but memory barriers prevent all forms of optimization from passing the memory barrier.

This is not true. `volatile` prevents the reordering by the compiler, which is does work on in-order single-core single-threaded CPUs. Memory barrier work on every system.

Summing up: You are using `volatile` for something it is historically not intended to be used. The consensus among C programmer is to not do what you are doing. Please provide reasonable and strong technical justification for that, or stop using `volatile` there.

Some links to read:
- [Arch D. Robison - Volatile: Almost Useless for Multi-Threaded Programming](https://web.archive.org/web/20181124154026/https://software.intel.com/en-us/blogs/2007/11/30/volatile-almost-useless-for-multi-threaded-programming)
- [John Regehr - Nine ways to break your systems code using volatile](https://blog.regehr.org/archives/28)
- [Jonathan Corbet et al. - Why the “volatile” type class should not be used](https://www.kernel.org/doc/html/latest/process/volatile-considered-harmful.html)

maribu:
Maybe @keestux or @jcarrano cares to join the discussion?

ZetaR60:
> @ZetaR60: I happily admit that my knowledge about the technical details of `volatile` are limited. But I am aware of the consensus among C developers regarding the usage of `volatile` and we should stick to that, unless we have a very, very good reason to not do so.

You are making two _faulty generalizations_ here. The first is that because _most_ developers aren't aware of the nuances of the use of volatile that _I_ am not aware of the nuances of its use. If your question was "should a developer who is not aware of the nuances of volatile use it", I would answer _no_, which is the consensus among C developers. The other faulty generalization is that since _in typical environments_ there are better tools than volatile, that _in all environments_ there are better tools than volatile. Your Corbet and Robinson citations apply specifically to those typical environments, and your Regehr citation _explicitly agrees_ with my assessment:

> If you are writing code for an in-order embedded processor and have little or no infrastructure besides the C compiler, you may need to lean more heavily on volatile.


Furthermore, you cannot adjudicate the nuances of a consensus regarding a particular construction if _you do not understand that construction_. The most that should be done in this circumstance is determine whether the developer in question _has a detailed and nuanced understanding_ of the construction, and has significant reasons for the construction's use based on that nuanced understanding. If that determination has been made, as it has been _adequately demonstrated_ in this circumstance, strict imperatives involving its use _should stop there_. That imperative process should only continue if the one adjudicating has _specific and nuanced_ reasons based in a good understanding that preferably could be backed up by _primary sources_ if disagreement continues. You have freely admitted that you currently lack that nuanced understanding, and I do not fault you for that. I, however, feel that falling back on tertiary sources, i.e. _opinion pieces_, without a nuanced understanding of the underlying reasons for those opinions is not proper in this circumstance. Also, I am dismayed by the zealousness with which you are persuing this issue.


> The degree by which ordering is guaranteed when using `volatile` accesses depends on who you ask. This is a very bad feature if compilers give different guarantees on that

It _does not depend_. It is _explicity guaranteed_ by the specification, which I have demonstrated. Those who are claiming that that it does not need to read the specification! If a compiler does not give the same guarantee as the specification then _that compiler is broken_.


> It is totally useless on systems that have out of order execution or multiple processor (cores). This is not yet relevant for RIOT, but it seems to be a matter of time until out of order processors hit the embedded market. Also RIOT is already running on the ESP32 (using only one core), so RIOT might gain support for running on more than one processor (core) soon.

I do not have experience on out of order execution and cannot provide rigorous comments on it, except that it is not germaine to the current circumstance. The sentiment that volatile is "totally useless" on multi-processor is _generally true_. However, like many principles in programming, it is _improper_ to universalize prohibitions, and to try and adjudicate others that have a specific exception informed by a nuanced understanding when you do not have a nuanced understanding yourself. Also, opinions on the general state of programming will be _inapplicable_ to MCU systems without knowledge on the specific construction in the context of those differences.


> The Linux kernel guys are explicitly forbidding the use of `volatile` variables except for memory directly linked to hardware (and some “stupid legacy” code) https://www.kernel.org/doc/html/latest/process/volatile-considered-harmful.html Unlike me (as you pointed out) , those guys have the required expertise to draw the conclusion that `volatile` is considered harmful.
> 
> > Memory barriers should not be used as a replacement for volatile. Volatile prevents only reordering with other volatile accesses, but memory barriers prevent all forms of optimization from passing the memory barrier.
> 
> This is not true. `volatile` prevents the reordering by the compiler, which is does work on in-order single-core single-threaded CPUs. Memory barrier work on every system.

Their claims apply specifically to systems that have more services available to accomplish tasks that require volatile. This does _not apply_ in this circumstance because the necessary servies _are not available_ in ISR mode, and making ISRs more complex is _highly undesirable_ in RIOT. Also, memory barriers are not the correct construct to apply here, because preventing reordering with other volatile accesses and enforcing repeated reads/writes is _exactly_ what is needed here, and preventing all forms of optimization from passing the memory barrier is _undesirable_.


> Summing up: You are using `volatile` for something it is historically not intended to be used. The consensus among C programmer is to not do what you are doing. Please provide reasonable and strong technical justification for that, or stop using `volatile` there.

My use of volatile is _specifically and explicitly_ given as an example in _the C99 standard document_ as a _correct and necessary_ use of volatile! You have not provided any evidence that what I am doing is not according to consensus of C programmers with a nuanced understanding of volatile because your _tertiary_ sources apply to circumstances that are _not applicable here_, and one of your sources _explicitly states that volatile may be required in this circumstance_. Also, I have _provided resonable and strong technical justification_ in prior posts, and you have apparently _ignored it_!

ZetaR60:
If you want an easier to understand description of what I am using volatile for: I am using `rtt_state.ext_cnt` and `rtc_state.time` _as-if_ they are registers. This is done to extend `TCNT2` to a higher bit size. So, basically, think of them as registers that just also happen to be ordinary memory locations.

ZetaR60:
I have pulled the volatile variables out of the state structs to make it clearer that they are being treated differently, and I have added a note about why those variables are volatile and a warning not to remove volatile.

jcarrano:
The way I see volatile in the context of this PR is that the timer extension variables are effectively "software registers". While there may be a way to make this work without marking them as volatile they behave, from the point of view if code running _outside_ of ISR, as hardware registers so it is quite convenient to treat them as such.

When interrupts are disabled they are regular variables and in that case the volatile is not needed. I don't think it is a big issue. The solution wold be to split the code, the ISR code sees a not volatile declaration and the user-facing code sees volatile (using for example `extern volatile ...`). Again, I would not consider it a big deal or worth the trouble.

Now, @ZetaR60, what is not clear to me is whether the RTT/RTC functions are safe to call from within an interrupt. Reading the code suggests it's not as the overflow would not get handled immediately. If that is correct, it should be documented as a limitation.

ZetaR60:
@jcarrano 
> Now, @ZetaR60, what is not clear to me is whether the RTT/RTC functions are safe to call from within an interrupt. Reading the code suggests it's not as the overflow would not get handled immediately. If that is correct, it should be documented as a limitation.

My latest commits should fix this issue. I _believe_ that all code is now safe from timer events, other ISRs interrupting the current thread, and running the code with interrupts disabled or in ISR mode. It is _not_ safe from concurrent access by multiple threads, except in the case where only state reads are occurring. However, so far as I can tell this is not generally expected of RIOT APIs, and it can be made concurrent access safe by disabling ISRs during access (like everything else).

ZetaR60:
It seems that making the code safe from concurrent access by multiple threads was easier than I thought. I think I have handled all of the various conditions that can occur; this is fairly difficult though, since there is multiple asynchronous considerations at almost every line of code, so I may have missed some things.

@maribu @jcarrano et al: I would like to merge and rebase as the list of changes is getting quite messy and out of date. Please let me know if this is acceptable.

maribu:
> I would like to squash and rebase

I'd say go ahead.

cpu/atmega_common/periph/rtc.c
```diff
@@ -0,0 +1,295 @@
...
+
+    do {
+        /* Operations must occur in this order: */
+        irq_flag = TIFR2 & (1 << TOV2);
+        time_tmp = time_cnt;
+        cnt_tmp = TCNT2;
+
+    /* Detect rollover after time_tmp read. If condition is met, then rollover
+     * just occured, and it is safe to read again now. */
+    } while ((time_tmp != time_cnt) || (irq_flag != (TIFR2 & (1 << TOV2))));
```
Josar: I´m quite sure that `TOV2` will ever be 0. But i might be wrong.

```
Bit 0 – TOV: Timer/Counter2, Overflow Flag
The TOV bit is set (one) when an overflow occurs in Timer/Counter2. TOV is cleared by hardware when
executing the corresponding interrupt handling vector.
```
As the interrupt is enabled `TOV2` will be cleared. Polling it only is usefull when no `IRQ routine` is used, imho. 
Maybe handling like in xtimer would be an idea.

https://github.com/RIOT-OS/RIOT/blob/8755a26716009958148f9b59440d2f6f3f75b0bc/sys/xtimer/xtimer_core.c#L89

ZetaR60: `TOV2` is checked in case interrupts are disabled. These two checks are not really polls, because the condition is only expected to occur zero or one times. It is just more space efficient than using `if`, since the compiler is not allowed to optimize this due to volatile.

benpicco:
Is there a TL;DR for the discussion?
Maybe it would be easier to have separate PRs for RTT and RTC?

A rebase would be in order any way.

benpicco:
When I run this on `arduino-mega2560`, the `tests/periph_rtt` test prints (with `ENABLE_DEBUG=1`)
```
2019-11-05 17:50:30,220 # Initializing the RTT driver
2019-11-05 17:50:30,221 # Initializing RTT
2019-11-05 17:50:30,239 # RTT waits until ASSR not busy
2019-11-05 17:50:31,412 # RTT initialized
2019-11-05 17:50:31,435 # RTT sleeps until safe to read TCNT2
2019-11-05 17:50:33,227 # RTT now: 0
2019-11-05 17:50:33,260 # Setting initial alarm to now + 5 s (160)
2019-11-05 17:50:33,303 # RTC sleeps until safe read TCNT2 and to write OCR2B
2019-11-05 17:50:33,552 # RTT set alarm TCNT2: 0, OCR2A: 160
2019-11-05 17:50:33,553 # RTT alarm interrupt active
2019-11-05 17:50:33,619 # Done setting up the RTT, wait for many Hellos
2019-11-05 17:50:33,638 # RTT sleeps until safe to read TCNT2
2019-11-05 17:50:33,835 # (RTT now: 0)
```

I'm a bit surprised by how long it stays in the `_asynch_wait()` loop.

benpicco:
How do these crystals work?
On my Mega 2560 Pro / `arduino-mega2560` I have two crystals: 12 MHz and 16 MHz and the RTT won't tick.

Mega 2560 Pro

I now have a `rss2-mega256rfr2` with a 32kHz XTAL and RTT works right out of the box with this PR.

herjulf:
@benpicco I think board is missing the 32kHz xtal needed for TIM2. This timer can be asynchronous and work even if we put power hungry timers in sleep mode. As a fallback for RDC is essential. But challenge is to sync the clocks.
The tests/periph_rtc seems not compatible as the update interval is 8s while the alarm is set to 2s in the test program.

roberthartung:
PING everyone, should we get this done?

benpicco:
RTT support has been merged now, should be possible to base RTC support off that if you set `RTT_FREQUENCY` to 32 Hz (which appears to be the minimal frequency)

maribu:
Maybe we could also just provide the RTC interface on top of the RTT? Unless the hardware internationally uses years, hours, minutes and seconds as units, I would consider the RTT interface the better option. And if one really wants an RTC like interface, a generic wrapper on top of RTT would solve this for all clock hardware (except those really internally using the broken up time / date fornat).

benpicco:
> Maybe we could also just provide the RTC interface on top of the RTT?

Something like this?

```C
int rtc_set_time(struct tm *time)
{
    uint32_t t = RTT_SEC_TO_TICKS(mktime(time));
    rtt_set_counter(t);

    return 0;
}

int rtc_get_time(struct tm *time)
{
    time_t t = RTT_TICKS_TO_SEC(rtt_get_counter());
    gmtime_r(&t, time);

    return 0;
}

int rtc_set_alarm(struct tm *time, rtc_alarm_cb_t cb, void *arg)
{
    time_t t = mktime(time);
    rtt_set_alarm(RTT_SEC_TO_TICKS(t), cb, arg);

    return 0;
}

int rtc_get_alarm(struct tm *time)
{
    time_t t = RTT_TICKS_TO_SEC(rtt_get_alarm());
    gmtime_r(&t, time);

    return 0;
}
```

(We can replace the POSIX functions with `rtc_mktime()` and `rtc_localtime()` from #13284 for smaller code and custom range)

kaspar030:
> Something like this?

There's already something like this somewhere in the codebase at least here: `cpu/kinetis/periph/rtc.c`.

benpicco:
The RTT overflows too fast to use it like that (the slowest is 32 Hz and it only has 24 bit), but we can make use of `rtc_tm_normalize()` from `periph_common/rtc.c`.

benpicco:
Both RTT and RTC support have now been merged.

PR #8845, opened 2018-03-28
Makefile.include: add %.hex and %.bin rules

cladmi:
### Contribution description
    
* Replace HEXFILE rule by a rule for each type.
* Add BINFILE variable
* Add 'elffile', 'hexfile' and 'binfile' named targets
* Cleanup of OFLAGS from -Oformat

### Issues/PRs references

Depends on ~~https://github.com/RIOT-OS/RIOT/pull/8844~~ ~~https://github.com/RIOT-OS/RIOT/pull/8846~~
Part of https://github.com/RIOT-OS/RIOT/pull/8838

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -308,7 +308,7 @@ BASELIBS += $(BINDIR)/$(APPLICATION_MODULE).a
 BASELIBS += $(APPDEPS)
 
 .PHONY: all link clean flash flash-only term doc debug debug-server reset objdump help info-modules
-.PHONY: print-size
+.PHONY: print-size elffile binfile hexfile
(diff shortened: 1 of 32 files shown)


cladmi:
When testing locally, I found out that `mips-malta` is broken with this as it was doing an `objcopy` without arguments before. Will fix.

However, I also found that the default goal is not `all` anymore for some boards, it was `link` but now its `elffile` so it does not fail when you do `make BOARD=mips-malta`.

I will do another PR for the second one.

kaspar030:
please squash

kaspar030:
murdock failure is unrelated.

Issue #8434, opened 2018-01-25
Lua support


danpetry:
## Description

There is an external developer who wants to build a modular piece of hardware with a coprocessor running RIOT, with potential to be a commercial product. For this, he wishes to use Lua to control part of it, and therefore Lua support could enable commercial use of RIOT. Additionally, dynamic language support (see also https://github.com/RIOT-OS/RIOT/pull/7796 and https://github.com/RIOT-OS/RIOT/pull/2968) could be an approach to achieve over-the-air update of an application and a multi-application development paradigm for the RAPstore project (RIOT app store).

There is a benefit/cost tradeoff: the above description is the benefit, and time spent to implement Lua support (and eg not develop other features) is the cost. Therefore part of the earlier stages of development will be to judge the total amount of time that would be required.

## Approach

The current approach followed is to adapt other FOSS projects which have implemented Lua in an embedded context. The ones that have been considered are as follows.

1. Currently the most promising vector of approach seems to be to use the [LuaRTOS](https://github.com/whitecatboard/Lua-RTOS-ESP32) project, where the Lua engine has been implemented on top of FreeRTOS, including a range of API bindings we would need to implement such as I2C, ADCs, and LoRAWAN. They have implemented supporting threads which can handle concurrent Lua programs with the same Lua state. The project would probably be pretty stable (as it is developed by a company, Whitecat, and used in their commercial product),  and it is currently [under active development](https://github.com/whitecatboard/Lua-RTOS-ESP32/commits/master). It also uses Google Blockly, a simple graphical programming language which is being considered to develop RIOT programs in Javascript. It uses a BSD license and the build system is based on Make.

2. [NutOS](http://www.egnite.de/en/product-overview/embedded-systems-ethernut-nutos/nutos-open-source-rtos/) also includes a Lua interpreter on top of an RTOS. It also uses a BSD license and a make-based build system. However, the Lua engine is not under current development (with the last substantial work done in 2012) and it seems that there is no implementation of supporting threads, just a callable library of functions.

3. [eLua](http://www.eluaproject.net/), a standalone Lua runtime environment for embedded devices, was also investigated in some detail. However, it was clear that the Lua interpreter was not designed to be separated from the rest of the program (which was designed to be loaded onto a microchip on its own, not used on top of an RTOS). Some of the reasons for this, along with other reasons that suggested that eLua wouldn't be a suitable approach, are as follows:

[...]

PR #8847, opened 2018-03-28
Lua support - initial PR

danpetry:
### Summary

This is an initial PR intended as a first step in development for adding Lua support. Lua is a lightweight scripting language suitable for embedded devices.

### Requirements

The requirements of this PR are to achieve the download and build into RIOT of the Lua interpreter, and demonstrate its working status via a simple example.

#### Package

Lua v5.3.4 is cloned from github and built into RIOT, ready for its API to be accessed. Currently, only the Lua print() function is supported, which is adequate to demonstrate a working state of the interpreter.

#### Example

A .lua file called main.lua containing a simple "hello world" program, residing in the RIOT application's home directory, is converted to a char array at compile time, and stored in a header file called main.lua.h. This header file is then #included into an application, and the converted script is run in the program via a helper function called lua_run_script.

## Issues/PRs references

https://github.com/RIOT-OS/RIOT/issues/8434

### DIFF
--- a/examples/lua/Makefile
+++ b/examples/lua/Makefile
@@ -1,0 +1,65 @@
+APPLICATION = lua
+
+# If no BOARD is found in the environment, use this default:
+BOARD ?= native
+
+# This has to be the absolute path to the RIOT base directory:
+RIOTBASE ?= $(CURDIR)/../..
+
+BOARD_INSUFFICIENT_MEMORY := bluepill calliope-mini cc2650-launchpad \
+	                     cc2650stk maple-mini microbit nrf51dongle \
+			     nucleo-f030r8 nucleo-f031k6 nucleo-f042k6 \
+			     nucleo-f070rb nucleo-f072rb nucleo-f103rb \
+			     nucleo-f302r8 nucleo-f303k8 nucleo-f334r8 \
+			     nucleo-f410rb nucleo-l031k6 nucleo-l053r8 \
+			     opencm904 spark-core stm32f0discovery
+
+BOARD_BLACKLIST := arduino-duemilanove arduino-mega2560 arduino-uno \
+                   chronos jiminy-mega256rfr2 mega-xplained mips-malta \
+		   msb-430 msb-430h pic32-clicker pic32-wifire telosb \
+		   waspmote-pro wsn430-v1_3b wsn430-v1_4 z1
+
+
+# Comment this out to disable code in RIOT that does safety checking
+# which is not needed in a production environment but helps in the
+# development process:
+DEVELHELP ?= 1
+
+# Change this to 0 show compiler invocation lines by default:
+QUIET ?= 1
+
+USEMODULE += ps
+
+ifneq ($(BOARD),native)
+  # This stack size is large enough to run Lua print() functions of
+  # various lengths. Other functions untested.
+  CFLAGS += -DTHREAD_STACKSIZE_MAIN=4096
+endif
+
+USEPKG += lua
+
+include $(RIOTBASE)/Makefile.include
+
+# The code below generates a header file from any .lua scripts in the
+# example directory. The header file contains a byte array of the
+# ASCII characters in the .lua script.
+
+LUA_PATH := $(BINDIR)/lua
+
+# add directory of generated *.lua.h files to include path
+CFLAGS += -I$(LUA_PATH)
+
+# generate .lua.h header files of .lua files
+LUA = $(wildcard *.lua)
+
+LUA_H := $(LUA:%.lua=$(LUA_PATH)/%.lua.h)
+
+$(LUA_PATH)/:
+	@mkdir -p $@
+
+$(LUA_H): | $(LUA_PATH)/
+$(LUA_H): $(LUA_PATH)/%.lua.h: %.lua
+
+	xxd -i $< | sed 's/^unsigned/const/g' > $@
+
+$(RIOTBUILD_CONFIG_HEADER_C): $(LUA_H)
--- a/examples/lua/README.md
+++ b/examples/lua/README.md
@@ -1,0 +1,22 @@
+### About
+
+This example shows how to write IoT applications using Lua.
+
+### Caveats
+
+Currently, the only actual function available is "print"
+
+### How to use
+
+Put your lua code into "main.lua" (check the example). The file will
+be included in your application as an ASCII byte array. This is done in the
+last 20 or so lines of the example's Makefile.
+
+The script will then be run immediately after RIOT has started up.
+
+### How to run
+
+Type `make flash term`
+
+Note: you may have to press `RESET` on the board (after the flash) if the board
+reboots faster than the terminal program can start..
--- a/examples/lua/main.c
+++ b/examples/lua/main.c
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2018 FU Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     examples
+ * @{
+ *
+ * @file
+ * @brief       Basic lua example application
+ *
+ * @author      Daniel Petry <daniel.petry@fu-berlin.de>
+ *
+ * @}
+ */
+
+#include <stdio.h>
+#include <errno.h>
+#include "lauxlib.h"
+#include "lualib.h"
+
+#include "main.lua.h"
+
+int lua_run_script(const char *buffer, size_t buffer_len)
+{
+
+    lua_State *L = luaL_newstate();
+
+    if (L == NULL) {
+        puts("cannot create state: not enough memory");
+        return ENOMEM;
+    }
+
+    luaL_openlibs(L);
+    luaL_loadbuffer(L, buffer, buffer_len, "lua input script");
+
+    if (lua_pcall(L, 0, 0, 0) != LUA_OK){
+        puts("Lua script running failed");
+        return EINTR;
+    }
+
+    lua_close(L);
+    return 0;
+}
+
+int main(void)
+{
+    puts("Lua RIOT build");
+    lua_run_script(main_lua, main_lua_len);
+    return 0;
+}
--- a/pkg/lua/Makefile
+++ b/pkg/lua/Makefile
@@ -1,0 +1,12 @@
+PKG_NAME=lua
+PKG_URL=https://github.com/lua/lua.git
+PKG_VERSION=e354c6355e7f48e087678ec49e340ca0696725b1
+PKG_LICENSE=MIT
+
+.PHONY: all
+
+all:
+	@cp Makefile.lua $(PKG_BUILDDIR)
+	"$(MAKE)" -C $(PKG_BUILDDIR) -f Makefile.lua
+
+include $(RIOTBASE)/pkg/pkg.mk
--- a/pkg/lua/Makefile.lua
+++ b/pkg/lua/Makefile.lua
@@ -1,0 +1,9 @@
+SRC := $(filter-out lua.c luac.c,$(wildcard *.c))
+
+# This builds for native using POSIX system calls and some extra libraries, and
+# removes a compiler warning that warns against using tmpnam().
+ifeq ($(BOARD),native)
+  CFLAGS += -DLUA_USE_LINUX
+endif
+
+include $(RIOTBASE)/Makefile.base
--- a/examples/lua/main.lua
+++ b/examples/lua/main.lua
@@ -1,0 +1,1 @@
+print("Hello world, this is lua!")
--- a/pkg/lua/Makefile.include
+++ b/pkg/lua/Makefile.include
@@ -1,0 +1,1 @@
+INCLUDES += -I$(PKGDIRBASE)/lua
--- a/pkg/lua/doc.txt
+++ b/pkg/lua/doc.txt
@@ -1,0 +1,6 @@
+/**
+ * @defgroup pkg_lua Lua ported to RIOT
+ * @ingroup  pkg
+ * @brief    Provides Lua support for RIOT
+ * @see      https://github.com/lua/lua
+ */
--- a/pkg/lua/patches/0001-Remove-dependency-on-nonexistent-RIOT-syscalls.patch
+++ b/pkg/lua/patches/0001-Remove-dependency-on-nonexistent-RIOT-syscalls.patch
@@ -1,0 +1,62 @@
+From b681cb20ee6bfc31b8ba23ec321140aae6e53e9d Mon Sep 17 00:00:00 2001
+From: danpetry <daniel.petry@fu-berlin.de>
+Date: Wed, 23 May 2018 14:09:17 +0200
+Subject: [PATCH 1/1] Remove dependency on nonexistent RIOT syscalls
+
+Os.rename and os.clock, in Lua, now always return an error message.
+Also, l_randomizePivot, which is used by table.sort, will always
+supply 0 rather than a pseudorandom number.
+---
+ loslib.c  | 7 ++-----
+ ltablib.c | 4 +++-
+ 2 files changed, 5 insertions(+), 6 deletions(-)
+
+diff --git a/loslib.c b/loslib.c
+index dd2bb378..f9e5cbd2 100644
+--- a/loslib.c
++++ b/loslib.c
+@@ -157,9 +157,7 @@ static int os_remove (lua_State *L) {
+ 
+ 
+ static int os_rename (lua_State *L) {
+-  const char *fromname = luaL_checkstring(L, 1);
+-  const char *toname = luaL_checkstring(L, 2);
+-  return luaL_fileresult(L, rename(fromname, toname) == 0, NULL);
++  return luaL_error(L, "This function is not implemented in RIOT yet");
+ }
+ 
+ 
+@@ -181,8 +179,7 @@ static int os_getenv (lua_State *L) {
+ 
+ 
+ static int os_clock (lua_State *L) {
+-  lua_pushnumber(L, ((lua_Number)clock())/(lua_Number)CLOCKS_PER_SEC);
+-  return 1;
++  return luaL_error(L, "This function is not implemented in RIOT yet");
+ }
+ 
+ 
+diff --git a/ltablib.c b/ltablib.c
+index 588bf40d..8895b653 100644
+--- a/ltablib.c
++++ b/ltablib.c
+@@ -235,13 +235,15 @@ static int unpack (lua_State *L) {
+ /* type for array indices */
+ typedef unsigned int IdxT;
+ 
+-
+ /*
+ ** Produce a "random" 'unsigned int' to randomize pivot choice. This
+ ** macro is used only when 'sort' detects a big imbalance in the result
+ ** of a partition. (If you don't want/need this "randomness", ~0 is a
+ ** good choice.)
+ */
++
++#define l_randomizePivot() 0
++
+ #if !defined(l_randomizePivot)		/* { */
+ 
+ #include <time.h>
+-- 
+2.17.0
+


danpetry:
@kaspar030 would you be interested in having a look at this?

examples/lua/lua.c
```diff
@@ -0,0 +1,658 @@
...
+#define LUA_MAXINPUT		512
+#endif
+
+#if !defined(LUA_INIT_VAR)
+#define LUA_INIT_VAR		"LUA_INIT"
+#endif
+
+#define LUA_INITVARVERSION	LUA_INIT_VAR LUA_VERSUFFIX
+
+
```
jia200x: Please remove extra empty line here

examples/lua/lua.c
```diff
@@ -0,0 +1,658 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "lua.h"
+
+#include "lauxlib.h"
+#include "lualib.h"
+
+#if !defined(LUA_PROMPT)
```
jia200x: #ifndef?

examples/lua/lua.c
```diff
@@ -0,0 +1,658 @@
+/*
```
jia200x: As far as I see, a large portion of this file is internal code. This should be in pkg/lua/contrib

examples/lua/main.c
```diff
@@ -0,0 +1,53 @@
...
+int lua_run_main_script(int argc, char **argv)
+{
+    (void)argc;
+    (void)argv;
+
+    return lua_handle_buffer(main_lua, main_lua_len);
+}
+
+static const shell_command_t shell_commands[] = {
+    { "lua", "Start a Lua prompt", lua_main},
```
jia200x: I think it would be good to have a single command here.
E.g `lua console`, `lua run`, etc.

examples/lua/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+### How to use
+
+Put your lua code into "main.lua" (check the example). The file will
+automatically be included when compiling the application. You can then execute the
+file using the lua_run_script command.
+
+### How to run
+
+Type `make flash term
```
jia200x: missing ` after "term"

examples/lua/lua.c
```diff
@@ -0,0 +1,658 @@
...
+#else				/* }{ */
+
+/* ISO C definition */
+#define lua_stdin_is_tty()	1  /* assume stdin is a tty */
+
+#endif				/* } */
+
+#endif				/* } */
+
+
```
jia200x: extra empty line

examples/lua/lua.c
```diff
@@ -0,0 +1,658 @@
...
+#define lua_stdin_is_tty()	_isatty(_fileno(stdin))
+
+#else				/* }{ */
+
+/* ISO C definition */
+#define lua_stdin_is_tty()	1  /* assume stdin is a tty */
+
+#endif				/* } */
+
+#endif				/* } */
```
jia200x: Could you add /* */ to mark the closing bracket? E.g /* lua_stdin_is_tty */

examples/lua/lua.c
```diff
@@ -0,0 +1,658 @@
...
+        ((void)L, fputs(p, stdout), fflush(stdout),  /* show prompt */ \
+        fgets(b, LUA_MAXINPUT, stdin) != NULL)  /* get line */
+#define lua_saveline(L,line)	{ (void)L; (void)line; }
+#define lua_freeline(L,b)	{ (void)L; (void)b; }
+
+#endif				/* } */
+
+#endif				/* } */
+
+
```
jia200x: extra empty lines here

examples/lua/lua.c
danpetry: A lot of this is implementation of a cli program, rather than the api - similar to e.g. examples/gcoap/gcoap_cli.c. So, is it best to think now about the API which is going to be exposed to an application and rearrange/separate out the lua.c file accordingly? My approach was to leave it to "evolve", but maybe the more usual approach is that code merged to master needs to be well thought out everywhere, is this true? If so will spend some time here...

danpetry: Further thoughts: just spoke to @miri64 who suggested I think more about what the user would need to know about to implement a program (goes in example) and what they don't need to know (goes in pkg)

jcarrano: I'm working on Lua now. The lua.c file is going away. It contains a REPL that is best implemented in Lua itself.

examples/lua/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+USEPKG += lua
+
+include $(RIOTBASE)/Makefile.include
+
+# The code below generates a header file from any .lua scripts in the
+# example directory. The header file contains a byte array of the
+# ASCII characters in the .lua script.
+
+LUA_PATH := $(BINDIR)/lua
```
jia200x: This lines can be moved to the pkg Makefiles and invoked when USEPKG+=lua. See e.g [here](https://github.com/RIOT-OS/RIOT/blob/master/pkg/openthread/Makefile#L16)

examples/lua/main.c
```diff
@@ -0,0 +1,52 @@
...
+int lua_run_main_script(int argc, char **argv)
+{
+    (void)argc;
+    (void)argv;
+
+    return lua_run_char_array(main_lua, main_lua_len);
+}
+
+static const shell_command_t shell_commands[] = {
+    { "run_lua_script", "Run main.lua", lua_run_main_script},
```
jia200x: this command can be dropped, since there's no configuration required to run the main.lua script

pkg/lua/Makefile
```diff
@@ -0,0 +1,13 @@
+PKG_NAME=lua
+PKG_URL=https://github.com/lua/lua.git
+PKG_VERSION=v5-3-4
+PKG_LICENSE=MIT
+
+.PHONY: all
+
+all:
+#	The lower case "makefile" is so that it copies over the lua makefile
```
jia200x: It's possible to specify a file with -f. It should be enough to copy the file

pkg/lua/contrib/lua_run_char_array.c
```diff
@@ -0,0 +1,155 @@
...
+
+#include <signal.h>
+#include <stdlib.h>
+#include "lauxlib.h"
+#include "lualib.h"
+
+static lua_State *globalL = NULL;
+
+static const char *progname = "lua";
+
```
jia200x: Here and below, please remove double empty lines. Should be just one empty line between 2 blocks of code

pkg/lua/contrib/lua_run_char_array.c
```diff
@@ -0,0 +1,155 @@
...
+
+/*
+** Hook set by signal function to stop the interpreter.
+*/
+static void lstop (lua_State *L, lua_Debug *ar) {
```
jia200x: although static, do you think it's possible to prefix these functions?

pkg/lua/contrib/lua_run_char_array.c
```diff
@@ -0,0 +1,155 @@
...
+
+
+/*
+** Function to be called at a C signal. Because a C signal cannot
+** just change a Lua state (as there is no proper synchronization),
+** this function only sets a hook that, when called, will stop the
+** interpreter.
+*/
+static void laction (int i) {
+  signal(i, SIG_DFL); /* if another SIGINT happens, terminate process */
```
jia200x: Identation should be 4 spaces

examples/lua/main.c
```diff
@@ -0,0 +1,52 @@
...
+
+#include <stdio.h>
+
+#include "msg.h"
+#include "shell.h"
+
+#include "main.lua.h"
+
+extern int lua_main(int argc, char **argv);
+extern int lua_run_char_array(const char *buffer, size_t buffer_len);
```
jia200x: This should be in a proper header file with it's Doxygen documentation. See e.g (here)[https://github.com/RIOT-OS/RIOT/blob/master/sys/include/base64.h#L56]

examples/lua/main.c
```diff
@@ -0,0 +1,52 @@
...
+ * @brief       Basic lua example application
+ *
+ * @author      Daniel Petry <daniel.petry@fu-berlin.de>
+ *
+ * @}
+ */
+
+#include <stdio.h>
+
+#include "msg.h"
```
jia200x: why is msg.h required here?

examples/lua/Makefile
jia200x: on second thoughts, maybe it's not bad to leave it as is. It explicitly tells the developer where's the file. But it could also be documented in the README.md. I'm OK with both ;)

jia200x:
@danpetry just wondering, I've check most of the functions are already implemented in external sources ([here](https://www.lua.org/source/5.3/lua.c.html) for instances). Is this file available in the Lua implementation? If so, maybe it could be a good idea to reuse as much as possible.

examples/lua/Makefile
danpetry: I guess this wasn't too clear - this path variable is only used in this example, to contain the path where the script should be parsed to. I've renamed this LUA_SCRIPT_HEADER_PATH to avoid confusion and documented it in the readme. Also, it now builds to examples/lua/scripts. What do you think?

jia200x:
@danpetry what differences exists between their CLI and the one proposed here? If it's not part of the official CLI implementation, I think using an external file might help maintainance (if possible though).

What do you think?

jia200x:
As discussed offline with @danpetry, the original lua.c can be reused. Only some small functions in contrib folder would be required.

examples/lua/Makefile
danpetry: I'd prefer to keep it as it is, for consistency with examples/javascript and pkg/jerryscript which do the same thing.

danpetry:
There are a few options for implementation of the pkg/contrib code

1. Have a cut-down version of lua.c in pkg/contrib

- Replicates some upstream code
- Avoids a patch, allowing code to be edited more readily for future development on this pkg
- Currently, this is what's done. (see pkg/contrib/lua_run_char_array.c)

2. patch the lua.c file.

- Some functions need to have their scope expanded by removing the static keyword. These are:

lstop - stops the interpreter
laction - sets a hook for lstop to stop the interpreter on a c signal
l_message - prints an error message
report - checks status and calls l_message if not ok
msghandler, docall, dochunk - run a chunk of lua code

- The lua_run_char_array function is then implemented in the pkg/contrib directory, and the functions above declared in that file as externs


Basically, it seems to be a choice between replication of upstream code or a patch. Opinions please? :)

danpetry:
So, here is another option which I think is better

The API wrapper code is not patched from Lua code, and it's not downstreamed either. It's re-written in a well-structured and maintainable way that can be extended in future iterations and included in the RIOT repo. It would need to be significantly different from the Lua code for this so neither patched nor downstreamed (nor a combination of both)

jia200x:
@danpetry can you provide a concrete example of this kind of wrapper?

If it's possible to call Lua instructions with the raw package, the current PR can be merged. I think the wrapper can come later, specially since there are other high level languages that might require also an abstraction layer.

danpetry:
ok so I'll move the implementation of the script parser to the example code?

jia200x:
@danpetry check the [current JS implementation](https://github.com/RIOT-OS/RIOT/tree/master/pkg/jerryscript). It fetches the original jerryscript code, and the application calls pkg functions.

I tend to think the Lua CLI can be added in another PR, so this one can easily gets merged with a simple example file showing actual Lua code. What do you think?

danpetry:
That sounds fine to me. Will do :)

examples/lua/main.c
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @}
+ */
+
+#include <stdio.h>
+#include "shell.h"
+
+#include "main.lua.h"
+
+extern int lua_run_char_array(const char *buffer, size_t buffer_len);
```
jia200x: This function should be included from a header file with its proper Doxygen documentation. As far as I see, it's intended to be called by the user

examples/lua/Makefile
jia200x: ACK

jia200x: maybe we can think about this later

examples/lua/main.c
```diff
@@ -0,0 +1,58 @@
...
+
+#include "main.lua.h"
+
+int lua_run_script (const char *buffer, size_t buffer_len ){
+
+    lua_State *L = luaL_newstate();
+
+    if (L == NULL) {
+        puts("cannot create state: not enough memory");
+        return EXIT_FAILURE;
```
jia200x: @danpetry are this return codes LUA specific?

examples/lua/main.c
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include "lauxlib.h"
+#include "lualib.h"
+
+#include "main.lua.h"
+
+int lua_run_script (const char *buffer, size_t buffer_len ){
```
jia200x: please remove space between function name and parentheses, and between `buffer_len` and parentheses.
Also, add a space between the closing parentheses and curly bracket

examples/lua/main.c
```diff
@@ -0,0 +1,58 @@
...
+
+    lua_State *L = luaL_newstate();
+
+    if (L == NULL) {
+        puts("cannot create state: not enough memory");
+        return EXIT_FAILURE;
+    }
+
+    luaL_openlibs(L);
+
```
jia200x: I would prefer to remove this extra line here

examples/lua/main.c
```diff
@@ -0,0 +1,58 @@
...
+    luaL_loadbuffer(L, buffer, buffer_len, "lua input script");
+
+    if (lua_pcall(L, 0, 0, 0) != LUA_OK){
+        puts("Lua script running failed");
+        return EXIT_FAILURE;
+    }
+
+    lua_close(L);
+
```
jia200x: also here

examples/lua/main.c
```diff
@@ -0,0 +1,58 @@
...
+    return EXIT_SUCCESS;
+}
+
+int main(void)
+{
+    puts("Lua RIOT build");
+
```
jia200x: and here

pkg/lua/Makefile
```diff
@@ -0,0 +1,13 @@
+PKG_NAME=lua
+PKG_URL=https://github.com/lua/lua.git
+PKG_VERSION=v5-3-4
```
jia200x: This should pointing to a commit. See e.g [here](https://github.com/RIOT-OS/RIOT/blob/master/pkg/jerryscript/Makefile#L3)

jia200x:
@danpetry please address the comments above, squash and I give the ACK. I tested and works as expected.

Then I need @kaspar030 approval (or dismissing the review ;) in order to merge.

danpetry:
@kaspar030 said a few weeks ago he's ok with us proceeding, he won't have a look at it.

Addressing the rest of the comments now.

jia200x:
@danpetry Ok

examples/lua/Makefile
danpetry: Where's the best place for this kind of discussion after a PR has been closed?

jia200x: Another PR/RFC should do the trick

examples/lua/main.c
danpetry: Putting the opening curly bracket on the next line as I see that seems to be the most common way of doing it

examples/lua/main.c
danpetry: No, they're defined in stdlib.h

examples/lua/main.c
jia200x: yes, sorry. Curly bracket here should be in next line

examples/lua/main.c
jia200x: I would prefer here `errno` (e.g ENOBUFS)

examples/lua/main.c
danpetry: Thanks. Re the coding conventions, how much will uncrustify catch if I run that? Or is it more a case of learning by doing? (With some leeway re personal preference in certain cases?)

examples/lua/main.c
danpetry: I've used ENOMEM for this one; EINTR for failure of the protected call to the lua engine, to run the script; and `return 0;` for success in line with most other examples I looked at.

jia200x: fine ;)

danpetry:
I've made the changes, squashed, rebased, and pushed.
However, now it doesn't build for ARM boards, for the reasons detailed in https://github.com/RIOT-OS/RIOT/issues/9051. I narrowed the step which stopped it building down to the rebasing step. So, there must have been a change on RIOT master which now requires linking in some new functions. Going to do some more investigation, but if anyone has any helpful info or suggestions on how to proceed in the meantime, that would be helpful.
Initial thoughts are to get https://github.com/RIOT-OS/RIOT/pull/9090 merged, but we need to understand why it linked before and not now, in case there are other implications.

danpetry:
Update: it was this PR, which essentially removed some newlib stubs, which stops it building: https://github.com/RIOT-OS/RIOT/pull/8795

pkg/lua/Makefile
```diff
@@ -3,7 +3,7 @@ PKG_URL=https://github.com/lua/lua.git
 PKG_VERSION=e354c6355e7f48e087678ec49e340ca0696725b1
 PKG_LICENSE=MIT
 PKG_BUILDDIR ?= $(BINDIRBASE)/pkg/$(BOARD)/$(PKG_NAME)
-
+#PKG_SOURCE_LOCAL=~/Sandbox/lua
```
jia200x: Why is this line here?

pkg/lua/Makefile.lua
```diff
@@ -1,9 +1,13 @@
...
-# This builds using POSIX system calls and some extra libraries, and
+# This builds for native using POSIX system calls and some extra libraries, and
 # removes a compiler warning that warns against using tmpnam().
 ifeq ($(BOARD),native)
 CFLAGS += -DLUA_USE_LINUX
 endif
 
+# This redefines a lua prng function, avoiding the need for the _times system
+# call, which RIOT doesn't provide
+CFLAGS += -D'l_randomizePivot()=0'
```
jia200x: I think it's a cleaner approach to remove this function with the patch

pkg/lua/patches/0001-Remove-the-Lua-functions-os.clock-and-os.rename.patch
```diff
@@ -0,0 +1,39 @@
...
+--- a/loslib.c
++++ b/loslib.c
+@@ -157,9 +157,7 @@ static int os_remove (lua_State *L) {
+ 
+ 
+ static int os_rename (lua_State *L) {
+-  const char *fromname = luaL_checkstring(L, 1);
+-  const char *toname = luaL_checkstring(L, 2);
+-  return luaL_fileresult(L, rename(fromname, toname) == 0, NULL);
++  return luaL_error(L, "This function is not implemented in RIOT yet");
```
jia200x: This errors can be in a buffer, I don't think RIOT compiles with "string pooling".

jcarrano: What's the advantage of putting the strings in buffers? String literals are all over the lua source code.

pkg/lua/Makefile.lua
jcarrano: The lua source is written in such a way that one is supposed to redefine things to change them. In this case, the definition of `l_randomizePivot()` is wrapped inside a `#ifndef l_randomizePivot`

jcarrano:
@jia200x @danpetry The next PR on lua will remove the REPL written in C (that is in lua.c) and replace it by a smaller and more elegant REPL written in Lua itself.

pkg/lua/Makefile.lua
danpetry: I'm for putting all the patches for this fixup together in one patch. eg

`#define l_randomizePivot() 0`

just above `if !defined(l_randomizePivot)`

sorts it out.

pkg/lua/patches/0001-Remove-the-Lua-functions-os.clock-and-os.rename.patch
danpetry: I agree with juan re this... I don't actually think it makes that much difference though.

jia200x: in theory saves some ROM bytes, but if's the whole source code it won't make a big difference

jia200x: ok, let's leave it like that then

jia200x:
ok, works as expected. Tested in samr21-xpro and native. ACK

miri64:
@aabadie @kaspar030 @smlng Murdock has an issue with this PR that we can't reproduce locally. Any idea what is wrong?

danpetry:
lauxlib.h is part of the pkg added as part of this PR. I suspect it's something to do with the way I've specified include paths. the two "debug commits" are there to get more info

miri64:
> @miri64 I guess the murdock slave don't have the LUA-dev package? That's why:
> 
> ```
> fatal error: lauxlib.h: No such file or directory
> ```

@smlng I tested the same application with one of the failing boards on my dev machine and installed nothing extra (granted this is no guarantee that something isn't missing, but I most certainly don't have liblua-dev installed ;-)) and it worked.

```
$ apt list --installed | grep -i "lua"

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

liblua5.1-0/xenial,now 5.1.5-8ubuntu1 amd64 [installed,automatic]
liblua5.2-0/xenial,now 5.2.4-1ubuntu1 amd64 [installed,automatic]
libluajit-5.1-2/xenial,now 2.0.4+dfsg-1 amd64 [installed,automatic]
libluajit-5.1-common/xenial,xenial,now 2.0.4+dfsg-1 all [installed,automatic]
libtexlua52/xenial,now 2015.20160222.37495-1 amd64 [installed,automatic]
libtexluajit2/xenial,now 2015.20160222.37495-1 amd64 [installed,automatic]
lua-lpeg/xenial,now 0.12.2-1 amd64 [installed,automatic]
texlive-luatex/xenial-updates,xenial-updates,xenial-security,xenial-security,now 2015.20160320-1ubuntu0.1 all [installed,automatic]
```

pkg/lua/Makefile.lua
```diff
@@ -0,0 +1,9 @@
+SRC := $(filter-out lua.c luac.c,$(wildcard *.c))
+
+# This builds for native using POSIX system calls and some extra libraries, and
+# removes a compiler warning that warns against using tmpnam().
+ifeq ($(BOARD),native)
+CFLAGS += -DLUA_USE_LINUX
```
aabadie: minor, this should be indented with 2 spaces

examples/lua/Makefile
```diff
@@ -0,0 +1,51 @@
...
+# development process:
+DEVELHELP ?= 1
+
+# Change this to 0 show compiler invocation lines by default:
+QUIET ?= 1
+
+USEMODULE += ps
+
+ifneq ($(BOARD),native)
+# This stack size is large enough to run Lua print() functions of
```
aabadie: indent this block (2 spaces)

pkg/lua/Makefile.lua
miri64: A frustrated Dan just came to my office. Can we stop asking for minor changes when a PR is already ACK'd, please? This is also something that scares away contributors, IMHO. If the PR is defective and the ACK was given wrongly, yes then asking for changes is definitely okay. But not like this! Yes code quality is important, but these kind of non-functional errors can also be fixed in post.

aabadie: I was pinged for a build issue, so I came, had a look at the package makefile, saw this and finally decided to **comment** this. I'm not requested any change.

tcschmidt: Yep, and I guess @miri64 hit the point: Reviewers should carefully work with authors to get things into a state at which code can be merged. For this reviews need structuring, streamlining, and pacing in the most constructive manner. 
This means for instance that code-style review should be completed at some early stage, but also means to aid authors in case of troubles with build or similar.
I guess that's a generic discussion, but a very important one: I've seen many old open PRs where authors have been exhausted by being pushed around randomly without hitting anywhere.

danpetry: I mean, I was just going to make the changes, as it would have been quicker than having a discussion about it. But the impression I get is that the PRs building up is a real problem and the review process itself should a) allow for valuable support of contributors by maintainers to get it into a good shape, which tbh has formed the bulk of my time on this PR; but also b) try not to duplicate work and streamline the process otherwise. I think it would be worth talking about tomorrow

miri64: > I was pinged for a build issue, so I came, had a look at the package makefile, saw this and finally decided to comment this. I'm not requesting any change.

Just to be clear, I appreciate that you only *commented* on this and did not make a change request.

> I didn't know an ACK by one reviewer should stop others from pointing out errors or possible improvements (as minor as they might be) anyway.

I already pointed out, that it's totally fine if the PR is defective, a late comment is ok. But it certainly gives a weird impression if a person that wasn't involved in the discussion before suddenly comes in and also comments on minor, non-functional details, when the PR was already ACK'd (even if they were pinged in for a completely different issue).

danpetry: I think it's worth talking about this over voice call tomorrow... not sure this is a great format for this kind of discussion... :)

jia200x:
ok, all green. Re-ACK.

smlng:
@jia200x then hit merge 🎉

jia200x:
>@jia200x then hit merge :tada:
pic

jia200x:
GO!

jia200x:
congratz @danpetry :tada: :tada:

danpetry:
Woop!

PR #7666, opened 2017-10-02
pm_layered: add test app

vincent-d:
Following #7385, this is a test app for pm_layered

PR #6445, opened 2017-01-20
msp430: Add newlib support

justinjoy:
This PR is made based on #4766. I rebased and took some commits from the original PR and they are little squashed, and some of them aren't used.
Unlike the original PR, I tend to use ldscripts from TI without changes.

Although I can build successfully, there are two issues. Please refer to the log below.
 - `Command not found` is shown when starting build process, but I failed to find exact point of that.
 -  As we talked in IRC, the generated `hex` is too large for `hello-world`.  (29Kb)

```
$ make BOARD=msb-430
make: command: Command not found
Building application "hello-world" for "msb-430" with MCU "msp430fxyz".

"make" -C /home/joykim/iot/RIOT/boards/msb-430
"make" -C /home/joykim/iot/RIOT/boards/msb-430-common
"make" -C /home/joykim/iot/RIOT/boards/msb-430-common/drivers
"make" -C /home/joykim/iot/RIOT/core
"make" -C /home/joykim/iot/RIOT/cpu/msp430fxyz
"make" -C /home/joykim/iot/RIOT/cpu/msp430_common
"make" -C /home/joykim/iot/RIOT/cpu/msp430fxyz/periph
"make" -C /home/joykim/iot/RIOT/drivers
"make" -C /home/joykim/iot/RIOT/drivers/periph_common
"make" -C /home/joykim/iot/RIOT/sys
"make" -C /home/joykim/iot/RIOT/sys/auto_init
"make" -C /home/joykim/iot/RIOT/sys/isrpipe
"make" -C /home/joykim/iot/RIOT/sys/newlib
"make" -C /home/joykim/iot/RIOT/sys/oneway-malloc
"make" -C /home/joykim/iot/RIOT/sys/tsrb
"make" -C /home/joykim/iot/RIOT/sys/uart_stdio
   text    data     bss     dec     hex filename
  10193     274    1074   11541    2d15 /home/joykim/iot/RIOT/examples/hello-world/bin/msb-430/hello-world.elf
```

PR #8848, opened 2018-03-28
tests: periph_pm: add peripheral test.

basilfx:
### Contribution description

This PR adds a test application for `periph_pm`. Although you need a way to measure power consumption to really use this test, I think its a nice way to play with (or debug) the power modes and observe what the CPU will do.

I had this one lying around, somewhere. Initially I used this (before `periph_pm`) to test the power modes on the EFM32 development kits, but rearranged it for `periph_pm`.

### Issues/PRs references

None

### DIFF
--- a/tests/periph_pm/Makefile
+++ b/tests/periph_pm/Makefile
@@ -1,0 +1,11 @@
+BOARD ?= slstk3401a
+include ../Makefile.tests_common
+
+# method `fflush()` is not defined for MSP-430 (#6445 will fix this)
+BOARD_BLACKLIST := chronos msb-430h msb-430 telosb wsn430-v1_3b wsn430-v1_4 z1
+
+FEATURES_OPTIONAL += periph_rtc
+
+USEMODULE += shell
+
+include $(RIOTBASE)/Makefile.include
--- a/tests/periph_pm/README.md
+++ b/tests/periph_pm/README.md
@@ -1,0 +1,14 @@
+Expected result
+===============
+You should be presented with the RIOT shell, providing you with commands to test the power
+management implementation by forcing the CPU to set a certain power mode. An external power
+measurement device is needed to observe the effect.
+
+Some power modes may require a certain event to wake up. Reset the CPU if needed.
+
+If a RTC peripheral is available, an additional command to temporarily unblock a power mode is
+available (provided that the CPU can wake-up from given power mode).
+
+Background
+==========
+Test the functionality of the platform's power management implementation.
--- a/tests/periph_pm/main.c
+++ b/tests/periph_pm/main.c
@@ -1,0 +1,248 @@
+/*
+ * Copyright (C) 2016-2018 Bas Stottelaar <basstottelaar@gmail.com>
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     tests
+ * @{
+ *
+ * @file
+ * @brief       Power management peripheral test.
+ *
+ * @author      Bas Stottelaar <basstottelaar@gmail.com>
+ * @author      Vincent Dupont <vincent@otakeys.com>
+ *
+ * @}
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "periph/pm.h"
+#ifdef MODULE_PM_LAYERED
+#ifdef MODULE_PERIPH_RTC
+#include "periph/rtc.h"
+#endif
+#include "pm_layered.h"
+#endif
+#include "shell.h"
+
+#ifdef MODULE_PM_LAYERED
+static int check_mode(int argc, char **argv)
+{
+    if (argc < 2) {
+        printf("Usage: %s <power mode>\n", argv[0]);
+        return -1;
+    }
+
+    return 0;
+}
+
+static int parse_mode(char *argv)
+{
+    uint8_t mode = atoi(argv);
+
+    if (mode >= PM_NUM_MODES) {
+        printf("Error: power mode not in range 0 - %d.\n", PM_NUM_MODES - 1);
+        return -1;
+    }
+
+    return mode;
+}
+
+#ifdef MODULE_PERIPH_RTC
+static int check_mode_duration(int argc, char **argv)
+{
+    if (argc != 3) {
+        printf("Usage: %s <power mode> <duration (s)>\n", argv[0]);
+        return -1;
+    }
+
+    return 0;
+}
+
+static int parse_duration(char *argv)
+{
+    int duration = atoi(argv);
+
+    if (duration < 0) {
+        puts("Error: duration must be a positive number.");
+        return -1;
+    }
+
+    return duration;
+}
+
+static void cb_rtc(void *arg)
+{
+    int level = (int)arg;
+
+    pm_block(level);
+}
+#endif /* MODULE_PERIPH_RTC */
+#endif /* MODULE_PM_LAYERED */
+
+static int cmd_off(int argc, char **argv)
+{
+    (void) argc;
+    (void) argv;
+
+    puts("CPU will turn off.");
+    fflush(stdout);
+
+    pm_off();
+
+    return 0;
+}
+
+static int cmd_reboot(int argc, char **argv)
+{
+    (void) argc;
+    (void) argv;
+
+    puts("CPU will reboot.");
+    fflush(stdout);
+
+    pm_reboot();
+
+    return 0;
+}
+
+#ifdef MODULE_PM_LAYERED
+static int cmd_block(int argc, char **argv)
+{
+    if (check_mode(argc, argv) != 0) {
+        return 1;
+    }
+
+    int mode = parse_mode(argv[1]);
+
+    if (mode < 0) {
+        return 1;
+    }
+
+    printf("Blocking power mode %d.\n", mode);
+    fflush(stdout);
+
+    pm_block(mode);
+
+    return 0;
+}
+
+static int cmd_set(int argc, char **argv)
+{
+    if (check_mode(argc, argv) != 0) {
+        return 1;
+    }
+
+    int mode = parse_mode(argv[1]);
+
+    if (mode < 0) {
+        return 1;
+    }
+
+    printf("CPU will enter power mode %d.\n", mode);
+    fflush(stdout);
+
+    pm_set(mode);
+
+    return 0;
+}
+
+static int cmd_unblock(int argc, char **argv)
+{
+    if (check_mode(argc, argv) != 0) {
+        return 1;
+    }
+
+    int mode = parse_mode(argv[1]);
+
+    if (mode < 0) {
+        return 1;
+    }
+
+    printf("Unblocking power mode %d.\n", mode);
+    fflush(stdout);
+
+    pm_unblock(mode);
+
+    return 0;
+}
+
+#ifdef MODULE_PERIPH_RTC
+static int cmd_unblock_rtc(int argc, char **argv)
+{
+    if (check_mode_duration(argc, argv) != 0) {
+        return 1;
+    }
+
+    int mode = parse_mode(argv[1]);
+    int duration = parse_duration(argv[2]);
+
+    if (mode < 0 || duration < 0) {
+        return 1;
+    }
+
+    printf("Unblocking power mode %d for %d seconds.\n", mode, duration);
+    fflush(stdout);
+
+    struct tm time;
+
+    rtc_get_time(&time);
+    time.tm_sec += duration;
+    mktime(&time);
+    rtc_set_alarm(&time, cb_rtc, (void *)mode);
+
+    pm_unblock(mode);
+
+    return 0;
+}
+#endif /* MODULE_PERIPH_RTC */
+#endif /* MODULE_PM_LAYERED */
+
+/**
+ * @brief   List of shell commands for this example.
+ */
+static const shell_command_t shell_commands[] = {
+    { "off", "turn off", cmd_off },
+    { "reboot", "reboot", cmd_reboot },
+#ifdef MODULE_PM_LAYERED
+    { "block", "block power mode", cmd_block },
+    { "set", "set power mode", cmd_set },
+    { "unblock", "unblock power mode", cmd_unblock },
+#ifdef MODULE_PERIPH_RTC
+    { "unblock_rtc", "temporary unblock power mode", cmd_unblock_rtc },
+#endif
+#endif
+    { NULL, NULL, NULL }
+};
+
+/**
+ * @brief   Application entry point.
+ */
+int main(void)
+{
+    char line_buf[SHELL_DEFAULT_BUFSIZE];
+
+    /* print test application information */
+#ifdef MODULE_PM_LAYERED
+    printf("This application allows you to test the CPU power management.\n"
+           "The available power modes are 0 - %d. Lower-numbered power modes\n"
+           "save more power, but may require an event/interrupt to wake up\n"
+           "the CPU. Reset the CPU if needed.\n",
+           PM_NUM_MODES - 1);
+#else
+    puts("This application allows you to test the CPU power management.\n"
+         "Layered support is not unavailable for this CPU. Reset the CPU if\n"
+         "needed.");
+#endif
+
+    /* run the shell and wait for the user to enter a mode */
+    shell_run(shell_commands, line_buf, SHELL_DEFAULT_BUFSIZE);
+
+    return 0;
+}


gebart:
Cool! Useful for reference measurements to compare with what the real application uses

aabadie:
Sorry, but it looks like a duplicate of #7666.

basilfx:
@aabadie Oops, yes. I did search for `periph_pm` before submitting :-)

@vincent-d Maybe we can merge both PR's?

vincent-d:
I have no time to work on this at the moment, so you can go ahead with this one! I'll try to review it. 

If it gets merged, we'll close mine.

basilfx:
Merged this with the work by @vincent-d in #7666.

I chose to add an additional `unblock_rtc` to simplify code/ifdefs. It only works if the CPU provides `pm_layered` and `periph_rtc`.

Screenshot below show a gap of three seconds of scheduled unblocking. The other gap is a normal unblock (with UART events inbetween).

schermafbeelding 2018-03-30 om 00 09 30

tests/periph_pm/README.md
```diff
@@ -0,0 +1,14 @@
+Expected result
+===============
+You should be presented with the RIOT shell, providing you with three commands to test the power
```
vincent-d: There are a bit more than three commands ;)

tests/periph_pm/Makefile
```diff
@@ -0,0 +1,11 @@
+APPLICATION = periph_pm
```
aabadie: You don't need to set this variable

tests/periph_pm/README.md
```diff
@@ -0,0 +1,14 @@
...
+management implementation by forcing the CPU to set a certain power mode. An external power
+measurement device is needed to observe the effect.
+
+Some power modes may require a certain event to wake up. Reset the CPU if needed.
+
+If a RTC peripheral is available, an additional command to temporary unblock a power mode is
```
aabadie: s/temporary/temporarily/

tests/periph_pm/main.c
```diff
@@ -0,0 +1,248 @@
...
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "periph/pm.h"
+#ifdef MODULE_PERIPH_RTC
+#include "periph/rtc.h"
+#endif
+#include "pm_layered.h"
```
aabadie: I think this include should be made conditional with `#ifdef MODULE_PM_LAYERED`

tests/periph_pm/main.c
```diff
@@ -0,0 +1,248 @@
...
+    if (mode < 0 || duration < 0) {
+        return 1;
+    }
+
+    printf("Unblocking power mode %d for %d seconds.\n", mode, duration);
+    fflush(stdout);
+
+    rtc_get_time(&time);
+    time.tm_sec += duration;
+    while (time.tm_sec > 60) {
```
aabadie: Do you think it would make sense to use mktime/localtime functions to automatically compute the correct `struct tm` ?

tests/periph_pm/main.c
```diff
@@ -0,0 +1,248 @@
...
+}
+#endif
+#endif
+
+int main(void)
+{
+    char line_buf[SHELL_DEFAULT_BUFSIZE];
+
+    /* print some information about the modes */
+    printf("This application allows you to test the power management\n"
```
aabadie: Not if power management is a peripheral. I would just say `to test the CPU power management`

tests/periph_pm/main.c
```diff
@@ -0,0 +1,248 @@
...
+static int cmd_unblock(int argc, char **argv);
+#ifdef MODULE_PERIPH_RTC
+static int cmd_unblock_rtc(int argc, char **argv);
+#endif
+#endif
+
+/**
+ * @brief   List of shell commands for this example.
+ */
+static const shell_command_t shell_commands[] = {
```
aabadie: In general this array is created just before the main function

tests/periph_pm/main.c
```diff
@@ -0,0 +1,248 @@
...
+#ifdef MODULE_PM_LAYERED
+    { "block", "block power mode", cmd_block },
+#endif
+    { "lowest", "enter lowest mode", cmd_lowest },
```
aabadie: Why is this command needed ? if one unblock a power mode with `unblock command`, then the main thread (only thread running) will be blocked (waiting for incoming commands), so the system switch to idle thread that automatically call `pm_set_lowest`.

tests/periph_pm/main.c
basilfx: I don't understand this. Can you explain?

Do you mean that I should move this just before the main method?

tests/periph_pm/main.c
basilfx: Removed `cmd_lowest`.

tests/periph_pm/main.c
basilfx: Fixed.

tests/periph_pm/main.c
basilfx: Yes. Fixed.

tests/periph_pm/README.md
basilfx: Fixed.

tests/periph_pm/Makefile
basilfx: Fixed.

tests/periph_pm/README.md
basilfx: Yes :-)

Fixed.

tests/periph_pm/main.c
basilfx: Not sure how I would use these methods to just add a certain amount of seconds.

I copied this part from #7666.

tests/periph_pm/main.c
aabadie: Yes, and then you don't need to declare the command function.

tests/periph_pm/main.c
aabadie: there can be side effects if the duration also increases hours. I think those functions handles that better (not tested though)

basilfx: Ah I see. Found something [here](https://stackoverflow.com/a/1860996), and it seems to work (ok, I didn't test whole hours, but minutes is OK).

tests/periph_pm/main.c
aabadie: Out-of-curiosity, do you confirm that it works as expected (in terms of power consumption) ?

tests/periph_pm/main.c
basilfx: Fixed.

tests/periph_pm/main.c
basilfx: Yes, I observe a difference when I change the power modes.

schermafbeelding_2018-03-30_om_10_45_48

aabadie: really nice!

basilfx:
Fixed the comments and rebased.

tests/periph_pm/main.c
```diff
@@ -0,0 +1,218 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "periph/pm.h"
+#ifdef MODULE_PM_LAYERED
+#ifdef MODULE_PERIPH_RTC
+#include "periph/rtc.h"
+#endif
+#endif
+#include "pm_layered.h"
```
aabadie: include is still out of the conditional `#ifdef`

basilfx: Oops, sorry. Fixed that one.

tests/periph_pm/main.c
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    printf("Blocking power mode %d.\n", mode);
+    fflush(stdout);
+
+    pm_block(mode);
+
+    return 0;
+}
+#endif
```
aabadie: Maybe add a comment which block is closed here (MODULE_PM_LAYERED)

tests/periph_pm/main.c
```diff
@@ -0,0 +1,218 @@
...
+    puts("CPU will reboot.");
+    fflush(stdout);
+
+    pm_reboot();
+
+    return 0;
+}
+
+#ifdef MODULE_PM_LAYERED
+static int cmd_set(int argc, char **argv)
```
aabadie: Maybe these functions could be grouped in the same `#ifdef MODULE_PM_LAYERED` above block.

tests/periph_pm/main.c
```diff
@@ -0,0 +1,218 @@
...
+}
+#endif
+#endif
+
+/**
+ * @brief   List of shell commands for this example.
+ */
+static const shell_command_t shell_commands[] = {
+#ifdef MODULE_PM_LAYERED
+    { "block", "block power mode", cmd_block },
```
aabadie: Same here, this line can be moved before the `unblock` line.

tests/periph_pm/Makefile
```diff
@@ -0,0 +1,10 @@
+BOARD ?= slstk3401a
+include ../Makefile.tests_common
+
+FEATURES_REQUIRED += periph_pm
```
aabadie: I don't think this is needed, this feature is made optional (see Makefile.dep). And that's probably why the CI is failing.

tests/periph_pm/main.c
basilfx: Wanted to keep it sorted alphabetically. Will move it to first one in the other MODULE_PM_LAYERED block.

tests/periph_pm/main.c
basilfx: Done.

tests/periph_pm/main.c
basilfx: Done

tests/periph_pm/Makefile
basilfx: Fixed.

tests/periph_pm/main.c
```diff
@@ -0,0 +1,216 @@
...
+static void cb_rtc(void *arg)
+{
+    int level = (int)arg;
+
+    pm_block(level);
+}
+#endif /* MODULE_PERIPH_RTC */
+#endif /* MODULE_PM_LAYERED */
+
+static int cmd_off(int argc, char **argv)
```
aabadie: `cmd_off` and `cmd_reboot` should be move to the top and then all functions that requires MODULE_PM_LAYERED could be grouped in the same conditional `#ifdef`.

tests/periph_pm/main.c
```diff
@@ -0,0 +1,216 @@
...
+int main(void)
+{
+    char line_buf[SHELL_DEFAULT_BUFSIZE];
+
+    /* print some information about the modes */
+    printf("This application allows you to test the CPU power management.\n"
+           "The available power modes are 0 - %d. Lower-numbered power modes\n"
+           "save more power, but may require an event/interrupt to wake up\n"
+           "the CPU. Reset the CPU if needed.\n",
+           PM_NUM_MODES - 1);
```
aabadie: The use of PM_NUM_MODE should be excluded for boards not providing the pm_layered module.

tests/periph_pm/main.c
basilfx: I'd like to keep it this way: I now have the commands grouped together, and utility functions on top.

tests/periph_pm/main.c
basilfx: Fixed.

basilfx:
Few MSP430-based board don't compile due to missing `fflush`. Any ideas?

gebart:
If we can get #6445 finished all such problems would go away.

basilfx:
Thanks @gebart for pointing that out.

Since #6445 is not active, I propose to either blacklist the boards, or to put `#ifndef MODULE_MSP430_COMMON` around `fflush`.

I prefer the latter, since the test itself is still valuable.

aabadie:
> I propose to either blacklist the boards

+1 for that solution (already used in other places)

basilfx:
@aabadie Can I squash? Murdock is almost happy now :-)

tests/periph_pm/main.c
```diff
@@ -0,0 +1,222 @@
...
+
+#ifdef MODULE_PM_LAYERED
+static int parse_mode(int argc, char **argv)
+{
+    if (argc < 2) {
+        printf("Error: power mode expected (0 - %d).\n", PM_NUM_MODES - 1);
+        return -1;
+    }
+
+    int mode = atoi(argv[1]);
```
aabadie: mode could be an `uint8_t`

tests/periph_pm/main.c
```diff
@@ -0,0 +1,222 @@
...
+
+    if (mode < 0 || mode >= (int) PM_NUM_MODES) {
```
aabadie: then here there's no need to check that mode is negative

tests/periph_pm/main.c
```diff
@@ -0,0 +1,222 @@
...
+
+    pm_reboot();
+
+    return 0;
+}
+
+#ifdef MODULE_PM_LAYERED
+static int cmd_block(int argc, char **argv)
+{
+    int mode = parse_mode(argc, argv);
```
aabadie: mode could be uint8_t

tests/periph_pm/main.c
```diff
@@ -0,0 +1,222 @@
...
+
+    if (mode < 0) {
```
aabadie: parse_mode function is already testing this

tests/periph_pm/main.c
```diff
@@ -0,0 +1,222 @@
...
+    pm_block(mode);
+
+    return 0;
+}
+
+static int cmd_set(int argc, char **argv)
+{
+    int mode = parse_mode(argc, argv);
+
+    if (mode < 0) {
```
aabadie: same comment about parse_mode

tests/periph_pm/main.c
```diff
@@ -0,0 +1,222 @@
...
+    fflush(stdout);
+
+    pm_set(mode);
+
+    return 0;
+}
+
+static int cmd_unblock(int argc, char **argv)
+{
+    int mode = parse_mode(argc, argv);
```
aabadie: parse_mode

tests/periph_pm/main.c
```diff
@@ -0,0 +1,222 @@
...
+
+#ifdef MODULE_PERIPH_RTC
+static int cmd_unblock_rtc(int argc, char **argv)
+{
+    struct tm time;
+
+    int mode = parse_mode(argc, argv);
+    int duration = parse_duration(argc, argv);
+
+    if (mode < 0 || duration < 0) {
```
aabadie: again parse_mode

tests/periph_pm/main.c
basilfx: This would indicate a error code, so I need to check this here.

tests/periph_pm/main.c
basilfx: This would indicate a error code, so I need to check this here.

tests/periph_pm/main.c
basilfx: This would indicate a error code, so I need to check this here.

tests/periph_pm/main.c
basilfx: Can you explain?

aabadie: arf, sorry, re-read the parse_mode function and testing the negative returned value is perfectly fine. The comment about uint8_t (in parse_mode) still stands.

tests/periph_pm/main.c
basilfx: Fixed.

tests/periph_pm/main.c
basilfx: Fixed.

tests/periph_pm/main.c
basilfx: Fixed.

basilfx:
Processed feedback and rebased.

basilfx:
@aabadie Good enough to squash?

tests/periph_pm/main.c
```diff
@@ -0,0 +1,222 @@
...
+#include "periph/rtc.h"
+#endif
+#include "pm_layered.h"
+#endif
+#include "shell.h"
+
+#ifdef MODULE_PM_LAYERED
+static int parse_mode(int argc, char **argv)
+{
+    if (argc < 2) {
```
aabadie: This should be moved to each command function handler and if a wrong number of arguments is given, the command prints a usage message `Usage: unblock <power mode>` or `Usage: unblock_rtc <power mode> <duration>`.
Otherwise the error message doesn't give any hint on how to correctly use the command. Maybe introduce a `_check_usage(int argc, int expected_argc)` function for this ?

Then the parse_mode and parse_duration could be simplified with this signature:
```c
static int _parse_mode(const char * mode)
static int _parse_duration(const char * duration)
```

basilfx:
@aabadie Have improved argument parsing + usage message. Is this better?

aabadie:
Then go!

PR #8854, opened 2018-03-29
net/gcoap: Internalize definition of state struct

kb2ma:
### Contribution description

Moves the definition of gcoap_state_t struct from gcoap.h to gcoap.c. Use of the struct is completely internal to the implementation, so the external definition is unnecessary. As shown in #8767, the external definition is painful for C++ users due to the use of atomic headers.

### Issues/PRs references

Partially addresses #8767.

### DIFF
--- a/sys/include/net/gcoap.h
+++ b/sys/include/net/gcoap.h
@@ -213,11 +213,9 @@
 #define NET_GCOAP_H
 
 #include <stdint.h>
-#include <stdatomic.h>
 
 #include "net/ipv6/addr.h"
 #include "net/sock/udp.h"
-#include "mutex.h"
 #include "net/nanocoap.h"
 #include "xtimer.h"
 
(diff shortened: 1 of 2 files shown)


kb2ma:
Murdock failed for MSP430 platforms with errors like:

> ../../../../cpu/msp430_common/include/stdatomic.h:212:9: error: expected specifier-qualifier-list before 'int_least8_t'

So, I added a fixup commit to move the stdint include just before the stdatomic include in gcoap.c, as it was in gcoap.h. @gebart, please confirm this change does not set off any alarms for you.

The builds fail now only due to the fixup. If we're OK, I'll squash and commit.

gebart:
That's a bug in the stdatomic.h header I guess. Keep the stdint.h include in gcoap.h as well though, since that header uses uint8_t etc from that file. (you might want to add stddef.h as well there because it uses size_t, but that's just a "while you're at it" kind of suggestion)

gebart:
@kb2ma Impressive build failure finding skills 😉

kb2ma:
> Keep the stdint.h include in gcoap.h as well though, since that header uses uint8_t etc from that file.

That leads to an interesting policy question. I thought about leaving stdint.h in gcoap.h, but both sock/udp.h and nanocoap.h already include it. So, it just seemed like clutter to keep it in gcoap.h, since it will certainly depend on those other two headers for the foreseeable future. Does that sway your guidance?

> Impressive build failure finding skills

Thanks, but I failed! I actually have a PR [checklist](https://github.com/kb2ma/RIOT/wiki/RIOT-PR-Checklist), and I ran it, but I ignored the results. :-( The code change was so simple and it ran on my samr21-xpro. So I figured that maybe something changed in the build mechanism to generate a false indication. Blaming the tool is so lame. ;-)

gebart:
@kb2ma 
>That leads to an interesting policy question. I thought about leaving stdint.h in gcoap.h, but both sock/udp.h and nanocoap.h already include it. So, it just seemed like clutter to keep it in gcoap.h, since it will certainly depend on those other two headers for the foreseeable future. Does that sway your guidance?

If sock/udp.h and nanocoap.h belonged to the same module as gcoap.h I would agree, but I don't think that it is robust to depend on other modules including other headers for us, even though the stdint.h header will be included by many API headers in RIOT.

PR #8856, opened 2018-03-30
dist/tools/openocd: add the possibility to do a reset before halt in openocd debug command

aabadie:
### Contribution description

This PR tries to address this [comment](https://github.com/RIOT-OS/RIOT/pull/8790#discussion_r177503888) by using an environment variable to trigger a reset before halt in the openocd debug command.

It's required for the b-l072z-lrwan1 board (at least), so the config of this board is updated accordingly.

### Issues/PRs references

Addresses the same issue than #8790 but in a cleaner way.

### DIFF
--- a/boards/b-l072z-lrwan1/Makefile.include
+++ b/boards/b-l072z-lrwan1/Makefile.include
@@ -12,5 +12,8 @@ include $(RIOTMAKE)/tools/serial.inc.mk
 export DEBUG_ADAPTER ?= stlink
 export STLINK_VERSION ?= 2-1
 
+# call a 'reset halt' command before starting the debugger
+export OPENOCD_DBG_START_CMD = -c 'reset halt'
+
 # this board uses openocd
 include $(RIOTMAKE)/tools/openocd.inc.mk
--- a/dist/tools/openocd/openocd.sh
+++ b/dist/tools/openocd/openocd.sh
@@ -83,6 +83,10 @@
 # Debugger flags, will be passed to sh -c, remember to escape any quotation signs.
 # Use ${DBG_DEFAULT_FLAGS} to insert the default flags anywhere in the string
 : ${DBG_FLAGS:=${DBG_DEFAULT_FLAGS} ${DBG_EXTRA_FLAGS}}
+# Initial target state when using debug, by default a 'halt' request is sent to
+# the target when starting a debug session. 'reset halt' can also be used
+# depending on the type of target.
+: ${OPENOCD_DBG_START_CMD:=-c 'halt'}
 # This is an optional offset to the base address that can be used to flash an
 # image in a different location than it is linked at. This feature can be useful
 # when flashing images for firmware swapping/remapping boot loaders.
@@ -201,7 +205,7 @@ do_debug() {
             -c 'gdb_port ${GDB_PORT}' \
             -c 'init' \
             -c 'targets' \
-            -c 'halt' \
+            ${OPENOCD_DBG_START_CMD} \
             -l /dev/null & \
             echo \$! > $OCD_PIDFILE" &
     # Export to be able to access these from the sh -c command lines, may be


aabadie:
Maybe @gebart or @kYc0o could help here.

boards/b-l072z-lrwan1/Makefile.include
```diff
@@ -12,5 +12,8 @@ include $(RIOTMAKE)/tools/serial.inc.mk
 export DEBUG_ADAPTER ?= stlink
 export STLINK_VERSION ?= 2-1
 
+# do a reset before halt when starting debugger
+export PRE_DEBUG_RESET_HALT ?= 1
```
kYc0o: I think the variable should be declared somewhere more general, then set it to one here. Maybe makefiles/vars.inc.mk ?

dist/tools/openocd/openocd.sh
```diff
@@ -178,6 +178,13 @@ do_flash() {
 do_debug() {
     test_config
     test_elffile
+
+    # Configure halt command: prepend a reset if required by the board config
+    PRE_RESET_HALT=''
+    if [ -n "${PRE_DEBUG_RESET_HALT}" ]; then
+        PRE_RESET_HALT='reset'
+    fi
+
```
kYc0o: I think instead of doing this you can evaluate the variable directly on the code, since it's used only there. Each command anyways gets executed separately.

boards/b-l072z-lrwan1/Makefile.include
kYc0o: Well I guess this anyways will be refactored for all the ST boards including a ST-Link debugger, so it's fine for now.

aabadie: > this anyways it will be refactored for all the ST boards embedding a ST-Link debugger

Are you referring to #8475 ?

dist/tools/openocd/openocd.sh
aabadie: I prefer keeping it this way. The command prefix is defined before, so there's no inline bash code added when the openocd command is built.

aabadie: But in this case it has to be defined globally and that's exactly what I wanted to avoid.

aabadie: By the way, I changed the previous code by using a local variable. Still works.

kYc0o: > But in this case it has to be defined globally and that's exactly what I wanted to avoid.

All the variables that are used in this script are global, they are available because they're exported since this script is out of the scope of the Makefiles. Thus, there's no real concept of "global" variable here, all of them are.

dist/tools/openocd/openocd.sh
```diff
@@ -178,6 +178,13 @@ do_flash() {
 do_debug() {
     test_config
     test_elffile
+
+    # Configure halt command: prepend a reset if required by the board config
+    local _pre_reset_halt=''
```
aabadie: @kYc0o, that's why I changed my code that now update a local variable by checking if the global variable is set.

aabadie: By the way, this strategy was inspired by the use of PRE_FLASH_CHECK_SCRIPT. The idea is exactly the same.

kYc0o: > By the way, this strategy was inspired by the use of PRE_FLASH_CHECK_SCRIPT. The idea is exactly the same.

That's a condition to execute a whole script and analysing its output, not just setting a variable. What I'm trying to avoid here is the declaration of variables and conditions which are IMHO not necessary, and by simply declaring the variable in makefiles/vars.inc.mk and setting it in your board Makefile it's enough. It's even more flexible since for other boards (which we might not be aware of yet) other commands might be necessary, thus you can just set this variable for your needs e.g. `PRE_DEBUG_RESET_HALT += reset halt reset` if you need to reset twice or whatever.

aabadie: Looks like we disagree

dist/tools/openocd/openocd.sh
```diff
@@ -178,6 +178,13 @@ do_flash() {
...
+    if [ -n "${PRE_DEBUG_RESET_HALT}" ]; then
```
cladmi: For me, the `PRE` here is wrong, because it changes the command from `halt` to `reset halt` it's not a previous command. Also, introducing a new really specific variable only for this specific way of handling the case seems bad to me.

cladmi: I would prefer the following

    # A meaningful comment on what this is :p
    : ${OPENOCD_EXTRA_INIT:=-c 'halt'}

And in the given board set

    export OPENOCD_DEBUG_HALT_CMD = -c 'reset halt'

aabadie:
@cladmi, I re-read the [openocd commands documentation](http://openocd.org/doc/html/General-Commands.html).
You are right that`halt` and `reset halt` are 2 different commands: the former just halt the target (which doesn't work with stm32l0 when used with some `reset_options`) while the latter performs a reset and then a halt (this is why I used PRE prefix).
Now I'm wondering if `halt` should just be replaced by `reset halt` like I did initially since `reset halt` seems more robust.

aabadie:
@cladmi @kYc0o I updated the code following your suggestions. Let's move forward.

dist/tools/openocd/openocd.sh
```diff
@@ -97,6 +97,10 @@
 # the file (default).
 # Valid values: elf, hex, s19, bin (see OpenOCD manual for more information)
 : ${IMAGE_TYPE:=}
+# Initial target state when using debug, by default a 'halt' request is sent to
+# the target when starting a debug session. 'reset halt' can also be used
+# depending on the type of target.
+: ${OPENOCD_EXTRA_DEBUG_INIT_CMD:=-c 'halt'}
```
cladmi: Some naming remarks, in that case it is not an `EXTRA` command like the other one as it's mandatory, so this `EXTRA` part should be removed.
To match your comment, maybe put more `START_CMD` as the name `INIT` means something with openocd (`reset init`, `init`)

Also I would prefer it defined along the `DBG` variables, but that's personal taste.

boards/b-l072z-lrwan1/Makefile.include
```diff
@@ -12,5 +12,8 @@ include $(RIOTMAKE)/tools/serial.inc.mk
 export DEBUG_ADAPTER ?= stlink
 export STLINK_VERSION ?= 2-1
 
+# do a reset before halt when starting debugger
```
cladmi: Same here, it is not a `reset` before `halt` but a `reset halt` openocd command.
http://openocd.org/doc/html/General-Commands.html#resetcommand

aabadie: Indeed "The optional parameter specifies what should happen after the reset" so there's a reset before halt (halt a an optional parameter) :)

aabadie:
Renamed and moved the variable as suggested by @cladmi

cladmi:
@aabadie Reviewing this PR made me find something you will like :) I will do a PR

aabadie:
> For the fact that it's working I trust you.

It does work :)

> Please squash.

Done

kYc0o:
I'd like to change the title though, it doesn't really reflect what it was done here.

aabadie:
> I'd like to change the title though, it doesn't really reflect what it was done here.

Again, it does, see this [comment](https://github.com/RIOT-OS/RIOT/pull/8856#discussion_r182455759).

kYc0o:
IMHO you're not adding the possibility to make a reset before halt, but rather you add a different command called `reset halt`. So maybe something like "openocd: replace `halt` command by `OPENOCD_DBG_START_CMD`", or something like that. Also, I'd say you should squash your commits into one, to allow a proper bisect if needed.

aabadie:
> Again, it does, see this comment.

Have you read the documentation ? I can put what I wrote there (again):

> "The optional parameter specifies what should happen after the reset"

so there's a reset before halt since halt is the optional parameter when using "reset halt".

> Also, I'd say you should squash your commits into one, to allow a proper bisect if needed.

I don't understand the purpose of your comment. The first commit updates the openocd script by introducing a parametrizable variable. The second commit set this variable to what works with the b-l072z-lrwan1 board. The 2 commits touch 2 different things (tools and boards) and this is very common practice into RIOT.

kYc0o:
Well, I still don't agree about the title (I had problems to recognise what actually PRs are doing by the title, even merging something that was already PRed before just because the title was misleading), but let's go. About the commits is ok because they are in the good order, I just want to enforce working commits because I also recently ran in a big hassle doing bisect.

PR #8861, opened 2018-04-01
cpu: lpc1768: provide periph_pm

basilfx:
### Contribution description

This provides power management for the LPC1768. It's based on [this](http://www.po-star.com/public/uploads/20120319123122_141.pdf) application note.

Tested using #8848 on the Seeeduino Arch-Pro. Although I could only measure milli-amps, I do see a difference when forcing different power modes. 

### Issues/PRs references

None

### DIFF
--- a/cpu/lpc1768/periph/pm.c
+++ b/cpu/lpc1768/periph/pm.c
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2018 Bas Stottelaar <basstottelaar@gmail.com>
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     cpu_lpc1768
+ * @ingroup     drivers_periph_pm
+ * @{
+ *
+ * @file
+ * @brief       Implementation of the power management peripheral
+ *
+ * @author      Bas Stottelaar <basstottelaar@gmail.com>
+ * @}
+ */
+
+#include "periph/pm.h"
+
+void pm_set(unsigned mode)
+{
+    switch (mode) {
+        case 0:
+            /* enter sleep mode */
+            LPC_SC->PCON = 0x00;
+            cortexm_sleep(0);
+            break;
+        case 1:
+            /* enter deep sleep mode */
+            LPC_SC->PCON = 0x00;
+            cortexm_sleep(1);
+            break;
+        case 2:
+            /* enter power down mode */
+            LPC_SC->PCON = 0x01;
+            cortexm_sleep(1);
+            break;
+    }
+}
+
+void pm_off(void)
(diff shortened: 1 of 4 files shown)


basilfx:
Decided to modify the board to measure power on the 3V3 rail. It's still connected to additional IC's, but it should give an idea.

power

cpu/lpc1768/periph/pm.c
```diff
@@ -0,0 +1,49 @@
...
+            /* enter power down mode */
+            LPC_SC->PCON = 0x01;
+            cortexm_sleep(1);
+            break;
+        case 1:
+            /* enter deep sleep mode */
+            LPC_SC->PCON = 0x00;
+            cortexm_sleep(1);
+            break;
+        case 2:
```
kaspar030: Should this be the default case, or is there a use-case for blocking mode 2?

basilfx: The way I [interpret](https://github.com/RIOT-OS/RIOT/blob/master/sys/include/pm_layered.h#L21) `pm_layered.h` is as follows:

* Mode 0 to `PM_NUM_MODES - 1` apply power management
* Mode `PM_NUM_MODES` (the implicit one) doesn't.

That's what I have implemented here (`PM_NUM_MODES` is 3).

kaspar030: OK!

It's just that the mode ```PM_NUM_MODES``` sets the CPU to idle on the other cortex-m. So I was wondering if ```LPC_SC->PCON = 0x00;``` means "no power management other than CPU idle", in which case I suggest setting ```PM_NUM_MODES=2```.

basilfx: `LPC_SC->PCON = 0x00;` is needed to ensure that it doesn't erroneously enter deep sleep on `__WFI()` (just idle).

So I think I need to change `PM_NUM_MODES` to 2, so that the implicit power mode is idle (`__WFI()`), right?

kaspar030: Yes!

basilfx: OK, thanks! I'll create another PR to fix this.

basilfx: Ok, thanks! I'll create another PR to fix this.

Just one more question: should I then add the default case, or `case 2` in combination with an `assert(mode <= PM_NUM_MODES)`?

IMHO `mode > PM_NUM_MODES` is wrong.

kaspar030: > IMHO mode > PM_NUM_MODES is wrong.

True, but we should rather catch that in pm_layered.c instead of addong those asserts to every pm_set implementation. IMO the default case will cause the least maintanance in the future. ;)

basilfx: Created #8874 to address this.

PR #5734, opened 2016-08-05
drivers/l1-nvram-eeprom: stm32l1 internal EEPROM driver

Cr0s:
Hello everyone,
I'm glad to introduce a simple driver for the STM32L1 Data EEPROM based on the RIOT's NVRAM interface.

Tested on my STM32L151CCU6-based device with testing app located in tests/driver_l1_nvram_eeprom

PR #8862, opened 2018-04-01
drivers/periph: add basic API to read/write data to EEPROM

aabadie:
### Contribution description

This PR adds:
- a basic API for reading/writing data to the internal EEPROM present on certain CPUs
- support for this to the stm32l0 and stm32l1 CPU families
- a basic test application with a shell

The test application has been tested on b-l072z-lrwan1 and nucleo-l152 boards.

Since reading and writing data to EEPROM has a very similar workflow than the flashpage implementation, the base eeprom functions are implemented in the same `flashpage.c` file. This allows for example to reuse the lock/unlock functions.
The problem is that it makes this feature dependent from the flashpage feature. Any advice on a better approach will be welcome.

Maybe this PR will interest @kYc0o, @vincent-d or @haukepetersen 

### Issues/PRs references

none

### DIFF
--- a/cpu/stm32_common/periph/flashpage.c
+++ b/cpu/stm32_common/periph/flashpage.c
@@ -58,6 +58,11 @@ static void _unlock(void)
         KEY_REG = FLASH_KEY1;
         KEY_REG = FLASH_KEY2;
     }
+}
+
+static void _unlock_flash(void)
+{
+    _unlock();
 
 #if defined(CPU_FAM_STM32L0) || defined(CPU_FAM_STM32L1)
     DEBUG("[flashpage] unlocking the flash program memory\n");
@@ -90,7 +95,7 @@ static void _erase_page(void *page_addr)
 #endif
 
    /* unlock the flash module */
-    _unlock();
+    _unlock_flash();
 
     /* make sure no flash operation is ongoing */
     DEBUG("[flashpage] erase: waiting for any operation to finish\n");
@@ -153,7 +158,7 @@ void flashpage_write_raw(void *target_addr, const void *data, size_t len)
 #endif
 
     DEBUG("[flashpage_raw] unlocking the flash module\n");
-    _unlock();
+    _unlock_flash();
 
     DEBUG("[flashpage] write: now writing the data\n");
 #if !(defined(CPU_FAM_STM32L0) || defined(CPU_FAM_STM32L1))
@@ -201,3 +206,19 @@ void flashpage_write(int page, const void *data)
         flashpage_write_raw(page_addr, data, FLASHPAGE_SIZE);
     }
 }
+
+#ifdef MODULE_PERIPH_EEPROM
+uint8_t eeprom_read_byte(uint32_t pos)
+{
+    DEBUG("Reading data from EEPROM at pos %lu\n", pos);
+    return *(uint8_t *)(EEPROM_START_ADDR + pos);
+}
+
+void eeprom_write_byte(uint32_t pos, uint8_t data)
+{
+    DEBUG("Writing data '%c' to EEPROM at pos %lu\n", data, pos);
+    _unlock();
+    *(uint8_t *)(EEPROM_START_ADDR + pos) = data;
+    _lock();
+}
+#endif
--- a/drivers/include/periph/eeprom.h
+++ b/drivers/include/periph/eeprom.h
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2018 Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_periph_eeprom EEPROM driver
+ * @ingroup     drivers_periph
+ * @brief       Low-level EEPROM interface
+ *
+ * @{
+ * @file
+ * @brief       Low-level eeprom driver interface
+ *
+ * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
+ *
+ */
+
+#ifndef PERIPH_EEPROM_H
+#define PERIPH_EEPROM_H
+
+#include <stdint.h>
+
+#include "periph_cpu.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef EEPROM_SIZE
+#error "periph/eeprom: EEPROM_SIZE is not defined"
+#endif
+
+#ifndef EEPROM_START_ADDR
+#error "periph/eeprom: EEPROM_START_ADDR is not defined"
+#endif
+
+/**
+ * @brief   Read a byte at the given position in eeprom
+ *
+ * @param[in]  pos      position to read
+ *
+ * @return the read byte
+ */
+uint8_t eeprom_read_byte(uint32_t pos);
+
+/**
+ * @brief   Read @p len bytes from the given position
+ *
+ * @param[in]  addr     start position in eeprom
+ * @param[out] data     byte address to write to
+ * @param[in]  len      the number of bytes to read
+ */
+void eeprom_read(uint32_t pos, void *data, uint8_t len);
+
+/**
+ * @brief   Write a byte at the given position
+ *
+ * @param[in] pos       position to write
+ * @param[in] data      byte address to write to
+ */
+void eeprom_write_byte(uint32_t pos, uint8_t data);
+
+/**
+ * @brief   Write @p len bytes at the given position
+ *
+ * @param[in] pos       start position in eeprom
+ * @param[in] data      byte address to write to
+ * @param[in] len       the number of bytes to read
+ */
+void eeprom_write(uint32_t pos, const void *data, uint8_t len);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* PERIPH_EEPROM_H */
+/** @} */
--- a/drivers/periph_common/eeprom.c
+++ b/drivers/periph_common/eeprom.c
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2018 Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers
+ * @{
+ *
+ * @file
+ * @brief       Common eeprom functions implmentation
+ *
+ * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
+ *
+ * @}
+ */
+
+#include <string.h>
+#include "cpu.h"
+#include "assert.h"
+
+/* guard this file, must be done before including periph/eeprom.h */
+#if defined(EEPROM_SIZE) && defined(EEPROM_START_ADDR)
+
+#include "periph/eeprom.h"
+
+void eeprom_read(uint32_t pos, void *data, uint8_t len)
+{
+    assert(pos + len < EEPROM_SIZE);
+
+    while(len--) {
+        *(uint8_t *)data++ = eeprom_read_byte(pos++);
+    }
+}
+
+void eeprom_write(uint32_t pos, const void *data, uint8_t len)
+{
+    assert(pos + len < EEPROM_SIZE);
+
+    while(len--) {
+        eeprom_write_byte(pos++, *(uint8_t *)data++);
+    }
+}
+
+#endif
--- a/tests/periph_eeprom/Makefile
+++ b/tests/periph_eeprom/Makefile
@@ -1,0 +1,8 @@
+BOARD ?= b-l072z-lrwan1
+include ../Makefile.tests_common
+
+FEATURES_REQUIRED += periph_eeprom
+
+USEMODULE += shell
+
+include $(RIOTBASE)/Makefile.include
--- a/tests/periph_eeprom/main.c
+++ b/tests/periph_eeprom/main.c
@@ -1,0 +1,108 @@
+/*
+ * Copyright (C) 2018 Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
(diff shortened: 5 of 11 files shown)


drivers/include/periph/eeprom.h
```diff
@@ -0,0 +1,81 @@
...
+uint8_t eeprom_read_byte(uint32_t pos);
+
+/**
+ * @brief   Read @p len bytes from the given position
+ *
+ * @param[in]  pos      start position in eeprom
+ * @param[out] data     byte address to write to
+ * @param[in]  len      the number of bytes to read
+ */
+void eeprom_read(uint32_t pos, uint8_t *data, uint8_t len);
```
kaspar030: Are eeproms always <256b? Otherwise, I suggest using size_t for read/write length.

kaspar030:
Would you mind implementing this on a very different platform, e.g., for some atmega? That usually shows whether the interface makes sense.

aabadie:
> Would you mind implementing this on a very different platform, e.g., for some atmega? 

Looking at AVR Atmega1284 datasheet, this seems rather simple. I'll give it a try (already saw things that might be changed...)

aabadie:
@kaspar030 comments addressed + added implementation for Atmega CPUs. Tested on nucleo-l152 and Arduino Mega2560.

aabadie:
Do you think this one could be part of the release @kaspar030 ?

cpu/stm32_common/periph/flashpage.c
```diff
@@ -201,3 +206,24 @@ void flashpage_write(int page, const void *data)
         flashpage_write_raw(page_addr, data, FLASHPAGE_SIZE);
     }
 }
+
+#ifdef MODULE_PERIPH_EEPROM
```
smlng: I think this files is missing the `#include "periph/eeprom.h"`? But would also suggest to not add this to flashpage driver but rather have a separate file.

aabadie: > I think this files is missing the #include "periph/eeprom.h"?

Agreed

> But would also suggest to not add this to flashpage driver but rather have a separate file.

How do you share code between different periph ? Here I'm reusing the lock/unlock functions from flashpage periph, since they are the same.

aabadie: @smlng, I moved the _lock/_unlock into a shared file (between flashpage and eeprom periph drivers) and provided the eeprom peripheral driver into a separate file: now there's no dependency between flashpage and eeprom (for STMs).

smlng:
@aabadie are you aware of #5734, not sure if there is anything to take over?

aabadie:
> are you aware of #5734

Yes, but I admit I missed it when I started to work on this one.

> not sure if there is anything to take over?

I don't know. Looking at the proposed design in #5734, it seems very (too) specific so it cannot be generalized to other CPU families. I also think that it's not following RIOT's peripheral drivers designs.

aabadie:
The good thing with #5734 is the better error management. The API is also different but less POSIX I think.

aabadie:
Now that I'm thinking at it again, the interesting concept in #5734 is the use of the nvram API. I think it would be easy to adapt #8862 to nvram (following what is already done in #5734)

aabadie:
@kaspar030, @smlng do you have other comments here ?

cpu/atmega_common/periph/eeprom.c
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+#include <stdint.h>
+
+#include "cpu.h"
+#include "periph/eeprom.h"
+
+uint8_t eeprom_read_byte(uint32_t pos)
+{
+    /* Wait for completion of previous write */
```
smlng: mhm, I thought we are reading? So maybe the comment should better state generically:
```
Wait for completion of previous operation
```

cpu/stm32_common/periph/eeprom.c
```diff
@@ -0,0 +1,48 @@
...
+extern void _lock(void);
+extern void _unlock(void);
+
+#ifndef EEPROM_START_ADDR
+#error "periph/eeprom: EEPROM_START_ADDR is not defined"
+#endif
+
+uint8_t eeprom_read_byte(uint32_t pos)
+{
+    DEBUG("Reading data from EEPROM at pos %lu\n", pos);
```
smlng: maybe also add `assert(pos < EEPROM_SIZE)`?

cpu/stm32_common/periph/eeprom.c
```diff
@@ -0,0 +1,48 @@
...
+    return *(uint8_t *)(EEPROM_START_ADDR + pos);
+}
+
+void eeprom_write_byte(uint32_t pos, uint8_t data)
+{
+    DEBUG("Writing data '%c' to EEPROM at pos %lu\n", data, pos);
```
smlng: same here, adding an `assert` might be a good idea?

cpu/stm32_common/periph/eeprom.c
```diff
@@ -0,0 +1,48 @@
...
+extern void _unlock(void);
+
+#ifndef EEPROM_START_ADDR
+#error "periph/eeprom: EEPROM_START_ADDR is not defined"
+#endif
+
+uint8_t eeprom_read_byte(uint32_t pos)
+{
+    DEBUG("Reading data from EEPROM at pos %lu\n", pos);
+    return *(uint8_t *)(EEPROM_START_ADDR + pos);
```
smlng: question: no locking while reading needed?

drivers/include/periph/eeprom.h
```diff
@@ -0,0 +1,78 @@
...
+uint8_t eeprom_read_byte(uint32_t pos);
+
+/**
+ * @brief   Read @p len bytes from the given position
+ *
+ * @param[in]  pos      start position in eeprom
+ * @param[out] data     byte array to write to
+ * @param[in]  len      the number of bytes read
+ */
+void eeprom_read(uint32_t pos, uint8_t *data, size_t len);
```
smlng: IMHO these functions should return `int` for the number bytes actually read or written, or an error otherwise, applies to write functions below, too.

smlng: this would also be more in line with other read/write functions in RIOT and Linux, in general.

drivers/periph_common/eeprom.c
```diff
@@ -0,0 +1,48 @@
...
+#include <string.h>
+#include "cpu.h"
+#include "assert.h"
+
+/* guard this file, must be done before including periph/eeprom.h */
+#if defined(EEPROM_SIZE)
+
+#include "periph/eeprom.h"
+
+void eeprom_read(uint32_t pos, uint8_t *data, size_t len)
```
smlng: following my comment above having a proper return value, this could be something like:

```
if ((pos < EEPROM_SIZE) && (len < EEPROM_SIZE)) {
    int rc;
    for (rc = 0; (rc < len) && (pos + rc < EEPROM_SIZE); ++rc) {
        data[rc] = eeprom_read_byte(pos + rc);
    } 
    return rc;
}
return (-1);
```

applies for write function below too

kaspar030: Not the most readable. :smirk: 

Maybe just copy "len" and return it?

kaspar030:
@smlng I think the API is fine as is (very slim, not much error checking). It would be very easy to add extensive error checking on top. It is difficult to remove it when the API dictates it.

aabadie:
@smlng, I added the `assert` as you suggested and changed the api of `eeprom_read` and `eeprom_write` functions so that they returns the number of bytes read/written (there's no complex error checking though, it's just a copy of `len`).

Let me know if it's ok for you now

aabadie:
@smlng, @kaspar030, comments addressed :)

cpu/atmega_common/periph/eeprom.c
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @file
+ * @brief       Low-level EEPROM driver implementation for ATmega family
+ *
+ * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
+ * @}
+ */
+
+#include <stdint.h>
+
```
mali: include <assert.h> is missing

mali:
Tested with Uno/Duemilanove :+1:

aabadie:
@smlng @kaspar030, let me know when/if I can squash this one. Thanks

drivers/periph_common/eeprom.c
```diff
@@ -0,0 +1,56 @@
...
+/* guard this file, must be done before including periph/eeprom.h */
+#if defined(EEPROM_SIZE)
+
+#include "periph/eeprom.h"
+
+size_t eeprom_read(uint32_t pos, uint8_t *data, size_t len)
+{
+    assert(pos + len < EEPROM_SIZE);
+
+    size_t ret = len;
```
smlng: though it doesn't safe binary size or anything, but IMHO still a bit more readable:

```
    for (size_t i = 0; i < len; ++i) {
        data[i] = eeprom_read_byte(pos++);
    }
    return len;
```

drivers/periph_common/eeprom.c
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    return ret;
+}
+
+size_t eeprom_write(uint32_t pos, const uint8_t *data, size_t len)
+{
+    assert(pos + len < EEPROM_SIZE);
+
+    size_t ret = len;
```
smlng: same here:

```
    for (size_t i = 0; i < len; ++i) {
        eeprom_write_byte(pos++, data[i]);
    }
    return len;
```

tests/periph_eeprom/main.c
```diff
@@ -0,0 +1,110 @@
...
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+
+#include "shell.h"
+
+#include "periph/eeprom.h"
+
+static char buffer[42];
```
smlng: buffer size should be a marco/define

tests/periph_eeprom/main.c
```diff
@@ -0,0 +1,110 @@
...
+{
+    if (argc < 3) {
+        printf("usage: %s <pos> <count>\n", argv[0]);
+        return 1;
+    }
+
+    uint32_t pos = atoi(argv[1]);
+    uint8_t count = atoi(argv[2]);
+
+    if (!count) {
```
smlng: also check wether it fits the buffer size

tests/periph_eeprom/main.c
```diff
@@ -0,0 +1,110 @@
...
+        puts("Count should be greater than 0");
+        return 1;
+    }
+
+    if (pos + count >= EEPROM_SIZE) {
+        puts("Failed: cannot read out of eeprom bounds");
+        return 1;
+    }
+
+    size_t ret = eeprom_read(pos, (uint8_t *)buffer, count);
```
smlng: unsafe buffer handling, might overflow

aabadie:
@smlng, fixed your new comments and tested on  b-l072z-lrwan1 board. Still works well.

tests/periph_eeprom/main.c
```diff
@@ -0,0 +1,119 @@
...
+    buffer[count] = '\0';
```
smlng: if you want to do this you must check for `count >= BUFFER_SIZE` above.

aabadie: nope, the buffer has an extra byte, see [here](https://github.com/RIOT-OS/RIOT/pull/8862/files#diff-0ec3102ff4d926d871687a7b42380deaR33)

smlng: oops, oversaw that 😬

drivers/periph_common/eeprom.c
```diff
@@ -0,0 +1,52 @@
...
+/* guard this file, must be done before including periph/eeprom.h */
+#if defined(EEPROM_SIZE)
+
+#include "periph/eeprom.h"
+
+size_t eeprom_read(uint32_t pos, uint8_t *data, size_t len)
+{
+    assert(pos + len < EEPROM_SIZE);
+
+    for (size_t i = 0; i < len; ++i) {
```
kaspar030: here and elsewhere: please use ```i++```.

aabadie: Done and hope this is the last one ;)

smlng: @kaspar030 why, if it makes a difference I always thought `++i` is better than `i++`

kaspar030: 1. consistency. In RIOT, 'for.*i++' vs 'for.*++i' is about four to one.
2. In C, it doesn't make a difference. BUT "for(...;i++)' is considered ideomatic C, and thus more readable.

aabadie: Ok, I'm fine with that. Now it should be all good here, right ?

cpu/atmega_common/periph/eeprom.c
```diff
@@ -0,0 +1,60 @@
...
+#include "cpu.h"
+#include "periph/eeprom.h"
+
+uint8_t eeprom_read_byte(uint32_t pos)
+{
+    assert(pos < EEPROM_SIZE);
+
+    /* Wait for completion of previous operation */
+    while (EECR & (1 << EEPE));
```
kaspar030: here and below: needs to be changed to ```while() {}```.

tests/periph_eeprom/Makefile
```diff
@@ -0,0 +1,9 @@
+BOARD ?= b-l072z-lrwan1
```
kaspar030: why this board? Maybe chose one that is more common?

aabadie: This is my dev board actually and it's also provided by iotlab ;)

cpu/atmega_common/periph/eeprom.c
aabadie: changed

aabadie:
and squashed

aabadie:
Yeah! Thanks @smlng @kaspar030 @mali for reviewing and testing.

PR #8863, opened 2018-04-02
 tests/fmt_print: improved

Josar:
tests/fmt_print: improved

Added test for unsigned int 16/32/64 to decimal string conversion.

Besides that the 64bit conversion seems to make problems on Native.
See the README.

### DIFF
--- a/sys/fmt/fmt.c
+++ b/sys/fmt/fmt.c
@@ -221,6 +221,7 @@ size_t fmt_u32_dec(char *out, uint32_t val)
 
     if (out) {
         char *ptr = out + len;
+        *ptr = '\0';
         do {
             *--ptr = (val % 10) + '0';
         } while ((val /= 10));
--- a/tests/fmt_print/README.md
+++ b/tests/fmt_print/README.md
@@ -1,0 +1,32 @@
+Expected result
+===============
+This application Test the formatter.
+
+Example:
+
+```
+RIOT native interrupts/signals initialized.
+LED_RED_OFF
+LED_GREEN_ON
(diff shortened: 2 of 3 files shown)


sys/fmt/fmt.c
```diff
@@ -221,6 +221,7 @@ size_t fmt_u32_dec(char *out, uint32_t val)
 
     if (out) {
         char *ptr = out + len;
+        *ptr = '\0';
```
kaspar030: this has to go, ```fmt_*()``` don't null-terminate.

kaspar030:
There are unittests for fmt, maybe improve them? This extra test application is fir the ```print_*()``` family of functions, as their output cannot be tested by the unittest framework.

btw, the unittests also fail on native, I'm investigating.

I'll close this, as doing the tests at two places doesn't make sense. If you feel the unittests need improvement, please open another PR.

kaspar030:
> btw, the unittests also fail on native, I'm investigating.

The null-termination change breaks ```fmt_u64_dec()```.

Josar:
Unittest runs on native and jiminy. My Problem seems to be vanished.

Only thing which then is maybe missleading is the description.
https://github.com/RIOT-OS/RIOT/blob/c23d85ab3ab40bbff3245f6ca08e760fe7247fda/sys/include/fmt.h#L140

As for my understanding a string is terminated.

kaspar030:
> As for my understanding a string is terminated.

I think that must be documented better. All fmt functions only write the converted characters without null termination.

PR #6472, opened 2017-01-24
cpu/lpc1768: add gpio peripheral (+ board adaptions)

basilfx:
I needed GPIO on the LPC1768 (more specific, on the Seeeduino Arch Pro), but I realized it wasn't there. Since I need it for a demo tomorrow, I decided to add it :-)

So far output mode works, but that's all I tested. Have to dig into interrupts, which seem to be a bit more complex.

Adapted `mbed_lpc1768` and `seeeduino-arch_pro` boards, including SAUL support for GPIO. I realize that they overlap a lot, but the boards are not quite similar.

WIP, needs a lot of cleanup.

PR #8866, opened 2018-04-03
boards: seeeduino_arch-pro: use hex file.

basilfx:
### Contribution description

LPC1768 needs a checksum inserted, which is not performed by OpenOCD. I use `lpc_checksum` for that, which requires a HEX file instead of an ELF or BIN file.

This was working before, but somewhere it broke. This PR ensures that the HEX file is used when inserting the checksum.

### Issues/PRs references

#6472


basilfx:
> It would be great to find the reason.

I had an old RIOT clone (april 2017) that worked, and after rebasing it didn't. So I think that the OpenOCD refactoring (october 2017 IIRC) caused it.

PR #8869, opened 2018-04-03
tests: add on-hardware execution for selected tests

kYc0o:
### Contribution description

As promised, here is some additions of tests that can be performed at least in native, regardless of existing peripherals. It needs a more deeper check though but I'd say it's at least WIP.

### Issues/PRs references
#8801


kaspar030:
The samr21 test nodes are down. I suspect they got shut down by the electricity test at FU last week.
@kYc0o are you at the premises and can take a look at the raspis in emmanuels office?

kYc0o:
>  The samr21 test nodes are down. I suspect they got shut down by the electricity test at FU last week.
@kYc0o are you at the premises and can take a look at the raspis in emmanuels office?

Yup... anyways, I guess the tests are done for native anyways? Before you removed the label it has already found 3 or 4 errors, but I cannot check them anymore...

kYc0o:
Ok, I checked and actually there was no router there anymore. I reinstalled it, can you check if everything is ok?

kYc0o:
Strangely the failing tests on native work OK locally in my OS X, besides the pkg_fatfs_vfs test, which needs more setup and I'll remove from the list.

Can someone with Linux check those tests? Basically:

- [ ] pkg_u8g2
- [ ] pthread_barrier

kaspar030:
> @kaspar030 can you check consistency on timing when launching scripts and the board reacts?

??

kYc0o:
I mean in the Pi's which perform the tests. Now all native tests are passing, it's only for the samr21's, and at least for the first failure I suspect there's something to do with timing. The node might be ready before the test script starts, or the opposite, thus we'll have random failures like this one.

kaspar030:
> The node might be ready before the test script starts, or the opposite, thus we'll have random failures like this one.

Possible. The node just executes "make test". Can you test that locally (with samr21-xpro)?

kYc0o:
> Possible. The node just executes "make test". Can you test that locally (with samr21-xpro)?

I did it and it passes most of the tests, but some (long calculation, e.g. libfixmath) tests are more likely to have random failures. What I notice on the two runs we did with this PR is that also shell interaction tests can also be affected.

kaspar030:
> I did it and it passes most of the tests, but some (long calculation, e.g. libfixmath) tests are more likely to have random failures.

So ```make flash-only && make test``` works fine when run a couple of times in a loop? Can you try using edbg and picocom?

Long calculation shouldn't cause trouble unless the 5 minute timeout for the whole job is hit.
I also have the impression that this is some buffering problem: ```help\n> lp\ncommand not found...```

kaspar030:
> Seems legit, I'll try it. However I wonder why it didn't fail in the previous run.

See #8879.

kYc0o:
> See #8879.

On it.

kYc0o:
I skipped the problematic tests for now. I have another branch with those which I can submit whenever the testrunner problem (#8211 ) is solved.

kaspar030:
the rng test failed now and in the last run. maybe remove for now?

kYc0o:
Removed for all, changed to native. Squashed directly.

kaspar030:
hm, ps_schedstatistics and struct_tm_utility failed.

kYc0o:
Sorry, I forgot to rebase. Though it's strange the previous run didn't happen...

kYc0o:
Rebased.

kaspar030:
> Sorry, I forgot to rebase. Though it's strange the previous run didn't happen...

actually murdock rebases before running the tests, thus the problem is also in master. Seems like there's more flakyness with pexpect.

kYc0o:
> actually murdock rebases before running the tests, thus the problem is also in master. Seems like there's more flakyness with pexpect.

True, I forgot that fact. Then we should think about anything more resilient. 

```console
> help
help
help

days_in              Tells you the number of days in a month.
leap_year            Tells you if a supplied year is a leap year.
doomsday             Tells you the wday Doomsday of the supplied year.
day                  Tells you the day of the supplied date.
> days_in
days_in


> Timeout in expect script at "child.expect_exact('Usage: days_in <Month[1..12]>')" (tests/struct_tm_utility/tests/01-run.py:32)
```

This doesn't look like a synchronisation issue, isn't it?

kaspar030:
> This doesn't look like a synchronisation issue, isn't it?

Nope, looks more like the node isn't receiving whatever pexpect sends. The terminal has echoed it, though.

kYc0o:
It just happened again with ps_schedstatistics. No synchronisation issues anymore.

kaspar030:
> But it seems that the trouble is also some characters lost in the serial line.

I tried reproducing this. Manually sending the test jobs to one of the RasPis always succeeds, unless jobs are queued and thus run right after another. Maybe this is some kind of buffering error?

kaspar030:
I noticed that one node always failed, while the other always succeeded. Rebooting the node helped. So much for flakyness...

kaspar030:
This now succeeded two times, and I cannot reproduce the lost data anymore. So I assume any trouble is unrelated to the tests that this PR enables.

PR #8832, opened 2018-03-25
boards: stk3200: add support

basilfx:
### Contribution description

This PR add supports for the Silicon Labs STK3200, which I already PR'd before in #8295. The files are similar to the SLTB001a and STK3600, STK3700 and SLSTK3401a already supported. This is a Cortex M0+ based board, with a low-power memory LCD that is supported by U8g2.

I can provide remote access to the hardware, if needed.

### Issues/PRs references

#8585
#8520
#8630
#8295

PR #8873, opened 2018-04-03
pkg: gecko_sdk: update version

basilfx:
### Contribution description

This PR updates `gecko-sdk` to vendor version 2.2 (which includes emlib 5.4.0, [changelog](https://github.com/basilfx/RIOT-gecko-sdk/blob/master/dist/emlib/Changes_emlib.txt)). Most notable are the changes towards being MISRAC 2012 compatible and GCC 7.x fixes.

Unfortunately, this package updates requires updated EFM32 vendor headers. which is 99% of this PR.  Yes, most changes are version numbers, but they have renamed a few crucial defines. I therefore suggest to focus on a8e57ff only.

I cannot split this PR since the vendor headers need the updated `gecko-sdk`, and vice versa.

Tested `examples/default` on the stk3600, sltb001a and slstk3401a and they all work.

### Issues/PRs references

#8832, #8824 and #8802 need their vendor headers updated as well.

### DIFF
--- a/cpu/efm32/families/efm32gg/include/vendor/efm32gg990f1024.h
+++ b/cpu/efm32/families/efm32gg/include/vendor/efm32gg990f1024.h
@@ -2,10 +2,10 @@
  * @file efm32gg990f1024.h
  * @brief CMSIS Cortex-M Peripheral Access Layer Header File
  *        for EFM32GG990F1024
- * @version 5.3.3
+ * @version 5.4.0
  ******************************************************************************
  * # License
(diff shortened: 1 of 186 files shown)


basilfx:
Tested the default example for the `stk3600`, `sltb001a` and `slstk3401a` and they all work.

kYc0o:
Should we merge this before the referenced PRs?

basilfx:
Yes, I'll update the other PRs tonight (they should fail during build now).

PR #8878, opened 2018-04-04
pkg/ccn-lite:enable CS dump in shell command

PeterKietzmann:
### Contribution description

This PR adds a new feature to the CCN-lite shell command `ccnl_cont` which dumps all content items in the cache. Please note that without a version bump to the below referenced PR, this will result in some formating errors. A version bump will be PRed soon.

### Issues/PRs references

[https://github.com/cn-uofbasel/ccn-lite/pull/224](https://github.com/cn-uofbasel/ccn-lite/pull/224)

### DIFF
--- a/sys/shell/commands/sc_ccnl.c
+++ b/sys/shell/commands/sc_ccnl.c
@@ -30,7 +30,6 @@
 
 static unsigned char _int_buf[BUF_SIZE];
 
-static const char *_default_content = "Start the RIOT!";
 static unsigned char _out[CCNL_MAX_PACKET_SIZE];
 
 /* usage for open command */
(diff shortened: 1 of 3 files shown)


PeterKietzmann:
Yes, I had the same impression although I tried to orientate at the `ccnl_fib` command and had no better (not over-complex) solution in mind. Do you have a proposal how you liked the command better?

cgundogan:
I had an offline discussion with @mfrey and that's the outcome:

1.`ccnl_cs` is a better name for `ccnl_cont`.
2. remove the "default" content that is bound when just specifying a name
3. instead show the `usage` if only one parameter is provided

What's still necessary is an `rm` sub-command to delete CS entries. But that can be done in a follow-up PR.

@mfrey is that correct?

PeterKietzmann:
@cgundogan, @mfrey I like your idea. Will change the PR soon if no one objects.

mfrey:
> @mfrey is that correct?

Sounds good to me!

PeterKietzmann:
@mfrey, @cgundogan like this? Will adapt the README afterwards

PeterKietzmann:
Now I changed the README as well. If you 're fine with the "SQASHEME" commit I'd actually squash it so we can trigger CI.

sys/shell/commands/sc_ccnl.c
```diff
@@ -75,21 +74,24 @@ int _ccnl_open(int argc, char **argv)
...
             "%% %s /riot/peter/schmerzl RIOT\n",
-            argv, argv, argv);
+            argv, argv);
 }
 
 int _ccnl_content(int argc, char **argv)
 {
     if (argc < 2) {
+        ccnl_cs_dump(&ccnl_relay);
+        return -1;
```
cgundogan:
```c
return 0;
```

PeterKietzmann:
@cgundogan thanks for the review. I addressed your change request and immediately squashed. Furthermore, I changed the commit message slightly

cgundogan:
cpp check:
```
sys/shell/commands/sc_ccnl.c:112: error (uninitvar): Uninitialized variable: buf
```

PR #8879, opened 2018-04-04
tests: enable ps_schedstatistics and struct_tm_utility on-hardware CI tests

kaspar030:
### Contribution description

fixes & CI-enables ps_schedstatistics, struct_tm_utility

### Issues/PRs references

#8869

### DIFF
--- a/tests/ps_schedstatistics/Makefile
+++ b/tests/ps_schedstatistics/Makefile
@@ -11,6 +11,8 @@ USEMODULE += ps
 USEMODULE += schedstatistics
 USEMODULE += printf_float
 
+TEST_ON_CI_WHITELIST += all
+
 include $(RIOTBASE)/Makefile.include
 
 test:
--- a/tests/ps_schedstatistics/tests/01-run.py
+++ b/tests/ps_schedstatistics/tests/01-run.py
@@ -39,6 +39,8 @@ def _check_startup(child):
 
 
 def _check_help(child):
+    child.sendline('')
+    child.expect('>')
     child.sendline('help')
     child.expect_exact('Command              Description')
     child.expect_exact('---------------------------------------')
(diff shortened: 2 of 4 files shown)


kaspar030:
- force-pushed flake8 complaint fix

kYc0o:
I was trying to solve an issue now in local. Apparently the HIL attached is "faster" than the script, so your fix works because the node is ready before the test. Thus, waiting for a prompt after an enter works correctly. However, in my local test the node is "slower" thus waiting for a prompt doesn't work since the node comes after the enter is given.

Any ideas? I'll run it again to see if it succeeds after a third run.

kYc0o:
Ok I just tried something. Usually shell based tests will react to 'help' command, so I suggest to change the empty newline with such a command. By doing that in a for loop (just in case) at least 3 times works in my local setup.

@kaspar030 would you agree on that? I can push the modification if you want, for instance:

```diff
diff --git a/tests/rng/tests/01-run.py b/tests/rng/tests/01-run.py
index 3d4e86e5b..758a14df0 100755
--- a/tests/rng/tests/01-run.py
+++ b/tests/rng/tests/01-run.py
@@ -9,9 +9,15 @@
 import sys
 import os
 import re
+import pexpect
 
 
 def testfunc(child):
+    for x in range(0, 2):
+        child.sendline('help')
+        index = child.expect(['>', pexpect.TIMEOUT])
+        if index == 0:
+            break
     # RNG source
     child.sendline("source 0")
     child.sendline("seed 1337")
```

kYc0o:
> It is possible that your terminal starts so slow that it ignores the newline and eats the prompt before interacting with pexpect.

I think the problem is pyterm. When I send "enters" it happens nothing, while with other terminals it returns the prompt.

Anyways, I still have some tests that don't succeed with picoterm (see #8869), but I think I came with a solution...

PR #8883, opened 2018-04-05
sys/random: remove printf in xorshift

PeterKietzmann:
### Contribution description

I don't see any reason for this printf and find it more annoying than helpful. Furthermore, when used with native the node crashes at that position.


PeterKietzmann:
@gebart thanks for the fast feedback. Removed that include and immediately squashed.

PR #8886, opened 2018-04-05
xtimer: Add XTIMER_SHIFT = 6 automatic option

gebart:
### Contribution description

Just adding another step in the conditionals for XTIMER_SHIFT 6. Discovered that it was missing when trying to use a 15625 Hz xtimer instead of 1 MHz as an experiment.


gebart:
It has something to do with which version of cppcheck is running on Travis vs Murdock

PR #8543, opened 2018-02-09
periph_rtt: remove RTT_NUMOF guard and init from periph_common

vincent-d:
### Contribution description

This align header with other periph (i.e. removes `RTT_NUMOF` guard which is not needed anymore).

I also added a call to `rtt_init()` from `periph_init()` but I'm not sure if it will work on every platform correctly. For instance, mulle seems to need to call `rtt_init()` earlier in the boot process.
So I can still remove the second commit if we don't agree on it.

### Issues/PRs references

None

PR #8890, opened 2018-04-05
cpu: efm32: throw error on including rtc and rtt together.

basilfx:
### Contribution description

On the EFM32, RTT and RTC use the same hardware peripheral. It doesn't make sense to include both of them, especially if #8543 gets merged.

This PR will throw an error if you include both the RTT and RTC driver.

To test, create/modify an application which includes both, then verify that it will not build.

### Issues/PRs references

#8543

### DIFF
--- a/cpu/efm32/Makefile.features
+++ b/cpu/efm32/Makefile.features
@@ -2,6 +2,9 @@ FEATURES_PROVIDED += periph_cpuid
 FEATURES_PROVIDED += periph_flashpage
 FEATURES_PROVIDED += periph_pm
 
+FEATURES_CONFLICT += periph_rtc:periph_rtt
+FEATURES_CONFLICT_MSG += "On the EFM32, the RTC and RTT map to the same hardware peripheral."
+
 ifeq (1,$(EFM32_TNRG))
   FEATURES_PROVIDED += periph_hwrng
 endif


kaspar030:
Maybe just set ```FEATURES_CONFLICT += periph_rtc:periph_rtt```?

basilfx:
I didn't know of that.

However, it seems that it doesn't work in combination with `FEATURES_OPTIONAL`. For instance, `examples/default` includes RTC as optional, and settings `FEATURES_REQUIRED` to include RTT doesn't throw an error. If I check `riotbuild.h`, both are included.

basilfx:
@kaspar030 Any idea?

kaspar030:
> @kaspar030 Any idea?

I think we need to fix the feature conflict stuff... @cladmi maybe you could take a look?
@basilfx Is this issue very pressing for you? I propose we keep the PR open, and if feature conflicts are not fixed until beginning of next week, we merge this PR as workaround.

cladmi:
I think the problem is that the check is using `FEATURES_REQUIRED` and not `FEATURES_USED`. I will verify it would solve it.

cladmi:
Yes that's it, I will do a PR.

cladmi:
With https://github.com/RIOT-OS/RIOT/pull/8925 you will be able to use FEATURES_CONFLICT and FEATURES_CONFLICT_MSG to show a warning.

They are used for `stm32f4discovery` for reference https://github.com/RIOT-OS/RIOT/blob/master/boards/stm32f4discovery/Makefile.features#L19

basilfx:
I have updated/rebased this PR to use `FEATURES_CONFLICT`. It should work whenever #8925 gets merged.

Issue #8896, opened 2018-04-07
cpu/atmega_common: Context switching race condition


ZetaR60:
On the ATmega there is a race condition where instructions after AVR_CONTEXT_SWAP_TRIGGER are executed before the context swap actually occurs. In my testing it seems that the thread_yield_higher function call has enough delay that instructions after it are not executed before the swap. However, this may change under environmental conditions that change clock rate or propagation delays such as: clock frequency (used 8MHz), voltage (used 3.3V), parasitics attached to the pin (appears to not be significant), temperature, part variations, and so forth.

To reproduce: Run an instruction immediately after AVR_CONTEXT_SWAP_TRIGGER, and within another thread, that both set conflicting states. Test for which state was actually set first.

Expected results: Conditions changed in second thread occurs before any instructions placed after AVR_CONTEXT_SWAP_TRIGGER.

Actual results: At least one instruction after AVR_CONTEXT_SWAP_TRIGGER is executed before context swap occurs.

Versions:
- avr-gcc 7.2.0
- avr-libc 2.0.0
- mega-xplained board

I will be submitting tests/thread_race after I finish writing this.

PR #8897, opened 2018-04-07
tests/thread_race: add test for race conditions

ZetaR60:
This adds a test for race conditions in context switching, especially for AVR_CONTEXT_SWAP_TRIGGER on the ATmega.

Tests for problem in #8896

### DIFF
--- a/tests/thread_race/Makefile
+++ b/tests/thread_race/Makefile
@@ -1,0 +1,10 @@
+include ../Makefile.tests_common
+
+DISABLE_MODULE += auto_init
+
+TEST_ON_CI_WHITELIST += all
+
+include $(RIOTBASE)/Makefile.include
+
+test:
+	tests/01-run.py
--- a/tests/thread_race/main.c
+++ b/tests/thread_race/main.c
@@ -1,0 +1,133 @@
+/*
+ * Copyright (C) 2018 Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     test
+ * @{
+ *
+ * @file
+ * @brief       Test for race conditions in context switching
+ *
+ * May have false negatives.
+ *
+ * @author      Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * @}
+ */
+
+#include <stdio.h>
+
+#include "irq.h"
+#include "sched.h"
+#include "thread.h"
+
+char iqr_check_stack[THREAD_STACKSIZE_DEFAULT];
+
+static volatile uint8_t irq_occurred;
+
+static void *_thread_irq_check(void *arg)
+{
+    (void)arg;
+
+    while (1) {
+        irq_occurred = 1;
+
+        thread_sleep();
+    }
+
+    return NULL;
+}
+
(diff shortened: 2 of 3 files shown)


ZetaR60:
Fixed for #8904

tests/thread_race/main.c
```diff
@@ -0,0 +1,147 @@
...
+
+    irq_restore(old_state);
+}
+
+static void _spin(void)
+{
+    /* Volatile so it is not messed with by optimizations */
+    volatile uint8_t i;
+
+    for (i=0; i<255; i++);
```
kYc0o: Here and below (for the if-else statements), please follow coding conventions. You may also run uncrustify on this file to correct it automatically.

tests/thread_race/main.c
```diff
@@ -0,0 +1,147 @@
...
+}
+
+int main(void)
+{
+    puts("Context swap race condition test application\n");
```
kYc0o: Here and below, no need of `\n` in the `puts()` function.

tests/thread_race/main.c
```diff
@@ -0,0 +1,147 @@
...
+#include <avr/cpufunc.h>
+#include <avr/interrupt.h>
+#include <stdio.h>
+
+#include "irq.h"
+#include "sched.h"
+#include "thread.h"
+
+/* board.h for AVR_CONTEXT_SWAP_TRIGGER */
+#include "board.h"
```
kYc0o: Here and below, I don't like the non-generic nature on this tests, and why AVR should be treated in such special way... I see more convenient to have a generic test which works for all platforms without `#ifdef`s.

If this is due to #8896, please make this PR dependent on #8904, on which `AVR_CONTEXT_SWAP_TRIGGER` doesn't exist anymore.

tests/thread_race/main.c
```diff
@@ -0,0 +1,147 @@
...
+    _thread_wake_wo_yield(pid);
+
+#ifdef AVR_CONTEXT_SWAP_TRIGGER
+    AVR_CONTEXT_SWAP_TRIGGER;
+#else
+    thread_yield_higher();
+#endif
+
+    /* Delay so we are not testing for race conditions also */
+    _spin();
```
kYc0o: Can you explain a bit better this spin? If I understood correctly, this test is actually used to test race conditions isn't it?

tests/thread_race/main.c
ZetaR60: Sorry, still new to RIOT conventions. Uncrustified.

tests/thread_race/main.c
ZetaR60: Removed.

ZetaR60:
Squashed with requested changes incorporated.

ZetaR60:
@kYc0o Squashed. Thanks!

tests/thread_race/Makefile
```diff
@@ -0,0 +1,5 @@
+include ../Makefile.tests_common
+
+DISABLE_MODULE += auto_init
+
```
kYc0o: I forgot this part. I think this test should be run by the CI on hardware. Can you please add:

```makefile
TEST_ON_CI_WHITELIST += all
```

just here?

ZetaR60: Added support for CI hardware testing.

kYc0o:
Removed tests on hardware because unrelated errors block this PR. However the test added by this PR passed.

kYc0o:
Let's go!

PR #8898, opened 2018-04-07
cpu/atmega_common: context switching rewrite

### LINKED ISSUE #8896: cpu/atmega_common: Context switching race condition

ZetaR60:
WIP context switching rewrite for the ATmegas. It has three goals:
- Fix #8896 (context switching race condition)
- Remove code defining context switches from board.h instances and put it in thread_arch.c
- Provide a second context switching method using a timer/counter for pin constrained boards

Race conditions are solved by waiting in thread_yield_higher for a condition set by the context switching interrupt.
- In the case of the pin option: thread_yield_higher sets the pin high and waits for it to be set low, and the ISR sets it low.
- In the case of the counter option: thread_yield_higher sets the counter prescaler to run at full speed and then waits for the counter to be at its max value (CTC mode prevents it from ever counting up to this value), and the ISR sets the prescaler back to no clock and sets the counter to the max value.

The code that allows periph/gpio.c to generically handle pins has been moved to include/periph_cpu_pins.h so that it can be used by both the gpio code and the context switching code. Context switching with pins can then use the GPIO_PIN like everything else. This is necessary for preventing the context switching fix from making board.h sloppy. When using the pin option AVR_CONTEXT_SWAP_INTERRUPT_VECT must still be set due to preprocessor limitations. When using the counter option, you just need to set which counter is used.

References: Fixes #8896. Race conditions are tested for by #8897. Some other issues may be found to be context switching issues.

### DIFF
--- a/cpu/atmega_common/include/periph_cpu_pins.h
+++ b/cpu/atmega_common/include/periph_cpu_pins.h
@@ -1,3 +1,28 @@
+/*
+ * Copyright (C) 2015 HAW Hamburg
+ *               2016 INRIA
+ *               2018 Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     cpu_atmega_common
+ * @{
+ *
+ * @file
+ * @brief       CPU specific definitions for internal pin handling
+ *
+ * @author      René Herthel <rene-herthel@outlook.de>
+ * @author      Francisco Acosta <francisco.acosta@inria.fr>
+ * @author      Laurent Navet <laurent.navet@gmail.com>
+ * @author      Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * @}
+ */
+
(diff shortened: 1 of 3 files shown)


ZetaR60:
Superseded by #8904

Issue #5745, opened 2016-08-12
AVR thread yield


jthacker:
I believe the AVR thread yield is not implemented correctly. `thread_arch_yield` (aka `thread_yield_higher`) is currently implemented as a naked function that will begin execution of the new thread at the same level as the calling point. In the atmega version of `timer.c`, `thread_yield_higher` is called in an ISR, which can result in long running functions at this level possible halting the execution of all other threads. https://github.com/RIOT-OS/RIOT/blob/master/cpu/atmega_common/periph/timer.c#L193

I discovered this when working with xtimer. It would consistently run the spin loop version instead of sleeping the thread due to running at interrupt level. https://github.com/RIOT-OS/RIOT/blob/master/sys/xtimer/xtimer.c#L45

A possible method for eliminating this issue is to use an interrupt driven thread swap like on the other platforms. For example, on the cortex platform, an interrupt is triggered and will be serviced as normal, the resulting context swap will place the new thread at the normal execution level since it will only start execution after returning the thread swapping ISR returns. Unlike the cortex platform, AVRs do not have a dedicated software driven interrupt for context swaps.

I have implemented a method for doing this by using one of the external interrupt pins, while keeping it as an output and driving it through software. https://github.com/RIOT-OS/RIOT/compare/master...jthacker:avr_thread_arch_isr_stack_usage

`CONTEXT_SWAP_INIT`, `CONTEXT_SWAP_INT_VECT` and `CONTEXT_SWAP_TRIGGER` would need to implemented by each board since this requires sacrificing a physical pin on each board, and may be dependent on the layout. Obviously this is not ideal. For this reason I have opened this as an issue and not a pull request to see if firstly there is agreement with the issue at hand and to discuss possible alternative solutions.

kaspar030:
Wow, this might explain a lot of trouble we've been having with AVR's interrupts behaving ... weird.

kaspar030:
> Obviously this is not ideal.

Quick googling suggests that this might be the only way to have proper ISR handling. So a PR would be very welcome.

kYc0o:
I agree with Kaspar, this could be the origin of several trouble we have especially with timers, maybe this could fix it.

@jthacker feel free to open a PR from your branch. Notice that not all external interrupts are implemented for this MCU family, so maybe we can complete the support and use a free pin for context switching.

roberthartung:
For me this breaks the default example I  get a kernel panic because the msg_queue is not initialized

PR #8904, opened 2018-04-09
cpu/atmega_common: return to non-interrupt context swaps

### LINKED ISSUE #8896: cpu/atmega_common: Context switching race condition

ZetaR60:
Second attempt. See also, first attempt: #8898

This removes the ISR based context swaps on the ATmegas. They were originally introduced in #5745. ISR swaps were introduced to handle the problems with context swaps while inside an ISR. However, there aren't really any differences in the execution environment within an ISR on the ATmega, which means that there are simpler ways to handle this problem.

The behavior is as follows:
- If thread_yield_higher is called outside of an ISR, it changes context and then changes the program counter using the 'ret' (subroutine return) instruction. This is the same behavior as prior to #5745.
- If thread_yield_higher is called within an ISR, it sets sched_context_switch_request and does nothing else.
- The function thread_yield_isr is introduced, which is meant to be called within ISRs to yield. It changes the context, tells RIOT it is no longer within an ISR (using __exit_isr) and then changes the program counter using the 'reti' (interrupt return) instruction.

tests/thread_* on mega-xplained all passed (including tests/thread_race added by #8897). Also tested with xtimer_* on mega-xplained and all passed except for xtimer_longterm (did not run this test) and xtimer_now64_continuity which showed some inaccuracy. This may be due to the low xtimer accuracy on mega-xplained however.

References:
- Fixes #8896
- Race conditions tested for by #8897
- First attempt at solving race conditions #8898
- ISR context swaps introduced by #5745

### DIFF
--- a/boards/common/arduino-atmega/include/board_common.h
+++ b/boards/common/arduino-atmega/include/board_common.h
@@ -60,32 +60,6 @@ extern "C" {
 /** @} */
 
 /**
- * @brief   Context swap defines
- *
- * Setup to use PC5 which is pin change interrupt 13 (PCINT13)
- * This emulates a software triggered interrupt
- */
-#ifdef CPU_ATMEGA328P
-#define AVR_CONTEXT_SWAP_INIT do { \
-            DDRD |= (1 << PD7); \
-            PCICR |= (1 << PCIE2); \
-            PCMSK2 |= (1 << PCINT23); \
-} while (0)
-#define AVR_CONTEXT_SWAP_INTERRUPT_VECT  PCINT2_vect
-#define AVR_CONTEXT_SWAP_TRIGGER   PORTD ^= (1 << PD7)
-#endif
-
-#ifdef CPU_ATMEGA2560
-#define AVR_CONTEXT_SWAP_INIT do { \
-    DDRJ |= (1 << PJ6); \
-    PCICR |= (1 << PCIE1); \
-    PCMSK1 |= (1 << PCINT15); \
-} while (0)
-#define AVR_CONTEXT_SWAP_INTERRUPT_VECT  PCINT1_vect
-#define AVR_CONTEXT_SWAP_TRIGGER   PORTJ ^= (1 << PJ6)
-#endif
-
-/**
  * @name    xtimer configuration values
  * @{
  */
--- a/boards/jiminy-mega256rfr2/include/board.h
+++ b/boards/jiminy-mega256rfr2/include/board.h
@@ -76,21 +76,6 @@ extern "C" {
 /** @} */
 
 /**
- * @name Context swap defines
- * This emulates a software triggered interrupt
- * @{
- */
-#define AVR_CONTEXT_SWAP_INIT do { \
-        DDRE |= (1 << PE7); \
-        EICRB |= (1 << ISC70); \
-        EIMSK |= (1 << INT7); \
-        sei(); \
-} while (0)
-#define AVR_CONTEXT_SWAP_INTERRUPT_VECT  INT7_vect
-#define AVR_CONTEXT_SWAP_TRIGGER   PORTE ^= (1 << PE7)
-/** @} */
-
-/**
  * @name xtimer configuration values
  * @{
  */
--- a/boards/mega-xplained/include/board.h
+++ b/boards/mega-xplained/include/board.h
@@ -50,22 +50,6 @@ extern "C" {
 #define UART_STDIO_DEV       (UART_DEV(1))
 
 /**
- * @name    Context swap defines
- *
- * Setup to use PD7 which is pin change interrupt 31 (PCINT31)
- * This emulates a software triggered interrupt
- * @{
- */
-#define AVR_CONTEXT_SWAP_INIT do { \
-            DDRD |= (1 << PD7); \
-            PCICR |= (1 << PCIE3); \
-            PCMSK3 |= (1 << PCINT31); \
-} while (0)
-#define AVR_CONTEXT_SWAP_INTERRUPT_VECT  PCINT3_vect
-#define AVR_CONTEXT_SWAP_TRIGGER   PORTD ^= (1 << PD7)
-/** @} */
-
-/**
  * @name    xtimer configuration values
  *
  * Xtimer runs at 8MHz / 64 = 125kHz
--- a/boards/waspmote-pro/include/board.h
+++ b/boards/waspmote-pro/include/board.h
@@ -146,21 +146,6 @@ extern "C" {
 /** @} */
 
 /**
- * @brief Context swap defines
- * Setup to use PB5 which is pin change interrupt 5
- * This emulates a software triggered interrupt
- **/
-#define AVR_CONTEXT_SWAP_INIT do { \
-    DDRB |= (1 << PB5); \
-    PCICR |= (1 << PCIE0); \
-    PCMSK0 |= (1 << PCINT5); \
-} while (0)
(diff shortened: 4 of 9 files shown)


ZetaR60:
When #8904 is applied to #8842 it fixes one race condition, but reveals another. Requires more testing to see if it is caused by #8904.

ZetaR60:
Issue in #8842 is unrelated to #8904. See #8914

ZetaR60:
Results of a large number of tests with trunk and fix applied to trunk side by side. I did not examine what each test was doing in detail, as long as the output looked the same and didn't look like a failure.

```
Test				Trunk				Fix
-------------------------------------------------------------------------------------------------------------------
events				reports success			identical output
evtimer_msg			received expected output	identical, except slightly different times reported
evtimer_underflow		received expected output	identical output
fault_handler			received expected output	identical output
irq				reports success			identical output
isr_yield_higher		thread done after success	ordering different
memarray			received expected output	identical output, except slightly different
								base and current addresses
msg_avail			reports success			identical output
msg_send_receive		reports success			identical output
msg_try_receive			received expected output	identical output
mutex_order			hangs after T7 locked		correct according to README
mutex_unlock_and_sleep		received expected output	identical output
posix_semaphore			threads do not report waking	received expected output
				in TEST2, TEST3
posix_time			received expected output	identical output
ps_schedstatistics		hangs after threads,		identical output
				and before shell
rmutex				hangs after T4 (prio 4,		T7 and T4 switched from README
				depth 0): unlocked mutex
sched_testing			received expected output	identical output
sizeof_tcb			reports success			identical output
thread_basic			received expected output	identical output
thread_cooperation		reports success			identical output
thread_exit			received expected output	identical output
thread_flags			reports success			identical, except trunk reports 100ms timeout
								as 160us and fix reports as 34648us
thread_flags_xtimer		reports success			identical output
thread_flood			reports success			identical output
thread_msg			reports success			identical output
thread_msg_block_wo_queue	received expected output	identical output
thread_msg_block_w_queue	received expected output	identical output
thread_msg_seq			reports success, but did not	reports success, did receive
				receive "THREAD nr3 (pid:5) end."
thread_priority_inversion	received expected output	identical output
trickle				TRICKLE_RESET			FAIL
warn_conflict			nothing to do			identical
xtimer_drift			received expected output	identical, except slightly different times reported
xtimer_hang			reports success			identical output
xtimer_msg			received expected output	identical, except slightly different times reported
xtimer_msg_receive_timeout	reports success			identical output
xtimer_now64_continuity		reports FAILURE,		identical output
				min=10, avg=10, max=1040
xtimer_periodic_wakeup		received expected output	identical, except slightly different times reported
								(identical min/max error reported)
xtimer_remove			reports success			identical output
xtimer_reset			received expected output	identical, except slightly different times reported
xtimer_usleep			received expected output	identical, except slightly different times reported
xtimer_usleep_short		reports success			identical output
```

tests of interest (significant divergences):
- isr_yield_higher
- mutex_order
- posix_semaphore
- rmutex
- thread_flags
- thread_msg_seq
- trickle

I will not be able to follow up on this right away. I will be out of town for a week.

kYc0o:
Testing on arduino-mega2560:

  - isr_yield_higher **Almost OK, it prints the success message a bit before the timer gets triggered, but it's OK IMHO.**

  - mutex_order **OK**

  - posix_semaphore **Fails, but has better results than master. Here I think it has something to do with xtimer or UART speed.**

  - rmutex **OK**

  - thread_flags **Almost OK. Again timer issues.**

  - thread_msg_seq **OK**

  - trickle **Fail. Timing issues. Although it makes use of random which is not really supported on AVR.**

I will try to investigate the timing issues, but I'd say this PR is a step forward.

Can you just confirm that this is getting to the exact previous implementation? I mean, before the context switching was implemented with interrupts.

kYc0o:
> Can you just confirm that this is getting to the exact previous implementation? I mean, before the context switching was implemented with interrupts.

I just looked by myself and see that this is somehow a new approach, thus is not exactly coming back to a previous state.

ZetaR60:
@kYc0o The old approach did not handle context swaps inside ISRs correctly. The context swap would run inside the ISR and the other thread would be executing in ISR mode. That was why the old approach was changed to use the interrupts. Non-ISR handling should be the same with this fix.

Differences within an ISR from the old approach:
- thread_yield and thread_yield_higher do not save / restore context or yield, but set sched_context_switch_request instead.
- thread_yield_isr is used at the end of ISRs so that if a yield was attempted (and sched_context_switch_request was set) then it: saves/restores contexts, exits from the ISR (which really only means it tells RIOT it is no longer in an ISR by using __exit_isr), and then issues a return from interrupt instruction (which actually switches the program counter).

This fixes the problem the old code had with ISRs.

gebart:
@ZetaR60 from your textual description it sounds like the avr with this PR behaves like the Cortex-M port does

kYc0o:
@Josar can you test this in your jiminy board?

Josar:
@kYc0o if i undersstand correctly running `tests/xtimer_msg` should suffice to prove if this works. The test produces expected output.

ZetaR60:
Rebased on master and merge conflicts solved.

ZetaR60:
thread_flags looks okay to me after a closer look. 01-run.py expects ``thread(): received flags: 0x0006`` and it is intentionally waited for by _thread (because 0x2 | 0x4 == 0x6). The timeout is intentional (not an error); testing the timeout functionality I guess? The nominal value for the timeout should be 100ms (or 100,000us), so it is about 60-70ms off. But this is better than the previous behavior (160us for me), which is almost an order of magnitude off. I suggest we consider this one to be okay.

Still looking into posix_semaphore and trickle.

ZetaR60:
The issues with posix_semaphore seem to be related to a xtimer problem and not to this fix. There is a separate issue for that problem: #9049. I think posix_semaphore would pass with #8904 if #9049 was fixed.

This is probably also the cause of the trickle failure.

ZetaR60:
The problems with trickle is not caused by the same underlying issue as posix_semaphore. There seems to be some problems with how tests/trickle works, at least on ATmegas, and is probably unrelated to this fix. Issue for tests/trickle: #9052

I have investigated the three main problem tests (thread_flags, posix_semaphore, and trickle). If the rest of you think that the other tests are okay or have only insignificant differences, then I think that all the currently outstanding problems with this fix have been solved.

kYc0o:
Ok, I understand. Let's tackle the other problems separately.

ZetaR60:
@kYc0o Squashed. Thanks for your help!

PR #8911, opened 2018-04-10
Makefile.include: fix missing target for libraries

### LINKED ISSUE #8910: tests: Unittests don't build
#### Description
Unittests don't build in current master.
Quick bisect we did with @miri64 points to #8844 as introducing this problem.

```
cd tests/unittests
make
```

 or

```
make tests-<module>
```

#### Expected results
It should build.

#### Actual results
It does not build. Sample error:

```
make: *** No rule to make target '/home/vagrant/RIOT/tests/unittests/bin/samr21-xpro/tests-tweetnacl.a', needed by '/home/vagrant/RIOT/tests/unittests/bin/samr21-xpro/tests_unittests.elf'.  Stop.

```

#### Versions
Issue detected on Linux and OS X

cladmi:
Unittests add libraries in 'BASELIBS' which do not have any rules to be built as
they are built by 'application.inc.mk' and the DIRS variable.
So make complains about missing target for the unittests archives.

The fix tells these files are generated when building '$(APPLICATION_MODULE).a'.

The bug was introduced by #8844

Fixes #8910

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -352,10 +352,13 @@ endif # RIOTNOLINK
 $(ELFFILE): $(BASELIBS)
 	$(Q)$(_LINK) -o $@
 
-# All modules are built by application.inc.mk makefile
 $(BINDIR)/$(APPLICATION_MODULE).a: $(RIOTBUILD_CONFIG_HEADER_C) $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS)
 	$(Q)DIRS="$(DIRS)" "$(MAKE)" -C $(APPDIR) -f $(RIOTMAKE)/application.inc.mk
 $(BINDIR)/$(APPLICATION_MODULE).a: FORCE
+
+# Other modules are built by application.inc.mk and packages building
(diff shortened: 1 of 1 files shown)


cladmi:
I will try a PR to show that murdock fails to detect it.

bergzand:
This fixes #8910 for me. I'm not familiar enough with the makefiles to review the code here though.

cladmi:
Even when doing `make clean; make all` it works, the commits for murdock will be removed before merging.

@emmanuelsearch can you test ?

emmanuelsearch:
`make clean` does help indeed.

cladmi:
The `_SUBMAKE_LIBS` is at that point not really defined to the files built by the application makefile because `$(BASELIBS)` does not have its final value.
But as the `$(ELFFILE)` rule also uses the same value, the fix works.

I added an issue to track this problem of use before define. https://github.com/RIOT-OS/RIOT/issues/8913

cladmi:
Fixed the rule because some modules are built by packages makefile.

cladmi:
Rebased and squashed message

PR #8912, opened 2018-04-10
Test murdock failed to detect bug in Makefile

cladmi:
### Contribution description

Murdock was not finding the problem described in https://github.com/RIOT-OS/RIOT/issues/8910

~~Reason is that archives files are kept between build, and so when running `make clean all` the files exist and make did not complain.~~ (I mixed with buildtest command)

As murdock is run with `clean all`, there is a rule that say `$(BASELIBS) (other_stuff): clean` so make was not complaining.

The goal of this PR is just to show that it fails in murdock without fix in https://github.com/RIOT-OS/RIOT/pull/8911

### Issues/PRs references

Issue https://github.com/RIOT-OS/RIOT/issues/8910

### DIFF
--- a/.murdock
+++ b/.murdock
@@ -120,7 +120,9 @@ compile() {
 
     # compile
     CCACHE_BASEDIR="$(pwd)" BOARD=$board RIOT_CI_BUILD=1 \
-        make -C${appdir} clean all -j${JOBS:-4}
+        make -C${appdir} clean -j${JOBS:-4}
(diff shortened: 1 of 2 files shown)


cladmi:
I modified the wrong file before…

cladmi:
Yeah, build is failing as expected.
This should not be mergedl, it is only to show that https://github.com/RIOT-OS/RIOT/pull/8911 fixes work in murdock when doing `make clean; make all`.

cladmi:
Not needed anymore, it was only to show #8911 fixed the problem.

PR #8916, opened 2018-04-11
murdock: unify dwq env list, pass NIGHTLY and RUN_TESTS to workers

kaspar030:
### Contribution description

murdock uses the NIGHTLY and RUN_TESTS environment variables to determine whether to run the on-hardware tests, thus those need to be passed to sub-jobs.


kaspar030:
thx!

Issue #8831, opened 2018-03-25
[RFC] gcoap adaptation to #8772


kb2ma:
[ Edited to add gcoap_opt_add_string() ]

I have an outline now for adaptation of gcoap for #8772. Overall, I don't see any issues. The existing gcoap API can remain, and I'll add a couple of functions for options. For implementation, I'll add a stack-based buffer for compiling options and an additional memcpy(). The overall gcoap approach remains -- a higher-level API to CoAP.

The gcoap functions for adding options will be like:

   gcoap_opt_add_buf(pdu, optnum, buf_ptr, buf_len)
   gcoap_opt_add_string(pdu, optnum, string_ptr, separator)
   gcoap_opt_add_value(pdu, optnum, uint32)
   
The implementation appends option values to the packet buffer as they are provided by the user. The user is _not_ required to add Options in order by option number. These values will be in the format of the raw data, like a query string. gcoap will use the new coap_optpos_t options[] array to store the position of these options in the buffer. gcoap will move the packet payload pointer as options are added, so it will be ready for the user when they decide to write the payload. You can see an example of this approach below.

```
   +-------+---------+---------+---------+----- ... --+
      hdr  |  path   | uint32  |  query  |    payload
           |         |         |         |
        opt_pos   opt_pos   opt_pos    payload*
        Uri-Path  Observe   Uri-Query
```     
Finally, gcoap_finish() will sort the Options into order by option number. It will create a stack buffer, and use the nanocoap functions like coap_put_option() to write the contents in order to the stack buffer. Finally, gcoap will copy the stack buffer to the packet buffer and do the final memmove to reposition the payload, as it does currently.

bergzand:
@kb2ma overall I like the idea. Do I understand correctly that the options are only added to the `coap_optpos_t` array and are only written to the buffer when the packet is finalised, and thus that the order of adding options doesn't matter?

kb2ma:
Cool, thanks for the confirmation. Yes, option positions are stored in the coap_optpos_t array. The option values are actually stored in the packet buffer as a scratchpad as the user adds them. gcoap also maintains the payload pointer while building the options, so the user can copy that after specifying the last option. Then when the packet is finalized in the gcoap_finish() call, gcoap uses the coap_optpos_t array and the raw options in the packet buffer to write the real CoAP  options in the proper order to a stack-allocated buffer. Finally the stack buffer contents are copied back to the packet buffer, ready to send.

[...]

PR #8920, opened 2018-04-11
net/nanocoap: add generic handling for string-based options

haukepetersen:
### Contribution description
Many string-based options (e.g. `COAP_OPT_URI`, `COAP_OPT_LOCATION_PATH`) share a similar structure: they consist of string 'snippets', that are split using some kind of separation character, typically `/`. This PR adds a more generic function for setting and reading these kind of options to `nanocoap`.

Please let me know what you think of this approach - once consensus is reached I will spend the time to add the missing doxygen...

### Issues/PRs references
Follow-up on #8772

### DIFF
--- a/sys/include/net/nanocoap.h
+++ b/sys/include/net/nanocoap.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2016-17 Kaspar Schleiser <kaspar@schleiser.de>
+ *               2018 Freie Universität Berlin
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -18,6 +19,7 @@
  *
  * @author      Kaspar Schleiser <kaspar@schleiser.de>
  * @author      Ken Bannister <kb2ma@runbox.com>
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
  */
 
 #ifndef NET_NANOCOAP_H
@@ -63,9 +65,11 @@ extern "C" {
  */
 #define COAP_OPT_URI_HOST       (3)
 #define COAP_OPT_OBSERVE        (6)
+#define COAP_OPT_LOCATION_PATH  (8)
 #define COAP_OPT_URI_PATH       (11)
 #define COAP_OPT_CONTENT_FORMAT (12)
 #define COAP_OPT_URI_QUERY      (15)
+#define COAP_OPT_LOCATION_QUERY (20)
 #define COAP_OPT_BLOCK2         (23)
 #define COAP_OPT_BLOCK1         (27)
 /** @} */
@@ -450,17 +454,87 @@ size_t coap_put_option(uint8_t *buf, uint16_t lastonum, uint16_t onum, uint8_t *
 size_t coap_put_option_ct(uint8_t *buf, uint16_t lastonum, uint16_t content_type);
 
 /**
- * @brief   Insert URI encoded option into buffer
+ * @brief   Encode the given string as multi-part option into buffer
+ *
+ * @param[out]  buf         buffer to write to
+ * @param[in]   lastonum    number of previous option (for delta calculation),
+ *                          or 0 if first option
+ * @param[in]   optnum      option number to use
+ * @param[in]   string      string to encode as option
+ * @param[in]   separator   character used in @p string to separate parts
+ *
+ * @return      number of bytes written to @p buf
+ */
+size_t coap_opt_put_string(uint8_t *buf, uint16_t lastonum, uint16_t optnum,
+                           const char *string, char separator);
+
+/**
+ * @brief   Convenience function for inserting URI_PATH option into buffer
  *
  * @param[out]  buf         buffer to write to
  * @param[in]   lastonum    number of previous option (for delta calculation),
  *                          or 0 if first option
  * @param[in]   uri         ptr to source URI
- * @param[in]   optnum      option number to use (e.g., COAP_OPT_URI_PATH)
  *
  * @returns     amount of bytes written to @p buf
  */
-size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uint16_t optnum);
+static inline size_t coap_opt_put_uri_path(uint8_t *buf, uint16_t lastonum,
+                                           const char *uri)
+{
+    return coap_opt_put_string(buf, lastonum, COAP_OPT_URI_PATH, uri, '/');
+}
+
+/**
+ * @brief   Convenience function for inserting URI_QUERY option into buffer
+ *
+ * @param[out]  buf         buffer to write to
+ * @param[in]   lastonum    number of previous option (for delta calculation),
+ *                          or 0 if first option
+ * @param[in]   uri         ptr to source URI
+ *
+ * @returns     amount of bytes written to @p buf
+ */
+static inline size_t coap_opt_put_uri_query(uint8_t *buf, uint16_t lastonum,
+                                            const char *uri)
+{
+    return coap_opt_put_string(buf, lastonum, COAP_OPT_URI_QUERY, uri, '&');
+}
+
+/**
+ * @brief   Convenience function for inserting LOCATION_PATH option into buffer
+ *
+ * @param[out]  buf         buffer to write to
+ * @param[in]   lastonum    number of previous option (for delta calculation),
+ *                          or 0 if first option
+ * @param[in]   location    ptr to string holding the location
+ *
+ * @returns     amount of bytes written to @p buf
+ */
+static inline size_t coap_opt_put_location_path(uint8_t *buf,
+                                                uint16_t lastonum,
+                                                const char *location)
+{
+    return coap_opt_put_string(buf, lastonum, COAP_OPT_LOCATION_PATH,
+                               location, '/');
+}
+
+/**
+ * @brief   Convenience function for inserting LOCATION_QUERY option into buffer
+ *
+ * @param[out]  buf         buffer to write to
+ * @param[in]   lastonum    number of previous option (for delta calculation),
+ *                          or 0 if first option
+ * @param[in]   location    ptr to string holding the location
+ *
+ * @returns     amount of bytes written to @p buf
+ */
+static inline size_t coap_opt_put_location_query(uint8_t *buf,
+                                                 uint16_t lastonum,
+                                                 const char *location)
+{
+    return coap_opt_put_string(buf, lastonum, COAP_OPT_LOCATION_QUERY,
+                               location, '&');
+}
 
 /**
  * @brief    Generic block option getter
@@ -585,10 +659,30 @@ ssize_t coap_opt_finish(coap_pkt_t *pkt, uint16_t flags);
 unsigned coap_get_content_type(coap_pkt_t *pkt);
 
 /**
- * @brief   Get the packet's request URI
+ * @brief   Read a full option as null terminated string into the target buffer
  *
- * This function decodes the pkt's URI option into a "/"-seperated and
- * NULL-terminated string.
+ * This function is for reading and concatenating string based, multi-part CoAP
+ * options like COAP_OPT_URI_PATH or COAP_OPT_LOCATION_PATH. It will write all
+ * parts of the given option into the target buffer, separating the parts using
+ * the given @p separator. The resulting string is `\0` terminated.
+ *
+ * @param[in]   pkt         packet to read from
+ * @param[in]   optnum      absolute option number
+ * @param[out]  target      target buffer
+ * @param[in]   max_len     size of @p target
+ * @param[in]   separator   character used for separating the option parts
+ *
+ * @return      -ENOSPC if the complete option does not fit into @p target
+ * @return      nr of bytes written to @p target (including '\0')
+ */
+ssize_t coap_opt_get_string(const coap_pkt_t *pkt, uint16_t optnum,
+                            uint8_t *target, size_t max_len, char separator);
+
+/**
+ * @brief   Convenience function for getting the packet's URI_PATH
+ *
+ * This function decodes the pkt's URI option into a "/"-separated and
+ * '\0'-terminated string.
  *
  * Caller must ensure @p target can hold at least NANOCOAP_URI_MAX bytes!
  *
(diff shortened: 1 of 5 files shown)


kaspar030:
ACK for the idea!

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -504,6 +512,8 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
     return bufpos - buf;
 }
 
+
```
kaspar030: ...

kaspar030:
@kb2ma I'm about to ACK, would you like to take a quick look?

sys/include/net/nanocoap.h
```diff
@@ -421,7 +439,27 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
...
+ * @param[in]   pkt         packet to read from
+ * @param[in]   option      absolute option number
+ * @param[out]  target      target buffer
+ * @param[in]   max_len     size of @p target
+ * @param[in]   separator   character used for separating the option parts
+ *
+ * @return      -ENOSPC if the complete option does fit into @p target
+ * @return      nr of bytes written to @p target (including '\0')
+ */
+int coap_get_opt_string(const coap_pkt_t *pkt, uint16_t optnum,
```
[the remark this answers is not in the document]
bergzand: `ssize_t` since error codes are negative?

sys/include/net/nanocoap.h
```diff
@@ -434,7 +494,32 @@ unsigned coap_get_content_type(coap_pkt_t *pkt);
  * @returns     -ENOSPC     if URI option is larger than NANOCOAP_URI_MAX
  * @returns     nr of bytes written to @p target (including '\0')
  */
-int coap_get_uri(coap_pkt_t *pkt, uint8_t *target);
+static inline int coap_get_uri(const coap_pkt_t *pkt, uint8_t *target)
```
bergzand: If @kaspar030 [comment](https://github.com/RIOT-OS/RIOT/pull/8920/files#r180752431) is applied, the return type also propagates to this function.

sys/include/net/nanocoap.h
```diff
@@ -434,7 +494,32 @@ unsigned coap_get_content_type(coap_pkt_t *pkt);
...
+ * Caller must ensure @p target can hold at least 2 bytes!
+ *
+ * @param[in]   pkt     pkt to work on
+ * @param[out]  target  buffer for location path
+ * @param[in]   max_len size of @p target in bytes
+ *
+ * @returns     -ENOSPC     if URI option is larger than @p max_len
+ * @returns     nr of bytes written to @p target (including '\0')
+ */
+static inline int coap_get_location_path(const coap_pkt_t *pkt,
```
bergzand: And to this one

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -131,9 +133,9 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
     return 0;
 }
 
-uint8_t *coap_find_option(coap_pkt_t *pkt, unsigned opt_num)
+uint8_t *coap_find_option(const coap_pkt_t *pkt, unsigned opt_num)
 {
-    coap_optpos_t *optpos = pkt->options;
+    coap_optpos_t *optpos = (coap_optpos_t *)pkt->options;
```
bergzand: If you mark `*optpos` as const you don't need the cast here.

haukepetersen:
Just a remark before I forget it again: somebody would need to go through all of the `nanocoap` code to check/fix the usage of `const` for function parameters. In this PR, I only touched the occurrences that were 'in my path'... 

Comments look all valid, will address them ASAP

bergzand:
There is a typo in the commit message of the second commit "LOCACTION_PATH" -> "LOCATION_PATH".

kb2ma:
> I'm about to ACK, would you like to take a quick look?

Yes, thanks. I'll post my feedback later today.

sys/include/net/nanocoap.h
```diff
@@ -421,7 +461,27 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
...
+ * parts of the given option into the target buffer, separating the parts using
+ * the given @p separator. The resulting string is `\0` terminated.
+ *
+ * @param[in]   pkt         packet to read from
+ * @param[in]   option      absolute option number
+ * @param[out]  target      target buffer
+ * @param[in]   max_len     size of @p target
+ * @param[in]   separator   character used for separating the option parts
+ *
+ * @return      -ENOSPC if the complete option does fit into @p target
```
kb2ma: Should be "does not fit".

haukepetersen:
rebased, squashed, and fixed comments:
- renamed 'get' function to `coap_get_option_string()` -> more consistent with `coap_put_option_` functions
- changed return type for `coap_get_option_string()` and its users to `ssize_t`
- make `optpos` in `coap_find_option` const
- fixed some typos in doxygen

haukepetersen:
and fixed the typo in the commit message :-)

kb2ma:
I'm going to put my review here to keep it all together, rather than put some of it in the code.

coap_put_option_location_path() -- shouldn't this be coap_put_option_location() to include Location-Path and Location-Query? This approach would be consistent with coap_put_option_uri(), which includes Uri-Path and Uri-Query.

I have a couple of higher level concerns. Why do coap_[get|put]_option_string() require the separator as an argument? Isn't this implicit in the option number argument? For example, Uri-Path always uses '/', and Uri-Query uses '&' except for the initial '?'.

What about options with an opaque data type, like ETag? WIll this require a coap_[get|put]_option_buf()? Why not roll up string and opaque options into a single function? Once again, the option number defines the behavior.

I think the fundamental API is:
- coap_get_option_buf(pkt, optnum, uint8_t *target, max_len)
- coap_put_option_buf(buf, lastonum, optnum, target, uint8_t *source)
- coap_get_option_value(pkt, optnum, uint32_t *target) 
- coap_put_option_value(buf, lastonum, optnum, value)

Any other option-specific functions likely are inline on top of those.

gcoap then simplifies the put options by removing the requirement to specify them in order, as shown in #8831.

kaspar030:
> Why not roll up string and opaque options into a single function?

Isn't the difficulty with string functions that one e.g, path is split into multiple occurences of the option?

kb2ma:
I have been experimenting with the adaptation of gcoap with the API I described above. The problem with this approach is that it means nanocoap would need to maintain lookup tables for the configuration of an option. Examples -- is it repeatable, what is the separator character, how to match the option number to a position in the table.

This approach would allow nanocoap to validate options more carefully. However, I think this work is two steps away from where we are today. The next step is just to get options working based on #8772. So, I agree that the API needs functions to get/set string values. If/when we get to option definitions, the string functions can be deprecated.

So how about this for an updated API. I do think the use of 'coap_opt' for the prefix improves usability, as @kaspar030 [suggested](https://github.com/RIOT-OS/RIOT/pull/8772#issuecomment-378625669).

    coap_opt_get_string(pkt, optnum, uint8_t *target, max_len, separator)
    coap_opt_put_string(buf, lastonum, optnum, char *source, separator)
    coap_opt_get_buf(pkt, optnum, uint8_t *target, max_len)
    coap_opt_put_buf(buf, lastonum, optnum, target, uint8_t *source, source_len)
    coap_opt_get_uint(pkt, optnum, uint32_t *target)
    coap_opt_put_uint(buf, lastonum, optnum, value)

haukepetersen:
> coap_put_option_location_path() -- shouldn't this be coap_put_option_location() 

yes, that's better indeed. Will fix.

> I have a couple of higher level concerns. Why do coap_[get|put]_option_string() require the separator as an argument? Isn't this implicit in the option number argument? For example, Uri-Path always uses '/', and Uri-Query uses '&' except for the initial '?'.

As you stated below, making this explicit depending on the option number implies that we need to keep some kind of a lookup table to map the behavior. This is IMHO overkill (especially considering nanocaop's scope to be as lightweight as possible). Further, the separator character in the string representation is something specific chosen by the nanocoap user, and is not really for making the string nice to read for humans, so there is IMHO no need for having e.g. a query string start with `?` instead of starting it also with an `&`...

> So how about this for an updated API.

I like it. I will adapt the functions touched by this PR and propose that we open an issue for adapting the rest of the option handling functions in separate PR(s). How does that sound?

kaspar030:
> So how about this for an updated API.

Thinking about this:

```
coap_opt_get_string(pkt, optnum, uint8_t *target, max_len, separator)
coap_opt_put_string(buf, lastonum, optnum, char *source, separator)
```

... the get functions get a packet, the put functions a buffer.
@kb2ma intends to add higher level "put" type functions that work on a pkt, into gcoap. IMO, those can actually reside in nanocoap.

I propose using ```coap_opt_put_<type>_raw()``` for the put functions that write directly into a buffer, then introduce higher-level functions without _raw as described in #8831.

haukepetersen:
Just talked to @kaspar030 offline and he convinced me about the `_raw` suffix -> adapted the PR.

kb2ma:
I re-read my last comment, and I'd like to be sure I understand @kaspar030's suggestion. Are you saying you would like the function below in nanocoap?

`coap_opt_add_string(pkt, optnum, char *source, separator)
`
Should a nanocoap user be able to use this function? It requires that the last option number written is maintained while writing the packet.

In contrast, gcoap does not require options to be added in order. So, a coap_opt_add_string(pkt) is troublesome because the option being added may have a negative optnum delta relative to the previous option.

kaspar030:
> Are you saying you would like the function below in nanocoap?

Yes, that was my intention. I hoped that your proposal for "add" functions can be implemented efficient enough to be considered "low-level". :)

My hope was that using those functions, the user would just fill the pkt's option array with pointers to the raw data, and some kind of "coap_option_finish(pkt)" would then "put" the options into the pkt's buffer, in order, converted. I think that is pretty much your proposal for gcoap.
I just realized that the option array currently just cannot fit even a pointer to the option data...

kb2ma:
I just added a [wiki page](https://github.com/RIOT-OS/RIOT/wiki/CoAP-API-Options-Update) with a proposal for updating the nanocoap/gcoap APIs to support options as have been discussed in this PR. The proposal includes a plan for nanocoap to use a coap_pkt_t to build a message, as well as adaptation for the gcoap API. It may still need some work at finalization time on the nanocoap side, but I don't know/use that part of the API as others do.

Let's beat it up before we code it down. ;-)

kb2ma:
Let's discuss in #9048 to avoid confusion on the present PR. Thanks.

kb2ma:
> I just realized that the option array currently just cannot fit even a pointer to the option data...

@kaspar030, I think the current implementation of coap_optpos_t, based on offsets, works well enough. I have added most of the coap_opt_add... API for nanocoap (#9085) and use it in gcoap (#9156). gcoap does not sort options. As a rule, I recommend that the user adds options in order. Otherwise, code and structures get complicated. I implemented sorting in #9144, but I'd like to avoid *requiring* it due to the bloat. If someone wants it for convenience, it's available for nanocoap and gcoap.

I also really like how the gcoap and nanocoap APIs are converging. We're developing this integrated,  stepwise sequence of choices for the user, from low-level buffer writing functions up to a single `gcoap_request()` call.

haukepetersen:
@kb2ma @kaspar030: its been a while, what do we do about this PR? Do we have an alternative or can we just get it merged as an intermediate or something? I desperately want to get #7428 in, and this PR is the only thing standing in the way... How should we proceed to get support for URI-query stings into `gcoap` in a timely manner?!

kb2ma:
@haukepetersen, with #9085 merged, I think this PR is almost ready. My only question is whether we want to merge some option infrastructure first.

The infrastructure is #9474 to modularize use of the 'coap_opt_xxx' API, which helps to isolate and eventually deprecate the original options functions. See the rationale in that PR, including the module documentation at the top of nanocoap.h. In this case, we'll need to move some of the functions here to different implementation files. For example, move coap_opt_get_string() to opt2.c.

Beyond 9474, I think the only other change is to the name of the coap_opt_put_xxx_raw() functions. We don't need "_raw" now that the pkt-based functions are named "coap_opt_add_xxx".

haukepetersen:
Ok, I am completely lost. So parts of this PR  (`coap_opt_add_string()` and `coap_opt_add_uint()`) were copied from this PR into #9085, but they dropped the `_raw` in the naming. But the rest of this PR is still to be merged?

> My only question is whether we want to merge some option infrastructure first.

NOO, please let us get this PR in first. I desperately need the handling for the location path and location query options and I am waiting for this for more than a year now, as there seems always one more step of remodeling before to be done... So please, let me have my feature first and then simply include those changes in any upcoming remodeling.

haukepetersen:
ok, I hope I understood the changes now. I tried to adjust the `nanocaop.h` header to the changes done in #9085, does this look about right?

kb2ma:
I am working on a response, and will post it later today. I'm afraid there still is some misunderstanding.

Apologies for the confusion. I hear you loud and clear on any other refactoring. I'm sure we can wrap this up quickly.

kb2ma:
I just added [PR 44](https://github.com/haukepetersen/RIOT/pull/44) in @haukepetersen's repository. That work makes this PR code-complete IMO. I describe the rationale in that PR, and hopefully the code gets us to a merge quicker. :-) Happy to discuss.

haukepetersen:
Thanks @kb2ma! Now I start to remember/understand what the option handling concept was all about :-) I agree with your proposed changes and merged them in, will do one more verification round on my side now and then I hope this PR should be ready to go.

kb2ma:
Excellent! Since you're working with Uri-Query options, remember that coap_parse() currently does not set the coap_pkt_t.qs attribute for gcoap. I'll watch for your OK for final review.

haukepetersen:
alright, fixed one more alignment issue and did some (successful) testing. Also cleaned up the commits and squashed. On my part this PR should be ready for final review!

kb2ma:
Tested OK on native and samr21-xpro, with the code updates in my review. Test included use of rdcli-simple example to register with libcoap coap-rd server. Good stuff!

haukepetersen:
addressed comments and squashed. Thanks for the testing!

haukepetersen:
@kaspar030, @bergzand: are your change requests satisfies as well?

kb2ma:
Still don't see the fix to sock.c and Murdock is unhappy with documentation.

haukepetersen:
you are right, I missed that one

haukepetersen:
ok, should be fixed now.

kb2ma:
Looks good to me! Since you've got an open question out there, I'll leave it to you to merge when ready.

haukepetersen:
Lets wait for the other two to approve or dismiss their review -> @kaspar030, @bergzand

bergzand:
I'll try to take a look at this PR today to see if I can spot any remaining issues.

haukepetersen:
awesome, thanks guys!

-> let's go!

haukepetersen:
dammit, 10 seconds too late :-)

PR #8924, opened 2018-04-11
pkg/spiffs: remove c11 dependency and clean-up makefile

### LINKED ISSUE #8876: spiffs: compile problems on new platform xtensa-lx106 (ESP8266)
I ran into compilation problems when I tried to use ```spiffs``` with my MTD flash device implementation in my RIOT port for the ESP8266 (xtensa-lx106). I got a lot of the following error message, one for each standard type.

```
.../newlib-xtensa/xtensa-lx106-elf/include/stdatomic.h:187:9: error: function definition declared 'typedef'
  typedef _Atomic(_Bool)   atomic_bool; 
```

Reason is propably that the xtensa version of gcc is not fully C11 standard compliant.

```xtensa-lx106-elf-gcc (crosstool-NG crosstool-ng-1.22.0-60-g37b07f6) 4.8.5```

The newlib version is 3.0.0.

I solved the problem for the moment, by adding a new variable ```SPIFFS_STD_OPTION``` in my ```Makefile.include``` and changing the the ```Makefile``` in ```spiffs``` as following:
```
# Compilation of spiffs fails in <stdatomic.h> on some platforms when
# option -std=c11 is set. Use SPIFFS_STD_FLAG to overwrite this option.
ifneq ($(SPIFFS_STD_OPTION), )
    CFLAGS += $(SPIFFS_STD_OPTION)
else
    CFLAGS += -std=c11
endif
```
It would be great if this could be changed also in the master. Otherwise, I could provide it with my upcoming pull request for my RIOT port. Maybe, there is a better way to solve this problem, I had no better idea.

vincent-d:
### Contribution description

A C11 dependency was added to mask a warning about unnamed struct/union.
It can be solved by disabling `-Wpedantic` for the package, to avoid pulling c11 for compiler which do not support it.
I also cleaned up the makefiles.

### Issues/PRs references

Should fix #8876

### DIFF
--- a/pkg/spiffs/Makefile
+++ b/pkg/spiffs/Makefile
@@ -3,8 +3,6 @@ PKG_URL=https://github.com/pellepl/spiffs.git
 PKG_VERSION=287148c46587089c4543a21eef2d6e9e14b88364
 PKG_LICENSE=MIT
 
-CFLAGS += -std=c11
-
 .PHONY: all
 
 all: git-download
--- a/pkg/spiffs/Makefile
+++ b/pkg/spiffs/Makefile
@@ -1,19 +1,13 @@
 PKG_NAME=spiffs
 PKG_URL=https://github.com/pellepl/spiffs.git
 PKG_VERSION=287148c46587089c4543a21eef2d6e9e14b88364
-PKG_BUILDDIR ?= $(PKGDIRBASE)/$(PKG_NAME)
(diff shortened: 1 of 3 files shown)


kaspar030:
@cladmi usually I'd say the same, but this PR seems trivial enough not to waste time on splitting the commit, no?

PR #8925, opened 2018-04-11
Makefile.include: check FEATURES_CONFLICT against FEATURES_USED

cladmi:
The FEATURES_CONFLICT check should be done on used features to also
test for included optional features.

### Contribution description

This make the FEATURES_CONFLICT test detect conflict with optional features included.

It can be tested with: `FEATURES_OPTIONAL="periph_dac periph_spi" make -C examples/hello-world/`
```
$ FEATURES_OPTIONAL="periph_dac periph_spi" make -C examples/hello-world/ 
BOARD=stm32f4discovery
make: Entering directory '/home/cladmi/git/work/RIOT/examples/hello-world'
The following features may conflict: periph_dac periph_spi
Rationale: On stm32f4discovery boards there are the same pins for the DAC and/or SPI_0.

EXPECT undesired behaviour!
Building application "hello-world" for "stm32f4discovery" with MCU "stm32f4".
...
```

### Issues/PRs references

It will allow using FEATURES_CONFLICT for https://github.com/RIOT-OS/RIOT/pull/8890

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -515,11 +515,11 @@ ifneq (, $(filter all, $(if $(MAKECMDGOALS), $(MAKECMDGOALS), all)))
     EXPECT_ERRORS := 1
   endif
 
-  # Test if any required feature conflict with another one.
-  CONFLICT := $(foreach var,$(FEATURES_CONFLICT),$(if $(filter $(words $(subst :, ,$(var))),$(words $(filter $(FEATURES_REQUIRED),$(subst :, ,$(var))))),$(subst :, ,$(var))))
+  # Test if any used feature conflict with another one.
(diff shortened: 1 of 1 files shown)


basilfx:
Works as expected and I can repeat the same for the EFM32.

ACK

basilfx:
(since this affects the Makefiles, do we need another ACK, @cladmi, @kaspar030?)

kaspar030:
I wonder if there's some logic we can implement so that having two optional but conflicting features would select one, or that requiring one conflicting with an optional would just not select the optional.

cladmi:
@kaspar030 I think the second case could be quite easily done but it would be a new feature where here it's just fixing the check.

With the way it's done now, to handle the one required and one optional case, I would do it with a state in in the Makefile.dep iterations. In the first iterations, optional features are not taken in to account, then when USEMODULE is stable, keep iterating and integrate the optional features that do not conflict until USEMODULE is stable again.

kaspar030:
ok!

PR #8926, opened 2018-04-11
drivers/sx127x: various improvements

### LINKED ISSUE #8935: drivers/sx127x: SX1272 driver doesn't work reliably
#### Description

While testing the Semtech Loramac package on a Nucleo board with an SX1272 radio device plugged on it, I realized that it was not working at all.

With more testing, I also noticed an issue with the driver test application in tests/drivers_sx127x when used on a sx1272: it's not possible to send/receive raw radio packets right after a reboot. I opened #8926 that fixes this particular problem.

But there is still an issue remaining with sending/receiving packets when using SF7 radio parameter and I think this is what breaks the loramac support with sx1272 (because the join procedure always uses this configuration).

For the radio driver application:
- Flash tests/driver_sx127x application on a nucleo-64 boards (eg. nucleo-l073 or nucleo-l152) that has a sx1272 radio:
```
DRIVER=sx1272 make BOARD=nucleo-l073 -C tests/driver_sx127x flash term
```
- Flash the same application on nucleo-64 that has a sx1276 radio:
```
make BOARD=nucleo-l073 -C tests/driver_sx127x flash term
```
- Put one of the board in RX mode using the `listen` shell command
- Send a packet from the other board using `send` (example `send test`)
- Repeat the operation with board exchanged: the second board in listen mode and the first board sending

For the Semtech-loramac package:
- Flash the tests/pkg_semtech-loramac application on a nucleo with sx1272 radio:
```
DRIVER=sx1272 make BOARD=nucleo-l073 -C tests/pkg_semtech-loramac flash term
```
- Configure the loramac stack:
```
> loramac set deveui 0000000000000000
> loramac set appeui 0000000000000000
> loramac set appkey 00000000000000000000000000000000
```
- Join the network, using OTAA
```
> loramac join otaa
```

The problem is already present in master and 2018.01-branch.

#### Expected results

- For the radio driver: A packet is received on the listening board
- For the semtech loramac package: The device can join the network

#### Actual results

- For the radio driver: No packet is received on the listening board. If one resets radio parameters (on both boards), it works again: `setup 125 12 8`.
- For the semtech loramac package: The device cannot join the network

aabadie:
### Contribution description

This PR provides several enhancements/fixes to the sx127x driver and in particular to the sx1272 variant:
- improve the debug messages output by making them consistent. It uses a `[sx127x] <message>`  pattern everywhere, this improves the readability of debug messages
- remove the `sx127x_rx_chain_calibration` from the sx1272 variant as it is not used in the original semtech code and I think this is causing troubles with the driver
- slightly reorder initialization sequence (similar to the semtech device driver)

I started to work on that because I had problem when using the sx1272 with the loramac package ~, it's currently broken in fact (but I'm sure it worked before)~.
During my tests I realized that the driver test application was also broken with sx1272 (unless one explicit uses the `setup` command once). At least this PR fixes that ~but there's still broken things (using SF7 doesn't work actually, and I think this explains why loramac doesn't work with sx1272)~.

### Issues/PRs references

fixes #8935

### DIFF
--- a/drivers/sx127x/sx127x.c
+++ b/drivers/sx127x/sx127x.c
@@ -115,23 +115,22 @@ int sx127x_init(sx127x_t *dev)
 
 void sx127x_init_radio_settings(sx127x_t *dev)
 {
-    sx127x_set_freq_hop(dev, LORA_FREQUENCY_HOPPING_DEFAULT);
-    sx127x_set_iq_invert(dev, LORA_IQ_INVERTED_DEFAULT);
+    sx127x_set_modem(dev, SX127X_MODEM_DEFAULT);
+    sx127x_set_channel(dev, SX127X_CHANNEL_DEFAULT);
     sx127x_set_bandwidth(dev, LORA_BW_DEFAULT);
     sx127x_set_spreading_factor(dev, LORA_SF_DEFAULT);
     sx127x_set_coding_rate(dev, LORA_CR_DEFAULT);
-    sx127x_set_fixed_header_len_mode(dev, LORA_FIXED_HEADER_LEN_MODE_DEFAULT);
     sx127x_set_crc(dev, LORA_PAYLOAD_CRC_ON_DEFAULT);
-    sx127x_set_symbol_timeout(dev, LORA_SYMBOL_TIMEOUT_DEFAULT);
-    sx127x_set_preamble_length(dev, LORA_PREAMBLE_LENGTH_DEFAULT);
-    sx127x_set_payload_length(dev, LORA_PAYLOAD_LENGTH_DEFAULT);
+    sx127x_set_freq_hop(dev, LORA_FREQUENCY_HOPPING_DEFAULT);
     sx127x_set_hop_period(dev, LORA_FREQUENCY_HOPPING_PERIOD_DEFAULT);
-
+    sx127x_set_fixed_header_len_mode(dev, LORA_FIXED_HEADER_LEN_MODE_DEFAULT);
+    sx127x_set_iq_invert(dev, LORA_IQ_INVERTED_DEFAULT);
+    sx127x_set_payload_length(dev, LORA_PAYLOAD_LENGTH_DEFAULT);
+    sx127x_set_tx_power(dev, SX127X_RADIO_TX_POWER);
+    sx127x_set_preamble_length(dev, LORA_PREAMBLE_LENGTH_DEFAULT);
+    sx127x_set_symbol_timeout(dev, LORA_SYMBOL_TIMEOUT_DEFAULT);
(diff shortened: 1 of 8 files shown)


aabadie:
For the record, the problems I found are already there in master and 2018.01-branch

drivers/sx127x/sx127x.c
```diff
@@ -252,11 +258,11 @@ static int _init_peripherals(sx127x_t *dev)
     res = spi_init_cs(dev->params.spi, dev->params.nss_pin);
 
     if (res != SPI_OK) {
-        DEBUG("sx127x: error initializing SPI_%i device (code %i)\n",
+        DEBUG("[sx127x] error initializing SPI_%i device (code %i)\n",
```
dylad: error:

drivers/sx127x/sx127x.c
```diff
@@ -252,11 +258,11 @@ static int _init_peripherals(sx127x_t *dev)
...
                   dev->params.spi, res);
         return 0;
     }
 
-    DEBUG("sx127x: peripherals initialized with success\n");
+    DEBUG("[sx127x] peripherals initialized with success\n");
```
dylad: peripheral ?

aabadie: There are several peripheral initialized by this function: gpios and spi. But I admit that only the spi initialization return status is taken into account. I'll change that.

drivers/sx127x/sx127x.c
aabadie: Will rephrase

dylad:
BTW @aabadie, 
Could you open an issue with all the problems you encounter with SX1272 please ?
It would be great to fix everything for the release.

aabadie:
> Could you open an issue with all the problems you encounter with SX1272 please ?

Done: #8935

> It would be great to fix everything for the release.

This is what I'm trying to do

drivers/sx127x/sx127x.c
aabadie: Fixed

drivers/sx127x/sx127x.c
aabadie: refactored the error management in the drivers and test application

aabadie:
@dylad, is it ok for you now ?

dylad:
@aabadie This look good but I would like to run some tests monday or tuesday first before ACK.

kaspar030:
@dylad Monday would be nice. :)

Or would it make sense to split out any fixes if the rest won't make feature freeze, which is today?)

dylad:
@kaspar030 Don't worry, my setup is ready. I'll test that right now !

kaspar030:
> @kaspar030 Don't worry, my setup is ready. I'll test that right now !

Awesome!

PR #8927, opened 2018-04-11
tests/periph_rtt: Initialize timer target from current reading

gebart:
### Contribution description

This avoids false test failures when the RTT retains its old counter value across reboots/resets.

### Issues/PRs references

Detected while refactoring the RTT driver on Kinetis

### DIFF
--- a/tests/periph_rtt/main.c
+++ b/tests/periph_rtt/main.c
@@ -23,6 +23,7 @@
 
 #include <stdio.h>
 #include <stdint.h>
+#include <inttypes.h>
 
 #include "cpu.h"
 #include "periph_conf.h"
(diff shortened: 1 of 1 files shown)


tests/periph_rtt/main.c
```diff
@@ -51,9 +52,12 @@ int main(void)
     puts("Initializing the RTT driver");
     rtt_init();
 
-    puts("Setting initial alarm");
-    last = TICKS_TO_WAIT;
-    rtt_set_alarm(TICKS_TO_WAIT, cb, 0);
+    uint32_t now = rtt_get_counter();
+    printf("RTT now: %" PRIu32 "\n", now);
+
+    last = now + TICKS_TO_WAIT;
```
aabadie: what about also masking the `last` value like it's done in the callback function ? Maybe the 'set alarm' part could be factorized.
I'll test on ST based platform tomorrow but otherwise it looks ok.

gebart: Added masking as suggested.

gebart:
squashed

PR #8929, opened 2018-04-11
teensy31: Fix off-by-1 error in clock dividers

gebart:
### Contribution description

The comment says the config should yield 24 MHz flash clock, but the
settings were configured to divide-by-3. (48 MHz / 3)

### DIFF
--- a/boards/teensy31/include/periph_conf.h
+++ b/boards/teensy31/include/periph_conf.h
@@ -44,7 +44,7 @@ static const clock_config_t clock_config = {
      * should have better accuracy than the internal slow clock, and lower power
      * consumption than using the 16 MHz crystal and the OSC0 module */
     .clkdiv1 = SIM_CLKDIV1_OUTDIV1(0) | SIM_CLKDIV1_OUTDIV2(0) |
-               SIM_CLKDIV1_OUTDIV3(2) | SIM_CLKDIV1_OUTDIV4(2),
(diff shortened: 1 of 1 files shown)


astralien3000:
In that case the error is also [there](https://github.com/gebart/RIOT/blob/4ba9d2b28018f814f71b6578ceb451c72a8b71e0/boards/mulle/include/periph_conf.h#L50), if I understand well (yes, I copy/pasted ^^).

gebart:
Thanks! I will post a PR with a fix there too.

astralien3000:
You are doing that in an other PR, I can go with this one ?

gebart:
@astralien3000 by the way, do you have RTC crystals on all your Teensy 3.1 boards? I noticed that the schematics on https://www.pjrc.com/teensy/schematic.html say that the 32 kHz crystal is not placed from factory but has to be mounted by the user. Maybe it would be better to have the default configuration use the internal reference oscillator (FEI mode) instead of the external crystal (FEE mode)?

astralien3000:
I don't think I have the RTC crystal on my teensy. But well, I was only able to use USB with the PEE mode.

gebart:
@astralien3000 then, is this default config broken? Does your teensy boot with this config?

astralien3000:
It does boot correctly in the current (master branch) configuration.
The PEE config is not merged because USB need more work.

PR #8928, opened 2018-04-11
kinetis: Refactor clock initialization code

gebart:
### Contribution description

- Move RTC oscillator configuration to the clock_config struct and handle it from kinetis_mcg_init
- Modularize kinetis_mcg_init to improve readability
- Merge boolean settings in clock_config_t into a single clock_flags field

### Issues/PRs references

Used by #8814, #8930, #8933

PR #8930, opened 2018-04-11
kinetis: Refactor RTT driver

gebart:
### Contribution description

- Keep counter value between resets
- Let kinetis_mcg_init manage the RTC oscillator, to avoid disrupting the core clock in certain configurations
- Remove some unnecessary macros for hardware abstraction; all Kinetis CPUs which have an RTC only have a single RTC instance, and the ISRs and hardware registers are named the same way in all Kinetis CPU headers.

### Issues/PRs references

Depends on #8928

### DIFF
--- a/boards/frdm-k22f/include/periph_conf.h
+++ b/boards/frdm-k22f/include/periph_conf.h
@@ -265,21 +265,6 @@ static const spi_conf_t spi_config[] = {
 #define I2C_0_PORT_CFG               (PORT_PCR_MUX(I2C_0_PIN_AF) | PORT_PCR_ODE_MASK)
 /** @} */
 
-/**
-* @name RTT and RTC configuration
-* @{
-*/
-#define RTT_NUMOF                    (1U)
-#define RTC_NUMOF                    (1U)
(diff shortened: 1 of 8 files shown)


gebart:
Rebased after #8928 was merged

aabadie:
Well, this PR was also touching periph_conf files that will probably be changed with I2C rework... Why not waiting ?

kYc0o:
I think what we need is a certitude that the embargo is also in place for **any** change on all the periph_conf files. Therefore, if that's the case, we urge to have the I2C branch in really super fast, otherwise we block more than a half of the current PRs.

aabadie:
You just have to look in the periph_conf if there's already some I2C configured, which was the case here. I think @gebart will probably take the I2C rework opportunity for updating old style (using defines) to new style (using array of struct) I2C configuration.
Unfortunately, the I2C configuration is just above the RTT config so there's a high chance that it will raise a conflict.

> we urge to have the I2C branch in really super fast, otherwise we block more than a half of the current PRs

+1. So what are we waiting for ?

PR #8932, opened 2018-04-11
nanocoap: add server-side block2 support

bergzand:
### Contribution description

Making use of the new option handling and the work done by @kaspar030 with block1 support, this PR extends that to block2 support. The `/.well-known/core` URI is also adapted to support block2.

commit 54b275c (block2 support, not the example) adds 412 bytes to the text size of the nanocoap example on a samr21-xpro

### Issues/PRs references

Depends on #8788

### DIFF
--- a/examples/nanocoap_server/coap_handler.c
+++ b/examples/nanocoap_server/coap_handler.c
@@ -17,11 +17,49 @@
 /* internal value that can be read/written via CoAP */
 static uint8_t internal_value = 0;
 
+static const uint8_t block2_intro[] = "This is RIOT (Version: ";
+static const uint8_t block2_board[] = " running on a ";
+static const uint8_t block2_mcu[] = " board with a ";
+
 static ssize_t _riot_board_handler(coap_pkt_t *pkt, uint8_t *buf, size_t len, void *context)
 {
     (void)context;
     return coap_reply_simple(pkt, COAP_CODE_205, buf, len,
             COAP_FORMAT_TEXT, (uint8_t*)RIOT_BOARD, strlen(RIOT_BOARD));
+}
+
+static ssize_t _riot_block2_handler(coap_pkt_t *pkt, uint8_t *buf, size_t len, void *context)
+{
+    (void)context;
+    coap_block_slicer_t slicer;
+    coap_block2_init(pkt, &slicer);
+    uint8_t *payload = buf + coap_get_total_hdr_len(pkt);
+
+    uint8_t *bufpos = payload;
+
+    bufpos += coap_put_option_ct(bufpos, 0, COAP_FORMAT_TEXT);
+    bufpos += coap_opt_put_block2(bufpos, COAP_OPT_CONTENT_FORMAT, &slicer, 1);
+    *bufpos++ = 0xff;
+
+    /* Add actual content */
+    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, block2_intro, sizeof(block2_intro));
+    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, (uint8_t*)RIOT_VERSION, sizeof(RIOT_VERSION));
+    bufpos += coap_blockwise_put_char(&slicer, bufpos, ')');
+    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, block2_board, sizeof(block2_board));
+    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, (uint8_t*)RIOT_BOARD, sizeof(RIOT_BOARD));
+    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, block2_mcu, sizeof(block2_mcu));
+    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, (uint8_t*)RIOT_MCU, sizeof(RIOT_MCU));
+    /* To demonstrate individual chars */
+    bufpos += coap_blockwise_put_char(&slicer, bufpos, ' ');
+    bufpos += coap_blockwise_put_char(&slicer, bufpos, 'M');
+    bufpos += coap_blockwise_put_char(&slicer, bufpos, 'C');
+    bufpos += coap_blockwise_put_char(&slicer, bufpos, 'U');
+    bufpos += coap_blockwise_put_char(&slicer, bufpos, '.');
+
+
+    unsigned payload_len = bufpos - payload;
+    return coap_block2_build_reply(pkt, COAP_CODE_205,
+                                   buf, len, payload_len, &slicer);
 }
 
 static ssize_t _riot_value_handler(coap_pkt_t *pkt, uint8_t *buf, size_t len, void *context)
@@ -109,6 +147,7 @@ const coap_resource_t coap_resources[] = {
     COAP_WELL_KNOWN_CORE_DEFAULT_HANDLER,
     { "/riot/board", COAP_GET, _riot_board_handler, NULL },
     { "/riot/value", COAP_GET | COAP_PUT | COAP_POST, _riot_value_handler, NULL },
+    { "/riot/ver", COAP_GET, _riot_block2_handler, NULL },
     { "/sha256", COAP_POST, _sha256_handler, NULL },
 };
 
--- a/sys/include/net/nanocoap.h
+++ b/sys/include/net/nanocoap.h
@@ -12,6 +12,31 @@
  * @ingroup     net
  * @brief       Provides CoAP functionality optimized for minimal resource usage
  *
+ * # Create a Block-wise Response (Block2)
+ *
+ * Block-wise is a CoAP extension (RFC 7959) to divide a large payload across
+ * multiple physical packets. This section describes how to write a block-wise
+ * payload for a response, and is known as Block2. (Block1 is for a block-wise
+ * payload in a request.) See _riot_board_handler() in the nanocoap_server
+ * example for an example handler implementation.
+ *
+ * Start with coap_block2_init() to read the client request and initialize a
+ * coap_slicer_t struct with the size and location for this slice of the
+ * overall payload. Then write the block2 option in the response with
+ * coap_opt_put_block2(). The option includes an indicator ("more") that a
+ * slice completes the overall payload transfer. You may not know the value for
+ * _more_ at this point, but you must initialize the space in the packet for
+ * the option before writing the payload. The option is rewritten later.
+ *
+ * Next, use the coap_blockwise_put_xxx() functions to write the payload
+ * content. These functions use the coap_block_slicer_t to enable or disable
+ * actually writing the content, depending on the current position within the
+ * overall payload transfer.
+ *
+ * Finally, use the convenience function coap_block2_build_reply(), which
+ * finalizes the packet and calls coap_block2_finish() internally to update
+ * the block2 option.
+ *
  * @{
  *
  * @file
@@ -62,8 +87,10 @@ extern "C" {
  * @name    Nanocoap specific maximum values
  * @{
  */
-#define NANOCOAP_NOPTS_MAX      (16)
-#define NANOCOAP_URI_MAX        (64)
+#define NANOCOAP_NOPTS_MAX          (16)
+#define NANOCOAP_URI_MAX            (64)
+#define NANOCOAP_BLOCK_SIZE_EXP_MAX  (6)  /**< Maximum size for a blockwise
+                                            *  transfer as power of 2 */
 /** @} */
 
 #ifdef MODULE_GCOAP
@@ -144,6 +171,16 @@ typedef struct {
     int more;                       /**< -1 for no option, 0 for last block,
                                           1 for more blocks coming          */
 } coap_block1_t;
+
+/**
+ * @brief Blockwise transfer helper struct
+ */
+typedef struct {
+    size_t start;                   /**< Start offset of the current block  */
+    size_t end;                     /**< End offset of the current block    */
+    size_t cur;                     /**< Offset of the generated content    */
+    uint8_t *opt;                   /**< Pointer to the placed option       */
+} coap_block_slicer_t;
 
 /**
  * @brief   Global CoAP resource list
@@ -412,6 +449,17 @@ int coap_get_blockopt(coap_pkt_t *pkt, uint16_t option, uint32_t *blknum, unsign
 int coap_get_block1(coap_pkt_t *pkt, coap_block1_t *block1);
 
 /**
+ * @brief    Block2 option getter
+ *
+ * @param[in]   pkt     pkt to work on
+ * @param[out]  block2  ptr to preallocated coap_block1_t structure
+ *
+ * @returns     0 if block2 option not present
+ * @returns     1 if structure has been filled
+ */
+int coap_get_block2(coap_pkt_t *pkt, coap_block1_t *block2);
+
+/**
  * @brief   Insert block1 option into buffer
  *
  * @param[out]  buf         buffer to write to
@@ -489,6 +537,23 @@ ssize_t coap_opt_add_uint(coap_pkt_t *pkt, uint16_t optnum, uint32_t value);
  * @return        total number of bytes written to buffer
  */
 ssize_t coap_opt_finish(coap_pkt_t *pkt, uint16_t flags);
+
+/**
+ * @brief   Insert block2 option into buffer
+ *
+ * When calling this function to initialize a packet with a block2 option, the
+ * more flag must be set to prevent the creation of an option with a length too
+ * small to contain the size bit.
+ *
+ * @param[out]  buf         buffer to write to
+ * @param[in]   lastonum    number of previous option (for delta calculation),
+ *                          must be < 23
+ * @param[in]   slicer      coap blockwise slicer helper struct
+ * @param[in]   more        more flag (1 or 0)
+ *
+ * @returns     amount of bytes written to @p buf
+ */
+size_t coap_opt_put_block2(uint8_t *buf, uint16_t lastonum, coap_block_slicer_t *slicer, bool more);
 
 /**
  * @brief   Get content type from packet
@@ -603,6 +668,83 @@ static inline ssize_t coap_get_location_query(const coap_pkt_t *pkt,
     return coap_opt_get_string(pkt, COAP_OPT_LOCATION_QUERY,
                                target, max_len, '&');
 }
+
+/**
+ * @brief Initialize a block2 slicer struct for writing the payload
+ *
+ * This function determines the size of the response payload based on the
+ * size requested by the client in @p pkt.
+ *
+ * @param[in]   pkt         packet to work on
+ * @param[out]  slicer      Preallocated slicer struct to fill
+ */
+void coap_block2_init(coap_pkt_t *pkt, coap_block_slicer_t *slicer);
+
+/**
+ * @brief Finish a block2 response
+ *
+ * This function finalizes the block2 response header
+ *
+ * Checks whether the `more` bit should be set in the block2 option and
+ * sets/clears it if required.  Doesn't return the number of bytes as this
+ * overwrites bytes in the packet, it doesn't add new bytes to the packet.
+ *
+ * @param[inout]  slicer      Preallocated slicer struct to use
+ */
+void coap_block2_finish(coap_block_slicer_t *slicer);
+
+/**
+ * @brief   Build reply to CoAP block2 request
+ *
+ * This function can be used to create a reply to a CoAP block2 request
+ * packet. In addition to @ref coap_build_reply, this function checks the
+ * block2 option and returns an error message to the client if necessary.
+ *
+ * @param[in]   pkt         packet to reply to
+ * @param[in]   code        reply code (e.g., COAP_CODE_204)
+ * @param[out]  rbuf        buffer to write reply to
+ * @param[in]   rlen        size of @p rbuf
+ * @param[in]   payload_len length of payload
+ * @param[in]   slicer      slicer to use
+ *
+ * @returns     size of reply packet on success
+ * @returns     <0 on error
+ */
+ssize_t coap_block2_build_reply(coap_pkt_t *pkt, unsigned code,
+                                uint8_t *rbuf, unsigned rlen, unsigned payload_len,
+                                coap_block_slicer_t *slicer);
+
+/**
+ * @brief Add a single character to a block2 reply.
+ *
+ * This function is used to add single characters to a CoAP block2 reply. It
+ * checks whether the character should be added to the buffer and ignores it
+ * when the character is outside the current block2 request.
+ *
+ * @param[in]   slicer      slicer to use
+ * @param[in]   bufpos      pointer to the current payload buffer position
+ * @param[in]   c           character to write
+ *
+ * @returns     Number of bytes writen to @p bufpos
+ */
+size_t coap_blockwise_put_char(coap_block_slicer_t *slicer, uint8_t *bufpos, char c);
+
+/**
+ * @brief Add a byte array to a block2 reply.
+ *
+ * This function is used to add an array of bytes to a CoAP block2 reply. it
+ * checks which parts of the string should be added to the reply and ignores
+ * parts that are outside the current block2 request.
+ *
+ * @param[in]   slicer      slicer to use
+ * @param[in]   bufpos      pointer to the current payload buffer position
+ * @param[in]   c           byte array to copy
+ * @param[in]   len         length of the byte array
+ *
+ * @returns     Number of bytes writen to @p bufpos
+ */
+size_t coap_blockwise_put_bytes(coap_block_slicer_t *slicer, uint8_t *bufpos,
+                                const uint8_t *c, size_t len);
 
 /**
  * @brief   Helper to decode SZX value to size in bytes
--- a/sys/net/application_layer/nanocoap/nanocoap.c
+++ b/sys/net/application_layer/nanocoap/nanocoap.c
@@ -546,11 +546,39 @@ size_t coap_put_option_ct(uint8_t *buf, uint16_t lastonum, uint16_t content_type
     }
 }
 
+static unsigned _size2szx(size_t size)
+{
+    unsigned szx = 0;
+    assert(size <= 1024);
+
+    while (size) {
+        size = size >> 1;
+        szx++;
+    }
+    /* Size exponent + 1 */
+    assert(szx >= 5);
+    return szx - 5;
+}
+
+static unsigned _slicer_blknum(coap_block_slicer_t *slicer)
+{
+    size_t blksize = slicer->end - slicer->start;
+    size_t start = slicer->start;
+    unsigned blknum = 0;
+
+    while (start > 0) {
+        start -= blksize;
+        blknum++;
+    }
+    return blknum;
+}
+
 static size_t coap_put_option_block(uint8_t *buf, uint16_t lastonum, unsigned blknum, unsigned szx, int more, uint16_t option)
 {
     uint32_t blkopt = (blknum << 4) | szx | (more ? 0x8 : 0);
     size_t olen = _encode_uint(&blkopt);
-    return coap_put_option(buf, lastonum, option, (uint8_t*)&blkopt, olen);
+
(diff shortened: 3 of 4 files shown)


sys/include/net/nanocoap.h
```diff
@@ -263,6 +276,29 @@ typedef struct {
...
+typedef struct {
+    size_t start;                   /**< Start offset of the current block  */
+    size_t end;                     /**< End offset of the current block    */
+    size_t cur;                     /**< Offset of the  generated content   */
+    uint8_t *opt;                   /**< Pointer to the placed option       */
+} coap_blockhelper_t;
+
+/**
+>>>>>>> 01c4c5265... nanocoap: Add server side block2 support
+>>>>>>> nanocoap: Add server side block2 support
```
miri64: Another one ;-).

bergzand: First :coffee:, otherwise half this PR is going to consist of me removing merge conflict indicators :disappointed:

bergzand: I hope they're all gone now

kb2ma:
Making progress on review. See [WIP](https://github.com/kb2ma/RIOT/commit/a267749e38d015b81269988aefb182462ced249a) as C-style comments in code. Overall approach and structure seem fine. Right now I'm thinking about how the block payload is written.

I plan to finish the merge for #8920, and then finish this review. That PR provides examples for the naming of option functions, which I think we should follow here.

sys/include/net/nanocoap.h
```diff
@@ -49,6 +49,7 @@ extern "C" {
  */
 #define NANOCOAP_NOPTS_MAX      (16)
 #define NANOCOAP_URI_MAX        (64)
+#define NANOCOAP_BLOCK_SZX_MAX  (6)
```
kb2ma: Do we really need this nanocoap-specific definition? Why not redefine COAP_BLOCKWISE_SZX_MAX to 6. According to sec 2.2, the allowed values are 0 to 6.

sys/include/net/nanocoap.h
```diff
@@ -498,6 +520,20 @@ size_t coap_put_option_block1(uint8_t *buf, uint16_t lastonum, unsigned blknum,
  */
 size_t coap_put_block1_ok(uint8_t *pkt_pos, coap_block1_t *block1, uint16_t lastonum);
 
+/**
```
kb2ma: Please change to coap_opt_put_block2() to be consistent with #9085. We'll need to change the equivalent block1 function name in a follow-on PR.

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -550,6 +550,11 @@ int coap_get_block1(coap_pkt_t *pkt, coap_block1_t *block1)
     return (block1->more >= 0);
 }
 
+int coap_get_block2(coap_pkt_t *pkt, coap_block1_t *block2)
+{
+    return coap_get_blockopt(pkt, COAP_OPT_BLOCK2, &block2->blknum, &block2->szx);
```
kb2ma: Populate the block2 'more' attribute?

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -595,32 +605,106 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
     return bufpos - buf;
 }
 
+size_t coap_block2_init(uint8_t* buf, uint16_t lastonum, coap_pkt_t *pkt, coap_blockhelper_t *blk)
+{
+    uint32_t blknum;
+    unsigned szx;
+    if(coap_get_blockopt(pkt, COAP_OPT_BLOCK2, &blknum, &szx) >= 0) {
```
kb2ma: Add a comment that we are reading the client request?

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -595,32 +605,106 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
...
+    int more = (blk->cur > blk->end) ? 0x80 : 0;
+
+    coap_put_option_block2(blk->opt, COAP_OPT_BLOCK2 - delta, blknum, szx, more);
+    return coap_build_reply(pkt, code, rbuf, rlen, payload_len);
+}
+
+size_t coap_blockwise_put_char(coap_blockhelper_t *blk, uint8_t *bufpos, char c)
+{
+    /* Only copy the char if it is within the window */
+    if (blk->start <=  blk->cur && blk->cur < blk->end) {
```
kb2ma: Need parens for clarity.

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -595,32 +605,106 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
...
+size_t coap_blockwise_put_bytes(coap_blockhelper_t *blk, uint8_t *bufpos,
+                                 const uint8_t *c, size_t len)
+{
+    size_t str_len = 0;    /* Length of the string to copy */
+
+    /* Calculate start offset of the supplied string */
+    size_t str_offset = (blk->start > blk->cur) ? blk->start - blk->cur : 0;
+
+    /* Check for string before or beyond window */
+    if (blk->cur >= blk->end || str_offset > len) {
```
kb2ma: Need parens for clarity, also for l.671.

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -595,32 +605,106 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
...
     uint8_t *bufpos = payload;
 
     bufpos += coap_put_option_ct(bufpos, 0, COAP_CT_LINK_FORMAT);
+    bufpos += coap_block2_init(bufpos, COAP_OPT_CONTENT_FORMAT, pkt, &blk);
     *bufpos++ = 0xff;
 
     for (unsigned i = 0; i < coap_resources_numof; i++) {
         if (i) {
-            *bufpos++ = ',';
+            bufpos += coap_blockwise_put_char(&blk, bufpos, ',');
```
kb2ma: Does it make sense to have a single function that handles both a block write as well as a plain payload write (like *bufpos++ )? This would help avoid buffer overruns in both situations. Handles a plain payload if the blk pointer is NULL. How about calling it 'coap_payload_put_\[char|bytes\]()?

sys/include/net/nanocoap.h
```diff
@@ -285,6 +286,16 @@ typedef struct {
                                           1 for more blocks coming          */
 } coap_block1_t;
 
+/**
```
kb2ma: Looking at use of this struct, would it be more descriptive to call it 'coap_blockbuilder_t' or something like that?

sys/include/net/nanocoap.h
```diff
@@ -285,6 +286,16 @@ typedef struct {
...
+ * @brief Blockwise transfer helper struct
+ */
+typedef struct {
+    size_t start;                   /**< Start offset of the current block  */
+    size_t end;                     /**< End offset of the current block    */
+    size_t cur;                     /**< Offset of the  generated content   */
```
kb2ma: Extra space between 'the' and 'generated'.

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -595,32 +605,106 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
     return bufpos - buf;
 }
 
+size_t coap_block2_init(uint8_t* buf, uint16_t lastonum, coap_pkt_t *pkt, coap_blockhelper_t *blk)
+{
+    uint32_t blknum;
+    unsigned szx;
+    if(coap_get_blockopt(pkt, COAP_OPT_BLOCK2, &blknum, &szx) >= 0) {
+        /* Use the smallest block size */
+        if (NANOCOAP_BLOCK_SZX_MAX - 4 < szx ) {
```
kb2ma: I'm confused, what does 'smallest' mean here? The smallest value of szx is 0. Also, watch the spacing in l.614 and 615.

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -595,32 +605,106 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
...
+        }
+    }
+    blk->opt = buf;
+    blk->start = blknum * coap_szx2size(szx);
+    blk->end = blk->start + coap_szx2size(szx);
+    blk->cur = 0;
+    return coap_put_option_block2(buf, lastonum, blknum, szx, 1);
+}
+
+ssize_t coap_block2_build_reply(coap_pkt_t *pkt, unsigned code,
```
kb2ma: Why not refactor the code here to create a coap_block2_finish() function. This would pair semantically with coap_block2_init(), and allow gcoap to use the function as well. Then this function, if still worthwhile, could just call coap_block2_finish. Please see comment about the blockwise put functions though.

bergzand:
Thanks for the review @kb2ma, I'll try to fix and rebase this PR asap.

sys/net/application_layer/nanocoap/nanocoap.c
bergzand: I'll clarify the comment here. This should describe that the block size used is the smallest block size among what the client requested and the maximum that nanocoap can handle. ([rfc7959 - sec 2.4 second paragraph](https://tools.ietf.org/html/rfc7959#section-2.4))

sys/include/net/nanocoap.h
bergzand: I'm also changing the `coap_get_block2` to `coap_opt_get_block2`

sys/include/net/nanocoap.h
bergzand: Done

sys/include/net/nanocoap.h
bergzand: I've added some comments. Please let me know what you think and if we should handle this differently

sys/net/application_layer/nanocoap/nanocoap.c
bergzand: I think I did what you requested. Please let me know if the new situation is usable for gcoap

bergzand:
@kb2ma I think I've fixed all of your issues besides your suggestion for a generic put_bytes. I have to take a look at that one later and see how the additional checks would change things.

sys/include/net/nanocoap.h
kb2ma: I get it now, but then what should this value represent? When I see 'SZX' in the name, I think of the value in the block option value (exponent - 4), not the value of the exponent. My comments in the implementation of coap_block2_init() show the confusion.

So, from my perspective the value of NANOCOAP_BLOCK_SZX_MAX should be 2 to map directly to the meaning of coap_block1_t.szx and COAP_BLOCKWISE_SZX_MAX. Using 2 for the value also would save the two subtractions in coap_block2_init().

Otherwise, it would make more sense to me to call this macro NANOCOAP_BLOCK_EXP_MAX, or something like that.

sys/net/application_layer/nanocoap/nanocoap.c
kb2ma: Exactly what I was thinking, thanks.

sys/include/net/nanocoap.h
kb2ma: coap_opt_put block2() looks good. See my comment on coap_opt_get_block2() though.

sys/net/application_layer/nanocoap/nanocoap.c
kb2ma: See comment above on definition of NANOCOAP_BLOCK_SZX_MAX in nanocoap.h. We can resolve there.

sys/net/application_layer/nanocoap/nanocoap.c
kb2ma: Agree to resolve later as you look into it further.

sys/include/net/nanocoap.h
bergzand: I understand and agree with your arguments here. How about naming it `NANOCOAP_BLOCK_SIZE_EXP_MAX`? I'm against using the raw SZX value (2 in this case) because I want it to be as human readable as possible without too much room for mistakes.

sys/include/net/nanocoap.h
```diff
@@ -762,6 +799,91 @@ static inline ssize_t coap_get_location_query(const coap_pkt_t *pkt,
...
+ * preceding options are added
+ *
+ * @param[in]   buf         buffer to write to
+ * @param[in]   lastonum    last option number (must be <27)
+ * @param[in]   pkt         packet to work on
+ * @param[out]  blk         Preallocated blockbuilder struct to fill
+ *
+ * @returns     amount of bytes written to @p buf
+ */
+size_t coap_block2_init(uint8_t* buf, uint16_t lastonum, coap_pkt_t *pkt, coap_blockbuilder_t *blk);
```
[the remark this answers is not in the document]
bergzand: Let me try something here, I think I understand what you want and it makes sense to me.

bergzand:
@kb2ma What makes more sense to you, passing the individual block values (size, block number and "more") to the `coap_opt_put_block2` or passing the `coap_blockbuilder_t` struct?

sys/include/net/nanocoap.h
```diff
@@ -569,6 +581,17 @@ int coap_get_blockopt(coap_pkt_t *pkt, uint16_t option, uint32_t *blknum, unsign
...
+/**
+ * @brief    Block2 option getter
+ *
+ * @param[in]   pkt     pkt to work on
+ * @param[out]  block1  ptr to preallocated coap_block1_t structure
+ *
+ * @returns     0 if block2 option not present
+ * @returns     1 if structure has been filled
+ */
+int coap_opt_get_block2(coap_pkt_t *pkt, coap_block1_t *block2);
```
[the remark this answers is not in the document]
kb2ma: Yes, please. Apologies for my contribution to any misunderstanding.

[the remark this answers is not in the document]
bergzand: No worries :)

sys/include/net/nanocoap.h
kb2ma: Name sounds fine. I would expect further development over negotiation of the size of the payload, so it will be helpful to think in those terms instead of the adjusted value encoded in the option.

kb2ma:
I think passing individual values makes sense at the lowest levels -- coap_get_blockopt() and coap put_option_block() (eventually to become coap_opt_put_block()). However, above that it seems more convenient to use a single coap_block1_t rather than its attributes. (I assume you meant coap_block1_t rather then coap_blockbuilder_t.)

So, yes, I think it makes more sense to use a coap_block1_t  in coap_opt_put_block2().

bergzand:
> At the risk of you throwing things at me, would coap_block_serializer_t or coap_block_payload_t be better than coap_blockbuilder_t? I'm not asking you to change this, just thinking about the best name as my understanding evolves. [ducking and running]

As long as this PR is not yet finalized I don't mind you suggesting new names. How about `coap_block_slicer_t`. I think serializer here is technically incorrect since the payload is not really transformed. I suggested slicer here since the payload is sliced into individual segments

kb2ma:
> How about coap_block_slicer_t.

+1! Nicely descriptive

sys/include/net/nanocoap.h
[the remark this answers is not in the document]
kb2ma: Following up on my comments above. I created #10036 with nanocoap module documentation. Might be worth a look.

bergzand:
> +1! Nicely descriptive

Done!

Now back to the other issues remaining :)

sys/include/net/nanocoap.h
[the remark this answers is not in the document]
bergzand: @kb2ma I've reworked it to use only a `coap_block_slicer_t` here. It required 2 helper function which added 20 bytes of additional flash compared to previous fixup state. This prevents us from having to put that `coap_block1_t` struct on the stack during the full handler processing.

Please let me know if this is workable for gcoap since I didn't completely follow your instructions :)

bergzand:
@kb2ma I think I've fixed all your comments so far, please let me know if I skipped one.

kb2ma:
Thanks for the update, @bergzand. I am reviewing as I write this. My goal is a concrete response as soon as possible.

sys/include/net/nanocoap.h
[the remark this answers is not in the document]
kb2ma: Still using old name. I fixed in [7a67661](https://github.com/bergzand/RIOT/pull/3/commits/7a67661e7774b5e6611b1a6b71d87a7b4d548512).

sys/include/net/nanocoap.h
```diff
@@ -762,6 +802,91 @@ static inline ssize_t coap_get_location_query(const coap_pkt_t *pkt,
...
+ * preceding options are added
+ *
+ * @param[in]   buf         buffer to write to
+ * @param[in]   lastonum    last option number (must be <27)
+ * @param[in]   pkt         packet to work on
+ * @param[out]  slicer      Preallocated slicer struct to fill
+ *
+ * @returns     amount of bytes written to @p buf
+ */
+void coap_block2_init(coap_pkt_t *pkt, coap_block_slicer_t *slicer);
```
kb2ma: Documentation out of sync with declaration. I fixed in [b6b452d](https://github.com/bergzand/RIOT/pull/3/commits/b6b452d6af2d14d78b09a04749d7d8bfee6b7213).

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -691,32 +731,115 @@ ssize_t coap_opt_finish(coap_pkt_t *pkt, uint16_t flags)
...
+    slicer->end = slicer->start + coap_szx2size(szx);
+    slicer->cur = 0;
+}
+
+void coap_block2_finish(coap_pkt_t *pkt, coap_block_slicer_t *slicer)
+{
+    int option_len;
+    uint16_t delta;
+    /* Retrieve the block2 option from the constructed packet */
+    _parse_option(pkt, slicer->opt, &delta, &option_len);
```
kb2ma: Use of pkt is not valid here. The pkt referred to here is the _response_ packet, not the request packet as in coap_block2_init(). This struct has not been initialized for the response. In fact, when using the buffer-based functions for the response, the pkt never is updated for the response.

All we really need to determine is the delta value. We can directly use _decode_value(). The only issue is the third parameter, which points to the beginning of the payload to be sure the function does not read past the end of the header. We *know* it won't read past the end of the header because the option already was written before getting here. So, we can just point to the location past the longest possible value for a block2 option. 

That's what I did in [81a818b](https://github.com/bergzand/RIOT/pull/3/commits/81a818b55dff53267c087a12bf2b1e917e71b53d).

bergzand:
> Besides the issues above, I also added module documentation in [d1e999d](https://github.com/bergzand/RIOT/pull/3/commits/d1e999d42bd2a0b4abb531fc583f51180cb8f261). This documentation seems odd on its own, but I'll combine with #10036 when that's merged in the next few days. Also, I fixed some typos in [c126ca8](https://github.com/bergzand/RIOT/pull/3/commits/c126ca85df33a7d8806df7c7023b15e942165fe6).

Sounds good to me

> With these changes, no failures when testing. I had not really used block before this, but the design and @bergzand's implementation are well done! I used the nanocoap server example to test /.well-known/core and /riot/description. I tried adjusting the lengths of the output to catch off-by-1 problems. I tried various block lengths (16, 32, etc.).

Thanks, good to hear

> Let's wrap this up and get it merged! If @bergzand is OK with these fixups, from my perspective all we need is to rebase to get the new coap header file.

I put some small fixes in the first paragraph of the module docs (mainly block -> block-wise) to match the rfc wording a bit closer.

sys/include/net/nanocoap.h
```diff
@@ -12,6 +12,31 @@
  * @ingroup     net
  * @brief       Provides CoAP functionality optimized for minimal resource usage
  *
+ * # Create a Block-wise Response (Block2)
+ *
+ * Block-wise is a CoAP extension (RFC 7959) to divide a large payloads across
```
kb2ma: payloads -> payload

Thanks for the update to use "block-wise". I had not really absorbed the use of the term in the RFC. I get the value of it now and agree we should use it similarly.

bergzand: > payloads -> payload

Fixed

> Thanks for the update to use "block-wise". I had not really absorbed the use of the term in the RFC. I get the value of it now and agree we should use it similarly.

I really didn't want to start stacking PR based change requests so I decided to push a fixup after merging your changes.

kb2ma: Thanks for the update and no worries. Output is good and we're getting it done. :-)

kb2ma:
The merge looks good, @bergzand. Please rebase and squash. Even if we need to squash one more time it will be worth it because there are so many commits now.

bergzand:
> The merge looks good, @bergzand. Please rebase and squash. Even if we need to squash one more time it will be worth it because there are so many commits now.

Done!

examples/nanocoap_server/coap_handler.c
```diff
@@ -108,6 +146,7 @@ ssize_t _sha256_handler(coap_pkt_t* pkt, uint8_t *buf, size_t len, void *context
 const coap_resource_t coap_resources[] = {
     COAP_WELL_KNOWN_CORE_DEFAULT_HANDLER,
     { "/riot/board", COAP_GET, _riot_board_handler, NULL },
+    { "/riot/description", COAP_GET, _riot_block2_handler, NULL },
```
kb2ma: _riot_block2_handler() plus the resource added to coap_resources adds as much text to the nanocoap_server executable as all of the block2 functions added to nanocoap.c.

I can't really argue with the implementation of _riot_block2_handler() as an example -- except that "/riot/description" is too long for a CoAP path. So shortened it to "/riot/ver" in [d80d15b](https://github.com/bergzand/RIOT/pull/4/commits/d80d15b6bdef85464121716b96c74989783f5343).

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -704,32 +744,117 @@ ssize_t coap_opt_finish(coap_pkt_t *pkt, uint16_t flags)
...
+void coap_block2_finish(coap_block_slicer_t *slicer)
+{
+    assert(slicer->opt);
+
+    /* The third parameter for _decode_value() points to the end of the header.
+     * We don't know this position, but we know we can read the option because
+     * it's already in the buffer. So just point past the option. */
+    uint8_t *pos = slicer->opt + 1;
+    uint16_t delta = _decode_value(*slicer->opt >> 4, &pos, slicer->opt + 3);
+    int more = (slicer->cur > slicer->end) ? 0x80 : 0;
```
kb2ma: Use 1 rather than 0x80 to clearly indicate the result is a boolean value, as in [223e761](https://github.com/bergzand/RIOT/pull/4/commits/223e761f589f2422e0cfe878b85325737a70f4e3).

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -576,6 +601,13 @@ int coap_get_block1(coap_pkt_t *pkt, coap_block1_t *block1)
     return (block1->more >= 0);
 }
 
+int coap_get_block2(coap_pkt_t *pkt, coap_block1_t *block2)
+{
+    block2->more = coap_get_blockopt(pkt, COAP_OPT_BLOCK2, &block2->blknum,
+            &block2->szx);
+    return block2->more;
```
kb2ma: The return value must indicate if the option is present, as with coap_get_block1(). See the fix in [f466972](https://github.com/bergzand/RIOT/pull/4/commits/f4669724c8c04503db48055a3eddb45016d7a95e).

sys/include/net/nanocoap.h
```diff
@@ -490,6 +538,23 @@ ssize_t coap_opt_add_uint(coap_pkt_t *pkt, uint16_t optnum, uint32_t value);
...
+/**
+ * @brief   Insert block2 option into buffer
+ *
+ * When calling this function to initialize a packet with a block2 option, the
+ * more flag must be set to prevent the creation of an option with a length too
+ * small to contain the size bit.
+ *
+ * @param[out]  buf         buffer to write to
+ * @param[in]   lastonum    number of previous option (for delta calculation),
+ *                          must be < 27
```
kb2ma: Typo, should be 23, as in [e2d3564](https://github.com/bergzand/RIOT/pull/4/commits/e2d3564c1e0335fe6df8f37e2f20ee9830d772e6).

kb2ma:
As @bergzand and I discussed, the block code needs some unit tests. For example, this will verify the fix for the issue with coap_get_block2(), which is not used in the nanocoap_server example. However, we agreed to add these tests in a follow-up.

kb2ma:
Given Murdock [issue](https://ci.riot-os.org/RIOT-OS/RIOT/8932/536b3ba038e22dd2779ac832fa8d265cd4714743/output.html#error0), need to blacklist WSN430? Alternatively could remove the /riot/description example to free text space. Can reference coap_well_known_core_default_handler() to illustrate how to use block2.

examples/nanocoap_server/coap_handler.c
bergzand: I agree with the shorter path

sys/net/application_layer/nanocoap/nanocoap.c
bergzand: Whoops!

sys/net/application_layer/nanocoap/nanocoap.c
bergzand: Ah, must have mixed up the API functions here

sys/include/net/nanocoap.h
bergzand: oops :sob:

bergzand:
> My only suggestion for future work on memory/code space is to consider adding the block value to coap_block_slicer_t. This would remove the need for _size2szx() and _slicer_blknum() at the price of 4 bytes of RAM.

I'm not sure what the correct tradeoff is here. If I compile the example with these functions not flagged with `static`, they occupy 20 and 16 bytes of flash, so we're talking about trading 36B of flash for 4B of stack usage.

> Given Murdock [issue](https://ci.riot-os.org/RIOT-OS/RIOT/8932/536b3ba038e22dd2779ac832fa8d265cd4714743/output.html#error0), need to blacklist WSN430? Alternatively could remove the /riot/description example to free text space. Can reference coap_well_known_core_default_handler() to illustrate how to use block2.

I prefer to have an as simple as possible example for this, IMHO the `coap_well_known_core_default_handler()` is not exactly trivial. In this case I'd prefer to blacklist the two boards, there are still other 16 bit boards that should be able to compile/run the example

kb2ma:
> I'm not sure what the correct tradeoff is here.

I agree that given the relative byte sizes of the tradeoff (RAM vs flash) it's a tough call. At least we've flagged it for future reference.

> In this case I'd prefer to blacklist the two boards

It really gets into larger questions about the content of examples, and it's definitely not worth holding up this PR. For example, it would be easy enough in the future to create a 'block' app in the [applications](https://github.com/RIOT-OS/applications) repository for `/riot/ver` and other uses of block. OTOH, it was definitely worthwhile to look at the flash and RAM sizes before merging.

kb2ma:
So, I'd say were there, and just in time given the soft freeze on 10/16. :-) Please squash.

kb2ma:
OK to blacklist the WSN boards in the squash.

bergzand:
> So, I'd say were there, and just in time given the soft freeze on 10/16. :-) Please squash.

Squashed

> OK to blacklist the WSN boards in the squash.

And added to the insufficient memory list.

kb2ma:
Thanks for the tag, @bergzand. I was just focused on waiting for Murdock to finish. :-)

bergzand:
:tada: thanks for the reviews @kb2ma!

kYc0o:
Yaaay!!! Thanks o everyone involved here! This feature is really really appreciated!

jia200x:
Awesome work! Thanks to everyone involved! :)

PR #8933, opened 2018-04-12
kinetis: Refactor PIT timer driver implementation

gebart:
### Contribution description

Cleaned up the PIT driver implementation significantly. Simplify the implementation by stopping/starting the prescaler channel instead of the counter. The refactor also fixes some random deadline misses in the old driver where the timer would occasionally overshoot by some 0.5-2 ms. The cause of the bug is unknown, just that the broken behaviour has not been observed with this refactored driver. The measurable driver runtime overhead seem to be lower in the refactored driver. Also fixed a bug where setting a new target on a stopped timer would modify the paused timer_read value.

### Issues/PRs references

depends on #8928. Use #8531 for testing

### DIFF
--- a/cpu/kinetis/periph/timer.c
+++ b/cpu/kinetis/periph/timer.c
@@ -177,6 +177,8 @@ static inline void _pit_set_counter(uint8_t dev)
     PIT->CHANNEL[ch].TFLG = PIT_TFLG_TIF_MASK;
     /* Restore previous timer state */
     PIT->CHANNEL[ch].TCTRL = pit[dev].tctrl;
+    /* Set the next cycle to reload the maximum value */
+    PIT->CHANNEL[ch].LDVAL = PIT_MAX_VALUE;
 }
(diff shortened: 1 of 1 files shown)


gebart:
Rebased after #8928 was merged

tcschmidt:
@aabadie can you perform an initial review?

aabadie:
I don't feel enough confident with kinetis to review this, but maybe @kYc0o or @astralien3000 could be of better help here.

astralien3000:
I don't have my teensy right now. I will try to test as soon as possible.

gebart:
I recommend #8531 for testing timer implementations

astralien3000:
Sorry for the long waiting. I tested xtimer_msg, periph_timer, and bench_timers (from #8531) on a teensy31. Everything is working fine. I also tested bench_timers without this PR, which was showing errors. So, ACK !

PR #8936, opened 2018-04-12
net/rdcli_simple: specify RD address as string

haukepetersen:
### Contribution description
One of the findings of a CoRE RD plugtest this week was, that it is quite cumbersome in practice, to specify the RD server address in binary format as required by the current `rdcli_simple` implementation. With this PR the RD server address is now specified as string and parsed at run-time, making it much easier to specify the RD server address.

The downside is some run-timer overhead (1 call to `ipv6_addr_from_string()`), and a memory overhead of ~120 bytes ROM (`rdcli_simple` example @ `samr21-xpro`). But I would say this is worth it.

### Issues/PRs references
none

### DIFF
--- a/sys/include/net/rdcli_simple.h
+++ b/sys/include/net/rdcli_simple.h
@@ -30,8 +30,9 @@ extern "C" {
  * @brief   Error codes used by the rdcli_simple implementation
  */
 enum {
-    RDCLI_SIMPLE_OK  = 0,       /**< all ok */
-    RDCLI_SIMPLE_ERR = -1,      /**< return for formatting and network errors */
+    RDCLI_SIMPLE_OK     =  0,   /**< all good */
+    RDCLI_SIMPLE_NOADDR = -1,   /**< on address conversion errors */
+    RDCLI_SIMPLE_ERROR  = -2,   /**< on other errors */
 };
 
 /**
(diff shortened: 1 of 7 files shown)


sys/net/application_layer/rdcli_simple/rdcli_simple.c
```diff
@@ -25,36 +25,45 @@
...
 int rdcli_simple_register(void)
 {
+    sock_udp_ep_t remote = {
+        .family    = AF_INET6,
+        .netif     = SOCK_ADDR_ANY_NETIF,
+        .port      = RDCLI_SERVER_PORT
+    };
+
+    /* parse RD server address */
+    if (ipv6_addr_from_str((ipv6_addr_t *)&remote.addr,
```
bergzand: Shouldn't this be `&remote.addr.ipv6`? (I know that this works too.)

haukepetersen: True, that would be more easy to read. Will fix.

sys/include/net/rdcli_simple.h
```diff
@@ -26,9 +26,21 @@
 extern "C" {
 #endif
 
+/**
+ * @brief   Error codes used by the rdcli_simple implementation
+ */
+enum {
+    RDCLI_SIMPLE_OK  = 0,       /**< all ok */
+    RDCLI_SIMPLE_ERR = -1,      /**< return for formatting and network errors */
```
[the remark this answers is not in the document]
haukepetersen: Will do - as it basically comes for free :-)

haukepetersen:
addressed comments and squashed.

Also improved the error handling in the `rdcli_simple_standalone` code

haukepetersen:
thanks!

PR #8938, opened 2018-04-12
pkg/cn-cbor: Bump version, remove patch

bergzand:
Patches for unaligned memory access have been added upstream and can be removed in our package.


bergzand:
You're right, completely missed that. I've added the parameter to the cflags in `Makefile.cn-cbor`, this looks like the right place to me, but I'm not that good with packages.

kaspar030:
ACK, please squash!

bergzand:
squashed!

PR #8895, opened 2018-04-06
pkg: add support for libcose

bergzand:
### Contribution description

[libcose](https://github.com/bergzand/libcose) is a library implementing parts of the [COSE](https://tools.ietf.org/html/rfc8152) spec. As of v0.2 Signing and verification with ed25519 based keys is supported. This package relies on cn-cbor, a block allocator and a crypto library (either TweetNaCl or LibSodium at this moment). Part of the tests used by libcose is reworked as a unittest test to provide RIOT specific tests.

COSE is used by both SUIT and OSCORE for security. The reason to keep this library out of the RIOT tree is to ensure it's stand alone and generic nature so that other projects can benefit from it.

### Issues/PRs references

Depends on #7651, had to rebase that for cn-cbor fixes, so commit history is a bit messy at the moment.

PR #8947, opened 2018-04-13
pkg/cn-cbor: compile with context pointer support

bergzand:
### Contribution description

This enables `USE_CONTEXT` when compiling cn-cbor. With this a cn_cbor_context struct with application specific calloc/free calls is used for the allocation of `cn_cbor` structs. Now that there is a memarray module in RIOT, enabling this allows us to use the memarray for `cn_cbor` struct allocation instead of having to use POSIX calloc and free calls.

The test is adapted to use a struct with `NULL` values, this causes cn-cbor to fall back to calloc/free.

### Issues/PRs references

Also required for building the libcose unittest from #8895 together with the cn-cbor test

### DIFF
--- a/pkg/cn-cbor/Makefile.include
+++ b/pkg/cn-cbor/Makefile.include
@@ -1,1 +1,4 @@
 INCLUDES += -I$(PKGDIRBASE)/cn-cbor/include
+
+#enable context pointer
+CFLAGS += -DUSE_CBOR_CONTEXT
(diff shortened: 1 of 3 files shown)


cladmi:
@lorenz9314 maybe you can also take a look

kYc0o:
I like better using an actual memory allocator than the previous dummy example. Thanks for the efforts!

PR #7610, opened 2017-09-15
cpu/atmega: WIP implementation of pin change interrupts

roberthartung:
Fixes #7609: First implementation of pin change interrupts.
Feedback much appreciated @lebrush 

Known problems:
- Pin change interrupts do not work on the PCINT for the context swap!

PR #8951, opened 2018-04-16
cpu/atmega_common: external interrupt fix and refactor

ZetaR60:
On the ATmegas there are several issues currently with external interrupts in cpu/atmega_common/periph/gpio.c:
- GPIO_EXT_INT_NUMOF is incorrectly defined for higher interrupt vectors, preventing their use.
- Order of operations with EICRB write was incorrect for gpio_init_int, preventing the use of some interrupts.
- Condition checking in gpio_init_int for ATmega1284p is wrong, preventing the use of one interrupt (this one is my fault).
- There is this mess of conditionals with lots of CPU ifdefs.

This PR:
- Removes the messy conditionals, and removes all CPU ifdefs.
- Finds if a pin is an interrupt and the interrupt number by looping over an array of pins defined by CPU_ATMEGA_EXT_INTS. This is placed in periph_cpu.h of each cpu.
- Corrects the bugs listed above.

Tested and working using tests/periph_gpio on ATmega1284p / mega-xplained, at least for INT2 (could not test INT0-1, since they are on the serial port).

### DIFF
--- a/cpu/atmega_common/periph/gpio.c
+++ b/cpu/atmega_common/periph/gpio.c
@@ -30,6 +30,7 @@
 #include "cpu.h"
 #include "periph/gpio.h"
 #include "periph_conf.h"
+#include "periph_cpu.h"
 
 #define GPIO_BASE_PORT_A        (0x20)
 #define GPIO_OFFSET_PORT_H      (0xCB)
@@ -45,13 +46,13 @@
 #elif defined(INT3_vect)
 #define GPIO_EXT_INT_NUMOF      (4U)
 #elif defined(INT4_vect)
-#define GPIO_EXT_INT_NUMOF      (4U)
+#define GPIO_EXT_INT_NUMOF      (5U)
 #elif defined(INT5_vect)
-#define GPIO_EXT_INT_NUMOF      (4U)
+#define GPIO_EXT_INT_NUMOF      (6U)
 #elif defined(INT6_vect)
-#define GPIO_EXT_INT_NUMOF      (4U)
+#define GPIO_EXT_INT_NUMOF      (7U)
 #elif defined(INT7_vect)
-#define GPIO_EXT_INT_NUMOF      (4U)
+#define GPIO_EXT_INT_NUMOF      (8U)
 #else
 #define GPIO_EXT_INT_NUMOF      (2U)
 #endif
@@ -109,6 +110,21 @@ static inline uint16_t _pin_addr(gpio_t pin)
     return (_port_addr(pin) - 0x02);
 }
 
+static inline int8_t _int_num(gpio_t pin)
+{
+    uint8_t num;
+    const gpio_t ext_ints[] = CPU_ATMEGA_EXT_INTS;
+
+    /* find pin in ext_ints array to get the interrupt number */
+    for (num = 0; num < GPIO_EXT_INT_NUMOF; num++) {
+        if (pin == ext_ints[num]) {
+            return num;
+        }
+    }
+
+    return -1;
+}
+
 int gpio_init(gpio_t pin, gpio_mode_t mode)
 {
     switch (mode) {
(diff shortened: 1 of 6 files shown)


cpu/atmega_common/periph/gpio.c
```diff
@@ -154,35 +164,27 @@ int gpio_init_int(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
...
-    if (pin_num < 4) {
-        EICRA |= (flank << (pin_num * 2));
+    /* apply flank to interrupt number int_num */
+    if (int_num < 4) {
+        EICRA |= (flank << (int_num * 2));
     }
 #if defined(EICRB)
     else {
-        EICRB |= (flank << (pin_num * 2) % 4);
+        EICRB |= (flank << ((int_num % 4) * 2));
```
kYc0o: Can you tell more about this change? At a first glance it looks like the same, but I know there's a difference. I also wonder it worked in the past.

ZetaR60: Here is an example if int_num == 6:
- Old behavior: multiply by 2 (== 12), then modulo 4 (remainder of 12/4 = 0), ergo use ISC40 and ISC41 in EICRB (wrong).
- New behavior: modulo 4 (remainder of 6/4 = 2), then multiply by 2 (2*2 = 4), ergo use ISC60 and ISC61 in EICRB (right).

kYc0o: Ok, thanks for the explanation!

ZetaR60:
One thing that concerns me: if GPIO_EXT_INT_NUMOF is different the size of the CPU_ATMEGA_EXT_INTS array, then non-existent members of the array can be addressed. Normally this isn't a problem, since this is a low-level definition (and wouldn't be changed often). It is difficult to not use two definitions however, since prior to using CPU_ATMEGA_EXT_INTS to define an array it's size to sizeof is undefined. Doing something like sizeof(CPU_ATMEGA_EXT_INTS)/sizeof(1) might work, but seems like a bad idea (might be overly dependent on compiler behavior). Thoughts?

ZetaR60:
Can't believe I missed this super obvious solution to my concern posted above.

ZetaR60:
Hmm. It is failing compile with the change I just made. I am not able to properly debug this at the moment (I am out of town). If you wish to undo / fix / change / squash the code, feel free.

ZetaR60:
Solved the problem anyway... by fixing yet another existing bug in cpu/atmega_common/periph/gpio.c. Wow. The order of the ifdefs was reversed from what it should be.

roberthartung:
What about Pin Change Interrupts? See #7610

ZetaR60:
@roberthartung My understanding of the situation: There are multiple bugs that this PR fixes. Since RIOT is in feature freeze for the new release, this is probably going to get added (as a bug fix) for the release before #7610 gets added. So, #7610 should be tweaked to be compatible with #8951. Frustrating, I know. I had some stuff that was ready but didn't make it in as well.

I also have some other suggestions for #7610, which I will post over there.

ZetaR60:
ping @kYc0o

cpu/atmega_common/periph/gpio.c
```diff
@@ -192,38 +194,12 @@ int gpio_init_int(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
...
-    if (_port_num(pin) == PORT_D) {
-        EIMSK |= (1 << (_pin_num(pin) - 2));
-    }
-    else {
-        EIMSK |= (1 << _pin_num(pin));
-    }
-#else
-    EIMSK |= (1 << _pin_num(pin));
-#endif
+    EIMSK |= (1 << _int_num(pin));
```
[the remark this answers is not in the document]
ZetaR60: The pin validity is checked in gpio_init_int and afterwards presumed valid.

ZetaR60:
@kYc0o Is there anything else this PR needs?

kYc0o:
Looks ok for me, but need to test, afterwards I can give my tested ACK. However I still wonder if this PR conflicts with #7610, I'd like to have @roberthartung 's opinion on this.

ZetaR60:
ping @roberthartung

roberthartung:
@ZetaR60 @kYc0o Go ahead with this one. Afterwards @TorbenPetersen can work on #7610 once #8993 is merged.

kYc0o:
@roberthartung thanks! @ZetaR60 you might squash so the CI tests will pass.

ZetaR60:
Squashed and double checked with tests/periph_gpio. Everything is looking good and ready to go. Thanks for the help with this PR!

PR #8952, opened 2018-04-16
boards/common/atmega: gracefully handle CKDIV8 fuse

ZetaR60:
So far as I can tell, all ATmega chips ship with the CLKDIV8 fuse bit set. Some boards ship with the fuse unprogrammed, others do not (such as the mega-xplained board). If this fuse is programmed, the system clock starts with the wrong frequency. This PR gracefully handles initialization on ATmega boards that start with a divided clock. Incidentally, this also allows setting a custom division setting in board.h.

Tested on ATmega1284p / mega-xplained.

### DIFF
--- a/boards/common/atmega/board.c
+++ b/boards/common/atmega/board.c
@@ -23,10 +23,15 @@
 #include "irq.h"
 #include "periph/gpio.h"
 
+#ifndef CPU_ATMEGA_CLK_SCALE_INIT
+#define CPU_ATMEGA_CLK_SCALE_INIT    CPU_ATMEGA_CLK_SCALE_DIV1
+#endif
+
 void led_init(void);
 
 void board_init(void)
 {
+    atmega_set_prescaler(CPU_ATMEGA_CLK_SCALE_INIT);
     atmega_stdio_init();
     cpu_init();
     led_init();
--- a/cpu/atmega_common/include/cpu.h
+++ b/cpu/atmega_common/include/cpu.h
@@ -93,6 +93,37 @@ __attribute__((always_inline)) static inline void cpu_print_last_instruction(voi
 }
 
 /**
+ * @brief   ATmega system clock prescaler settings
+ *
+ * Some CPUs may not support the highest prescaler settings
+ */
+enum {
+    CPU_ATMEGA_CLK_SCALE_DIV1 = 0,
+    CPU_ATMEGA_CLK_SCALE_DIV2 = 1,
+    CPU_ATMEGA_CLK_SCALE_DIV4 = 2,
+    CPU_ATMEGA_CLK_SCALE_DIV8 = 3,
+    CPU_ATMEGA_CLK_SCALE_DIV16 = 4,
+    CPU_ATMEGA_CLK_SCALE_DIV32 = 5,
+    CPU_ATMEGA_CLK_SCALE_DIV64 = 6,
+    CPU_ATMEGA_CLK_SCALE_DIV128 = 7,
+    CPU_ATMEGA_CLK_SCALE_DIV256 = 8,
+    CPU_ATMEGA_CLK_SCALE_DIV512 = 9,
+};
+
+/**
+ * @brief   Initializes system clock prescaler
+ */
+static inline void atmega_set_prescaler(uint8_t clk_scale)
+{
+    /* Enable clock change */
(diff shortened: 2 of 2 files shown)


bergzand:
@ZetaR60 Do I understand correctly that this effectively overrides the CLKDIV8 fuse setting?

boards/common/atmega/board.c
```diff
@@ -23,10 +23,24 @@
...
+#ifndef CPU_ATMEGA_CLK_SCALE_INIT
+#define CPU_ATMEGA_CLK_SCALE_INIT    CPU_ATMEGA_CLK_SCALE_DIV1
+#endif
+
 void led_init(void);
 
+static inline void set_prescaler(void)
+{
+    /* Enable clock change */
+    CLKPR = 0x80;
```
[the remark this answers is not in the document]
ZetaR60: Fixed.

ZetaR60:
@bergzand Yes. When the CLKDIV8 fuse bit is set, it initializes the clock prescaler as set to divide the clock by 8. This just overwrites that change with the chosen clock prescaler setting.

boards/common/atmega/board.c
```diff
@@ -23,10 +23,25 @@
 #include "irq.h"
 #include "periph/gpio.h"
 
+#ifndef CPU_ATMEGA_CLK_SCALE_INIT
+#define CPU_ATMEGA_CLK_SCALE_INIT    CPU_ATMEGA_CLK_SCALE_DIV1
+#endif
+
 void led_init(void);
 
+static inline void set_prescaler(void)
```
bergzand: This function looks CPU-related to me. Would it be problematic to move it to `cpu/atmega_common/include/cpu.h` together with the `#define` you added above? Maybe rename it to `atmega_set_prescaler` then.

bergzand:
Tested it on an arduino mega2560 with the timer_periodic_wake example, works perfectly, booting is a lot slower though, but that's to be expected :)

boards/common/atmega/board.c
ZetaR60: Note to others: bergzand and I discussed this in IRC. There are some issues with moving it due to wanting to have the ability to override CPU_ATMEGA_CLK_SCALE_INIT in board.h, which would require adding an include for board.h to cpu.h if it is moved. We haven't decided yet the best way to handle this.

boards/common/atmega/board.c
```diff
@@ -23,10 +23,25 @@
 #include "irq.h"
 #include "periph/gpio.h"
 
+#ifndef CPU_ATMEGA_CLK_SCALE_INIT
+#define CPU_ATMEGA_CLK_SCALE_INIT    CPU_ATMEGA_CLK_SCALE_DIV1
```
bergzand: This should go into `./boards/common/arduino-atmega/include/periph_conf.h` preferably under the `CLOCK_CORECLOCK` definition and with a small bit of doxygen.

Can be done without the `#ifndef` construction but I don't mind keeping it.

boards/common/atmega/board.c
bergzand: This one into `cpu/atmega_common/include/cpu.h` as `static inline void atmega_set_prescaler(void)`.

boards/common/atmega/board.c
ZetaR60: Problem though: not all boards using boards/common/atmega are Arduinos.

boards/common/atmega/board.c
ZetaR60: Fixed, but also takes the clock scale as an argument, which means that cpu.h doesn't need another include.

boards/common/atmega/board.c
kYc0o: I think we can solve that problem in a follow up PR. Discussed a bit offline with @bergzand and maybe found a solution.

ZetaR60:
Did another quick test just to make sure that I made no mistakes when moving things around. Everything is working fine.

Shall we now push the remaining movement of code to a refactor PR, squash, and merge?

bergzand:
@ZetaR60 Squash please, I think this is good to go, but I'll let @kYc0o do the final ACK.

ZetaR60:
Squashed. Thanks for helping with this!

ZetaR60:
@bergzand I think you forgot to approve the changes from your review.

PR #8953, opened 2018-04-16
gnrc_lwmac: use ximer for recording radio duty over RTT

zhuoshuguo:
### Contribution description
This PR simply uses xtimer to record LWMAC's radio duty-cycle instead of using (currently) the RTT module, thus to enable more node of supporting radio duty-cycle recording/printing.


fjmolinas:
This PR could remove the clash with rtt...

PR #8962, opened 2018-04-16
boards*: cleanup doxygen

### LINKED ISSUE #6466: doc: @brief where @name is supposed to be
In many places in the doc this structure is used:

```C
/**
 * @brief name of group
 * @{
 */
#define FOOBAR (0)   /**< some doc that sometimes isn't here, but that's okay */
/** @} */
```

The intend is to add FOOBAR to a group named "name of group", but this is not what that results in. Rather is "name of group" parsed as the brief for FOOBAR, while the original `@brief` (if present) is parsed as its `@details`. The correct command would be `@name`.

Here is a list of top-level directories that should be addressed:
- [ ] boards (#8962)
- [ ] cpus
- [ ] drivers

aabadie:
- replace brief with name where required
- remove some unwanted empty lines
- fix missing closing doxygen name block

### Contribution description

I found #6466 was still opened and went through the board doxygen documentation. There remaining occurence of `@brief` where `@name` should be used.

This PR fixes the ones I found + a few other minor things.

### Issues/PRs references

Partially addresses #6466

### DIFF
--- a/boards/arduino-mkr1000/include/board.h
+++ b/boards/arduino-mkr1000/include/board.h
@@ -37,7 +37,7 @@ extern "C" {
 #define ARDUINO_LED         (6U)
 
 /**
- * @brief   LED pin definitions and handlers
+ * @name    LED pin definitions and handlers
  * @{
  */
 #define LED0_PIN            GPIO_PIN(PA, 20)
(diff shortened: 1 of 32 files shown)


aabadie:
All green, let's merge this one.

PR #8974, opened 2018-04-18
sys/binsearch: Add binary search module.

jcarrano:
### Description

There are places (for example, the shell and the Lua interpreter I'm working on) where one has to search an array of structs (a table), which is generally ordered.

This code allows to do this in a generic way. Also, it uses binary search so it should be faster than iteration for bigger arrays.

There are two functions and two macros. The macros are a convenience to avoid having to use `sizeof` and `offsetof` all the time.

### DIFF
--- a/sys/binsearch/Makefile
+++ b/sys/binsearch/Makefile
@@ -1,0 +1,1 @@
+include $(RIOTBASE)/Makefile.base
--- a/sys/binsearch/binsearch.c
+++ b/sys/binsearch/binsearch.c
@@ -1,3 +1,11 @@
+/*
+ * Copyright (C) 2017 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
 /**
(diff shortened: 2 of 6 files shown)


gebart:
Is there any benefit to using this over stdlib.h `bsearch()`?

gebart:
A more flexible approach is to use a comparison function pointer argument, which allows adapting it to comparing all kinds of data, not only strings.

jcarrano:
@gebart : It could be implemented in terms of bsearch, but in the end you'll end up with even more code. The idea of adding a comparison callback is quite easy to implement, but I don't know how useful it will be.

gebart:
In my opinion, a custom comparison I'd extremely useful. Even just reversing the sort order can be trivially implemented with a custom comparison function (`cmp(a, b) { return a > b;}` vs. `cmp(a, b) { return a < b;}`)

I don't like adding code which does the same thing as some standard library part unless there is any special circumstance that makes the standard library version unsuitable (e.g. the fmt library which uses a lot less memory than the libc printf family of functions).
The standard library components are usually pretty well tested, and we don't have to maintain the implementations.

PR #8978, opened 2018-04-18
drivers: add driver for FXOS8700 3-axis accelerometer/magnetometer

Hyungsin:
This PR provides a driver for the FXOS8700 sensor, which reads 3-axis acceleration and magnetic field.
Driver implementation style is similar to HDC1000 driver.

### DIFF
--- a/drivers/fxos8700/fxos8700.c
+++ b/drivers/fxos8700/fxos8700.c
@@ -1,0 +1,235 @@
+/*
+ * Copyright (C) 2018 UC Berkeley
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @file
+ * @brief       Driver for the FXOS8700 3-axis accelerometer/magnetometer
+ *
+ * @author      Michael Andersen <m.andersen@cs.berkeley.edu>
+ * @author      Hyung-Sin Kim <hs.kim@cs.berkeleyedu>
+ */
+
+
+#include "periph/i2c.h"
+#include "xtimer.h"
+#include "fxos8700.h"
+
+#define ENABLE_DEBUG (0)
+#include "debug.h"
+
+#define I2C_SPEED                I2C_SPEED_FAST
+
+#define FXOS8700_FULLDATA_LENGTH (12)
+
+static int fxos8700_read_regs(const fxos8700_t* dev, uint8_t reg, uint8_t* data, size_t len)
+{
+    i2c_acquire(dev->p.i2c);
+    if (i2c_read_regs(dev->p.i2c, dev->p.addr, reg, (char*) data, len) <= 0) {
+        DEBUG("[fxos8700] Can't read register 0x%x\n", reg);
+        i2c_release(dev->p.i2c);
+        return FXOS8700_BUSERR;
+    }
+    i2c_release(dev->p.i2c);
+
+    return FXOS8700_OK;
+}
+
+static int fxos8700_write_regs(const fxos8700_t* dev, uint8_t reg, uint8_t* data, size_t len)
+{
+    i2c_acquire(dev->p.i2c);
+    if (i2c_write_regs(dev->p.i2c, dev->p.addr, reg, (char*) data, len) <= 0) {
+        DEBUG("[fxos8700] Can't write to register 0x%x\n", reg);
+        i2c_release(dev->p.i2c);
+        return FXOS8700_BUSERR;
+    }
+    i2c_release(dev->p.i2c);
+
+    return FXOS8700_OK;
+}
+
+int fxos8700_init(fxos8700_t* dev, const fxos8700_params_t *params)
+{
+    uint8_t config;
+
+    if ((params->addr < 0x1C) || (params->addr > 0x1F)) {
+        DEBUG("[fxos8700] Invalid address\n");
+        return FXOS8700_ADDRERR;
+    }
+    dev->p.addr = params->addr;
+    dev->p.i2c = params->i2c;
+    dev->p.acc_range = params->acc_range;
+    dev->p.renew_interval = params->renew_interval;
+
+    i2c_acquire(dev->p.i2c);
+    if (i2c_init_master(dev->p.i2c, I2C_SPEED) != 0) {
+        DEBUG("[fxos8700] Can't initialize I2C master\n");
+        i2c_release(dev->p.i2c);
+        return FXOS8700_NOBUS;
+    }
+
+    if (i2c_read_regs(dev->p.i2c, dev->p.addr, FXOS8700_REG_WHO_AM_I, &config, 1) <= 0) {
+        i2c_release(dev->p.i2c);
+        DEBUG("[fxos8700] Could not read WHOAMI\n");
+        return FXOS8700_NOBUS;
+    }
+    if (config != FXOS8700_WHO_AM_I_VAL) {
+        i2c_release(dev->p.i2c);
+        DEBUG("[fxos8700] WHOAMI is wrong (%2x)\n", config);
+        return FXOS8700_NODEV;
+    }
+    i2c_release(dev->p.i2c);
+
+    /* Configure the ODR to maximum (400Hz in hybrid mode) */
+    config = FXOS8700_REG_CTRL_REG1_ODR__400HZ & FXOS8700_REG_CTRL_REG1_MASK__ODR;
+    if (fxos8700_write_regs(dev, FXOS8700_REG_CTRL_REG1, &config, 1) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
+    }
+    dev->config = config;
+    /* Activate hybrid mode */
+    config = FXOS8700_REG_M_CTRL_REG1_HMS__HYBRID & FXOS8700_REG_M_CTRL_REG1_MASK__HMS;
+    if (fxos8700_write_regs(dev, FXOS8700_REG_M_CTRL_REG1, &config, 1) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
+    }
+    /* Set burst read mode (accel + magnet together) */
+    config = FXOS8700_REG_M_CTRL_REG2_MASK__HYB_AUTOINC_MODE;
+    if (fxos8700_write_regs(dev, FXOS8700_REG_M_CTRL_REG2, &config, 1) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
+    }
+    /* Set accelerator's full-scale range */
+    if (fxos8700_read_regs(dev, FXOS8700_REG_XYZ_DATA_CFG, &config, 1) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
+    }
+    config &= ~FXOS8700_REG_XYZ_DATA_CFG_MASK__FS;
+    config |= dev->p.acc_range & FXOS8700_REG_XYZ_DATA_CFG_MASK__FS;
+    if (fxos8700_write_regs(dev, FXOS8700_REG_XYZ_DATA_CFG, &config, 1) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
+    }
+
+    /* initial read for caching operation */
+    if (fxos8700_read(dev, &dev->acc_cached, &dev->mag_cached) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
+    }
+    dev->last_read_time = xtimer_now_usec();
+
+    return FXOS8700_OK;
+}
+
+
+int fxos8700_set_active(const fxos8700_t* dev)
+{
+    uint8_t config = dev->config | FXOS8700_REG_CTRL_REG1_MASK__ACTIVE;
+    if (fxos8700_write_regs(dev, FXOS8700_REG_CTRL_REG1, &config, 1) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
+    }
+    return FXOS8700_OK;
+}
+
+int fxos8700_set_idle(const fxos8700_t* dev)
+{
+    uint8_t config = dev->config & ~FXOS8700_REG_CTRL_REG1_MASK__ACTIVE;
+    if (fxos8700_write_regs(dev, FXOS8700_REG_CTRL_REG1, &config, 1) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
+    }
+    return FXOS8700_OK;
+}
+
+int fxos8700_read(const fxos8700_t* dev, fxos8700_measurement_t* acc,
+                  fxos8700_measurement_t* mag)
+{
+    uint8_t data[12];
+    uint8_t ready = 0;
+
+    if (fxos8700_set_active(dev)) {
+        return FXOS8700_BUSERR;
+    }
+
+    while (!(ready & FXOS8700_REG_STATUS_MASK__XYZ_READY)) {
+        fxos8700_read_regs(dev, FXOS8700_REG_STATUS, &ready, 1);
+    }
+    while (!(ready & FXOS8700_REG_M_DR_STATUS_MASK__XYZ_READY)) {
+        fxos8700_read_regs(dev, FXOS8700_REG_M_DR_STATUS, &ready, 1);
+    }
+
+    /* Read all data at once */
+    if (fxos8700_read_regs(dev, FXOS8700_REG_OUT_X_MSB, &data[0], FXOS8700_FULLDATA_LENGTH)) {
+        return FXOS8700_BUSERR;
+    }
+
+    if (fxos8700_set_idle(dev)) {
+        return FXOS8700_BUSERR;
+    }
+
+    /* Read accelerometer */
+    if (acc) {
+#if FXOS8700_USE_ACC_RAW_VALUES
+        acc->x = (int16_t) ((data[0] << 8) | data[1]) >> 2;
+        acc->y = (int16_t) ((data[2] << 8) | data[3]) >> 2;
+        acc->z = (int16_t) ((data[4] << 8) | data[5]) >> 2;
+#else
+        int32_t acc_raw_x = (int16_t) ((data[0] << 8) | data[1]) >> 2;
+        int32_t acc_raw_y = (int16_t) ((data[2] << 8) | data[3]) >> 2;
+        int32_t acc_raw_z = (int16_t) ((data[4] << 8) | data[5]) >> 2;
+        switch(dev->p.acc_range) {
+            case FXOS8700_REG_XYZ_DATA_CFG_FS__2G:
+                acc->x = (int16_t) ((acc_raw_x * 244) / 100);
+                acc->y = (int16_t) ((acc_raw_y * 244) / 100);
+                acc->z = (int16_t) ((acc_raw_z * 244) / 100);
+                break;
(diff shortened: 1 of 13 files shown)


gebart:
Nice! This sensor can be found on the frdm-kw41z board.

drivers/fxos8700/include/fxos8700_params.h
```diff
@@ -0,0 +1,75 @@
...
+{
+    FXOS8700_PARAMS
+};
+
+/**
+ * @brief   Additional meta information to keep in the SAUL registry
+ */
+static const saul_reg_info_t fxos8700_saul_info[] =
+{
+    HDC1000_SAUL_INFO
```
gebart: Copy pasta

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,203 @@
...
+        return FXOS8700_NOBUS;
+    }
+
+    rv = i2c_read_regs(dev->p.i2c, dev->p.addr, 0x0D, &config, 1);
+    if (rv != 1) {
+        i2c_release(dev->p.i2c);
+        DEBUG("[fxos8700] Could not read WHOAMI (%d)\n", rv);
+        return FXOS8700_NOBUS;
+    }
+    dev->whoami = config;
```
gebart: Is there any use case for saving this value? It seems like it is never used by the implementation.

drivers/include/fxos8700.h
```diff
@@ -0,0 +1,293 @@
...
+/**
+ * @brief   Parameters needed for device initialization
+ */
+typedef struct {
+    i2c_t i2c;              /**< I2C device that sensor is connected to */
+    uint8_t addr;           /**< I2C address of this particular sensor */
+} fxos8700_params_t;
+
+/**
+  * @brief   Device descriptor for a AT30TSE75x device
```
gebart: Copy pasta, and weird indentation

drivers/include/fxos8700.h
```diff
@@ -0,0 +1,293 @@
...
+  * @{
+  */
+typedef struct {
+    fxos8700_params_t p;
+    uint8_t whoami;
+} fxos8700_t;
+
+/**
+  * @brief   Individual hybrid measurement
```
gebart: Weird indent

drivers/include/fxos8700.h
```diff
@@ -0,0 +1,293 @@
...
+        uint8_t  zyxow:1;     /*!< bit:      0  Start Conversion Event In          */
+        uint8_t  zow:1;       /*!< bit:      1  Synchronization Event In           */
+        uint8_t  yow:1;       /*!< bit:  2.. 3  Reserved                           */
+        uint8_t  xow:1;       /*!< bit:      0  Start Conversion Event In          */
+        uint8_t  zyxdr:1;     /*!< bit:      1  Synchronization Event In           */
+        uint8_t  zdr:1;       /*!< bit:  2.. 3  Reserved                           */
+        uint8_t  ydr:1;       /*!< bit:      0  Start Conversion Event In          */
+        uint8_t  xdr:1;       /*!< bit:      1  Synchronization Event In           */
+    } bit;                    /*!< Structure used for bit  access                  */
+    uint8_t reg;              /*!< Type      used for register access              */
```
gebart: Here and everywhere else: use `/**` (and of course `/**<`) for doxygen comments in RIOT.

drivers/include/fxos8700.h
```diff
@@ -0,0 +1,293 @@
...
+} FXOS8700_STATUS_Type;
```
gebart: This type, and the other bit fields below, are not actually used by the implementation.
Additionally, it is not required by a compiler to pack the bit fields in the struct, so this bit representation will not be guaranteed to match the hardware register bits. It is more robust to use bit mask macros or other bit operations in the implementation. The compiler should be smart enough to optimize the accesses in a good way when one of the operands is a numeric literal or static constant.

gebart:
The cache does not handle multiple instances.
I understand that your projects use the caching to reduce the power consumption while reading multiple successive reads, but I wonder if we should generalize the cache into its own module and let it wrap any sensor driver, with configurable cache time to live. Most sensors have the same kind of read function API where you pass a driver instance as the first argument and an output buffer as the next argument with error codes as the return value, that would be easily wrapped by a generic cache implementation.

gebart:
Otherwise we will likely end up with various cache implementations in each sensor driver, with plenty of opportunity for mistakes and difficulties testing them in a reliable way.

gebart:
I think it is a good verification to read the whoami register during init, to check that the device is responding and that the connected device is actually a FXOS8700 sensor by comparing the actual value against the expected value. Just that it does not need to be saved in the context struct.

Hyungsin:
@gebart. Thank you for the quick review! I resolved you comments.

WHOAMI comparison is still here but its value is not saved now.

Regarding the cache, I don't know how many sensors in RIOT are relevant (one sensor should provide multiple types of information). If this cache functionality is useful for many other sensors, we can think of generalizing it. But I think this should be another PR.

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /* Read accelerometer */
+    if (acc) {
+        int32_t acc_raw_x = (int32_t) ((data[0]<<6) | (data[1]>>2));
+        int32_t acc_raw_y = (int32_t) ((data[2]<<6) | (data[3]>>2));
+        int32_t acc_raw_z = (int32_t) ((data[4]<<6) | (data[5]>>2));
+        acc->x = (int16_t)(acc_raw_x*244/1000);
+        acc->y = (int16_t)(acc_raw_y*244/1000);
+        acc->z = (int16_t)(acc_raw_z*244/1000);
```
gebart: Code style: add spaces around operators

gebart: Also, for clarity it may be better to have parentheses around `(raw * 244) / 1000`, though it changes nothing in the generated code since the associativity is left to right for the multiplication and division operators.

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,205 @@
...
+    }
+    /* Read magnetometer */
+    if (mag) {
+        mag->x = (int16_t) ((data[6] <<8) | data[7]);
+        mag->y = (int16_t) ((data[8] <<8) | data[9]);
+        mag->z = (int16_t) ((data[10]<<8) | data[11]);
```
gebart: Spaces around bit shift operators

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,205 @@
...
+    }
+    return FXOS8700_OK;
+}
+
+int fxos8700_read_cached(const fxos8700_t *dev, fxos8700_measurement_t* acc, fxos8700_measurement_t* mag)
+{
+    uint32_t now = xtimer_now_usec();
```
gebart: Optional: It may be less computationally expensive to use xtimer_now and convert the renew interval to ticks instead, since you can precompute that us to tick conversion at compilation time rather than do a conversion from ticks to us in every read. See this as an optional comment, and maybe just remember for future drivers, I assume this function will be called somewhat infrequently since the renew interval is set to 1 second.

gebart:
Are you using this sensor externally or integrated on any dev board? Do you have any updated board configurations you want to add here?

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,205 @@
...
+
+static fxos8700_measurement_t acc_cached, mag_cached;
+static uint32_t last_read_time;
+
+static int fxos8700_read_regs(fxos8700_t* dev, uint8_t reg, uint8_t* data, size_t len)
+{
+    i2c_acquire(dev->p.i2c);
+    if(i2c_read_regs(dev->p.i2c, dev->p.addr, reg, (char*) data, len) <= 0) {
+        DEBUG("[fxos8700] Can't read register 0x%x\n", reg);
+        i2c_release(dev->p.i2c);
```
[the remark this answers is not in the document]
Hyungsin: I followed the style of the current HDC1000 implementation. Is this style a strict requirement?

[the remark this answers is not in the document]
aabadie: No, that was more a suggestion for future maintainance, technically it doesn't change anything.

Hyungsin:
@gebart @aabadie 
Thank you for the reviews and I resolved comments again. 👍 

@gebart, I'm using FXOS8700 on the hamilton board, which I plan to include in RIOT.
Basically, all sensor/clock/radio drivers I have modified/added so far are related to the hamilton.

drivers/fxos8700/fxos8700.c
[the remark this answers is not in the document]
Hyungsin: I see. Thanks!

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,205 @@
...
+
+#ifndef FXOS8700_RENEW_INTERVAL
+#define FXOS8700_RENEW_INTERVAL 1000000ul
+#endif
+
+#define ACCEL_ONLY_MODE         (0x00)
+#define MAG_ONLY_MODE           (0x01)
+#define HYBRID_MODE             (0x03)
+
+static fxos8700_measurement_t acc_cached, mag_cached;
```
aabadie: Instead of those static variable; why no having them as attributes of `fxos8700_t` ?
Because otherwise this will be a problem with multiple instances of this device as already reported by @gebart [here](https://github.com/RIOT-OS/RIOT/pull/8978#issuecomment-382612342).That could be a best temporary solution until we have a module for managing cache.
Maybe saul could do it ?

Hyungsin: I agree and updated this PR reflecting this comment. Thanks!

drivers/include/fxos8700.h
```diff
@@ -0,0 +1,277 @@
...
+    i2c_t i2c;                /**< I2C device that sensor is connected to */
+    uint8_t addr;             /**< I2C address of this particular sensor */
+    uint32_t renew_interval;  /**< Interval for cache renewal */
+} fxos8700_params_t;
+
+/**
+ * @brief   Device descriptor for a FXOS8700 device
+ */
+typedef struct {
+    fxos8700_measurement_t acc_cached;
```
aabadie: The attributes of this struct are missing documentation, same with the fxos8700_measurement_t struct

aabadie:
Could you also add a test application ?

Hyungsin:
@aabadie, sure. I added an application. Thanks!

Hyungsin:
Now all checks have passed!

tests/driver_fxos8700/Makefile
```diff
@@ -0,0 +1,7 @@
+include ../Makefile.tests_common
+
+BOARD_BLACKLIST := msb-430 msb-430h telosb wsn430-v1_3b wsn430-v1_4 z1
```
aabadie: is this needed ?

tests/driver_fxos8700/README.md
```diff
@@ -0,0 +1,12 @@
...
+This is a manual test application for the FXOS8700 driver.
+
+# Usage
+This test application will initialize the FXOS8700 sensor with the configuration
+as specified in the default `fxos8700_params.h` file.
+
+If you want to use this test application with different parameters, you can
+simply override the default FXOS8700_PARAMS.
+
+After initialization, the sensor reads the pulse counter values every second and
```
aabadie: copy-paste (remaining occurrence of pulse counter)

tests/driver_fxos8700/main.c
```diff
@@ -0,0 +1,56 @@
...
+    else {
+        printf("[OK]\n");
+    }
+
+    while (1) {
+        /* Acceleration and Magnetic field reading */
+        fxos8700_measurement_t acc, mag;
+        if (fxos8700_read(&dev, &acc, &mag) != FXOS8700_OK) {
+            printf("Read failure\n");
+        } else {
```
aabadie: the else should on next line

tests/driver_fxos8700/main.c
```diff
@@ -0,0 +1,56 @@
...
+#include "xtimer.h"
+#include "fxos8700_params.h"
+
+#define SLEEP       US_PER_SEC
+
+int main(void)
+{
+    fxos8700_t dev;
+
+    printf("FXOS8700 driver test application\n");
```
aabadie: maybe use puts instead of printf

tests/driver_fxos8700/main.c
```diff
@@ -0,0 +1,56 @@
...
+
+    /* Initialization */
+    if (fxos8700_init(&dev, &fxos8700_params[0])) {
+        printf("[Failed]");
```
aabadie: 'puts' here and in other places below

tests/driver_fxos8700/main.c
```diff
@@ -0,0 +1,56 @@
...
+    while (1) {
+        /* Acceleration and Magnetic field reading */
+        fxos8700_measurement_t acc, mag;
+        if (fxos8700_read(&dev, &acc, &mag) != FXOS8700_OK) {
+            printf("Read failure\n");
+        } else {
+            printf("acceleration: (%d, %d, %d)\n", acc.x, acc.y, acc.z);
+            printf("magnetic field: (%d, %d, %d)\n", mag.x, mag.y, mag.z);
+        }
+        xtimer_usleep(SLEEP);
```
aabadie: since you are waiting 1s, you can use xtimer_sleep with 1 as parameter. Then you can also drop the define SLEEP above

drivers/include/fxos8700.h
```diff
@@ -0,0 +1,277 @@
...
+
+#include <stdint.h>
+#include "periph/i2c.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+/**
```
aabadie: Maybe put those defines in a separate internal header file (see other drivers)

tests/driver_fxos8700/main.c
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (C) 2017 UC Berkeley
```
aabadie: 2018

gebart:
I think something is wrong with the scale configuration. This is the SAUL output I see when I have the board completely still on my desk:

```
2018-04-20 17:23:36,600 - INFO # Reading from #14 (fxos8700|SENSE_ACCEL)
2018-04-20 17:23:36,602 - INFO # Data:	[0]   3742 mg
2018-04-20 17:23:36,603 - INFO # 	[1]   3957 mg
2018-04-20 17:23:36,605 - INFO # 	[2]    962 mg
2018-04-20 17:23:36,605 - INFO # 
2018-04-20 17:23:36,607 - INFO # Reading from #15 (fxos8700|SENSE_MAG)
2018-04-20 17:23:36,609 - INFO # Data:	[0]    763 mGs
2018-04-20 17:23:36,611 - INFO # 	[1]    271 mGs
2018-04-20 17:23:36,612 - INFO # 	[2]    683 mGs
2018-04-20 17:23:36,613 - INFO # 
```

The length (2-norm) of the acceleration vector should be close to 1000 mg, but we have some 5500 mg

gebart:
(or my sensor is broken)

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,204 @@
...
+        return FXOS8700_BUSERR;
+    }
+    i2c_release(dev->p.i2c);
+
+    return FXOS8700_OK;
+}
+
+static int fxos8700_write_regs(const fxos8700_t* dev, uint8_t reg, uint8_t* data, size_t len)
+{
+    i2c_acquire(dev->p.i2c);
```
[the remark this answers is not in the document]
Hyungsin: i2c_release is added now :)

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,204 @@
...
+    if (config != FXOS8700_WHO_AM_I_VAL) {
+        i2c_release(dev->p.i2c);
+        DEBUG("[fxos8700] WHOAMI is wrong (%2x)\n", config);
+        return FXOS8700_NODEV;
+    }
+
+    /* Configure the ODR to maximum (400Hz in hybrid mode) */
+    config = 0x00;
+    if (fxos8700_write_regs(dev, FXOS8700_REG_CTRL_REG1, &config, 1) != FXOS8700_OK) {
+        return FXOS8700_BUSERR;
```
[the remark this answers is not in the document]
Hyungsin: fxos8700_write_regs and fxos8700_read_regs have their own i2c_release.

Hyungsin:
@aabadie, @gebart, I addressed your comments.
and @gebart, let me test this with my device also and I will let you know. It worked before but there could be some bugs while rearranging codes.

Hyungsin:
@gebart, I think it works at least in my case (hamilton).
-------------------------------------------
Dev: fxos8700	Type: SENSE_ACCEL
Data:	[0] 15mg
        	[1] 1005mg
	        [2] 156mg
-------------------------------------------
Can you retry?

gebart:
I will check again tomorrow

gebart:
My frdm-kw41z is giving strange values, I will try again with a frdm-k22f board tomorrow

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,205 @@
...
+
+    while (!(ready & 0x08)) {
+        fxos8700_read_regs(dev, FXOS8700_REG__STATUS, &ready, 1);
+    }
+    while (!(ready & 0x08)) {
+        fxos8700_read_regs(dev, FXOS8700_REG_M_DR_STATUS, &ready, 1);
+    }
+
+    /* Read all data at once */
+    if (fxos8700_read_regs(dev, FXOS8700_REG_OUT_X_MSB, &data[0], 12)) {
```
gebart: change magic `12` into sizeof something or a macro definition.

drivers/fxos8700/fxos8700.c
```diff
@@ -0,0 +1,205 @@
...
+
+    if (fxos8700_set_idle(dev)) {
+        return FXOS8700_BUSERR;
+    }
+
+    /* Read accelerometer */
+    if (acc) {
+        int32_t acc_raw_x = (int32_t) ((data[0] << 6) | (data[1] >> 2));
+        int32_t acc_raw_y = (int32_t) ((data[2] << 6) | (data[3] >> 2));
+        int32_t acc_raw_z = (int32_t) ((data[4] << 6) | (data[5] >> 2));
```
gebart: the shift of 6 loses the sign bit

Hyungsin: Very good point. Thanks!!

Hyungsin:
@gebart, I updated, but I think this driver does not lose any bit regardless of resolution mode since it provides "FXOS8700_USE_ACC_RAW_VALUES" mode.

gebart:
:+1:

gebart:
@Hyungsin in future PRs, please create fixup/squash commits for changes you add after the initial review to make the review process easier. You can amend typos or other minor corrections, but for anything a little bigger, such as when you added the different scaling factor configurations in this PR, use a separate commit. Then squash the commits when the reviewer requests it. It makes it easier for the reviewer to know what has changed between iterations.

Hyungsin:
@gebart, good point. I should keep this in mind!

aabadie:
All green here. Let's merge.

PR #8982, opened 2018-04-19
pkg/semtech_loramac: handle non OK mcps confirm messages

aabadie:
### Contribution description

This PR is built on top of #8798 and #8639 and provides a fallback when no OK MCPS confirm message is received from the network after a TX.

It doesn't solve the fact that no message can be received when using ABP activation but at least it prevents the mac thread from being stuck forever.

For simplicity, non OK status are all treated as error (so we don't distinguish all the cases). See [here](https://github.com/Lora-net/LoRaMac-node/blob/v4.4.0/src/mac/LoRaMac.h#L575) for more information.

### Issues/PRs references

Based on ~#8798~ and ~#8639~

### DIFF
--- a/pkg/semtech-loramac/contrib/semtech_loramac.c
+++ b/pkg/semtech-loramac/contrib/semtech_loramac.c
@@ -171,6 +171,11 @@ static void mcps_confirm(McpsConfirm_t *confirm)
                 break;
         }
     }
+    else {
+        msg_t msg;
+        msg.type = MSG_TYPE_LORAMAC_TX_CNF_FAILED;
+        msg_send(&msg, semtech_loramac_pid);
+    }
 }
 
 /* MCPS-Indication event function */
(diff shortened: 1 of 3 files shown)


dylad:
Ran some tests on my side, OTAA still works as expected and it also fix some issues I encounter before like shell stucks after sending a message if public : off or class : C. Now I cannot block the shell (at least in OTAA, no clue about ABP).
I'll take a deeper look at the changes. If I can I'll also try ABP activation.

dylad:
Seems like I have the same issue as @jia200x with ABP : 
```
INFO #  loramac tx "BCDA"
INFO # Failed: mac is busy
```

aabadie:
@dylad can you try again ? I rebased the  branch on the latest master which should contain the fix for "mac is busy".

dylad:
Indeed now it works with ABP too ! @jia200x does it solve your issue too ?

jia200x:
This solves the blocked shell issue for me. Good!

aabadie:
@jia200x, I rebased this one, now that #8639 has been merged

jia200x:
Perfect! Let's wait Murdock

jia200x:
@dylad are you OK with this PR?

dylad:
@jia200x I'm fine with this PR.

jia200x:
Great! Then ACK&GO

PR #8986, opened 2018-04-19
sys/event: allow calling event_post multiple times

haukepetersen:
### Contribution description
Currently, `event_post()` triggers a failed assertion when called for an event, that is already queued (somewhere). In some use-cases, it is however useful, to be able to call `event_post()` more than once for the same event, for example when using events similar to `thread_flags`.

This PR modifies the `event_post()` function slightly, so that it can be call multiple times for the same event, simply leaving the event untouched in case it is already queued.

### Issues/PRs references
none

### DIFF
--- a/sys/event/event.c
+++ b/sys/event/event.c
@@ -23,11 +23,12 @@ void event_queue_init(event_queue_t *queue)
 
 void event_post(event_queue_t *queue, event_t *event)
 {
-    assert(!event->list_node.next);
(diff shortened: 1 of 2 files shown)


cladmi:
Mistake, not counting as a API change as it is not removing any functionality.

haukepetersen:
@kaspar030 ping

haukepetersen:
thanks!

PR #7523, opened 2017-08-29
core: introduce crossfile arrays (xfa)

kaspar030:
At many places, we're hitting a problem with C: arrays have to be defined within one compilation unit. The language itself has no way of doing otherwise, unless the linker helps a little.

This PR adds some macros and the added linker script entries in order to allow adding members to a static array from different compilation units. See the included tests/xfa for an example.

The main idea is that from anywhere, code can do sth like this:

~~```XFA(mydriver_params_t, 0, mydriver_params) onboard = { .spi=blah, .foo=bar}```~~
```XFA(mydriver_params, 0) mydriver_params_t onboard = { .spi=blah, .foo=bar}```

This would add ```onboard``` to the linker section .xfa.mydrivers_params.
Entries in those sections are sorted by priority (```0``` in this case) and name (```onboard``` in this case).

Something like that would be very convenient for:

- auto-init functions (declare them with the code, not in sys/auto_init/...)
- shell commands (same, no more sys/shell/commands.c)
- driver parameters (board adds onboard params to params array, user can simply add more)
- saul drivers

We've always tried to avoid "linker script black magic", but after years of dodging (and coming up with schemes like in #7519) and having unified the ldscripts per platform, I think we can afford adding those two lines.

I see this PR as a proof-of-concept and I'm happy to hear your opinion!

Currently only native, atmega2560 and cortexm have the necessary linker script support.
Waiting for #5757 (needs ugly hacks without, look for hack[12] in the example).
Solves #4290, #7519 and some others.

Todo:
- [ ] merge #5757
- [ ] add linkerscripts for msp430*
- [ ] add linkerscripts for remaining AVRs
- [ ] add linkerscripts for MIPS

PR #8987, opened 2018-04-19
 make: process include and dep for external modules

cladmi:
Allow using `EXTERNAL_MODULE_DIRS` for complete modules with configuration and dependencies as if they were in RIOT.

It gives them the same possibilities as packages when all files are already there and do not need to be downloaded.

### Contribution description

Process `Makefile.include` for external modules. It is included after the others
so it could overwrite some of the configuration if wanted.

Process `Makefile.dep` for external modules. It is included before the others so
it could be parsed before setting 'default' values to dependencies.

#### Compile time testing

I added a test that checks at compile time the header inclusion, linking and parsing dependencies.

#### Pre-contrib

~~I needed to make `USEMODULE_INCLUDES_ ` an immediate variable to allow using `MAKEFILE_LIST` to define an include directory. Not really sure why though :/ I know why the fix makes it work, not why it is required in the first place.~~
In the external module makefile include, I need to use an immediate variable to evaluate `MAKEFILE_LIST` directly.

### Issues/PRs references

Would allow using modules without modifying RIOT with a mechanism like https://github.com/RIOT-OS/RIOT/pull/7523

### DIFF
--- a/doc/doxygen/src/creating-modules.md
+++ b/doc/doxygen/src/creating-modules.md
@@ -74,6 +74,9 @@ The external module can optionally define the following files:
   API headers include paths to the `USEMODULE_INCLUDES` variable.
 * `Makefile.dep` file to set module dependencies
 
+An example can be found in
+[`tests/external_module_dirs`](https://github.com/RIOT-OS/RIOT/tree/master/tests/external_module_dirs)
+
 Pseudomodules                                                  {#pseudomodules}
 =============
 Pseudomodules are modules that do not have any code. Their main use cases are
--- a/doc/doxygen/src/creating-modules.md
+++ b/doc/doxygen/src/creating-modules.md
@@ -67,8 +67,12 @@ their dependencies.
 Modules outside of RIOTBASE                      {#modules-outside-of-riotbase}
 ===========================
 Modules can be defined outside `RIOTBASE`. In addition to add it to `USEMODULE`
-the user needs to add the path to the module to `EXTERNAL_MODULE_DIRS` and add
-the include path to the API definitions to `INCLUDES`.
+the user needs to add the module path to `EXTERNAL_MODULE_DIRS`.
+
+The external module can optionally define the following files:
+* `Makefile.include` file to set global build configuration like `CFLAGS` or add
+  API headers include paths to the `USEMODULE_INCLUDES` variable.
+* `Makefile.dep` file to set module dependencies
 
 Pseudomodules                                                  {#pseudomodules}
 =============
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -5,6 +5,10 @@ OLD_USEPKG := $(sort $(USEPKG))
 # include board dependencies
 -include $(RIOTBOARD)/$(BOARD)/Makefile.dep
 
+# include external modules dependencies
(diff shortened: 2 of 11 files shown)


doc/doxygen/src/creating-modules.md
```diff
@@ -67,8 +67,13 @@ their dependencies.
 Modules outside of RIOTBASE                      {#modules-outside-of-riotbase}
 ===========================
 Modules can be defined outside `RIOTBASE`. In addition to add it to `USEMODULE`
-the user needs to add the path to the module to `EXTERNAL_MODULE_DIRS` and add
-the include path to the API definitions to `INCLUDES`.
+the user needs to add the path to the module to `EXTERNAL_MODULE_DIRS`.
+The external module can define a `Makefile.include` file to add the include path
```
kaspar030: This is misleading."Makefile.include" will always be included by every submake, thus allows global definitions. That just happens to also be useful for changing the include path.

kaspar030:
Nice one, and one step towards ```USEMODULE_GIT += https://github.com/me/module```.

doc/doxygen/src/creating-modules.md
cladmi: It is not included by every submake, just by the main `Makefile.include`.

But yeah it could be used for more than just setting `USEMODULE_INCLUDES`, I could add a note on setting CFLAGS for example.

kaspar030: > It is not included by every submake

Right. But any variable that is exported will be available in every submake.

cladmi: I will reformulate in this direction.

cladmi:
Rebased the conflicting file.

smlng:
ping @kaspar030

cladmi:
Rebased on master. As it was merged after writing this PR, I found out that currently this PR does not talk about defining `BUILDDEP` in the external modules and does not use it in tests. This would show more how to define an external "package" like module. I can add this in followup PR.

cladmi:
Ping @kaspar030

kYc0o:
@kaspar030 are you satisfied with the rewording?

kYc0o:
@kaspar030 ping.

jia200x:
> @jia200x would you like to have this for the release? If so, we need to dismiss @kaspar030 review.

Would be nice to have this for the Release. @kaspar030 last call?

jia200x:
> @jia200x would you like to have this for the release? If so, we need to dismiss @kaspar030 review.

Would be nice to have this for the Release. @kaspar030 last call?

kaspar030:
> are you satisfied with the rewording?

Yes.

kaspar030:
> are you satisfied with the rewording?

Yes.

kaspar030:
> are you satisfied with the rewording?

Yes.

cladmi:
Updated the last comment.

kaspar030:
lgtm

kaspar030:
@cladmi pls squash!

jia200x:
murdock is happy. Let's merge it then ;)

jia200x:
@cladmi thank you for this contribution

cladmi:
Thanks all for reviewing and merging.

Issue #8979, opened 2018-04-19
Received packets processing and thread change Timing


Josar:
I ping from an RPI to the jiminy and see that the timings are somehow getting bigger and then jumping back to smaller and again getting bigger.

Thus i added pin toggels to xtimer and to the transciever interrupts. What i can see is that the requested ping is only processed after a xtimer overflow. I would assume that the packet is proccessed as soon as possible.

The timings from received and transmitted can be seen at the right side of the second figure.

So my question would be why does the thread not change imidiatly after the packet is recieved?

Any Ideas? 

![whatsapp image 2018-04-19 at 09 58 06](https://user-images.githubusercontent.com/6164061/38980045-153c268a-43bc-11e8-85a6-4116788fdd99.jpeg)
![whatsapp image 2018-04-19 at 09 56 50](https://user-images.githubusercontent.com/6164061/38980049-16af3f66-43bc-11e8-948d-895ba6a4feb6.jpeg)

gebart:
Does your problem still occur if you apply #8951?

Josar:
Yes atmega based. Thanks for the hint, i will look into it.

Josar:
@gebart could not find `cortexm_isr_end `, but you made me thinking in the right direction.
Compared the timer isr with my isr implementation and found missing information in sched.h

The interrupt has to call the scheduler to execute the thread change.
```
    if (sched_context_switch_request) {
        thread_yield();
    }
```

Now the ping time is at 20ms.
Thanks a lot.

gebart:
No problem :+1:

Josar:
As i see it in `msp430_common` this is included in `__exit_isr` as i would have expected it to be, in `atmega_common` it is not. Instead it is redundatnly written in every isr

gebart:
IMO the msp430 and cortex approach is preferable. This is something that was refactored for the Cortex-M platforms in #6171. Feel free to provide a similar PR for the atmega if you feel like you have the time to do it.

PR #6171, opened 2016-11-30
cpu/cortexm: added and use central yield_from_isr() function

haukepetersen:
a) adds a little convenience
b) one step into using the periph drivers without RIOT

PR #8988, opened 2018-04-19
cpu/atmega_common: __exit_isr thread_yield

Josar:
As proposed by @gebart in #8979 this PR adds `thread_yield();` to `__exit_isr` similar to #6171.

This avoids duplicated code. 
Additionaly code was uncrustified.

### DIFF
--- a/cpu/atmega_common/include/cpu.h
+++ b/cpu/atmega_common/include/cpu.h
@@ -35,7 +35,8 @@
 
 #include <avr/interrupt.h>
 #include "cpu_conf.h"
-
+#include "sched.h"
+#include "thread.h"
 /**
  * For downwards compatibility with old RIOT code.
  * TODO: remove once core was adjusted
@@ -52,7 +53,7 @@ extern "C" {
 extern volatile uint8_t __in_isr;
 
 /**
- * @brief Flag entering of an ISR
+ * @brief Run this code on entering interrupt routines
  */
 static inline void __enter_isr(void)
 {
@@ -60,10 +61,22 @@ static inline void __enter_isr(void)
 }
 
 /**
- * @brief Flag exiting of an ISR
+ * @brief   Exit ISR mode and yield with a return from interrupt. Use at the
+ * end of ISRs in place of thread_yield_higher. If thread_yield is needed, use
+ * thread_yield followed by thread_yield_isr instead of thread_yield alone.
+ */
+void thread_yield_isr(void);
+
+/**
+ * @brief Run this code on exiting interrupt routines
  */
 static inline void __exit_isr(void)
 {
+    if (sched_context_switch_request) {
+        thread_yield();
+        __in_isr = 0;
+        thread_yield_isr();
+    }
     __in_isr = 0;
 }
 
@@ -127,13 +140,6 @@ static inline void atmega_set_prescaler(uint8_t clk_scale)
  * @brief   Initializes avrlibc stdio
  */
 void atmega_stdio_init(void);
-
-/**
- * @brief   Exit ISR mode and yield with a return from interrupt. Use at the
- * end of ISRs in place of thread_yield_higher. If thread_yield is needed, use
- * thread_yield followed by thread_yield_isr instead of thread_yield alone.
- */
-void thread_yield_isr(void);
 
 #ifdef __cplusplus
 }
--- a/cpu/atmega_common/include/cpu.h
+++ b/cpu/atmega_common/include/cpu.h
@@ -88,20 +88,20 @@ void cpu_init(void);
 /**
  * @brief   Print the last instruction's address
  */
-__attribute__((always_inline)) static inline void cpu_print_last_instruction(void)
+ static inline void __attribute__((always_inline)) cpu_print_last_instruction(void)
 {
     uint8_t hi;
     uint8_t lo;
     uint16_t ptr;
 
-    __asm__ volatile( "in __tmp_reg__, __SP_H__  \n\t"
+    __asm__ volatile ("in __tmp_reg__, __SP_H__  \n\t"
                       "mov %0, __tmp_reg__       \n\t"
-                      : "=g"(hi) );
+                      : "=g" (hi));
 
-    __asm__ volatile( "in __tmp_reg__, __SP_L__  \n\t"
+    __asm__ volatile ("in __tmp_reg__, __SP_L__  \n\t"
                       "mov %0, __tmp_reg__       \n\t"
-                      : "=g"(lo) );
-    ptr = hi<<8 | lo;
+                      : "=g" (lo));
+    ptr = hi << 8 | lo;
     printf("Stack Pointer: 0x%04x\n", ptr);
 }
 
--- a/cpu/atmega_common/include/cpu.h
+++ b/cpu/atmega_common/include/cpu.h
@@ -44,7 +44,8 @@
 #include "irq.h"
 
 #ifdef __cplusplus
-extern "C" {
+extern "C"
+{
 #endif
 
 /**
@@ -88,7 +89,7 @@ void cpu_init(void);
 /**
  * @brief   Print the last instruction's address
  */
- static inline void __attribute__((always_inline)) cpu_print_last_instruction(void)
+static inline void __attribute__((always_inline)) cpu_print_last_instruction(void)
 {
     uint8_t hi;
     uint8_t lo;
@@ -111,13 +112,13 @@ void cpu_init(void);
  * Some CPUs may not support the highest prescaler settings
  */
 enum {
-    CPU_ATMEGA_CLK_SCALE_DIV1 = 0,
-    CPU_ATMEGA_CLK_SCALE_DIV2 = 1,
-    CPU_ATMEGA_CLK_SCALE_DIV4 = 2,
-    CPU_ATMEGA_CLK_SCALE_DIV8 = 3,
-    CPU_ATMEGA_CLK_SCALE_DIV16 = 4,
-    CPU_ATMEGA_CLK_SCALE_DIV32 = 5,
-    CPU_ATMEGA_CLK_SCALE_DIV64 = 6,
+    CPU_ATMEGA_CLK_SCALE_DIV1   = 0,
+    CPU_ATMEGA_CLK_SCALE_DIV2   = 1,
+    CPU_ATMEGA_CLK_SCALE_DIV4   = 2,
+    CPU_ATMEGA_CLK_SCALE_DIV8   = 3,
+    CPU_ATMEGA_CLK_SCALE_DIV16  = 4,
+    CPU_ATMEGA_CLK_SCALE_DIV32  = 5,
+    CPU_ATMEGA_CLK_SCALE_DIV64  = 6,
     CPU_ATMEGA_CLK_SCALE_DIV128 = 7,
     CPU_ATMEGA_CLK_SCALE_DIV256 = 8,
     CPU_ATMEGA_CLK_SCALE_DIV512 = 9,
--- a/cpu/atmega_common/periph/timer.c
+++ b/cpu/atmega_common/periph/timer.c
@@ -29,18 +29,18 @@
 #include "periph/timer.h"
 #include "periph_conf.h"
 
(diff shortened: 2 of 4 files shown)


cpu/atmega_common/periph/uart.c
```diff
@@ -166,45 +166,35 @@ void uart_poweroff(uart_t uart)
...
-    if (sched_context_switch_request) {
-        thread_yield();
-    }
+    __exit_isr();
 }
 
 #ifdef UART_0_ISR
 ISR(UART_0_ISR, ISR_BLOCK)
 {
-    __enter_isr();
```
gebart: I prefer the original design where enter isr is the first line of each ISR

Josar: Is this just a preference or are there any reasons. As i see there is no benefit in settings the `__in_isr` flag as early as possible, as there is no chance that code could be invoked which could use this information. 
Setting the flag prevents code in the callback to sleep or do other things which schould not be done in interrupts, so setting it before the callback should be sufficient, imho.

I change it as it is like this in all other interrupt handler and thought it would also increase consistency.

gebart: Personal preference. The consistency helps when reviewing the code and also reduces the risk of missing it if someone uses this periph driver as a template for some other driver for the same CPU.

cpu/atmega_common/include/cpu.h
```diff
@@ -75,20 +81,20 @@ void cpu_init(void);
 /**
  * @brief   Print the last instruction's address
  */
-__attribute__((always_inline)) static inline void cpu_print_last_instruction(void)
+ static inline void __attribute__((always_inline)) cpu_print_last_instruction(void)
```
kaspar030: unrelated

cpu/atmega_common/include/cpu.h
```diff
@@ -75,20 +81,20 @@ void cpu_init(void);
...
 {
     uint8_t hi;
     uint8_t lo;
     uint16_t ptr;
 
-    __asm__ volatile( "in __tmp_reg__, __SP_H__  \n\t"
+    __asm__ volatile ("in __tmp_reg__, __SP_H__  \n\t"
```
kaspar030: unrelated

cpu/atmega_common/periph/uart.c
kaspar030: +1

Josar: @gebart maybe we just have a language problem as consistency translates into two different meanings in german. Do you mean consistency in regards as it stays at it was or consistency as it is similar to other code of the atmegas?
https://github.com/RIOT-OS/RIOT/blob/22aa6b9cc5c0ae95663209a3bdeb13e6b5630e63/cpu/atmega_common/periph/gpio.c#L264-L274
https://github.com/RIOT-OS/RIOT/blob/22aa6b9cc5c0ae95663209a3bdeb13e6b5630e63/cpu/atmega_common/periph/timer.c#L169-L189

gebart: I guess you are right regarding consistency. I am not very engaged in the atmega development, sorry. 
What I was trying to push for is to have a strict template for the ISR definitions which should be the same across all drivers for the CPU family, which you showed that this is. Consistency makes refactoring the driver code easier in the future. I am happy as long as this design matches the other periph drivers for the atmega.

cpu/atmega_common/include/cpu.h
```diff
@@ -34,8 +34,10 @@
 #include <stdint.h>
 
 #include <avr/interrupt.h>
+#include "board.h"
```
gebart: is this include actually used?

gebart:
My concerns have been addressed.

Josar:
@kaspar030 can you approve it like this?

Josar:
@kaspar030 ping 😏

cpu/atmega_common/include/cpu.h
```diff
@@ -52,19 +53,23 @@ extern "C" {
...
 {
     __in_isr = 1;
 }
 
 /**
- * @brief Flag exiting of an ISR
+ * @brief Run this code on exiting interrupt routines
  */
-static inline void __exit_isr(void)
+static inline void __attribute__((always_inline)) __exit_isr(void)
```
kaspar030: why always inline?

cpu/atmega_common/include/cpu.h
```diff
@@ -52,19 +53,23 @@ extern "C" {
 extern volatile uint8_t __in_isr;
 
 /**
- * @brief Flag entering of an ISR
+ * @brief Run this code on entering interrupt routines
  */
-static inline void __enter_isr(void)
+static inline void __attribute__((always_inline)) __enter_isr(void)
```
kaspar030: why always inline? Don't trust the compiler?

cpu/atmega_common/include/cpu.h
Josar: As it will be executed in an ISR, i thought speed could matter.
And if optimization is off it won't 
> GCC does not inline any functions when not optimizing unless you specify the ‘always_inline’ attribute for the function, like this:
> 
> ```
>/* Prototype.  */
> inline void foo (const char) __attribute__((always_inline));
>```

But as the statement is the attribute has to be after the function, i am not sure but, i think it does not matter.

This [benchmark ](https://indico.cern.ch/event/386232/sessions/159923/attachments/771039/1057534/always_inline_performance.pdf) makes me think maybe `inline` only is better, nevertheless i think i forgot to remove it.

Josar:
@gebart and @kaspar030 rebased onto #8904 and added neccessary changes.

ZetaR60:
This looks okay to me, but note that it introduces a yield in the periph/gpio.c ISR where there previously was none.

Josar:
@ZetaR60 i think that is not fully correct, it request a thread yield if there was a `sched_context_switch_request` in the callback handler. Which i would expect to happen, as the `__in_isr` blocks the context switch and at the end of an isr it has to be executed. Otherwise it would be lost or executed with an other thread_yield.

But thinking further is it neccessary to duplicate the code?

```
static inline void __exit_isr(void)
{
     if (sched_context_switch_request) {
         thread_yield();
        thread_yield_isr();
     }
   __in_isr = 0;
}
```
`thread_yield_isr();` is the same as `thread_yield_higher();` so just bringing `__in_isr = 0` before the `if` and setting reti to the end will have the same result with less code, or not ?

```
static inline void __exit_isr(void)
{
     __in_isr = 0;
     if (sched_context_switch_request) {
         thread_yield();
     }
    __asm__ volatile ("reti");
}
```

ZetaR60:
Yes, you are right about my comment being erroneous. Sorry for the lazy thought process.

There are problems with moving ```__asm__ volatile("reti");``` to a different area from ```__context_restore();``` because the context change directly manipulates the program counter, and returning from a function also manipulates the program counter. You would have to add ```__attribute__((always_inline)) inline``` to thread_yield / thread_yield_isr, which means you would have a copy of all the contents of __context_save and __context_restore in every ISR.

Josar:
@ZetaR60 seems it was too late for my brain, of course moving `reti` does not work.

Maybe we could agree on just putting the content of the `if` to thread_yield_isr();` ?

I don`t imagine a situation where saving the context and running the scheduler is necessary if there was not a thread change request. 

Further thoughts, I think there should be a function like `sched_thread_change_needed` to avoid saving and restoring the context when a yield or yield higher is issued but there is no higher.
As of now `sched_run` checks this, after the context was allready saved.

ZetaR60:
@Josar can you be a bit more precise which things you are referring to? Currently nothing is run at the end of an ISR if there was no attempt to yield within the callback, because sched_context_switch_request is not set. Are the other things you are mentioning proposed changes to core/thread.c ? Currently this PR is a refactor; perhaps discussion of changes of behavior should be moved to another PR specific to that change? There might be some improvements to be made in the way I rewrote context swapping, but that rewrite took a lot of testing to do properly.

ZetaR60:
Those sound like reasonable changes. They would involve changing behavior though, so the extensive testing done in #8904 probably ought to be re-checked in a PR that changes it.

Josar:
@ZetaR60 thinking further i think this can be optimized.

After deciding that a context swap is allowed in an ISR there is no need of preventing an context swap in an ISR, the only thing which has to be taken care of is to return with `reti`. The rest are remains which only add overhead.

As it is now we check for isr set the flag, return to caller, call next function, check for flag and do the swap.
Removing this will not only encrease swapping in interrupts but also increase interrupts with no yield as the flag check is removed. 

```
void thread_yield_higher(void) {
        __context_save();
        sched_run();
        __context_restore();

    if (irq_is_in() == 0) {
        __asm__ volatile("ret");
    } else {
        __exit_isr();
        __asm__ volatile ("reti");
    }
}
```

and removing the if irq in the scheduler

```
    if (!on_runqueue || (current_prio > other_prio)) {
        // if (irq_is_in()) {
        //     DEBUG("sched_switch: setting sched_context_switch_request.\n");
        //     sched_context_switch_request = 1;
        // }
        // else {
            DEBUG("sched_switch: yielding immediately.\n");
            thread_yield_higher();
        // }
    }
    else {
        DEBUG("sched_switch: continuing without yield.\n");
    }
```

first test with `test/xtimer_msg` shows expected results.

Josar:
@kYc0o would you mind having a look at this?

ZetaR60:
@Josar This change to thread_yield_higher is a pretty major change in behavior. Could you create another PR for it?

Josar:
@kaspar030 any further remarks?

Josar:
@kaspar030 some further requests?

Josar:
@kaspar030 maybe you could find some time?

ZetaR60:
I ran some tests, and got a couple failures:

```
thread_basic			OK
thread_cooperation		OK
thread_exit			OK
thread_flags			OK
thread_flags_xtimer		Reports failure
thread_flood			OK
thread_msg			OK
thread_msg_block_wo_queue	OK
thread_msg_block_w_queue	OK
thread_msg_seq			OK
xtimer_hang			OK
xtimer_periodic_wakeup		OK
xtimer_remove			Reports failure
xtimer_reset			OK
xtimer_usleep			OK
xtimer_usleep_short		OK
```

I think we need a bit more extensive examination. @Josar I suggest rebasing on trunk and running tests side by side with your code and the exact same revision without your code.

Josar:
Rebased and fixed.
All above Tests run successfull.

PS: it would be nice to have a readme in each  test with the expected output.

ZetaR60:
Confirmed that the previously failed tests are now succeeding on mega-xplained. Also did another proofread of the changes, and they look fine. Let me know if I should run any other tests on mega-xplained.

Josar:
@kaspar030 any further remarks?

Josar:
@kaspar030 could you have a second look?

ZetaR60:
@Josar You are being a bit aggressive with pinging Kaspar. I know things often move slowly, but I don't think repeating requests that much is going to help.

kYc0o:
@Josar please squash.

kYc0o:
All the tests in https://github.com/RIOT-OS/RIOT/pull/8988#issuecomment-392869767 passed on arduino-mega2560.

Josar:
I uncrustified the touched files, hope this is allright.

cpu/atmega_common/periph/timer.c
```diff
@@ -87,17 +88,21 @@ int timer_init(tim_t tim, unsigned long freq, timer_cb_t cb, void *arg)
     uint8_t pre = 0;
 
     /* make sure given device is valid */
-    if (tim >= TIMER_NUMOF) {
+    if (tim >= TIMER_NUMOF)
+    {
```
aabadie: Is this change the result of uncrustify ? Then I think there's an issue with the uncrustify configuration file. The opening curly brace should be on the same line as the if.

Josar:
@aabadie i think that the uncrustifying went right this time.
@kYc0o rebased and squashed.

kYc0o:
Perfect! Thanks for your contribution!

PR #8992, opened 2018-04-20
evtimer: optimized adding

Josar:
Minor changes to optimize adding an event timer.

Removes variable  `delta_sum` and three calculations.


kaspar030:
> Minor changes to optimize adding an event timer.

can you elaborate? +19 -6 doesn't seem minor.

Josar:
@kaspar030 most of the +19 are comments and DEBUG().

Josar:
@smlng better like this?

smlng:
looks good to me now, @miri64 and @kaspar030 any objections?

kaspar030:
> looks good to me now, @miri64 and

Nope, let's go. Thanks @Josar!

PR #8993, opened 2018-04-20
sys/cb_mux: Callback multiplexer

ZetaR60:
This is a rough sketch of a callback multiplexer, intended primarily for use in CPUs that have multiple pin interrupts tied to a single interrupt vector (but could be useful for other things too). It uses a singly linked list to allow callbacks to be dynamically added and deleted as needed so that all interrupts are usable without wasting memory when few are used.

Posted early in the development process to facilitate discussion in #7610.

### DIFF
--- a/sys/include/cb_mux.h
+++ b/sys/include/cb_mux.h
@@ -1,0 +1,143 @@
+/*
+ * Copyright (C) 2018 Acutam Automation, LLC
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    sys_cb_mux Callback multiplexer
+ * @ingroup     sys
+ * @brief       cb_mux provides utilities for storing, retrieving, and managing
+ *              callback information in a singly linked list.
+ *
+ * If an API provides the ability to call multiple callbacks, cb_mux can
+ * simplify handling of an arbitrary number of callbacks by requiring memory
+ * for a cb_mux entry to be passed along with other arguments. The cb_mux entry
+ * is then attached to a list using cb_mux_add. The code implementing that API
+ * can manage the list using the various utility functions that cb_mux provides.
+ *
+ * @{
+ *
+ * @file
+ * @brief       cb_mux interface definitions
+ *
+ * @author      Matthew Blue <matthew.blue.neuro@gmail.com>
+ */
+
+#ifndef CB_MUX_H
+#define CB_MUX_H
+
+#include <stdint.h>
+
+/* For alternate cb_mux_cbid_t */
+#include "periph_cpu.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef HAVE_CB_MUX_CBID_T
+/**
+ * @brief   cb_mux identifier type
+ */
+typedef unsigned int cb_mux_cbid_t;
+#endif
+
+/**
+ * @brief   cb_mux callback type
+ */
+typedef void (*cb_mux_cb_t)(void *);
+
+/**
+ * @brief   cb_mux list entry structure
+ */
+typedef struct cb_mux {
+    struct cb_mux *next;  /**< next entry in the cb_mux list */
+    cb_mux_cbid_t cbid;   /**< identifier for this callback */
+    void *info;           /**< optional extra information */
+    cb_mux_cb_t cb;       /**< callback function */
+    void *arg;            /**< argument for callback function */
+} cb_mux_t;
+
+/**
+ * @brief   cb_mux iterate function callback type for cb_mux_iter
+ */
+typedef void (*cb_mux_iter_t)(cb_mux_t *, void *);
+
+/**
+ * @brief   Add a new entry to the end of a cb_mux list
+ *
+ * @param[in] head   double pointer to first list entry
+ * @param[in] entry  entry to add
+ */
+void cb_mux_add(cb_mux_t **head, cb_mux_t *entry);
+
+/**
+ * @brief   Remove a entry from a cb_mux list
+ *
+ * @param[in] head   double pointer to first list entry
+ * @param[in] entry  entry to remove
+ */
+void cb_mux_del(cb_mux_t **head, cb_mux_t *entry);
+
+/**
+ * @brief   Find an entry in the list by ID
+ *
+ * @param[in] head      pointer to first list entry
+ * @param[in] cbid_val  ID to find
+ *
+ * @return pointer to the list entry
+ */
+cb_mux_t *cb_mux_find_cbid(cb_mux_t *head, cb_mux_cbid_t cbid_val);
+
+/**
+ * @brief   Find the entry with the lowest ID
+ *
+ * If there are multiple hits, this returns the oldest.
+ *
+ * @param[in] head   pointer to first list entry
+ *
+ * @return pointer to the list entry
+ */
+cb_mux_t *cb_mux_find_low(cb_mux_t *head);
+
+/**
+ * @brief   Find the entry with the highest ID
+ *
+ * If there are multiple hits, this returns the oldest.
+ *
+ * @param[in] head   pointer to first list entry
+ *
+ * @return pointer to the list entry
+ */
+cb_mux_t *cb_mux_find_high(cb_mux_t *head);
+
+/**
+ * @brief   Find the lowest unused ID
+ *
+ * Returns highest possible ID on failure
+ *
+ * @param[in] head   pointer to first list entry
+ *
+ * @return lowest unused ID
+ */
+cb_mux_cbid_t cb_mux_find_free_id(cb_mux_t *head);
+
+/**
+ * @brief   Run a function on every item in the cb_mux list
+ *
+ * @param[in] head  pointer to first list entry
+ * @param[in] func  function to run on each entry
+ * @param[in] arg   argument for the function
+ */
+void cb_mux_iter(cb_mux_t *head, cb_mux_iter_t func, void *arg);
+
+#ifdef __cplusplus
+}
+#endif
+
+/** @} */
+
+#endif /* CB_MUX_H */
--- a/sys/cb_mux/Makefile
+++ b/sys/cb_mux/Makefile
@@ -1,0 +1,1 @@
+include $(RIOTBASE)/Makefile.base
--- a/sys/cb_mux/cb_mux.c
+++ b/sys/cb_mux/cb_mux.c
@@ -1,0 +1,124 @@
+/*
+ * Copyright (C) 2018 Acutam Automation, LLC
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup   sys_cb_mux
+ * @{
+ *
+ * @file
+ * @brief   cb_mux implementation
+ *
+ * @author  Matthew Blue <matthew.blue.neuro@gmail.com>
+ * @}
+ */
+
+#include "cb_mux.h"
+#include "utlist.h"
+
+void cb_mux_add(cb_mux_t **head, cb_mux_t *entry)
+{
+    LL_APPEND(*head, entry);
+}
+
+void cb_mux_del(cb_mux_t **head, cb_mux_t *entry)
+{
+    LL_DELETE(*head, entry);
+}
+
+cb_mux_t *cb_mux_find_cbid(cb_mux_t *head, cb_mux_cbid_t cbid_val)
+{
+    cb_mux_t *entry;
+
+    LL_SEARCH_SCALAR(head, entry, cbid, cbid_val);
+
+    return entry;
+}
+
+cb_mux_t *cb_mux_find_low(cb_mux_t *head)
+{
+    cb_mux_t *entry;
+    cb_mux_t *entry_low = NULL;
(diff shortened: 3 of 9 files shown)


sys/cbmplex/cbmplex.c
```diff
@@ -0,0 +1,76 @@
...
+cbmplex_t *cbmplex_find_cbid(cbmplex_t *head, uint8_t cbid_val)
+{
+    cmbplex_t *entry;
+
+    LL_SEARCH_SCALAR(head, entry, cbid, cbid_val);
+
+    return entry;
+}
+
+cbmplex_t *cbmplex_find_flags(cbmplex_t *head, uint8_t flags, uint8_t mask)
```
bergzand: Is there a use case for having different flags and masks? Why not use `flags` as a mask?

sys/include/cbmplex.h
```diff
@@ -0,0 +1,112 @@
...
+ */
+typedef void (*cbmplex_cb_t)(void *);
+
+/**
+ * @brief   cbmplex list entry structure
+ */
+typedef struct cbmplex {
+    struct cbmplex *next;
+    uint8_t flags;
+    uint8_t cbid;
```
bergzand: What is the idea behind the callback ID?

gebart:
regarding name, "mux" is a more common shorthand for "multiplexer" (in electronics)

bergzand:
"cb_mux", or "callback_mux" maybe then

sys/cbmplex/cbmplex.c
ZetaR60: mask tells the function which bits to search, and flags tells the function what they should be set to. I will put a comment to clarify this.

sys/include/cbmplex.h
ZetaR60: This is whatever the system using the multiplexer uses to identify the callback. In the case of pin interrupts, it would be the pin number.

sys/cbmplex/cbmplex.c
bergzand: Of course -_-

sys/include/cbmplex.h
bergzand: Not sure I understand what you mean. How does the "system" here know the `cbid`? It is not passed as argument in the callback.

ZetaR60:
Name changed to cb_mux

ZetaR60:
Here is an example usage.

Initialization:
- foo code calls gpio_init_int for a pin change interrupt (PCINT)
- gpio.c creates a cb_mux entry and attaches it to its list of PCINTs. It puts the cb and arg into cb_mux's cb and arg, and the flank into flags. It gives it an cbid equal to the pin gpio_t
- since PCINTs on this system don't have a flank register and trigger on any change (like on the ATMega), the current state of that pin is also put into flags
- gpio.c then enables the PCINT for that pin

Pin change event (on foo's pin):
- gpio.c's handler for PCINTs triggers
- the handler searches using cb_mux_find_cbid for each entry that had a pin change event
- the handler checks the flank in those entries to see if the callback needs to be executed, and executes foo's callback
- the handler updates the current state of each pin (option to use cb_mux_update)

sys/include/cbmplex.h
ZetaR60: See example.

ZetaR60:
Note: There is an issue with the current definition of cbid. Different systems define gpio_t differently, some use up to a uint32_t. Setting cbid to uint32_t seems really big for small systems, and setting it to gpio_t seems wrong if it is not being used to hold pins. Suggestions?

ZetaR60:
Bergzand and I have been discussion potential issues with the API being too specific to pin interrupts. I have tried to make it more generic:
- I have removed flags and flag searches in favor of a generic ```void *info```, so it can be handled however the code using this API wishes.
- cbid is now 32 bits, which will support more use cases. I do feel that it would be nice for it to be smaller, but that would require restricting the potential uses.
- cb_mux_nextid is replaced by cb_mux_find_hilo_id, which should be more flexible in potential applications.
- cb_mux_update is renamed cb_mux_iter, which more generically describes what it does.

I am thinking of another example use case with timers, e.g. allowing multiple callbacks with RTC/RTT implementations:
- cbid is the alarm time for the given callback
- when the current alarm goes off, it uses cb_mux_find_hilo_id to find the next highest time and sets that as the next alarm

kYc0o:
Have you measured the overhead in response time of this approach against direct callbacks (which costs memory of course)?

kYc0o:
Ok, so if we merge this asap you'll be able to provide such a test soon? I want to move forward for interrupt handling on AVR.

ZetaR60:
@kYc0o I intend to provide the test in this PR.

ZetaR60:
Notice to interested parties, especially @kYc0o and @roberthartung:

There are some fundamental problems with using cb_mux for its original intended purpose: getting rid of callback limitations in APIs like periph/gpio.h. This is due to the inability for the implementation to allocate memory for new list entries (i.e. no malloc) in a manner transparent to the code using the API. I suspect that an array will be more appropriate for statically allocated memory. For these reasons I will be scrapping cb_mux unless someone really wants it or has a compelling reason to keep it.

Sorry for my oversight in not noticing this, and thanks to @roberthartung for pointing it out.

gebart:
@ZetaR60 you can work around the memory allocation issues using xfa (#7523, #9105)

ZetaR60:
@gebard If I am understanding xfa correctly, the memory is still statically allocated at compile time (although in a very neat way). The problem is that something like a periph/gpio.h implementation has no way of knowing how many times pin init functions are going to be called, so I don't see how statically allocated memory would work. Of course the API could be changed to pass memory, but then xfa would not be needed. Or you could redefine the function calls in the API to be macros that hide memory allocation, but that could introduce problems when they are assumed to be function calls.

I have been discussing the issue with @kYc0o and the different possible ways around the allocation issues. I think the best compromise so far for use cases like PCINTs is to define an array of structs that contain the callback information. For example, if you had 4 entries, you would be able to use 0-4 PCINTs at a time, but you would get an error if you tried to use 5. You could then allow this number to be changed in periph_conf.h on a board-by-board basis.

kaspar030:
> possible ways around the allocation problem

Maybe if we'd introduce a gpio_int_t type, which any user of a gpio interrupt needs to provide? That struct could contain necessary per-platform fields. It would be an API change, but we could move the allocation to wherever the interrupt is used.

ZetaR60:
A new type would work. I did interrupts in #9054 by using an array to hold the information, and the max number of entries in the list is defined in board.h. That would be a potential alternative that doesn't require an API change.

ZetaR60:
Rebased on trunk and added periph/gpio.h support for cb_mux. Nothing _actually_ supports it yet, except for the API; but everything should silently ignore the passed gpio_int_t entry.

@kaspar030 Is this approximately what you were suggesting?

It is going to be a lot of work implementing cb_mux for all of the cpus. If we are going with this method (rather than using arrays), how should we proceed?

ZetaR60:
I have added a comprehensive test routine, and also used it to fix many bugs with the API:

```
[liveuser@localhost-live cb_mux]$ make BOARD=mega-xplained test
./tests/01-run.py
/home/liveuser/Desktop/RIOT_cbmplex/dist/tools/pyterm/pyterm -p "/dev/ttyACM0" -b "9600"
2018-05-20 20:37:18,644 - INFO # Connect to serial port /dev/ttyACM0
Welcome to pyterm!
Type '/exit' to exit.
2018-05-20 20:37:22,646 - INFO # 

2018-05-20 20:37:22,746 - INFO # main(): This is RIOT! (Version: 2018.07-devel-235-gb4a04-localhost-live-RIOT_cbmplex)
2018-05-20 20:37:22,812 - INFO # cb_mux test routine
2018-05-20 20:37:22,814 - INFO # Test list addition, retrieval, execution of 5 CBs
2018-05-20 20:37:22,846 - INFO # Callback 0 executed
2018-05-20 20:37:22,912 - INFO # Callback 1 executed
2018-05-20 20:37:22,914 - INFO # Callback 2 executed
2018-05-20 20:37:22,916 - INFO # Callback 3 executed
2018-05-20 20:37:22,918 - INFO # Callback 4 executed
2018-05-20 20:37:23,014 - INFO # Test list deletion of CB 0, 2, 4, execution of 1, 3
2018-05-20 20:37:23,016 - INFO # Callback 1 executed
2018-05-20 20:37:23,018 - INFO # Callback 3 executed
2018-05-20 20:37:23,112 - INFO # Test execution of CB with lowest ID (1)
2018-05-20 20:37:23,114 - INFO # Callback 1 executed
2018-05-20 20:37:23,146 - INFO # Test execution of CB with highest ID (3)
2018-05-20 20:37:23,147 - INFO # Callback 3 executed
2018-05-20 20:37:23,213 - INFO # Re-adding list entries (0, 2, 4) by finding next free ID
2018-05-20 20:37:23,214 - INFO # Added entry 0
2018-05-20 20:37:23,247 - INFO # Added entry 2
2018-05-20 20:37:23,248 - INFO # Added entry 4
2018-05-20 20:37:23,314 - INFO # Test iteration of a function over list
2018-05-20 20:37:23,317 - INFO # Entry 0 was updated correctly
2018-05-20 20:37:23,347 - INFO # Entry 1 was updated correctly
2018-05-20 20:37:23,415 - INFO # Entry 2 was updated correctly
2018-05-20 20:37:23,418 - INFO # Entry 3 was updated correctly
2018-05-20 20:37:23,448 - INFO # Entry 4 was updated correctly
2018-05-20 20:37:23,547 - INFO # Tests complete!

[liveuser@localhost-live cb_mux]$ echo $?
0
```

```
Installed compiler toolchains 
-----------------------------
             native gcc: missing
      arm-none-eabi-gcc: missing
                avr-gcc: avr-gcc (Fedora 7.2.0-1.fc27) 7.2.0
       mips-mti-elf-gcc: missing
             msp430-gcc: missing
   riscv-none-embed-gcc: missing
                  clang: missing

Installed compiler libs
-----------------------
   arm-none-eabi-newlib: missing
    mips-mti-elf-newlib: missing

Installed development tools
---------------------------
                  cmake: cmake version 3.11.0
               cppcheck: missing
                doxygen: missing
                 flake8: missing
                    git: git version 2.14.2
             coccinelle: missing
```

ZetaR60:
Rebased to separate the API changes out into #9159

ZetaR60:
Added a simple benchmark at @kYc0o 's request.

```
[liveuser@localhost-live cb_mux_bench]$ make BOARD=mega-xplained test
./tests/01-run.py
/home/liveuser/Desktop/RIOT_cbmplex/dist/tools/pyterm/pyterm -p "/dev/ttyACM0" -b "9600"
2018-05-22 18:11:45,030 - INFO # Connect to serial port /dev/ttyACM0
Welcome to pyterm!
Type '/exit' to exit.
2018-05-22 18:11:48,732 - INFO # 

2018-05-22 18:11:48,800 - INFO # main(): This is RIOT! (Version: 2018.07-devel-231-g316235-localhost-live-RIOT_cbmplex)
2018-05-22 18:11:48,833 - INFO # cb_mux benchmark application
2018-05-22 18:11:48,899 - INFO # Populating cb_mux list with 20 items
2018-05-22 18:11:48,901 - INFO # Finding the last list entry
2018-05-22 18:11:48,932 - INFO # List walk time: 80 us
2018-05-22 18:11:49,032 - INFO # Walk time less than threshold of 200 us
2018-05-22 18:11:49,033 - INFO # [SUCCESS]

[liveuser@localhost-live cb_mux_bench]$ echo $?
0
```

kYc0o:
I'd ACK in general this PR, however is failing for samr21-xpro. I ran Murdock and HIL tests to know where else is failing.

kYc0o:
Basically it's a printf problem. Just try to use as much as possible portable data types, so their size depend on the architecture.

ZetaR60:
printf fixed. Tested again to double check on mega-xplained, and both tests pass. Murdock looks good with the cb_mux tests, but has unrelated failures on ps_schedstatistics and shell board tests.

You might notice the double cast: this is because gcc doesn't like to cast from a pointer to an integer of a different size, even if it is guaranteed to be equal or smaller.

ZetaR60:
@bergzand Do you think everything here is good to go?

sys/include/cb_mux.h
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright (C) 2018 Acutam Automation, LLC
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup   sys_cb_mux
```
aabadie: this doxygen group is not defined. you can use a `@defgroup` above this line and maybe also add a bit of documentation.

ZetaR60: ```@defgroup``` and summary added.

ZetaR60:
Unrelated board failures on Murdock again. I am attempting to resubmit the rebased commits to get it to succeed.

ZetaR60:
I gave Murdock several tries but it is still failing here:
```
tests/struct_tm_utility/samr21-xpro
tests/ps_schedstatistics/samr21-xpro
tests/shell/samr21-xpro
```

ZetaR60:
Created #9220 to try to fix the Murdock issue.

aabadie:
I removed the `CI:run tests` because of unrelated failures, at least the added cb_mux test script passes. Let's merge when Murdock is green.

aabadie:
All green now, merging. Thanks @ZetaR60 !

PR #8994, opened 2018-04-20
pkg/tlsf: add documentation.

jcarrano:
The doc.txt for TLSF was empty, so it wouldn't show up in Doxygen. I wrote a brief description and also cross linked it from the "Oneway Malloc" docs, so that people looking for memory allocation can find it.

### DIFF
--- a/pkg/tlsf/doc.txt
+++ b/pkg/tlsf/doc.txt
@@ -1,0 +1,17 @@
+/**
+ * @defgroup pkg_tlsf	Two-Level Segregated Fit memory allocator
+ * @ingroup  pkg
+ * @brief    TLSF is a general purpose dynamic memory allocator specifically
+ *            designed to meet real-time requirements:
+ * @see      http://www.gii.upv.es/tlsf/
+ * @see      https://github.com/mattconte/tlsf
+ *
+ * TLSF provides an implementation of malloc/realloc/free/etc with the following
+ * characteristics:
+ *
+ *  - O(1) Performance
+ *  - Works on a user supplied block of memory instead of a global heap.
+ *  - Efficient both in terms of memory overhead and processor time.
+ *  - Low fragmentation.
+ *
+ */
--- a/sys/oneway-malloc/include/malloc.h
+++ b/sys/oneway-malloc/include/malloc.h
@@ -20,6 +20,8 @@
  *
  * @note        You should prefer statically allocated memory whenever possible.
  *
+ * @see         pkg_tlsf
+ *
  * @{
  * @file
  *


jcarrano:
Also, there's a big patch being applied, and it's not very clear what it does. It changes the TLSF api, which is a problem since now there is no way for the user to find the correct docs for the new api.
Perhaps a proper fork of TLSF could be made with the patch already applied?

jia200x:
hi @jcarrano 

The patch allows the usage of TLSF as the global malloc/free implementation. Concerning the isolated TLSF environments, I'm pretty sure @OlegHahm or @Kijewski has more information about why it was implemented this way.

>and also cross linked it from the "Oneway Malloc" docs

Not sure if it's the best to put the tlsf link there, since Oneway Malloc is a specific implementation for platforms that don't implement malloc and free in the toolchain. Maybe a Memory Allocator entry which points to Oneway Malloc and TLSF? Opinions?

pkg/tlsf/doc.txt
```diff
@@ -0,0 +1,17 @@
+/**
+ * @defgroup pkg_tlsf	Two-Level Segregated Fit memory allocator
+ * @ingroup  pkg
+ * @brief    TLSF is a general purpose dynamic memory allocator specifically
+ *            designed to meet real-time requirements:
```
jia200x: Colon in the end. Should be dot.

pkg/tlsf/doc.txt
```diff
@@ -0,0 +1,17 @@
...
+ * @see      https://github.com/mattconte/tlsf
+ *
+ * TLSF provides an implementation of malloc/realloc/free/etc with the following
+ * characteristics:
+ *
+ *  - O(1) Performance
+ *  - Works on a user supplied block of memory instead of a global heap.
+ *  - Efficient both in terms of memory overhead and processor time.
+ *  - Low fragmentation.
+ *
```
jia200x: As far as I know, this implementation/adoption doesn't support 16-bit platforms and should be pointed out here. @OlegHahm can you confirm this?

sys/oneway-malloc/include/malloc.h
```diff
@@ -20,6 +20,8 @@
  *
  * @note        You should prefer statically allocated memory whenever possible.
  *
+ * @see         pkg_tlsf
```
jia200x: [See discussion](https://github.com/RIOT-OS/RIOT/pull/8994#issuecomment-383069117)

jcarrano:
@jia200x 
> Oneway Malloc is a specific implementation for platforms that don't implement malloc and free in the toolchain
If that's the case, then it's a good reason for linking the docs. If the platform does not implement malloc/free then the user will try to look at an external implementation.

About the patch, that's another issue, I asked about it in the devel mailing list. For this PR I only care about documentation.

pkg/tlsf/doc.txt
jcarrano: It's not building for 16 bit, see #4612

jcarrano:
I think we can drop this one now the the _real_ tlsf patch is being merged.

kYc0o:
@jcarrano Can you reference your real patch? and if it's already merged?

jcarrano:
@kYc0o #9006 Included a `doc.txt` as part of the changes.

PR #8995, opened 2018-04-20
sys/arduino: drop cpp feature dependency

### LINKED ISSUE #8548: arduino-uno does not provide FEATURES_PROVIDED=cpp
#### Description
C++ support is required to build Arduino code. Building the arduino-hello-world example with BOARD=arduino-uno works just fine, but the build system is warning that feature cpp is missing for the board.

```
cd examples/arduino_hello-world
make BOARD=arduino-uno
```

#### Expected results
No warnings

#### Actual results
```
There are unsatisfied feature requirements: cpp


EXPECT ERRORS!
```
but compilation succeeds (using both BUILD_IN_DOCKER=1 and my local avr-gcc installation)

#### Versions
riotbuild docker image b329486bb877

avr-gcc: avr-gcc (Gentoo 7.2.0-r1 p1.1) 7.2.0

kaspar030:
### Contribution description

Currently, the arduino module depends on the "cpp" feature. As AVR's don't have ```libstdc++```, they don't provide that feature, as otherwise all cpp examples would break.
But the AVR does provide a cpp compiler and thus the arduino module compiles just fine.

This PR removes the cpp dependency from arduino module.

### Issues/PRs references

Fixes #8548.

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -534,7 +534,6 @@ endif
 
 ifneq (,$(filter arduino,$(USEMODULE)))
   FEATURES_REQUIRED += arduino
-  FEATURES_REQUIRED += cpp
   USEMODULE += xtimer
 endif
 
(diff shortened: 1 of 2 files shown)


cladmi:
I will ACK with the PR description added in commit message.

PR #8996, opened 2018-04-20
make: don't create BINDIR for flash-only

smlng:
### Contribution description

This PR omits the creation of `BINDIR` for `make flash-only` (as used by murdock when running tests on boards), otherwise (i.e. currently) an empty directory is created which isn't used or needed. 

### Issues/PRs references

None

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -275,12 +275,6 @@ CFLAGS += -DBOARD_$(BOARDDEF)=\"$(BOARD)\" -DRIOT_BOARD=BOARD_$(BOARDDEF)
 CFLAGS += -DCPU_$(CPUDEF)=\"$(CPU)\" -DRIOT_CPU=CPU_$(CPUDEF)
 CFLAGS += -DMCU_$(MCUDEF)=\"$(MCU)\" -DRIOT_MCU=MCU_$(MCUDEF)
 
-# OSX fails to create empty archives. Provide a wrapper to catch that error.
-ifneq (0, $(shell mkdir -p $(BINDIR); $(AR) rc $(BINDIR)/empty-archive.a 2> /dev/null; \
-            echo $$?; rm -f $(BINDIR)/empty-archive.a 2>&1 > /dev/null))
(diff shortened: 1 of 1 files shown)


smlng:
> I do not like adding code to the BUILD_BEFORE_FLASH handling.

mhm, for me its more adding to handling of all make targets except of the special `flash-only`. At least that's what's the if-statement is about, hence it make (IMHO) sense to add this handling there. The check is only needed when building and linking, so it shouldn't be run for `flash-only` in any case.

cladmi:
>  At least that's what's the if-statement is about,

The if-statement is only a hack because we can't do `flash: $(ELFFILE)` without triggering a rebuild.

Parsing `MAKECMDGOALS` treats "goals" as equivalent of "command arguments" which should not be done ideally for makefiles. The real way would to have done it with a variable configuration:
 `make flash NO_REBUILD=1` but is horrible to use so I accepted the `flash-only` hack.

smlng:
> Can it be checked by doing the $(AR) rc /dev/null instead ?

maybe, but I would like to avoid making an unnecessary function call, because for `flash-only` the `ar`  tool isn't required anyways.

cladmi:
I tested, it does not work with `/dev/null`.

Does "ar rc dist/empty.a" fails on osx ? Because it could be a solution.
On Linux it does not modify the file and does not produce error.


Changing the behavior depending on `MAKECMDGOALS` here is still for me a really bad idea.
And the argument of  "removing just one command" is for me not enough to introduce a strange thing there

On native, where `flash-only` does nothing:

    $ strace -f make BOARD=native -C examples/hello-world flash-only 2>&1  | grep exec  | wc -l
    310

So -1 does not justify the edge case.

cladmi:
BTW checking the output of `strace` is scary, it looks like we are doing `lstat` on every file in the repository… I will check why.

smlng:
closing, as this should be handled differently

cladmi:
Question, when is `ar-wrapper` even needed ? What would break if it is removed ?
Because the condition for using it could be changed to just detect it.

PR #9002, opened 2018-04-23
make: Add 'BUILDDEPS' variable

cladmi:
Extend/fix `APPDEPS` as a new variable to declare general build dependencies.

This PR proposes to add a new `BUILDDEPS` variable instead of fixing `APPDEPS`.
It is only a proposal and is here to show how it could be used.
Fixing `APPDEPS` and making it evolve is also a possibility, with other sub-prs, I just want to start talking about it and justify the pull requests that would change `APPDEPS`.

### Contribution description

`BUILDDEPS` are files / make targets that should be build before compiling.
It can include packages source download, generating headers, modules.
    
It is the equivalent of `APPDEPS` but not limited to the application (as documented).
It cannot be done right now with `APPDEPS` as it is used in `BASELIBS` and fixing it requires changing mips using it for source files.

#### Other solution

Another solution is to fix `APPDEPS` now:
* Replace:
    * https://github.com/RIOT-OS/RIOT/blob/af70c0844c5622aef2fd89b75f5ed96adfda47d0/boards/pic32-clicker/Makefile.include#L4
    * https://github.com/RIOT-OS/RIOT/blob/af70c0844c5622aef2fd89b75f5ed96adfda47d0/boards/pic32-wifire/Makefile.include#L4
* Revert https://github.com/RIOT-OS/RIOT/pull/4906 that puts is in `BASELIBS` and so in the linker command
* Extend `APPDEPS` documentation.
* Maybe rename it to something like `BUILDDEPS`.

### Issues/PRs references

This would be necessary to do https://github.com/RIOT-OS/RIOT/pull/7654 properly.
With https://github.com/RIOT-OS/RIOT/pull/8987 it would allow defining external packages by setting a dependency that downloads the source files.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -364,6 +364,9 @@ LINKFLAGPREFIX ?= -Wl,
 
 DIRS += $(EXTERNAL_MODULE_DIRS)
 
+# Define dependencies required for building (headers, downloading source files,)
+BUILDDEPS += $(RIOTBUILD_CONFIG_HEADER_C)
+
 # Save value to verify it is not modified later
 _BASELIBS_VALUE_BEFORE_USAGE := $(BASELIBS)
 
@@ -387,7 +390,7 @@ endif # RIOTNOLINK
 $(ELFFILE): $(BASELIBS)
 	$(Q)$(_LINK) -o $@
 
-$(BINDIR)/$(APPLICATION_MODULE).a: $(RIOTBUILD_CONFIG_HEADER_C) $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS) $(BUILDDEPS)
+$(BINDIR)/$(APPLICATION_MODULE).a: $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS) $(BUILDDEPS)
 	$(Q)DIRS="$(DIRS)" "$(MAKE)" -C $(APPDIR) -f $(RIOTMAKE)/application.inc.mk
 $(BINDIR)/$(APPLICATION_MODULE).a: FORCE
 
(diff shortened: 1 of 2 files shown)


cladmi:
@neiljay I may also need some of your input. Do you remember why you put the assembly files as APPDEPS and not just using the module ? 
https://github.com/RIOT-OS/RIOT/blob/af70c0844c5622aef2fd89b75f5ed96adfda47d0/boards/pic32-clicker/Makefile.include#L4

cladmi:
I found another line that should be using `APPDEPS`: https://github.com/RIOT-OS/RIOT/blob/6d96eaa491d0ac5eb7e0deab875c4d9d62fcd126/examples/javascript/Makefile#L50

cladmi:
Rebased and inline fixed a whitespace error.

Makefile.include
```diff
@@ -364,6 +364,11 @@ LINKFLAGPREFIX ?= -Wl,
 
 DIRS += $(EXTERNAL_MODULE_DIRS)
 
+# Define dependencies required for building (headers, dowloading source files,)
```
kaspar030: typo: downloading

makefiles/vars.inc.mk
```diff
@@ -17,6 +17,7 @@ export USEMODULE             # Sys Module dependencies of the application. Set i
 export USEPKG                # Pkg dependencies (third party modules) of the application. Set in the application's Makefile.
 export DISABLE_MODULE        # Used in the application's Makefile to suppress DEFAULT_MODULEs.
 export APPDEPS               # Files / Makefile targets that need to be created before the application can be build. Set in the application's Makefile.
+# BUILDDEPS                  # Files / Makefile targets that need to be created before starting to build. Set by modules / build system included files.
```
kaspar030: took me a while to parse "Set by modules / build system included files.". What does "Set by build system included files." mean? Maybe just drop the whole "Set by" phrase?

cladmi: I was writing in the same way as the line before but it does not help, I removed it.

cladmi:
I inline fixed the download and removing the last part of the line as mentioned.

Issue #8984, opened 2018-04-19
RFC: Set `PSEUDOMODULES` in packages `Makefile.include`


cladmi:
#### Description

I was the one suggesting @bergzand to set `PSEUDOMODULES` in his `libcose` package `Makefile.include` while developing so he would not need to modify the other RIOT files.

https://github.com/RIOT-OS/RIOT/blob/6c39d2d621d15cd6f252545c791bbc381b1b48bf/pkg/libcose/Makefile.include#L11-L12

I also thought it could be the occasion to ask if it is something we may want to do for packages as they already provide `Makefile.include` and `Makefile.dep`

As it has been merged, it is even more important to ask the question.

I plan to do a PR to also allow having a `Makefile.include` and `Makefile.dep` for `EXTERNAL_MODULE_DIRS`. So these directories would need to do this also.

#### Question

For packages, do we want to set the `PSEUDOMODULES` variables in the package `Makefile.include` instead of the centralized file ?

I would like to have your feedback and arguments if you are against.

#### Next steps

Depending on the result, either, change the libcose package, or add documentation saying that it's possible and wanted, a comment at the bottom of `pseudomodules.inc.mk` and update packages accordingly.

cladmi:
It would also be the way of doing it for external modules with: https://github.com/RIOT-OS/RIOT/pull/8987

cladmi:
I added a PR that proposes to add this in the documentation https://github.com/RIOT-OS/RIOT/pull/9003

jia200x:
> For packages, do we want to set the PSEUDOMODULES variables in the package Makefile.include instead of the centralized file ?

The idea sounds nice to me!

cladmi:
It is now supported and documented.

PR #9003, opened 2018-04-23
pseudomodules: allow defining them in pkg/PKG/Makefile.include

cladmi:
### Contribution description

Officially allow packages to define pseudomodules in their Makefile.include as
done in the libcose package.

https://github.com/RIOT-OS/RIOT/blob/6c39d2d621d15cd6f252545c791bbc381b1b48bf/pkg/libcose/Makefile.include#L11-L12

### Issues/PRs references

Proposed implementation of this RFC https://github.com/RIOT-OS/RIOT/issues/8984

### DIFF
--- a/makefiles/pseudomodules.inc.mk
+++ b/makefiles/pseudomodules.inc.mk
@@ -131,3 +131,5 @@ NO_PSEUDOMODULES += periph_common
 PSEUDOMODULES += auto_init_skald
 PSEUDOMODULES += skald_ibeacon
 PSEUDOMODULES += skald_eddystone
+
+# Packages may also add modules to PSEUDOMODULES in their `Makefile.include`.


kaspar030:
I think Makefile.dep is a better place, despite it's misleading name.

cladmi:
> I think Makefile.dep is a better place, despite it's misleading name.

Can you elaborate on this ? for me it is more `.include` that describes a module/package where `.dep` handles only setting dependency.

One of the "problem" with Makefile.dep is that it needs to be parsed several times during USEMODULE resolution. I imagine it is even the reason to have these dependencies in a separate file from Makefile.include.

kaspar030:
> Can you elaborate on this ? for me it is more .include that describes a module/package where .dep handles only setting dependency.

Sorry, yes. Makefile.dep is considered for info-boards-supported, but that actually doesn't matter for pseudomodules. Thus I think I changed my mind and now proclaim the opposite.

cladmi:
The 'PSEUDOMODULES' is still there in libcose, so I think it now should be officially documented that it can be done.

makefiles/pseudomodules.inc.mk
```diff
@@ -129,3 +129,6 @@ NO_PSEUDOMODULES += periph_common
 PSEUDOMODULES += auto_init_skald
 PSEUDOMODULES += skald_ibeacon
 PSEUDOMODULES += skald_eddystone
+
+
```
bergzand: Any reason for the duplicate newline here?

cladmi: I did it on purpose to separate the block where pseudomodules are defined and with the second block with the documentation.
I often only keep one newline between related things and use two to distinguish between blocks.

It is only personal style leaking here and I can change it.

bergzand: I think I slightly prefer a single line, but I don't mind having two newlines here if you think it makes things more readable.

cladmi: I update it then.

cladmi:
I rebased and changed the whitespace inline.

cladmi:
Murdock happy.

PR #8989, opened 2018-04-19
drivers: add driver for APDS9007 illuminance sensor

Hyungsin:
This PR provides a driver implementation for APDS9007 illuminance sensor.
APDS9007 gives raw voltage value which is proportional to "lux".
This driver reads this voltage value through ADC, which needs to be converted to "lux" according to passive components. The voltage->lux conversion is out of the scope of this PR.

Note: 
APDS9007's illuminance output is 3 to 70,000 lux, which cannot be captured by a 16 bit variable.
But raw voltage output is 0 to 3000 mV, which is durable.

### Dependencies

https://github.com/RIOT-OS/RIOT/pull/9013

PR #7826, opened 2017-10-24
samd21/clock: add OSCULP32K/DFLL option

Hyungsin:
This PRR adds OSCULP32K/DFLL clock option for samd21.
This also allows xtimer to use 32kHz OSCULP32K/RTT timer to provide low power sleep mode.  

@thomaseichinger may be a good reviewer.

PR #7823, opened 2017-10-24
Sensor driver: PIR-based occupancy sensing

Hyungsin:
This PR provides a driver for EKMB PIR motion sensor.

The current driver operation is:
1) When the sensor detects motion, the connected GPIO pin goes high. Otherwise, it goes low.
2) The current driver supports to detect occupancy in % (how long has the GPIO been high?). 
3) It supports AUTO_INIT and SAUL.

PR #9013, opened 2018-04-24
board: add hamilton board support

Hyungsin:
This PR provides an initial driver for the Hamilton board [1,2]. @immesys. 

image

Hamilton has  samr21e18a SoC (Cortex M0+ CPU and AT86RF233 radio), HDC1080 (temperature and humidity sensors), FXOS8700 (3-axis accelerator and magnetometer), TMP006 (object temperature), APDS9007 (light sensor), Button, LED, and EKMB (PIR motiron sensor). It provides JTAG as the programming/debugging interface. Its idle current is ~6 uA when a 32 kHZ RTC is running.

Some sensors have not been merged yet: APDS9007 (#8989), EKMB (#7823), FXOS8700 (#8978).
Low power clock configuration has not been merged yet: #7826. 
 
Therefore, this PR, as an initial support, provides a test application which tests operation of HDC1080, TMP006, Button, and LED without low idle current.
 

### References
[1] https://bets.cs.berkeley.edu/publications/2017buildsys_hamiltondemo.pdf
[2] https://bets.cs.berkeley.edu/publications/2018sensys_hamilton.pdf

### DIFF
--- a/boards/hamilton/Makefile.dep
+++ b/boards/hamilton/Makefile.dep
@@ -1,0 +1,14 @@
+ifneq (,$(filter gnrc_netdev_default netdev_default,$(USEMODULE)))
+  USEMODULE += at86rf233
+endif
+
+ifneq (,$(filter saul_default,$(USEMODULE)))
+  USEMODULE += fxos8700
+  USEMODULE += hdc1000
+  USEMODULE += pir
+  USEMODULE += pulse_counter
+  USEMODULE += saul_gpio
+  USEMODULE += tmp006
+  # ToDo: peripherals to be added in the future
+  #USEMODULE += apds9007
+endif
--- a/boards/hamilton/Makefile.features
+++ b/boards/hamilton/Makefile.features
@@ -1,0 +1,14 @@
+# Put defined MCU peripherals here (in alphabetical order)
+FEATURES_PROVIDED += periph_adc
+FEATURES_PROVIDED += periph_gpio periph_gpio_irq
+FEATURES_PROVIDED += periph_i2c
+FEATURES_PROVIDED += periph_pwm
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
+FEATURES_PROVIDED += periph_spi
+FEATURES_PROVIDED += periph_timer
+
+# The board MPU family (used for grouping by the CI system)
+FEATURES_MCU_GROUP = cortex_m0_2
+
+-include $(RIOTCPU)/samd21/Makefile.features
--- a/boards/hamilton/Makefile.include
+++ b/boards/hamilton/Makefile.include
@@ -1,0 +1,16 @@
+# define the cpu used by Hamilton
+export CPU = samd21
+export CPU_MODEL = samr21e18a
+
+# debugger config
+export JLINK_DEVICE := atsamr21e18a
+export OBJDUMPFLAGS += --disassemble --source --disassembler-options=force-thumb
+export OFLAGS := -O binary --gap-fill 0xff
+
+# Configure terminal, hamilton doesn't provide any UART, thus use RTT
+TERMPROG = $(RIOTTOOLS)/jlink/jlink.sh
+TERMFLAGS = term_rtt
+
+USEMODULE += stdio_rtt
+
+include $(RIOTMAKE)/tools/jlink.inc.mk
--- a/boards/hamilton/board.c
+++ b/boards/hamilton/board.c
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2016 University of California, Berkeley
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     boards_hamilton
+ * @{
+ *
+ * @file
+ * @brief       Board specific implementations for the Hamilton mote
+ *
+ * @author      Michael Andersen <m.andersen@berkeley.edu>
+ *
+ * @}
+ */
+
+#include <stdio.h>
+
+#include "board.h"
+#include "periph/gpio.h"
+
+void board_init(void)
+{
+    /* initialize the on-board LED */
+    gpio_init(LED0_PIN, GPIO_OUT);
+
+    /* initialize the CPU */
+    cpu_init();
+}
--- a/boards/hamilton/include/board.h
+++ b/boards/hamilton/include/board.h
@@ -1,0 +1,141 @@
+/*
+ * Copyright (C) 2016 UC Berkeley
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    boards_hamilton HamiltonIoT Hamilton
+ * @ingroup     boards
+ * @brief       Support for the HamiltonIoT Hamilton board.
+ * @{
+ *
+ * @file
+ * @brief       Board specific definitions for the Hamilton board
+ *
+ * @author      Michael Andersen <m.andersen@cs.berkeley.edu>
+ * @author      Hyung-Sin Kim <hs.kim@cs.berkeley.edu>
+ */
+
+#ifndef BOARD_H
+#define BOARD_H
+
+#include "cpu.h"
+#include "periph_conf.h"
+#include "periph_cpu.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @name    xtimer configuration
+ * @{
+ */
+#define XTIMER_DEV          TIMER_DEV(1)
+#define XTIMER_CHAN         (0)
+#define XTIMER_OVERHEAD     (0)
+/** @} */
+
+/**
+ * @name AT86RF233 configuration
+ * @{
+ */
+#define AT86RF2XX_PARAM_SPI      SPI_DEV(0)
+#define AT86RF2XX_PARAM_SPI_CLK  SPI_CLK_5MHZ
+#define AT86RF2XX_PARAM_CS       GPIO_PIN(PB, 31)
+#define AT86RF2XX_PARAM_INT      GPIO_PIN(PB, 0)
+#define AT86RF2XX_PARAM_SLEEP    GPIO_PIN(PA, 20)
+#define AT86RF2XX_PARAM_RESET    GPIO_PIN(PB, 15)
+/** @} */
+
+/**
+ * @name    LED pin definitions and handlers
+ * @{
+ */
+#define LED0_PIN                 GPIO_PIN(0, 19)
+
+#define LED0_PORT                PORT->Group[0]
+#define LED0_MASK                (1 << 19)
+
+#define LED0_ON                  (LED0_PORT.OUTCLR.reg = LED0_MASK)
+#define LED0_OFF                 (LED0_PORT.OUTSET.reg = LED0_MASK)
+#define LED0_TOGGLE              (LED0_PORT.OUTTGL.reg = LED0_MASK)
+/** @} */
+
+/**
+ * @name    Button pin definitions
+ * @{
+ */
+#define BTN0_PORT                PORT->Group[0]
+#define BTN0_PIN                 GPIO_PIN(0, 18)
+#define BTN0_MODE                GPIO_IN_PU
+/** @} */
+
+/**
+ * @name FXOS8700 configuration
+ * Note that another fxos8700 operation option, FXOS8700_USE_ACC_RAW_VALUES,
+ * need to be set according to the application purposes
+ * @{
+ */
+#define FXOS8700_PARAM_I2C            I2C_DEV(0)
+#define FXOS8700_PARAM_ADDR           (0x1E)
+#define FXOS8700_PARAM_RENEW_INTERVAL (1000000ul)
+/** @} */
+
+/**
+ * @name HDC1080 configuration
+ * @{
+ */
+#define HDC1000_PARAM_I2C             I2C_DEV(0)
+#define HDC1000_PARAM_ADDR            (0x40)
+#define HDC1000_PARAM_RES             HDC1000_14BIT
+#define HDC1000_PARAM_RENEW_INTERVAL  (1000000ul)
+/** @} */
+
+/**
+ * @name EKMB (PIR motion sensor) configuration
+ * @{
+ */
+#define PIR_PARAM_GPIO          GPIO_PIN(PA, 6)
+#define PIR_PARAM_ACTIVE_HIGH   (1)
+/** @} */
+
+/**
+ * @name PULSE_COUNTER configuration
+ * @{
+ */
+#define PULSE_COUNTER_GPIO          BTN0_PIN
+#define PULSE_COUNTER_GPIO_FLANK    GPIO_FALLING
+/** @} */
+
+/**
+ * @name TMP006 configuration
+ * Note that two other tmp006 operation options, TMP006_USE_LOW_POWER and
+ * TMP006_USE_RAW_VALUES, need to be set according to the application purpose
+ * @{
+ */
+#define TMP006_PARAM_I2C            I2C_DEV(0)
+#define TMP006_PARAM_ADDR           (0x44)
+#define TMP006_PARAM_RATE           TMP006_CONFIG_CR_AS2
+/** @} */
+
+/**
+ * @name ToDo: APDS9007 configuration
+ * @{
+ */
+/** @} */
+
+/**
+ * @brief Initialize board specific hardware, including clock, LEDs and std-IO
+ */
+void board_init(void);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* BOARD_H */
+/** @} */
--- a/boards/hamilton/include/periph_conf.h
+++ b/boards/hamilton/include/periph_conf.h
@@ -1,0 +1,242 @@
+/*
+ * Copyright (C) 2014-2015 Freie Universität Berlin
+ *               2016 UC Berkeley
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     boards_hamilton
+ * @{
+ *
+ * @file
+ * @brief       Configuration of CPU peripherals for the Hamilton mote
+ *
+ * @author      Thomas Eichinger <thomas.eichinger@fu-berlin.de>
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Peter Kietzmann <peter.kietzmann@haw-hamburg.de>
+ * @author      Michael Andersen <m.andersen@berkeley.edu>
+ * @author      Hyung-Sin Kim <hs.kim@berkeley.edu>
+ */
+
+#ifndef PERIPH_CONF_H
+#define PERIPH_CONF_H
+
+#include <stdint.h>
+
+#include "cpu.h"
+#include "periph_cpu.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @name   External oscillator and clock configuration
+ *
+ * There are three choices for selection of CORECLOCK:
+ *
+ * - usage of the 48 MHz DFLL fed by external oscillator running at 32 kHz
+ * - usage of the PLL fed by the internal 8MHz oscillator divided by 8
+ * - usage of the internal 8MHz oscillator directly, divided by N if needed
+ *
+ *
+ * The PLL option allows for the usage of a wider frequency range and a more
+ * stable clock with less jitter. This is why this option is default.
+ *
+ * The target frequency is computed from the PLL multiplier and the PLL divisor.
+ * Use the following formula to compute your values:
+ *
+ * CORECLOCK = ((PLL_MUL + 1) * 1MHz) / PLL_DIV
+ *
+ * NOTE: The PLL circuit does not run with less than 32MHz while the maximum PLL
+ *       frequency is 96MHz. So PLL_MUL must be between 31 and 95!
+ *
+ *
+ * The internal Oscillator used directly can lead to a slightly better power
+ * efficiency to the cost of a less stable clock. Use this option when you know
+ * what you are doing! The actual core frequency is adjusted as follows:
+ *
+ * CORECLOCK = 8MHz / DIV
+ *
+ * NOTE: A core clock frequency below 1MHz is not recommended
+ *
+ * @{
+ */
+#define CLOCK_USE_PLL       (1)
+
+#if CLOCK_USE_PLL
+/* edit these values to adjust the PLL output frequency */
+#define CLOCK_PLL_MUL       (47U)               /* must be >= 31 & <= 95 */
+#define CLOCK_PLL_DIV       (1U)                /* adjust to your needs */
+/* generate the actual used core clock frequency */
+#define CLOCK_CORECLOCK     (((CLOCK_PLL_MUL + 1) * 1000000U) / CLOCK_PLL_DIV)
+#elif CLOCK_USE_XOSC32_DFLL
+/* Settings for 32 kHz external oscillator and 48 MHz DFLL */
+#define CLOCK_CORECLOCK     (48000000U)
+#define CLOCK_XOSC32K       (32768UL)
+#define CLOCK_8MHZ          (1)
+#define GEN2_ULP32K         (1)
+#else
+/* edit this value to your needs */
+#define CLOCK_DIV           (1U)
+/* generate the actual core clock frequency */
+#define CLOCK_CORECLOCK     (8000000 / CLOCK_DIV)
+#endif
+/** @} */
+
+/**
+ * @name    RTC configuration
+ * @{
+ */
+#define RTC_NUMOF           (1U)
+#define RTC_DEV             RTC->MODE2
+/** @} */
+
+/**
+ * @name    RTT configuration
+ * @{
+ */
+#define RTT_NUMOF           (1U)
+#define RTT_DEV             RTC->MODE0
+#define RTT_IRQ             RTC_IRQn
+#define RTT_IRQ_PRIO        10
+#define RTT_ISR             isr_rtc
+#define RTT_MAX_VALUE       (0xffffffff)
+#define RTT_FREQUENCY       (32768U)    /* in Hz. For changes see `rtt.c` */
+#define RTT_RUNSTDBY        (1)         /* Keep RTT running in sleep states */
+/** @} */
+
+/**
+ * @name    Timer peripheral configuration
+ * @{
+ */
+#define TIMER_NUMOF         (2U)
+#define TIMER_0_EN          1
+#define TIMER_1_EN          1
+
+/* Timer 0 configuration */
+#define TIMER_0_DEV         TC3->COUNT16
+#define TIMER_0_CHANNELS    2
+#define TIMER_0_MAX_VALUE   (0xffff)
+#define TIMER_0_ISR         isr_tc3
+
+/* Timer 1 configuration */
+#define TIMER_1_DEV         TC4->COUNT32
+#define TIMER_1_CHANNELS    2
+#define TIMER_1_MAX_VALUE   (0xffffffff)
+#define TIMER_1_ISR         isr_tc4
+/** @} */
+
+/**
+ * @name ADC Configuration
+ * @{
+ */
+#define ADC_0_EN                           1
+#define ADC_MAX_CHANNELS                   14
+/* ADC 0 device configuration */
+#define ADC_0_DEV                          ADC
+#define ADC_0_IRQ                          ADC_IRQn
+
+/* ADC 0 Default values */
+#define ADC_0_CLK_SOURCE                   0 /* GCLK_GENERATOR_0 */
+#define ADC_0_PRESCALER                    ADC_CTRLB_PRESCALER_DIV512
+
+#define ADC_0_NEG_INPUT                    ADC_INPUTCTRL_MUXNEG_GND
+#define ADC_0_GAIN_FACTOR_DEFAULT          ADC_INPUTCTRL_GAIN_1X
+#define ADC_0_REF_DEFAULT                  ADC_REFCTRL_REFSEL_INT1V
+
+static const adc_conf_chan_t adc_channels[] = {
+    /* port, pin, muxpos */
+    {GPIO_PIN(PA, 6), ADC_INPUTCTRL_MUXPOS_PIN6},
+    {GPIO_PIN(PA, 7), ADC_INPUTCTRL_MUXPOS_PIN7},
+    {GPIO_PIN(PA, 8), ADC_INPUTCTRL_MUXPOS_PIN16},
+};
+
+#define ADC_0_CHANNELS                     (3U)
+#define ADC_NUMOF                          ADC_0_CHANNELS
+/** @} */
+
+/**
+ * @name PWM configuration
+ * @{
+ */
+#define PWM_0_EN            1
+#define PWM_1_EN            1
+#define PWM_NUMOF           (PWM_0_EN + PWM_1_EN)
+#define PWM_MAX_CHANNELS    2
+
+/* PWM device configuration */
+#if PWM_NUMOF
+static const pwm_conf_t pwm_config[] = {
+#if PWM_0_EN
+    {TCC1, {
+        /* GPIO pin, MUX value, TCC channel */
+        {GPIO_PIN(PA, 6), GPIO_MUX_E, 0},
+        {GPIO_PIN(PA, 7), GPIO_MUX_E, 1}
+    }},
+#endif
+#if PWM_1_EN
+    {TCC0, {
+        /* GPIO pin, MUX value, TCC channel */
+        {GPIO_PIN(PA, 18), GPIO_MUX_F, 2},
+        {GPIO_PIN(PA, 19), GPIO_MUX_F, 3}
+    }},
+#endif
+};
+#endif
+/** @} */
+
+/**
+ * @name SPI configuration
+ * @{
+ */
+static const spi_conf_t spi_config[] = {
+    {
+        .dev      = &SERCOM4->SPI,
+        .miso_pin = GPIO_PIN(PC, 19),
+        .mosi_pin = GPIO_PIN(PB, 30),
+        .clk_pin  = GPIO_PIN(PC, 18),
+        .miso_mux = GPIO_MUX_F,
+        .mosi_mux = GPIO_MUX_F,
+        .clk_mux  = GPIO_MUX_F,
+        .miso_pad = SPI_PAD_MISO_0,
+        .mosi_pad = SPI_PAD_MOSI_2_SCK_3
+    }
+};
+
+#define SPI_NUMOF           (sizeof(spi_config) / sizeof(spi_config[0]))
+/** @} */
+
+/**
+ * @name    I2C configuration
+ * @{
+ */
+static const i2c_conf_t i2c_config[] = {
+    {
+        .dev      = &(SERCOM3->I2CM),
+        .speed    = I2C_SPEED_FAST,
+        .scl_pin  = GPIO_PIN(PA, 17),
+        .sda_pin  = GPIO_PIN(PA, 16),
+        .mux      = GPIO_MUX_D,
+        .gclk_src = GCLK_CLKCTRL_GEN_GCLK0,
+        .flags    = I2C_FLAG_NONE
+     }
+};
+#define I2C_NUMOF          (sizeof(i2c_config) / sizeof(i2c_config[0]))
+
+/**
+ * @name Random Number Generator configuration
+ * @{
+ */
+#define RANDOM_NUMOF       (0U)
+/** @} */
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* PERIPH_CONF_H */
+/** @} */
--- a/boards/hamilton/Makefile
+++ b/boards/hamilton/Makefile
@@ -1,0 +1,3 @@
+MODULE = board
+
+include $(RIOTBASE)/Makefile.base
--- a/boards/hamilton/include/gpio_params.h
+++ b/boards/hamilton/include/gpio_params.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2015 Freie Universität Berlin
+ *               2015 Kaspar Schleiser <kaspar@schleiser.de>
+ *               2016 UC Berkeley
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup   boards_hamilton
+ * @{
+ *
+ * @file
+ * @brief     Board specific configuration of direct mapped GPIOs
+ *
+ * @author    Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author    Kaspar Schleiser <kaspar@schleiser.de>
+ * @author    Sebastian Meiling <s@mlng.net>
+ * @author    Michael Andersen <m.andersen@berkeley.edu>
+ */
+
+#ifndef GPIO_PARAMS_H
+#define GPIO_PARAMS_H
+
+#include "board.h"
+#include "saul/periph.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief    GPIO pin configuration
+ */
+static const  saul_gpio_params_t saul_gpio_params[] =
+{
+    {
+        .name = "LED(red)",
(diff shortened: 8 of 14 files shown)


Hyungsin:
Any review here?

MichelRottleuthner:
Where to get one of these boards, is it publicly available somewhere?

Hyungsin:
Yes, you can get one from https://hamiltoniot.com/. I sent 5 hamiltons to @emmanuelsearch too.

Hyungsin:
Since #8978 was merged, now this PR includes FXOS8700.

boards/hamilton/include/periph_conf.h
```diff
@@ -0,0 +1,226 @@
...
+ * The PLL option allows for the usage of a wider frequency range and a more
+ * stable clock with less jitter. This is why this option is default.
+ *
+ * The target frequency is computed from the PLL multiplier and the PLL divisor.
+ * Use the following formula to compute your values:
+ *
+ * CORECLOCK = ((PLL_MUL + 1) * 1MHz) / PLL_DIV
+ *
+ * NOTE: The PLL circuit does not run with less than 32MHz while the maximum PLL
+ *       frequency is 96MHz. So PLL_MULL must be between 31 and 95!
```
MichelRottleuthner: PLL_MUL~L~

MichelRottleuthner:
I don't have a board here to actually test. @emmanuelsearch can you or someone at FU test this on real hardware? Or can you spare one of the five boards you have for some time? (I assume @tcschmidt would be willing to ship it from FU to HAW and back)

boards/hamilton/Makefile.dep
```diff
@@ -0,0 +1,15 @@
+ifneq (,$(filter gnrc_netdev_default netdev_default,$(USEMODULE)))
+  USEMODULE += at86rf233
+  USEMODULE += auto_init_gnrc_netif
```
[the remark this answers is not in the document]
Hyungsin: What is a counter example? As saul_default includes "auto_init_saul" for sensor initialization, I think it is very convenient to include "auto_init_gnrc_netif" for radio initialization.

boards/hamilton/Makefile.features
```diff
@@ -0,0 +1,17 @@
...
+FEATURES_PROVIDED += periph_adc
+FEATURES_PROVIDED += periph_cpuid
+FEATURES_PROVIDED += periph_gpio
+FEATURES_PROVIDED += periph_i2c
+FEATURES_PROVIDED += periph_pwm
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
+FEATURES_PROVIDED += periph_spi
+FEATURES_PROVIDED += periph_timer
+FEATURES_PROVIDED += periph_pm
```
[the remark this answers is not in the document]
Hyungsin: done

boards/hamilton/dist/reset.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+# This script resets a CC2538SF53 target using JLink called
```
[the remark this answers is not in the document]
Hyungsin: done

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * @defgroup    boards_hamilton
+ * @ingroup     boards
+ * @brief       Support for the Hamilton board.
+ * @{
+ *
+ * @file
+ * @brief       Board specific definitions for the Hamilton
+ *              board
```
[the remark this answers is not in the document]
Hyungsin: done

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright (C) 2018 UC Berkeley
```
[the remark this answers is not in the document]
Hyungsin: done 2016 is more correct

boards/hamilton/include/periph_conf.h
Hyungsin: done

boards/hamilton/include/periph_conf.h
```diff
@@ -0,0 +1,226 @@
...
+#define ADC_0_PRESCALER                    ADC_CTRLB_PRESCALER_DIV512
+
+#define ADC_0_NEG_INPUT                    ADC_INPUTCTRL_MUXNEG_GND
+#define ADC_0_GAIN_FACTOR_DEFAULT          ADC_INPUTCTRL_GAIN_1X
+#define ADC_0_REF_DEFAULT                  ADC_REFCTRL_REFSEL_INT1V
+
+static const adc_conf_chan_t adc_channels[] = {
+    /* port, pin, muxpos */
+    {GPIO_PIN(PA, 6), ADC_INPUTCTRL_MUXPOS_PIN6},      /* EXT1, pin 3 */
+    {GPIO_PIN(PA, 7), ADC_INPUTCTRL_MUXPOS_PIN7},      /* EXT1, pin 4 */
```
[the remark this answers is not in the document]
Hyungsin: done

cpu/sam0_common/periph/cpuid.c
```diff
@@ -24,6 +24,7 @@
 #include <string.h>
 
 #include "periph/cpuid.h"
+#include "board.h"
```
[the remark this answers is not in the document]
Hyungsin: It is related because fb_eui64 is defined in board.h.

sys/luid/luid.c
```diff
@@ -59,4 +64,5 @@ void luid_base(void *buf, size_t len)
         out[i % len] ^= cid[i];
     }
 #endif
+#endif
```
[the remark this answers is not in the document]
Hyungsin: done

Hyungsin:
@MichelRottleuthner, I reflected most comments.
Regarding the test application, I do think that providing a complete test application is good for users to experience full features of the board quickly, without tedious trials and errors. This is to mitigate any misunderstanding and hasty frustration.

The hamilton is designed for low-power sensing and wireless reporting, which needs to be shown in this test application. I'm thinking working on this application further to show that hamilton periodically wakes up, senses all information, and broadcasts a packet including the information. The idle current should be ~6uA.

If redundancy is only thing that matters, please allow me to maintain this application :)

boards/hamilton/include/periph_conf.h
```diff
@@ -0,0 +1,226 @@
...
+ * @name PWM configuration
+ * @{
+ */
+#define PWM_0_EN            1
+#define PWM_1_EN            1
+#define PWM_NUMOF           (PWM_0_EN + PWM_1_EN)
+#define PWM_MAX_CHANNELS    2
+/* for compatibility with test application */
+#define PWM_0_CHANNELS      PWM_MAX_CHANNELS
+#define PWM_1_CHANNELS      PWM_MAX_CHANNELS
```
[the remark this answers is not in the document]
Hyungsin: It is true, but no harm to have this just in case, right?

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,151 @@
...
+ * @{
+ */
+extern const uint64_t* const fb_sentinel;
+extern const uint64_t* const fb_flashed_time;
+extern const uint8_t*  const fb_eui64;
+extern const uint16_t* const fb_device_id;
+extern const uint64_t* const fb_designator;
+extern const uint8_t*  const fb_aes128_key;
+extern const uint8_t*  const fb_25519_pub;
+extern const uint8_t*  const fb_25519_priv;
```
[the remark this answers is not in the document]
Hyungsin: will do

boards/hamilton/Makefile.dep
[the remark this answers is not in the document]
MichelRottleuthner: Yes it is convenient, but it should be upon the application to decide whether to use auto_init or not. Counter example: run `grep -r --include "Makefile*" "USEMODULE += auto_init_gnrc_netif"` from RIOT root dir and you will see that hamilton is the only board that adds this dependency, all other occurrences are from tests or examples (i.e. applications).

boards/hamilton/include/periph_conf.h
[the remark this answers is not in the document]
MichelRottleuthner: no harm, but we shouldn't add unnecessary code ;) I think you are safe to remove it here.  The use of this define was removed [some time ago](https://github.com/RIOT-OS/RIOT/commit/4b4cc5a6b3a3889ca7dd4a8df21caba462370aad#diff-b2daa6756b90bf7ebdcc56d8ac1b66a7).

cpu/sam0_common/periph/cpuid.c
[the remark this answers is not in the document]
MichelRottleuthner: sorry, I actually meant the same line in tmp006.h

drivers/include/tmp006.h
```diff
@@ -80,6 +80,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include "periph/i2c.h"
+#include "board.h"
```
MichelRottleuthner: This is used for `TMP006_USE_LOW_POWER`, right? It looks a bit misplaced to me, that the driver of a temperature sensor needs to know something about the board it is used on, except for it's parameters. Is it really needed to define this in board.h? I mean, the board will work without this setting and so will the driver -> this looks application-specific to me. If there really is a good reason to set `TMP006_USE_LOW_POWER` and `TMP006_USE_RAW_VALUES` in board.h then these should be handled like the other `TMP006_PARAM*` values via tmp006_params.h. But then this change should also be introduced by a seperate PR I think.

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * @name FXOS8700 configuration
+ * @{
+ */
+#define FXOS8700_PARAM_I2C            I2C_DEV(0)
+#define FXOS8700_PARAM_ADDR           (0x1E)
+#define FXOS8700_PARAM_ACC_RANGE      FXOS8700_REG_XYZ_DATA_CFG_FS__8G
+#define FXOS8700_PARAM_RENEW_INTERVAL (1000000ul)
+#define FXOS8700_USE_ACC_RAW_VALUES   (0)
```
MichelRottleuthner: Is this really needed here? `FXOS8700_USE_ACC_RAW_VALUES` isn't handled by fxos8700_params.h so overwriting this by the application will not be possible. See my other comment regarding tmp006.h

emmanuelsearch:
@Hyungsin @MichelRottleuthner the plan was to test this week and integrate in CI in Berlin with @kaspar030 
Stay tuned, we target Friday.

MichelRottleuthner:
> Regarding the test application, I do think that providing a complete test application is good for users to experience full features of the board quickly, without tedious trials and errors. This is to mitigate any misunderstanding and hasty frustration.

I agree on that and I like that you kept an eye on usability/simplicity, but at the moment the application doesn't include something specific to this board, does it? Have you tried the saul test appllication? It should do almost the same as this test app with less code.

>I'm thinking working on this application further to show that hamilton periodically wakes up, senses all information, and broadcasts a packet including the information. The idle current should be ~6uA.

Then it may be better to issue a PR once the application has this implemented ;)

@emmanuelsearch thanks!

boards/hamilton/include/periph_conf.h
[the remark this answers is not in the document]
Hyungsin: OK. I resolved the issue. But FYI, samr21-xpro driver also has this part.

cpu/sam0_common/periph/cpuid.c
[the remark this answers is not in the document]
Hyungsin: Alright :)

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,158 @@
...
+ * @{
+ */
+#define FXOS8700_PARAM_I2C                I2C_DEV(0)
+#define FXOS8700_PARAM_ADDR               (0x1E)
+#ifndef FXOS8700_PARAM_ACC_RANGE
+#define FXOS8700_PARAM_ACC_RANGE          FXOS8700_REG_XYZ_DATA_CFG_FS__8G
+#endif
+#ifndef FXOS8700_PARAM_RENEW_INTERVAL
+#define FXOS8700_PARAM_RENEW_INTERVAL     (1000000ul)
+#endif
```
Hyungsin: This part is related to #9386.

Hyungsin:
@MichelRottleuthner, I agree. So I removed the test application and tmp006.h from this PR.

MichelRottleuthner:
I'm fine with the changes now, please rebase and squash once all dependant PRs are merged

Hyungsin:
added PIR configuration since #7823 is about to be merged.

aabadie:
@Hyungsin, is it possible to remove parts related to I2C from this PR ? This way, it won't be blocked anymore by the I2C refactoring embargo.
Is #9386 the last opened PR required by this PR ?

tcschmidt:
@emmanuelsearch did you manage to do the testing? If not, consider handing the device over this Friday so that I can transfer it to @MichelRottleuthner

emmanuelsearch:
@tcschmidt yes we are in the process of testing (we only minor comments @kaspar030 can perhaps give more details?)
Basically we are in a position where we could add this board to the CI (once @aabadie concerns with I2C are cleared).

tcschmidt:
@emmanuelsearch @kaspar030 fine- would be good to document test results in this PR for future reference.

Hyungsin:
@aabadie, sure. Just for a double check, you mean I2C_configuration in "periph_conf.h", not "board.h", right?

Hyungsin:
@aabadie updated.
It is also fine to merge this PR before #9386. Before merging #9386, HDC1000_PARAM_RENEW_INTERVAL will just be a redundant parameter which does not impact any operation.

boards/hamilton/Makefile.features
```diff
@@ -0,0 +1,16 @@
+# Put defined MCU peripherals here (in alphabetical order)
+FEATURES_PROVIDED += periph_adc
+FEATURES_PROVIDED += periph_cpuid
```
aabadie: This feature is provided by the CPU, you can just add `-include $(RIOTCPU)/samd21/Makefile.features` at the end of this file.
Same for `periph_pm` and `cpp`

boards/hamilton/board.c
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @}
+ */
+
+#include <stdio.h>
+
+#include "board.h"
+#include "periph/gpio.h"
+
+const uint64_t* const fb_sentinel     = ((const uint64_t* const)0x3fc00);
```
aabadie: What are these variables for ?

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,166 @@
...
+#define AT86RF2XX_PARAM_INT      GPIO_PIN(PB, 0)
+#define AT86RF2XX_PARAM_SLEEP    GPIO_PIN(PA, 20)
+#define AT86RF2XX_PARAM_RESET    GPIO_PIN(PB, 15)
+/** @} */
+
+/**
+ * @name    LED pin definitions and handlers
+ * @{
+ */
+#define LED_PIN                  GPIO_PIN(0, 19)
```
aabadie: Use `LED0_PIN` to be consistent with other boards. And this way an application that is using the `LEDx_PIN` can be used on different platforms without any change.

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * @name FXOS8700 configuration
+ * Note that another fxos8700 operation option, FXOS8700_USE_ACC_RAW_VALUES,
+ * need to be set according to the application purposes
+ * @{
+ */
+#define FXOS8700_PARAM_I2C                I2C_DEV(0)
+#define FXOS8700_PARAM_ADDR               (0x1E)
+#ifndef FXOS8700_PARAM_ACC_RANGE
```
aabadie: I don't think you need this `#ifndef` but maybe you have a good reason ?

cpu/sam0_common/periph/cpuid.c
```diff
@@ -33,6 +34,12 @@
 
 void cpuid_get(void *id)
 {
+#ifdef HAS_FACTORY_BLCOK
```
aabadie: Typo, it should be `HAS_FACTORY_BLOCK` instead

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,166 @@
...
+ */
+extern const uint64_t* const fb_sentinel;
+extern const uint64_t* const fb_flashed_time;
+extern const uint8_t*  const fb_eui64;
+extern const uint16_t* const fb_device_id;
+extern const uint64_t* const fb_designator;
+extern const uint8_t*  const fb_aes128_key;
+extern const uint8_t*  const fb_25519_pub;
+extern const uint8_t*  const fb_25519_priv;
+#define FB_SENTINEL_VALUE 0x27c83f60f6b6e7c8
```
aabadie: What is this value for ?

cpu/sam0_common/periph/cpuid.c
aabadie: I'm not sure including a board specific header in a cpu file is a good idea.

boards/hamilton/include/board.h
Hyungsin: FXOS8700_PARAM_ACC_RANGE can be changed depending on applications. I personally don't use this default parameter for my application.

cpu/sam0_common/periph/cpuid.c
Hyungsin: I have no preference so any idea to get around this issue is welcome :)

boards/hamilton/Makefile.features
Hyungsin: done

boards/hamilton/include/board.h
Hyungsin: done

cpu/sam0_common/periph/cpuid.c
Hyungsin: done

boards/hamilton/include/board.h
aabadie: Ok, so you should remove the range related defines since this is not board specific.

Hyungsin: done

boards/hamilton/Makefile.features
```diff
@@ -0,0 +1,13 @@
...
+FEATURES_PROVIDED += periph_pwm
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
+FEATURES_PROVIDED += periph_spi
+FEATURES_PROVIDED += periph_timer
+
+# The board MPU family (used for grouping by the CI system)
+FEATURES_MCU_GROUP = cortex_m0_2
+
+include $(RIOTCPU)/samd21/Makefile.features
```
aabadie: you need a `-` at the beginning of the line (I don't remember the exact reason, Makefile experts can explain ;) )

Hyungsin: done

boards/hamilton/include/board.h
```diff
@@ -0,0 +1,157 @@
...
+#define AT86RF2XX_PARAM_RESET    GPIO_PIN(PB, 15)
+/** @} */
+
+/**
+ * @name    LED pin definitions and handlers
+ * @{
+ */
+#define LED0_PIN                  GPIO_PIN(0, 19)
+
+#define LED_PORT                  PORT->Group[0]
```
aabadie: Should be LED0_PORT (you forgot this one)

boards/hamilton/include/board.h
Hyungsin: @immesys, please answer to this question.

boards/hamilton/board.c
Hyungsin: @immesys, please answer to this question

boards/hamilton/include/board.h
Hyungsin: done

boards/hamilton/dist/flash.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
```
aabadie: I think this file is probably not needed: you can reuse the script in `dist/tools/jlink` by setting JLink specific variable in including `jlink.inc.mk` in the Makefile.include of that board.

aabadie: This comment also stands for reset.sh and debug.sh

Hyungsin: done

Hyungsin:
@aabadie, I think I resolved all issues. Please check again :)
1. Redundant sh files are all gone now.
2. Factory-generated parameters are removed since these are used only at the HamiltonIoT factory.
3. Accordingly, cpuid.c and luid.c are not changed, so zero issues regarding these.

Hyungsin:
rebased

kaspar030:
> but I guess others already tested it

iI did a brief hello-world test about two weeks ago, that worked fine (only make term is broken, see below). I'll re-check current PR state tomorrow.

For integration of hamilton boards into the CI, the hardware is in place, but "make term" needs work. By default, it uses segger_rtt for stdio, which in turn needs the corresponding JLink tool to be started in the background. It would be nice if that would happen automagically, but that should not be part of this PR.

MichelRottleuthner:
Murdock still lists some errors related to RTC and RTT. Makefile.features lists both as supported but `periph_conf.h` doesn't provide configuration. While thinking about it, IMO those configs should actually go to cpu code, but that's a thing for later discussion/PRs.

Hyungsin:
@MichelRottleuthner, good finding! I resolved the issue, but it still has some errors due to PIR test application (double definition). I opened #9488 to resolve the issue, which includes a very simple fix in Makefile of the PIR test application.

Hyungsin:
All checks have passed

aabadie:
Let's wait for @kaspar030 final decision.

MichelRottleuthner:
ping @kaspar030

kaspar030:
Soon :) (Can't access the board right now, probably Tue or Wed.)

Hyungsin:
@kaspar030, did you have a chance to test it?

kaspar030:
> @kaspar030, did you have a chance to test it?

Somehow I can't get output via Segger RTT anymore.

kaspar030:
Stepping through main (e.g., of hello-world) works like a charm, but there's no output.

kaspar030:
@Hyungsin did anything change since end of June that could cause this?

Hyungsin:
@kaspar030 Not really, hmm, I can test again :)

Hyungsin:
@kaspar030, ping please.
Also, RIOT has been updated, which causes Murdock errors. What did cause the errors?

kYc0o:
I just got one at hand and gets stuck in `make term`:

```console
Script processing completed.

JLinkRTTClient 
###RTT Client: ************************************************************ 
###RTT Client: *               SEGGER Microcontroller GmbH                * 
###RTT Client: *   Solutions for real time microcontroller applications   * 
###RTT Client: ************************************************************ 
###RTT Client: *                                                          * 
###RTT Client: *       (c) 2012 - 2016  SEGGER Microcontroller GmbH       * 
###RTT Client: *                                                          * 
###RTT Client: *     www.segger.com     Support: support@segger.com       * 
###RTT Client: *                                                          * 
###RTT Client: ************************************************************ 
###RTT Client: *                                                          * 
###RTT Client: * SEGGER J-Link RTT Client   Compiled May 28 2018 17:00:15 * 
###RTT Client: *                                                          * 
###RTT Client: ************************************************************ 

###RTT Client: -----------------------------------------------
###RTT Client: Connecting to J-Link RTT Server via localhost:19021 .......................................................................................................
```

and so on... the `...` never end.

Is there any trick to make the segger_rtt working?

kaspar030:
> Is there any trick to make the segger_rtt working?

You'll have to start the jlink server in another shell (on my system, it is called ```JLinkRemoteServerCLExe```).

kYc0o:
Ok apparently one needs to connect first with `JLinkExe -device atsamr21e18a -speed 4000 -if SWD` before doing a `make term`.

@Hyungsin can you add this as an OS X variant?

Otherwise tests are successful.

kYc0o:
> Ok apparently one needs to connect first with JLinkExe -device atsamr21e18a -speed 4000 -if SWD before doing a make term.

A bit more testing reveals that this is also true for Linux machines. I'll try to provide a diff with the required commands to make this smoothly integrated into single commands.

boards/hamilton/Makefile.include
```diff
@@ -0,0 +1,13 @@
+# define the cpu used by Hamilton
+export CPU = samd21
+export CPU_MODEL = samr21e18a
+
+# debugger config
+export JLINK_DEVICE := atsamr21e18a
+export TERMPROG := JLinkRTTClient
```
kYc0o: RIOT already uses pyterm as a client for RTT. Please change to

```diff
diff --git a/boards/hamilton/Makefile.include b/boards/hamilton/Makefile.include
index f301c2731..432c79321 100644
--- a/boards/hamilton/Makefile.include
+++ b/boards/hamilton/Makefile.include
@@ -4,7 +4,8 @@ export CPU_MODEL = samr21e18a
 
 # debugger config
 export JLINK_DEVICE := atsamr21e18a
-export TERMPROG := JLinkRTTClient
+TERMPROG = $(RIOTTOOLS)/jlink/jlink.sh
+TERMFLAGS = term_rtt
 export OBJDUMPFLAGS += --disassemble --source --disassembler-options=force-thumb
 export OFLAGS := -O binary --gap-fill 0xff
 ```

Optionally, you might add the usage of `JLinkRTTClient` as a non-default setting to use it separately.

**Comment by kYc0o** [comment]
By the way, all shell/uart related tests don't succeed. I suspect is related to uart not being used for getting the commands like normally is, maybe another patch is needed for those tests using RTT.

**Comment by Hyungsin** [comment]
@kYc0o, Thank you for working on this PR!

I can change the PR as you suggested.

Hamilton does not have UART so it does not support any UART-related stuff and does not have UART_DEV. This configuration was OK with Murdock before but some latest changes to RIOT made Murdock unhappy...

**Comment by kYc0o** [comment]
> Hamilton does not have UART so it does not support any UART-related stuff and does not have UART_DEV. This configuration was OK with Murdock before but some latest changes to RIOT made Murdock unhappy...

I suggest to blacklist the board, it will reduce the Murdock builds and anyways we cannot test them, so they will be automatically skipped by the CI HIL tests.

However, we need to find a way to perform the tests using another strategy, maybe a combination of openocd + JLink tools.

**Comment by kYc0o** [comment]
> Hamilton does not have UART so it does not support any UART-related stuff and does not have UART_DEV. This configuration was OK with Murdock before but some latest changes to RIOT made Murdock unhappy...

I've found the problem. 9bd23636432da7ba5c2ebb97586941debcf91fcd changed the name from rtt_stdio to stdio_rtt.

Please rebase on master and change that on your Makefile.include. Though, it might be misplaced... But that's another discussion.

I have another 3 PRs before this can be merged.

**Comment by kYc0o** [comment]
This PR now depends on #9896 and #9916.

The patch needed once that PR is merged and this PR is rebased is as following:

```diff
diff --git a/boards/hamilton/Makefile.include b/boards/hamilton/Makefile.include
index f301c2731..b878fff70 100644
--- a/boards/hamilton/Makefile.include
+++ b/boards/hamilton/Makefile.include
@@ -4,10 +4,13 @@ export CPU_MODEL = samr21e18a
 
 # debugger config
 export JLINK_DEVICE := atsamr21e18a
-export TERMPROG := JLinkRTTClient
 export OBJDUMPFLAGS += --disassemble --source --disassembler-options=force-thumb
 export OFLAGS := -O binary --gap-fill 0xff
 
-USEMODULE += rtt_stdio
+# Configure terminal, hamilton doesn't provide any UART, thus use RTT
+TERMPROG = $(RIOTTOOLS)/jlink/jlink.sh
+TERMFLAGS = term_rtt
+
+USEMODULE += stdio_rtt
```

With this everything should work.

kYc0o:
@Hyungsin ping.

Hyungsin:
@kYc0o, I think I resolved all things! Thanks!

Hyungsin:
@kYc0o, Yes I pushed!

kYc0o:
Ok I just retriggered Murdock to be sure everything is OK. Still #9916 needs to be merged before.

Hyungsin:
@kYc0o, Murdock seems happy except the pending PR issue. Look forward to #9916 merged!

Hyungsin:
rebased

kYc0o:
While recompiling I got the following error:

```console
snake:RIOT facosta$ BOARD=hamilton make -C examples/default/ flash term
make: Entering directory '/Users/facosta/git/RIOT-OS/RIOT/examples/default'
There are unsatisfied feature requirements: periph_gpio_irq


EXPECT ERRORS!


Building application "default" for "hamilton" with MCU "samd21".
```

This is maybe related to #9992. Can you verify?

Hyungsin:
@kYc0o, resolved. Please check again!

kYc0o:
For some reason some tests cannot be executed. For instance tests/shell doesn't answer anymore, though it seems to flash it correctly as the prompt is displayed, but when I type some command I have not answer. @Hyungsin can you check if you experience this behaviour too?

emmanuelsearch:
@Hyungsin ping

Hyungsin:
@emmanuelsearch @kYc0o, I haven't tested shell with hamilton and don't know about the implementation. Does the shell app use UART?

kYc0o:
I was investigating a bit but I couldn't find the problem.

Basically, if you compile and flash examples/default or examples/gnrc_networking everything works as expected. Some tests pass too, but when you try `BOARD=hamilton make -C tests/shell flash test` it fails. Actually, even `make term` doesn't seem to respond. The test uses the shell which is also used in the examples, so it should work, but for some reason it doesn't. It depends on stdio which for most of the boards is stdio_uart, but as we know for hamilton is stdio_rtt. I don't know if that has something to do with the failure though...

Hyungsin:
@kYc0o, Previously, we executed "make debug-server", "make debug", "make term" separately. But you tried to merge all of these. I'm wondering if this change causes some problems in the shell test.

tcschmidt:
@emmanuelsearch Where is the hardware now? Who can continue testing? Maybe @danpetry ?

emmanuelsearch:
@tcschmidt the boards are at FU Berlin if @danpetry wants to take over (which would be great!)

danpetry:
Sure, I'll take over.

tcschmidt:
@MichelRottleuthner @aabadie add labels characterizing your parts of the review here?

danpetry:
I've just performed the following test:
1. Flash examples/default
2. Sanity check operation of all SAUL-supported sensors, with the exception of the PIR because the board doesn't have one on it. By "sanity check" I mean observe that the readings are realistic, and change in the expected way (e.g. by tilting the board and seeing the accelerometer change)

Test rationale was that I'm not testing the sensor drivers here (hence no accurate readings needed) and that initial support is present (RIOT runs and operates generally as expected), and I agree with the rationale that smaller bugs can be fixed later on. ACK on testing from me, any objections to this testing strategy?

danpetry:
@kYc0o do you have any unadressed points you wish to be addressed ?

kaspar030:
@danpetry can you try if you can get the rtt.py example of https://github.com/square/pylink to interact with the hamilton?

danpetry:
```
danielpetry@inetm04:~/Sandbox/pylink/examples$ python rtt.py atsamr21e18a

connecting to JLink...
connecting to atsamr21e18a...
Traceback (most recent call last):
  File "rtt.py", line 142, in <module>
    sys.exit(main(sys.argv[1]))
  File "rtt.py", line 117, in main
    jlink.connect(target_device)
  File "build/bdist.linux-x86_64/egg/pylink/jlink.py", line 141, in wrapper
  File "build/bdist.linux-x86_64/egg/pylink/jlink.py", line 941, in connect
pylink.errors.JLinkException: Unsupported device was connected to.
```

kaspar030:
Could be [this](https://github.com/square/pylink/issues/38) issue.

Try:
```
# Option A: Copy the library to your libraries directory.
$ cp libjlinkarm.so /usr/local/lib/

# Option B: Add SEGGER's J-Link library path to your libraries path.
$ export LD_LIBRARY_PATH=/path/to/SEGGER/JLink:$LD_LIBRARY_PATH
```
(probably the latter. On my system, the path is ```/opt/SEGGER/JLink```)

cladmi:
Just a note for further improvements, while testing on my machine, I need a delay between the server creation and connecting to it during `term` here:

https://github.com/RIOT-OS/RIOT/blob/7e77ac7ace75ca9643250e0c747a5e65074d3fef/dist/tools/jlink/jlink.sh#L241-L249

It is unrelated to the board port but was required for me to run any tests. I can do a PR for it.

danpetry:
Ok, dropping the last `a` from the device name made it work.

With the rtt.py example, I get the same behaviour as with `make flash term`. I.e., examples/default can be communicated to, but tests/shell cannot. (After doing `make flash` in each respective directory. The jlinkexe program reports that for both of these applications, flashing is successful

kaspar030:
> In principle, then, would you prefer to fix this before merging?

Well, ```tests/shell``` is one of our most basic tests.

Just to confirm - when running ```"make debug-server", "make debug", "make term" separately```, does the application work? If yes, we're having some problem with the test target. In that case, merging would be alright.

 If not, it might be something deeper.

danpetry:
Same results, examples/default works and tests/shell does not.

Hyungsin:
@kaspar030, sorry, my schedule is crazy these days... I might have some time to work on this in two weeks. @danpetry, so.. although you tried to do debug-server, debug, term, all separately, did it not work?

danpetry:
@Hyungsin yes, correct.

Hyungsin:
@danpetry, @kaspar030, I just came back.. sorry! For the shell application, commenting out "DISABLE_MODULE += auto_init" in Makefile terminates the problem. Now shell seems working!

Hyungsin:
Without #10849 merged, you should use two separate terminals.
Type make debug at one terminal and type make term at the other one.
I see the results as below:

Hyungsin:
Rebased

danpetry:
Hi, sorry for the delay! I'll be responsive now, we could try and get this in for the upcoming release perhaps?
I'm getting sporadic behaviour. I have successfully run `tests/shell`, but then immediately afterwards I flashed and tried to communicate `default/examples` and the session crashed at the point at which I tried to send commands to the node. By unplugging and plugging back in the segger, I could get it working again, but when flashing a second application it also crashed. This behaviour is also observed when I flash the applications the other way round. 
I ran a debug session in one pane and a RIOT term in another, as described above. Output below
image
@Hyungsin are you able to recreate, or have any immediate insights into this behaviour?
The auto_init disabling has been here since the `tests/` folder was created, see [here](https://github.com/RIOT-OS/RIOT/blob/d46c7e00d462d7f5f66cf1d59aa156fe72c2e968/tests/shell/Makefile). @Kijewski or @OlegHahm do you have any recollection of why this was put in?

Hyungsin:
@danpetry, I'm not really sure about this. In my experience, JLink has connection is not always robust... Does turning power off and on before flashing a new app help?

danpetry:
Yes, power cycling of the segger helps (I did this by unplugging and plugging in the usb) but I think we should not require power cycling to re-flash. I just checked with @kaspar030 on his computer/toolchain and he's having similar problems to me. IMO we should solve this so that the support is solid. I'll have a closer look at this and try and isolate the problem. @Hyungsin would you be able to post your Jlink versions and platform (looks like ubuntu 16.04?)

Re the removal of auto_init, initially this was to isolate the test code better. As you point out, now we need it for RTT so should put it back in. It's unlikely to have any adverse effects but let's test on a few other selected boards just in case.

kaspar030:
> Re the removal of auto_init, initially this was to isolate the test code better. As you point out, now we need it for RTT so should put it back in. It's unlikely to have any adverse effects but let's test on a few other selected boards just in case.

I'm just realizing that this would be fine for tests/shell, but may cause trouble in other tests.

Hyungsin:
@danpetry, my Ubuntu version is 16.04, JLink Firmware is J-Link V10, and Hardware is V10.10.

Hyungsin:
@danpetry, how is it going? Can we include this PR in the upcoming release?

tcschmidt:
There is still a change request by @kYc0o - has this been resolved?

Hyungsin:
@tcschmidt, yes 👍

tcschmidt:
> 
> 
> @tcschmidt, yes 👍

@kYc0o can you confirm?
@danpetry can this be merged then?

danpetry:
This is still not working over this side - we've tested on a number of dev machines to make sure it's not something to do with the setup of mine. @Hyungsin I've got exactly the same OS and Jlink versions as you. Looking into this further.

cladmi:
I added a PR that could at least, show when flashing fails with the board instead on returning 0.
https://github.com/RIOT-OS/RIOT/pull/11303

cladmi:
> > Re the removal of auto_init, initially this was to isolate the test code better. As you point out, now we need it for RTT so should put it back in. It's unlikely to have any adverse effects but let's test on a few other selected boards just in case.
> 
> I'm just realizing that this would be fine for tests/shell, but may cause trouble in other tests.

I am thinking about something as I currently have a similar issue that when `auto_init` is disabled I cannot rely on it to have something started.
The issue is that `auto_init` cannot be enabled by default as one test may not want one init function to be called.
So more the fact that "auto_init" enables all the available "auto_init" functions by default.

It would be a good solution to just disable the auto inclusion of all init functions but still use the requested ones. It would require some refactoring  but I think it would still be beneficial when using `xfa` so could still make sense to do it now.
By changing the `#ifdef MODULE_MODULE_NAME` to `#ifdef MODULE_AUTO_INIT_MODULE_NAME`  or `#ifdef defined(MODULE_AUTO_INIT_MODULE_NAME) && $(MODULE_MODULE_NAME)` and doing the requested definitions in the dependencies handling, `auto_init` could be split between `init` which calls the base function and `auto_init` that enables everything by default.

danpetry:
I've found that the following sequence of operations works reliably, for `examples/default`, `tests/shell`, and `examples/hello-world`.

w1: `BOARD=hamilton make flash debug`
w2: `BOARD=hamilton make term`                       (can get serial output from w3 at this point) 
w1: `(gdb) c`                                        (can give serial input in w3 at this point)

If the second and third steps are swapped, no operation is possible, which is why it wasn't working for me before. I.e. pyterm doesn't seem to connect correctly when the program is actually running. @Hyungsin is this the same on your side?

@cladmi has raised the point that this doesn't allow running automated tests. i.e. `make test` doesn't work. What's the usual requirement for a new board concerning `make test`? Is it acceptable to document this and raise an issue? I'm conscious also that this would probably be an extra requirement for @Hyungsin that he wasn't aware of, so I'm happy to help out if we decide it's necessary.

Hyungsin:
@danpetry, @cladmi, alright. Please figure out where we should go.

danpetry:
Here's my assessment of the issues and a proposed course of action.

*auto_init/xtimer*
Stdio_rtt has a dependency on xtimer being initialised. It's meant to act the same way as stdio_uart - it shares the API functions `stdio_read`, `stdio_write` and `stdio_init`. However, one difference is that stdio_rtt is dependent on xtimer, whereas stdio_uart is not. And xtimer is initialised centrally, by auto_init, so it would break current paradigms to initialise it in stdio_rtt. But we've currently got this "all-or-nothing" approach to initializing modules in auto_init, which effectively means that to make RTT work in current paradigms we have to initialise everything. Ok usually, but not for tests.
Personally I think if we want to make stdio_rtt a transparent swap-in for stdio_uart, we should configure an RTT timer in the board config code and use this instead of xtimer. Then we wouldn't need auto_init.

*make test*
Checking the Ruuvitag and Thingy52, it seems they have the same treatment - i.e. stdio is provided over RTT by default. So, no `make test` available.

*Proposed way forward*
For `make test`, I suggest we stick with current paradigms and merge as-is, document the flashing method (obviously), and raise an issue describing the situation. For the xtimer dependency issue, currently we don't have an RTT stdio which imitates the uart stdio transparently, and so it shouldn't share a test. I would propose for now a shell_rtt test with xtimer_init() in the main() function, and raise an issue to merge this into the other shell test once the dependency on xtimer is removed.

danpetry:
Comment above updated after some offline discussion with @miri64 . ACK from me and I'll wait before merging for a short while as stated.

Hyungsin:
@danpetry, great! Then should I restore DISABLE=auto_init for the shell test?

danpetry:
@Hyungsin yes, please restore the shell test makefile back to the way it was, squash and rebase. I don't see any blocking comments above so I'll re-test and merge once you've done this :)

Hyungsin:
@danpetry, done!

danpetry:
Re-tested and merged. Thanks for your patience and your contribution! :100:

danpetry:
https://github.com/RIOT-OS/RIOT/pull/11341 adds instructions on RTT shell access for all boards that use it

danpetry:
https://github.com/RIOT-OS/RIOT/issues/11344 and https://github.com/RIOT-OS/RIOT/issues/11343 describe the RTT-related issues with the shell

danpetry:
As described above, these issues were pre-existing and so being raised and dealt with outside the scope of this PR.

PR #9015, opened 2018-04-24
pkg/pkg.mk: Do not sign when applying patches

cladmi:
Disable signing every applied patches for user that have gpgsign by default.

### Issues/PRs references

Build system asking for gpg key passphrase when you have in your configuration
```
[commit]
    gpgsign = True
```


bergzand:
Works for me, building packages with tests doesn't ask me for the passphrase anymore

cladmi:
Further improvement could be to not use the user git configuration but I did not manage to find the way of doing it.

PR #9017, opened 2018-04-25
core/clist: make clist_foreach() return break-causing node

kaspar030:
### Contribution description

This PR makes clist_foreach(), which "breaks" out of list traversal if the supplied helper function returns non-zero, return the corresponding node instead of void.

This makes it easier to use clist_foreach() to search for nodes.

This should not break any users, as just the return value is added. Also, code size is unchanged on samr21-xpro.


cladmi:
I like this one and yeah would also like to have the test updated.

kaspar030:
> Please also update the corresponding unittest function

done!

cladmi:
@aabadie do you ACK too ? if yes I let you merge after murdock is happy.

PR #9011, opened 2018-04-24
makefile.iotlab.single: add IoT-LAB testbed support for one node

cladmi:
Add support to do flash/reset/term on an IoT-LAB node.
It should also allow running test using 'testrunner'.

It handles both running from your computer or from the ssh frontend.

    makefile.iotlab.single: add IoT-LAB testbed support for one node
    
    Add support to do flash/reset/term on an IoT-LAB node.
    It also allow running test using 'testrunner'.
    
    Configuration variables are:
    
    * `IOTLAB_NODE` which should be set to your node url
      * The full url including site to use from your computer `m3-1.grenoble.iot-lab.info`
      * The short url when used on the IoT-LAB frontend `m3-1`
    * `IOTLAB_EXP_ID` for your experiment id for flash and reset.
      By default it tries to use your currently running experiment if you have only one
    * `IOTLAB_USER`: is read from `${HOME}/.iotlabrc` as saved by `iotlab-auth`
      * It is expected to have run `iotlab-auth` beforehand.

Usage is 

    make IOTLAB_NODE=m3-1.iot-lab.info BOARD=iotlab-m3 all flash term

Also, it is possible to detect automatically a first node to use from available ones:

    make IOTLAB_NODE=auto-ssh BOARD=iotlab-m3 all flash term

TODO:

* [x] Verify it's working on the frontend (not tested yet)
* [x] Run the release tests with it to find issues
* [x] Reword first commit message

### Issues/PRs references

Allow running 'make test' using IoT-LAB nodes.

PR #9018, opened 2018-04-25
dist/testbed-support: use new cli-tools command names

aabadie:
### Contribution description

While looking at #9011, I realized that `Makefile.iotlab` in `dist/testbed-support` was still using the old iotlabcli command names. The new tools are now called `iotlab-xx` and are deployed on all ssh frontends of iotlab.

By the way, if someone is still using them locally from his machine, they should be updated using pip:
```
pip install iotlabcli --upgrade
```

This PR is just replacing the old `xx-cli` command names with the new `iotlab-xx` (and avoids the deprecation warning displayed when using the old names)

This PR also removes `iotlab-auth` from the .PHONY list, since it seems useless.

### Issues/PRs references

None

### DIFF
--- a/dist/testbed-support/Makefile.iotlab
+++ b/dist/testbed-support/Makefile.iotlab
@@ -1,14 +1,14 @@
-.PHONY: iotlab-auth iotlab-exp iotlab-flash iotlab-reset iotlab-term iotlab-check-exp
+.PHONY: iotlab-exp iotlab-flash iotlab-reset iotlab-term iotlab-check-exp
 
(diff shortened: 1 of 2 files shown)


cladmi:
Code change looks ok, did you re-tested it ?

aabadie:
> did you re-tested it ?

not yet ;)

aabadie:
> did you re-tested it ?

Now I did and it works, without the deprecation warnings. I just pushed a change updating a leftover in the README (directly squashed).

PR #9019, opened 2018-04-25
tests/pkg_cayenne_lpp: fix inconsistent buffer values across architectures (native, boards)

aabadie:
### Contribution description

This PR is an attempt to have a more robust output check with the pkg_cayenne-lpp test application.
I also don't think it's the goal of the test target to verify that the generated payload is valid.
But PRs on the [upstream project](https://github.com/aabadie/cayenne-lpp) are very welcome ;)

It should be useful with #9007.

I tested it on native and nucleo-l073, the ouput are sligthly different but the test target passes.

### Issues/PRs references

#9007

### DIFF
--- a/tests/pkg_cayenne-lpp/main.c
+++ b/tests/pkg_cayenne-lpp/main.c
@@ -19,41 +19,90 @@
  */
 
 #include <stdio.h>
+#include <string.h>
 
 #include "cayenne_lpp.h"
 
+#define TEST_BUFFER1 { 0x03, 0x67, 0x01, 0x10, 0x05, 0x67, 0x00, 0xff }
+#ifdef BOARD_NATIVE
+#define TEST_BUFFER2 { 0x01, 0x67, 0xFF, 0xD8, \
+                       0x06, 0x71, 0x04, 0xD1, 0xFB, 0x2F, 0x00, 0x00 }
+#else
+#define TEST_BUFFER2 { 0x01, 0x67, 0xFF, 0xD7, \
+                       0x06, 0x71, 0x04, 0xD2, 0xFB, 0x2E, 0x00, 0x00 }
+#endif
+#define TEST_BUFFER3 { 0x01, 0x88, 0x06, 0x76, \
+                       0x5E, 0xF2, 0x96, 0x0A, 0x00, 0x03, 0xE8 }
+
 static cayenne_lpp_t lpp;
 
-static void _print_buffer(cayenne_lpp_t *lpp)
+static const uint8_t test_buffer1[] = TEST_BUFFER1;
+static const uint8_t test_buffer2[] = TEST_BUFFER2;
+static const uint8_t test_buffer3[] = TEST_BUFFER3;
+
+static void _print_buffer(const uint8_t *buffer, size_t len, const char *msg)
 {
-    for (uint8_t i = 0; i < lpp->cursor; ++i) {
-        printf("%02X", lpp->buffer[i]);
+    printf("%s: ", msg);
+    for (uint8_t i = 0; i < len; i++) {
+        printf("%02X", buffer[i]);
     }
-    puts("");
 }
 
 int main(void)
 {
-    puts("Cayenne LPP test application");
+    puts("Cayenne LPP test application\n");
 
     /* generate payload like the one given as example at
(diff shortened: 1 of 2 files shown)


aabadie:
@cladmi, I know I'm disabling the verification of the generated payload, and I did that on purpose (see the initial comment of this PR)
`I also don't think it's the goal of the test target to verify that the generated payload is valid.` This verification should be part of the upstream library. But this is debatable.

cladmi:
How can you test the library with itself ? If the library does not work it could say ok when it's not good.

However if you would like to do the test on the node and not in python (as we tend to), you should add a "_compare_buffer(&lpp, expected_1)" and not do only a "_print_buffer".

aabadie:
> How can you test the library with itself ?

Adding unit tests upstream would help (this is something I have in mind). But it wouldn't solve the issue with rounding float values being different depending on the architecture. The upstream library is tested on CircleCI, i.e. on x86 arch.

At least, you pointed to origin of the problem (floating point precision) and adding the following (taken from [there](https://www.linuxtopia.org/online_books/an_introduction_to_gcc/gccintro_70.html):
```mk
ifeq (native, $(BOARD))
  CFLAGS += -msse2 -mfpmath=sse
endif
```
to the test application makefile gives consistent results between native and other boards. Maybe this should be added to the package makefile itself. Or do you see better solutions ?

cladmi:
Maybe even to native in general.

aabadie:
> Maybe even to native in general.

indeed, but I have no idea of the consequences of such a change :)

cladmi:
> indeed, but I have no idea of the consequences of such a change :)

Yes. What I mean by this is for the moment it's ok to have a special `native` case in the test.

Setting a special floating point if cayenne is included changes the behavior of all floats of the application so maybe not that good.

Asking if this floating point configuration could be made default to be more accurate to nodes float handling is another separate topic.

kaspar030:
> CFLAGS += -msse2 -mfpmath=sse

That won't work on ARM.

kaspar030:
> But it wouldn't solve the issue with rounding float values being different depending on the architecture.

If the library doesn't clearly define floating point behaviour (which would IMO make it way too complex), float values can always end up different when using CPU instructions directly. Better change the tests so only limited precision is tested.

aabadie:
> Better change the tests so only limited precision is tested.

I prefer keeping the values used in the test because they are the ones given by the [cayenne lpp documentation](https://mydevices.com/cayenne/docs_stage/lora/#lora-cayenne-low-power-payload) (this gives us a ground truth).
Since only a few values (3) in the output buffer are affected, I just adapted it with a regexp and added a comment. Tested on native, samr21-xpro and nucleo-l152

cladmi:
I still prefer an explicit `if`, because it really explains what is on native and what is on a board.
Whereas with the regex, it gives 4 valid values instead on 2, and does not say which one to expect.

The test could also be implemented in main.c and not in Python. But I could do that in a further PR.

aabadie:
> The test could also be implemented in main.c and not in Python.

I pushed a couple of commits that does that. The idea is to let the preprocessor decide which buffer is correct depending on the target. It's much cleaner.

aabadie:
@cladmi, updated with the expected value being displayed when each test fails.

tests/pkg_cayenne-lpp/main.c
```diff
@@ -19,41 +19,87 @@
...
+    puts("Test 1:");
+    puts("-------");
     cayenne_lpp_add_temperature(&lpp, 3, 27.2);
     cayenne_lpp_add_temperature(&lpp, 5, 25.5);
-    _print_buffer(&lpp);
+    _print_buffer(&lpp, "Result");
+    /* check buffer */
+    if (memcmp(lpp.buffer, test_buffer1, lpp.cursor) != 0) {
+        puts("");
+        _print_buffer(&lpp, "Expected");
```
cladmi: Nop, it prints the same one :p 

```
Result: 0167FFD7067104D2FB2E0000
Expected: 0167FFD7067104D2FB2E0000 FAILED
```

aabadie:
@cladmi, my last commit is crap, I'll rework it. Sorry

aabadie:
Done, thanks for reviewing and testing @cladmi !

PR #9020, opened 2018-04-25
pkg/tinycbor: Initial import

bergzand:
### Contribution description

Initial import of the [tinycbor](https://github.com/intel/tinycbor) package.

From an initial impression it is less memory hungry than the cn-cbor package.

Test is based on the cn-cbor test.

### Issues/PRs references

depends on #8793 due to a header naming conflict

### DIFF
--- a/pkg/tinycbor/Makefile
+++ b/pkg/tinycbor/Makefile
@@ -1,0 +1,11 @@
+PKG_NAME=tinycbor
+PKG_URL=https://github.com/intel/tinycbor
+PKG_VERSION=v0.5.1
+PKG_LICENSE=MIT
+
+.PHONY: all
+
+all: git-download
+	"$(MAKE)" -C $(PKG_BUILDDIR)/src -f $(CURDIR)/Makefile.tinycbor
+
+include $(RIOTBASE)/pkg/pkg.mk
--- a/tests/pkg_tinycbor/main.c
+++ b/tests/pkg_tinycbor/main.c
@@ -1,0 +1,104 @@
+/*
+ * Copyright (C) 2018 Freie Universität Berlin
+ * Copyright (C) 2018 Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     tests
+ * @{
+ *
+ * @file
+ * @brief       Tests tinycbor package
+ *
+ * @author      Koen Zandberg <koen@bergzand.net>
+ *
+ * @}
+ */
+
+#include <string.h>
+
+#include "embUnit.h"
+#include "embUnit/embUnit.h"
+#include "fmt.h"
+#include "cbor.h"
+
+const char *tests[] = {
+    "00",                  /* 0                       */
+    "01",                  /* 1                       */
+    "17",                  /* 23                      */
+    "1818",                /* 24                      */
+    "190100",              /* 256                     */
+    "1a00010000",          /* 65536                   */
+    "20",                  /* -1                      */
+    "37",                  /* -24                     */
+    "3818",                /* -25                     */
+    "390100",              /* -257                    */
+    "3a00010000",          /* -65537                  */
+    "4161",                /* h"a"                    */
+    "6161",                /* "a"                     */
+    "80",                  /* []                      */
+    "8100",                /* [0]                     */
+    "820102",              /* [1,2]                   */
+    "818100",              /* [[0]]                   */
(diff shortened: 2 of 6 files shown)


bergzand:
I've added some code to the makefile to effectively use `tinycbor/cbor.h" in the code. The sed rules are to adapt the current cbor code. I can create a patch for it instead if preferred.

With this the conflict between the current cbor module and this package is solved.

pkg/tinycbor/Makefile
```diff
@@ -0,0 +1,15 @@
+PKG_NAME=tinycbor
+PKG_URL=https://github.com/intel/tinycbor
+PKG_VERSION=v0.5.1
+PKG_LICENSE=MIT
+
+.PHONY: all
+
+all: git-download
+	sed -i 's/"cbor.h"/"tinycbor\/cbor.h"/' $(PKG_BUILDDIR)/src/*.h
```
kaspar030: ```*.[hc]```?

tests/pkg_tinycbor/main.c
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @}
+ */
+
+#include "tinycbor/cbor.h"
+
+#include "fmt.h"
+#include "embUnit.h"
+#include "embUnit/embUnit.h"
+#include <string.h>
```
kaspar030: please put the standard header on top

tests/pkg_tinycbor/main.c
```diff
@@ -0,0 +1,106 @@
...
+
+static void test_parse(void)
+{
+    static size_t idx;
+    char *tests[] = {
```
kaspar030: would it save stack here to make these file-scope and const?

tests/pkg_tinycbor/main.c
```diff
@@ -0,0 +1,106 @@
+/*
```
kaspar030: This is laid out like our unittests, right?

bergzand:
Noticed that `Makefile.tinycbor` was not behaving as expected. Fixed that in the last fixup commit.

cladmi:
I have an idea to does not require changing the package.

As the `cbor` package is deprecated, I think it would be better to remove `cbor.h` from the standard `sys/include` like this ?

``` diff
diff --git a/sys/Makefile.include b/sys/Makefile.include
index 24b8c72..aac2784 100644
--- a/sys/Makefile.include
+++ b/sys/Makefile.include
@@ -6,6 +6,10 @@ ifneq (,$(filter crypto,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/include/crypto
 endif
 
+ifneq (,$(filter cbor,$(USEMODULE)))
+  USEMODULE_INCLUDES += $(RIOTBASE)/sys/cbor/include
+endif
+
 ifneq (,$(filter fib,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/sys/posix/include
 endif
diff --git a/sys/include/cbor.h b/sys/cbor/include/cbor.h
similarity index 100%
rename from sys/include/cbor.h
rename to sys/cbor/include/cbor.h
```

bergzand:
#9055 is merged, so I removed the sed rules.

bergzand:
No longer waiting on a PR, ready for review/merge

tests/pkg_tinycbor/tests/01-run.py
```diff
@@ -0,0 +1,22 @@
...
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import os
+import sys
+
+
+def testfunc(child):
+    child.expect(r"OK \(\d+ tests\)")
```
kYc0o: Can you please add the amount of tests that are actually expected to be OK? Is it possible?

bergzand: Done

bergzand:
I've enabled more strict checking for the cbor streams

bergzand:
I have a few changes open upstream to fix the compilation issues for the 8 and 16 bit platforms. For now I'll disable those architectures for this package, but hopefully we can enabled them again with the next release of tinycbor

bergzand:
@kYc0o Murdock agrees now. Can I squash again?

@kaspar030 Are your issues solved here?

kaspar030:
(please squash!)

bergzand:
Squashed!

kYc0o:
Good, fro now it's ok that it works only for 32bit platforms, let's wait for the next release, hopefully the upstream changes are merged by then.

Thanks!

PR #9021, opened 2018-04-25
cpu/nrf52: add PWM implementation

SemjonKerner:
This is an initial peripheral pwm implementation for nordic nrf52 cpu. This implementation provides all functions in pwm.h. It implements nordics hardware pwm. Also this PR adds nrf52dk board configs.

### DIFF
--- a/boards/common/nrf52xxxdk/include/periph_conf.h
+++ b/boards/common/nrf52xxxdk/include/periph_conf.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016-2017 Freie Universität Berlin
+ * Copyright (C) 2016-2018 Freie Universität Berlin
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -107,6 +107,19 @@ static const i2c_conf_t i2c_config[] = {
 };
 #define I2C_NUMOF           (sizeof(i2c_config) / sizeof(i2c_config[0]))
 /** @} */
+
+#if defined(BOARD_NRF52DK) || defined(BOARD_NRF52840DK)
+/**
+ * @name   PWM configuration
+ * @{
+ */
+static const pwm_conf_t pwm_config[] = {
+    { NRF_PWM0, { 28, 29, 30, 31 } }
+};
+#define PWM_NUMOF           (sizeof(pwm_config) / sizeof(pwm_config[0]))
+/** @} */
+
+#endif
 
 #ifdef __cplusplus
 }
--- a/cpu/nrf52/include/periph_cpu.h
+++ b/cpu/nrf52/include/periph_cpu.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2017 Freie Universität Berlin
+ * Copyright (C) 2015-2018 Freie Universität Berlin
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -29,6 +29,11 @@ extern "C" {
  * @brief   System core clock speed, fixed to 64MHz for all NRF52x CPUs
  */
 #define CLOCK_CORECLOCK     (64000000U)
+
+/**
+ * @name    Peripheral clock speed (fixed to 16MHz for nRF52 based CPUs)
+ */
+#define PERIPH_CLOCK        (16000000U)
 
 /**
  * @brief   Redefine some peripheral names to unify them between nRF51 and 52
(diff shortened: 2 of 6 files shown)


boards/common/nrf52xxxdk/include/periph_conf.h
```diff
@@ -108,6 +108,17 @@ static const i2c_conf_t i2c_config[] = {
 #define I2C_NUMOF           (sizeof(i2c_config) / sizeof(i2c_config[0]))
 /** @} */
 
+/**
+ * @name   PWM configuration
+ * @{
+ */
+static const pwm_conf_t pwm_config[] = {
+    { NRF_PWM0, { 23, 24, 25, 26 } }
```
PeterKietzmann: P0.23 doesn't seem to work here. Does it on your board?

cpu/nrf52/periph/pwm.c
```diff
@@ -0,0 +1,184 @@
...
+    DEBUG("cnt: 0x%08x\n", (int)dev(pwm)->SEQ[0].CNT);
+    DEBUG("refresh: 0x%08x\n", (int)dev(pwm)->SEQ[0].REFRESH);
+    DEBUG("enddelay: 0x%08x\n", (int)dev(pwm)->SEQ[0].ENDDELAY);
+
+    /* start sequence */
+    dev(pwm)->TASKS_SEQSTART[0] = 1;
+
+    DEBUG("PWM started\n");
+
+    return real_clk;
```
PeterKietzmann: I assume this value has to be adapted in case of  `PWM_CENTER` ?
```
init 0 1 10000 100
The pwm frequency is set to 10000

init 0 2 10000 100
The pwm frequency is set to 20000
```

However, frequency is 10k in both cases (as expected)

cpu/nrf52/periph/pwm.c
```diff
@@ -0,0 +1,184 @@
...
+
+    /* left aligned */
+    if (pwm_seq[pwm][channel] & POL_MASK) {
+        pwm_seq[pwm][channel] = value | POL_MASK;
+    }
+    /* right aligned */
+    else {
+        pwm_seq[pwm][channel] = dev(pwm)->COUNTERTOP - value;
+    }
+
```
PeterKietzmann: Can you please spend a note about the center mode here?

boards/common/nrf52xxxdk/include/periph_conf.h
SemjonKerner: @PeterKietzmann Yes, it works for me on nrf52dk. Is it possible, that you are testing a different board with P0.23 already in use by another module?

cpu/nrf52/periph/pwm.c
SemjonKerner: @PeterKietzmann i intend on dropping center mode support. I already checked that there is no support in center mode for other CPUs, also center mode is rearly used anyway. If center mode has no special relevance to you, i will just drop it.

PeterKietzmann: I'm okay with dropping that feature.

boards/common/nrf52xxxdk/include/periph_conf.h
PeterKietzmann: To be honest, I don't remember. **But** in my opinion any board that provides the PWM feature and makes use of this *periph_conf.h* should work with the provided configuration of **at least** be documented.

SemjonKerner: I double check that, but i am confident 23 works for me, because i mostly used that pin for debugging.

PeterKietzmann: Would be great if you could check the manuals of the other nrf52 boards that we support. I think it's just 2-3...

SemjonKerner:
@PeterKietzmann Okay, so i moved the pins and restricted the pwm pin configuration to both boards nrf52dk and nrf52840dk. I decided to use ifdefs, because it felt excessive to create new periph_conf files for both boards just to add 4 common pins each.
Also i dropped center mode.

cpu/nrf52/include/periph_cpu.h
```diff
@@ -108,6 +113,49 @@ typedef struct {
...
+
+/**
+ * @brief   Override the PWM mode definitions
+ * @{
+ */
+#define HAVE_PWM_MODE_T
+typedef enum {
+    PWM_LEFT       = PWM_MODE(0, 1),    /**< left aligned PWM */
+    PWM_RIGHT      = PWM_MODE(0, 0),    /**< right aligned PWM */
+    PWM_CENTER     = PWM_MODE(1, 1),    /**< not supported */
```
PeterKietzmann: Should we remove this?

SemjonKerner: I don't have a strong opinion about that. As i saw in other cpus (samd21,efm32) center mode is not supported but PWM_CENTER is set anyway. I did not question this design, but i think it gives easy access / guidance to anyone who wants to implement center mode in the future.

PeterKietzmann: Fine I leave it up to you. Please squash

PeterKietzmann:
`warning: Member PWM_NUMOF (macro definition) of file boards/common/nrf52xxxdk/include/periph_conf.h is not documented`

PeterKietzmann:
@SemjonKerner still around? Please fix Mudrocks complaints. My ACK holds

SemjonKerner:
@PeterKietzmann murdock is green

PR #9032, opened 2018-04-26
tests/mutex_unlock_and_sleep: increase timeout

cladmi:
### Contribution description

Fix test on wsn430 failing because of timeout too small.
In total, the test takes **10 minutes** on wsn430-v1_4.

I also re-run the beginning of the tests on arduino-uno and arduino-mega2560 to verify the timeout is ok (it take ~20seconds on them).

### Alternative

Another solution, could be to not do 100k iterations but only 10k.
I do not know if there is a reason for such a high number ?

I could remove the increased timeout if 10k is enough.

### Issues/PRs references

Tests for the release.

### DIFF
--- a/tests/mutex_unlock_and_sleep/main.c
+++ b/tests/mutex_unlock_and_sleep/main.c
@@ -27,7 +27,11 @@ static volatile int indicator;
 static kernel_pid_t main_pid;
 static char stack[THREAD_STACKSIZE_DEFAULT];
 
+#ifdef BOARD_NATIVE
 static const unsigned KITERATIONS = 100;
+#else
+static const unsigned KITERATIONS = 10;
+#endif
 
 static void *second_thread(void *arg)
 {
--- a/tests/mutex_unlock_and_sleep/main.c
+++ b/tests/mutex_unlock_and_sleep/main.c
@@ -27,6 +27,8 @@ static volatile int indicator;
 static kernel_pid_t main_pid;
 static char stack[THREAD_STACKSIZE_DEFAULT];
 
+static const unsigned KITERATIONS = 100;
+
 static void *second_thread(void *arg)
 {
     (void) arg;
@@ -42,6 +44,7 @@ static void *second_thread(void *arg)
 int main(void)
 {
     uint32_t count = 0;
+    uint32_t kcount = 0;
 
     indicator = 0;
     main_pid = thread_getpid();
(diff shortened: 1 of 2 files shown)


kaspar030:
> the test takes 10 minutes on wsn430-v1_4.

So the arduino is 30 times faster than msp430 in this test? That seems wrong....

Anyhow, reducing the number of iterations seems valid. I'd suggest using some power of two so the modulo doesn't need to do a division (which might be absymally slow in msp430's gcc4.x software emulation). I'd rather not have these 10 minutes in our test stuite for such a simple test.

cladmi:
Oh, sorry, it is the iterations, so the required timeout, that took 20 seconds on arduino, so it was ~7 seven minutes in total.

cladmi:
I tried something to have different values in the main.c.
How I did it is not consistent because I do not know how to check for the architecture in a macro.

 * one for native
 * one default one
 * one for 16bit platforms

Adding an 8bit case may also be good.

Should I go with something like this or do you have another idea ?

cladmi:
I got another problem with timings on the `posix_semaphore` as the timeout margin is too low on wsn430.
So this asks what to do in a more generic "native, fast cpu, slow cpu" separation and how to handle it in tests.

kaspar030:
> I'd suggest using some power of two so the modulo doesn't need to do a division (which might be absymally slow in msp430's gcc4.x software emulation).

Did you try this? Maybe the slowness comes from that modulo.

cladmi:
There are no wsn430 available at the moment I will try when possible.

cladmi:
With 32768 instead of 100k it takes from 8.6 to 10.5 seconds per iteration instead of the 10.5 seconds it would be in theory by dividing the 32 seconds I had before.
So not a huge time gain here, it could be changed to remove the modulo but it would not solve the whole time.

cladmi:
With removing the modulo completely, and doing `(count == 10000)` it takes 8-10 instead of the 32 that I had before. So dropping the modulo is WAY better than doing a power of two modulo.

It actually fits in the 10 seconds thing.

cladmi:
I am happy with what I have:
* Removing the modulo completely, (a power of two modulo on a 32 bit was problematic)
* I divided the number of iterations by 10 to be in a 10 second timeout on wsn430
* I reduced the number of tested iterations to 10.

What do you think ? Can I rebase/squash ?

kaspar030:
pls squash & rebase!

cladmi:
Rebased and split into different commits that explains more the changes.

I will do a final re-test on wsn430 and iotlab-m3, I forgot before pushing.

cladmi:
I updated the timings info in the commit messages for wsn430.

Tests are working.

PR #9035, opened 2018-04-26
drivers/srf04: range finder sensor

SemjonKerner:
This is an initial driver implementation for the srf04 ultra sonic range finder. The driver provides functions for initializing the gpio, triggering a measurement and reading out the measured data. This PR also provides a very simple test.

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -405,6 +405,12 @@ ifneq (,$(filter srf02,$(USEMODULE)))
   USEMODULE += xtimer
 endif
 
+ifneq (,$(filter srf04,$(USEMODULE)))
+  USEMODULE += xtimer
+  FEATURES_REQUIRED += periph_gpio
+  FEATURES_REQUIRED += periph_gpio_irq
+endif
+
 ifneq (,$(filter srf08,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
   USEMODULE += xtimer
--- a/drivers/include/srf04.h
+++ b/drivers/include/srf04.h
@@ -1,0 +1,110 @@
+/*
+ * Copyright (C) 2018 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_srf04 srf04 ultra sonic range finder
+ * @ingroup     drivers_sensors
+ * @brief       Device driver for the srf04 ultra sonic range finder
+ * @{
+ *
+ * @file
+ * @brief       Device driver for the srf04 ultra sonic range finder
+ *
+ * @author      Semjon Kerner <semjon.kerner@fu-berlin.de>
+ */
+
+#ifndef SRF04_H
+#define SRF04_H
+
+#include <stdint.h>
+#include <stdio.h>
+
+#include "periph/gpio.h"
+#include "xtimer.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Status and error return codes
+ */
+enum {
+    SRF04_OK            =  0,   /**< exit without error */
+    SRF04_ERR_INVALID   = -1,   /**< error no valid measurement available*/
+    SRF04_ERR_MEASURING = -2,   /**< error sensor is measuring*/
+    SRF04_ERR_GPIO      = -3,   /**< error initializing gpio*/
+};
+
+/**
+ * @brief   GPIO pins for srf04 device
+ */
+typedef struct {
+    gpio_t trigger;     /**< GPIO Port the trigger pin is connected to */
+    gpio_t echo;        /**< GPIO Port the echo pin is connected to */
+} srf04_params_t;
+
+/**
+ * @brief   Device descriptor for srf04 sensor
+ */
+typedef struct {
+    srf04_params_t p;   /**< GPIO Ports of device */
+    int distance;       /**< raw time of flight distance */
+    uint32_t time;      /**< timestamp of trigger or echo */
+} srf04_t;
+
+/**
+ * @brief   Initialize gpio and interrupt
+ *
+ * @param[out] dev      device descriptor of sensor to initialize
+ * @param[in] params    init param struct holding gpio trigger and echo pins
+ *
+ * @return              SRF04_OK on success
+ * @return              SRF04_GPIO on gpio init failure
+ */
+int srf04_init(srf04_t *dev, const srf04_params_t *params);
+
+/**
+ * @brief   Triggers measurement
+ *
+ * @param[in]  dev      device descriptor of sensor
+ */
+void srf04_trigger(const srf04_t *dev);
+
+/**
+ * @brief   Returns time of flight in ms
+ *
+ * @note    should not be invoked within 50 ms after triggering
+ *
+ * @param[in]  dev      device descriptor of sensor
+ *
+ * @return              time of flight in ms
+ * @return              SRF04_MEASURING if measurement is in progress
+ * @return              SRF04_INVALID if no valid measurement is available
+ */
+int srf04_read(const srf04_t* dev);
+
+/**
+ * @brief   Convenience function triggers a measurement and returns distance
+ *
+ * @note    This function will return after 50 ms once new data is available
+ *
+ * @param[in]  dev      device descriptor of sensor
+ *
+ * @return              time of flight in mm
+ * @return              SRF04_MEASURING if measurement is in progress
+ * @return              SRF04_INVALID if no valid measurement is available
+ */
+int srf04_get_distance(const srf04_t* dev);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* SRF04_H */
+/** @} */
--- a/drivers/srf04/include/srf04_params.h
+++ b/drivers/srf04/include/srf04_params.h
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2018 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_srf04
+ *
+ * @{
+ * @file
+ * @brief       Default configuration for srf04 devices
+ *
+ * @author      Semjon Kerner <semjon.kerner@fu-berlin.de>
+ */
+
+#ifndef SRF04_PARAMS_H
+#define SRF04_PARAMS_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "board.h"
+#include "srf04.h"
+
+/**
+ * @name   Default configuration parameters for SRF04 device
+ * @{
+ */
+#ifndef SRF04_PARAM_TRIGGER
+#define SRF04_PARAM_TRIGGER     GPIO_PIN(0,13)
+#endif
+#ifndef SRF04_PARAM_ECHO
+#define SRF04_PARAM_ECHO        GPIO_PIN(0,14)
+#endif
+
+#ifndef SRF04_PARAMS
+#define SRF04_PARAMS                            \
+    {   .trigger        = SRF04_PARAM_TRIGGER,  \
+        .echo           = SRF04_PARAM_ECHO,     \
+    }
+#endif
+/** @} */
+
+/**
+ * @brief   SRF04 configuration
+ */
+static const srf04_params_t srf04_params[] = {
+    SRF04_PARAMS
+};
+
+#define SRF04_NUMOF     (sizeof(srf04_params) / sizeof(srf04_params[0]))
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* SRF04_PARAMS_H */
+/** @} */
--- a/drivers/srf04/srf04.c
+++ b/drivers/srf04/srf04.c
@@ -1,0 +1,95 @@
+/*
+ * Copyright (C) 2018 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup drivers_srf04
+ * @{
+ *
+ * @file
+ * @brief       driver for srf04 ultra sonic range finder
+ *
+ * @author      Semjon Kerner <semjon.kerner@fu-berlin.de>
+ *
+ * @}
+ */
+
+#include "srf04.h"
+#include "srf04_params.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+#define SRF04_DISTANCE  (584U)
+#define SRF04_SAMPLE_PERIOD (50U * US_PER_MS)
+
+static void _cb(void *arg)
+{
+    uint32_t t = xtimer_now_usec();
+
+    srf04_t* dev = (srf04_t*)arg;
+    if (dev->distance > SRF04_ERR_MEASURING) {
+        dev->distance = SRF04_ERR_MEASURING;
+        dev->time = t;
+    } else {
+        gpio_irq_disable(dev->p.echo);
+        dev->distance = (t - dev->time);
+    }
+}
+
+int srf04_init(srf04_t* dev, const srf04_params_t *params)
+{
+    dev->p = *params;
+
+    dev->distance = SRF04_ERR_INVALID;
+    dev->time = 0;
+
+    if (gpio_init(dev->p.trigger, GPIO_OUT) != 0) {
+        DEBUG("[srf04] Error: could not initialize GPIO trigger pin\n");
+        return SRF04_ERR_GPIO;
+    }
+
+    if (gpio_init_int(dev->p.echo, GPIO_IN, GPIO_BOTH, _cb, (void*)dev) != 0) {
+        DEBUG("[srf04] Error: could not initialize GPIO echo pin\n");
+        return SRF04_ERR_GPIO;
+    }
+
+    gpio_irq_disable(dev->p.echo);
+
+    return SRF04_OK;
+}
+
+void srf04_trigger(const srf04_t* dev)
+{
+    if (dev->distance == SRF04_ERR_MEASURING) {
+        return;
+    }
+
+    gpio_irq_enable(dev->p.echo);
+
+    gpio_set(dev->p.trigger);
+    xtimer_usleep(10);
+    gpio_clear(dev->p.trigger);
+}
+
+int srf04_read(const srf04_t* dev)
+{
+    return dev->distance;
+}
+
+int srf04_get_distance(const srf04_t* dev)
+{
+    /* trigger new reading */
+    srf04_trigger(dev);
+    /* give the sensor the required time for sampling */
+    xtimer_usleep(SRF04_SAMPLE_PERIOD);
+    /* get the result */
+    if (dev->distance >= SRF04_OK) {
+        return ((dev->distance * 100) / SRF04_DISTANCE);
+    }
+    return dev->distance;
+}
--- a/tests/driver_srf04/Makefile
+++ b/tests/driver_srf04/Makefile
@@ -1,0 +1,9 @@
+include ../Makefile.tests_common
+
+BOARD ?= nrf52dk
+
+RIOTBASE ?= $(CURDIR)/../..
+
+USEMODULE += srf04
+
+include $(RIOTBASE)/Makefile.include
--- a/tests/driver_srf04/README.md
+++ b/tests/driver_srf04/README.md
@@ -1,0 +1,17 @@
+tests/driver_srf04
+================
+This example shows the usage of an srf04 module.
+The application uses a timer and two gpio pins.
+The module is an ultrasonic range finder with a small protocol [1]:
+- trigger pin is raised high for 10 us by calling trigger function
+- after a sample period of 50 ms the value can be read out
+
+[1] www.robot-electronics.co.uk/htm/srf04tech.htm
+
+Usage
+=====
+
+Build, flash and start the application:
+
+    export BOARD=your_board
+    make flash term
--- a/tests/driver_srf04/main.c
+++ b/tests/driver_srf04/main.c
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2018 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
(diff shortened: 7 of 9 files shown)


drivers/include/srf04.h
```diff
@@ -0,0 +1,101 @@
...
+typedef struct {
+    srf04_state_t state;
+    gpio_t trigger;
+    gpio_t echo;
+    uint32_t distance;
+    uint32_t time;
+} srf04_params_t;
+
+/**
+ * @brief   GPIO pins for hc-sr04_t device
```
gebart: Wrong docs

drivers/include/srf04.h
```diff
@@ -0,0 +1,101 @@
...
+    SRF04_GPIO    = -1,              /**< error initializing gpio*/
+    SRF04_INT     = -2,              /**< error initializing gpio interrupt*/
+};
+
+typedef enum {
+    SRF04_MEASURING,
+    SRF04_IDLE,
+} srf04_state_t;
+
+typedef struct {
```
gebart: Missing docs

drivers/include/srf04.h
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * @brief   Status and error return codes
+ */
+enum {
+    SRF04_OK      =  0,              /**< exit without error */
+    SRF04_GPIO    = -1,              /**< error initializing gpio*/
+    SRF04_INT     = -2,              /**< error initializing gpio interrupt*/
+};
+
+typedef enum {
```
gebart: Missing docs

drivers/include/srf04.h
```diff
@@ -0,0 +1,101 @@
...
+ * @return              SRF04_GPIO on gpio init failure
+ */
+int srf04_init(srf04_t *dev, const gpio_t trigger, const gpio_t echo);
+
+/**
+ * @brief   Triggers measurement
+ *
+ * @param[in]  dev      device descriptor of sensor
+ *
+ * @return              SRF04_OK
```
gebart: Does it always return this value? Is there any reason to have a return value then?

drivers/srf04/include/srf04_params.h
```diff
@@ -0,0 +1,65 @@
...
+        .trigger        = SRF04_PARAM_TRIGGER,  \
+        .echo           = SRF04_PARAM_ECHO,     \
+        .distance       = (0U),                 \
+        .time           = (0U),                 \
+    }
+#endif
+/** @} */
+
+/**
+ * @brief   BMX055 configuration
```
gebart: Wrong docs

drivers/include/srf04.h
```diff
@@ -0,0 +1,101 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Status and error return codes
+ */
+enum {
+    SRF04_OK      =  0,              /**< exit without error */
+    SRF04_GPIO    = -1,              /**< error initializing gpio*/
```
aabadie: I would slightly update the name of the error enums by introducing ERR: SRF04_ERR_GPIO, etc

drivers/include/srf04.h
```diff
@@ -0,0 +1,101 @@
...
+    gpio_t echo;
+    uint32_t distance;
+    uint32_t time;
+} srf04_params_t;
+
+/**
+ * @brief   GPIO pins for hc-sr04_t device
+ */
+typedef struct {
+    srf04_params_t p;
```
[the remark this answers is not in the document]
aabadie: According to [this page](https://github.com/RIOT-OS/RIOT/wiki/Guide:-Writing-a-device-driver-in-RIOT#device-descriptor-and-parameter-configuration), this is a recommended pattern. But your concern sounds valid. We have the same question in #7067. @haukepetersen what's your opinion ?

drivers/srf04/include/srf04_params.h
```diff
@@ -0,0 +1,65 @@
...
+#ifndef SRF04_PARAM_ECHO
+#define SRF04_PARAM_ECHO        GPIO_PIN(0,14)
+#endif
+
+#ifndef SRF04_PARAMS
+#define SRF04_PARAMS                            \
+    {   .state          = SRF04_IDLE,           \
+        .trigger        = SRF04_PARAM_TRIGGER,  \
+        .echo           = SRF04_PARAM_ECHO,     \
+        .distance       = (0U),                 \
```
aabadie: Why do you need distance in the initialization parameters ? Same question for time and I also saw below that state is also added in the parameter struct. I think those 3 should go in the device descriptor.

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,83 @@
...
+int srf04_init(srf04_t* dev, const gpio_t trigger, const gpio_t echo)
+{
+    dev->p.state = SRF04_IDLE;
+    dev->p.trigger = trigger;
+    dev->p.echo = echo;
+    dev->p.distance = 0;
+    dev->p.time = 0;
+
+    if ((gpio_init(dev->p.trigger, GPIO_OUT) != 0) |
+        (gpio_init(dev->p.echo, GPIO_IN) != 0)){
```
aabadie: missing space before {

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,83 @@
...
+        DEBUG("[srf04] Error: could not initialize GPIO\n");
```
aabadie: I would split this check in 2: one for the trigger pin, one for the echo pin

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,83 @@
...
+        return SRF04_GPIO;
+    }
+
+    if (gpio_init_int(dev->p.echo, GPIO_IN, GPIO_BOTH, _cb, (void*)dev) != 0) {
```
aabadie: If echo is initialized here as interrupt, the previous initialization (without interrupt) seems useless

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,83 @@
...
+    gpio_irq_disable(dev->p.echo);
+
+    return SRF04_OK;
+}
+
+int srf04_trigger(const srf04_t* dev)
+{
+    gpio_irq_enable(dev->p.echo);
+
+    //trigger on falling edge for 10us
```
aabadie: use C-style comment

tests/driver_srf04/Makefile
```diff
@@ -0,0 +1,13 @@
+APPLICATION = driver_srf04
+
+BOARD ?= calliope-mini
+
+RIOTBASE ?= $(CURDIR)/../..
+
+FEATURES_REQUIRED = periph_gpio
+
+USEMODULE += srf04
+USEMODULE += xtimer
```
aabadie: Since xtimer is used internally by the driver, it should be added as a dependency of the driver module in drivers/Makefile.dep

tests/driver_srf04/Makefile
```diff
@@ -0,0 +1,13 @@
...
+USEMODULE += shell
```
aabadie: The test application is not using the shell so this can be removed.

tests/driver_srf04/README.md
```diff
@@ -0,0 +1,14 @@
+tests/driver_srf04
+================
+This example shows the usage of an srf04 module.
+The application exerts a timer and two gpio.
```
aabadie: exerts ?

tests/driver_srf04/README.md
```diff
@@ -0,0 +1,14 @@
...
+The module is an ultrasonic range finder with a small protocoll,
```
aabadie: typo: protocol**l**

tests/driver_srf04/README.md
```diff
@@ -0,0 +1,14 @@
...
+hence an implementation can be kept simple.
```
aabadie: I don't think this comment is very useful. Better detail the mentioned protocol ? or give a link to a datasheet.

tests/driver_srf04/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Usage
+=====
+
+Build, flash and start the application:
+```
+export BOARD=your_board
+make -j 2 flash term
```
aabadie: Missing closing \`\`\`
Or just indent the block with 4 spaces (and remove the opening \`\`\`). I would put the command line on a single line and remove `-j2`

tests/driver_srf04/main.c
```diff
@@ -0,0 +1,48 @@
...
+
+int main(void)
+{
+    printf("SRF04 range finder example\n");
+
+    srf04_t dev;
+
+    srf04_init(&dev, srf04_params[0].trigger, srf04_params[0].echo);
+
+    while(1) {
```
aabadie: missing space after `while`

tests/driver_srf04/main.c
```diff
@@ -0,0 +1,48 @@
...
+
+#include "srf04_params.h"
+#include "srf04.h"
+#include "xtimer.h"
+
+#define SAMPLE_PERIOD       (50U * US_PER_MS)
+
+int main(void)
+{
+    printf("SRF04 range finder example\n");
```
aabadie: Use `puts` without `\n`

tests/driver_srf04/main.c
```diff
@@ -0,0 +1,48 @@
...
+    srf04_t dev;
+
+    srf04_init(&dev, srf04_params[0].trigger, srf04_params[0].echo);
+
+    while(1) {
+        srf04_trigger(&dev);
+
+        xtimer_usleep(SAMPLE_PERIOD);
+
+        printf("D: %d mm\n", (((srf04_read(&dev) * 100) / 292) / 2));
```
aabadie: Here, I think it would be better to extend the driver API with a `srf04_read_distance` that return the mesure in `mm`. The computation seems weird (x * 100 / 292 / 2). And better define constants in the driver header instead of hard coded values.

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,93 @@
...
+#endif
+
+static bool state;
+
+static void _cb(void *arg)
+{
+    uint32_t t = xtimer_now_usec();
+
+    srf04_t* dev = (srf04_t*)arg;
+    if (dev->state == SRF04_IDLE) {
```
haukepetersen: we could simply use `dev->distance` for keeping the state:
```c
if (dev->distance > SRF04_MEASURING) {
    .... // start new measurement
}
else {
    ... // save actual distance time
} 
```

drivers/include/srf04.h
```diff
@@ -0,0 +1,124 @@
...
+void srf04_trigger(const srf04_t *dev);
+
+/**
+ * @brief   Returns time of flight in ms
+ *
+ * @note    should not be invoked within 50 ms after triggering
+ *
+ * @param[in]  dev      device descriptor of sensor
+ *
+ * @return              time of flight in ms
```
haukepetersen: add something like
```
@return              time of flight in ms (>= SRF04_OK)
@return  SRF04_MEASURING if measurement is in progress
@return SRF04_INVALID if no valid measurement is available
```

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,93 @@
...
+        dev->state = SRF04_IDLE;
+        dev->distance = (t - dev->time);
+    }
+}
+
+int srf04_init(srf04_t* dev)
+{
+    dev->p = srf04_params[0];
+
+    dev->state = SRF04_IDLE;
```
haukepetersen: remove

drivers/include/srf04.h
```diff
@@ -0,0 +1,124 @@
...
+    gpio_t echo;
+} srf04_params_t;
+
+/**
+ * @brief   Device descriptor for srf04 sensor
+ */
+typedef struct {
+    srf04_params_t p;
+    srf04_state_t state;
+    uint32_t distance;
```
haukepetersen: remove `state` and change to `int distance;`

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,93 @@
...
+    dev->time = 0;
+
+    if (gpio_init(dev->p.trigger, GPIO_OUT) != 0) {
+        DEBUG("[srf04] Error: could not initialize GPIO trigger pin\n");
+        return SRF04_ERR_GPIO;
+    }
+
+    if (gpio_init_int(dev->p.echo, GPIO_IN, GPIO_BOTH, _cb, (void*)dev) != 0) {
+        DEBUG("[srf04] Error: could not initialize GPIO echo pin\n");
+        return SRF04_ERR_INT;
```
haukepetersen: I think I would simplify here and drop the explicit `ERR_INT` value in favor of returning `SRF04_ERR_GPIO` in both cases.

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,93 @@
...
+
+    gpio_irq_disable(dev->p.echo);
+
+    return SRF04_OK;
+}
+
+void srf04_trigger(const srf04_t* dev)
+{
+    gpio_irq_enable(dev->p.echo);
+
```
haukepetersen: maybe it makes sense to add a check so we don't trigger a new measurement while one is in progress:
```c
if (dev->distance == SRF04_MEASURING) {
    return;
}
```

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,93 @@
...
+}
+
+int srf04_read(const srf04_t* dev)
+{
+    return dev->distance;
+}
+
+int srf04_read_distance(const srf04_t* dev)
+{
+    return ((dev->distance * 100) / SRF04_DISTANCE);
```
haukepetersen: here we should also add a check:
```c
if (dev->distance >= SRF04_OK) {
    return ((dev->distance * 100) / SRF04_DISTANCE);
}
return dev->distance;
```

tests/driver_srf04/main.c
```diff
@@ -0,0 +1,48 @@
...
+    srf04_t dev;
+
+    srf04_init(&dev);
+
+    while (1) {
+        srf04_trigger(&dev);
+
+        xtimer_usleep(SAMPLE_PERIOD);
+
+        printf("D: %d mm\n", srf04_read_distance(&dev));
```
haukepetersen: with my proposed changes, it could be nice to check the return value of `srf04_read_distance()` and print an explicit error message in case the value is not valid (`val < SRF04_OK`).

drivers/include/srf04.h
```diff
@@ -0,0 +1,124 @@
...
+ * @brief   Returns time of flight in mm
+ *
+ * @note    should not be invoked within 50 ms after triggering
+ *
+ * @param[in]  dev      device descriptor of sensor
+ *
+ * @return              time of flight in mm
+ */
+int srf04_read_distance(const srf04_t* dev);
+
```
haukepetersen: last proposal: how about a convenience function that combines `_trigger()` and `read_distance()` together with the mandatory wait time?! This could be useful for very simple use cases (as e.g. the example application below...).

PeterKietzmann:
@SemjonKerner I guess you 'll ping the reviewers once you addressed their comments?

PeterKietzmann:
No stress from my side. In general it makes sense to inform or even ping reviewers after you addressed their comments. @haukepetersen, @gebart, @aabadie if your comments were addressed, please remove the change request.

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,98 @@
...
+        dev->time = t;
+    } else {
+        gpio_irq_disable(dev->p.echo);
+        dev->distance = (t - dev->time);
+    }
+}
+
+int srf04_init(srf04_t* dev)
+{
+    dev->p = srf04_params[0];
```
gebart: This limits the driver to a single hardware sensor. I suggest adding an argument to the init function, passing a pointer to a params struct. See other device drivers in the tree, for example https://github.com/RIOT-OS/RIOT/blob/master/drivers/io1_xplained/io1_xplained.c#L37-L39

drivers/Makefile.dep
```diff
@@ -321,6 +321,10 @@ ifneq (,$(filter srf02,$(USEMODULE)))
   USEMODULE += xtimer
 endif
 
+ifneq (,$(filter srf04,$(USEMODULE)))
+  USEMODULE += xtimer
```
aabadie: The driver requires `periph_gpio` internally, you also need to add:
```
FEATURES_REQUIRED += periph_gpio
```

tests/driver_srf04/Makefile
```diff
@@ -0,0 +1,11 @@
+APPLICATION = driver_srf04
+
+BOARD ?= nrf52dk
+
+RIOTBASE ?= $(CURDIR)/../..
+
+FEATURES_REQUIRED = periph_gpio
```
aabadie: This is not needed anymore with this added in the driver module dependencies.

tests/driver_srf04/Makefile
```diff
@@ -0,0 +1,11 @@
+APPLICATION = driver_srf04
```
aabadie: Use `include ../Makefile.tests_common` instead of that line.

drivers/include/srf04.h
```diff
@@ -0,0 +1,136 @@
...
+    SRF04_ERR_INVALID   = -1,   /**< error no valid measurement available*/
+    SRF04_ERR_MEASURING = -2,   /**< error sensor is measuring*/
+    SRF04_ERR_GPIO  = -3,   /**< error initializing gpio*/
+};
+
+/**
+ * @brief   GPIO pins for srf04 device
+ */
+typedef struct {
+    gpio_t trigger;
```
aabadie: The attributes of this struct are not documented, same for the `srf04_t` struct below.

drivers/include/srf04.h
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * @brief defines sensor required sample time
+ */
+#define SRF04_SAMPLE_PERIOD (50U * US_PER_MS)
+
+/**
+ * @brief   Status and error return codes
+ */
+enum {
+    SRF04_OK        =  0,   /**< exit without error */
```
aabadie: Maybe align a bit the values and comments

SemjonKerner:
Hey @haukepetersen , @gebart , @aabadie finally there is a new commit fixing your requested changes.

SemjonKerner:
@haukepetersen @gebart @aabadi just a small reminder

tests/driver_srf04/main.c
```diff
@@ -0,0 +1,46 @@
...
+{
+    puts("SRF04 range finder example");
+
+    srf04_t dev;
+    srf04_init(&dev, &srf04_params[0]);
+    int distance;
+
+    while (1) {
+        distance = srf04_get_distance(&dev);
+        if (distance < SRF04_OK) {
```
aabadie: This check is weird since the function can return a distance or error codes (negative values). Maybe just compare to 0 ? Or use a switch and address each potential error code ?

SemjonKerner: Comparing to SRF04_OK is the same thing, since the struct defines it as zero and all errors would evaluate negative. I could distinguish between errors, but i deem it not necessary.

SemjonKerner: I thought about your concern and now i believe to understand what you meant. Distance should only be SRF04_OK (zero), when srf04_init returns successfully. After the first get_distance (wich happens before) it will either have a valid distance (greater or equal to 0) or give an error. So comparing to SRF04_OK is just meant to be a clean way of checking for new data.
As the sensor only measures values from a minimum distance of i think 3 cm it should always be greater then zero when successfull.

drivers/include/srf04.h
```diff
@@ -0,0 +1,122 @@
...
+extern "C" {
+#endif
+
+/**
+ * @brief defines distance divisor
+ *
+ * @note for inch define "SRF04_DISTANCE (1480U)"
+ */
+#ifndef SRF04_DISTANCE
+#define SRF04_DISTANCE  (584U)
```
haukepetersen: as discussed offline: drop the support for inches and move this define into the source file

drivers/include/srf04.h
```diff
@@ -0,0 +1,122 @@
...
+#endif
+
+/**
+ * @brief defines sensor required sample time
+ */
+#define SRF04_SAMPLE_PERIOD (50U * US_PER_MS)
```
haukepetersen: this is static and has nothing to do with the interface -> move into source file

drivers/include/srf04.h
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * @return              time of flight in ms
+ * @return              SRF04_MEASURING if measurement is in progress
+ * @return              SRF04_INVALID if no valid measurement is available
+ */
+int srf04_read(const srf04_t* dev);
+
+/**
+ * @brief   Convenience function triggers a measurement and returns distance
+ *
```
haukepetersen: I would add a comment about the blocking nature of this function (will only return after the new value is acquired, which takes ~50ms..).

drivers/srf04/srf04.c
```diff
@@ -0,0 +1,93 @@
...
+    srf04_trigger(dev);
+    /* give the sensor the required time for sampling */
+    xtimer_usleep(SRF04_SAMPLE_PERIOD);
+    /* get the result */
+    if (dev->distance >= SRF04_OK) {
+        return ((dev->distance * 100) / SRF04_DISTANCE);
+    }
+    return dev->distance;
+}
+
```
haukepetersen: duplicate newline :-)

tests/driver_srf04/main.c
```diff
@@ -0,0 +1,46 @@
...
+#include "srf04_params.h"
+#include "srf04.h"
+
+int main(void)
+{
+    puts("SRF04 range finder example");
+
+    srf04_t dev;
+    srf04_init(&dev, &srf04_params[0]);
+    int distance;
```
haukepetersen: scope: cppcheck will probably complain that `distance` should be define inside the while loop below

tests/driver_srf04/main.c
```diff
@@ -0,0 +1,46 @@
...
+
+#include "srf04_params.h"
+#include "srf04.h"
+
+int main(void)
+{
+    puts("SRF04 range finder example");
+
+    srf04_t dev;
+    srf04_init(&dev, &srf04_params[0]);
```
haukepetersen: for demonstration purposes, it is good practice to check the return value of the init function, e.g.
```c
if (srf04_init(dev, params) != SRF04_OK) {
  puts("error initializing...");
  return 1;
}
```
or similar..

tests/driver_srf04/main.c
haukepetersen: I'd say the check is fine. Maybe more intuitive would simply be to test for `distance < 0`, semantically considering the `0` as distance value instead of some arbitrarily defined named value. But I'd say this is up to the creative freedom of the author :-)

SemjonKerner:
@haukepetersen fixes are commited.
@gebart are you still looking into this?

gebart:
My comments seem to have been addressed

haukepetersen:
all green -> lets go

@SemjonKerner nice job, thanks!

PR #9037, opened 2018-04-27
tests/xtimer_usleep_short: add expected runtime

Josar:
In some cases the sleep time is very long compared to the expected sleep time.
This pr adds a functionality to test against the expected sleep time.
If the sleep time is far of the expected sleep time an error is thrown.

### DIFF
--- a/tests/xtimer_usleep_short/main.c
+++ b/tests/xtimer_usleep_short/main.c
@@ -22,16 +22,24 @@
 #define TEST_USLEEP_MIN (0)
 #define TEST_USLEEP_MAX (500)
 
+#define TEST_TIME (125250)
+
 int main(void)
 {
     xtimer_sleep(3);
     printf("This test will call xtimer_usleep for values from %d down to %d\n",
             TEST_USLEEP_MAX, TEST_USLEEP_MIN);
 
+    uint32_t start, sleeping_time=0;
+
     for (int i = TEST_USLEEP_MAX; i >= TEST_USLEEP_MIN; i--) {
         printf("going to sleep %d usecs...\n", i);
+        start = xtimer_now_usec();
         xtimer_usleep(i);
+        sleeping_time += xtimer_now_usec() - start;
     }
+
+    printf("Slept for %" PRIu32 " expected %" PRIu32"\n", sleeping_time, TEST_TIME);
 
     puts("[SUCCESS]");
 
--- a/tests/xtimer_usleep_short/main.c
+++ b/tests/xtimer_usleep_short/main.c
@@ -22,13 +22,23 @@
 #define TEST_USLEEP_MIN (0)
 #define TEST_USLEEP_MAX (500)
 
+#ifdef BOARD_NATIVE
+/* native can sometime take more time to respond as it is not real time */
+#define TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US (1000)
+#else
+#define TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US (20)
+#endif /* BOARD_NATIVE */
+
 int main(void)
 {
     xtimer_sleep(3);
     printf("This test will call xtimer_usleep for values from %d down to %d\n",
            TEST_USLEEP_MAX, TEST_USLEEP_MIN);
 
-    uint32_t test_time = 0, sleeping_time = 0;
+    printf("Expected delay margin is %d us\n",
+           TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US);
+
+    uint32_t test_time = 0, sleeping_time = 0, margin_fault = 0;
 
     for (int i = TEST_USLEEP_MAX; i >= TEST_USLEEP_MIN; i--) {
         printf("going to sleep %d us\n", i);
@@ -36,14 +46,30 @@ int main(void)
         xtimer_usleep(i);
         uint32_t slept = xtimer_now_usec() - start;
         printf("Slept for      %" PRIu32 " us\n", slept);
+
+        if (slept < (unsigned int)i) {
+            puts("Timeout to short");
+            ++margin_fault;
+        }
+        else if (slept > (unsigned int)i + TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US) {
+            puts("Timeout longer than expected margin.");
+            ++margin_fault;
+        }
+
         sleeping_time += slept;
         test_time += i;
     }
 
-    printf("Slept for %" PRIu32 " expected %" PRIu32 "\n",
+    printf("Slept for %" PRIu32 " us expected %" PRIu32 " us\n",
            sleeping_time, test_time);
 
-    puts("[SUCCESS]");
-
-    return 0;
+    if (margin_fault != 0) {
+        printf("Sleep delay margin was not kept for %" PRIu32 " times\n", margin_fault);
+        puts("[FAILED]");
+        return 1;
+    }
+    else {
+        puts("[SUCCESS]");
+        return 0;
+    }
 }
--- a/tests/xtimer_usleep_short/main.c
+++ b/tests/xtimer_usleep_short/main.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2017 HAW-Hamburg
+ *               2018 Josua Arndt
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -14,6 +15,7 @@
  * @brief       xtimer_usleep_short test application
  *
  * @author      Michel Rottleuthner <michel.rottleuthner@haw-hamburg.de>
+ * @author      Josua Arndt <jarndt@ias.rwth-aachen.de>
  * @}
  */
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


tests/xtimer_usleep_short/tests/01-run.py
```diff
@@ -25,6 +25,15 @@ def testfunc(child):
             break
         i = i - 1
 
+    child.expect(u"Slept for (\\d+) expected (\\d+)", timeout=3)
+    runtime = int(child.match.group(1))
+    expected = int(child.match.group(2))
+
+    if runtime < expected*0.9 or runtime > expected*1.1:
```
aabadie: This should be tested in the test application not by the pexpect script

tests/xtimer_usleep_short/main.c
```diff
@@ -22,17 +22,25 @@
...
 
+#define TEST_TIME (125250)
+
 int main(void)
 {
     xtimer_sleep(3);
     printf("This test will call xtimer_usleep for values from %d down to %d\n",
             TEST_USLEEP_MAX, TEST_USLEEP_MIN);
 
+    uint32_t start, sleeping_time=0;
```
aabadie: put spaces around `=` operator

tests/xtimer_usleep_short/main.c
```diff
@@ -22,17 +22,25 @@
 #define TEST_USLEEP_MIN (0)
 #define TEST_USLEEP_MAX (500)
 
+#define TEST_TIME (125250)
```
aabadie: How did you determine this value ?

Josar: http://www.wolframalpha.com/input/?i=500%2B499%2B498%2B+...+%2B+1

aabadie: Then threre's direct relationship with TEST_USLEEP_MAX, so I think it's useless.

Josar: I don't understand what you mean, but changed the value to be calculated.

aabadie:
This looks more to what I had in mind now. I tested on arduino-zero, microbit and an stm32l4 based board. It only works on the latter, the other ones exceeds the expected test time (even with 20% error tolerance).
What hardware are you using @Josar ?

Josar:
I just add some improvements to the xtimer while testing the jiminy board.
Maybe the tollerance has to be adapted or the changes from #8990 have to be applied.

smlng:
I would make the check in the python test script instead, i.e. just print actual an expected sleep timings and then make any check/verification in the test later on. That makes it more flexible.

Josar:
@aabadie 
> This should be tested in the test application not by the pexpect script

@smlng 

>  check in the python test script instead,

Could you please discuss this and come to a conclusion!

Josar:
Added reports for each sleep, changed the test to be done in python and made the deviation configurable.

Josar:
@smlng and @cladmi any thoughts?

smlng:
I'm still (and in general) in favour of making any checks and verifications in the python script. Thus let the C test program be simple with limited output which also reduces binary size. I also think its easier to make maths for verification in Python than in C.

Josar:
@aabadie @smlng any thoughts about this?

tests/xtimer_usleep_short/tests/01-run.py
```diff
@@ -10,21 +10,31 @@
 import sys
 import pexpect
 
+DEVIATION = .1
```
aabadie: This is unused

tests/xtimer_usleep_short/main.c
```diff
@@ -22,18 +22,54 @@
...
+           TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US);
+
+    uint32_t test_time = 0, sleeping_time = 0, margin_fault = 0;
 
     for (int i = TEST_USLEEP_MAX; i >= TEST_USLEEP_MIN; i--) {
-        printf("going to sleep %d usecs...\n", i);
+        printf("going to sleep %d us\n", i);
+        uint32_t start = xtimer_now_usec();
         xtimer_usleep(i);
+        uint32_t slept = xtimer_now_usec() - start;
```
smlng: making two `xtimer_now` calls is not necessary when saving the last (from previous iteration) and also assuming that all the other calls cost only marginal time compared to the sleep. `xtimer_now` is rather expensive, anyways.

tests/xtimer_usleep_short/main.c
```diff
@@ -22,18 +22,54 @@
...
+        printf("Slept for      %" PRIu32 " us\n", slept);
+
+        if (slept < (unsigned int)i) {
```
smlng: as already written several times: I'm against making any checks within in the C-code and would rather do this in the  (python) test script. We should keep the C test code simple avoiding unnecessary maths operations and also increasing the code size with such addition code and output. Its already the case that many tests don't fit on all boards just because of excessive output.

tests/xtimer_usleep_short/main.c
Josar: The print of  `"going to sleep %d us\n"`, if i am not mistaken, needs:
20 * 10 * 1000/9600Buad = 20,8 ms

Then the second print will also need some time.
This test goes from 500ms down to 1ms. So is the time marginal or not?

tests/xtimer_usleep_short/main.c
Josar: Maybe there is a way to get the python script print sequentially?

The idea behind this is to make it obvious if the `XTIMER_BACKOFF` border or something else causes the error.
Thus the print has to be exactly after the sleep, not 5 sleeps after it.

Adding the slept time is neccessary to get a hint of errors like this https://github.com/RIOT-OS/RIOT/pull/9211

tests/xtimer_usleep_short/main.c
cladmi: I am also in favor of doing two times `xtimer_now` it helped timing tests correctly, printf takes time.

tests/xtimer_usleep_short/main.c
cladmi: There are boards that we cannot reboot with `make reset` so cannot run the automated python tests on them. If the test can alone say it is working I find it better.
All the tests based on `embunit` are doing asserts on the node and python only checks for `[SUCCESS]` already.

Also, here they are not `math` operations, they are simple integer comparison and addition.

tests/xtimer_usleep_short/tests/01-run.py
```diff
@@ -12,19 +12,27 @@
 
 
 def testfunc(child):
-    child.expect(u"This test will call xtimer_usleep for values from \\d+ down to \\d+\r\n")
+    child.expect(u"This test will call xtimer_usleep for values from (\d+) down to (\d+)")
```
cladmi: Now that we are using python3 none of the string required the `u` in front anymore.

tests/xtimer_usleep_short/tests/01-run.py
```diff
@@ -12,19 +12,27 @@
...
+    sleep_max = int(child.match.group(1))
+    sleep_min = int(child.match.group(2))
 
-    i = 500
+    i = sleep_max - sleep_min
 
     while (i >= 0):
         try:
-            child.expect(u"going to sleep \\d+ usecs...\r\n", timeout=3)
+            child.expect(u"going to sleep (\d+) us", timeout=3)
```
cladmi: Here you can replace `(\d+)` by the value of `i` or if renamed `delay`. It would prevent issue when testing with text taken from node before `make reset`.

tests/xtimer_usleep_short/tests/01-run.py
```diff
@@ -12,19 +12,27 @@
...
+            sleep_exp = int(child.match.group(1))
```
cladmi: sleep_exp would not be necessary as it would have matched against the value of `i`.

tests/xtimer_usleep_short/tests/01-run.py
```diff
@@ -12,19 +12,27 @@
...
+
+            child.expect(u"Slept for      (\d+) us", timeout=3)
+
         except pexpect.TIMEOUT:
-            print("xtimer stuck when trying to sleep %d usecs" % (i+1))
+            print("xtimer stuck when trying to sleep %d us" % (sleep_exp))
             print("[FAILED]")
             break
```
cladmi: This `break` should be an `exit(1)`, not print `failed` then `success`…
It's a good time to fix it.

tests/xtimer_usleep_short/main.c
```diff
@@ -22,18 +22,54 @@
...
+#ifdef BOARD_NATIVE
+/* native can sometime take more time to respond as it is not real time */
+#define TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US (1000)
+#else
+#define TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US (20)
+#endif /* BOARD_NATIVE */
+
 int main(void)
 {
     xtimer_sleep(3);
```
cladmi: This `xtimer_sleep(3)` break my tests when doing `make flash test` on some boards as the test could match the `child.expect(u"This test will call xtimer_usleep for values from (\d+) down to (\d+)")` line before `make reset` is done.

tests/xtimer_usleep_short/main.c
```diff
@@ -22,18 +22,54 @@
 #define TEST_USLEEP_MIN (0)
 #define TEST_USLEEP_MAX (500)
 
+#ifdef BOARD_NATIVE
+/* native can sometime take more time to respond as it is not real time */
+#define TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US (1000)
+#else
+#define TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US (20)
```
cladmi: The margin may not be enough as mentioned in #8990.
I need to do the xtimer_configure to check if it helps getting in the margin.

tests/xtimer_usleep_short/main.c
```diff
@@ -22,18 +22,54 @@
...
+           TEST_USLEEP_MAX, TEST_USLEEP_MIN);
+
+    printf("Expected delay margin is %d us\n",
+           TEST_XTIMER_USLEEP_SHORT_SLEEP_MARGIN_US);
+
+    uint32_t test_time = 0, sleeping_time = 0, margin_fault = 0;
 
     for (int i = TEST_USLEEP_MAX; i >= TEST_USLEEP_MIN; i--) {
-        printf("going to sleep %d usecs...\n", i);
+        printf("going to sleep %d us\n", i);
```
cladmi: Inconsistent with first letter lowercase.

cladmi:
I can do a PR on your branch if you want for the updating script changes (on monday though).

tests/xtimer_usleep_short/main.c
Josar: Any suggestions on how to fix this? Increase the sleep?

Josar:
@aabadie @smlng @cladmi tried to address all requested changes. However, the probelm with testing in C-Code or not is still not settled.

Josar:
@aabadie @smlng @cladmi did you find a consensus for the testing?

smlng:
I'm still against having any evaluation in the (test) application, specifically this delay margin. IMHO its up to the human tester or a high layer test framework to decide whether the test passes or fails according to the values reported/printed. It might also be that we only want to evaluate if a PR improves or degrades the sleep time precision, and don't want to look that much into the actual margin.

Josar:
So @smlng you and @aabadie do not agree. Is there a possibility that you work out a solution?

Josar:
@cladmi @aabadie any further remarks?

Issue #9043, opened 2018-04-27
enc28j60: packet tranmission problem


gschorcht:
Sometimes, either directly after a reset or after some time of operation, the ENC28J60 stops the transmission of packets. Transmit interrupts are then not generated. A similar behaviour is described in the errata sheet section 12, but only for half duplex mode.

gschorcht:
It seems that the transmission logic sometimes hangs. In that case, after writing the data to send into the buffer and starting the transmission process with ```cmd_bfs(dev, REG_ECON1, -1, ECON1_TXRTS);```, it seems that nothing happens. The transmission process is neither finished nor aborted. Neither the ```EIR.TXIF``` nor ```EIR.TXERIF``` are set. When ```ns_send``` is called next time, the ```ECON1_TXRTS``` is still set. It seems to help to reset the transmission logic in that case. After resetting the tranmission logic, the next packet can send without any problems. Of course, the packet that was still not send is destroyed in that case.

PR #9044, opened 2018-04-27
enc28j60: fix of #9043

gschorcht:
This pull request solves both problems:

1. Sometimes, either directly after reset or after some time operation, the ENC28J80 doesn't generate interrupts when packet transmission has been finished. Workaround has been found in chip errata.

2. Compilation errors when MODULE_NETSTATS_L2 is enabled.


gschorcht:
After a lot of more stress tests it became clear, that the workaround of packet transmission problem doesn't help always. I have to investigate the problem and a workaround a bit more.

cladmi:
You can already do a PR for the `MODULE_NETSTATS_L2` fix as this one is a compilation issue.

PR #9046, opened 2018-04-28
uuid: Initial import of RFC4122 UUID functions

bergzand:
Provides functions for type 3, 4 and 5 UUID generations.

Type 3: MD5 name based
Type 4: Full (pseudo)random
Type 5: SHA-1 name based

UUID type 1 is timestamp based and requires an accurate time source. For
this reason it is left out of this implementation. UUID type 2 is not
defined in RFC 4122 and thus also not included here

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -728,6 +728,11 @@ ifneq (,$(filter tlsf-malloc,$(USEMODULE)))
   USEPKG += tlsf
 endif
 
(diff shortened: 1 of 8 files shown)


bergzand:
Added an `uuid_equal` function for easy comparison. Uncrustified and adapted the test.

kYc0o:
And rebase.

bergzand:
rebased and squashed!

Issue #8212, opened 2017-12-06
Fix common issues with shell scripts


bergzand:
shellcheck is an application to statically check shell scripts for common errors. Most of these consists of small issues with the shell scripts where the problem might only appear in certain edge cases.

As an example, one of the issues is that shellcheck warns for is not having quotes around `${VAR}`. This issue leads in the headerguards check to vague errors as soon as there are spaces in `$PWD`.

I think this could also be a useful addition to the CI to check the shell scripts for common mistakes.

[1]: https://github.com/koalaman/shellcheck

gebart:
I like, more static tests 👍

aabadie:
Indeed, very nice. Would be great to have something similar for Makefiles ;)

Lotterleben:
Since both PRs are merged, can this issue be closed?

PeterKietzmann:
It seems so. @bergzand please reopen if I'm mistaken

PR #9056, opened 2018-05-01
tools/flake8: fix shellcheck issues

bergzand:
### Contribution description

Fix a number of issues with the `dist/tools/flake8/check.sh` script as indicated by shellcheck.

mainly quoting issues to prevent word splitting/globbing and a few cases of variables in printf statements ([SC2059](https://github.com/koalaman/shellcheck/wiki/SC2059))

### Issues/PRs references

None

### DIFF
--- a/dist/tools/flake8/check.sh
+++ b/dist/tools/flake8/check.sh
@@ -9,9 +9,9 @@
 
 FLAKE8_CMD="python3 -m flake8"
 
-if tput colors &> /dev/null && [ $(tput colors) -ge 8 ]; then
-    CERROR="\e[1;31m"
-    CRESET="\e[0m"
+if tput colors &> /dev/null && [ "$(tput colors)" -ge 8 ]; then
+    CERROR=$'\033[1;31m'
+    CRESET=$'\033[0m'
 else
     CERROR=
     CRESET=
@@ -19,7 +19,7 @@ fi
 
 DIST_TOOLS=${RIOTBASE:-.}/dist/tools
 
-. ${DIST_TOOLS}/ci/changed_files.sh
+. "${DIST_TOOLS}/ci/changed_files.sh"
 
 FILES=$(FILEREGEX='(?=*.py$|pyterm$)' changed_files)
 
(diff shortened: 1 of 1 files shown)


dist/tools/flake8/check.sh
```diff
@@ -9,17 +9,17 @@
...
+    CRESET=$'\033[0m'
 else
     CERROR=
     CRESET=
 fi
 
 DIST_TOOLS=${RIOTBASE:-.}/dist/tools
 
-. ${DIST_TOOLS}/ci/changed_files.sh
+. "${DIST_TOOLS}/ci/changed_files.sh"
```
aabadie: shellcheck still reports a warning with this line. Maybe add this comment on the line before to remove it:
```sh
# shellcheck source=/dev/null
```
See [here](https://github.com/koalaman/shellcheck/wiki/SC1090) for information

aabadie:
BTW, what about adding a shellcheck check script for the CI ?

dist/tools/flake8/check.sh
aabadie: Or use 
```sh
# shellcheck disable=SC1090
```
?

bergzand: I'd rather disable this check in a future CI script, similar to the flake8.conf. This to prevent having to add an exception to almost every script

bergzand:
> BTW, what about adding a shellcheck check script for the CI ?

#8212 (No progress from my side so far)

dist/tools/flake8/check.sh
```diff
@@ -29,16 +29,16 @@ then
 fi
 
 ${FLAKE8_CMD} --version &> /dev/null || {
-    printf "${CERROR}$0: cannot execute \"${FLAKE8_CMD}\"!${CRESET}\n"
+    printf "%s%s: cannot execute \"%s\"!%s\n" "${CERROR}" "$0" "${FLAKE8_CMD}" "${CRESET}"
     exit 1
 }
 
-ERRORS=$(${FLAKE8_CMD} --config=${DIST_TOOLS}/flake8/flake8.cfg ${FILES})
+ERRORS=$(${FLAKE8_CMD} --config="${DIST_TOOLS}/flake8/flake8.cfg" "${FILES}")
```
bergzand: Quoting `${FILES}` here is an issue since it is a space separated list of files. Quoting it changes it to a single argument here. The proper way to solve this would be to use a bash array, but this would restrict these scripts from POSIX shell to bash as interpreter.

aabadie: > this would restrict these scripts from POSIX shell to bash as interpreter.

Is this really a problem since the script is explicitly using bash (see first line) ?

bergzand: > Is this really a problem since the script is explicitly using bash (see first line) ?

Unless somebody really wants to keep it POSIX shell compliant, no. But quoting `FILES` here is going to break things now.

aabadie: Then let's remove the quotes for now. The shellcheck script is not yet in the check scripts.

bergzand: Done

bergzand:
squashed!

PR #9070, opened 2018-05-04
tools/check: use RIOTTOOLS variable

cladmi:
### Contribution description

Use RIOTTOOLS variable for tools/check

Follow up to #9067 and part of #8821

### Testing

I ran `make static-tests` manually and also tests scripts manually from the RIOT directory.

### Issues/PRs references

Follow up to #9067 and part of #8821

### DIFF
--- a/dist/tools/coccinelle/check.sh
+++ b/dist/tools/coccinelle/check.sh
@@ -6,9 +6,8 @@
 # General Public License v2.1. See the file LICENSE in the top level
 # directory for more details.
 
-: ${RIOTBASE:=$(pwd)}
-
-. ${RIOTBASE}/dist/tools/ci/changed_files.sh
+: ${RIOTTOOLS:=./dist/tools}
+. ${RIOTTOOLS}/ci/changed_files.sh
 
 EXIT_CODE=0
 
@@ -26,7 +25,7 @@ indent() {
 
 coccinelle_checkone() {
     OUT="$(spatch --very-quiet \
-        --macro-file-builtins ${RIOTBASE}/dist/tools/coccinelle/include/riot-standard.h \
+        --macro-file-builtins ${RIOTTOOLS}/coccinelle/include/riot-standard.h \
         --sp-file $patch ${FILES} | filter)"
 
     if [ -n "$OUT" ]; then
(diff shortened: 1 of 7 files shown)


cladmi:
Sure I will do and test with your PR.

bergzand:
Thanks!

cladmi:
I changed the lines modifying RIOTTOOLS but not the ones with `OUT="$(...)"` because not really sure how to fix them.

cladmi:
I see some problems in the `COCCINELLE_WARNONLY` variable handling in the script… Like it's not checked and the second line  setting `COCCINELLE_WARNONLY` should not be there I think.

dist/tools/coccinelle/check.sh
```diff
@@ -6,9 +6,8 @@
 # General Public License v2.1. See the file LICENSE in the top level
 # directory for more details.
 
-: ${RIOTBASE:=$(pwd)}
-
-. ${RIOTBASE}/dist/tools/ci/changed_files.sh
+: "${RIOTTOOLS:=./dist/tools}"
```
smlng: shouldn't this better be `: "${RIOTTOOLS:=$(pwd)/dist/tools}"`

dist/tools/cppcheck/check.sh
```diff
@@ -8,7 +8,8 @@
 # General Public License v2.1. See the file LICENSE in the top level
 # directory for more details.
 
-. ${RIOTBASE:+${RIOTBASE}/}dist/tools/ci/changed_files.sh
+: "${RIOTTOOLS:=./dist/tools}"
```
smlng: dito, see above

dist/tools/externc/check.sh
```diff
@@ -8,7 +8,8 @@
 # General Public License v2.1. See the file LICENSE in the top level
 # directory for more details.
 
-. ${RIOTBASE:+${RIOTBASE}/}dist/tools/ci/changed_files.sh
+: "${RIOTTOOLS:=./dist/tools}"
```
smlng: dito

dist/tools/coccinelle/check.sh
cladmi: Some of the files had './dist/tools' before, but putting a safe default value as example is better.
One could use a 'cd' in one command and only fail when executed without 'RIOTTOOLS' set.

cladmi:
Changes addressed, I also removed the wrongly added 'shellcheck' script.

kaspar030:
@smlng you good here?

PR #9075, opened 2018-05-04
 sens0: Simple sensor device interface 

gebart:
### Contribution description

Provide an easy way to interface new sensors without writing a lot of boilerplate code.
This driver provides a basic polled interface for digital sensors, currently only I2C is supported, but it will be a small effort to add SPI support later.

The output format from the sens0 read API was chosen as `phydat_t`, which makes this a good match for using with SAUL. 

An example application is provided in tests/driver_sens0. The example application is configured for the onboard accelerometer on FRDM-K64F, and an optional external FRDM-STBC-AGM01 IMU 9DOF shield.

This driver is meant to only provide a basic polled read functionality of digital sensors and will not support interrupts, FIFOs, caching, or other more advanced features, which are usually more difficult to generalize into a generic interface like this.

#### Background

Three important observations:

- Many sensing applications only need the basic functionality of the sensor, i.e. polling for a measurement reading every X ms and then doing something with that value in software.
- Many digital sensors follow this operating procedure: Set up sensor parameters (sample rate, resolution, full scale range etc.), read measurements from output registers.
- Many digital sensors support both I2C or SPI communication on the same chip, but the hardware register map and operating procedures are the same, or at least closely similar, regardless whether the register read/write operations happen over SPI or I2C.

Interfacing with a basic digital sensor in RIOT today requires writing or copying a whole bunch of boilerplate code, just to get some readings from the sensor. The Sens0 driver is meant to eliminate as much as possible of the boilerplate and leaving the developer to only create a simple hardware specification, which contains information about what registers to write during init, where to find the output, and how to parse the raw output into something usable. This will obviously limit the functionality compared to writing a specific driver from the ground up, but it also takes a lot less time to configure.

As an example, I wrote the fully working FXAS21002C hardware interface specification in some 15 minutes, without any prior experience with that particular sensor.

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -298,6 +298,16 @@ ifneq (,$(filter sdcard_spi,$(USEMODULE)))
   USEMODULE += xtimer
 endif
 
+ifneq (,$(filter sens0,$(USEMODULE)))
+  USEMODULE += sens0_specs
+endif
+
+ifneq (,$(filter sens0_i2c,$(USEMODULE)))
+  USEMODULE += sens0
+  FEATURES_REQUIRED += periph_i2c
+  USEMODULE += xtimer
+endif
+
 ifneq (,$(filter sht11,$(USEMODULE)))
   USEMODULE += xtimer
 endif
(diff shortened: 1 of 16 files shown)


gebart:
I have prepared an auto init implementation which uses xfa (#7523) for the device configuration instances, but let's merge this without auto init first to ease review.

PR #9077, opened 2018-05-04
enc28j60: improvements to fix #9043

### LINKED ISSUE #9043: enc28j60: packet tranmission problem

gschorcht:
This pull request tries to improve the stability of ```enc28j60``` driver as following:

1. Problem reported in #9043 was that sometimes the transmission logic seems to hang. The transmission that was started in```nd_send``` function before, is neither finished nor aborted and the ```ECON1.TXRTS``` flag keeps set all the time. The only way to abort the hanging transmission and to send a new packet in ```nd_send``` function is to reset the transmission logic. The fix (workaround) tries to solve this problem by saving the system time in ```dev->tx_time``` once a transmission is started. It then checks the ```ECON1.TXRTS``` flag when the ```nd_send``` function is executed next time. If the ```ECON1.TXRTS``` flag is still set, it checks the time that has passed since ```dev->tx_time```. If it is greater than 2.4287 ms, it supposes that the transmission logic hangs and resets it to send the packet. Otherwise it simply returns with ```EBUSY``` error code to indicate a busy device.

2. The errata sheet section 14 reports that even values in ```ERXRDPT``` may corrupt receive buffer as well as the next packet pointer. ```ERXRDPT``` need to be set always at odd addresses. The fix defines two macros ```NEXT_TO_ERXRDPT``` and ```ERXRDPT_TO_NEXT``` to implement a workaround based on ```BUF_RX_END``` as suggested in the errata sheet. For that purpose ```BUF_RX_END``` has to point to the last byte in the recieve buffer to odd address. Therefore ```BUF_*_END``` macros are defined to odd addresses and always point to the last byte.

3. Received data length should never result into negative values. Therefore, I have changed ```size_t size``` to ```uint16_t size```. This might be wrong since the return value and the parameter ```max_len``` are both signed. However, I had a crash, where ```size``` from ```head``` resulted into -4 which was less than the ```max_len``` and the following read receive buffer were executed.

### DIFF
--- a/drivers/enc28j60/enc28j60.c
+++ b/drivers/enc28j60/enc28j60.c
@@ -70,10 +70,18 @@
  * RX must start at buffer address 0x0000 (see errata sheet, section 5)
  */
 #define BUF_TX_START                ((BUFFER_SIZE / 4) * 3)
-#define BUF_TX_END                  (BUFFER_SIZE - 2)
+#define BUF_TX_END                  (BUFFER_SIZE - 1)
 #define BUF_RX_START                (0)
-#define BUF_RX_END                  (BUF_TX_START - 2)
+#define BUF_RX_END                  (BUF_TX_START - 1)
 
+/**
+ * @brief  Maximum transmission time
+ *
+ * The time in us that is required to send an Ethernet frame of maximum length
+ * (Preamle + SFD + 1518 byte) at 10 Mbps in full duplex mode with a guard
+ * period of 9,6 us. This time is used as time out for send operations.
+ */
+#define MAX_TX_TIME                 (1230U)
 
 static void switch_bank(enc28j60_t *dev, int8_t bank)
 {
@@ -248,6 +256,28 @@ static int nd_send(netdev_t *netdev, const iolist_t *iolist)
 
     mutex_lock(&dev->devlock);
 
+    if (cmd_rcr(dev, REG_ECON1, -1) & ECON1_TXRTS) {
+        /* there is already a transmission in progress */
+        if (xtimer_now_usec() - dev->tx_time > MAX_TX_TIME * 2) {
+            /*
+             * if transmission time exceeds the double of maximum transmission
+             * time, we suppose that TX logic hangs and has to be reset
+             */
+            cmd_bfs(dev, REG_ECON1, -1, ECON1_TXRST);
+            cmd_bfc(dev, REG_ECON1, -1, ECON1_TXRST);
+        }
+        else {
+            /*
+             * otherwise we suppose that the transmission is still in progress
+             * and return EBUSY
+             */
+            mutex_unlock(&dev->devlock);
+            return -EBUSY;
+        }
+    }
+
+    /* set TX start pointer */
+    cmd_w_addr(dev, ADDR_TX_START, BUF_TX_START);
     /* set write pointer */
     cmd_w_addr(dev, ADDR_WRITE_PTR, BUF_TX_START);
     /* write control byte and the actual data into the buffer */
@@ -260,6 +290,8 @@ static int nd_send(netdev_t *netdev, const iolist_t *iolist)
     cmd_w_addr(dev, ADDR_TX_END, cmd_r_addr(dev, ADDR_WRITE_PTR) - 1);
     /* trigger the send process */
     cmd_bfs(dev, REG_ECON1, -1, ECON1_TXRTS);
+    /* set last transmission time for timeout handling */
+    dev->tx_time = xtimer_now_usec();
 
 #ifdef MODULE_NETSTATS_L2
     netdev->stats.tx_bytes += c;
@@ -269,23 +301,34 @@ static int nd_send(netdev_t *netdev, const iolist_t *iolist)
     return c;
 }
 
+/*
+ * Section 14 of errata sheet: Even values in ERXRDPT may corrupt receive
+ * buffer as well as the next packet pointer. ERXRDPT need to be set always
+ * at odd addresses. Following macros determine odd ERXRDPT from next packet
+ * pointer and vise versa. Next packet pointer is always at even address
+ * because of hardware padding.
+ */
+#define NEXT_TO_ERXRDPT(n) ((n == BUF_RX_START || n - 1 > BUF_RX_END) ? BUF_RX_END : n - 1)
+#define ERXRDPT_TO_NEXT(e) ((e >= BUF_RX_END) ? BUF_RX_START : e + 1)
+
 static int nd_recv(netdev_t *netdev, void *buf, size_t max_len, void *info)
 {
     enc28j60_t *dev = (enc28j60_t *)netdev;
     uint8_t head[6];
-    size_t size;
+    uint16_t size;
     uint16_t next;
 
     (void)info;
     mutex_lock(&dev->devlock);
 
     /* set read pointer to RX read address */
-    cmd_w_addr(dev, ADDR_READ_PTR, cmd_r_addr(dev, ADDR_RX_READ));
+    uint16_t rx_rd_ptr = cmd_r_addr(dev, ADDR_RX_READ);
+    cmd_w_addr(dev, ADDR_READ_PTR, ERXRDPT_TO_NEXT(rx_rd_ptr));
     /* read packet header */
     cmd_rbm(dev, head, 6);
     /* TODO: care for endianess */
     next = (uint16_t)((head[1] << 8) | head[0]);
-    size = (size_t)((head[3] << 8) | head[2]) - 4;  /* discard CRC */
+    size = (uint16_t)((head[3] << 8) | head[2]) - 4;  /* discard CRC */
 
     if (buf != NULL) {
 #ifdef MODULE_NETSTATS_L2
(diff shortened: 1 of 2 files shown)


drivers/enc28j60/enc28j60.c
```diff
@@ -248,10 +248,27 @@ static int nd_send(netdev_t *netdev, const iolist_t *iolist)
 
     mutex_lock(&dev->devlock);
 
-#ifdef MODULE_NETSTATS_L2
-    netdev->stats.tx_bytes += count;
-#endif
+    if (cmd_rcr(dev, REG_ECON1, -1) & ECON1_TXRTS)
+    {
```
smlng: please adapt to RIOT coding style, see [wiki](https://github.com/RIOT-OS/RIOT/wiki/Coding-conventions)

drivers/enc28j60/enc28j60.c
```diff
@@ -248,10 +248,27 @@ static int nd_send(netdev_t *netdev, const iolist_t *iolist)
...
+        /* there is already a transmission in progress */
+        if (xtimer_now_usec() - dev->tx_time > 2429)
```
smlng: please use macros or constants (with proper documentation) instead of magic number like `2429`

drivers/enc28j60/enc28j60.c
```diff
@@ -248,10 +248,27 @@ static int nd_send(netdev_t *netdev, const iolist_t *iolist)
...
+        {
+            /* if transmission time exceeds 2.4287 ms, we suppose that
```
smlng: ah here is some docu, still would prefer to use macro/constant

drivers/enc28j60/enc28j60.c
gschorcht: I guess you would like to have the opening curly brace at the end of the line according to the old ANSI C coding style :-( Personally, I prefere
```
if (condition) 
{
   do_something;
}
else
{
  do_something_else;
}
```
over 
```
if (condition)  {
   do_something;
} else {
  do_something_else;
}
```
because the structure seems to be more clear and much better readable from my point of view, especially in larger code.

smlng: to be blunt: that's why RIOT (as every other larger project) has coding conventions for consistent style and to avoid re-occuring discussions on that matter.

drivers/enc28j60/enc28j60.c
gschorcht: The problem was, that I had no idea how to name this timeout since it is not defined as a parameter in the IEEE802.3 standard. Therefore, I simply used the value as defined in the errata sheet even though they defined this timeout for half duplex mode. In switched environments with full-duplex this time should be much smaller. The transmission of an Ethernet frame with maximum length of 1526 byte (including the 8 bytes preamble and start frame delimiter) should not take longer than 1.2208 ms at 10 Mbps. So, the magic time defined in the errata sheet seems to be the double of the maximum packet transmission time with some additional guard period between the transmissions.

I'm not sure whether RIOT is able at all to send a second packet within these 1.2208 ms. If not, we could remove the timeout guard completely. In that case, it would be enough to test the ```ECON1_TXRTS``` flag before next packet is sent and to reset the transmission logic, if the previous transmission has not been finished.

drivers/enc28j60/enc28j60.c
gschorcht: Sure, I can introduce a macro/constant definition.

gschorcht:
@sming I made the changes as requested.

tcschmidt:
@smlng ping: there was a typo (username) in the note of @gschorcht

drivers/enc28j60/enc28j60.c
```diff
@@ -260,6 +286,12 @@ static int nd_send(netdev_t *netdev, const iolist_t *iolist)
     cmd_w_addr(dev, ADDR_TX_END, cmd_r_addr(dev, ADDR_WRITE_PTR) - 1);
     /* trigger the send process */
     cmd_bfs(dev, REG_ECON1, -1, ECON1_TXRTS);
+    /* set last transmission time for timeout handling */
+    dev->tx_time = xtimer_now_usec();
+
+#ifdef MODULE_NETSTATS_L2
+    netdev->stats.tx_bytes += c;
```
smlng: IMHO no need to do this twice 😉 see below

gschorcht: @sming Sure, I don't know how this could happen :-(

gschorcht:
@sming I have changed as requested.

tcschmidt:
@gschorcht repeated typo in username - use @smlng

gschorcht:
@smlng I have changed as requested.

gschorcht:
@sming Is there anything else required for the merge?

gschorcht:
@smlng Is there anything else required for the merge? The status is still 'requested changes'.

drivers/enc28j60/enc28j60.c
```diff
@@ -70,10 +70,18 @@
...
+#define BUF_RX_END                  (BUF_TX_START - 1)
 
+/**
+ * @brief  Maximum transmission time
+ *
+ * The time in us that is required to send an Ethernet frame of maximum length
+ * (Preamle + SFD + 1518 byte) at 10 Mbps in full duplex mode with a guard
+ * period of 9,6 us. This time is used as time out for send operations.
+ */
+#define MAX_TX_TIME                 1230
```
smlng: please use parentheses, i.e. `#define MAX_TX_TIME                 (1230U)`

drivers/enc28j60/enc28j60.c
```diff
@@ -248,6 +256,24 @@ static int nd_send(netdev_t *netdev, const iolist_t *iolist)
 
     mutex_lock(&dev->devlock);
 
+    if (cmd_rcr(dev, REG_ECON1, -1) & ECON1_TXRTS) {
+        /* there is already a transmission in progress */
+        if (xtimer_now_usec() - dev->tx_time > MAX_TX_TIME * 2) {
+            /* if transmission time exceeds the double of maximum transmission
+               time, we suppose that TX logic hangs and has to be reset */
```
smlng: here and below: please uses `*` at the beginning of subsequent lines in multi-line comments

PeterKietzmann:
I have the hardware on my desk (connected to a samr21-xpro). @gschorcht could you add a short description how to reproduce the bug so I can easily verify that the PR is a fix? @smlng you already provided the review, would complete it by testing? I'll give you the hardware then...

PeterKietzmann:
What I usually do is using `gnrc_networking` with two network interfaces. For the hardware that I'll give you, add this to your application Makefile:
```
ifneq (,$(filter samr21-xpro,$(BOARD)))
    USEMODULE += enc28j60
    GNRC_NETIF_NUMOF := 2

    ENC_SPI ?= SPI_DEV\(1\)
    ENC_CS  ?= GPIO_PIN\(0,18\)
    ENC_INT ?= GPIO_PIN\(0,28\)
    ENC_RST ?= GPIO_PIN\(0,13\)

    # export SPI and pins
    CFLAGS += -DENC28J60_PARAM_SPI=$(ENC_SPI)
    CFLAGS += -DENC28J60_PARAM_CS=$(ENC_CS)
    CFLAGS += -DENC28J60_PARAM_INT=$(ENC_INT)
    CFLAGS += -DENC28J60_PARAM_RESET=$(ENC_RST)
endif
```

gschorcht:
@PeterKietzmann @smlng 
Problem 1: Unfortunately, the problem described was a sporadic problem. It happened sometimes directly after reset, sometimes after some minutes, but sometimes after one day and or never. I was running into the problem when I played with enc28j60 and my esp8266 RIOT port by pinging an arbitrary node in the LAN. So I would just suggest just to let the node ping any node in your LAN for an endless time. As I remember it should be enough to ping every second.

Problem 2: Errata sheet says "The receive hardware may corrupt the circular receive buffer (including the Next Packet Pointer and receive status vector fields) when an even value is programmed into the ERXRDPTH:ERXRDPTL registers." It seems also to be a sporadic problem. When I was tried to fix problem 1, I implemented the workaround as described there. I would suggest to do a stress test by sending a lot of packets with an arbitrary length as fast as you can from another node to the node with enc28j60.

Problem 3: Should be covered by testing problem 2.

smlng:
Tested ACK, please squash

smlng:
(minor) downside: increases binary size by approx. 150B using samr21-xpro

PeterKietzmann:
Please fix:
```
enc28j60.c: In function 'nd_recv':
enc28j60.c:346:9: error: comparison of unsigned expression < 0 is always false [-Werror=type-limits]
```

gschorcht:
@PeterKietzmann Yes, I already saw it and I'm working on it. I'm just wondering why I don't get the error on my native platform while murdock gives this error.

gschorcht:
@PeterKietzmann Done

PeterKietzmann:
thanks!

PR #9079, opened 2018-05-04
cpu/stm32f4: add support for STM32F423 line

vincent-d:
### Contribution description

This adds support for stm32f423 line and fix stm32f413 (all part numbers now supported).

### Issues/PRs references

None


kaspar030:
@kYc0o @aabadie do you have a board to test?

kYc0o:
Unfortunately I have no board to test this...

vincent-d:
AFAIK, there is no board on the market with an stm32f423. But we need this for our developments.

I don't know how we can proceed for cpu variants which have no boards, but it doesn't make sense to maintain this in a fork IMO.
Maintaining a board outside of RIOT repo is easy, but there is no way of doing this for a cpu. And I think it would not be valuable to be able to do it anyway.

kYc0o:
Actually this PR doesn't hurt, so it could be merged. However the commit for info-stm32 it's a bit out of context... But it doesn't hurt either.

I'm not against merging this as-is, so if @kaspar030 agrees we can merge it.

PR #9080, opened 2018-05-04
newlib-nano: add include directory

toonst:
### Contribution description
In libnewlib-arm-none-eabi version 2.4.0.20160527-3 the include directory is called nano iso newlib-nano. This fix checks for the existence of either directories and adds it to the include path.

### DIFF
--- a/makefiles/libc/newlib.mk
+++ b/makefiles/libc/newlib.mk
@@ -64,7 +64,8 @@ ifeq ($(TOOLCHAIN),llvm)
 endif
 
 ifeq (1,$(USE_NEWLIB_NANO))
-  NEWLIB_NANO_INCLUDE_DIR ?= $(NEWLIB_INCLUDE_DIR)/newlib-nano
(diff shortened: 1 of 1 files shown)


tcschmidt:
@gebart @kaspar030 ready for approval?

kYc0o:
I'm sorry to come until now, but this PR broke compilation for ARM platforms on OS X. I'll investigate for a solution.

kYc0o:
@smlng @emmanuelsearch  can you verify just to check if it's not only my local setup?

jcarrano:
This PR broke compilation in ArchLinux. As far as I can tell, it's not really the fault of this change, but rather the newlib package in Arch is not installing the nano directory properly (see https://bugs.archlinux.org/task/50481 reported by someone at OTAkeys). @kYc0o Maybe in OSX there is a similar problem.

As a workaround, can we fall back to what we were doing previousl if neither the "nano" or "newlib-nano" directories are found?

kYc0o:
@jcarrano please report your output in #9214 .

PR #8062, opened 2017-11-16
make: add board grouping

kaspar030:
This PR adds infrastructure to group boards (e.g., by architecture, by bit-width, ...).

Another commit shows how this can be used within application makefiles for black/whitelisting.

PR #8292, opened 2017-12-19
makefiles/modules.inc.mk: remove setting deprecated FEATURE_PERIPH_ macros

cladmi:
This PR removes setting the `FEATURES_PERIPH_` macros as they are completely unused and said deprecated.

There will be no reference after https://github.com/RIOT-OS/RIOT/pull/8227 which was even set manually.
Was cleaned up also by:  https://github.com/RIOT-OS/RIOT/pull/8226

PR #9081, opened 2018-05-04
make: add architecture features and feature blacklisting

kaspar030:
### Contribution description

This PR adds negative feature dependencies (feature blacklisting), allowing a test or module to specify that it will not work if a certain feature is provided.

This is then used to specify architectures or some of their traits as features, and as a next step, use those to blacklist tests.

Previously, if a module or test did not work on e.g., <32bit platforms or with the msp430 toolchain, all boards had to be specified in ```BOARD_BLACKLIST```, and that list needed to be maintained.

With this PR it is possible to blacklist a whole architecture using e.g., ```FEATURES_BLACKLIST += arch_avr8 arch_msp430```.
*edit* Also, it is now possible to do e.g., ```FEATURES_REQUIRED += arch_32bit```.

### Issues/PRs references

#8062


kaspar030:
(accidently added some leftover test folders, fixing...)

Makefile.include
```diff
@@ -514,6 +514,9 @@ ifneq (, $(filter all, $(if $(MAKECMDGOALS), $(MAKECMDGOALS), all)))
     EXPECT_ERRORS := 1
   endif
 
+  # turn provided but blacklisted features into required "!<feature>"
+  FEATURES_REQUIRED += $(addprefix !,$(sort $(filter $(FEATURES_PROVIDED), $(FEATURES_BLACKLIST))))
```
cladmi: I would prefer adding another variable name and/or other if case for this check.
These sections are hard enough to read, so adding an explicit name simplify this.

makefiles/info-global.inc.mk
```diff
@@ -27,6 +29,7 @@ define board_missing_features
   include $(RIOTBASE)/Makefile.dep
 
   FEATURES_MISSING := $$(sort $$(filter-out $$(FEATURES_PROVIDED), $$(FEATURES_REQUIRED)))
+  FEATURES_MISSING += $$(addprefix !,$$(sort $$(filter $$(FEATURES_PROVIDED), $$(FEATURES_BLACKLIST))))
```
cladmi: Same here I would prefer adding another variable name and/or other if case for this check.
These sections are hard enough to read, so adding an explicit name simplify this.

Makefile.include
kaspar030: You mean, instead of directly adding the blacklisted but available features to FEATURES_REQUIRED, create an intermediate FEATURES_BLACKLISTED and use that in the if below?

I thought about that, but decided to rather go with a nice comment... I think ```warning: feature "!arch_16bit" missing, expect errors``` pretty much explains what is happening, and extra variables clutter code.

Do you insist?

cladmi:
I found the problem why the naming, features re-usage and definition did not fit me. The lack of clarity was hiding the problem. I think it is currently incorrectly using the feature concepts.

The `arch` definitions and the blacklisting are done on provided features, and not on `used` features which is for me wrong.

### From my understanding:

`FEATURES_PROVIDED` are hardware (including toolchain) features that are available for this board. It can require having code to make them available to the application. In that case, it means first enabling this feature by having it put in `FEATURES_USED`, which currently happens only if it is mentioned in `FEATURES_REQUIRED` or `FEATURES_OPTIONAL`.

Then the build system only process them if they are in `FEATURES_USED`.
It was even reinforced when changing the `FEATURES_CONFLICT` handling: https://github.com/RIOT-OS/RIOT/pull/8925 as there was no reason to care about disabled provided features.

In the current implementation these `arch_%` features will almost never be in `FEATURES_REQUIRED` and still be always *enabled* because of course, the architecture is 16bit.

If it is required to have this information available in  the code, like for examples in tests where it would be easier to use a macro than a `sizeof(int) == 2`, it would somehow require re-exporting these enabled features. As done before https://github.com/RIOT-OS/RIOT/pull/8292

However, it means only exporting *used* features, not the available but unused ones.

Also, if for any reason, some module needs to blacklist against another feature than `arch_` it would complained about disabled features as in https://github.com/RIOT-OS/RIOT/pull/8925.

### Ideas

Re-using the concept of FEATURES is I think adapted here, but not as provided ones.
I would add these `arch_` ones directly to a used/enabled features list, or a new adapted name and not in the provided list.

cladmi:
The problem is even more hidden because both feature blacklisting, which should be global to all features, and adding arch features was done in the same pr.

kaspar030:
> I found the problem why the naming, features re-usage and definition did not fit me. The lack of clarity was hiding the problem. I think it is currently incorrectly using the feature concepts.

I don't see a problem.

FEATURES_PROVIDED -> "the platform provides a feature"
FEATURES_REQUIRED -> "this feature must be provided"
FEATURES_OPTIONAL -> "this feature will be used if available"
FEATURES_USED -> "list of all required or optional and provided features"

The latter two are used for periph_* in order to select periph_% modules when appropriate.

This PR adds:

FEATURES_BLACKLIST -> "this feature must not be provided by the platform" 

In order to re-use fiddly make loops, each feature in FEATURES_BLACKLIST gets added to FEATRUES_REQUIRED with prefixed "!" as "!<feature_name>".

> The arch definitions and the blacklisting are done on provided features, and not on used features which is for me wrong.

Some features are provided and can lead to code linked in when used, some are "always there". It is a matter of definition.
We can argue about the naming, or the meaning, or definition, or personal feelings, or whatever, forever. Or we can add two lines of dokumentation and take a useful improvement.

kaspar030:
Huh, why did I close this?

> But if for any reason, you want to do an app/library that should not work if, for example periph_pm is not included. Then the blacklisting as it is now would prevent using it if the board/cpu as pm support even if not activated and compiled.

If an app should *not work* if periph_pm is *not included*, that turns into it should *only work* if it is *included*. Thus periph_pm should be a required feature.

If you meant, an app or library should *only work* if periph_pm is not included (even though it is available as a feature), then this is out of scope for features, and should be handled by the module system. periph_pm is both a module and a feature. If actually used, the feature turns into a used module.

cladmi:
> If you meant, an app or library should only work  if periph_pm is not included (even though it is available as a  feature), then this is out of scope for features, and should be handled  by the module system. periph_pm is both a module and a feature. If  actually used, the feature turns into a used module.

As said, I personally understand that BLACKLIST should be applied to used features not on available ones, like FEATURES_CONFLICT.


As we have different understanding on the semantic, I would like to have other point of view on this. It is an API design, where there are subjective choices, so having other developers feedback could help.

cladmi:
I will also do a documentation PR to summarize what was said here on the existing FEATURES variables.

jcarrano:
I agree with @cladmi in that the filtering should be made based on the features that are USED and not only PROVIDED.

> FEATURES_BLACKLIST -> "this feature must not be provided by the platform"

I'd say `FEATURES_BLACKLIST -> "the current build should not have this feature enabled"`

The architecture is always USED. so `arch_*` should be in USED. Now, how come a feature is used when it is not provided? Therefore, for consistency, it should also be PROVIDED (possibly by the same package).

I find it kind of weird that some module can require to *not* have a feature. It is a bit misleading: how can something stop working because a feature is added. In fact, I cannot think of an example where this is used that is not for covering a bug. This is not bad in itself, software has bugs and there should be a way to declare "x does not work with y", so we need to be able to do this.

IMO, the issue with "!" is mixing the interface with the implementation. The interface is how the user tells the system "this feature should/should not be here". The implementation is the mechanics of how these settings are stored an operated with. I favour a straightforward implementation, where things are named by what they are, and if there is a need for a "nice" interface where the used defines one variable that contains everything, it still can be done.

kaspar030:
> The `arch` definitions and the blacklisting are done on provided features, and not on `used` features which is for me wrong.

I changed this to blacklist based on *used* features, and added that all provided arch_* features are always used.

kaspar030:
@cladmi @jcarrano would you take another look?

kaspar030:
I've rebased this and dropped some of the commits that put the arch_features in use, as they get stale quickly (IMO showing the need for this PR).

cladmi:
I would indeed prefer a separate line for the error message. But I could fix it on my side afterwards if it is really a problem. I am moving the `FEATURES_` variables definitions in `Makefile.features` to remove the duplication.

When you see "feature missing negation something" does it means it is not missing? And better understand it directly if it needs 2 lines of code than thinking every time.

I would prefer to split the `FEATURES_BLACKLIST` in a separate PR as it deserves its own reference for the future instead of mixing it with the architectures definition and all.

I agree with automatically including the `arch_` features.

Makefile.include
kaspar030: addressed

kaspar030:
I migrated this to the improved Makefile.features. I think all comments have been addressed but this:

> I would prefer to split the `FEATURES_BLACKLIST` in a separate PR as it deserves its own reference for the future instead of mixing it with the architectures definition and all.

We split PR's for easier reviewing, not for references. For that we use commits.

@cladmi care to take another look? This is relevant for #12406.

smlng:
I like this, just one comment: to me it would make sense to change `arch_{8|16|32}bit` to `arch_bit_{8|16|32}` to allow for generic filtering in features if we (for what ever reason) want to get the arch bits during a build.

smlng:
please squash if ready

smlng:
partly tested, don't know about some corner cases.

cpu/arm7_common/Makefile.features
```diff
@@ -1,2 +1,5 @@
 FEATURES_PROVIDED += cpp
 FEATURES_PROVIDED += periph_pm
+FEATURES_PROVIDED += arch_arm7
```
aabadie: Please keep alphabetical order, here and in other modified places below.

smlng: good point

fjmolinas:
> `FEATURES_BLACKLIST_GLOBAL` should also be part of `info-build`.

Sorry I think it is actually `FEATURES_BLACKLIST` here.

cpu/arm7_common/Makefile.features
kaspar030: fixed

kaspar030:
> Sorry I think it is actually `FEATURES_BLACKLIST` here.

addressed

cpu/esp32/Makefile.features
```diff
@@ -1,5 +1,8 @@
 # Features that are provided by the CPU independent on the board
 FEATURES_PROVIDED += cpp
+# Features that are provided by CPU independent on the board
```
aabadie: There is a problem with this comment and the alphabetical order. `cpp` should be after `esp32` I think.

kaspar030: they are in different lists.

aabadie: `cpp` is in the same FEATURES_PROVIDED list as arch_xx. See arm7_common features list.

makefiles/info.inc.mk
```diff
@@ -60,6 +60,8 @@ info-build:
 	@echo '         $(or $(sort $(FEATURES_PROVIDED)), -none-)'
 	@echo 'FEATURES_MISSING (only non optional features):'
 	@echo '         $(or $(FEATURES_MISSING), -none-)'
+	@echo 'FEATURES_BLACKLIST (blacklisted features):'
+	@echo '         $(or $(FEATURES_BLACKLIST), -none-)'
```
aabadie:
```suggestion
	@echo '         $(or $(sort $(FEATURES_BLACKLIST)), -none-)'
```

makefiles/info-global.inc.mk
```diff
@@ -9,6 +9,8 @@ FEATURES_CONFLICT_GLOBAL := $(FEATURES_CONFLICT)
 FEATURES_CONFLICT_MSG_GLOBAL := $(FEATURES_MSG_CONFLICT)
 DISABLE_MODULE_GLOBAL := $(DISABLE_MODULE)
 DEFAULT_MODULE_GLOBAL := $(DEFAULT_MODULE)
+FEATURES_BLACKLIST_GLOBAL := $(FEATURES_BLACKLIST)
+DISABLE_MODULE_GLOBAL := $(DISABLE_MODULE_GLOBAL)
```
aabadie: This line seems invalid and DISABLE_MODULE_GLOBAL is already set 3 lines above.

makefiles/info.inc.mk
kaspar030: ok

cpu/esp32/Makefile.features
kaspar030: moved arch features up to CPU defined features

makefiles/info-global.inc.mk
kaspar030: removed

cpu/esp32/Makefile.features
```diff
@@ -1,5 +1,8 @@
 # Features that are provided by the CPU independent on the board
+FEATURES_PROVIDED += arch_32bit
+FEATURES_PROVIDED += arch_esp32
 FEATURES_PROVIDED += cpp
+# Features that are provided by CPU independent on the board
```
aabadie: I think this comment could be removed. It's almost the same as the one line 1.

kaspar030: maybe, but that would be an unrelated change.

aabadie: This is an addition from this PR, so not something that is changed. Please remove.

kaspar030: removed

Makefile.features
```diff
@@ -22,9 +22,11 @@ endif
 # Resolve FEATURES_ variables
 # Their value will only be complete after resolving dependencies
 
-# Features that are required by the application but not provided by the BSP
+# Features that are required by the application but not provided by the BSP and
+# features that are used but blacklisted (prepended with "!").
 # Having features missing may case the build to fail.
-FEATURES_MISSING = $(sort $(filter-out $(FEATURES_PROVIDED),$(FEATURES_REQUIRED)))
+FEATURES_MISSING = $(sort $(filter-out $(FEATURES_PROVIDED),$(FEATURES_REQUIRED)) \
```
fjmolinas: You are changing the definition of `FEATURES_MISSING` and not updating the doc:

https://github.com/RIOT-OS/RIOT/blob/3eae2f7b792d60f48006d8338c91a73cb4e0a99a/makefiles/info.inc.mk#L61

Also why is a used blacklisted feature a `FEATURES_MISSING`? To me the usage doesn't fit with the name. This would be more of a  `FEATURES_USED_BLACKLISTED`.

kaspar030: Used but blacklisted features get missed as `!<feature>`. Doc is here:

https://github.com/kaspar030/RIOT/blob/3eae2f7b792d60f48006d8338c91a73cb4e0a99a/Makefile.features#L25-L29

There's also `_features_used_blacklisted`.

fjmolinas:
```
Used but blacklisted features get missed as !<feature>. Doc is here:

https://github.com/kaspar030/RIOT/blob/3eae2f7b792d60f48006d8338c91a73cb4e0a99a/Makefile.features#L25-L29
```

I referenced `RIOT/makefiles/info.inc.mk`, not the code.

> ```There's also _features_used_blacklisted.``` 

I'm not arguing this, I'm saying names should make sense with the definiton. A missing `FEATURE` is not a blacklisted used `FEATURE`, why does it make sense adding `_features_used_blacklisted` to `FEATURES_MISSING`. Is this just to re-use the `FEATURES_MISSING` error handling?

If I do `BOARD=esp8266-olimex-mod make -C tests/lwip info-build`, I get:

```
FEATURES_MISSING (only non optional features):
         !arch_esp8266
FEATURES_BLACKLIST (blacklisted features):
         arch_esp8266
```

Before, `FEATURES_MISSING` was self-documented because the name made sense. Now if I do not look at the code, I do not undersand what is going on with `!arch_esp8266`, is `!arch_esp8266` missing? its not missing? (I do not like double negation). In #11492 an effort was made for names to make sense.

kaspar030: > Now if I do not look at the code, I do not undersand what is going on with `!arch_esp8266`, is `!arch_esp8266` missing?

Do *you* know or do you assume that is what's gonna happen to someone?
I think if this shows up:

```
FEATURES_MISSING (only non optional features):
         !arch_esp8266
FEATURES_BLACKLIST (blacklisted features):
         arch_esp8266
```

it is pretty clear what's happening.

But yes, this is re-using FEATURES_MISSING.

> A missing FEATURE is not a blacklisted used FEATURE, why does it make sense adding _features_used_blacklisted to FEATURES_MISSING.

Well, it is adding `!feature`, not `feature`, to FEATURES_MISSING. For C devs there should be some meaning there.
Before, features could just be "missing" (e.g., "not provided"). Now, with `!feature`, they can be "used but blacklisted". A more fitting name might be FEATURE_REQUIREMENTS_UNFULFILLED" or something similar.

Again, we're in a situation where the machinery works but the names are not to everyone's liking.
How about we go with the functionality and move the naming struggle elsewhere?

fjmolinas: > Do you know or do you assume that is what's gonna happen to someone?

Yes since I know, to me it doesn't make sense.

> How about we go with the functionality and move the naming struggle elsewhere?

 ...

fjmolinas: I'll dismiss my review.

kaspar030:
Thanks!

Issue #9048, opened 2018-04-29
[RFC] CoAP API updates for options


kb2ma:
#### Description
Discussion for nanocoap/gcoap API updates to support use of options as defined in #8772. See [CoAP API Options Update](https://github.com/RIOT-OS/RIOT/wiki/CoAP-API-Options-Update) wiki page for the proposal.

Also see #8920, which triggered the proposal.

kb2ma:
I just modified the page to require use of coap_opt_finish(), including addition of a parameter to specify that a payload is expected. This addition allows writing the payload marker at this point. Previously, the idea was to maintain the payload marker at all times to avoid use of coap_opt_finish() if no options. I think that approach is too error prone and susceptible to off-by-one.

kb2ma:
Building up to a WIP PR in my [nanocoap/build_request_pkt branch](https://github.com/kb2ma/RIOT/tree/nanocoap/build_request_pkt).

kb2ma:
I just modified the page to remove the 'flags' attribute from coap_pkt_t. Instead, the user requests to sort options via a bitwise parameter in coap_opt_finish(). The addition of the sort function is controlled by a compile time macro.

PR #9085, opened 2018-05-05
net/nanocoap: Build message with coap_pkt_t

kb2ma:
### Contribution description
Adds an API to nanocoap to use a coap_pkt_t to build a message, including options. Builds on existing buffer-based functions. Use of the coap_pkt_t struct makes it easier to manage options.

Includes:

- coap_pkt_init()
- coap_opt_add_string()
- coap_opt_add_uint()
- coap_opt_finish()

This approach also allows integration by gcoap, bringing the APIs closer together. 

### Issues/PRs references
Implements ideas described in #9048 and #8831. Provides an API useful to #8920.

### DIFF
--- a/sys/include/net/nanocoap.h
+++ b/sys/include/net/nanocoap.h
@@ -17,6 +17,7 @@
  * @brief       nanocoap API
  *
  * @author      Kaspar Schleiser <kaspar@schleiser.de>
+ * @author      Ken Bannister <kb2ma@runbox.com>
  */
 
 #ifndef NET_NANOCOAP_H
@@ -225,6 +226,18 @@ extern "C" {
 /** @} */
 
 /**
+ * @name coap_opt_finish() flag parameter values
+ *
+ * Directs packet/buffer updates when user finishes adding options
+ * @{
+ */
+/** @brief    no special handling required */
+#define COAP_OPT_FINISH_NONE     (0x0000)
+/** @brief    expect a payload to follow */
+#define COAP_OPT_FINISH_PAYLOAD  (0x0001)
+/** @} */
+
+/**
  * @brief   Raw CoAP PDU header structure
  */
 typedef struct __attribute__((packed)) {
@@ -391,6 +404,22 @@ ssize_t coap_build_hdr(coap_hdr_t *hdr, unsigned type, uint8_t *token,
                        size_t token_len, unsigned code, uint16_t id);
 
 /**
+ * @brief   Initialize a packet struct, to build a message buffer
+ *
+ * @pre  buf              CoAP header already initialized
+ * @post pkt.flags        all zeroed
+ * @post pkt.payload      points to first byte after header
+ * @post pkt.payload_len  set to maximum space available for options + payload
+ *
+ * @param[out]   pkt        pkt to initialize
+ * @param[in]    buf        buffer to write for pkt, with CoAP header already
+ *                          initialized
+ * @param[in]    len        length of buf
+ * @param[in]    header_len length of header in buf, including token
+ */
+void coap_pkt_init(coap_pkt_t *pkt, uint8_t *buf, size_t len, size_t header_len);
+
+/**
  * @brief   Insert a CoAP option into buffer
  *
  * This function writes a CoAP option with nr. @p onum to @p buf.
@@ -498,6 +527,52 @@ size_t coap_put_option_block1(uint8_t *buf, uint16_t lastonum, unsigned blknum, 
  * @returns     amount of bytes written to @p pkt_pos
  */
 size_t coap_put_block1_ok(uint8_t *pkt_pos, coap_block1_t *block1, uint16_t lastonum);
+
+/**
+ * @brief   Encode the given string as option(s) into pkt
+ *
+ * Use separator to split string into multiple options.
+ *
+ * @post pkt.payload advanced to first byte after option(s)
+ * @post pkt.payload_len reduced by option(s) length
+ *
+ * @param[in,out] pkt         pkt referencing target buffer
+ * @param[in]     optnum      option number to use
+ * @param[in]     string      string to encode as option
+ * @param[in]     separator   character used in @p string to separate parts
+ *
+ * @return        number of bytes written to buffer
+ * @return        -ENOSPC if no available options
+ */
+ssize_t coap_opt_add_string(coap_pkt_t *pkt, uint16_t optnum, const char *string, char separator);
+
+/**
+ * @brief   Encode the given uint option into pkt
+ *
+ * @post pkt.payload advanced to first byte after option
+ * @post pkt.payload_len reduced by option length
+ *
+ * @param[in,out] pkt         pkt referencing target buffer
+ * @param[in]     optnum      option number to use
+ * @param[in]     value       uint to encode
+ *
+ * @return        number of bytes written to buffer
+ * @return        <0 reserved for error but not implemented yet
+ */
+ssize_t coap_opt_add_uint(coap_pkt_t *pkt, uint16_t optnum, uint32_t value);
+
+/**
+ * @brief   Finalizes options as required and prepares for payload
+ *
+ * @post pkt.payload advanced to first available byte after options
+ * @post pkt.payload_len is maximum bytes available for payload
+ *
+ * @param[in,out] pkt         pkt to update
+ * @param[in]     flags       see COAP_OPT_FINISH... macros
+ *
+ * @return        total number of bytes written to buffer
+ */
+ssize_t coap_opt_finish(coap_pkt_t *pkt, uint16_t flags);
 
 /**
  * @brief   Get content type from packet
--- a/sys/net/application_layer/nanocoap/nanocoap.c
+++ b/sys/net/application_layer/nanocoap/nanocoap.c
@@ -383,6 +383,15 @@ ssize_t coap_build_hdr(coap_hdr_t *hdr, unsigned type, uint8_t *token, size_t to
     return sizeof(coap_hdr_t) + token_len;
 }
 
+void coap_pkt_init(coap_pkt_t *pkt, uint8_t *buf, size_t len, size_t header_len)
+{
+    memset(pkt, 0, sizeof(coap_pkt_t));
+    pkt->hdr = (coap_hdr_t *)buf;
+    pkt->token = buf + sizeof(coap_hdr_t);
+    pkt->payload = buf + header_len;
+    pkt->payload_len = len - header_len;
+}
+
 static int _decode_value(unsigned val, uint8_t **pkt_pos_ptr, uint8_t *pkt_end)
 {
     uint8_t *pkt_pos = *pkt_pos_ptr;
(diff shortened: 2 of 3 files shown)


kb2ma:
I plan to add a couple more tests, including for the Content-Format uint option. I think that then I can remove the WIP label. The coap_opt_get_string() function can be added to #8920. We can create a PR like #8920 for uint options.

kb2ma:
Squashed WIP commits; ready for review. The unit tests provide decent coverage. To give the work some live testing, I created a CLI app based on the gcoap example. See my [riot-nano-cli](https://github.com/kb2ma/riot-nano-cli) repo.

As I pulled this together, a couple of issues arose in my mind:

- Do we need to create some kind of #ifdef to distinguish use of pkt-based functions to build a message vs. the traditional buffer based functions? There will be a transitional period until the buffer-based functions, like coap_put_option_block1() can be replaced.

- We should take advantage of the information in the packet struct to update/replace coap_put_option(). The change is to inform the function of the maximum buffer length available, and return -ENOSPC if not sufficient. The option header length can be more than 1, so the caller really can't predict if there is enough space. I've added asserts for now.

kb2ma:
@kaspar030, I assume that the basic approach here is acceptable. I am starting on a PR based on this one to adapt gcoap to it. This will allow us to remove the gcoap-specific #ifdefs added in #8772 and to reverse the stack size increase in #9000.

kb2ma:
Squashed. No one has commented, and this simplifies a follow-on PR to sort options that I'm creating.

sys/include/net/nanocoap.h
```diff
@@ -499,6 +524,50 @@ size_t coap_put_option_block1(uint8_t *buf, uint16_t lastonum, unsigned blknum,
...
 
+/**
+ * @brief   Encode the given string as option(s) into pkt
+ *
+ * Use separator to split string into multiple options.
+ *
+ * @post pkt.payload advanced to first byte after option(s)
+ * @post pkt.payload_len reduced by option(s) length
+ *
+ * @param[inout]  pkt         pkt referencing target buffer
```
bergzand: This should be `@param[in,out]` according to [doxygen](https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdparam)

sys/include/net/nanocoap.h
```diff
@@ -224,6 +225,15 @@ extern "C" {
 #define COAP_BLOCKWISE_SZX_MAX  (7)
 /** @} */
 
+/**
+ * @name coap_opt_finish() flag parameter values
```
bergzand: A small description of the purpose of these flags would be helpful here.

sys/include/net/nanocoap.h
```diff
@@ -499,6 +524,50 @@ size_t coap_put_option_block1(uint8_t *buf, uint16_t lastonum, unsigned blknum,
...
+ */
+ssize_t coap_opt_add_string(coap_pkt_t *pkt, uint16_t optnum, const char *string, char separator);
+
+/**
+ * @brief   Encode the given uint option into pkt
+ *
+ * @post pkt.payload advanced to first byte after option
+ * @post pkt.payload_len reduced by option length
+ *
+ * @param[inout]  pkt         pkt referencing target buffer
```
bergzand: Same as above

sys/include/net/nanocoap.h
```diff
@@ -499,6 +524,50 @@ size_t coap_put_option_block1(uint8_t *buf, uint16_t lastonum, unsigned blknum,
...
+ * @post pkt.payload_len reduced by option(s) length
+ *
+ * @param[inout]  pkt         pkt referencing target buffer
+ * @param[in]     optnum      option number to use
+ * @param[in]     string      string to encode as option
+ * @param[in]     separator   character used in @p string to separate parts
+ *
+ * @return        number of bytes written to buffer
+ */
+ssize_t coap_opt_add_string(coap_pkt_t *pkt, uint16_t optnum, const char *string, char separator);
```
bergzand: Is it possible for this function to return error codes? If not, can we change the return type to `size_t` instead of `ssize_t`?

sys/include/net/nanocoap.h
```diff
@@ -499,6 +524,50 @@ size_t coap_put_option_block1(uint8_t *buf, uint16_t lastonum, unsigned blknum,
...
+ * @post pkt.payload advanced to first byte after option
+ * @post pkt.payload_len reduced by option length
+ *
+ * @param[inout]  pkt         pkt referencing target buffer
+ * @param[in]     optnum      option number to use
+ * @param[in]     value       uint to encode
+ *
+ * @return        number of bytes written to buffer
+ */
+ssize_t coap_opt_add_uint(coap_pkt_t *pkt, uint16_t optnum, uint32_t value);
```
bergzand: Same with the return type as for `coap_opt_add_string`

sys/include/net/nanocoap.h
```diff
@@ -499,6 +524,50 @@ size_t coap_put_option_block1(uint8_t *buf, uint16_t lastonum, unsigned blknum,
...
+
+/**
+ * @brief   Finalizes options as required and prepares for payload
+ *
+ * @post pkt.payload advanced to first available byte after options
+ * @post pkt.payload_len is maximum bytes available for payload
+ *
+ * @param[inout]  pkt         pkt to update
```
bergzand: `[in,out]`

sys/include/net/nanocoap.h
```diff
@@ -499,6 +524,50 @@ size_t coap_put_option_block1(uint8_t *buf, uint16_t lastonum, unsigned blknum,
...
+ * @param[in]     flags       see COAP_OPT_FINISH... macros
+ *
+ * @return        total number of bytes written to buffer
+ */
+ssize_t coap_opt_finish(coap_pkt_t *pkt, uint16_t flags);
```
bergzand: How many flags are we expecting here, would `uint8_t` suffice?

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -595,6 +604,80 @@ size_t coap_put_option_uri(uint8_t *buf, uint16_t lastonum, const char *uri, uin
...
 }
 
+/* Common functionality for addition of an option */
+static ssize_t _add_opt_pkt(coap_pkt_t *pkt, uint16_t optnum, uint8_t *val,
+                            size_t val_len)
+{
+    uint16_t lastonum = (pkt->options_len)
+            ? pkt->options[pkt->options_len - 1].opt_num : 0;
+    assert(optnum >= lastonum);
+
```
bergzand: I think you should check here whether there is still an empty position in the options array, to prevent array-out-of-bounds bugs

sys/include/net/nanocoap.h
```diff
@@ -390,6 +400,21 @@ ssize_t coap_handle_req(coap_pkt_t *pkt, uint8_t *resp_buf, unsigned resp_buf_le
...
+ * @post pkt.payload points to first byte after header
+ * @post pkt.payload_len set to maximum space available for options + payload
+ *
+ * @param[out]   pkt        pkt to initialize
+ * @param[in]    buf        buffer to write for pkt, with CoAP header already
+ *                          initialized
+ * @param[in]    len        length of buf
+ * @param[in]    header_len length of header in buf, including token
+ */
+void coap_pkt_init(coap_pkt_t *pkt, uint8_t *buf, size_t len, size_t header_len);
```
bergzand: I think an explicit @pre to indicate that there should be an initialized `coap_hdr_t` in the buffer would be nice. Somebody is going to miss the note with the `buf` param there :)

sys/include/net/nanocoap.h
kb2ma: Sure thing.

sys/include/net/nanocoap.h
kb2ma: Will do.

sys/include/net/nanocoap.h
kb2ma: This is an interesting API design question. These `coap_opt_add...()` functions rely on `_add_opt_pkt()`, which in turn relies on `coap_put_option()`. IMO the latter can be improved to return -ENOSPC because the user can't always determine in advance if the option will be too long for the available buffer space. See the questions I pose in a comment [above](https://github.com/RIOT-OS/RIOT/pull/9085#issuecomment-387958364).

So, I'm sort of anticipating that we will update these functions eventually to return an error code in that case. Does it make sense to update the API *now* for these new functions so the user does not need to update their code in the future? Besides, use of a signed value will never really be a problem in this case. Why box in the API if you don't need to?

OTOH, maybe this whole idea of overrunning the buffer when writing options is so unlikely that the present use of an assert after the option is added is the best/simplest solution.

sys/include/net/nanocoap.h
kb2ma: Might we need more than eight flags? Another interesting API question. We already have used two (add-payload, and sort-options in another PR) I rounded up here figuring it didn't really matter in terms of space, and it allows for future flexibility. Conceivably we also could use the high-order byte if some future flag requires an associated value.

sys/net/application_layer/nanocoap/nanocoap.c
kb2ma: Excellent idea!

sys/include/net/nanocoap.h
kb2ma: I agree completely.

kb2ma:
Thanks for the review, @bergzand. Pushed fixups.

sys/include/net/nanocoap.h
bergzand: Ok. `uint16_t` sounds good then. Could you please reflect this in the flag defines, so `0x0001` instead of `0x1` to have the data size a bit more explicit there.

sys/include/net/nanocoap.h
bergzand: `ssize_t` sounds good to me. I agree with you reason that these function might start returning errors. This error could not only be caused by a buffer overflow with the pdu buffer space, but also with the option array being completely filled.

I tend to favor a "soft" error return code here over an assertion. Mostly because it is possible for external CoAP requests to influence these options, and it would be bad if an assertion could be remotely triggered with some specific request (causing a packet of death vulnerability).

I think it would be best to add an `@return negative on error` to the doxygen both here and with the other "coap_opt_add" functions to indicate the need for error checking.

kb2ma: Please see the exact wording and let me know if any concerns. Also I updated the implementation of `coap_opt_add_string()` to return -ENOSPC. The input string may contain multiple separators, and it might be difficult for the caller to determine this in advance.

sys/include/net/nanocoap.h
kb2ma: Done.

emmanuelsearch:
@bergzand @kb2ma are there a lot of to dos left here?

kb2ma:
I believe I've addressed @bergzand's comments. It would be nice to at least have a high-level "ok with the approach" confirmation from @kaspar030 since this change is a significant addition to nanocoap. It is a new API to build a message instead of the historical, purely buffer-based approach.

bergzand:
@kaspar030 On a samr21-xpro, this patch adds 26 bytes of flash usage to the nanocoap-server example, that is still unmodified without any usage of these functions.
(measured with [cosy](https://github.com/haukepetersen/cosy))

kaspar030:
@kb2ma The approach looks good to me -> high-level ACK.

kb2ma:
@bergzand, thanks for measuring. I plan to review and report. I haven't tried cosy yet.

@kaspar030, thanks for confirmation.

bergzand:
@kb2ma I'd like to test this a bit more and then ACK it. Feel free to squash in the mean time.

kb2ma:
Rebased and squashed.

Relative to master, I see the difference in size of the text sections as 52 bytes. I attribute this to 

- realization of _encode_uint(), which is called from coap_opt_add_uint() as well as the historical coap_put_option_block()
- the longer branch name

In the proposed [module doc](https://kb2ma.github.io/riot-coap-api/nanocoap-api-doc.html) for nanocoap, I include macros to exclude the code, which clears up this problem. If this is important to do now, I can add to this PR, or we can put it in a follow-on PR.

bergzand:
> If this is important to do now, I can add to this PR, or we can put it in a follow-on PR.

Let's do this in a follow up.

PR #9090, opened 2018-05-07
 sys/newlib: Bugfix, add missing syscalls (only stubs)

### LINKED ISSUE #9051: Linker failure when using `times` or `rename`
#### Description

When building for an ARM platform, certain functions in newlib are not implemented and cause a linker error. 

Try to compile a program that uses [`times`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/times.html) or [`rename`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/rename.html).

For example:

```c
#include <stdio.h>
#include <sys/times.h>

/*
#include <string.h>
#include <errno.h>
#include <reent.h>

int _link_r(struct _reent *ptr, const char *old_name, const char *new_name)
{
    (void)old_name;
    (void)new_name;

    ptr->_errno = ENOSYS;

    return -1;
}

clock_t _times(struct tms *ptms)
{
    memset(ptms, 0, sizeof(*ptms));

    return 0;
}
*/
int main(void)
{
    struct tms tm;
    clock_t t = times(&tm);

    puts("Hello World!");

    printf("You are running RIOT on a(n) %s board.\n", RIOT_BOARD);
    printf("This board features a(n) %s MCU.\n", RIOT_MCU);
    printf("Time %lu.\n", t);
    rename("/in", "/out");

    puts("renamed!!!");

    return 0;
}
```

#### Expected results

Compilation succeeds, even though `times` may return irrelevant results an `rename` will alwa

#### Actual results

Compilation fails at the linker stage with error:

```
/opt/gcc-arm-none-eabi-7-2017-q4-major/bin/../lib/gcc/arm-none-eabi/7.2.1/../../../../arm-none-eabi/lib/thumb/v7e-m/libc_nano.a(lib_a-timesr.o): In function `_times_r':
timesr.c:(.text._times_r+0x2): undefined reference to `_times'
/opt/gcc-arm-none-eabi-7-2017-q4-major/bin/../lib/gcc/arm-none-eabi/7.2.1/../../../../arm-none-eabi/lib/thumb/v7e-m/libc_nano.a(lib_a-linkr.o): In function `_link_r':
linkr.c:(.text._link_r+0xe): undefined reference to `_link'
```

#### Versions

Operating system: Debian Testing and Stable
Build environment:
 - arm-none-eabi-gcc: arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2017-q4-major) 7.2.1 20170904 (release) [ARM/embedded-7-branch revision 255204]
 - arm-none-eabi-newlib: "2.5.0"

jcarrano:
Add small stubs for _times and _link_r so that code using times() or link()/rename() can still compile. The functions are marked as not implemented and return invalid codes.

### Issues/PRs references

Fixes #9051 .

### DIFF
--- a/sys/newlib_syscalls_default/syscalls.c
+++ b/sys/newlib_syscalls_default/syscalls.c
@@ -45,6 +45,8 @@
 
 #include "uart_stdio.h"
 
+#include <sys/times.h>
+
 #ifdef MODULE_XTIMER
 #include <sys/time.h>
 #include "div.h"
@@ -462,6 +464,23 @@ int _unlink_r(struct _reent *r, const char *path)
 #endif /* MODULE_VFS */
 
 /**
+ * Create a hard link (not implemented).
+ *
+ * @todo    Not implemented.
+ *
+ * @return  -1. Sets errno to ENOSYS.
+ */
+int _link_r(struct _reent *ptr, const char *old_name, const char *new_name)
+{
+    (void)old_name;
+    (void)new_name;
+
+    ptr->_errno = ENOSYS;
+
+    return -1;
+}
+
+/**
  * @brief Query whether output stream is a terminal
  *
  * @param r     TODO
@@ -516,3 +535,18 @@ int _gettimeofday_r(struct _reent *r, struct timeval *restrict tp, void *restric
     return -1;
 }
 #endif
+
+/**
+ * Current process times (not implemented).
+ *
+ * @param[out]  ptms    Not modified.
+ *
+ * @return  -1, this function always fails. errno is set to ENOSYS.
+ */
+clock_t _times_r(struct _reent *ptr, struct tms *ptms)
+{
+    (void)ptms;
+    ptr->_errno = ENOSYS;
+
+    return (-1);
+}


sys/newlib_syscalls_default/syscalls.c
```diff
@@ -516,3 +535,17 @@ int _gettimeofday_r(struct _reent *r, struct timeval *restrict tp, void *restric
...
+ * Current process times (not implemented).
+ *
+ * @param[out]  ptms    Filled with zeros.
+ *
+ * @return  Zero.
+ */
+clock_t _times(struct tms *ptms)
+{
+    memset(ptms, 0, sizeof(*ptms));
+
```
kaspar030: does errno have to be cleared? Spec says, errno indicates error on error...

jcarrano: @kaspar030 You are right, this is an example from libgloss (part of newlib):
```
/* _times -- no clock, so return an error.  */
clock_t _times (struct tms *buf)
{
  errno = EINVAL;
  return (-1);
}
```
I'm fixing it.

kaspar030:
please squash

basilfx:
@jcarrano The build failed, because of unused parameters.

```
syscalls.c: In function '_times':
syscalls.c:546:28: error: unused parameter 'ptms' [-Werror=unused-parameter]
 clock_t _times(struct tms *ptms)
```

You can mark them as unused by adding `(void) ptms;` to `_times`.

sys/newlib_syscalls_default/syscalls.c
```diff
@@ -516,3 +535,18 @@ int _gettimeofday_r(struct _reent *r, struct timeval *restrict tp, void *restric
...
+ *
+ * @param[out]  ptms    Not modified.
+ *
+ * @return  -1, this function always fails.
+ */
+clock_t _times(struct tms *ptms)
+{
+    (void)ptms;
+
+    errno = EINVAL;
```
kaspar030: Shouldn't this also use the errno variable from struct reent?

jcarrano: I copied it from libgloss, which is part of newlib.

jia200x:
@jcarrano any news on this?

jcarrano:
@jia200x I addressed @kaspar030 review, and squashed. Btw, I'm also working intermittently on what I think is a better fix (#9258  ), but this one should work for now.

tcschmidt:
@kaspar030 do changes satisfy your concerns?

danpetry:
OK I'm satisfied with the approach here, on the basis that it replicates the approach elsewhere and that that approach is being addressed more comprehensively in https://github.com/RIOT-OS/RIOT/pull/9258. Also, the added functions aren't actually used elsewhere meaning, for example, they won't have a memory impact anywhere except in the Lua interpreter. Would be good to address the above PR ASAP to avoid errors if other people try to use the related newlib functions.

PR #9095, opened 2018-05-07
murdock: enable nrf52dk for on-hardware tests

kaspar030:
### Contribution description

This PR enables on-hardware tests on two nrf52dk boards.

nrf52dk uses ```$(HEXFILE)``` as flashfile. One of the commits makes an exception for this. IMO, this is ugly, but will go away as soon as #8838 is merged.

### Issues/PRs references

#8838

### DIFF
--- a/.murdock
+++ b/.murdock
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-: ${TEST_BOARDS_AVAILABLE:="samr21-xpro"}
+: ${TEST_BOARDS_AVAILABLE:="nrf52dk samr21-xpro"}
 : ${TEST_BOARDS_LLVM_COMPILE:="iotlab-m3 native nrf52dk mulle nucleo-f401re samr21-xpro slstk3402a"}
 
 export RIOT_CI_BUILD=1


cladmi:
Looks good.
Still funny that the current hack is set `FLASHFILE:=$(HEXFILE)` when it's still a .bin :D 

Does the patch for the test work ? it could help for other shell based tests that from time to time fail on samr21.

I tried running it but it currently failed, so waiting on your input for re-running it.

tcschmidt:
ping @MrKevinWeiss

tests/ps_schedstatistics/tests/01-run.py
```diff
@@ -40,8 +41,10 @@ def _check_startup(child):
 
 def _check_help(child):
     child.sendline('')
-    child.expect('>')
+    child.expect('> ')
```
MrKevinWeiss: Instead of adding sleep can you just `child.expect('> ', timeout=0.1)`?

tests/ps_schedstatistics/tests/01-run.py
```diff
@@ -40,8 +41,10 @@ def _check_startup(child):
...
+    time.sleep(0.1)
     child.sendline('help')
```
MrKevinWeiss: Same here

tests/ps_schedstatistics/tests/01-run.py
kaspar030: No, that would make pexpect give '> ' more time, right? I intended to add a slight delay after receiving the prompt, before sending the next line.

tcschmidt:
@MrKevinWeiss @smlng is this PR still needed given the establishing HIL testing technology and deployment?

MrKevinWeiss:
@kaspar030 Do you have the murdock workers connected?  Now that we have the [compile_and_test_for_board](https://github.com/RIOT-OS/Release-Specs/blob/master/02-tests/compile_and_test_for_board.py) any maintainer can run these tests on there desk.  Also I found lots of failures that we due to things unrelated to the test (ie. libfixmath, xtimer_periodic_wakeup).  Plus some cases where the flasher either doesn't work or get held up and the only way to flash is by pressing the reset button at the right time.

How would you like to proceed?

kaspar030:
> Do you have the murdock workers connected?

One is connected, but somehow the incoming UART is broken. I haven't gotten around to debugging that, yet.

MrKevinWeiss:
I am not sure but I also think IoT labs supports a whole bunch of boards now.  They may have something available.

kaspar030:
> I am not sure but I also think IoT labs supports a whole bunch of boards now. They may have something available.
> Now that we have the compile_and_test_for_board any maintainer can run these tests on there desk. 
> is this PR still needed given the establishing HIL testing technology and deployment?

All of those don't support running tests per-PR as part of the CI.

emmanuelsearch:
@kaspar030 I am in the office today if there's anyway to physically help with the attached nrf52dk

kaspar030:
> @kaspar030 I am in the office today if there's anyway to physically help with the attached nrf52dk

Yes, if you can find a second board, you could replace the nrf that's attached to the Pi. It might be a hardware problem. The nrf is connected to the Pi named "pi-36f90aef".

emmanuelsearch:
@kaspar030 can you try again? I think the USB cable was faulty. I changed to a new one.

kaspar030:
> @kaspar030 can you try again? I think the USB cable was faulty. I changed to a new one.

Seems to work. Thanks!

kaspar030:
One of the two NRFs is working now, the other one isn't. I'll investigate.

kaspar030:
I have fixed the USB power cycling and updated the JLINK firmwares. Now the UARTs seem to work reliably. 

How do we proceed here? I just did a couple of runs with no error. I suggest merging.

kaspar030:
Interestingly, the benchmarks seem to run a slightly faster with llvm builds on nrf52dk. On samr21-xpro, gnu is faster. :confused:

kaspar030:
This is not WIP anymore. Anyone?

cladmi:
The last commit with the `murdoc.inc.mk` can be discarded. The value is found automatically with the current master. The test could be re-run after.

```
BOARD=nrf52dk make --no-print-directory -C examples/hello-world/ info-debug-variable-FLASHFILE
/home/harter/work/git/RIOT/examples/hello-world/bin/nrf52dk/hello-world.bin
```

What did you do for the JLink update ? I may need the same thing as sometime my jlink uart is mixed and could be added in the doc (other than this PR but asking as it was mentioned here).

kaspar030:
> The last commit with the `murdoc.inc.mk` can be discarded.

Nice. rebased & dropped that commit.

kaspar030:
> What did you do for the JLink update ?

It was updated automatically when I flashed the board with JLink on my workstation, which had a very recent JLinkExe (I think from this year even).

cladmi:
Ok, so my board may have been updated already.

When looking in the murdock output, some files are not recognized as being `utf-8` and are forced to be downloaded:

https://ci.riot-os.org/RIOT-OS/RIOT/9095/e3729be0358ec5a5e952c254fc97867c4acd85a4/output/run_test/tests/ssp/nrf52dk:gnu.txt

I get a file download.It opens as `chinese` in gedit, but gvim handles it correctly.

When opening the url for `samr21-xpro` it works correctly.

https://ci.riot-os.org/RIOT-OS/RIOT/9095/e3729be0358ec5a5e952c254fc97867c4acd85a4/output/run_test/tests/ssp/samr21-xpro:gnu.txt

cladmi:
"file" does not recognize the file as `utf-8` It looks like there are `^@` characters that it does not like. Apparently null characters https://en.wikipedia.org/wiki/Null_character

kaspar030:
> "file" does not recognize the file as `utf-8` It looks like there are `^@` characters that it does not like. Apparently null characters https://en.wikipedia.org/wiki/Null_character

At least it is defined how Firefox does this:

From [here](https://developer.mozilla.org/en-US/docs/Mozilla/How_Mozilla_determines_MIME_Types):

```
When the Content-Type sent by the server is one of (case-sensitively)

    text/plain
    text/plain; charset=ISO-8859-1
    text/plain; charset=iso-8859-1

and the server did not send a Content-Encoding header, Mozilla will sniff the first block of data it gets and check for non-text bytes. Text bytes are 9-13, 27, and 31-255. When encountering a non-text byte, the helper app dialog will be shown, showing the MIME type corresponding to the extension of the file.
```

kaspar030:
@smlng can we try forcing all ```*.txt``` to be ```text/plain; charset=utf-8```?

kaspar030:
Anyhow, the output issue is unrelated to this PR.

@cladmi @smlng do you see any reason not to merge this?

MrKevinWeiss:
@kaspar030 hmmm I can't see the tests being run on the hardware in the murdock output, shouldn't the samr21 and nrf52dk be triggered and show up in the worker stats?
```
--- worker stats:
e5 total: 3326 pass: 3326 fail: 0 avg: 2.1s
e52 total: 3602 pass: 3602 fail: 0 avg: 2.0s
haw-icc total: 6968 pass: 6968 fail: 0 avg: 2.3s
ifive total: 602 pass: 602 fail: 0 avg: 3.2s
mobi3.inet.haw-hamburg.de total: 4759 pass: 4759 fail: 0 avg: 3.3s
mobi7.inet.haw-hamburg.de total: 2897 pass: 2897 fail: 0 avg: 3.1s
mobi8.inet.haw-hamburg.de total: 2661 pass: 2661 fail: 0 avg: 4.4s
riot-ci.inet.haw-hamburg.de total: 1144 pass: 1144 fail: 0 avg: 3.4s
riotbuild total: 8504 pass: 8504 fail: 0 avg: 3.3s

```

cladmi:
The output issue will be seen on the first failed test for `nrf52dk` and somebody clicking on the test result. But if they open it in a good editor, it should be ok.

kaspar030:
@cladmi yes but this issue is also present for samr21-xpro, it is not nrf52dk specific.

kaspar030:
@MrKevinWeiss results are there now. Care to confirm?  there's a race when setting both CI labels at the same time, it sometimes causes one to not be recognized by murdock.

MrKevinWeiss:
All good, thanks @kaspar030

cladmi:
> @cladmi yes but this issue is also present for samr21-xpro, it is not nrf52dk specific.

I did not know, I only saw it for `nrf52dk`.

PR #9096, opened 2018-05-07
cpu/stm32: add CPU_LINE_* variable and use it for stm32f0 and stm32f4

vincent-d:
### Contribution description

Most of the stm32 configuration can be shared by 'line'. This PR adds a variable CPU_LINE_* (e.g. CPU_LINE_STM32F401) which can be used to group configurations. This avoid having to list all cpus from the line with `defined(...)`.

I changed only in stm32f4 so far, to show how it works and maybe discuss a bit. The rest can be done in other PRs anyway.

### Issues/PRs references

Based on #9079

### DIFF
--- a/cpu/stm32_common/Makefile.include
+++ b/cpu/stm32_common/Makefile.include
@@ -23,8 +23,16 @@ export INCLUDES += -I$(RIOTCPU)/stm32_common/include
 
 include $(RIOTCPU)/stm32_common/stm32_mem_lengths.mk
 
+# Get CPU_LINE_ variable
+-include $(RIOTCPU)/$(CPU)/stm32_line.mk
+CPU_LINE ?= $(shell echo $(CPU_MODEL) | cut -c -9 | tr 'a-z-' 'A-Z_')xx
+
+# Set CFLAGS
+export CFLAGS += -D$(CPU_LINE) -DCPU_LINE_$(CPU_LINE)
+
 info-stm32:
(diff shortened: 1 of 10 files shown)


vincent-d:
Rebased

vincent-d:
Ping?

vincent-d:
I had some new thoughts and tried f0 as well. It lead me to change slightly the `CPU_LINE_*` variables to match the one defined by ST in the headers.

We now define:
- `STM32Fxxxxx` to be able to include `stm32f0xx.h` or `stm32f4xx.h` which will select the right include.
- ̀CPU_LINE_STM32Fxxxxxx` to be use internally in RIOT files

It should simplify quite a bit the addition of new CPU variants.
It should work on f0 and f4.

kYc0o:
Maybe I ACKed too fast? Have you added support for more CPUs?

vincent-d:
> Have you added support for more CPUs?

No, not yet. So far, only f0 and f4 are ported to use the new variables, but there are no new part number support added by this PR.

vincent-d:
In follow up PRs, we should port other families to use these variables and maybe add new part numbers if needed.

kYc0o:
Yeah, I just verified again and everything is OK here. I have no means to test everything... So I'd say we merge and add more support for more CPUs later.

vincent-d:
Let's go then!

Issue #9093, opened 2018-05-07
EFM32 (sltb001a) is missing FEATURES_PROVIDED=cpp


Mario0x1:
#### Description
FEATURES_PROVIDED=cpp is not set in cpu/efm32/Makefile.features despite being able to build and run examples/riot_and_cpp 

#### Steps to reproduce the issue
```
cd examples/riot_and_cpp
make BOARD=sltb001a flash
```

#### Expected results

>make BOARD=sltb001a

informs that

>...
> There are unsatisfied feature requirements: cpp
> 
> 
> EXPECT ERRORS!
> 
> 
> Building application "riot_and_cpp" for "sltb001a" with MCU "efm32".
> ...

It should fail if RIOT-gecko-sdk does not support cpp.

#### Actual results
example/riot_and_cpp can be build and flashed and returns

> main(): This is RIOT! (Version: 2018.04-devel-1009-g4f4e-hoss-214)
> 
> ************ RIOT and C++ demo program ***********
> 
> ******** Hello, you're in thread #2 ********
> We'll test C++ class and methods here!
> Instanciating Object [constructor called]
> Im starting!
> 
> -= Test overloading functions =-
> Hello!
> Hello![int: 42]
> Hello![float: ]
> 
> -= Test namespace =-
> typing std::vector is obsolete when 'using namespace std;'
> The vector vInts has been filled with 3 numbers.
> 
> -= Test iterator =-
> The content of vInts = { 1 3 2 }
> Destroying Object [destructor called]
> Im shutting down!
> ******** Hello, you're in thread #3 ********
> We'll test some C functions here!
> 
> -= hello function =-
> Hello world in a C function
> 
> -= day_of_week function =-
> day 13, month 6, year 2014 is Friday
> 
> This demo ends here, press Ctrl-C to exit (if you're on native)!

#### Versions
SEGGER J-Link Commander V6.22a
Gecko_sdk: https://github.com/basilfx/RIOT-gecko-sdk d381e526d68a2d0c951f37040c1c2e168ac66cd6

@basilfx

basilfx:
I've created #9097 to address this bug.

PR #9097, opened 2018-05-07
cpu: efm32: fix for cpp support.

basilfx:
### Contribution description

This fixes a missing include of `cpu/cortexm_common/Makefile.features`, which defines C++ support and `periph_pm`. 

I don't know why I missed that one, but there was no reason to leave it out.

### Issues/PRs references

#9093

### DIFF
--- a/cpu/efm32/Makefile.features
+++ b/cpu/efm32/Makefile.features
@@ -1,6 +1,5 @@
 FEATURES_PROVIDED += periph_cpuid
 FEATURES_PROVIDED += periph_flashpage
-FEATURES_PROVIDED += periph_pm
 
 FEATURES_CONFLICT += periph_rtc:periph_rtt
 FEATURES_CONFLICT_MSG += "On the EFM32, the RTC and RTT map to the same hardware peripheral."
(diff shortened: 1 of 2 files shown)


basilfx:
@mario0x1 can you acknowledge if this fixes it for you?

cpu/efm32/Makefile.features
```diff
@@ -1,10 +1,11 @@
...
-FEATURES_PROVIDED += periph_pm
 
 FEATURES_CONFLICT += periph_rtc:periph_rtt
 FEATURES_CONFLICT_MSG += "On the EFM32, the RTC and RTT map to the same hardware peripheral."
 
 ifeq (1,$(EFM32_TNRG))
   FEATURES_PROVIDED += periph_hwrng
 endif
+
+-include $(RIOTCPU)/cortexm_common/Makefile.features
```
gebart: Do this without minus (don't ignore missing). I know the other platforms have the minus as well, but it is better to do it with error checking so that we will get feedback on typos etc.

basilfx: Done.

basilfx:
Depends on #9098.

Mario0x1:
@basilfx this fixes the issue for me

gebart:
Yes please

basilfx:
I thought that #9098 would fix the errors, but it turns out that 107bc8c was still needed to fix build errors.

basilfx:
@gebart Can you check once more, and ACK if you agree?

basilfx:
Thanks!

PR #9098, opened 2018-05-07
tests: unittests: disable for new efm32 boards.

basilfx:
### Contribution description

Blacklist the (new) EFM32 boards in `tests/unittests` so that the build doesn't fail. I forgot to add them. That's why #9097 fails.

### Issues/PRs references

#9097


basilfx:
Thanks!

PR #9101, opened 2018-05-08
makefiles/info.inc.mk: add a info-debug-variable-% target

cladmi:
### Contribution description

Add a target to print an internal variable value.

Usage: make info-debug-variable-VARIABLENAME

Example: 

    make info-debug-variable-ELFFILE
    /path/to/riot/examples/hello-world/bin/native/hello-world.elf

Suggestions on the target name are welcomed.

### Issues/PRs references

I want to show that an internal value did not change with a refactoring.
Could also help having internal informations.

### DIFF
--- a/makefiles/info.inc.mk
+++ b/makefiles/info.inc.mk
@@ -1,6 +1,7 @@
 .PHONY: info-objsize info-buildsizes info-build info-boards-supported \
         info-features-missing info-modules info-cpu \
-        info-features-provided info-features-required
+        info-features-provided info-features-required \
+        info-debug-variable-%
 
 info-objsize:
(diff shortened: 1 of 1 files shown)


cgundogan:
```sh
make -pn | grep ELFFILE
```
? :smirk: Nevertheless, yeah, good idea!

cladmi:
@cgundogan this "make -pn | grep ELFFILE" will never work, as I define the rule with a %, so that you can also do `make info-debug-variable-RIOTBASE`.

However, maybe adding an `info-debug-variable-VARIABLE_NAME` that explains it could make it listed in `make help`.

PR #9104, opened 2018-05-09
cortexm: ldscript refactor

gebart:
### Contribution description

This is a rewrite of the cortexm_base ldscript to make it more similar to the default ldscripts for native and other platforms. Sections which do not need a runtime relocation (e.g. .text, .bss), except for some special cases (.preinit_array, .init_array, .fini_array etc.) do not need to have a output segment specified, as long as the attribute flags are set properly on the memory segments (#9089).

### Issues/PRs references

based on #9089


gebart:
Fixed an alignment issue with the initialization code on Cortex-M0

cladmi:
To simplify reviewing, can you put in the commit message a link to the file you used a reference?

cladmi:
I started reviewing by getting the default one and modifying it to look like our own to help me reviewing, I will try something next week in this direction. My goal is to really show what is needed  for RIOT which is not in the default one. So adding commits before that adds the default ldscript and adapts it before replacing our own.

Question, I first changed the indentation of the default ldscript to our own, I could do the opposite to really see the what are our required modifications from the default. Which way would you prefer ? Keeping our style or adopting the original one ?

To simplify comparing versions with mine, could you squash the changes that are not ". = ALIGN(4)" related ?

gebart:
@cladmi I inserted a commit with the default ldscript before the RIOT specific changes to make this easier to review.

gebart:
rebased and addressed build failures

gebart:
ping @cladmi 
This is useful to get #8711 further along

cladmi:
@gebart I am not forgetting it, no worry :)

But it is I think "only" for xfa https://github.com/RIOT-OS/RIOT/pull/9105 that also needs `ld -r` and I tried to focus first on the `ld -r` dependency as it is a major change that I would like ready so it can have enough review time.

So my current pipeline for this was more `ld -r` -> `cortexm ldscript` -> `xfa`.

If you would like to have it  before `ld -r`, for also other reasons, I can try to dedicate some time now.

PR #9105, opened 2018-05-09
core: introduce crossfile arrays (xfa), refactored

jnohlgard:
### Contribution description

A rebase of #7523. The alternate approach used here is to avoid introducing platform specific ldscripts and instead use the same approach as was used for the native platform in #7523. The only exception here is the AVR arch which needs a specific version of the patch ldscript to ensure that the xfa sections are placed where the code can use them. The AVR is a Harvard architecture which means that we can't simply place data in the code memory segment without adding extra code for accessing it.

TODO: Static check of linking xfa sections as suggested in https://github.com/RIOT-OS/RIOT/pull/7523#issuecomment-387024914

### Issues/PRs references

rebase/refresh of #7523 
based on #9104 #9089


kaspar030:
Thanks for taking over! I like using patch ldscripts, seems much cleaner. I think when I tried I didn't realize the order of "-T" matters.

Keep in mind, xfa in general still needs fixing the linking process /#5757 or probably better #8711).

kaspar030:
@gebart would you mind rebasing? The diff is quite large as is.

gebart:
Will do soon ish, when I am back at the computer

gebart:
@kaspar030 Feel free to rebase and force push to my branch if you like

cladmi:
I tried a static test:

``` diff
diff --git a/tests/xfa/Makefile b/tests/xfa/Makefile
index abe5a68..d5b1763 100644
--- a/tests/xfa/Makefile
+++ b/tests/xfa/Makefile
@@ -3,3 +3,11 @@ APPLICATION = xfa
 
+
+all: static-test
+static-test: $(ELFFILE)
+       $(eval xfastart := $(shell $(PREFIX)readelf --symbols $< |\
+              grep ' xfatest_const$$' | awk '{printf "0x%s", $$2}'))
+       $(eval xfaend   := $(shell $(PREFIX)readelf --symbols $< |\
+              grep ' xfatest_const_end$$' | awk '{printf "0x%s", $$2}'))
+       bash -c "[[ $(xfastart) < $(xfaend) ]]"
```

Only tested that I get an error when changing the sign, not really tested on a non implemented board.

cladmi:
When commenting `LINKFLAGS += -Txfa.ld` the static-test fails, yeah.

cladmi:
For msp430 based boards using `msp430-gcc-4.6.3` I get the following error:

`error: unknown option after ‘#pragma GCC diagnostic’ kind [-Werror=pragmas]`

```
msp430-gcc --version
msp430-gcc (GCC) 4.6.3 20120301 (mspgcc LTS 20120406 unpatched)
```
I disabled it with:
```
diff --git a/cpu/msp430_common/Makefile.include b/cpu/msp430_common/Makefile.include
index ab6e0cf..b34ce66 100644
--- a/cpu/msp430_common/Makefile.include
+++ b/cpu/msp430_common/Makefile.include
@@ -7,6 +7,9 @@ export CFLAGS += -DCPU_MODEL_$(MODEL)
 export UNDEF += $(BINDIR)/msp430_common/startup.o
 export USEMODULE += msp430_common msp430_common_periph periph_common
 
+# Ignore unknown pragmas (for xfa with gcc 4.6.3)
+CFLAGS += -Wno-pragmas
+
 DEFAULT_MODULE += oneway_malloc
 
 # include the msp430 common Makefile
```

With the `static-test`, the `pic32` patch and the warning disabled I can run `buildtest` with success.

cladmi:
I successfully run the xfa test with `iotlab-m3`, `iotlab-a8-m3`, and `wsn430-v1_4`.
I cannot test with `samr21-xpro` and `arduino-zero` it looks like saclay site is not working properly.

I could re-run all the automated tests these ones when it works again.

cladmi:
Hmm, in fact, the real problem is that `samr21-xpro` and `arduino-zero` do not show any output with this branch. I tested xfa, hello-world, and default.
They both use a `samd21` cpu.

gebart:
@cladmi could you test on those boards with only the cortexm_common ldscript refactoring applied?

cladmi:
Bisecting tells me it is the first bad commit:

`cortexm_common: Update ldscript to be more similar to binutils default`

gebart:
Updated:
 - Added unit test for verifying XFA linking at runtime
 - Added static check in tests/xfa Makefile
 - Corrected typos/mistakes in `XFA_USE` macros

gebart:
The PIC32 ldscripts are messy and difficult to work with... Trying to find a solution for now

gebart:
I found a solution to get what I believe is a correct memory map for PIC32 XFA. I don't have any hardware to test on though. Could someone run unittests/tests-core or tests/xfa on a pic32-clicker or pic32-wifire please?

cladmi:
I will now be able to do on `pic32-wifire` using https://github.com/RIOT-OS/RIOT/pull/9259

cladmi:
For `pic32-wifire`

`unittests` tests work for `tests-core`. The whole tests do not work but it is also the case even without this.

There is currently no `test` command in xfa, but here is the output

```
Type '/exit' to exit.
2018-05-31 18:07:56,063 - INFO # �main(): This is RIOT! (Version: 2018.07-devel-479-g0bc32-inetm02-pr/9105/introduce_crossfile_arrays__xfa__refactored)
2018-05-31 18:07:56,064 - INFO # Cross file array test
2018-05-31 18:07:56,065 - INFO # xfatest[2]:
2018-05-31 18:07:56,066 - INFO # [0] = 1, "xfatest1"
2018-05-31 18:07:56,067 - INFO # [1] = 2, "xfatest2"
2018-05-31 18:07:56,067 - INFO # xfatest_const[2]:
2018-05-31 18:07:56,068 - INFO # [0] = 123, "xfatest_const1"
2018-05-31 18:07:56,069 - INFO # [1] = 45, "xfatest_const2"
2018-05-31 18:08:13,671 - INFO # Exiting Pyterm
```

cladmi:
I am re-running all automated tests for `pic32-wifire` with this PR and will compare to master.

cladmi:
I added a python test script with a `test` command.

cladmi:
> Fixed CM0 alignment issues (#9104)
This fixed `samr21-xpro` and `arduino-zero` tests.

cladmi:
Question, are the `cortexm` linkerscript changes required or could they be separated ?

I may be easier to have first here the minimal patch required for xfa, if there is one and later take care of updating the linkerscript. No need to separate now but at least I am interested to know what is required and I could do the split if this blocks the merge sometime.

I should take care of reviewing and helping with `ld -r` before.

gebart:
Rebased.

The cortex-m linker script refactor is not strictly required, but it would require different changes to the cortex M ldscript to place the XFA sections correctly without this refactoring. With this change, both native and cortex-m can use the core ldscript for XFA. With the current version of the cortexm_common ldscript in master, the XFA sections end up outside of the relocate area in the map, so they are not copied to RAM during libc init.

jia200x:
@gebart what else is needed here? Chances to get it by the soft freeze? (next Tuesday)

gebart:
@jia200x we should focus on #8711 first, it will simplify the implementation of this PR and no/fewer linker hacks necessary to pull in the right symbols in the linking stage.

jia200x:
> @jia200x we should focus on #8711 first, it will simplify the implementation of this PR and no/fewer linker hacks necessary to pull in the right symbols in the linking stage.

Ok, thanks for the information!

cladmi:
It was missing the `Waiting for other PR` label, it is now fixed.

aabadie:
XFA support was merged in #15002, closing.

PR #9108, opened 2018-05-09
tests/xtimer_periodic_wakeup: all states

Josar:
This PR adds additional test for all states in the xtimer_periodic_wakeup() function.
Untill now there was no dependency of the parameter XTIMER_PERIODIC_SPIN and XTIMER_PERIODIC_RELATIVE which ensures to test all possible configuration.

in some cases only spinning and relativ timestamps where tested.

The test is now changed to test spinning, relativ timestamp and also absolut timestamping.

### DIFF
--- a/tests/xtimer_periodic_wakeup/main.c
+++ b/tests/xtimer_periodic_wakeup/main.c
@@ -22,33 +22,52 @@
 #include "xtimer.h"
 #include "periph_conf.h"
 
-#ifndef NUMOF
-#define NUMOF   (256)
-#endif
+#define NUMOF   (3 * XTIMER_PERIODIC_SPIN)
+
 
 int32_t res[NUMOF];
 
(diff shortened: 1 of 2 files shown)


cladmi:
I successfully ran the tests on the following boards:

* `samr21-xpro`: `Min/max error: 7/40`
* `iotlab-`: `Min/max error: 1/21`
* `wsn430-v1_4`: `Min/max error: 14/129`
* `arduino-zero`: `Min/max error: 7/46`

I did not review the code changes though.

Josar:
Ping

PeterKietzmann:
@cladmi I think this is a perfect Hack'n'ACK candidate, don't you think so?

PeterKietzmann:
@cladmi now that you already ran some tests would you mind to provide a review?

PR #9113, opened 2018-05-10
w5100: fixed RX interrupt handling 

enzingerm:
### Fix for W5100 driver RX interrupt handling
When testing with an W5100 ethernet shield I came across an error where the system freezed when receiving packets. While debugging, I noticed that the `ir` variable holding the interrupt register was never updated (read again from the device) which led the loop to run forever and thus freezing my system. With this fix, receiving packets works fine.

### DIFF
--- a/drivers/w5100/w5100.c
+++ b/drivers/w5100/w5100.c
@@ -34,7 +34,6 @@
 #define ENABLE_DEBUG        (0)
 #include "debug.h"
 
-
 #define SPI_CONF            SPI_MODE_0
 #define RMSR_DEFAULT_VALUE  (0x55)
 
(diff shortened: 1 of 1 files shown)


kYc0o:
I have no way to test this and I didn't find the hardware in FU. Should we merge for this release?

dobby880:
We tested it at LMU with a Nucleo F091 and the following board if that is any helpful for you!

https://www.amazon.de/keyestudio-W5100-Ethernet-Erweiterung-Mega2560-mit-Kartenslot/dp/B01GO2JB4S/ref=sr_1_5?ie=UTF8&qid=1531827121&sr=8-5&keywords=arduino+ethernet

kYc0o:
@dobby880 and your test was successful with this PR?

dobby880:
Yes, that's the setup we provided for @enzingerm . 
If you need more tests, I can do them with other boards (arduino m0, nucleo f411, arduino due) and another w5100 module https://www.mouser.de/ProductDetail/WIZnet/WIZ810MJ?qs=sGAEpiMZZMtNlGR3Dbecs0TQSbQSUOhHEWo8PUSa%252bWU%3d but not before end of July

kYc0o:
@PeterKietzmann do you mind to re-ACK?

PR #9114, opened 2018-05-11
add 2018.04 release notes

kaspar030:
### Contribution description

PR adds 2018.04 release notes.


aabadie:
cayenne-lpp package is not in the list

kaspar030:
> cayenne-lpp package is not in the list

yup, fixed

Issue #6857, opened 2017-04-04
cc110x: not working since SPI driver interface rework


patgrosse:
Since #4780 I cannot use networking on the msba2 board. I used the gnrc_networking example on two msba2 boards and I can't even access the shell anymore because the code in the `main()` is not reached. Possibly this is an issue/task for #6437.

Used example: gnrc_networking

With 513b20ffd328934c58af169e2bce0c0a01eddee2:
```
2017-04-04 14:18:45,213 - INFO # Connect to serial port /dev/ttyUSB0
Welcome to pyterm!
Type '/exit' to exit.
2017-04-04 14:18:46,216 - INFO # cc110x: initialized with address=34 and channel=0
[no more console output, shell is not started]
```

With b101c704c7d901275aefea331b28023403328de1 (commit right before):
```
2017-04-04 14:20:29,366 - INFO # Connect to serial port /dev/ttyUSB0
Welcome to pyterm!
Type '/exit' to exit.
2017-04-04 14:20:30,368 - INFO # cc110x: initialized with address=34 and channel=0
2017-04-04 14:20:30,369 - INFO # main(): This is RIOT! (Version: 2017.01-devel-824-gb101c-patrick-ubuntu-HEAD)
2017-04-04 14:20:30,369 - INFO # RIOT network stack example application
2017-04-04 14:20:30,369 - INFO # All up, running the shell now
> 
[shell is started correctly, networking seems good to me]
```

haukepetersen:
Hi, sorry for the late reply! Just did a little digging into the driver: when I introduced some delays into the `cc110x_get_reg_robust()` function (through DEBUG output in my case, the initialization and the shell work fine. So my guess is, that we have some timing issue in the driver.

@kaspar030: would you mind to have a quick look?

haukepetersen:
to go more specific, inserting a simple `DEBUG("foo");` between `cpsr = irq_disable();` and `cc110x_cs(dev);` in `cc110x_get_reg_robust()`, then it works fine, taking out the DEBUG it while compaing `res1` and `res2`...

maribu:
Hi,

the loop

```C
do {
        res1 = spi_transfer_reg(dev->params.spi, SPI_CS_UNDEF,
                                (addr | CC110X_READ_BURST), CC110X_NOBYTE);
        res2 = spi_transfer_reg(dev->params.spi, SPI_CS_UNDEF,
[...]

PR #9124, opened 2018-05-12
boards/msbiot: Configured cc110x as default netdev

maribu:
### Contribution description

boards/msbiot/Makefile.include:
 - Use module "cc110x" if netdev_default or gnrc_netdev_default
   is used

boards/msbiot/include/cc110x_params.h:
 - Added configuration for the CC1101 transceiver


boards/msbiot/Makefile.include
```diff
@@ -14,3 +14,8 @@ export STLINK_VERSION ?= 2
 
 # this board uses openocd
 include $(RIOTMAKE)/tools/openocd.inc.mk
+
+# add driver for CC1101 sub-gigahertz transceiver as default netdev
+ifneq (,$(filter netdev_default gnrc_netdev_default,$(USEMODULE)))
+	USEMODULE += cc110x
```
aabadie: indent should be only 2 spaces

boards/msbiot/include/cc110x_params.h
```diff
@@ -0,0 +1,46 @@
+/*
```
aabadie: A better solution would be to move this file in `drivers/cc110x/include` then using `#ifndef/#define/#endif` it will be possible to set default values in the driver code that can be overridden in the boards. The msba2 could also be updated.
See how it's done with at86rf2xx driver and samr21-xpro/iotlab-m3 boards.

boards/msbiot/Makefile.include
maribu: Thanks for the hint. I will fix this. Is there a reason why RIOT uses 4 spaces in C code, but 2 in Makefiles? This seems inconsistent.

I personally use smart tabs [1](https://www.stream9.net/blog/post/spaces-or-tabs-the-answer-is-both), [2](https://www.emacswiki.org/emacs/SmartTabs) for just this reason: You don't need to argue about indent width, because it because an editor option everyone can configure to her/his liking. And if needed, to different width on a per file type bases.

boards/msbiot/include/cc110x_params.h
maribu: Thanks for the hint. I will update it accordingly.

Btw: Does the cc110x work for those boards? Last time I checked the driver didn't work with either msba2 or msbiot. The driver was developed when `spi_acquire` was a rather cheap operation of acquiring a mutex, not setting up the SPI bus. Before the SPI API was changed so that `spi_acquire` also configured the SPI bus, the driver worked on both msba2 and msbiot. So optimizing the driver to call `spi_acquire` only once at the very beginning of every interaction with the hardware and `spi_release` at the very end might not only improve performance, but also fix the issue with msba2 and msbiot boards.

maribu:
@aabadie: Thanks for pointing out how to improve the pull request. I believe I have addressed the issues you mentioned now.

maribu:
@aabadie: Thanks for pointing out how to improve the pull request. I believe I have addressed the issues you mentioned now.

drivers/cc110x/include/cc110x_params.h
```diff
@@ -0,0 +1,75 @@
...
+extern "C" {
+#endif
+
+/**
+ * @name    Default parameters for the cc110x driver
+ *
+ * These values are based on the msbiot board
+ * @{
+ */
+#ifndef CC110X_SPI
```
aabadie: I would call this parameter `CC110X_PARAM_SPI`, same for other parameters (add `_PARAM` after `CC110X`).

drivers/cc110x/include/cc110x_params.h
```diff
@@ -0,0 +1,75 @@
...
+/** @} */
+
+
+/**
+ * @name    CC110X configuration
+ * @{
+ */
+const cc110x_params_t cc110x_params[] = {
+    {
+        .spi  = CC110X_SPI,  ///< SPI bus the cc110x tranceiver is connected to
```
aabadie: No need for adding doxygen documentation. This struct is already "documented" in the cc110x.h file (I looked at it and it should be updated in a follow PR).

drivers/cc110x/include/cc110x_params.h
```diff
@@ -0,0 +1,75 @@
...
+#endif
+
+#ifndef CC110X_GDO1
+#define CC110X_GDO1                GPIO_PIN(PORT_A,  6)
+#endif
+
+#ifndef CC110X_GDO2
+#define CC110X_GDO2                GPIO_PIN(PORT_C,  5)
+#endif
+/** @} */
```
aabadie: Please also add a `CC110X_PARAMS` define, like it's done in at86rf2xx driver. This way it allows a developer to define a global params define for multiple devices. With your change, this is not possible.

boards/msba2/include/board.h
```diff
@@ -55,6 +55,17 @@ extern "C" {
 #define XTIMER_SHOOT_EARLY  3
 /** @} */
 
+/**
+ * @name    Configure conntect CC1101 (radio) device)
+ * @{
+ */
+#define CC110X_SPI          0
```
aabadie: Use `SPI_DEV(0)`

boards/msba2/include/board.h
```diff
@@ -55,6 +55,17 @@ extern "C" {
...
+#define CC110X_CS           53
```
aabadie: Those values are weird, is it possible to find something with GPIO_PIN(PORT_NUM, PIN_NUM) ? I know it's out of the scope of this PR, I'm just commenting

maribu: In `cpu/lpc2387/include/periph_cpu.h:57` I found this
```C
#define GPIO_PIN(port, pin) (port*32 + pin)
```

So I guess it should be like this
```
CS   = 53 = GPIO_PIN(PORT_B, 21)
GDO0 = 27 = GPIO_PIN(PORT_A, 27)
GDO1 = 55 = GPIO_PIN(PORT_B, 23)
GDO2 = 28 = GPIO_PIN(PORT_A, 28)
```

I will check this tomorrow on the hardware. Would be nice to have no magic numbers in the code.

maribu: No need to test, this will obviously yield the same value and thus has to be correct.

maribu: I think this is better:

```
CS   = 53 = GPIO_PIN(1, 21)
GDO0 = 27 = GPIO_PIN(0, 27)
GDO1 = 55 = GPIO_PIN(1, 23)
GDO2 = 28 = GPIO_PIN(0, 28)
```

This would be consistent with the rest of the MSBA2 `board.h` and the pin labels at the board diagram.

board diagram

aabadie:
@maribu, I allowed myself to push an extra commit in your branch that changes your PR to what I have in mind:
- the default params of cc110x uses the values for msba2, since it was already the case before
- only board.h in msbiot needs to have the param macro defined

What would be great now is that you squash your commits in 2:
- one related to the changes in msba2 and the move of the params.h file to drivers
- one related to the msbiot board change

Then it will be perfect. Maybe someone else could also test that the msba2 board is still working (I don't have one).

maribu:
@aabadie: Done.

The reviewing will sadly be a bit problematic:
 - I'm 100% sure that all existing msbiot boards are in my office
 - There aren't too many MSBA2 boards as well - maybe our work group is also the proud owner of all remaining and operational boards?
- The cc110x driver is known to not work [since more than a year](https://github.com/RIOT-OS/RIOT/issues/6857). So actually checking that this PR doesn't break anything requires the fix first
     - Hopefully I'll have some time to work on that soon

maribu:
I was wrong, sort of:

```
--- fe80::ff:fe00:51 ping statistics ---
100 packets transmitted, 10 received, 90% packet loss, time 189.06358086 s
rtt min/avg/max = 16.621/1.663/16.650 ms
```

Above when using the original pull request (without moving cc110x_params.h to the driver) over a distance of ~10cm.

```
main(): This is RIOT! (Version: 2018.07-devel-195-g33116-faultier2go)
gcoap example app
All up, running the shell now
> cc110x: initialized with address=81 and channel=0
[CC110X spi] fatal error
```

With the current PR. Over SWD I get the following backtrace:
```
#1  0x08000912 in _shoot (timer=<optimized out>) at /home/maribu/Repos/software/RIOT/sys/xtimer/xtimer_core.c:159
```

I do believe that the PR does not contain the bug that causes the fatal error, but merely triggers it. But still it might be reasonable to not merge the PR until the cc110x issue is resolved

aabadie:
> But still it might be reasonable to not merge the PR until the cc110x issue is resolved

+1

maribu:
Of course it was an evil typo. After having that fixed, it works again (at least as bad as before...)

maribu:
Of course it was an evil typo. After having that fixed, it works again (at least as "good" as before, with 90% loss rate).

I will check for the MSBA2 tomorrow as well. But I believe the MSBA2 is not even booting since the update of the SPI API.

maribu:
Hi,

I solved the issue with the MSBA2 not booting with the cc110x driver in PR https://github.com/RIOT-OS/RIOT/pull/9154. Once PR https://github.com/RIOT-OS/RIOT/pull/9154 is applied, the cc110x driver works with and without this PR. So I can confirm, that this PR does  not break thinks.

Below is the output of one of our MSBA2 boards after applying this PR and PR https://github.com/RIOT-OS/RIOT/pull/9154 on top:

```
2018-05-19 11:40:51,393 - INFO # Connect to serial port /dev/ttyUSB0
Welcome to pyterm!
Type '/exit' to exit.
2018-05-19 11:40:52,401 - INFO # random: NO SEED AVAILABLE!
2018-05-19 11:40:52,407 - INFO # cc110x: initialized with address=34 and channel=0
2018-05-19 11:40:52,414 - INFO # main(): This is RIOT! (Version: 2018.07-devel-224-g30a4e-faultier2go-msbiot)
2018-05-19 11:40:52,417 - INFO # gcoap example app
2018-05-19 11:40:52,420 - INFO # All up, running the shell now
> 2018-05-19 11:40:52,425 - INFO #  ping6 fe80::ff:fe00:22
2018-05-19 11:40:52,433 - INFO # 12 bytes from fe80::ff:fe00:22: id=83 seq=1 hop limit=64 time = 0.351 ms
2018-05-19 11:40:53,431 - INFO # 12 bytes from fe80::ff:fe00:22: id=83 seq=2 hop limit=64 time = 0.351 ms
2018-05-19 11:40:54,468 - INFO # 12 bytes from fe80::ff:fe00:22: id=83 seq=3 hop limit=64 time = 0.350 ms
2018-05-19 11:40:54,471 - INFO # --- fe80::ff:fe00:22 ping statistics ---
2018-05-19 11:40:54,477 - INFO # 3 packets transmitted, 3 received, 0% packet loss, time 2.0620426 s
2018-05-19 11:40:54,481 - INFO # rtt min/avg/max = 0.350/0.350/0.351 ms
```

Kind regards,
Marian

drivers/cc110x/include/cc110x_params.h
```diff
@@ -0,0 +1,80 @@
...
+                                        .gdo1 = CC110X_PARAM_GDO1, \
+                                        .gdo2 = CC110X_PARAM_GDO2, \
+                                    }
+
+#endif
+/** @} */
+
+
+/**
+ * @name    CC110X configuration
```
aabadie: Missed that one. Here it should be `@brief` and the `@{`, `/** @} */` below can be removed.

maribu: Fixed :-)

maribu:
Hi,
I can confirm that the cc110x still works fine on the MSBA2 after this PR. For the MSB IoT it still works as good as before (90% loss rate at ping).

Regarding the MSB IoT: I have figured out some timing issues in the driver. The MSB IoT seems to be faster and enables interrupts on the GDO2 control pin before the CC1101 transceiver applies the new configuration. Because the interrupt triggers on both rising and falling edge, it will can off two times: Let's say the GDO2 signals by falling edge that more data can be uploaded to the TX fifo but the GDO2 was low before, it would have to go high first and later low again. Thus, only the second edge should trigger an interrupt, but the MSB IoT might run the ISR two times. Sadly those timing issues aren't the only problem. It seems the SPI transfer to the MSB IoT is not working completely reliable. I hope to figure that out as well. (It might have to do with how the CS is done.) Maybe I can solve that next weekend.

Kind regards,
Marian

PR #7165, opened 2017-06-09
kinetis: Use 9 bit UART config for 8E1, 8O1 modes

mfrey:
Pull request GH-7098 introduced the "8 data bits, even parity, 1 stop bit" mode to the Kinetis UART configuration. Unfortunately, the configuration misses the correct configuration since the current setting reduces the number of actual data bits to 7. Hence, the 9 bit mode needs to be enabled (so it becomes 8 data bits + 1 parity bit).

The following snippet is from the [MKW2xD Reference Manual](http://www.nxp.com/assets/documents/data/en/reference-manuals/MKW2xDxxxRM.pdf) and specifies the required changes.
image

PR #9127, opened 2018-05-13
cpu: efm32: provide non-standard UART modes

basilfx:
### Contribution description
I need non-standard UART modes for some (KNX) transceivers that want 8E1. This PR adds them as an option to the `uart_config` struct, similar to how it was done for some Kinetis boards. The default remains 8N1, for both USART and LEUART. 

Conversion methods were added to Gecko-SDK (see [this commit](https://github.com/basilfx/RIOT-gecko-sdk/commit/412328e299f6b9352f295257d5d86e40615c5963)), which minimizes the changes in the RIOT code and provides a clean abstraction for both USART and LEUART.

### Issues/PRs references
#5899, #7165

### DIFF
--- a/cpu/efm32/periph/uart.c
+++ b/cpu/efm32/periph/uart.c
@@ -26,8 +26,10 @@
 #include "periph/gpio.h"
 
 #include "em_usart.h"
+#include "em_usart_utils.h"
 #if LOW_POWER_ENABLED && defined(LEUART_COUNT) && LEUART_COUNT > 0
 #include "em_leuart.h"
+#include "em_leuart_utils.h"
 #endif
 
 /**
@@ -73,6 +75,11 @@ int uart_init(uart_t dev, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
 
         init.enable = usartDisable;
         init.baudrate = baudrate;
+#if EFM32_UART_MODES
+        init.databits = USART_DataBits2Def((uart_config[dev].mode >> 0) & 0xf);
+        init.stopbits = USART_StopBits2Def((uart_config[dev].mode >> 4) & 0xf);
+        init.parity = USART_Parity2Def((uart_config[dev].mode >> 8) & 0xf);
+#endif
 
         USART_InitAsync(uart, &init);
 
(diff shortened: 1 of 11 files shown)


basilfx:
Ping @kYc0o

cpu/efm32/periph/uart.c
```diff
@@ -104,6 +109,9 @@ int uart_init(uart_t dev, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
 
         init.enable = leuartDisable;
         init.baudrate = baudrate;
+        init.databits = LEUART_DataBits2Def((uart_config[dev].mode >> 0) & 0xf);
+        init.stopbits = LEUART_StopBits2Def((uart_config[dev].mode >> 4) & 0xf);
+        init.parity = LEUART_Parity2Def((uart_config[dev].mode >> 8) & 0xf);
```
kYc0o: So, the baud rate will only change on init, then it stays like that as I see? Which means once the UART is attached to a device it cannot be detached is it?

pkg/gecko_sdk/Makefile
```diff
@@ -1,6 +1,6 @@
 PKG_NAME=gecko_sdk
 PKG_URL=https://github.com/basilfx/RIOT-gecko-sdk
-PKG_VERSION=d381e526d68a2d0c951f37040c1c2e168ac66cd6
+PKG_VERSION=ee01ab57eaaa8805916601dd444d1b63f03037de
```
kYc0o: I see there's an update to the gecko sdk. Is this intended here?

basilfx: Yeah, the update is intended, as they provide the utility methods to convert bits to enums.

cpu/efm32/periph/uart.c
basilfx: Yes, that's true. #5899 proposes a runtime option, but it is not merged/fixed/proposed.

STM32 do the same, iirc.

kYc0o:
I also wonder how is this configurable at compile time. The only way to change the behaviour is to modify the source code, which seems not super practical. Maybe you can add an `#ifdef` on the headers (which would be also a kind of API change, but only as a parameter) and then pass the config through CFLAGS? Otherwise even if there are reasons to have this PR in, it doesn't really give the option and the behaviour stays unchanged (with some additional bytes to the final binary because of the new operations).

basilfx:
@kYc0o As we discussed on IRC, I have updated this PR to make this feature optional (in favor of code size, because the current boards/configuration doesn't make use of this feature). I did not make it configurable for each UART interface, since I think that if this is a desired feature, we should fix #5899.

I opted for a solution that I can reuse to make other parts of EFM32 optional, like disabling support for low-power peripherals (if not used). So the last two commits may seem like a lot, I want to re-use them and provide a uniform interface/place for EFM32 specific features/toggles.

basilfx:
Squashed and rebased.

PR #9129, opened 2018-05-14
mtd_spi_nor: add 4-byte address flash support

vincent-d:
### Contribution description

Add opcodes for 4-byte addresses commands with a SPI NOR flash, this is needed for flashes bigger than 128Mb.
The commands come from MX25L25645G datasheet.

### Issues/PRs references

None

### DIFF
--- a/drivers/include/mtd_spi_nor.h
+++ b/drivers/include/mtd_spi_nor.h
@@ -140,6 +140,13 @@ extern const mtd_desc_t mtd_spi_nor_driver;
  * sensible for default values. */
 extern const mtd_spi_nor_opcode_t mtd_spi_nor_opcode_default;
 
+/**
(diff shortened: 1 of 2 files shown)


vincent-d:
ping, this one is really small

gebart:
These memories are annoying. The older MX25L25635E uses a separate "enable 4-byte mode" command, but uses the same command set as the already existing default opcode list, the MX25L25735E does not have the EN4B command, but uses 4 byte addressing with the default opcodes too. And then this memory has separate opcodes for 4 byte addressing

vincent-d:
I see...

Can we maybe at least merge this as-is and improve later for this kind of memories? I guess a new flag would be needed to enable sending the EN4B command first. And we would also need another set of commands.

gebart:
Untested ACK

PR #8811, opened 2018-03-21
Basic support for CC430 radio module

baptiste-cartier:
### Contribution description

This is a basic implementation for the cc430 radio module, enabling the use of this cpu radio capabilities. It comes with a configuration example file with two example configurations, one in transmit mode and one in receive mode.

The goal of this implementation is to give a fast and simple way to use the cc430 radio module for basic communications.

PR #9131, opened 2018-05-15
driver : Netdev implementation for cc430

baptiste-cartier:
### Contribution description

Netdev implementation for the cc430 cpu, based on the cc110x driver.

cc430 has a integrated cc1101-like radio module which is used through register manipulation and does not use any communication bus (SPI...) , therefor the cc110x driver can not be used.

This netdev implementation heavily based on the cc110x driver for RIOT, most of the credit goes to the authors of the  cc110x driver.

This implementation should offer the same capabilities as the cc110x driver, and responds to the same netopt_t options.

### DIFF
--- a/cpu/cc430/Makefile
+++ b/cpu/cc430/Makefile
@@ -2,4 +2,9 @@ MODULE = cpu
 
 DIRS = $(RIOTCPU)/msp430_common periph
 
+# cc430_rf radio driver
+ifneq (,$(filter cc430_rf,$(USEMODULE)))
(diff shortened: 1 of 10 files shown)


PeterKietzmann:
@OlegHahm, @kaspar030 aren't you familiar with that device?

PeterKietzmann:
@baptiste-cartier commit-wise I'm not entirely sure but I assume this PR depends on #8811, is that correct?

tcschmidt:
Ping @baptiste-cartier

PR #9137, opened 2018-05-15
 tests/unittests: add empty list clist_foreach case 

cladmi:
### Contribution description

Add clist_foreach empty list test case which should:

* not call `func`
* return `NULL`

### Issues/PRs references

Follow up to and depending on #9017

### DIFF
--- a/tests/unittests/tests-core/tests-core-clist.c
+++ b/tests/unittests/tests-core/tests-core-clist.c
@@ -237,13 +237,19 @@ static int _foreach_test_trampoline(clist_node_t *node, void *arg)
 
 static void test_clist_foreach(void)
 {
+    void *res;
     list_node_t *list = &test_clist;
+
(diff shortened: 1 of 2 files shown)


aabadie:
Please rebase (#9017 was just merged)

tests/unittests/tests-core/tests-core-clist.c
```diff
@@ -237,13 +237,19 @@ static int _foreach_test_trampoline(clist_node_t *node, void *arg)
 
 static void test_clist_foreach(void)
 {
+    void *res;
```
aabadie: why not moving the declaration of `res` line 24 ?

cladmi: It is used three times after to do `res = clist_foreach(...)`, so it makes these lines the same.

PR #6837, opened 2017-03-31
atmega328p/startup.c: Clear watchdog flag

zqad:
When the atmega328p is restarted by the watchdog, the WDRF bit in
the MCUSR register will also set the WDE bit, enabling the
watchdog. Since we use the watchdog for reboot (see
cpu/atmega_common/periph/pm.c), this means that the CPU will just
loop-reboot when a reboot is issued.

From the Atmel documentation about WDE in WDTCSR:
  "WDE is overridden by WDRF in MCUSR. This means that WDE is
  always set when WDRF is set. To clear WDE, WDRF must be cleared
  first. This feature ensures multiple resets during conditions
  causing failure, and a safe start-up after the failure."

PR #9141, opened 2018-05-15
boards/common/atmega: initialize watchdog

### LINKED ISSUE #6836: Issuing a reboot on ATmega328p makes the board loop-reboot
Consider this sample program:
```C
#include <stdio.h>
#include <periph/pm.h>

int main(void)
{
	puts("Rebooting!\n");
	pm_reboot();
	return 0;
}
```

Running it on the board gives this result on my serial console:
```main(): This is RIOT! (Version: 2017.04-devel-638-gdf32c-lisen)
Rebooting!



main(): This 
             
main(): This é
              
main(): This 
             
main(): This é
```

The error is due to the usage of the watchdog as the reboot method, without properly clearing the WDRF bit in MCUSR. This means that the watchdog will be activated with a 250ms timeout on the next reboot.

I have prepared a patch which I will add as a pull request to this issue. Feel free to give feedback on it, in case you find anything noteworthy with it.

Thanks,
/Jonas

ZetaR60:
This initializes the watchdog on ATmega boards if a watchdog reset occurred, preventing reboot loops when using the reboot command.

Adapted from #6837 

Fixes #6836

### DIFF
--- a/boards/common/atmega/board.c
+++ b/boards/common/atmega/board.c
@@ -31,6 +31,7 @@ void led_init(void);
 
 void board_init(void)
 {
+    atmega_wdog_init();
     atmega_set_prescaler(CPU_ATMEGA_CLK_SCALE_INIT);
     atmega_stdio_init();
     cpu_init();
--- a/cpu/atmega_common/include/cpu.h
+++ b/cpu/atmega_common/include/cpu.h
@@ -124,6 +124,23 @@ static inline void atmega_set_prescaler(uint8_t clk_scale)
 }
 
 /**
+ * @brief   Initializes watchdog
+ */
+static inline void atmega_wdog_init(void)
+{
+    /* Reset the watchdog if the reboot was triggered due to it */
+    if (MCUSR & (1 << WDRF)) {
+        MCUSR &= ~(1 << WDRF);
+
+        /* Enable watchdog change */
+        WDTCSR |= 1 << WDCE;
+
+        /* Clear settings within 4 cycles */
+        WDTCSR = 0;
+    }
+}
+
+/**
  * @brief   Initializes avrlibc stdio
  */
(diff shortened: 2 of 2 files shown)


ZetaR60:
Note: @mali reports that this PR does not fix #6836 on some boards, for unknown reasons.

ZetaR60:
If I am not mistaken, interrupts are enabled for the first time only after atmega_wdog_init is being run (at the end of board_init).

mali:
@ZetaR60 , you're right about interrupts. Any thoughts about the WDE bit ?

ZetaR60:
@mali WDE should be cleared when all of WDTCSR is set to 0.

I have an idea: What if some versions of the Arduino bootloader are clearing MCUSR / WDRF, but not WDTCSR ? The atmega_wdog_init function would do nothing (because it does not know the board was watchdog reset), but it would still have the watchdog set to reboot again. Could you try #9141 again, but with the condition removed (so that atmega_wdog_init always clears the watchdog no matter what)?

ZetaR60:
@Josar I am not very familiar with the section attribute and how this is modifying the startup process. Could you submit an alternate PR so that we could test the method as you intend it to be written? I think that since there is no established API for getting the reset cause it should be a secondary priority (unless a specific board needs it, like Jiminy), and the main goal should be getting all CPUs to reset properly.

ZetaR60:
ping @Josar

Josar:
@ZetaR60 i will have a look into it. Maybe after https://github.com/RIOT-OS/RIOT/pull/9130 was merged as this will place all CPU depended stuff in one place and then the `init` section can be in one place for all CPUs. Maybe its then just a realocating of the else.

ZetaR60:
Fixed by #9866. Closing.

PR #9142, opened 2018-05-15
Interrupt context switch

Josar:
This is a PR to reduce overhead for platforms which allow ISR context swap.

As for now the procedure is as follows:
1. ISR enter
1. Calback
1. Do something, thread_yield, check for ISR, signalize with flag, return to caller.
1. Check for flag, swap context.

This can be reduced.
1. ISR enter
1. Calback
1. Do something, swap context.

As of now the feature can be enabled with a CFLAG.
```
CFLAGS= -DISR_CONTEXT_SWITCH_ALLOWED
```

and only the atmega platform was changed accordingly.

Following tests run successfull.
- tests/xtimer_msg_receive_timeout
- tests/xtimer_periodic_wakeup
- tests/xtimer_remove
- tests/xtimer_remove
- tests/xtimer_reset
- tests/xtimer_usleep
- tests/xtimer_usleep_short

The question is, how to implement this that it works for all platforms or can be enabled and disabled.
Where to place `ISR_CONTEXT_SWITCH_ALLOWED` ?
As for now it has to be placed at the test or example Makefile.
Any suggestions are welcome.

This is a followup of https://github.com/RIOT-OS/RIOT/pull/8988 

This is WIP. 

All comparison for `irq_is_in()` can be moved to `thread_yield_higher` thus moving the check from multiple places to the dependend platform.

### DIFF
--- a/core/msg.c
+++ b/core/msg.c
@@ -165,7 +165,7 @@ static int _msg_send(msg_t *m, kernel_pid_t target_pid, bool block, unsigned sta
         sched_set_status(target, STATUS_PENDING);
 
         irq_restore(state);
-        thread_yield_higher();
+
     }
 
     return 1;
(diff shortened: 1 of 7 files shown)


Josar:
This changes can not work with the current implementation of the xtimer.
As the xtimer unlocks multiple mutex and the ISR context will be lost if there are multiple timer to shoot in the same ISR.
Also the next compare value is not set before unlocking, thus changing the context.

PR #9145, opened 2018-05-16
trickle: remove msg_time member from struct

bergzand:
xtimer_set_msg64 offset is in μs and trickle is in ms, the timer values
should thus be multiplied with `US_PER_MS` and not with `MS_PER_SEC`.

Doesn't change any functionality, but prevents confusion with the timer
units.

### DIFF
--- a/sys/trickle/trickle.c
+++ b/sys/trickle/trickle.c
@@ -51,8 +51,8 @@ void trickle_interval(trickle_t *trickle)
     /* old_interval == trickle->I / 2 */
     trickle->t = random_uint32_range(old_interval, trickle->I);
 
-    trickle->msg_time = (trickle->t + diff) * MS_PER_SEC;
-    xtimer_set_msg64(&trickle->msg_timer, trickle->msg_time, &trickle->msg,
+    uint64_t msg_time = (trickle->t + diff) * US_PER_MS;
+    xtimer_set_msg64(&trickle->msg_timer, msg_time, &trickle->msg,
                      trickle->pid);
 }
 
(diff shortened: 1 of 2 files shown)


sys/trickle/trickle.c
```diff
@@ -51,7 +51,7 @@ void trickle_interval(trickle_t *trickle)
     /* old_interval == trickle->I / 2 */
     trickle->t = random_uint32_range(old_interval, trickle->I);
 
-    trickle->msg_time = (trickle->t + diff) * MS_PER_SEC;
+    trickle->msg_time = (trickle->t + diff) * US_PER_MS;
```
cladmi: I agree that from the documentation, `trickle->msg_time` is in `ms`, and that `xtimer_set_msg64` `offset` is in `µs`. However in that case it should be multiplied by `USEC_PER_MS` only for the `xtimer_set_msg64` call not when setting `trickle->msg_time`.

I do not really get what is the the `t + diff` unit is however. From what I see by doing grep, nothing else is using `msg_time` in the repository at the moment.

bergzand: Maybe @cgundogan can clarify why the `msg_time` is a member of the struct.

I agree that from the documentation, `msg_time` should be in ms, I can take another look to see if there are more units that can be clarified here.

bergzand: > why souldn't msg_time be a member of the struct? If several applications use trickle concurrently, then a distinct msg_time context is needed for all of them

Because it is only used in the `trickle_interval` interval function and only after the new value is calculated from the trickle struct values.

cgundogan: >I do not really get what is the the t + diff unit is however.

it's in `ms`.

cgundogan: @bergzand ah now I see! Yeah, in a previous iteration we used to timers to trigger the "new interval phase" and the "trigger callback phase". I would be fine removing that from the struct.

bergzand:
@cgundogan Removed the msg_time struct member.

bergzand:
Squashed. I've reworded the commit message to reflect the new changes

cgundogan:
What's up with travis?

PR #9146, opened 2018-05-16
tests/xtimer_now64: Start time measure after print

bergzand:
Printing can take a lot of time over a serial connection. On the atmega board it takes more than 2 ms, skewing the first measurement over the threshold value and causing a test failure. 

This PR moves the print statement above the first measumerement to prevent the skew.


bergzand:
> Makes somewhat sense, but could also simply remove the START output and adapt the test script accordingly.

Also a possible solution yes. Do you have any preference for either solution?

PR #9147, opened 2018-05-16
kinetis: GPIO: Enable the correct IRQn on CM0+

gebart:
### Contribution description

Address a problem where gpio_init_int activates the wrong IRQn in the NVIC on KW41Z when trying to set up interrupts on PORTC.

### Issues/PRs references

none

### DIFF
--- a/cpu/kinetis/periph/gpio.c
+++ b/cpu/kinetis/periph/gpio.c
@@ -113,6 +113,7 @@ static isr_ctx_t isr_ctx[CTX_NUMOF];
  */
 static uint32_t isr_map[ISR_MAP_SIZE];
 
+static const uint8_t port_irqs[] = PORT_IRQS;
 
 static inline PORT_Type *port(gpio_t pin)
 {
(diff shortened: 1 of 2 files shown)


cpu/kinetis/periph/gpio.c
```diff
@@ -218,8 +218,21 @@ int gpio_init_int(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
...
+        /* PORTA gets its own IRQ */
+        NVIC_EnableIRQ(PORTA_IRQn);
+    }
+    else {
+        /* All other ports are combined on the next IRQ number */
+        NVIC_EnableIRQ(PORTA_IRQn + 1);
+    }
+#else
+    /* Cortex-M4 parts have one IRQ per port */
     NVIC_EnableIRQ(PORTA_IRQn + port_num(pin));
```
smlng: all Kinetis header define `PORT_IRQS` so you could use a generic `NVIC_EnableIRQ(PORT_IRQS[port_num(pin)]);`, instead of this rather ugly `#ifdef` construct, or not? see https://github.com/RIOT-OS/RIOT/blob/master/cpu/kinetis/include/vendor/MKW41Z4.h#L5424 for instance

gebart: I had never noticed that before, thanks! I will update accordingly

smlng: 😄 me neither, but as I quickly grep'ed through the code I stumbled over it and it look kind of useful and exactly what would fit here...

gebart:
Updated, not yet tested on actual hardware

gebart:
tested on frdm-kw41z OK after change to PORT_IRQS

PR #9151, opened 2018-05-17
core/debug: fix debug.h use within ISRs

ZetaR60:
If you attempt to use DEBUG messages within ISRs on some platforms (tested on mega-xplained) it complains "Cannot debug, stack too small". This PR allows DEBUG messages to be used.


kYc0o:
Let's go then!

PR #9153, opened 2018-05-18
pkg/lua: Better Lua-RIOT integration

jcarrano:
### Better integration of the Lua interpreter with RIOT

The main idea behind this PR is that interpreted code should never be allowed to crash the host application. A secondary goal is to cover the majority of use cases in a few simple API calls, to remove the required boilerplate and make it easier to use the Lua package correctly.

- The lua intepreter itself, with some custom modifications:
  * Patches to reduce C-stack usage.
  * Modifications to the test module to enable it to work on RIOT.
  * The vanilla module loader is disabled as it is too dependent on files and dynamic loading.
- A new module loader intended to load source code and C-modules built-into the application.
- A custom memory allocator based on TLSF. This allows to limit the memory that Lua can consume and to not have memory leaks in case of an interpreter crash.
- A panic handler based on `setjmp()/longjmp()`. Thus, in case of a irrecoverable failure in the interpreter, control returns to the host application (by default, lua calls `abort()`)
- An example with an interactive Lua REPL.
- Procedures to easily launch Lua and handle errors.

**Note that only 32 bit CPUs will be supported.**

### Future work

The next PRs will focus on build system integration (i.e. being able to compile ``%.lua`` into an application).

### About the Lua patches

I'm patching lua, in a way that is perfectly compatible with the unmodified code. I'm running the test suite, and it passes. Patches are still patches, and they will never get merged into upstream lua, simply because the Lua people don't accept outside submissions (check the [FAQ](https://www.lua.org/faq.html#1.8) for more info). This is a reality we have to accept. Other projects do the same, they apply their own modifications.

### Issues/PRs references

Extends: #8847
Depends on: #9090 #9006 #9244


smlng:
please rebase

jcarrano:
Why is the Murdock state stuck in "Waiting for status to be reported"?

cgundogan:
@jcarrano Murdock only starts the build, once the `ready for CI build` flag is set

cpu/saml21/periph/pm.c
```diff
@@ -38,6 +38,7 @@ void pm_set(unsigned mode)
                 DEBUG("pm_set(): setting STANDBY mode.\n");
                 _mode = PM_SLEEPCFG_SLEEPMODE_STANDBY;
                 break;
+            default:
```
jia200x: This should be in a separate PR (gcc7 complaining?)

examples/lua_REPL/Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+BOARD_WHITELIST := frdm-k64f iotlab-m3 mulle nrf52840dk nucleo-f411re \
+				   stm32f4discovery
+
+# Comment this out to disable code in RIOT that does safety checking
+# which is not needed in a production environment but helps in the
+# development process:
+DEVELHELP ?= 1
+
+CFLAGS_OPT = -Og
```
jia200x: why? :)

examples/lua_REPL/main.c
```diff
@@ -0,0 +1,50 @@
+/*
```
jia200x: maybe this could be in the same `lua` example. What do you think?

jia200x: (basically the LUA Repl implementation is running LUA. So, should be enough to show how it works)

pkg/lua/Makefile
```diff
@@ -1,11 +1,11 @@
 PKG_NAME=lua
 PKG_URL=https://github.com/lua/lua.git
-PKG_VERSION=e354c6355e7f48e087678ec49e340ca0696725b1
+PKG_VERSION=v5-3-4
```
jia200x: is this v5-3-4 a git identifier?

pkg/lua/contrib/binsearch.c
```diff
@@ -0,0 +1,48 @@
...
+        int cmp = strncmp(str, target, n);
+
+        if (cmp == 0)
+            return mid;
+        else if (cmp < 0)
+            hi = mid;
+        else /* (cmp > 0) */
+            lo = mid + 1;
+    }
+    return BINSEARCH_NOTFOUND;
```
jia200x: if possible, prefer `errno` types

pkg/lua/contrib/lua_loadlib.c
```diff
@@ -0,0 +1,282 @@
...
+    if (load_result == LUAR_MODULE_NOTFOUND) {
+        load_result = _ll_searcher_builtin_c(L, name);
+    }
+
+    return load_result;
+}
+
+/* ======================== 'require' function ============================= */
+
+
```
jia200x: please remove extraline

pkg/lua/contrib/lua_loadlib.c
```diff
@@ -0,0 +1,282 @@
...
+            lua_pop(L, 1);              /* remove extra return */
+            luaL_addvalue(&msg);        /* concatenate error message */
+        }
+        else {
+            lua_pop(L, 2); /* remove both returns */
+        }
+    }
+}
+
+
```
jia200x: and here :)

sys/newlib_syscalls_default/syscalls.c
```diff
@@ -461,6 +463,23 @@ int _unlink_r(struct _reent *r, const char *path)
 }
 #endif /* MODULE_VFS */
 
+/**
```
jia200x: can you add these functions in a separate PR?

cpu/saml21/periph/pm.c
jcarrano: Yes, it is in #9244. As soon as that is merged, I will rebase and that change should disappear.

examples/lua_REPL/Makefile
jcarrano: To make it easier to debug the lua implementation and find the exact lines where error appeared. I can remove it.

pkg/lua/Makefile
jcarrano: Yes, it is a tag.

examples/lua_REPL/main.c
jcarrano: Yes, the thing is the REPL requires more resources, so it runs on less boards (part of that is because I'm enabling the lua test module).
My idea was to have the lua example showcase the basic API, and another example with a more complex app.

sys/newlib_syscalls_default/syscalls.c
jcarrano: It's in #9090

cpu/saml21/periph/pm.c
jia200x: ok!

examples/lua_REPL/main.c
jia200x: check the OpenThread test at #7149. It's possible to switch between Full Thread Device and Minimum Thread Device with a Makefile flag.

examples/lua_REPL/Makefile
jia200x: I would either remove it or add a comment on top

pkg/lua/Makefile
jia200x: ok!

sys/newlib_syscalls_default/syscalls.c
jia200x: ok!

pkg/lua/contrib/binsearch.c
jcarrano: I searched, but I cannot find an errno code that matches (they concern mostly external errors). Anyways, there is only one error that can be thrown by a search routine: "not found".

examples/lua_REPL/main.c
jcarrano: Are you suggesting I test the type of device and run the REPL or the Hello World accordingly?

jia200x: yes, something like that. Like:

```
/* call hello world */
#ifdef MODULE_LUA_REPL
/* call Lua repl module */
#endif
```

jcarrano: Then I will have two programs in one, which is what happens each time one has an `#ifdef`

jia200x: Yes. I tend to prefer one example showing key functionalities than several. It's only one opinion though, so maybe would be nice to hear other people here too.

MrKevinWeiss: I think it should be two separate examples if it is already done that way.  More obvious when looking at examples.  +1 to original implementation.

pkg/lua/contrib/binsearch.c
jia200x: ENOTFOUND should do the trick :)

examples/lua_REPL/main.c
jia200x: ok, let's leave it like that then. We can discuss it later in further PRs and don't block this one

jia200x: (or what @jcarrano thinks the best)

pkg/lua/contrib/binsearch.c
jcarrano: `ENOTFOUND` is not in `errno.h` and is not defined in RIOT either.

jia200x: you can use ENOENT as seen [here](https://github.com/RIOT-OS/RIOT/blob/af2c1538898db82006e006e43c3af5bbd099e8b4/sys/net/gnrc/netif/hdr/gnrc_netif_hdr.c#L65)

pkg/lua/include/lua_builtin.h
```diff
@@ -0,0 +1,66 @@
...
+ *
+ */
+
+#ifndef LUA_BUILTIN_H
+#define LUA_BUILTIN_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
jia200x: double empty line here

examples/lua_REPL/main.c
danpetry: +1 for separate examples for clarity reasons. Could rename `example/lua` to `example/lua_hello-world` or something. In general I think it's more important that the examples are all concise, useful and non-overlapping than that they are simply small in number

pkg/lua/contrib/binsearch.c
jcarrano: Mmmhh, I disagree with that usage. [The definition](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html) says it means "No such file or directory". I'll change anyways, it if is a sort of convention in RIOT to use it that way.

pkg/lua/contrib/lua_run.c
```diff
@@ -0,0 +1,275 @@
...
+ */
+static void *lua_tlsf_alloc(void *ud, void *ptr, size_t osize, size_t nsize)
+{
+    tlsf_t tlsf = ud;
+    (void)osize;
+
+    return tlsf_realloc(tlsf, ptr, nsize);
+}
+
+LUALIB_API lua_State *luaR_newstate(void *memory, size_t mem_size,
```
jia200x: @jcarrano @danpetry I'm not sure about these luaR_xxx functions. Are these Lua or RIOT specific? According to our [Coding Conventions](https://github.com/RIOT-OS/RIOT/wiki/Coding-conventions), RIOT code should be always snake_case( It's OK to call CamelCase functions though if they are defined in the original pkg)


What do you think?

jia200x: (it's also OK to write CamelCase when it's the implementation of a function required by the pkg, but couldn't find any luaR functions reference on internet)

pkg/lua/include/lua_run.h
```diff
@@ -0,0 +1,225 @@
...
+ *                      Currently this functionality is not supported and this
+ *                      must be set to 0.
+ * @param   panicf      Function to be passed to lua_atpanic. If set to NULL,
+ *                      a generic function that does nothing will be used.
+ *
+ * @return      the new state, or NULL if there is a memory allocation error.
+ */
+LUALIB_API lua_State *luaR_newstate(void *memory, size_t mem_size,
+                                    lua_CFunction panicf);
+
```
jia200x: double empty line

examples/lua_REPL/main.c
jia200x: I like @danpetry approach

examples/lua_REPL/README.md
```diff
@@ -0,0 +1,20 @@
+### About
```
jia200x: Could you provide a readme with the available commands? I ran the example but don't know what to do :)

examples/lua_REPL/Makefile
```diff
@@ -0,0 +1,53 @@
+APPLICATION = lua_repl
+
+# If no BOARD is found in the environment, use this default:
+BOARD ?= native
+
+# This has to be the absolute path to the RIOT base directory:
+RIOTBASE ?= $(CURDIR)/../..
+
+BOARD_WHITELIST := frdm-k64f iotlab-m3 mulle nrf52840dk nucleo-f411re \
```
jia200x: this test runs on `native` too. You should add it to BOARD_WHITELIST

pkg/lua/contrib/lua_run.c
jcarrano: There is no "nice" solution. I chose those names because those are replacements for `luaL_newstate`, `lual_openlibs`, etc, and the convention in lua is to have lua_xxxx for core functions and luaY_xxx for other functions.

pkg/lua/include/lua_builtin.h
jcarrano: We really need a tool to find these.

pkg/lua/contrib/lua_run.c
jia200x: I tend to think this code should follow RIOT standards, because it's an adoption for RIOT.
Maybe using wrappers as seen [here](https://github.com/RIOT-OS/RIOT/blob/master/pkg/semtech-loramac/include/semtech_loramac.h#L162) can do the trick.

Note is always possible to directly invoke pkg functions with their own style, but this code will live in Doxygen and strictly attached to the RIOT side

pkg/lua/include/lua_builtin.h
jia200x: that's the idea of #8519. Hope it gets into master soon

jia200x:
@jcarrano I set the Waiting For Other PR label in order to focus on the blocking PRs (#9090 and #9244)

pkg/lua/contrib/lua_run.c
jcarrano: There is still the rest of the lua api, named after lua conventions. Then I would have mixed conventions in the same code (lua_riot_newstate returning a lua_State).

pkg/lua/Makefile.lua
```diff
@@ -1,9 +1,9 @@
...
-# This builds for native using POSIX system calls and some extra libraries, and
-# removes a compiler warning that warns against using tmpnam().
-ifeq ($(BOARD),native)
-  CFLAGS += -DLUA_USE_LINUX
-endif
+CFLAGS += -fstack-usage -fconserve-stack \
+          -DLUA_MAXCAPTURES=16 \
+          -DL_MAXLENNUM=50
+#    Enable these options to debug stack usage
+#          -Wstack-usage=128 -Wno-error=stack-usage=128
```
danpetry: Note that now the examples build with warnings...
@jcarrano @jia200x don't want to do a blocking comment but wanted to highlight this

jcarrano: If you are referring to this:
```
 loslib.c:(.text.os_tmpname+0x29): warning: the use of `tmpnam' is dangerous, better use `mkstemp'
```
There is no standard C way of fixing it, without changing the OS module's API. Asking for a filename and then creating the file is a bad idea, but that's how the lua standard library is made to work.

jia200x: @jcarrano is it possible to patch that file?

danpetry: Is ` CFLAGS += -DLUA_USE_LINUX` a bad idea

jcarrano: It is already patched. Here is what happens:
1. The `os` module contains a function `os.tempname` which should return a filename suitable for a temporary file.
2. tmpnam is a C function that was supposed to do this before peopl realized that it's not good to generate a filename without creating the file at the same time: it exposes you to a race condition, the file could be created by someone else in the meanwhile.
3. mkstemp is an alternative. However it is not C-standard and not really compatible with the way `os.tempname` works. What unpatched lua does (if LUA_USE_LINUX is defined) is to create a file and then close it (it does not even delete it).

Conclusion: it is a bug in the Lua API.

jcarrano: @danpetry Yes, exactly. It causes different code to be built for native, which defeats the purpose of being able to test your app in your PC, and have that app be as close as possible to what will run on the MCU later.

danpetry: You could create another #define LUA_RIOT_NATIVE or something which only deals with the offending lines in loslib.c and leaves the rest?
It's my understanding that "no compiler warnings" is a golden rule, maybe Jose can clarify having been a maintainer for longer

jia200x: As far as I see, this problem vanishes if -DLUA_USE_LINUX is not present. Right? As @jcarrano pointed out in this thread, it might be a good idea to run these without the LUA_USE_LINUX flag. That should solve the issue.

jcarrano: We could remove the warning, but the code would still be wrong. It is the very definition of [`os.tmpname`](https://www.lua.org/manual/5.3/manual.html#pdf-os.tmpname) that is wrong, and the user should instead use [`io.tmpfile`](https://www.lua.org/manual/5.3/manual.html#pdf-io.tmpfile) instead. Removing the function would be a better idea, we break the "official" lua api, but again, people should not be using that anyways.

pkg/lua/contrib/binsearch.c
jia200x: ENOENT stands for Error No ENTry (or Error No ENTity), as seen [here](https://stackoverflow.com/questions/19902828/why-does-enoent-mean-no-such-file-or-directory). It's usually described as "No such file or directory", but can be used for more cases

pkg/lua/Makefile.lua
jcarrano: @jia200x No, it's exactly the opposite. The warning vanishes with -DLUA_USE_LINUX. But then, not only is the code different: the function does different things too!. See https://www.lua.org/manual/5.3/manual.html#pdf-os.tmpname . With DLUA_USE_LINUX it creates a file, which you must remove afterwards!!!!!

jia200x: @jcarrano Then I would vote for removing/patching the function from RIOT. There are other packages that already do this (e.g [LoRaMAC](https://github.com/RIOT-OS/RIOT/blob/master/pkg/semtech-loramac/patches/0001-adapt-for-RIOT.patch)).

If there's a not-so-desired behavior, it's OK to patch it.

jcarrano: @jia200x @danpetry Done, os.tmpname is gone. I am accepting the fact that we will have to patch lua and there is no way around.

jia200x: yes, I think it's ok to do it. Although those modules are part of the Lua implementation, they are not part of the "Lua interpreter". There are high chances Lua for RIOT is not exactly the same as Lua for Linux. Use cases tend to be different IMO

danpetry: Cool, worth documenting somewhere? Worth starting maybe a wiki page on Lua for RIOT or something which can grow?

jia200x: I tend to prefer Doxygen pages (or similar), since the wiki is hard to maintain. See #8516

examples/lua_REPL/repl.lua
```diff
@@ -0,0 +1,91 @@
...
+local _R_CONT = 1
+local _R_EXIT = 2
+local _R_ERROR = 3
+
+--[[ Read code from standard input (whatever stdin means for lua)
+    @return action_code     what the eval loop should do
+    @return code_or_msg     either code (_R_EVAL) or a message (_R_ERROR) or nil
+                            (_R_CONT, _R_EXIT).
+]]
+
```
danpetry: Could you make a few changes to make this easier to understand out-of-the-box - i.e. so that someone can come in with no help and learn a few lua features? Suggestions are:
- rename "maybe_code" to something like "is_executable_code" or something and explain that the three `if maybe_code == nil` sections check whether it's an expression, then a statement, then a line from a multiline
- briefly explain the coroutine functionality
- explain what the typical dual return codes for lua are for
- maybe explain that load() compiles chunks line by line and then pcall() executes the compiled bytecode  

I understand that comments come at a memory cost since it's all being parsed into the array. Would be good to find a good balance here. I think if you can go as far as possible with readable code itself that would be good, by renaming some of the variables. Potentially, could also write some notes in the readme?

examples/lua_REPL/repl.lua
```diff
@@ -0,0 +1,91 @@
...
+        return _R_CONT
+    end
+
+    local maybe_code, compile_err = load("return "..ln)
+
+    if maybe_code == nil then
+        maybe_code, compile_err = load(ln)
+    end
+
+    if maybe_code == nil then
```
danpetry: `if not maybe_code then` ?

examples/lua_REPL/repl.lua
```diff
@@ -0,0 +1,91 @@
+--[[
```
danpetry: As you mentioned it is probably not necessary to be thorough with the [Lua style guide](http://lua-users.org/wiki/LuaStyleGuide). One thing I did notice, though, is that indents are generally two spaces. This would also reduce the array size a bit.

examples/lua_REPL/main.c
```diff
@@ -0,0 +1,50 @@
...
+
+#include <stdio.h>
+#include <string.h>
+
+#include "lua_run.h"
+#include "repl.lua.h"
+
+#define MAIN_LUA_MEM_SIZE (40000)
+
+char lua_memory[MAIN_LUA_MEM_SIZE] __attribute__ ((aligned (__BIGGEST_ALIGNMENT__)));
```
danpetry: also, space between `aligned` and `(__BIGGEST_ALIGNMENT__)))`

examples/lua_REPL/main.c
```diff
@@ -0,0 +1,50 @@
...
+
+#define BARE_MINIMUM_MODS (LUAR_LOAD_BASE|LUAR_LOAD_IO|LUAR_LOAD_CORO)
```
danpetry: space between these constants, i.e. `(LUAR_LOAD_BASE | LUAR_LOAD_IO | LUAR_LOAD_CORO)`

examples/lua_REPL/main.c
```diff
@@ -0,0 +1,50 @@
...
+
+int main(void)
+{
+    printf("%p - %p %zu\n", lua_memory, lua_memory+MAIN_LUA_MEM_SIZE, sizeof(void*));
```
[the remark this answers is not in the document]
danpetry: also `lua_memory + MAIN_LUA_MEM_SIZE`

examples/lua_REPL/main.c
```diff
@@ -0,0 +1,50 @@
...
+
+    while (1) {
+        int status, value;
+        puts("This is Lua: starting interactive session\n");
+
+        status = luaR_do_buffer(repl_lua, repl_lua_len, lua_memory, MAIN_LUA_MEM_SIZE,
+                                BARE_MINIMUM_MODS, &value);
+
+        printf("Exited. status: %s, return code %d\n", luaR_strerror(status),
+                                                        value);
```
danpetry: Does this need a newline? if so, should be aligned with the opening bracket

examples/lua_REPL/main.c
danpetry: Sorry, remove the space I meant (looked at the wrong diff from uncrustify)

examples/lua/Makefile
```diff
@@ -29,12 +29,10 @@ DEVELHELP ?= 1
...
 QUIET ?= 1
 
-USEMODULE += ps
-
 ifneq ($(BOARD),native)
   # This stack size is large enough to run Lua print() functions of
   # various lengths. Other functions untested.
-  CFLAGS += -DTHREAD_STACKSIZE_MAIN=4096
+  CFLAGS += -DTHREAD_STACKSIZE_MAIN=3000
 endif
```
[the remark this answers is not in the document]
jcarrano: It's pretty empirical, trial and error .I don't like that magic number either. I also wanted to have a nice power-of-two number. Right now there's no easy way to get this number, and it's architecture-dependent, which means testing in one CPU doesn't tell you anything about the others. In the future we can instrument the lua functions to record the maximum stack usage, and then the user can choose a decent safety margin.

A complementary solution is to just reduce usage so much that it is no longer a problem.

examples/lua/Makefile
```diff
@@ -29,12 +29,10 @@ DEVELHELP ?= 1
 # Change this to 0 show compiler invocation lines by default:
 QUIET ?= 1
```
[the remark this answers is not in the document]
jcarrano: OK. I'm thinking "lua-basic".

examples/lua_REPL/Makefile
```diff
@@ -0,0 +1,53 @@
...
+BOARD_WHITELIST := frdm-k64f iotlab-m3 mulle native nrf52840dk nucleo-f411re \
+				   stm32f4discovery
+
+# Comment this out to disable code in RIOT that does safety checking
+# which is not needed in a production environment but helps in the
+# development process:
+DEVELHELP ?= 1
+
+# Uncomment the following lines to enable debugging features.
+#CFLAGS_OPT = -Og
```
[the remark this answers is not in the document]
jcarrano: It tries to achieve a decent level of optimization, but it leaves out any optimization which could interfere with debugging. If you compile with -O2 and then try to debug, you get all kinds of weird jumps.

examples/lua_REPL/main.c
```diff
@@ -0,0 +1,50 @@
...
+ * @}
+ */
+
+#include <stdio.h>
+#include <string.h>
+
+#include "lua_run.h"
+#include "repl.lua.h"
+
+#define MAIN_LUA_MEM_SIZE (40000)
```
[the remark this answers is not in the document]
jcarrano: Lua has the [collectgarbage](https://www.lua.org/manual/5.3/manual.html#pdf-collectgarbage) function that reports the memory in use. Anyways, a more accurate estimate can be obtained by using the TLSF walker functions. Maybe it would be a good idea to include a binding to this module in a future release. (BTW, I'm writing down the list of TODO items)

examples/lua_REPL/main.c
```diff
@@ -0,0 +1,50 @@
...
+
+char lua_memory[MAIN_LUA_MEM_SIZE] __attribute__ ((aligned (__BIGGEST_ALIGNMENT__)));
+
+#define BARE_MINIMUM_MODS (LUAR_LOAD_BASE|LUAR_LOAD_IO|LUAR_LOAD_CORO)
+
+int main(void)
+{
+    printf("%p - %p %zu\n", lua_memory, lua_memory+MAIN_LUA_MEM_SIZE, sizeof(void*));
+
+    while (1) {
```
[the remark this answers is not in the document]
jcarrano: If the Lua interpreter crashes (for example, it runs out of memory) or exits (maybe an unhandled exception in Lua code), then it's immediately restarted. Without the `while (1)` the user would have to restart the CPU.

examples/lua_REPL/repl.lua
jcarrano: Another note for the future: write a "partial minifier" that preserves line numbers and variable names.

examples/lua_REPL/repl.lua
jcarrano: I agree that it's poorly documented. However, teaching Lua is beyond the scope of our documentation. I would expect people who want to use lua to learn the core language elsewhere, and in RIOT we can limit ourselves to document only the RIOT-specific extensions and usage.

examples/lua/Makefile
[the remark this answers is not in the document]
bergzand: Would it be possible to rewrite it to something like
`CFLAGS += '-DTHREAD_STACKSIZE_MAIN=(THREAD_STACKSIZE_DEFAULT + 2048)'`
To keep it somewhat architecture dependent?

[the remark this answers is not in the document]
jcarrano: @bergzand Good one, I didn't know about that define.

examples/lua_REPL/main.c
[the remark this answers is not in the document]
jcarrano: @bergzand Oops! that break shouldn't be there!!

examples/lua_REPL/repl.lua
jcarrano: Two-space indents make me want to rip off my eyes.

pkg/lua/Makefile
```diff
@@ -1,11 +1,11 @@
...
 PKG_LICENSE=MIT
 
 .PHONY: all
 
-all:
+all: Makefile.lua
```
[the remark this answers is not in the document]
jcarrano: It forces a rebuild if the lua makefile changes.

pkg/lua/contrib/lua_loadlib.c
```diff
@@ -0,0 +1,280 @@
...
+    if (load_result == LUA_OK) {
+        return 2;
+    } else {
+        return luaL_error(L, "Module '%s' not found in C-builtins",
+                          lua_tostring(L, 1));
+    }
+}
+
+int luaR_getloader(lua_State *L, const char *name)
```
danpetry: Please could you add Doxygen comments to all the functions that have been added to the Lua C API?

pkg/lua/patches/0006-Cleanup-test-module.patch
```diff
@@ -0,0 +1,497 @@
+From 542105cafbba381ef40c414a4025ea25d20371a2 Mon Sep 17 00:00:00 2001
+From: Juan Carrano <j.carrano@fu-berlin.de>
+Date: Mon, 14 May 2018 15:19:14 +0200
+Subject: [PATCH 6/8] Cleanup test module.
+
+This patch makes several improvements to the test module intended
+to allow testing inside embedded environments:
+
+- It does not rely in the standalone intepreter (lua.c).
```
danpetry: Could you describe how to test the changes (even better, provide test snippets)? If, for any of the changes, this is simply something like "the interpreter doesn't crash" or "it compiles" then that's ok too but please describe.

pkg/lua/patches/0007-Add-a-proper-makefile.patch
```diff
@@ -0,0 +1,358 @@
...
++*.a
++lua
++lua-dbg
++*.patch
+diff --git a/Makefile b/Makefile
+new file mode 100644
+index 00000000..2735db8d
+--- /dev/null
++++ b/Makefile
+@@ -0,0 +1,116 @@
```
danpetry: Might be worth checking with Gaetan whether these changes are in line with the work he's doing on the build system. In other words, whether they incorporate the work he's doing or do things in a different way

pkg/lua/contrib/lua_run.c
```diff
@@ -0,0 +1,275 @@
...
+    switch (compilation_result) {
+    case LUAR_MODULE_NOTFOUND:
+        status = LUAR_NOMODULE;
+        goto luaR_do_error;
+    case LUA_ERRSYNTAX:
+        status = LUAR_COMPILE_ERR;
+        goto luaR_do_error;
+    case LUA_ERRMEM: /* fallthrough */
+    case LUA_ERRGCMM:
+    default:
```
danpetry: please also uncrustify. here and throughout the rest of the submitted code.

pkg/lua/contrib/lua_run.c
danpetry: My opinion is that it would be best to follow the RIOT conventions as far as possible. If the API is RIOT programmer-facing, then it would avoid the case where e.g. there are a bunch of different packages with their own coding style and the programmer sees a range of different styles. I think leaving in stuff which **has** to be left in is acceptable (and obv necessary) but better than having a real mix of styles

examples/lua_REPL/README.md
danpetry: @jcarrano pointed out that it's not our place to teach people to program in Lua (we wouldn't link to a JavaScript tutorial in the jerryscript pkg for example). Worth noting that at the moment we haven't done any bindings. The only officially supported function is print() as far as I'm aware

examples/lua/Makefile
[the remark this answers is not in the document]
danpetry: sounds good

examples/lua_REPL/Makefile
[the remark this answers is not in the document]
danpetry: thx!

examples/lua_REPL/main.c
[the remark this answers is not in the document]
danpetry: > I'm writing down the list of TODO items
Great, thanks :+1:

examples/lua_REPL/repl.lua
danpetry: ok cool, thanks for making the changes

examples/lua_REPL/repl.lua
danpetry: ok. That's a fair reaction tbh

pkg/lua/Makefile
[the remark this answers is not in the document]
danpetry: ok thanks!

pkg/lua/contrib/lua_run.c
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    if (buf == NULL) {
+        compilation_result = luaR_getloader(L, modname);
+    } else {
+        compilation_result = luaL_loadbufferx(L, buf, buflen, modname, "t");
+    }
+
+    switch (compilation_result) {
+    case LUAR_MODULE_NOTFOUND:
```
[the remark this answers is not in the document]
jcarrano: It would be more like moving the dirtiness to somewhere else. I don't like to rely on specific values of enums, and I'd rather treat them as symbols.

pkg/lua/contrib/lua_run.c
```diff
@@ -0,0 +1,275 @@
...
+    /* TODO: try to return some info about the error object. */
+
+    longjmp(*jump_buffer, 1);
+}
+
+static int luaR_do_module_or_buf(const char *buf, size_t buflen,
+                            const char *modname, void *memory, size_t mem_size,
+                            uint16_t modmask, int *retval)
+{
+    jmp_buf jump_buffer;
```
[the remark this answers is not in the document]
jcarrano: What do you mean by cleanup?

pkg/lua/patches/0007-Add-a-proper-makefile.patch
jcarrano: This makefile is not being used (it is overwritten). The patch is there because it's useful if one wants to compile the standalone interpreter.

pkg/lua/contrib/lua_run.c
jcarrano: Ok. One thing I'll keep, though, is the indentation of the code between `#ifdef`s. I find it quite hard to follow conditionals if they are not indented, and I don't see how a compile-time conditional would make a difference.

pkg/lua/contrib/lua_loadlib.c
```diff
@@ -0,0 +1,280 @@
...
+
+    if (load_result == LUAR_MODULE_NOTFOUND) {
+        load_result = _ll_searcher_builtin_c(L, name);
+    }
+
+    return load_result;
+}
+
+/* ======================== 'require' function ============================= */
+
```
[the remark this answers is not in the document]
jcarrano: Do we have a specific rule for function placement?

pkg/lua/contrib/lua_loadlib.c
jcarrano: See the header file.

pkg/lua/patches/0001-Remove-luaL_newstate.patch
```diff
@@ -0,0 +1,62 @@
...
+---
+ lauxlib.c | 25 -------------------------
+ lauxlib.h |  2 --
+ 2 files changed, 27 deletions(-)
+
+diff --git a/lauxlib.c b/lauxlib.c
+index 7b14ca4d..018a4347 100644
+--- a/lauxlib.c
++++ b/lauxlib.c
+@@ -1005,31 +1005,6 @@ LUALIB_API const char *luaL_gsub (lua_State *L, const char *s, const char *p,
```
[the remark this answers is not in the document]
jcarrano: Done (for some reason this comment is placed in the wrong file).

pkg/lua/contrib/lua_loadlib.c
```diff
@@ -0,0 +1,280 @@
...
+    { "preload", NULL },
+    { "searchers", NULL },
+    { "loaded", NULL },
+    { NULL, NULL }
+};
+
+
+static const luaL_Reg ll_funcs[] = {
+    { "require", ll_require },
+    { NULL, NULL }
```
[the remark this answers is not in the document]
jcarrano: Done (in doc.txt). Of course, it's a coarse readme, we need proper docs, with sections and multiple pages and code snippets and all.

pkg/lua/patches/0006-Cleanup-test-module.patch
jcarrano: Maybe I should add a link to the repo in which I have the modified Lua? I can be downloaded in a desktop and compiled just like normal lua (only better, because mine builds the debug and non-debug versions simultaneously), and can run the test suite just like normal lua.

tcschmidt:
@danpetry are comments of @jia200x addressed? Can we remove the blocking?

pkg/lua/contrib/lua_run.c
bergzand: I strongly agree with @jia200x and @danpetry here to follow RIOT coding conventions. mixing conventions is something that is often caused by external software. I would not wrap everything, but having at least our own contributed functions in our own convention also makes it easier for seasoned RIOT users as they know what to expect and prevents confusion with new users

Rewording our own functions to something like `lua_riot_funcname` is IMHO clear enough to indicate that it is our version of `funcname`.

(I think I accidentally rephrased @danpetry's arguments)

danpetry:
@tcschmidt I'm waiting for some test information and results from @jcarrano . After these are here I'll review the details at the same time, i.e. do it all together - hope that's ok, I prefer to do reviewing monolithically rather than piecemeal as it's more efficient.

danpetry:
Strongly agree with the splitting PRs. This will make it a lot easier to get future contributions merged

examples/lua_REPL/Makefile
```diff
@@ -0,0 +1,53 @@
...
+#CFLAGS += -DDEBUG_ASSERT_VERBOSE -DLUA_DEBUG
+
+# Change this to 0 show compiler invocation lines by default:
+QUIET ?= 1
+
+# This is a lot of stack, should be fixed in the future.
```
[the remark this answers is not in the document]
jcarrano: See the package's `doc.txt`

examples/lua_REPL/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+$(LUA_PATH)/:
+	@mkdir -p $@
+
+$(LUA_H): | $(LUA_PATH)/
+$(LUA_H): $(LUA_PATH)/%.lua.h: %.lua
+
+	xxd -i $< | sed 's/^unsigned/const/g' > $@
+
+$(RIOTBUILD_CONFIG_HEADER_C): $(LUA_H)
```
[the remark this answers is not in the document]
jcarrano: Yes, but in a future PR. Duplication is only one of the issues. That way of generating code is not very robust, but this change I CAN split, so I will.

examples/lua_REPL/Makefile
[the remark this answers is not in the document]
jcarrano: I want to keep it separate. Running the full test suite requires `-DLUA_DEBUG` but that doesn't mean I want to compile with `-Og` because it may take too much flash.

examples/lua_REPL/Makefile
```diff
@@ -0,0 +1,53 @@
+APPLICATION = lua_repl
+
+# If no BOARD is found in the environment, use this default:
+BOARD ?= native
+
+# This has to be the absolute path to the RIOT base directory:
+RIOTBASE ?= $(CURDIR)/../..
+
+BOARD_WHITELIST := frdm-k64f iotlab-m3 mulle native nrf52840dk nucleo-f411re \
+				   stm32f4discovery
```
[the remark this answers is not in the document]
jcarrano: Those are boards with enough flash to run Lua. Maybe other boards can run Lua too, but I don't know.

pkg/lua/include/lua_builtin.h
```diff
@@ -0,0 +1,65 @@
...
+/** Table containing all built in pure lua modules */
+extern const struct luaR_builtin_lua *const luaR_builtin_lua_table;
+/** Number of elements of luaR_builtin_lua_table */
+extern const size_t luaR_builtin_lua_table_len;
+
+/** Table containing all built in c lua modules */
+extern const struct luaR_builtin_c *const luaR_builtin_c_table;
+/** Number of elements of luaR_builtin_c_table */
+extern const size_t luaR_builtin_c_table_len;
```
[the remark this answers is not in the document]
jcarrano: Those are defined as weak symbols,where the default value is an empty table. This allows the code to compile and work even of there are no lua modules. The tables can be overridden by the application to provide additional functionality in the form of modules.
So, they are relevant for the application to define, but not to read.

pkg/lua/include/lua_loadlib.h
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+#ifndef LUA_LOADLIB_H
+#define LUA_LOADLIB_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define LUAR_MODULE_NOTFOUND 50
```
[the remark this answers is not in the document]
jcarrano: To ensure we don't collide with lua's error codes.

jcarrano:
@danpetry @bergzand I understand that the PR is big, but how do you propose to split it? The only split that could work is the package and the examples.

pkg/lua/include/lua_builtin.h
```diff
@@ -0,0 +1,65 @@
...
+ * Entry describing a c lua module built into the
+ * application binary.
+ */
+struct luaR_builtin_c {
+    const char *name;   /*!< Name of the module */
+    int (*luaopen)(lua_State *); /*!< Loader function. It must place the module
+                                  *  table at the top of the lua stack.
+                                  *  @todo Add better docs.
+                                  */
+};
```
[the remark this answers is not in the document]
jcarrano: Those structs make up tables that have to be redefined. I don't do the user any favour by hiding the fact that they are structs.

danpetry:
Let's keep it as it is for now, we're almost there. Let's talk about the granularity that works best for upcoming PRs.

examples/lua_REPL/Makefile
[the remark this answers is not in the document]
bergzand: It looks like my comment was not completely clear here. What I meant is that to me, the `-Og` option is not only useful here, but could be useful for debugging other examples as well. So what I would like to propose is an option along the lines of the current `DEVELHELP`, maybe called `OPTIMIZE_DEBUG` to enable `-Og` for a random example.

examples/lua_REPL/Makefile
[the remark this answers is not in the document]
bergzand: Could you at least add this information as a comment here. If you like, maybe you could quickly find all boards that have enough flash and ram to fit the example here.

As this is now, I don't think it is clear for an end user why these specific boards are whitelisted and when a board qualifies

pkg/lua/include/lua_builtin.h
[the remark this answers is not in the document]
bergzand: I don't think you have to hide the fact that they are structs, I'm just asking you to adhere to the [RIOT coding conventions on types](https://github.com/RIOT-OS/RIOT/wiki/Coding-conventions#types)

pkg/lua/include/lua_builtin.h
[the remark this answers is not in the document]
bergzand: Okay, would you mind adding a short explanation about this above these declarations?

bergzand:
> @danpetry @bergzand I understand that the PR is big, but how do you propose to split it? The only split that could work is the package and the examples.

While I agree with @danpetry that it is a bit late for a split, I think having the examples split from this PR already helps. I'm not sure if it is possible, but having (some of) the patches in a separate PR makes also for easier reviewing.

danpetry:
I'd really prefer if we didn't split this now. I've got a good handle on everything that's going on here now, and I'm almost there with reviewing. It wouldn't make sense at this stage.

bergzand:
> I'd really prefer if we didn't split this now. I've got a good handle on everything that's going on here now, and I'm almost there with reviewing. It wouldn't make sense at this stage.

Yes, I completely agree with you here, just wanted to give @jcarrano a few pointers on how to split PRs like these in the future

danpetry:
Ah ok cool :)

examples/lua_REPL/Makefile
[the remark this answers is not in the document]
jcarrano: Now I'm thinking on whitelisting only the boards that I can try myself. I could run the CI on all boards and then blacklist all that fail because of low memory, but that's not very robust: a later change to riot may cause the build to fail. How should I proceed?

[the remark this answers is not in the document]
tcschmidt: Check back with @cladmi .

pkg/lua/include/lua_builtin.h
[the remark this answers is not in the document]
jcarrano: > * The use of typedefs for structs and pointers is allowed.

"allowed" != "required"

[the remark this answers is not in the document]
bergzand: Ah you're right there. I still don't see the harm in adding a typedef here, users are not forced to use it, but if you really have a strong opinion here I won't block it.

pkg/lua/Makefile.lua
jcarrano: > * remove os.tmpname using the same approach as used with os.rename and os.clock (see first patch)
> * record that these three functions aren't supported in doc.txt

Done

examples/lua_REPL/main.c
danpetry: This is corrected

pkg/lua/include/lua_builtin.h
danpetry: This is corrected

examples/lua_REPL/README.md
danpetry: Link to the lua page and some brief info "works in the same way as the lua shell in the default installation"

pkg/lua/Makefile.lua
danpetry: Great, thanks.

examples/lua/Makefile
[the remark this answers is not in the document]
danpetry: has been changed

pkg/lua/Makefile
danpetry: This has gone back to a commit hash, because tags can change.

sys/newlib_syscalls_default/syscalls.c
danpetry: This is rebased

pkg/lua/contrib/lua_run.c
danpetry: This has been changed

examples/lua_REPL/main.c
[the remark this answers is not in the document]
danpetry: happy with current implementation. Will be nice when/if we get feedback on the examples, can raise PRs if it's not useful for people.

pkg/lua/contrib/lua_loadlib.c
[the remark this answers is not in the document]
danpetry: Yes. the doc.txt is great for now, thanks.

pkg/lua/contrib/lua_loadlib.c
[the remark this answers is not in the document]
danpetry: Not really, fine with your implementation.

pkg/lua/patches/0007-Add-a-proper-makefile.patch
danpetry: ok

pkg/lua/contrib/binsearch.c
```diff
@@ -0,0 +1,52 @@
...
+ * @brief   Generic binary search for tables containing strings.
+ * @author  Juan Carrano <j.carrano@fu-berlin.de>
+ *
+ */
+
+#include <string.h>
+
+#include "binsearch.h"
+
+int binsearch_str(const void *start, size_t offset, size_t stride, size_t nmemb,
```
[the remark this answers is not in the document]
jcarrano: It can be done with bsearch, sure but I'll be the same amount of code as we have now, getting an index instead of a pointer would be more hacky and we would still need the macros.

danpetry:
Thanks for making changes. However, lua_REPL example now doesn't compile, please could you debug?

jcarrano:
@danpetry Fixed. It seems that the "ä" from Universität was causing trouble, and I had to change function signatures to accept `uint8` instead of of `char` to avoid future problems.

examples/lua_REPL/README.md
```diff
@@ -0,0 +1,32 @@
+## Lua interactive interpreter
+
+### About
+
+This example shows how to run a [Lua](https://www.lua.org/) Read-Eval-Print loop.
+It works in a similar way to the lua shell found in the default lua installation
+in your operating system.
```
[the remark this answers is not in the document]
bergzand: @jcarrano poke :)

danpetry:
could you please squash the changes into one of the previous two commits and push them?

danpetry:
All comments addressed. ACK

bergzand:
@danpetry I could really appreciate it if you let me just ACK the PR next time instead of dismissing my review. I still had one or two open discussions... :(

PR #9156, opened 2018-05-19
net/gcoap: Use nanocoap pkt/options API

kb2ma:
### Contribution description
Updates gcoap to use the nanocoap pkt/options API in #9085. This work allows removal of the gcoap-specific attributes in coap_pkt_t, which add a painful 130 bytes. (Now you know why I've been working so hard lately!) 

Use of this API allows gcoap to fully build on nanocoap as a higher-level API. gcoap has always used a coap_pkt_t to build a message, and now nanocoap does too. See how gcoap_req_init() prepares for coap_build_hdr(), and then calls coap_pkt_init(), coap_opt_add_string() for the path, and coap_opt_add_uint() for the content format.

#### 2018-10-23 Update
Simplified implementation of gcoap_finish(), which results in a minor API change. gcoap_finish() adds a Content-Format option to the packet after writing the payload. The original implementation of this PR accommodated options with any other option number that may have been added to the packet. Today I updated the implementation to accept  options only with an option number less than Content-Format.

Fundamentally, the gcoap_finish() function has become obsolete as we have increased the use of CoAP options. The updated CoAP API now uses coap_opt_finish() to finalize the packet header *before* writing the payload. We plan to deprecate gcoap_finish() soon.

This change significantly simplifies gcoap_finish() until we deprecate it, and reduces the text space in the executable for this function by half, around 120 bytes. The cost is that existing code breaks when setting a Uri-Query option and then setting the format in gcoap_finish(). (Uri-Query is the only option with an option number greater than Content-Format that gcoap used before this PR.) In fact, the new Resource Directory implementation, cord_ep, had this problem, so I added a commit to this PR to fix it.

### Issues/PRs references
Depends on #10223

See [CoAP API Options Update](https://github.com/RIOT-OS/RIOT/wiki/CoAP-API-Options-Update) wiki page for background, including how gcoap uses it.

### DIFF
--- a/sys/include/net/gcoap.h
+++ b/sys/include/net/gcoap.h
@@ -253,30 +253,6 @@ extern "C" {
 #endif
 
 /**
- * @brief   Reduce payload length by this value for a request created with
- *          gcoap_req_init()
- *
- * Accommodates writing Content-Format option in gcoap_finish()
- */
-#define GCOAP_REQ_OPTIONS_BUF   (4)
-
-/**
- * @brief   Reduce payload length by this value for a respons created with
- *          gcoap_resp_init()
- *
- * Accommodates writing Content-Format option in gcoap_finish()
- */
-#define GCOAP_RESP_OPTIONS_BUF  (4)
-
-/**
- * @brief   Reduce payload length by this value for an observe notification
- *          created with gcoap_obs_init()
- *
- * Accommodates writing Content-Format option in gcoap_finish()
- */
-#define GCOAP_OBS_OPTIONS_BUF   (4)
-
-/**
  * @brief   Maximum number of requests awaiting a response
  */
 #ifndef GCOAP_REQ_WAITING_MAX
--- a/sys/include/net/gcoap.h
+++ b/sys/include/net/gcoap.h
@@ -253,26 +253,28 @@ extern "C" {
 #endif
 
 /**
- * @brief   Size of the buffer used to write options, other than Uri-Path, in a
- *          request
+ * @brief   Reduce payload length by this value for a request created with
+ *          gcoap_req_init()
  *
- * Accommodates Content-Format and Uri-Queries
+ * Accommodates writing Content-Format option in gcoap_finish()
  */
-#define GCOAP_REQ_OPTIONS_BUF   (40)
+#define GCOAP_REQ_OPTIONS_BUF   (4)
 
 /**
- * @brief   Size of the buffer used to write options in a response
+ * @brief   Reduce payload length by this value for a respons created with
+ *          gcoap_resp_init()
  *
- * Accommodates Content-Format.
+ * Accommodates writing Content-Format option in gcoap_finish()
  */
-#define GCOAP_RESP_OPTIONS_BUF  (8)
+#define GCOAP_RESP_OPTIONS_BUF  (4)
 
 /**
- * @brief   Size of the buffer used to write options in an Observe notification
+ * @brief   Reduce payload length by this value for an observe notification
+ *          created with gcoap_obs_init()
  *
- * Accommodates Content-Format and Observe.
+ * Accommodates writing Content-Format option in gcoap_finish()
  */
-#define GCOAP_OBS_OPTIONS_BUF   (8)
+#define GCOAP_OBS_OPTIONS_BUF   (4)
 
 /**
  * @brief   Maximum number of requests awaiting a response
@@ -506,6 +508,11 @@ void gcoap_register_listener(gcoap_listener_t *listener);
 
 /**
  * @brief   Initializes a CoAP request PDU on a buffer.
+
+ * @warning After you use this function, you may not add Options with option
+ * number less than COAP_OPT_URI_PATH. Otherwise, use the struct-based API
+ * described with @link net_nanocoap nanocoap @endlink to initialize the
+ * message. See the implementation of gcoap_req_init() itself as an example.
  *
  * @param[out] pdu      Request metadata
  * @param[out] buf      Buffer containing the PDU
@@ -525,8 +532,14 @@ int gcoap_req_init(coap_pkt_t *pdu, uint8_t *buf, size_t len,
 /**
  * @brief   Finishes formatting a CoAP PDU after the payload has been written
  *
- * Assumes the PDU has been initialized with gcoap_req_init() or
- * gcoap_resp_init().
+ * Assumes the PDU has been initialized with a gcoap_xxx_init() function, like
+ * gcoap_req_init().
+ *
+ * @warning To use this function, you only may have added an Option with
+ * option number less than COAP_OPT_CONTENT_FORMAT. Otherwise, use the
+ * struct-based API described with @link net_nanocoap nanocoap. @endlink With
+ * this API, you specify the format with coap_opt_add_uint(), prepare for the
+ * payload with coap_opt_finish(), and then write the payload.
  *
  * @param[in,out] pdu       Request metadata
  * @param[in] payload_len   Length of the payload, or 0 if none
--- a/sys/include/net/nanocoap.h
+++ b/sys/include/net/nanocoap.h
@@ -232,9 +232,6 @@ typedef struct {
     uint16_t options_len;                       /**< length of options array */
     coap_optpos_t options[NANOCOAP_NOPTS_MAX];  /**< option offset array     */
 #ifdef MODULE_GCOAP
-    uint8_t url[NANOCOAP_URI_MAX];              /**< parsed request URL      */
-    uint8_t qs[NANOCOAP_QS_MAX];                /**< parsed query string     */
-    uint16_t content_type;                      /**< content type            */
     uint32_t observe_value;                     /**< observe value           */
 #endif
 } coap_pkt_t;
--- a/sys/include/net/nanocoap.h
+++ b/sys/include/net/nanocoap.h
@@ -175,7 +175,7 @@ extern "C" {
 /**
  * @brief   Nanocoap-specific value to indicate no format specified
  */
-#define COAP_FORMAT_NONE        (65535)
+#define COAP_FORMAT_NONE        (UINT16_MAX)
 
 /**
  * @name    Nanocoap specific maximum values
--- a/sys/net/application_layer/gcoap/gcoap.c
+++ b/sys/net/application_layer/gcoap/gcoap.c
@@ -38,6 +38,15 @@
 #define GCOAP_RESOURCE_WRONG_METHOD -1
 #define GCOAP_RESOURCE_NO_PATH -2
 
+/*
+ * Reduce payload length by this value for a request created with
+ * gcoap_req_init(), gcoap_resp_init(), and gcoap_obs_init(), respectively.
+ * Accommodates writing Content-Format option in gcoap_finish().
+ */
+#define GCOAP_REQ_OPTIONS_BUF   (4)
+#define GCOAP_RESP_OPTIONS_BUF  (4)
+#define GCOAP_OBS_OPTIONS_BUF   (4)
+
 /* Internal functions */
 static void *_event_loop(void *arg);
 static void _listen(sock_udp_t *sock);
--- a/sys/net/application_layer/gcoap/gcoap.c
+++ b/sys/net/application_layer/gcoap/gcoap.c
@@ -38,20 +38,12 @@
 #define GCOAP_RESOURCE_WRONG_METHOD -1
 #define GCOAP_RESOURCE_NO_PATH -2
 
-/*
- * gcoap internal Content-Format option value. Not intended for use in a
- * transmitted packet. Must be a 3-byte unsigned value.
- */
-#define COAP_FORMAT_NO_PAYLOAD  (UINT16_MAX + 1)
-
 /* Internal functions */
 static void *_event_loop(void *arg);
 static void _listen(sock_udp_t *sock);
 static ssize_t _well_known_core_handler(coap_pkt_t* pdu, uint8_t *buf, size_t len, void *ctx);
-static ssize_t _write_options(coap_pkt_t *pdu, uint8_t *buf, size_t len);
 static size_t _handle_req(coap_pkt_t *pdu, uint8_t *buf, size_t len,
                                                          sock_udp_ep_t *remote);
-static ssize_t _finish_pdu(coap_pkt_t *pdu, uint8_t *buf, size_t len);
 static void _expire_request(gcoap_request_memo_t *memo);
 static void _find_req_memo(gcoap_request_memo_t **memo_ptr, coap_pkt_t *pdu,
                            const sock_udp_ep_t *remote);
@@ -343,9 +335,6 @@ static size_t _handle_req(coap_pkt_t *pdu, uint8_t *buf, size_t len,
                 memcpy(&memo->token[0], pdu->token, memo->token_len);
             }
             DEBUG("gcoap: Registered observer for: %s\n", memo->resource->path);
-            /* generate initial notification value */
-            uint32_t now       = xtimer_now_usec();
-            pdu->observe_value = (now >> GCOAP_OBS_TICK_EXPONENT) & 0xFFFFFF;
         }
 
     } else if (coap_get_observe(pdu) == COAP_OBS_DEREGISTER) {
@@ -430,29 +419,6 @@ static int _find_resource(coap_pkt_t *pdu, const coap_resource_t **resource_ptr,
 }
 
 /*
- * Finishes handling a PDU -- write options and reposition payload.
- *
- * Returns the size of the PDU within the buffer, or < 0 on error.
- */
-static ssize_t _finish_pdu(coap_pkt_t *pdu, uint8_t *buf, size_t len)
-{
-    ssize_t hdr_len = _write_options(pdu, buf, len);
-    DEBUG("gcoap: header length: %i\n", (int)hdr_len);
-
-    if (hdr_len > 0) {
-        /* move payload over unused space after options */
-        if (pdu->payload_len) {
-            memmove(buf + hdr_len, pdu->payload, pdu->payload_len);
-        }
-
-        return hdr_len + pdu->payload_len;
-    }
-    else {
-        return -1;      /* generic failure code */
-    }
-}
-
-/*
  * Finds the memo for an outstanding request within the _coap_state.open_reqs
  * array. Matches on remote endpoint and token.
  *
@@ -533,69 +499,6 @@ static ssize_t _well_known_core_handler(coap_pkt_t* pdu, uint8_t *buf, size_t le
                                       COAP_FORMAT_LINK);
     /* response content */
     return gcoap_finish(pdu, (size_t)plen, COAP_FORMAT_LINK);
-}
-
-/*
- * Creates CoAP options and sets payload marker, if any.
- *
- * Returns length of header + options, or -EINVAL on illegal path.
- */
-static ssize_t _write_options(coap_pkt_t *pdu, uint8_t *buf, size_t len)
-{
-    uint8_t last_optnum = 0;
-    (void)len;
-
-    uint8_t *bufpos = buf + coap_get_total_hdr_len(pdu);  /* position for write */
-
-    /* Observe for notification or registration response */
-    if (coap_get_code_class(pdu) == COAP_CLASS_SUCCESS && coap_has_observe(pdu)) {
-        uint32_t nval  = htonl(pdu->observe_value);
-        uint8_t *nbyte = (uint8_t *)&nval;
-        unsigned i;
-        /* find address of non-zero MSB; max 3 bytes */
-        for (i = 1; i < 4; i++) {
-            if (*(nbyte+i) > 0) {
-                break;
-            }
-        }
-        bufpos += coap_put_option(bufpos, last_optnum, COAP_OPT_OBSERVE,
-                                                       nbyte+i, 4-i);
-        last_optnum = COAP_OPT_OBSERVE;
-    }
-
-    /* Uri-Path for request */
-    if (coap_get_code_class(pdu) == COAP_CLASS_REQ) {
-        size_t url_len = strlen((char *)pdu->url);
-        if (url_len) {
-            if (pdu->url[0] != '/') {
-                DEBUG("gcoap: _write_options: path does not start with '/'\n");
-                return -EINVAL;
-            }
-            bufpos += coap_opt_put_uri_path(bufpos, last_optnum,
-                                            (char *)pdu->url);
-            last_optnum = COAP_OPT_URI_PATH;
-        }
-    }
-
-    /* Content-Format */
-    if (pdu->content_type != COAP_FORMAT_NONE) {
-        bufpos += coap_put_option_ct(bufpos, last_optnum, pdu->content_type);
-        last_optnum = COAP_OPT_CONTENT_FORMAT;
-    }
-
-    /* Uri-query for requests */
-    if (coap_get_code_class(pdu) == COAP_CLASS_REQ) {
-        bufpos += coap_opt_put_uri_query(bufpos, last_optnum,
-                                         (char *)pdu->qs);
-        /* uncomment when further options are added below ... */
-        /* last_optnum = COAP_OPT_URI_QUERY; */
-    }
-
-    /* write payload marker */
-    if (pdu->payload_len) {
-        *bufpos++ = GCOAP_PAYLOAD_MARKER;
-    }
-    return bufpos - buf;
 }
 
 /*
@@ -729,11 +632,7 @@ int gcoap_req_init(coap_pkt_t *pdu, uint8_t *buf, size_t len,
 {
     assert((path != NULL) && (path[0] == '/'));
 
-    (void)len;
-
     pdu->hdr = (coap_hdr_t *)buf;
-    memset(pdu->url, 0, NANOCOAP_URL_MAX);
-    memset(pdu->qs, 0, NANOCOAP_QS_MAX);
 
     /* generate token */
 #if GCOAP_TOKENLEN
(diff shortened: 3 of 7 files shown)


kb2ma:
I think code complete, but I'll give it a little time before removing WIP. I was able to remove most of the gcoap attributes in coap_pkt_t, with the nice reduction in size. I partially converted use of the observe value attribute, but removal will require a little work in nanocoap.  I'll leave that for a follow-on PR.

kb2ma:
Rebased and ready for review now that #9085 has been merged.

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -900,12 +862,22 @@ int gcoap_resp_init(coap_pkt_t *pdu, uint8_t *buf, size_t len, unsigned code)
     }
     coap_hdr_set_code(pdu->hdr, code);
 
-    /* Reserve some space between the header and payload to write options later */
-    pdu->payload      = buf + coap_get_total_hdr_len(pdu) + GCOAP_RESP_OPTIONS_BUF;
-    /* Payload length really zero at this point, but we set this to the available
-     * length in the buffer. Allows us to reconstruct buffer length later. */
-    pdu->payload_len  = len - (pdu->payload - buf);
-    pdu->content_type = COAP_FORMAT_NONE;
+    unsigned header_len  = sizeof(coap_hdr_t) + coap_get_token_len(pdu);
```
bergzand: This looks like a `coap_get_total_hdr_len(pdu);`

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -900,12 +862,22 @@ int gcoap_resp_init(coap_pkt_t *pdu, uint8_t *buf, size_t len, unsigned code)
...
+    bool has_obs_register = (coap_get_observe(pdu) == COAP_OBS_REGISTER);
+
+    pdu->options_len = 0;
+    pdu->payload     = buf + header_len;
+    pdu->payload_len = len - (pdu->payload - buf);
```
bergzand: `pdu->payload - buf` is equal to `header_len` right?

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -750,15 +661,10 @@ int gcoap_req_init(coap_pkt_t *pdu, uint8_t *buf, size_t len,
...
-        /* Payload length really zero at this point, but we set this to the available
-         * length in the buffer. Allows us to reconstruct buffer length later. */
-        pdu->payload_len  = len - (pdu->payload - buf);
-        pdu->content_type = COAP_FORMAT_NONE;
-
-        memcpy(&pdu->url[0], path, strlen(path));
+        coap_pkt_init(pdu, buf, len, hdrlen);
+        coap_opt_add_string(pdu, COAP_OPT_URI_PATH, path, '/');
+        /* placeholder until gcoap_finish() */
+        coap_opt_add_uint(pdu, COAP_OPT_CONTENT_FORMAT, COAP_FORMAT_NO_PAYLOAD);
```
bergzand: Is this to guarantee space for the content format in the buffer?

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -769,12 +675,68 @@ int gcoap_req_init(coap_pkt_t *pdu, uint8_t *buf, size_t len,
 
 ssize_t gcoap_finish(coap_pkt_t *pdu, size_t payload_len, unsigned format)
 {
-    /* reconstruct full PDU buffer length */
-    size_t len = pdu->payload_len + (pdu->payload - (uint8_t *)pdu->hdr);
+    unsigned format_offset = UINT_MAX;
```
bergzand: The use of this variable confuses me. It looks to me as if it is only used to remember if there was a content-format option

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -769,12 +675,68 @@ int gcoap_req_init(coap_pkt_t *pdu, uint8_t *buf, size_t len,
...
+                len_delta -= coap_put_option_ct(
+                                (uint8_t *)pdu->hdr + pdu->options[i].offset,
+                                (i > 0) ? pdu->options[i-1].opt_num : 0, format);
+            }
+        }
+        else if (format_offset < UINT_MAX) {
+            if (format == COAP_FORMAT_NONE) {
+                if (!move_pos) {
+                    move_pos = (uint8_t *)pdu->hdr + pdu->options[i].offset;
+                }
```
bergzand: These snippet  of 3 lines is in both branches of the `if`. Is it possible to move it outside of the `if` statement

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -769,12 +675,68 @@ int gcoap_req_init(coap_pkt_t *pdu, uint8_t *buf, size_t len,
...
 ssize_t gcoap_finish(coap_pkt_t *pdu, size_t payload_len, unsigned format)
 {
-    /* reconstruct full PDU buffer length */
-    size_t len = pdu->payload_len + (pdu->payload - (uint8_t *)pdu->hdr);
+    unsigned format_offset = UINT_MAX;
+    unsigned len_delta = 0;     /* buffer length to move forward */
+    uint8_t *move_pos = 0;      /* buffer position to move forward */
+
+    /* Remove or adjust Content-Format option for request, if any.
+     * Assumes current payload format is COAP_FORMAT_NO_PAYLOAD */
```
bergzand: I had a hard time understanding the code below. Maybe some additional comments would help the next person. Something like:

```
/* Remove or adjust Content-Format option for request, if any.
 * Assumes current payload format is COAP_FORMAT_NO_PAYLOAD
 * First the content format is updated and the option size difference is
 * stored in len_delta. Then the Option array is updated. Finally the options as 
 * placed in the buffer are moved to their updated position (offset) */
```

Feel free to adjust to your liking

sys/include/net/nanocoap.h
```diff
@@ -183,8 +178,19 @@ extern "C" {
...
+ * Not intended for use in a transmitted packet.
+ */
+#define COAP_FORMAT_NONE        (65535)
+/**
+ * @brief    nanocoap-internal value to indicate no payload present.
+ *
+ * Not intended for use in a transmitted packet. Must be a 3-byte unsigned
+ * value.
+ */
+#define COAP_FORMAT_NO_PAYLOAD  (65536)
```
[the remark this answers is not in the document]
kb2ma: Sure, no problem.

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -375,14 +370,20 @@ static int _find_resource(coap_pkt_t *pdu, coap_resource_t **resource_ptr,
...
+
     while (listener) {
         coap_resource_t *resource = listener->resources;
         for (size_t i = 0; i < listener->resources_len; i++) {
             if (i) {
                 resource++;
             }
 
-            int res = strcmp((char *)&pdu->url[0], resource->path);
+            int res = strcmp((char *)&uri[0], resource->path);
```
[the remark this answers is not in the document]
kb2ma: coap_get_uri() inserts the terminator, but I agree it's better to be safe.

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -1017,26 +988,25 @@ int gcoap_get_resource_list(void *buf, size_t maxlen, uint8_t cf)
 
 int gcoap_add_qstring(coap_pkt_t *pdu, const char *key, const char *val)
 {
-    size_t qs_len = strlen((char *)pdu->qs);
+    char qs[NANOCOAP_URI_MAX];
```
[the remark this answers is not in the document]
kb2ma: NANOCOAP_QS_MAX definition goes away in nanocoap.h.

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -1017,26 +988,25 @@ int gcoap_get_resource_list(void *buf, size_t maxlen, uint8_t cf)
...
+    size_t qs_len  = 0;
     size_t key_len = strlen(key);
     size_t val_len = (val) ? (strlen(val) + 1) : 0;
 
-    /* make sure if url_len + the new query string fit into the url buffer */
-    if ((qs_len + key_len + val_len + 2) >= NANOCOAP_QS_MAX) {
+    if ((key_len + val_len + 2) >= NANOCOAP_URI_MAX) {
```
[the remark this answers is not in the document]
kb2ma: Actually it's the query punctuation, but yes, let's describe the magic.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: I agree. It looks like I don't even need the header_len variable itself. I'll take a closer look.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: Yes, I'll make that clearer.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: Yes, that's right. I will generally add more comments to make the logic in gcoap_finish() clearer. This tortured nature of this code is why I plan to deprecate gcoap_finish() in favor of gcoap_opt_format_finish(), which is called after all options have been entered but before the payload has been written. See [API Options](https://github.com/RIOT-OS/RIOT/wiki/CoAP-API-Options-Update) page for details.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: I agree -- looks like an oversight.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: Yes, as I described in a reply above.

kb2ma:
Thanks for the detailed review, @bergzand! I plan to implement the adjustments above soon.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: That looks right to me. I'll work to simplify this section.

sys/net/application_layer/gcoap/gcoap.c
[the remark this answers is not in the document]
bergzand: Why not keep `NANOCOAP_QS_MAX` and use it here? To me it makes sense to separate these two defines, even if they have the same value.

sys/net/application_layer/gcoap/gcoap.c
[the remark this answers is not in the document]
bergzand: Small note here, I'm not sure whether strncmp is in the C99 libc. (It's not the case with strnlen by default)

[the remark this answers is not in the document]
kb2ma: Use of strncmp() adds 40 bytes to the samr21-xpro executable text space. Since coap_get_uri() documentation asserts that it inserts the terminator, would you be OK with beefing up the unit tests for that function as an alternative way to accomplish the goal?

sys/net/application_layer/gcoap/gcoap.c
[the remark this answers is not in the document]
kb2ma: I am inclined to leave as-is until we have a reason to differentiate max length of query string. Why add a constant we don't need? The use is internal/invisible to the user, so we can add it whenever necessary.

kb2ma:
Added fixups, except for a couple of items with replies above.

sys/net/application_layer/gcoap/gcoap.c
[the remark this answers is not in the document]
bergzand: Agreed, especially since it is a string build in our own functions.

sys/net/application_layer/gcoap/gcoap.c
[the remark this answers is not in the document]
bergzand: To me it feels as if the define is forcefully removed with this PR while there was still a use case for the define. For me defines are a way to make code more readable by giving a meaning to a constant. 

Using the `NANOCOAP_URI_MAX` define here where the use case is not to fix the size of a uri array feels dirty. 

So my opinion would be to keep the `NANOCOAP_QS_MAX`, maybe define the value as `NANOCOAP_URI_MAX`, but to at least keep it here for clarity reasons.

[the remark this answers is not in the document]
kb2ma: I understand your motivation. I removed the macro because nanocoap removed it from generic use in #8772, and left if for gcoap for historical reasons. I just pushed a fixup for gcoap only, as it was before this PR. We can reinstate NANOCOAP_QS_MAX generically for nanocoap when we work on merging #8920.

kb2ma:
Addressed latest comments, including a fixup.

sys/net/application_layer/gcoap/gcoap.c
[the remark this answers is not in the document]
kb2ma: Merged in tests in #9444.

kb2ma:
I plan to rebase on #9474. Actually I have rebased locally with no conflicts. I'd like to test a little to be sure there are no regressions.

kb2ma:
Rebased; tests OK.

kb2ma:
Rebased again due to recent commits. @bergzand, please confirm your concerns have been addressed.

kb2ma:
> @bergzand, please confirm your concerns have been addressed.

ping

sys/include/net/nanocoap.h
```diff
@@ -182,13 +181,24 @@ extern "C" {
...
+#define COAP_FORMAT_LINK           (40)
+#define COAP_FORMAT_OCTET          (42)
+#define COAP_FORMAT_JSON           (50)
+#define COAP_FORMAT_CBOR           (60)
+/**
+ * @brief    nanocoap-internal value to indicate no format specified.
+ *
+ * Not intended for use in a transmitted packet.
+ */
+#define COAP_FORMAT_NONE        (65535)
```
haukepetersen: minor thing: we could improve readability slightly by using `UINT16_MAX` (needs `#include <limits.h>`) here and below.

sys/include/net/nanocoap.h
```diff
@@ -182,13 +181,24 @@ extern "C" {
...
+/**
+ * @brief    nanocoap-internal value to indicate no payload present.
+ *
+ * Not intended for use in a transmitted packet. Must be a 3-byte unsigned
+ * value.
+ */
+#define COAP_FORMAT_NO_PAYLOAD  (65536)
```
haukepetersen: and just noticed: It seems to me this define is only used in gcoap.c. So I think it makes sense to make it local to that file, right?!

kb2ma:
Rebased to bring in #9926.

sys/include/net/nanocoap.h
kb2ma: Good idea. Actually, UINT16_MAX is in <stdint.h>. I was surprised to learn this, too. ;-)

sys/include/net/nanocoap.h
kb2ma: Absolutely, it's a gcoap internal value.

kb2ma:
Pushed review fixups. Next I'll investigate the assertion failure.

kb2ma:
> When trying to test this PR, I get however some failed assertions.

@haukepetersen, cord should work now. Please confirm.

For review, see the updated PR description. There are three new commits:

1. fdb6f34 from rebase on #10223
2. f09d941 to simplify gcoap_finish() and remove preallocation of Content-Format option
3. df9e6dc to fix cord_ep

kb2ma:
Added a warning comment to gcoap_finish() documentation to describe the new limitation on Options in the PDU.

kb2ma:
@haukepetersen, @bergzand

I'd like to get this one (and #10223) merged early this cycle to shake out any other issues. Thanks!

haukepetersen:
On it :-)

I confirm that `cord_ep` and `cord_epsim` are still working as expected, so the changes seem to work.

Will now take another (final?!) look at the code changes and hopefully this PR should be ready to go :-)

sys/include/net/gcoap.h
```diff
@@ -253,26 +253,28 @@ extern "C" {
...
- * @brief   Size of the buffer used to write options, other than Uri-Path, in a
- *          request
+ * @brief   Reduce payload length by this value for a request created with
+ *          gcoap_req_init()
  *
- * Accommodates Content-Format and Uri-Queries
+ * Accommodates writing Content-Format option in gcoap_finish()
  */
-#define GCOAP_REQ_OPTIONS_BUF   (40)
+#define GCOAP_REQ_OPTIONS_BUF   (4)
```
haukepetersen: Looking at this block of changes now, I had two thoughts that came to mind:
1. why not simply join all 3 `GCOAP_X_OPTIONS_BUF` defines so something like `GCOAP_OPTIONS_FINISH_OFFSET` (there is room for improvement on the name here...)
2. more importantly: why are these defines part of the general header? As far as I can tell, they are only used internally in the `gcoap` implementation, so shouldn't they rather be part of `gcoap.c`?

bergzand:
@haukepetersen Feel free to dismiss my review if it is blocking

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -748,15 +659,8 @@ int gcoap_req_init(coap_pkt_t *pdu, uint8_t *buf, size_t len,
...
-        pdu->payload      = buf + coap_get_total_hdr_len(pdu) + strlen(path)
-                                                              + GCOAP_REQ_OPTIONS_BUF;
-        /* Payload length really zero at this point, but we set this to the available
-         * length in the buffer. Allows us to reconstruct buffer length later. */
-        pdu->payload_len  = len - (pdu->payload - buf);
-        pdu->content_type = COAP_FORMAT_NONE;
-
-        memcpy(&pdu->url[0], path, strlen(path));
+        coap_pkt_init(pdu, buf, len - GCOAP_REQ_OPTIONS_BUF, hdrlen);
+        coap_opt_add_string(pdu, COAP_OPT_URI_PATH, path, '/');
```
haukepetersen: Just for clarification: the `coap_opt_add_x` functions do require to be called in the order of the applied option numbers, correct? So does this mean, that we can not add any option with an opt num < COAP_OPT_URI_PATH (11) after calling `gcoap_req_init()`? Or am I completely of track right now?

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -1023,26 +957,26 @@ int gcoap_get_resource_list(void *buf, size_t maxlen, uint8_t cf)
...
-    size_t qs_len = strlen((char *)pdu->qs);
+    char qs[NANOCOAP_QS_MAX];
+    size_t qs_len  = 0;
     size_t key_len = strlen(key);
     size_t val_len = (val) ? (strlen(val) + 1) : 0;
 
-    /* make sure if url_len + the new query string fit into the url buffer */
-    if ((qs_len + key_len + val_len + 2) >= NANOCOAP_QS_MAX) {
+    /* add 2 for '=' and null terminator */
+    if ((key_len + val_len + 2) >= NANOCOAP_QS_MAX) {
```
haukepetersen: the `+ 2` seems wrong, as the `=` character is already accounted for when computing `val_len` above, right?

I propose the following code for this function:
```c
int gcoap_add_qstring(coap_pkt_t *pdu, const char *key, const char *val)
{
    char qs[NANOCOAP_QS_MAX];
    size_t len = strlen(key);
    size_t val_len = (val) ? (strlen(val) + 1) : 0;

    /* test if the query string fits, account for the zero termination */
    if ((len + val_len + 1) >= NANOCOAP_QS_MAX) {
        return -1;
    }

    memcpy(&qs[0], key, len);
    if (val) {
        qs[len] = '=';
        /* the `=` character was already counted in `val_len`, so subtract it here */
        memcpy(&qs[len + 1], val, (val_len - 1));
        len += val_len;
    }
    qs[len] = '\0';

    return coap_opt_add_string(pdu, COAP_OPT_URI_QUERY, qs, '&');
}
```
Its slightly shorter (though the compiled binary has the same size)... I'd say up to you.

haukepetersen:
Also did some testing (under `native`) using the `gcoap` and the `cord` example apps -> all works as expected.

haukepetersen:
@bergzand will do (once it is actually blocking), thanks!

sys/include/net/gcoap.h
kb2ma: I thought about joining them, but I didn't think it was worth it. They are used only for gcoap_finish() which will be deprecated soon in favor of coap_opt_finish(). See the [wiki page](https://github.com/RIOT-OS/RIOT/wiki/CoAP-API-Options-Update) for the plan. I'm happy to make this change though if you prefer.

I will move them to `gcoap.c`. I wasn't thinking about API vs. implementation when I originally put them in the include.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: Yes, you're right that options must be added in order (except see last paragraph). gcoap_req_init() was meant to be a simple way to create a request for the common case where the user does not need to add lower-numbered options. The great thing about the rework of option handling and this PR is that we now have the option to write a message from scratch using coap_build_hdr() and coap_pkt_init(). You could even use gcoap_req_init() as a template for how to do that. 

Actually, now that I look at it, we could allow the path parameter for gcoap_req_init() to be NULL and not write the path option. I'll create a separate PR for that. I like it!

Also, open #9475 sorts options, so in that case options don't need to be added in order. :-)

sys/net/application_layer/gcoap/gcoap.c
kb2ma: Yes, you're right about `+ 2`. I misinterpreted the reason for the addition in the old code. It was for the '&' and the null terminator, not for the '=' and the null. However, we no longer need to consider the '&' for the local `qs` buffer. So, the test should be changed to `+ 1`.

Thanks for the suggested replacement. I'll review.

sys/net/application_layer/gcoap/gcoap.c
```diff
@@ -765,14 +669,42 @@ int gcoap_req_init(coap_pkt_t *pdu, uint8_t *buf, size_t len,
...
+/*
+ * Assumes pdu.payload_len attribute was reduced in gcoap_xxx_init() to
+ * ensure enough space in PDU buffer to write Content-Format option and
+ * payload marker here.
+ */
 ssize_t gcoap_finish(coap_pkt_t *pdu, size_t payload_len, unsigned format)
 {
-    /* reconstruct full PDU buffer length */
-    size_t len = pdu->payload_len + (pdu->payload - (uint8_t *)pdu->hdr);
+    if (pdu->options_len && payload_len) {
```
kb2ma: Actually, we make this friendlier by adding another condition here. We won't check the assertion if the user specifies there is not a Content-Format option (`format` is `COAP_FORMAT_NONE`).

kb2ma:
Thanks, @haukepetersen. Let me know your reactions on `GCOAP_X_OPTIONS_BUF` and any concerns about the change to the assertion check in gcoap_finish().

sys/include/net/gcoap.h
haukepetersen: In that case, joining the defines is indeed a wasted effort. So if you just move them as they are to `gcoap.c`, I am very happy.

sys/net/application_layer/gcoap/gcoap.c
haukepetersen: ok, then I understood it correctly. Should we as intermediate step include a `@waring` in the function doc to explicitly state that one should not add any options with opt num < COAP_OPT_URI_PATH when using this function? I'll leave this up to you...

sys/net/application_layer/gcoap/gcoap.c
haukepetersen: sounds valid, so +1

Side note: I think this check should be written as a single logical expression inside the `assert()` (intentional bad formatting...:
```c
assert( !(pdu->options_len) ||
        !(payload_len) || 
        (pdu->options[pdu->options_len-1].opt_num < COAP_OPT_CONTENT_FORMAT));
```
hope I did the logic correctly here...

haukepetersen:
done.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: I'll incorporate your code. I agree that yours is an improvement and cleans up the memcpy and length accounting for the value. It looks like we were writing two null terminators in the old code.

At the same time, I'll add a couple of unit tests -- with and without a value. The nature of the code and the amount of change in this reimplementation really call for it.

kb2ma: BTW, once this PR is merged, this function definitely should be reimplemented in nanocoap as coap_add_qstring(). It's no longer specific to gcoap. We can retain an inline for the existing function name.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: OK, I'll rework the assert. Your construct is more idiomatic although as you say the logic is a little harder to follow.

kb2ma:
Rebased on master for #10223.

sys/include/net/gcoap.h
kb2ma: Fixed in 88755f8.

sys/net/application_layer/gcoap/gcoap.c
kb2ma: Reimplemented gcoap_add_string() in 33210ac, and added test in 0e275f9.

kb2ma:
All comments addressed.

sys/net/application_layer/gcoap/gcoap.c
haukepetersen: > BTW, once this PR is merged, this function definitely should be reimplemented in nanocoap as coap_add_qstring() or something like that. It's no longer specific to gcoap. We can retain an inline for the existing function name.

+1!

kb2ma:
Done, and all code bits accounted for!

haukepetersen:
@bergzand:
> @haukepetersen Feel free to dismiss my review if it is blocking

done that :-)

haukepetersen:
now let's see what Murdock is saying, but seems we are pretty much ready to go (fingers crossed).

haukepetersen:
all green -> let's go!

PR #6576, opened 2017-02-08
periph/i2c: remodeling part 2: introduce new API

haukepetersen:
EDIT
~~this PR is reabsed on #6575~~

This PR introduces the actual new i2C API as factored out in #4926.

PR #9163, opened 2018-05-22
tools/bpt_interface: Interface for a hardware test device

MrKevinWeiss:
This is an interface and associated firmware for controlling function of the bpt (bluepill tester).  This is a board with (somewhat) qualified firmware (according to specifications) to be used as a slave tester for hardware.  It is intended for regression testing of hardware peripherals across all different boards.  It will start with I2C but add additional peripherals later.

### Contribution description

- Add a py module interface that supports a number of commands

### Issues/PRs references

#6576
#9162
Board firmware is located [here](https://github.com/MrKevinWeiss/Testing)


MrKevinWeiss:
I think I will just link this to my github with a commit hash and make it like a 3rd party tool...  https://github.com/MrKevinWeiss/Testing

PR #9168, opened 2018-05-23
tests/periph_i2c: Update to new I2C API

MrKevinWeiss:
Updated test for command shell interface to new i2c api.  Added some features and simplified a bit.

### Contribution description
-update to new api
-standardize variable names
-add new commands to match api closer
-remove i2c clock speed functionality

### Issues/PRs references
Issue #6577
Involves PR #9162, #6576, #9257

### DIFF
--- a/tests/periph_i2c/main.c
+++ b/tests/periph_i2c/main.c
@@ -25,207 +25,127 @@
 #include "periph/i2c.h"
 #include "shell.h"
 
+#ifndef GET_I2C_CLK_SPD
+#define GET_I2C_CLK_SPD(x)  (-1)
+#endif
+
 #define BUFSIZE        (128U)
 
 static int i2c_dev = -1;
 
-static i2c_speed_t map_speed[] = {
-    I2C_SPEED_LOW,
-    I2C_SPEED_NORMAL,
-    I2C_SPEED_FAST,
-    I2C_SPEED_FAST_PLUS,
-    I2C_SPEED_HIGH
-};
-
-int cmd_init_master(int argc, char **argv)
+int cmd_i2c_assign_dev(int argc, char **argv)
 {
-    int dev, speed, res;
+    int dev;
 
-    if (argc != 3) {
-        puts("Error: Init: Invalid number of arguments!");
-        printf("Usage:\n%s: [DEVICE] [SPEED]\n", argv[0]);
+    if (argc != 2) {
+        puts("Error: Invalid number of arguments!");
+        printf("Usage:\n%s: [DEVICE]\n", argv[0]);
         puts("    with DEVICE:");
         for (unsigned i = 0; i < I2C_NUMOF; i++) {
             printf("          %u -> I2C_%u\n", i, i);
         }
-        puts("         SPEED:");
-        puts("          0 -> SPEED_LOW (10kbit/s)");
-        puts("          1 -> SPEED_NORMAL (100kbit/s)");
-        puts("          2 -> SPEED_FAST (400kbit/s)");
-        puts("          3 -> SPEED_FAST_PLUS (1Mbit/s)");
-        puts("          4 -> SPEED_HIGH (3.4Mbit/s)\n");
         return 1;
     }
-
     dev = atoi(argv[1]);
-    speed = atoi(argv[2]);
-
-    if (speed < 0 || speed > 4) {
-        puts("Error: Speed value not in range (0 - 4)");
-        return 1;
-    }
-
-    res = i2c_init_master(dev, map_speed[speed]);
-    if (res == -1) {
-        puts("Error: Init: Given device not available");
-        return 1;
-    }
-    else if (res == -2) {
-        puts("Error: Init: Unsupported speed value");
+    if (dev < 0 || dev >= (int)I2C_NUMOF) {
+        puts("Error: Given device not available");
         return 1;
     }
     else {
         printf("I2C_%i successfully initialized as master!\n", dev);
         i2c_dev = dev;
+        return 0;
     }
-
-    return 0;
+    return 1;
 }
 
-int cmd_write(int argc, char **argv)
+int cmd_i2c_acquire(int argc, char **argv)
 {
     int res;
-    uint8_t addr;
-    int length = argc - 2;
-    uint8_t data[BUFSIZE];
-
+    (void)argv;
+    if (argc != 1) {
+        puts("Error: Aquire: Invalid number of arguments!");
+        return 1;
+    }
     if (i2c_dev < 0) {
         puts("Error: no I2C device was initialized");
         return 1;
     }
-    if (argc < 3) {
-        puts("Error: not enough arguments given");
-        printf("Usage:\n%s: ADDR BYTE0 [BYTE1 [BYTE_n [...]]]\n", argv[0]);
-        return 1;
-    }
-
-    addr = atoi(argv[1]);
-    for (int i = 0; i < length; i++) {
-        data[i] = atoi(argv[i + 2]);
-    }
-
-    if (length == 1) {
-        printf("i2c_write_byte(I2C_%i, 0x%02x, 0x%02x)\n", i2c_dev, addr, data[0]);
-        res = i2c_write_byte(i2c_dev, addr, data[0]);
-    }
-    else {
-        printf("i2c_write_bytes(I2C_%i, 0x%02x, [", i2c_dev, addr);
-        for (int i = 0; i < length; i++) {
-            printf(", 0x%02x", data[i]);
-        }
-        puts("])");
-        res = i2c_write_bytes(i2c_dev, addr, data, length);
-    }
-
-    if (res < 0) {
-        puts("Error: no bytes were written");
-        return 1;
-    }
-    else {
-        printf("I2C_%i: successfully wrote %i bytes to the bus\n", i2c_dev, res);
+    res = i2c_acquire(i2c_dev);
+    if (res == 0) {
+        printf("I2C_%i successfully acquired!\n", i2c_dev);
         return 0;
     }
+    else {
+        printf("Error: Device busy\n");
+        return 1;
+    }
+    return 1;
 }
 
-int cmd_write_reg(int argc, char **argv)
+int cmd_i2c_release(int argc, char **argv)
+{
+    (void)argv;
+    if (argc != 1) {
+        puts("Error: Aquire: Invalid number of arguments!");
+        return 1;
+    }
+    if (i2c_dev < 0) {
+        puts("Error: no I2C device was initialized");
+        return 1;
+    }
+    i2c_release(i2c_dev);
+    printf("I2C_%i successfully released!\n", i2c_dev);
+    return 0;
+}
+
+int cmd_i2c_read_reg(int argc, char **argv)
 {
     int res;
-    uint8_t addr, reg;
-    int length = argc - 3;
-    uint8_t data[BUFSIZE];
+    uint8_t addr, reg, flags;
+    uint8_t data[1];
 
     if (i2c_dev < 0) {
         puts("Error: no I2C device initialized");
         return 1;
     }
-    if (length < 1) {
+    if (argc < 3) {
         puts("Error: not enough arguments given");
-        printf("Usage:\n%s ADDR REG BYTE0 [BYTE1 ...]\n", argv[0]);
+        printf("Usage:\n%s ADDR REG FLAG\n", argv[0]);
         return 1;
     }
 
     addr = atoi(argv[1]);
     reg = atoi(argv[2]);
-    for (int i = 0; i < length; i++) {
-        data[i] = atoi(argv[i + 3]);
-    }
+    flags = atoi(argv[3]);
 
-    if (length == 1) {
-        printf("i2c_write_reg(I2C_%i, 0x%02x, 0x%02x, 0x%02x)\n",
-               i2c_dev, addr, reg, data[0]);
-        res = i2c_write_reg(i2c_dev, addr, reg, data[0]);
-    }
-    else {
-        printf("i2c_write_regs(I2C_%i, 0x%02x, 0x%02x, [", i2c_dev, addr, reg);
-        for (int i = 0; i < length; i++) {
-            printf("0x%02x, ", data[i]);
-        }
-        puts("])");
-        res = i2c_write_regs(i2c_dev, addr, reg, data, length);
-    }
+    printf("i2c_read_reg(I2C_%i, 0x%02x, 0x%02x)\n", i2c_dev, addr, reg);
+    res = i2c_read_reg(i2c_dev, addr, reg, data, flags);
 
-    if (res < 1) {
-        puts("Error: no bytes were written");
+    if (res == I2C_ADDR_NACK) {
+        puts("Error: addr NACK");
         return 1;
     }
-    else {
-        printf("I2C_%i: successfully wrote %i bytes to register 0x%02x\n", i2c_dev, res, reg);
-        return 0;
-    }
-}
-
-int cmd_read(int argc, char **argv)
-{
-    int res;
-    uint8_t addr;
-    int length;
-    uint8_t data[BUFSIZE];
-
-    if (i2c_dev < 0) {
-        puts("Error: no I2C device initialized");
+    else if (res == I2C_ERR) {
+        puts("Error: i2c return error");
         return 1;
     }
-    if (argc < 3) {
-        puts("Error: not enough arguments given");
-        printf("Usage:\n%s ADDR LENGTH\n", argv[0]);
-        return 1;
-    }
-
-    addr = atoi(argv[1]);
-    length = atoi(argv[2]);
-
-    if (length < 1 || length > (int)BUFSIZE) {
-        puts("Error: invalid LENGTH parameter given\n");
-        return 1;
-    }
-    else if (length == 1) {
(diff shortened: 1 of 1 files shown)


tests/periph_i2c/main.c
```diff
@@ -29,253 +29,264 @@
...
-        puts("          4 -> SPEED_HIGH (3.4Mbit/s)\n");
         return 1;
     }
 
     dev = atoi(argv[1]);
-    speed = atoi(argv[2]);
 
-    if (speed < 0 || speed > 4) {
-        puts("Error: Speed value not in range (0 - 4)");
+    res = i2c_init(dev);
```
dylad: i2c_init must not be call there. It only belongs to drivers/periph_common/init.c

tests/periph_i2c/main.c
```diff
@@ -29,253 +29,264 @@
...
+    if (res == -1) {
+        puts("Error: Init: Given device not available");
+        return 1;
+    }
+    else if (res == 0){
```
dylad: space before curly brace

MrKevinWeiss:
@dylad I tested with #9198 and it seems to function, can we merge?

dylad:
Let me (or someone else ) test it first and give an ACK.
Plus I would like to test on two different CPUs to validate this.

MrKevinWeiss:
@dylad So it is a bit heavy but should support most all API and have a few nice features for later.  Eventually I will do a minimal shell (binary based) but for now this should do.  Once everything is up to date and merged it would be nice to make buildtest and see if memory issues occur on any board (right now they all fail because the API isn't fully ported).  [Here](https://github.com/MrKevinWeiss/RIOT/tree/kw/samr2) is it in working with the samr21 (ie, the API is working).

tests/periph_i2c/main.c
```diff
@@ -25,265 +25,525 @@
...
     }
+    else {
+        printf("Success: i2c_%i successfully assigned as master\n", dev);
+        i2c_dev = dev;
+        return 0;
+    }
+    return 1;
+}
+
+int cmd_i2c_acquire(int argc, char **argv) {
```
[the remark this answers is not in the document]
MrKevinWeiss: Yup I know, I think if it is in the API it should be tested (the fact that it blocks shouldn't be a problem for the test since I should be able to expect timeout and provide resets).  If you don't wish to have this tested in the automated test we can remove it but I think since it is exposed to the user it should be tests.  Also some clarification on it's expected behavior would be nice (ie is it required to acquire before use? Can we release twice without getting an error?)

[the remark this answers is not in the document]
dylad: You're right I mess up things sorry. acquire doesn't belong to CPU driver but device driver so we must be able to test the acquire/release function. 
You can leave it as it.

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
+    }
+    else if (res == -EIO) {
+        printf("Error: EIO [%d]\n", -res);
+        return 1;
+    }
+    else if (res == -ETIMEDOUT) {
+        printf("Error: ETIMEDOUT [%d]\n", -res);
+        return 1;
+    }
+    else if (res == I2C_ACK) {
```
smlng: From the code below, this will never be used but is handled in the function with specific success output.
Meaning: this could be removed, or not?

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
+    if (i2c_dev < 0) {
+        puts("Error: no I2C device was initialized");
+        return 1;
+    }
+    return 0;
+}
+
+int print_i2c_error(int res) {
+    if (res == -EOPNOTSUPP) {
+        printf("Error: EOPNOTSUPP [%d]\n", -res);
```
smlng: here and below: does it make sense to print the number, too? IMHO simply using `puts("Error: EOPNOTSUPP!");` should be fine and enough.

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
-    I2C_SPEED_FAST_PLUS,
-    I2C_SPEED_HIGH
-};
 
-int cmd_init_master(int argc, char **argv)
-{
-    int dev, speed, res;
+int check_dev(void) {
+    if (i2c_dev < 0) {
+        puts("Error: no I2C device was initialized");
```
smlng: there is `ENODEV` which you could use here and print similarly to the other errors? See `print_i2c_error`, too

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
-        puts("          1 -> SPEED_NORMAL (100kbit/s)");
-        puts("          2 -> SPEED_FAST (400kbit/s)");
-        puts("          3 -> SPEED_FAST_PLUS (1Mbit/s)");
-        puts("          4 -> SPEED_HIGH (3.4Mbit/s)\n");
         return 1;
     }
-
     dev = atoi(argv[1]);
-    speed = atoi(argv[2]);
+    if (dev < 0 || dev >= (int)I2C_NUMOF) {
```
smlng: minor nit, but this can be simplified to:
```
if (dev < 0 || dev >= (int)I2C_NUMOF) {
    puts("Error: Given device not available");
    i2c_dev = -1;
    return 1;
}
printf("Success: i2c_%i successfully assigned as master\n", dev);
i2c_dev = dev;
return 0;
```

omitting the `else` branch because of the `return` in the `if` part and also the final `return 1` would've never been reached, too.

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
+
+int cmd_i2c_acquire(int argc, char **argv) {
+    int res;
+    (void)argv;
+
+    if (check_dev() == 1){
+        return 1;
+    }
+
+    if (argc != 1) {
```
smlng: why bother? You don't expect nor use arguments, so (IMHO) no need to check the number.

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
+    int res;
+
+    if (check_dev() == 1){
         return 1;
     }
 
-    res = i2c_init_master(dev, map_speed[speed]);
-    if (res == -1) {
-        puts("Error: Init: Given device not available");
+    if (argc != 1) {
```
smlng: dito, see above

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
+    if (argc == 4) {
+        flags = atoi(argv[3]);
+    }
+
+    printf("Command: i2c_read_reg(i2c_%i, 0x%02x, 0x%02x)\n", i2c_dev, addr, reg);
+    res = i2c_read_reg(i2c_dev, addr, reg, &data, flags);
+
+
+    if (res == I2C_ACK) {
+        printf("Success: i2c_%i successfully read 1 byte from reg 0x%02x: [",
```
smlng: why not having one `printf` only, instead of 2 + `puts`?

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
+    addr = atoi(argv[1]);
+    if (argc == 3) {
+        flags = atoi(argv[2]);
+    }
+
+    printf("Command: i2c_read_byte(i2c_%i, 0x%02x, 0x%02x)\n", i2c_dev, addr, flags);
+    res = i2c_read_byte(i2c_dev, addr, &data, flags);
+
+    if (res == I2C_ACK) {
+        printf("Success: i2c_%i successfully read 1 byte: [", i2c_dev);
```
smlng: nit picky: omit `successfully` because its already a `Success` 😉 

applies to other output above and below, too.

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
-        printf("i2c_write_bytes(I2C_%i, 0x%02x, [", i2c_dev, addr);
-        for (int i = 0; i < length; i++) {
-            printf(", 0x%02x", data[i]);
+        printf("Command: i2c_read_regs(i2c_%i, 0x%02x, 0x%02x, %i)\n", i2c_dev,
+            addr, reg, len);
+        res = i2c_read_regs(i2c_dev, addr, reg, data, len, flags);
+    }
+
+    if (res == I2C_ACK) {
+        printf("Success: i2c_%i successfully read %i bytes from reg 0x%02x: [",
```
smlng: IMHO its a bit confusing, that the user puts in the address of a register in decimal and its printed in HEX here, or not? At least from what I found `atoi` expects numbers of base 10, but you could use `strtol` with base 16 instead.

smlng: and btw. maybe this printing could be a helper function? Something like:

```
static inline print_i2c_read(i2c_t dev, uint16_t *reg, uint8_t *buf, size_t len)
{
    assert ((dev > 0) && (len > 0));
    printf("Success: i2c_%i read %i byte(s) ", dev, len);
    if (reg != Null) {
        printf("from reg 0x%02x ", reg);
    }
    printf(": [ ");
    for (unsigned i = 0; i < len; i++) {
        printf("0x%02x ", data[i]);
    }
    puts("]");
}
```

and then be use in `i2c_read_reg(s)/byte(s)`

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,547 @@
...
+    return print_i2c_error(res);
 }
 
-int cmd_write(int argc, char **argv)
-{
+int cmd_i2c_read_regs(int argc, char **argv) {
     int res;
     uint8_t addr;
-    int length = argc - 2;
+    uint8_t reg;
```
smlng: is this enough, because the API uses `uint16_t`, or not?

tests/periph_i2c/main.c
MrKevinWeiss: I'd like to leave it in, just in case somebody does something in the future it won't say unknown error since nothing specifies that one must handle the ACK first.

tests/periph_i2c/main.c
MrKevinWeiss: The number is more there if the python (or any other) interface wants to use it, since it is a standard errno it may be easy just to feed it into a standard decoding function and get more info (I also don't think it adds too much bloat).

tests/periph_i2c/main.c
MrKevinWeiss: Yup, nice catch, will do! [TODO]

tests/periph_i2c/main.c
MrKevinWeiss: Yup, will do. [TODO]

tests/periph_i2c/main.c
MrKevinWeiss: Yup, legacy stuff.  [TODO]

tests/periph_i2c/main.c
MrKevinWeiss: Yup, legacy stuff [TODO]

tests/periph_i2c/main.c
MrKevinWeiss: I may just do strtol (legacy used atoi), I know lots of times I like seeing the hex numbers (since most data sheets support that).  With the interface it gets abstracted out though.  Either way I'll improve somehow.  [TODO]

tests/periph_i2c/main.c
MrKevinWeiss: Yup legacy, will do. [TODO] x3

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,532 @@
...
-        puts("          4 -> SPEED_HIGH (3.4Mbit/s)\n");
-        return 1;
+        printf("0x%02x", buf[i]);
     }
+    puts("]");
+}
 
-    dev = atoi(argv[1]);
-    speed = atoi(argv[2]);
+static inline int get_num(char *str) {
```
keestux: Please put function opening curly on a new line

tests/periph_i2c/main.c
```diff
@@ -29,253 +29,264 @@
...
+        return 1;
     }
+    return 1;
+}
 
+int cmd_i2c_release(int argc, char **argv)
+{
+    (void)argv;
+    if (argc != 1) {
+        puts("Error: Aquire: Invalid number of arguments!");
```
keestux: Typo: acquire

tests/periph_i2c/main.c
```diff
@@ -29,19 +29,23 @@
...
 #define INVALID_ARGS    puts("Error: Invalid number of arguments");
 
 #define BUFSIZE        (128U)
 
 static int i2c_dev = -1;
 
-int print_i2c_error(int res)
-{
+
+int check_dev(void) {
```
keestux: More curlies that should go on a new line

tcschmidt:
@smlng @MrKevinWeiss can we resolve the controversial change request?

tests/periph_i2c/main.c
smlng: fine with me

dylad:
If @smlng thinks this PR is good to go after the final comment is adressed then I guess we can go :)

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,550 @@
...
-    I2C_SPEED_LOW,
-    I2C_SPEED_NORMAL,
-    I2C_SPEED_FAST,
-    I2C_SPEED_FAST_PLUS,
-    I2C_SPEED_HIGH
-};
+static inline void print_i2c_read(i2c_t dev, uint16_t *reg, uint8_t *buf,
+    size_t len)
+{
+    printf("Success: i2c_%i read %i byte(s) ", dev, len);
```
smlng: travis complains here, should `printf("Success: i2c_%i read %i byte(s) ", dev, (int)len);`  for instance.

smlng: casting length from `size_t` to `int` is the easiest, I guess.

smlng:
@dylad apart from the minor issue travis reports, this is good to go. And @MrKevinWeiss would be happy to get this done (over with 😬) and work on refining the test scripts eventually making I2C testing easier for everyone, that is: no need to merge different PRs together to run tests.

dylad:
@smlng I'm not an expert in testing/CI and so but this looks OK to me. If all comments + Travis are addressed, we can click on the big green button :)

tests/periph_i2c/main.c
smlng: @MrKevinWeiss can you fix this small issue before merge, because it will hit us anyway when this branch will be merged back into master.

MrKevinWeiss: oops missed that, sure!

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,550 @@
...
+
+int cmd_i2c_acquire(int argc, char **argv)
+{
+    int res;
 
-    if (speed < 0 || speed > 4) {
-        puts("Error: Speed value not in range (0 - 4)");
+    (void)argv;
+    (void)argc;
+    if (check_dev() == 1){
```
keestux: space before curly (there are more like these)

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,550 @@
...
     }
-    else {
-        printf("I2C_%i: successfully read %i bytes from reg 0x%02x:\n  [", i2c_dev, res, reg);
-        for (int i = 0; i < res; i++) {
-            printf("0x%02x, ", data[i]);
-        }
-        puts("]");
-        return 0;
+
+    switch (GET_I2C_CLK_SPD(i2c_dev)) {
```
keestux: The indenting of this switch/case statement is somewhat unusual. The Coding Convention does not say anything specific about it, but I like to see an extra indentation for the statements besides the "case".

    switch (GET_I2C_CLK_SPD(i2c_dev)) {
        case case I2C_SPEED_LOW:
            printf("Success: i2...");
            break;
        case I2C_SPEED_NORMAL:

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,550 @@
...
+        printf("from reg 0x%02x ", *reg);
+    }
+    printf(": [");
+    for (int i = 0; i < len; i++) {
+        if (i != 0){
+            printf(", ");
+        }
+        printf("0x%02x", buf[i]);
+    }
+    puts("]");
```
keestux: Why the mix of puts and printf?
Personally I prefer to see just printf's, for consistency.

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,550 @@
...
+static inline void print_i2c_read(i2c_t dev, uint16_t *reg, uint8_t *buf,
+    int len)
+{
+    printf("Success: i2c_%i read %i byte(s) ", dev, len);
+    if (reg != NULL) {
+        printf("from reg 0x%02x ", *reg);
+    }
+    printf(": [");
+    for (int i = 0; i < len; i++) {
+        if (i != 0){
```
keestux: space before curly

tests/periph_i2c/main.c
```diff
@@ -20,270 +20,550 @@
...
+            printf(", ");
+        }
+        printf("0x%02x", buf[i]);
+    }
+    puts("]");
+}
 
-int cmd_init_master(int argc, char **argv)
+static inline int get_num(char *str)
```
keestux: I'd like to see "const char *" for the argument

dylad:
If everything is addressed, I think you can squash.

smlng:
@dylad we squashed it, you wanna hit merge?

dylad:
Here we go !
Good job guys !

PR #9167, opened 2018-05-23
periph/i2c: add default fallback wrappers I2C functions

dylad:
### Contribution description
This PR provides the default fallback implementations of basic I2C functions.
Each CPU can overwrite those functions if it's needed.
The behaviour is actually borrowed from SPI bus.

PR #9169, opened 2018-05-23
drivers/periph: fix doxygen groups in implementation

aabadie:
### Contribution description

While reviewing #9167, I realized that some periph implementation files were not added to the right groups.

This PR fixes that.

### Issues/PRs references

None

### DIFF
--- a/drivers/periph_common/cpuid.c
+++ b/drivers/periph_common/cpuid.c
@@ -9,7 +9,7 @@
  */
 
 /**
- * @addtogroup  drivers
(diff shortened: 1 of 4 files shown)


gebart:
Weird that this is giving build failures about insufficient target RAM. I guess some toolchain update on the CI server is the culprit

aabadie:
it has not been rebase on latest master and contain the issue related to the wrong nucleo board name. I'll update asap.

aabadie:
All green, go

PR #9185, opened 2018-05-24
driver/at: add option to keep EOL character in readline, and make EOL configurable

vincent-d:
### Contribution description

With modem, I have some use cases where I want to keep all EOL characters in the response when calling `at_send_cmd_get_lines`. For instance when reading a file or an HTTP response.

This PR add a parameter to `at_send_cmd_get_lines` thus  to `at_readline` to not remove EOL characters. So far '\r' was removed.

It also let the user configure the expected response EOL characters with 2 macros.

(this is not yet tested)

### Issues/PRs references

Follow-up #7084


drivers/at/at.c
```diff
@@ -222,10 +232,12 @@ ssize_t at_readline(at_dev_t *dev, char *resp_buf, size_t len, uint32_t timeout)
...
             }
-            if (*resp_pos == '\r') {
-                continue;
+            if (sizeof(eol) > 2 && *resp_pos == eol[0]) {
+                if (!keep_eol) {
+                    continue;
+                }
             }
-            if (*resp_pos == '\n') {
+            if (*resp_pos == eol[sizeof(eol) - 1]) {
```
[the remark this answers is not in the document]
vincent-d: Changed to `if (*resp_pos == eol[sizeof(eol) - 2])`. The `sizeof(eol) - 1` was a mistake, as it should always be `\0`. Now the last character of `eol[]` is used, so either `AT_RECV_EOL_2` if both are set or `AT_RECV_EOL_1` if the former is empty.

aabadie:
@vincent-d, is this PR tested on your side ?
Otherwise, I tested it with the Arduino MKRWAN1300, using `\r` in EOL1 and an empty EOL2 and it works like a charm.
I'm inclined to ACK but maybe @kaspar030 is also interested by having a look/testing.

drivers/at/at.c
```diff
@@ -128,17 +131,21 @@ ssize_t at_send_cmd_get_lines(at_dev_t *dev, const char *command,
...
+            if (bytes_left) {
+                *pos++ = eol[sizeof(eol) - 2];
+                bytes_left--;
+            }
             continue;
         }
         else if (res > 0) {
             bytes_left -= res;
-            if ((res == 2) && (strncmp(pos, "OK", 2) == 0)) {
+            if ((res == (2 + (keep_eol ? 1 : 0))) && (strncmp(pos, "OK", 2) == 0)) {
```
kaspar030: here and below: if ```keep_eol```, it has values 0 or 1, thus the ternary operator can be removed. e.g., just use ```2 + keep_eol)```

vincent-d: changed

vincent-d:
tested with ublox-c030-u201

vincent-d:
squashed

aabadie:
@vincent-d, there are failing tests on Murdock ('unknown type name 'bool'', an include is missing I think)

vincent-d:
> there are failing tests on Murdock

Sorry, should be fixed (amended directly)

PR #9190, opened 2018-05-25
drivers/gpio_exp: map GPIO expanders onto gpio_t

ZetaR60:
This PR makes handling of GPIO expanders invisible to code using the GPIO pin API (periph/gpio.h). This is accomplished by reserving part of the range of values of gpio_t. When a call to the GPIO API uses a pin that falls within this range, it is parsed into a device ID that is looked up in the GPIO expansion registery and the call is redirected to the corresponding device.

Mostly done, but still a bit WIP.


ZetaR60:
I did a brief survey of how each system that defines gpio.c handles gpio_t, and the following systems look to be easy to add mapping support (in the same way as the atmega_common example):
- atmega_common
- cc26x0
- efm32
- ezr32wg
- lm4f120
- lpc2387
- mips_pic32_common
- nrf5x_common
- stm32f1

The following systems can be made compatible by raising the size of gpio_t to 16bit / 32bit (or changing how gpio_t is handled):
- kinetis
- lpc1768
- msp430fxyz

These systems use all of a 32bit gpio_t, and would require modifications to how gpio_t is used (unless gpio_t is raised to 64bits):
- cc2538
- sam0_common
- sam3
- stm32_common

ZetaR60:
Rebased to solve conflict. I have added two test routines: one for runtime and one for compile time. Neither require an actual GPIO expander (since the system will redirect to any set of functions if you tell it that it is a "driver"). I think the tests are complete enough to remove the explicit dependency on #9054 (edited from my original post).

ZetaR60:
Travis:
```
cpu/atmega_common/periph/gpio.c:120: style (variableScope): The scope of the variable 'exp_entry' can be reduced.
...etc
```
Uhm, no it can't?

gebart:
Re cppcheck: You can move the exp_entry declaration inside the if block

ZetaR60:
Thanks @gebart. I had thought that variable declaration in blocks other than the outer function block was not guaranteed by the standard (other than in for loops), but a search says otherwise.

ZetaR60:
Sorry for the large number of fixups. This PR is now tested and working properly on mega-xplained.

There may be an issue with the method of tests/driver_gpio_exp_coll, because some CPUs have a GPIO_PIN definition that uses a cast, which is not supported in preprocessor conditions. I may be able to work around it by defining a macro for ```(gpio_t)```, but I will have to do some experimenting.

```
[liveuser@localhost-live driver_gpio_exp]$ make BOARD=mega-xplained test
./tests/01-run.py
/home/liveuser/Desktop/RIOT_gpio_exp/dist/tools/pyterm/pyterm -p "/dev/ttyACM0" -b "9600"
2018-05-27 20:42:42,810 - INFO # Connect to serial port /dev/ttyACM0
Welcome to pyterm!
Type '/exit' to exit.
2018-05-27 20:42:43,814 - INFO # 

2018-05-27 20:42:46,713 - INFO # main(): This is RIOT! (Version: 2018.07-devel-386-gdf718-localhost-live-RIOT_gpio_exp)
2018-05-27 20:42:46,780 - INFO # gpio_exp test routine
2018-05-27 20:42:46,783 - INFO # Running gpio.h functions on pin 1
2018-05-27 20:42:46,812 - INFO # init on dev 0xbeef with pin 1
2018-05-27 20:42:46,879 - INFO # init_int on dev 0xbeef with pin 1
2018-05-27 20:42:46,880 - INFO # irq on dev 0xbeef with pin 1
2018-05-27 20:42:46,912 - INFO # irq on dev 0xbeef with pin 1
2018-05-27 20:42:46,980 - INFO # read on dev 0xbeef with pin 1
2018-05-27 20:42:46,982 - INFO # write on dev 0xbeef with pin 1
2018-05-27 20:42:47,013 - INFO # write on dev 0xbeef with pin 1
2018-05-27 20:42:47,078 - INFO # read on dev 0xbeef with pin 1
2018-05-27 20:42:47,081 - INFO # write on dev 0xbeef with pin 1
2018-05-27 20:42:47,112 - INFO # write on dev 0xbeef with pin 1
2018-05-27 20:42:47,178 - INFO # Running gpio.h functions on pin 2
2018-05-27 20:42:47,180 - INFO # init on dev 0xbeef with pin 2
2018-05-27 20:42:47,214 - INFO # init_int on dev 0xbeef with pin 2
2018-05-27 20:42:47,278 - INFO # irq on dev 0xbeef with pin 2
2018-05-27 20:42:47,279 - INFO # irq on dev 0xbeef with pin 2
2018-05-27 20:42:47,281 - INFO # read on dev 0xbeef with pin 2
2018-05-27 20:42:47,314 - INFO # write on dev 0xbeef with pin 2
2018-05-27 20:42:47,379 - INFO # write on dev 0xbeef with pin 2
2018-05-27 20:42:47,380 - INFO # read on dev 0xbeef with pin 2
2018-05-27 20:42:47,413 - INFO # write on dev 0xbeef with pin 2
2018-05-27 20:42:47,480 - INFO # write on dev 0xbeef with pin 2
2018-05-27 20:42:47,483 - INFO # Running gpio.h functions on pin 4
2018-05-27 20:42:47,514 - INFO # init on dev 0xbeef with pin 4
2018-05-27 20:42:47,580 - INFO # init_int on dev 0xbeef with pin 4
2018-05-27 20:42:47,583 - INFO # irq on dev 0xbeef with pin 4
2018-05-27 20:42:47,614 - INFO # irq on dev 0xbeef with pin 4
2018-05-27 20:42:47,681 - INFO # read on dev 0xbeef with pin 4
2018-05-27 20:42:47,683 - INFO # write on dev 0xbeef with pin 4
2018-05-27 20:42:47,713 - INFO # write on dev 0xbeef with pin 4
2018-05-27 20:42:47,780 - INFO # read on dev 0xbeef with pin 4
2018-05-27 20:42:47,783 - INFO # write on dev 0xbeef with pin 4
2018-05-27 20:42:47,813 - INFO # write on dev 0xbeef with pin 4
2018-05-27 20:42:47,879 - INFO # Running gpio.h functions on pin 8
2018-05-27 20:42:47,880 - INFO # init on dev 0xbeef with pin 8
2018-05-27 20:42:47,913 - INFO # init_int on dev 0xbeef with pin 8
2018-05-27 20:42:47,979 - INFO # irq on dev 0xbeef with pin 8
2018-05-27 20:42:47,982 - INFO # irq on dev 0xbeef with pin 8
2018-05-27 20:42:48,013 - INFO # read on dev 0xbeef with pin 8
2018-05-27 20:42:48,079 - INFO # write on dev 0xbeef with pin 8
2018-05-27 20:42:48,080 - INFO # write on dev 0xbeef with pin 8
2018-05-27 20:42:48,113 - INFO # read on dev 0xbeef with pin 8
2018-05-27 20:42:48,179 - INFO # write on dev 0xbeef with pin 8
2018-05-27 20:42:48,181 - INFO # write on dev 0xbeef with pin 8
2018-05-27 20:42:48,212 - INFO # Running gpio.h functions on pin 16
2018-05-27 20:42:48,280 - INFO # init on dev 0xbeef with pin 16
2018-05-27 20:42:48,283 - INFO # init_int on dev 0xbeef with pin 16
2018-05-27 20:42:48,312 - INFO # irq on dev 0xbeef with pin 16
2018-05-27 20:42:48,380 - INFO # irq on dev 0xbeef with pin 16
2018-05-27 20:42:48,382 - INFO # read on dev 0xbeef with pin 16
2018-05-27 20:42:48,412 - INFO # write on dev 0xbeef with pin 16
2018-05-27 20:42:48,480 - INFO # write on dev 0xbeef with pin 16
2018-05-27 20:42:48,482 - INFO # read on dev 0xbeef with pin 16
2018-05-27 20:42:48,512 - INFO # write on dev 0xbeef with pin 16
2018-05-27 20:42:48,578 - INFO # write on dev 0xbeef with pin 16
2018-05-27 20:42:48,580 - INFO # Running gpio.h functions on pin 32
2018-05-27 20:42:48,613 - INFO # init on dev 0xbeef with pin 32
2018-05-27 20:42:48,679 - INFO # init_int on dev 0xbeef with pin 32
2018-05-27 20:42:48,680 - INFO # irq on dev 0xbeef with pin 32
2018-05-27 20:42:48,713 - INFO # irq on dev 0xbeef with pin 32
2018-05-27 20:42:48,779 - INFO # read on dev 0xbeef with pin 32
2018-05-27 20:42:48,782 - INFO # write on dev 0xbeef with pin 32
2018-05-27 20:42:48,812 - INFO # write on dev 0xbeef with pin 32
2018-05-27 20:42:48,879 - INFO # read on dev 0xbeef with pin 32
2018-05-27 20:42:48,882 - INFO # write on dev 0xbeef with pin 32
2018-05-27 20:42:48,913 - INFO # write on dev 0xbeef with pin 32
2018-05-27 20:42:48,980 - INFO # Running gpio.h functions on pin 64
2018-05-27 20:42:48,982 - INFO # init on dev 0xbeef with pin 64
2018-05-27 20:42:49,013 - INFO # init_int on dev 0xbeef with pin 64
2018-05-27 20:42:49,079 - INFO # irq on dev 0xbeef with pin 64
2018-05-27 20:42:49,080 - INFO # irq on dev 0xbeef with pin 64
2018-05-27 20:42:49,113 - INFO # read on dev 0xbeef with pin 64
2018-05-27 20:42:49,179 - INFO # write on dev 0xbeef with pin 64
2018-05-27 20:42:49,182 - INFO # write on dev 0xbeef with pin 64
2018-05-27 20:42:49,213 - INFO # read on dev 0xbeef with pin 64
2018-05-27 20:42:49,279 - INFO # write on dev 0xbeef with pin 64
2018-05-27 20:42:49,282 - INFO # write on dev 0xbeef with pin 64
2018-05-27 20:42:49,313 - INFO # Running gpio.h functions on pin 128
2018-05-27 20:42:49,379 - INFO # init on dev 0xbeef with pin 128
2018-05-27 20:42:49,380 - INFO # init_int on dev 0xbeef with pin 128
2018-05-27 20:42:49,413 - INFO # irq on dev 0xbeef with pin 128
2018-05-27 20:42:49,481 - INFO # irq on dev 0xbeef with pin 128
2018-05-27 20:42:49,483 - INFO # read on dev 0xbeef with pin 128
2018-05-27 20:42:49,513 - INFO # write on dev 0xbeef with pin 128
2018-05-27 20:42:49,579 - INFO # write on dev 0xbeef with pin 128
2018-05-27 20:42:49,579 - INFO # read on dev 0xbeef with pin 128
2018-05-27 20:42:49,613 - INFO # write on dev 0xbeef with pin 128
2018-05-27 20:42:49,680 - INFO # write on dev 0xbeef with pin 128
2018-05-27 20:42:49,682 - INFO # Running notsup functions
2018-05-27 20:42:49,713 - INFO # (they should not print output)
2018-05-27 20:42:49,779 - INFO # Running notsup gpio.h functions on pin 0
2018-05-27 20:42:49,812 - INFO # Running notsup gpio.h functions on pin 1
2018-05-27 20:42:49,880 - INFO # Running notsup gpio.h functions on pin 2
2018-05-27 20:42:49,883 - INFO # Running notsup gpio.h functions on pin 3
2018-05-27 20:42:49,979 - INFO # Running notsup gpio.h functions on pin 4
2018-05-27 20:42:49,983 - INFO # Running notsup gpio.h functions on pin 5
2018-05-27 20:42:50,013 - INFO # Running notsup gpio.h functions on pin 6
2018-05-27 20:42:50,080 - INFO # Running notsup gpio.h functions on pin 7
2018-05-27 20:42:50,113 - INFO # Checking that all pins in range have init error using notsup
2018-05-27 20:42:50,214 - INFO # (lack of init error implies improper redirection)
2018-05-27 20:42:50,214 - INFO # [SUCCESS]

[liveuser@localhost-live driver_gpio_exp]$ echo $?
0
```

```
Installed compiler toolchains 
-----------------------------
             native gcc: missing
      arm-none-eabi-gcc: missing
                avr-gcc: avr-gcc (Fedora 7.2.0-1.fc27) 7.2.0
       mips-mti-elf-gcc: missing
             msp430-gcc: missing
   riscv-none-embed-gcc: missing
                  clang: missing

Installed compiler libs
-----------------------
   arm-none-eabi-newlib: missing
    mips-mti-elf-newlib: missing

Installed development tools
---------------------------
                  cmake: cmake version 3.11.0
               cppcheck: missing
                doxygen: missing
                 flake8: missing
                    git: git version 2.14.2
             coccinelle: missing
```

ZetaR60:
tests/driver_gpio_exp_coll now does its testing using the compiler using static_assert, rather than the preprocessor. It can understand the enum port definitions (no more hard-coding), casts, sizeof, etc.

This PR now requires #9279

ZetaR60:
#9279 has been merged: the dependency has been satisfied.

ZetaR60:
Rebased on trunk to bring in new CPU (fe310) and resolved conflicts. All CPUs with conventional GPIO definitions have support added, but some have collisions. None of the CPUs with odd GPIO definitions have been added yet. The compile time collision detection in driver_gpio_exp_coll is confirmed to be working properly and is detecting collisions in a couple of the CPUs. Example output when detecting a collision:

```
user@computer:~/Desktop/files/projects/acutam/firmware/RIOT_gpio_exp_docker/tests/driver_gpio_exp_coll$ make BOARD=bluepill
/home/user/Desktop/files/projects/acutam/firmware/RIOT_gpio_exp_docker/makefiles/toolchain/gnu.inc.mk:17: objcopy not found. Hex file will not be created.
docker run --rm -t -u "$(id -u)" \
    -v '/home/user/Desktop/files/projects/acutam/firmware/RIOT_gpio_exp_docker:/data/riotbuild/riotbase' \
    -v '/home/user/Desktop/files/projects/acutam/firmware/RIOT_gpio_exp_docker/cpu:/data/riotbuild/riotcpu' \
    -v '/home/user/Desktop/files/projects/acutam/firmware/RIOT_gpio_exp_docker/boards:/data/riotbuild/riotboard' \
    -v '/home/user/Desktop/files/projects/acutam/firmware/RIOT_gpio_exp_docker/makefiles:/data/riotbuild/riotmake' \
    -v '/home/user/Desktop/files/projects/acutam/firmware/RIOT_gpio_exp_docker:/data/riotbuild/riotproject' \
    -v /etc/localtime:/etc/localtime:ro \
    -e 'RIOTBASE=/data/riotbuild/riotbase' \
    -e 'CCACHE_BASEDIR=/data/riotbuild/riotbase' \
    -e 'RIOTCPU=/data/riotbuild/riotcpu' \
    -e 'RIOTBOARD=/data/riotbuild/riotboard' \
    -e 'RIOTMAKE=/data/riotbuild/riotmake' \
    -e 'RIOTPROJECT=/data/riotbuild/riotproject' \
    -e 'BOARD=bluepill' \
    -w '/data/riotbuild/riotproject/tests/driver_gpio_exp_coll/' \
    'riot/riotbuild:latest' make  'BOARD=bluepill'
Building application "tests_driver_gpio_exp_coll" for "bluepill" with MCU "stm32f1".

In file included from /data/riotbuild/riotproject/tests/driver_gpio_exp_coll/main.c:20:0:
/data/riotbuild/riotbase/core/include/assert.h:122:49: error: division by zero [-Werror=div-by-zero]
     enum { static_assert_failed_on_div_by_0 = 1 / (!!(cond)) }
                                                 ^
/data/riotbuild/riotproject/tests/driver_gpio_exp_coll/main.c:302:1: note: in expansion of macro 'static_assert'
 static_assert(COLL_RES, "[driver_gpio_exp_coll] Collision in GPIO_EXP_PIN");
 ^~~~~~~~~~~~~
/data/riotbuild/riotbase/core/include/assert.h:122:12: error: enumerator value for 'static_assert_failed_on_div_by_0' is not an integer constant
     enum { static_assert_failed_on_div_by_0 = 1 / (!!(cond)) }
            ^
/data/riotbuild/riotproject/tests/driver_gpio_exp_coll/main.c:302:1: note: in expansion of macro 'static_assert'
 static_assert(COLL_RES, "[driver_gpio_exp_coll] Collision in GPIO_EXP_PIN");
 ^~~~~~~~~~~~~
cc1: all warnings being treated as errors
/data/riotbuild/riotbase/Makefile.base:83: recipe for target '/data/riotbuild/riotproject/tests/driver_gpio_exp_coll/bin/bluepill/application_tests_driver_gpio_exp_coll/main.o' failed
make[1]: *** [/data/riotbuild/riotproject/tests/driver_gpio_exp_coll/bin/bluepill/application_tests_driver_gpio_exp_coll/main.o] Error 1
/data/riotbuild/riotbase/Makefile.include:359: recipe for target '/data/riotbuild/riotproject/tests/driver_gpio_exp_coll/bin/bluepill/application_tests_driver_gpio_exp_coll.a' failed
make: *** [/data/riotbuild/riotproject/tests/driver_gpio_exp_coll/bin/bluepill/application_tests_driver_gpio_exp_coll.a] Error 2
/home/user/Desktop/files/projects/acutam/firmware/RIOT_gpio_exp_docker/makefiles/docker.inc.mk:90: recipe for target '..in-docker-container' failed
make: *** [..in-docker-container] Error 2
```

ZetaR60:
This PR is ready for review: All targets with gpio support now have gpio_exp support. All collisions have been resolved with no modification to how gpios were previously used. Compile tests on all non-blacklisted targets show no errors, and a runtime test on mega-xplained passes.

kYc0o:
Thanks for your contribution! I think it can be reviewed soon. Can you squash so it's in a clean status?

ZetaR60:
Squashed.

ZetaR60:
Rebased on trunk and conflicts solved.

kYc0o:
I'm doing a review of this but I need you to squash the fixups. Can you do it?

ZetaR60:
Squashed.

sys/auto_init/Makefile
```diff
@@ -14,4 +14,8 @@ ifneq (,$(filter auto_init_can,$(USEMODULE)))
   DIRS += can
 endif
 
+ifneq (,$(filter auto_init_gpio_exp,$(USEMODULE)))
+  DIRS += gpio_exp
+endif
```
kYc0o: I'm not sure this belongs here. While it's necessary to add this into `DIRS`, it doesn't feel right because there's no auto_init code for this module... I'll check where it fits better.

drivers/include/gpio_exp.h
```diff
@@ -0,0 +1,320 @@
...
+        if (pin > GPIO_EXP_THRESH) { \
+            gpio_exp_t *exp_entry = gpio_exp_entry(pin); \
+            if (exp_entry == NULL) { \
+                return; \
+            } \
+            exp_entry->driver->write(exp_entry->dev, \
+                                     gpio_exp_pin(pin), value); \
+        } \
+    } \
+    while (0)
```
kYc0o: All this macro substitution should be in a static inline function, then handle the return code (if any) into the caller function. This way it's clear that this code is terminating the routine.

cpu/stm32f1/periph/gpio.c
```diff
@@ -207,6 +224,8 @@ void gpio_toggle(gpio_t pin)
 
 void gpio_write(gpio_t pin, int value)
 {
+    GPIO_INTERCEPT_WRITE(pin, value);
+
```
kYc0o: Here and in all other calls to this macro. I don't get clearly what this macro does (I know what it does because I looked at the whole code, but this snip is not self explanatory at all), especially if the following code is being executed. I'd say it needs to be surrounded by `#ifdef MODULE_GPIO_EXP` (or the name that corresponds) and add a return to explicitly exit from this function. I see that this might be tedious since there are a lot of calls, but unless someone have a better idea I'd do it that way.

sys/auto_init/Makefile
ZetaR60: This is directly analogous to ```sys/auto_init/saul```. The directory is not for use by the API itself, but rather for implementations to add their auto_init code to. My first GPIO expander PR (#9054), is currently awaiting the merge of the I2C rework, and once it is done it would have auto_init code in this folder.

drivers/include/gpio_exp.h
ZetaR60: Unfortunately it seems that this is a case where it is impossible to both be clear and be concise. To use inline functions would require:
1) an ```#ifdef``` to check if the module is enabled
2) an ```if``` statement to check if the pin needs redirecting
3) the redirection
4) a return, if required by either redirection or error

In order to have all of this logic, it would require a minimum of 5-6 lines (depending on whether the return needs to be on a separate line or not). I can add this logic, but unfortunately it would be duplicated for every macro instance. I feel that due to the duplication it is more appropriate to use a macro in this case. The macro incorporates all of the above logic.

cpu/stm32f1/periph/gpio.c
ZetaR60: I think the outline of the trade-offs in my response above also applies here.

ZetaR60:
Thanks for the review!

> I don't know if a completely new module is necessary for this but the overhead added to the GPIO interface might be too big. I'll investigate about alternatives, but I clearly get the need of a module like this.

I'm not sure what you mean here. What overhead is added? When the module is disabled there should be no additional instructions generated. When the module is enabled, but a CPU pin is requested, there is only a single numerical comparison to check if the pin is on a GPIO expander, and then it continues with the normal function call.

> My main issue is that here for the sake of code reuse we might overuse the gpio API, intended for internal CPU GPIOs.

Unfortunately, everything assumes that a "pin" is only on a CPU, and nowhere else. It is a reasonable assumption for simpler systems, however it is potentially crippling on more complex ones. The only alternative to checking in the GPIO API is rewriting every single call to the GPIO API to check for a GPIO expander and call the GPIO expander API. If the RIOT community set out to create more generic support for GPIOs, I think something similar to this PR would be required. In a generic GPIO interface, code that is calling any GPIO API should not be required to be aware that the pin in question is on a GPIO expander. That way everything maintains abstraction from the hardware's methods of communication.

kYc0o:
> I'm not sure what you mean here. What overhead is added? 

I'm talking specifically to code overhead, I didn't test yet the memory or CPU overhead, which I agree with you, it might be neglected.

Besides I think also that there's something we can reuse from this interface, but it needs to be done very carefully and with the less impact as possible for the users of this interface.

vincent-d:
I share concerns with @kYc0o regarding the 'overhead' of this PR, but more in term of maintainability for cpu code. I also understand and see the need of this feature.

I think that `periph` API are intended to remain as simple and light as possible and implement only cpu-specific code. Then I would see such a feature in a layer above `periph_gpio`, but the drawback would be to update every code using gpio... Another solution would be to switch `periph` API to a descriptor / function pointer -based API, but this topic is wider and has already been discussed a lot without reaching consensus I think.

I think @kaspar030 opinion might also be valuable here.

ZetaR60:
An alternative to the macro based intercept might be to rename all GPIO functions in ```cpu/*/periph/gpio.c``` to a low-level name (e.g. ```gpio_init``` becomes ```gpio_init_ll```), and then implement a GPIO function call as an inline like this:
```
inline int gpio_init(gpio_t pin, gpio_mode_t mode)
{
#ifdef MODULE_GPIO_EXP
    /* intercept code here */
#endif

    return gpio_init_ll(pin, mode);
}
```

This would of course involve renaming all of the GPIO implementations, but it would avoid confusion over the nature of the macro, and it would avoid the overhead of building function pointers into the GPIO interface.

kaspar030:
I'm with @kYc0o and @vincent-d, having to modify *all* cpu implementations is too much.

> Another solution would be to switch periph API to a descriptor / function pointer -based API, but this topic is wider and has already been discussed a lot without reaching consensus I think.

> An alternative to the macro based intercept might be to rename all GPIO functions in cpu/*/periph/gpio.c to a low-level name (e.g. gpio_init becomes gpio_init_ll), and then implement a GPIO function call as an inline like this:

There where even multiple prototypes with different tradeoffs.
For gpio, the current API optimizes very well (when using LTO), and we didn't (yet) come up with something that doesn't add significant overhead.

For this PR, I suggest concentrating on the actual expander API first and tackle integration later.

ZetaR60:
It was mentioned to me by @kYc0o that he performed a review of the entire PR, not just the intercept code. Can you confirm kYc0o?

There was an issue he mentioned to me with ```tests/driver_gpio_exp_coll``` default failing unknown CPUs. I can fix that now by changing unknown CPUs to default to only a partial test, which can run on any CPU.

cpu/atmega_common/periph/gpio.c
```diff
@@ -127,6 +129,8 @@ static inline int8_t _int_num(gpio_t pin)
 
 int gpio_init(gpio_t pin, gpio_mode_t mode)
 {
+    GPIO_INTERCEPT_INIT(pin, mode);
```
kaspar030: Having to add code to every single implementation needs to be avoided. Please find another way.

drivers/gpio_exp/include/gpio_exp_reg.h
```diff
@@ -0,0 +1,60 @@
...
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_gpio_exp
+ *
+ * @{
+ *
+ * @file
+ * @brief       GPIO expansion default registry
```
kaspar030: I don't quite like that gpio_exp comes with its own registry. Is this actually necessary?

ZetaR60: Yes. Expanders are addressed by a number that is used with ```GPIO_EXP_PIN(foo_dev, bar_pin)```. The number ```foo_dev``` cannot be known until linking the binary, because the board may have more than one GPIO expander, so ```foo_dev``` cannot be a macro or enum set by the driver. Also, order is important, so automatically building the registry with a linked list cannot be relied on, because changes to code may change the order of device numbers. This leaves a manually defined array as the remaining satisfactory choice, which would be set up as part of the board includes by the developer porting the board.

ZetaR60: Note: I think I was using the word "registry" too loosely here. It is not a registry in the sense of e.g. SAUL. I will change the name of the file to ```gpio_exp_conf.h```

sys/auto_init/Makefile
ZetaR60: I have removed this. Because the list of entries is not built dynamically like SAUL, drivers do not need auto_init code that is separate from the driver implementation.

drivers/gpio_exp/include/gpio_exp_reg.h
ZetaR60: Should be clearer now.

drivers/include/gpio_exp.h
ZetaR60: Removed and now uses separate functions for the low-level implementation.

drivers/include/gpio_exp.h
```diff
@@ -0,0 +1,320 @@
...
+ * @return      pin number
+ */
+gpio_t gpio_exp_pin(gpio_t gpio);
+
+#ifdef MODULE_GPIO_EXP
+
+/**
+ * @brief   Macro for intercepting and redirecting gpio_init calls
+ */
+#define GPIO_INTERCEPT_INIT(pin, mode) \
```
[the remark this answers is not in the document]
ZetaR60: Moved to non-macro in ```drivers/include/periph/gpio.h```.

cpu/atmega_common/periph/gpio.c
ZetaR60: Fixed. The intercepts are now implemented by splitting the GPIO functions into a low-level version.

ZetaR60:
The latest changes addresses some of the major concerns expressed so far:
- Use of macros
- Insertion of code into ```cpu/*/periph/gpio.c```
- Lack of clarity of API call interception

PR #9192, opened 2018-05-25
sys/uart_stdio: remove unused rx_cb from header

haukepetersen:
### Contribution description
The `uart_stdio_rx_cb()` definition seems to be some outdated leftover, as far as i can tell it is never used/implemented. So we might as well remove it.

### Issues/PRs references
none


haukepetersen:
and go.

PR #9194, opened 2018-05-25
board/cpu init: move cpu_init() from board_init() to startup

smlng:
### Contribution description

During "boot" of RIOT the initial startup (reset) function defined per CPU calls `board_init()` which then calls `cpu_init()` and maybe inits some LEDs and buttons afterwards. IMHO its a bit weird that `board_init` calls `cpu_init` instead of handling this in the startup function by calling `board_init` after `cpu_init`.

This PR moves the `cpu_init` function accordingly. I tested with Arduinos, PhyNode and Remote-RevB, which all work. However, some boards have very special startup routines so if anybody could check the following boards to verify they still work, would be great:

- [x] mbed_lpc1768
- [ ] msba2
- [ ] mulle (@gebart ?)
- [x] seeeduino_arch-pro
- [ ] waspmote-pro
- [ ] msp430 (on hold, not changed yet)
  - [ ] chronos
  - [ ] msb-430
  - [ ] telosb


kaspar030:
> +24 −661

Awesome! Something like this was long overdue.

basilfx:
I can do the Seeeduino Arch-Pro and the mbed LPC1768.

aabadie:
Nice one!

smlng:
The msp430 based are bit complicated to sort out, that's why I reverted the changes there for now.

boards/mbed_lpc1768/system.c
```diff
@@ -1,478 +0,0 @@
-/**************************************************************************//**
```
jia200x: why is this file removed?

smlng: oops, just for fun? 😄

jia200x: sounds good :)

smlng: ah no, because its basically the same file as for the seeedoino and that I moved to the cpu/lpc1768 so its shared with the mbed board, hence the remove.

jia200x: ok, cool! :)

basilfx:
mbed LPC1768 and Seeeduino Arch-Pro seem to work (tested `examples/default`).

tcschmidt:
@jia200x any changes to request? Or is the review completed?

smlng:
please note, this is still WIP (see label) need to sort out the msp430 based boards, i.e., don't merge

PR #9197, opened 2018-05-25
doc: fix quick guide command order

yegorich:
git checkout can only work if you've already changed to RIOT folder.


doc/doxygen/src/mainpage.md
```diff
@@ -55,8 +55,8 @@ call this the `native` port). Try it right now in your terminal window:
 
 ~~~~~~~{.sh}
 git clone git://github.com/RIOT-OS/RIOT.git # assumption: git is pre-installed
-git checkout <LATEST_RELEASE>
 cd RIOT
+git checkout <LATEST_RELEASE>
```
smlng: Nice catch!

PR #9198, opened 2018-05-25
cpu/sam0: update I2C driver to new API

dylad:
### Contribution description
This PR updates the I2C driver from sam0 families CPU and also update their periph_conf.h file to fit the new driver.
tested with #9193 #9195 and ADXL345

### Issues/PRs references

#6577

### DIFF
--- a/boards/sodaq-explorer/include/periph_conf.h
+++ b/boards/sodaq-explorer/include/periph_conf.h
@@ -196,34 +196,28 @@ static const spi_conf_t spi_config[] = {
  * @name I2C configuration
  * @{
  */
-#define I2C_NUMOF          (2U)
-#define I2C_0_EN            1
-#define I2C_1_EN            1
-#define I2C_2_EN            0
-#define I2C_3_EN            0
-#define I2C_IRQ_PRIO        1
+static const i2c_conf_t i2c_config[] = {
+    {
+        .dev      = &(SERCOM1->I2CM),
+        .speed    = I2C_SPEED_NORMAL,
+        .sda_pin  = GPIO_PIN(PA, 16),
+        .scl_pin  = GPIO_PIN(PA, 17),
+        .mux      = GPIO_MUX_C,
+        .gclk_src = GCLK_CLKCTRL_GEN_GCLK0,
+        .flags    = I2C_FLAG_NONE,
+    },
+    {
+        .dev      = &(SERCOM2->I2CM),
+        .speed    = I2C_SPEED_NORMAL,
+        .sda_pin  = GPIO_PIN(PA, 8),
+        .scl_pin  = GPIO_PIN(PA, 9),
+        .mux      = GPIO_MUX_C,
+        .gclk_src = GCLK_CLKCTRL_GEN_GCLK0,
+        .flags    = I2C_FLAG_NONE
+    }
+};
+#define I2C_NUMOF          (sizeof(i2c_config) / sizeof(i2c_config[0]))
 
-#define I2C_0_DEV           SERCOM1->I2CM
-#define I2C_0_IRQ           SERCOM1_IRQn
-#define I2C_0_ISR           isr_sercom1
-/* I2C 0 GCLK */
-#define I2C_0_GCLK_ID       SERCOM1_GCLK_ID_CORE
-#define I2C_0_GCLK_ID_SLOW  SERCOM1_GCLK_ID_SLOW
-/* I2C 0 pin configuration */
-#define I2C_0_SDA           GPIO_PIN(PA, 16)
-#define I2C_0_SCL           GPIO_PIN(PA, 17)
-#define I2C_0_MUX           GPIO_MUX_C
-
-#define I2C_1_DEV           SERCOM2->I2CM
-#define I2C_1_IRQ           SERCOM2_IRQn
-#define I2C_1_ISR           isr_sercom2
-/* I2C 1 GCLK */
-#define I2C_1_GCLK_ID       SERCOM2_GCLK_ID_CORE
-#define I2C_1_GCLK_ID_SLOW  SERCOM2_GCLK_ID_SLOW
-/* I2C 1 pin configuration */
-#define I2C_1_SDA           GPIO_PIN(PA, 8)
-#define I2C_1_SCL           GPIO_PIN(PA, 9)
-#define I2C_1_MUX           GPIO_MUX_C
 /** @} */
 
 /**
--- a/cpu/sam0_common/include/periph_cpu_common.h
+++ b/cpu/sam0_common/include/periph_cpu_common.h
@@ -15,6 +15,7 @@
  * @brief           Common CPU specific definitions for all SAMx21 based CPUs
  *
  * @author          Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author          Dylan Laduranty <dylan.laduranty@mesotic.com>
  */
 
 #ifndef PERIPH_CPU_COMMON_H
@@ -39,6 +40,16 @@ extern "C" {
 #define PERIPH_SPI_NEEDS_TRANSFER_BYTE
 #define PERIPH_SPI_NEEDS_TRANSFER_REG
 #define PERIPH_SPI_NEEDS_TRANSFER_REGS
+/** @} */
+
+/**
+ * @name   Use shared I2C functions
+ * @{
+ */
+#define PERIPH_I2C_NEED_READ_REG
+#define PERIPH_I2C_NEED_READ_REGS
+#define PERIPH_I2C_NEED_WRITE_REG
+#define PERIPH_I2C_NEED_WRITE_REGS
 /** @} */
 
 /**
@@ -231,6 +242,40 @@ typedef struct {
     spi_misopad_t miso_pad; /**< pad to use for MISO line */
     spi_mosipad_t mosi_pad; /**< pad to use for MOSI and CLK line */
 } spi_conf_t;
+/** @} */
+
+/**
+ * @brief   Available SERCOM I2C flag selections
+ */
+typedef enum {
+    I2C_FLAG_NONE            = 0x0,    /**< No flags set */
+    I2C_FLAG_RUN_STANDBY     = 0x1,    /**< run SERCOM in standby mode */
+} i2c_flag_t;
+
+/**
+ * @brief   Override I2C clock speed values
+ */
+#define HAVE_I2C_SPEED_T
+typedef enum {
+    I2C_SPEED_LOW       = 10000U,      /**< low speed mode:    ~10kbit/s */
+    I2C_SPEED_NORMAL    = 100000U,     /**< normal mode:       ~100kbit/s */
+    I2C_SPEED_FAST      = 400000U,     /**< fast mode:         ~400kbit/s */
+    I2C_SPEED_FAST_PLUS = 1000000U,    /**< fast plus mode:    ~1Mbit/s */
+    I2C_SPEED_HIGH      = 3400000U,    /**< high speed mode:   ~3.4Mbit/s */
+} i2c_speed_t;
+
+/**
+ * @brief   I2C device configuration
+ */
+typedef struct {
+    SercomI2cm *dev;        /**< pointer to the used I2C device */
+    i2c_speed_t speed;      /**< baudrate used for the bus */
+    gpio_t scl_pin;         /**< used SCL pin */
+    gpio_t sda_pin;         /**< used MOSI pin */
+    gpio_mux_t mux;         /**< alternate function (mux) */
+    uint8_t gclk_src;       /**< GCLK source which supplys SERCOM */
+    uint8_t flags;          /**< allow SERCOM to run in standby mode */
+} i2c_conf_t;
 
 /**
  * @brief   Set up alternate function (PMUX setting) for a PORT pin
--- a/cpu/sam0_common/periph/i2c.c
+++ b/cpu/sam0_common/periph/i2c.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2014 CLENET Baptiste
+ * Copyright (C) 2018 Mesotic SAS
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -16,11 +17,13 @@
  *
  * @author      Baptiste Clenet <bapclenet@gmail.com>
  * @author      Thomas Eichinger <thomas.eichinger@fu-berlin.de>
+ * @author      Dylan Laduranty <dylan.laduranty@mesotic.com>
  *
  * @}
  */
 
 #include <stdint.h>
+#include <errno.h>
 
 #include "cpu.h"
 #include "board.h"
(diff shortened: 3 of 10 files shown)


cpu/sam0_common/include/periph_cpu_common.h
```diff
@@ -41,6 +42,17 @@ extern "C" {
 #define PERIPH_SPI_NEEDS_TRANSFER_REGS
 /** @} */
 
+/**
+ * @brief   Use shared I2C functions
```
aabadie: Should be `@name`

cpu/sam0_common/include/periph_cpu_common.h
```diff
@@ -41,6 +42,17 @@ extern "C" {
...
+ * @{
+ */
+#define PERIPH_I2C_NEED_READ_REG
+#define PERIPH_I2C_NEED_READ_REGS
+#define PERIPH_I2C_NEED_WRITE_REG
+#define PERIPH_I2C_NEED_WRITE_REGS
+
```
aabadie: blank line could be removed

cpu/sam0_common/include/periph_cpu_common.h
```diff
@@ -231,6 +243,43 @@ typedef struct {
...
+ * @brief   Available SERCOM I2C flag selections
+ */
+typedef enum {
+    I2C_FLAG_NONE            = 0x0,    /**< No flags set */
+    I2C_FLAG_RUN_STANDBY     = 0x1,    /**< run SERCOM in standby mode */
+} i2c_flag_t;
+
+/**
+ * @brief   Override SPI clock speed values
+ * @{
```
aabadie: not needed

cpu/sam0_common/include/periph_cpu_common.h
```diff
@@ -231,6 +243,43 @@ typedef struct {
...
+ */
+#define HAVE_I2C_SPEED_T
+typedef enum {
+    I2C_SPEED_LOW       = 10000U,      /**< low speed mode:    ~10kbit/s */
+    I2C_SPEED_NORMAL    = 100000U,     /**< normal mode:       ~100kbit/s */
+    I2C_SPEED_FAST      = 400000U,     /**< fast mode:         ~400kbit/s */
+    I2C_SPEED_FAST_PLUS = 1000000U,    /**< fast plus mode:    ~1Mbit/s */
+    I2C_SPEED_HIGH      = 3400000U,    /**< high speed mode:   ~3.4Mbit/s */
+} i2c_speed_t;
+/** @} */
```
aabadie: not needed

cpu/sam0_common/include/periph_cpu_common.h
```diff
@@ -231,6 +243,43 @@ typedef struct {
...
+typedef struct {
+    SercomI2cm *dev;        /**< pointer to the used I2C device */
+    i2c_speed_t speed;      /**< baudrate used for the bus */
+    gpio_t scl_pin;         /**< used SCL pin */
+    gpio_t sda_pin;         /**< used MOSI pin */
+    gpio_mux_t mux;         /**< alternate function (mux) */
+    uint8_t gclk_src;       /**< GCLK source which supplys SERCOM */
+    uint8_t flags;          /**< allow SERCOM to run in standby mode */
+} i2c_conf_t;
+/** @} */
```
aabadie: not needed

cpu/sam0_common/periph/i2c.c
```diff
@@ -230,180 +187,79 @@ int i2c_release(i2c_t dev)
...
-        default:
-            return -1;
+    if (dev >= I2C_NUMOF) {
+        return -1;
     }
-
-    /* start transmission and send slave address */
-    if (_start(i2c, address, I2C_FLAG_READ) < 0) {
-        return 0;
+    if(!(flags & I2C_NOSTART))
```
aabadie: a space after `if` could be added

cpu/sam0_common/periph/i2c.c
```diff
@@ -230,180 +187,79 @@ int i2c_release(i2c_t dev)
...
+        /* start transmission and send slave address */
+        if (_start(bus(dev), (addr << 1) | I2C_READ) < 0) {
+            DEBUG("Start command failed\n");
+            return 0;
     }
     /* read data to register */
-    if (_read(i2c, data, length) < 0) {
+    if (_read(bus(dev), data, len) < 0) {
```
aabadie: a space after `if` could be added

cpu/sam0_common/periph/i2c.c
```diff
@@ -230,180 +187,79 @@ int i2c_release(i2c_t dev)
...
-            break;
-#endif
-        default:
-            return -1;
-    }
-
-    /* start transmission and send slave address */
-    if (_start(i2c, address, I2C_FLAG_WRITE) < 0) {
-        return 0;
+    if(!(flags & I2C_NOSTOP)) {
```
aabadie: a space after `if` could be added

cpu/sam0_common/periph/i2c.c
```diff
@@ -230,180 +187,79 @@ int i2c_release(i2c_t dev)
...
-    /* start transmission and send slave address */
-    if (_start(i2c, address, I2C_FLAG_WRITE) < 0) {
-        return 0;
+    if (dev >= I2C_NUMOF) {
+        return -1;
     }
-    /* send register address and wait for complete transfer to be finished */
-    if (_write(i2c, &reg, 1) < 0) {
-        return 0;
+    if(!(flags & I2C_NOSTART)) {
```
aabadie: a space after `if` could be added

cpu/sam0_common/periph/i2c.c
```diff
@@ -230,180 +187,79 @@ int i2c_release(i2c_t dev)
...
-    /* finish transfer */
-    _stop(i2c);
-    return length;
-}
-
-static void _i2c_poweron(SercomI2cm *sercom)
-{
-    if (sercom == NULL) {
-        return;
+    if(!(flags & I2C_NOSTOP)) {
```
aabadie: a space after `if` could be added

dylad:
@aabadie I just pushed the required changes.

boards/sodaq-explorer/include/periph_conf.h
```diff
@@ -196,34 +196,28 @@ static const spi_conf_t spi_config[] = {
...
+    {
+        .dev      = &(SERCOM1->I2CM),
+        .speed    = I2C_SPEED_NORMAL,
+        .sda_pin  = GPIO_PIN(PA, 16),
+        .scl_pin  = GPIO_PIN(PA, 17),
+        .mux      = GPIO_MUX_C,
+        .gclk_src = GCLK_CLKCTRL_GEN_GCLK0,
+        .flags    = I2C_FLAG_NONE,
+     },
+         {
```
aabadie: Minor: this opening curly brace is not aligned

cpu/sam0_common/include/periph_cpu_common.h
```diff
@@ -231,6 +242,40 @@ typedef struct {
...
+ * @brief   I2C device configuration
+ */
+typedef struct {
+    SercomI2cm *dev;        /**< pointer to the used I2C device */
+    i2c_speed_t speed;      /**< baudrate used for the bus */
+    gpio_t scl_pin;         /**< used SCL pin */
+    gpio_t sda_pin;         /**< used MOSI pin */
+    gpio_mux_t mux;         /**< alternate function (mux) */
+    uint8_t gclk_src;       /**< GCLK source which supplys SERCOM */
+    uint8_t flags;          /**< allow SERCOM to run in standby mode */
```
aabadie: The IRQ management has been removed by the refactoring. Is this intentional ? Do you think it could be set in an attribute of this struct ?

cpu/sam0_common/periph/i2c.c
```diff
@@ -45,168 +47,123 @@
...
-            break;
-#endif
-        default:
-            DEBUG("I2C FALSE VALUE\n");
-            return -1;
-    }
-
-    /* HACK: fixes cppcheck issue until #7588 is merged. */
-    if (I2CSercom == NULL) {
+    if (dev >= I2C_NUMOF) {
```
aabadie: You could use an assert instead.

```c
assert(dev < I2C_NUMOF);
```

cpu/sam0_common/periph/i2c.c
```diff
@@ -230,180 +187,79 @@ int i2c_release(i2c_t dev)
...
-
-    switch (dev) {
-#if I2C_0_EN
-        case I2C_0:
-            i2c = &I2C_0_DEV;
-            break;
-#endif
-        default:
-            return -1;
+    if (dev >= I2C_NUMOF) {
```
aabadie: same comment about the assert, there might be others

cpu/sam0_common/periph/i2c.c
```diff
@@ -230,180 +187,79 @@ int i2c_release(i2c_t dev)
...
+        return -1;
     }
-
-    /* start transmission and send slave address */
-    if (_start(i2c, address, I2C_FLAG_READ) < 0) {
-        return 0;
+    if (!(flags & I2C_NOSTART))
```
aabadie: Missing opening `{`, the if block should be surrounded with `{}` even if it's a one liner.  It improves readability.

dylad: Oops ...
Nice catch !

cpu/sam0_common/include/periph_cpu_common.h
dylad: Well it was useless for SAM0 but now you bring this to the stage... I don't know if it could be useful for others arch.
I guess we could still add something later if it's needed right ?

dylad:
@aabadie I've addressed your comments

cpu/sam0_common/include/periph_cpu_common.h
aabadie: I'm fine with keeping it like this for now.

aabadie: I'm fine with keeping it like this for now.

dylad:
Squashed

aabadie:
Travis is reporting trailing white spaces issues, can you have a look ?

MrKevinWeiss:
Here are the issues I found while testing this PR with the new periph_i2c test on a samr21.

I found that if you try to acquire twice it waits forever instead of timeout or return with busy message.  

It also appears that all read and write calls return I2C_ACK regardless of success.  A read/write reg will continue with the second action (after the initial write to reg) regardless of success.

I haven't been able to get a true successful read from my device  with this yet (but that may be my problem, I will confirm later).

It seems that the device doesn't need to be acquired to be used, I am not sure if it is intentional or not.

dylad:
> I found that if you try to acquire twice it waits forever instead of timeout or return with busy message.

This is because the if (xxxx) return -1; has been replace by an assertion. Maybe it was not a good idea ?

> It also appears that all read and write calls return I2C_ACK regardless of success. A read/write reg will continue with the second action (after the initial write to reg) regardless of success

I guess the old driver behaves the same and I still haven't find enough time to handle the error on the bus correctly. Maybe it will be done after the refactoring if I can't fix it quickly.

> It seems that the device doesn't need to be acquired to be used, I am not sure if it is intentional or not.

I think we will deal with it right now, because we must implement (in a near future ?) the acquire and release function with clock gating + pm_management. So it will be impossible to use the bus if acquire is not called.

cpu/sam0_common/periph/i2c.c
```diff
@@ -45,365 +47,208 @@
...
     return 0;
 }
 
 int i2c_acquire(i2c_t dev)
 {
-    if (dev >= I2C_NUMOF) {
-        return -1;
-    }
+    assert(dev < I2C_NUMOF);
     mutex_lock(&locks[dev]);
```
MrKevinWeiss: I like the assert, I assume once everything is ported we can either remove the int return (because all failures would be asserted) or at least update the comments in the i2c.h driver.

Should the mutex_lock be used or the mutex_trylock or should there be an option?

dylad: @aabadie what's your opinion about assert vs return err_code ?

aabadie: +1 for the assert. Brief comparing with spi API, the spi_init returns void, I think it should be the same for i2c (sorry for noticing this only now).

MrKevinWeiss:
So it appears that the _start fxn both sends the start, addr and automatically gets the first byte from that slave (feature or bug).  Disabling the smart mode doesn't seem to effect it and on the register level I can't see why it would do that.  Either way it should be fixed since read adds an extra byte (even though it isn't shown on the higher levels, the last byte seems to just remain in the dev->DATA.reg).

For i2c.c

in the fxn:
`static inline int _read(SercomI2cm *dev, uint8_t *data, int length)`
line 338
`while (count != length) {` should be `while (count != length - 1) {`

it seems like the last byte should only be read after the stop command to prevent it from reading another:

`int i2c_read_bytes(i2c_t dev, uint16_t addr, void *data, size_t len, uint8_t flags)`
line 204 add
`((uint8_t*)data[len - 1] = bus(dev)->dev->DATA.reg;`

maybe protect it with some
`assert(len > 0);`

I also don't know how this effects the flags (ie if no start then maybe len should be +1 and if no stop then reading the byte will still trigger another (unwanted) read).  I am a bit unclear on the no start functionality, should it send the addr without a start bit or no addr at all?

Also it would be nice to make the return codes match the API (maybe a separate PR).

dylad:
> I am a bit unclear on the no start functionality, should it send the addr without a start bit or no addr at all?
If NO_START is NOT selected, then we must send start bit AND the device address.
If NO_START is selected, then we must NOT send start and the devices address.
I'll try to look at it this week but I'm back to work so I don't have much time.
Thanks for the feedback !

dylad:
I'm currently working on a solution for the extra byte but I'm running out of hardware. I'll push some more commits soon !

MrKevinWeiss:
Let me know if I can help.  Also I noticed that the i2c_write_regs is not well defined.
Currently it goes 
-start-addr-reg0-stop- -start-addr-data...

if the 16 bit flag is set it goes 
-start-addr-reg0-reg1-stop -start-addr-data...

This appears to be two different write commands and shouldn't work for any sensors, the first byte of the data usually would be the register byte.
For example on the lis2dh with if I wanted to write 2,3,4 in register 5,6,7 I would go 
`i2c_write_regs(addr = 12, reg = 2, data = [5,6,7])`
I would get:
-start-12<<1-2-stop- 
which sets the pointer then 
-start-12<<1-5-6-7-stop-
this would actually write 6 and 7 in register 5 not 5 in reg 2, 6 in reg 3, and 7 in reg 4

I think it should either go:
-start-addr-reg0-data... which is the convention for writing to registers in lis2dh and many other ic's
for 16 bit 
-start-addr-reg0-reg1-data...

or
-start-addr-reg0-*start*-data...
but I have never seen this actually done.

Can anyone give me a sensor or slave device that follows either the current riot implementation or the no stop version?

MrKevinWeiss:
Well according to the [SAMR21 datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/SAM-R21_Datasheet.pdf) page 513 or 514 it appears impossible to support the no stop flag without having an extra unwanted read.  I think the closest you can come is that it sends a NACK instead of a ACK after the last byte is read if no stop bit is requested.  I would suggest for the read_bytes we just don't support that call with the stop flag.

dylad:
@MrKevinWeiss I just push a fix for the extra read byte without loosing support of NOSTART, NOSTOP
Please let me know if this solve your issue. I tested using a logic analyzer and I don't see any extra read byte now.

Now I'll cleanup some parts and work on the errno return code.

cpu/sam0_common/periph/i2c.c
```diff
@@ -471,46 +315,51 @@ static inline int _write(SercomI2cm *dev, const uint8_t *data, int length)
...
+            /* Prepare stop command before read last byte */
+            _stop(dev);
+        }
         /* Save data to buffer. */
         data[count] = dev->DATA.reg;
 
         /* Wait for response on bus. */
         if (_wait_for_response(dev, SAMD21_I2C_TIMEOUT) < 0)
-            return -1;
+            return I2C_ERR;
```
MrKevinWeiss: It seems to always fail here, maybe put a condition that it only checks if it is not the last byte.

cpu/sam0_common/periph/i2c.c
```diff
@@ -45,365 +47,207 @@
...
-    if (_write(I2CSercom, data, length) < 0) {
-        return 0;
+    ret = _write(bus(dev), data, len);
+    if (ret < 0) {
+        DEBUG("Write command failed\n");
+        return ret;
     }
-    _stop(I2CSercom);
-    return length;
+    return len;
```
MrKevinWeiss: Should this be ret not len?

cpu/sam0_common/periph/i2c.c
```diff
@@ -45,365 +47,207 @@
...
     }
-    _stop(i2c);
     /* return number of bytes sent */
-    return length;
-}
-
-int i2c_read_reg(i2c_t dev, uint8_t address, uint8_t reg, void *data)
-{
-    return i2c_read_regs(dev, address, reg, data, 1);
+    return len;
```
MrKevinWeiss: Should this be ret not len?

cpu/sam0_common/periph/i2c.c
```diff
@@ -45,365 +47,207 @@
...
-    gpio_t pin_scl = 0;
-    gpio_t pin_sda = 0;
-    gpio_mux_t mux;
-    uint32_t clock_source_speed = 0;
-    uint8_t sercom_gclk_id = 0;
-    uint8_t sercom_gclk_id_slow = 0;
+    return i2c_config[dev].dev;
+}
 
+void i2c_init(i2c_t dev)
```
MrKevinWeiss: Why no feedback if the init was successful, I imagine it would alter the program if something failed or is there an GET_INIT_STATE somewhere?

MrKevinWeiss:
@dylad I like moving the functionality to _read, much cleaner, just a few specific things that I have commented on.  This fixes the extra read problem if you have a stop bit but the problem is still there if you do not have a stop bit (I2C_NOSTOP).  This is an issue with the state machine of the hardware.  If you try to read data without the next state being a stop or nack or repeated start it will clk the slave for another byte.  The closest I have come is by disabling smart mode and withholding the ACK but it still doesn't seem correct to me.
screenshot from 2018-06-01 08-17-56

This is why I suggest we just return error for a read with NOSTOP.  The only reason to support a read NOSTOP is if we want to read a certain amount wait (with the scl line held down) then do a NOSTART read later.  I don't see this functionality being useful (this also means that we wouldn't need to support a NOSTART since we would never get into the state where it could be used).

gebart:
I had the same situation with the Kinetis hardware. I chose to let the hardware do another read when NOSTART is set and avoid performing a dummy read to clock the first byte when NOSTART is set. I don't know if it applies to the Sam hardware though

dylad:
@MrKevinWeiss I understand what you mean but I would like to keep this level of freedom for the driver. 
Of course, if we cannot find a suitable workaround to this limitation we impose this to the driver.
I'm sure we can find a solution to this :) At least, I hope...

MrKevinWeiss:
@gebart Yup seems to be the same situation.  If everyone agrees that we can assume anyone calling a read bytes with NOSTART will have previously called read bytes with NOSTOP or NOSTOP | NOSTART then it should be doable.  We could even add some static memory or certain reg check to know that it is in an inbetween state, I noticed that without a stop byte it appears that the scl stays low.  We could just say if the scl is low there is a valid previous byte in the register that we want.

@dylad Maybe I am getting too obsessy over this.  I also don't think saying the sam0_common cpu doesn't support it means that the driver doesn't support it (other cpus could have it).  I tend think if something doesn't work perfectly then it should be clearly not supported rather than hiding corner case errors (which end up costing the people using RIOT a lot of time).  With information above I will gear my tests to your decision.

dylad:
> If everyone agrees that we can assume anyone calling a read bytes with NOSTART will have previously called read bytes with NOSTOP or NOSTOP | NOSTART then it should be doable

I'm okay with that. I don't see any use case against this logic.

> We could just say if the scl is low there is a valid previous byte in the register that we want.

We can give it a try but if this issue is present in other CPU must ensure we can also use this workaround before.

> I tend think if something doesn't work perfectly then it should be clearly not supported rather than hiding corner case errors

I don't want to hide anything to the users. I really want to have a good  and reliable I2C API. I know this effort will cost us a lot of time and we will have more debate in a near futur but I'm sure we will find the perfect solution for everyone :)

gebart:
A cleaner solution for the read NOSTOP problem for the Kinetis CPUs is to switch the module to TX mode before reading the data register. Switching to TX does not send out any stop conditions or repeated starts or anything, but disables automatically triggering another reception when the data register is read. I don't know if the sam0 behaves like this, but it could be worth looking into.

dylad:
@MrKevinWeiss I'm a bit confused I reretest this PR in the current state and I don't have any issue with it (tested on SAML21-XPRO) and the ADXL345
For testing this, I replace the function
i2c_read_regs(BUS, ADDR, ACCEL_ADXL345_DATA_X0, result, 6, 0);
This function write the first register where the sensor holds its data and issue a 6 bytes read.
This has been replaced by
    i2c_write_byte(BUS, ADDR, ACCEL_ADXL345_DATA_X0, I2C_NOSTOP);
    i2c_read_bytes(BUS, ADDR, &result[0], 2, I2C_NOSTOP);
    i2c_read_bytes(BUS, ADDR, &result[2], 2, I2C_NOSTART | I2C_NOSTOP);
    i2c_read_bytes(BUS, ADDR, &result[4], 2, I2C_NOSTART);
Which is the equivalent.

And the output on Logic analyzer is OK

the previous frame : 


The new frame obtain : 



Datas changes of course but the form is good .START/STOP ACK/NACK are OK to me.
Did I miss something ??

MrKevinWeiss:
@dylad I think it is fine for now, sorry for being too nitpicky.  I was testing on a single call level with expected behavior, not as the full frame.  So if you just call i2c_read_bytes(BUS, ADDR, &result[0], 2, I2C_NOSTOP); then it would have 3 bytes read on the line.

Since you agree that we can make the assumptions the users of the api will not do silly things (like call two NOSTOP reads in a row and always end a NOSTOP read with a NOSTART read) then it should be alright.

It just the weird corner cases that I was thinking of.  Anyways, very nice job, sorry I was such a pain.  Looking forward to the merge (than your feedback on the periph_i2c).

keestux:
May I chime in (after a long time :-) ? I was looking at this I2C rework and I was trying the BMX280 driver on my SODAQ Autonomo (SAMD21). I'm using the branch, and rebased the bmx280 rework on top of it.

Result: i2c_read_reg is giving I2C_ERR (-3), but it did read the correct byte (BME id).

I verified that master works as expected.

keestux:
Debug output of i2c_read_regs when running tests/driver/_bmx280:

    Stop sent
    Wait for response.
    STATUS_ERR_TIMEOUT
    Read command failed

keestux:
@dylad What's the status of i2c_if_sam0? I notice that the SAM0 I2C functions still return len instead of a status.

keestux:
@dylad Sorry to nag, but the adxl345 driver is a dubious choice to test the I2C rework. Nowhere it checks the return value of the I2C functions.

dylad:
@MrKevinWeiss Don't get it wrong. There is absolutely no need to apologize, I really appreciate your efforts to get the best driver possible. In fact, we were not testing the same thing so I'll also look at a single call of read_bytes without stop. 
@keestux I'll investigate your issue but I don't have this sensor. Currently, I'm waiting for #9274 to get merge to rebase this branch on it. We're having issue with Murdock. 
I can fix the ret code soon if that helps you testing your sensor.

dylad:
> the adxl345 driver is a dubious choice to test the I2C rework

You're right but this is the only sensor I have atm.

dylad:
@keestux Do you have the same error when you disable the debug output ?

keestux:
@dylad Yes, I have the error without debug too. I only enabled debug to investigate the error.

dylad:
@keestux I don't know what's going on then. I did not change a lot of things in the driver and I cannot use logic analyzer without the device...
I'll try to take a look at the datasheet to see if I can find something.

keestux:
@dylad There is an essential change in the driver how the stop command is done.
Before it was a separate step in i2c_read_bytes or i2c_write_bytes. Now it is (optionally) part of _read. In the new situation it does _stop before reading the last byte and wait for response.

I'm not at all an I2C expert, so I cannot tell if your new approach is correct. All I can do is look at the result, and it's not working right now. Plus the return values are in my way.

If I do a quick hack (moving the _stop call to the end of _read) then I see it coming back to live. But that makes me wondering why it is working for you. Maybe it isn't, because you only have ADXL345. What you could do is to change ADXL345 and check the return values.

dylad:
The  new approach is borrowed from Atmel I2C driver in Atmel Studio. The STOP CMD is set before read the last byte in Atmel's driver. 
I'll retry with ADXL345 maybe the error is set too.

MrKevinWeiss:
I'm a bit busy now but I can look in depth tomorrow.  I will say at one point I noticed that this if (_wait_for_response(dev, SAMD21_I2C_TIMEOUT) < 0)  shouldn't be called if the stop bit was set.  Maybe you fixed that already or maybe that is not true but just didn't want anyone spinning their wheels unnecessarily...

dylad:
I'll look tonight and test. I'll also take another look to Atmel's driver.

dylad:
Indeed I'm facing the same error return. I compare with Atmel driver, it did not "wait for response" after reading the last byte of the frame as @MrKevinWeiss said. I guess it's logical because we have already all the bytes we want so there is no need to wait for anything else from the sensor.
I'll push a fix.
I'll also change the tests/periph_adxl345 later to include return code check.
Thanks for your help :)

keestux:
Meanwhile I also looked at ASF 3.34 sam0/drivers/sercom/i2c/i2c_samd20/i2c_master.c (lacking access to Atmel Studio, not for Linux)
If that is what you used for inspiration then I see many deviations. One of them is how STOP and ACK/NACK is handled. If we really want ASF as an inspiration source then we should try harder to use it.

keestux:
@dylad Your i2c_if_sam0 branch is in an inconsistent state. I'm getting build errors because of i2c_init

     void i2c_init(i2c_t dev)
versus

    int i2c_init(i2c_t dev);

dylad:
@keestux this should be fixed.
Sorry but I had to rebase the branch to the current new_i2c_if branch.

MrKevinWeiss:
@dylad So I tested the newest version, aside from fixing the init return and an write reg implementation in the driver/periph_common/i2c.c it seems like it is passing the basic tests.  It is not fully tested (still working on it) but this test should catch most problems.  You can view the results [here](https://gist.github.com/MrKevinWeiss/074d8b435cb4564d975903d816842120).  I would like to see this merged but I guess it depends on @aabadie

dylad:
Are we good now with the current state ? If there is any more changes or bugs please tell me so I can try to fix it asap. We need to move forward.

aabadie:
@dylad, I tested this PR on arduino-zero with #9298 and #9299, it works like a charm. I'm fine with merging this. Please squash (I think 2 commits will be ok) !
What do you think @MrKevinWeiss ?

MrKevinWeiss:
ACK, I have been waiting for this day my entire life!

keestux:
@dylad I need a little bit more time to test it...

cpu/sam0_common/periph/i2c.c
```diff
@@ -321,15 +323,17 @@ static inline int _write(SercomI2cm *dev, const uint8_t *data, int length)
...
             DEBUG("STATUS_ERR_PACKET_COLLISION\n");
             return -EAGAIN;
         }
 
         /* Wait for hardware module to sync */
         while (dev->SYNCBUSY.reg & SERCOM_I2CM_SYNCBUSY_MASK) {}
 
-        DEBUG("Written byte #%i to data reg, now waiting for DR to be empty again\n", buffer_counter);
+        DEBUG("Written byte #%i to data reg, now waiting for DR \
+              to be empty again\n", buffer_counter);
```
keestux: This is not good. The newline is escaped alright, but then there are 14 spaces instead of one.

cpu/sam0_common/periph/i2c.c
```diff
@@ -281,7 +283,7 @@ static int _start(SercomI2cm *dev, uint16_t addr)
 
     /* Wait for response on bus. */
     if (addr & I2C_READ) {
-        /* Some devices (e.g. SHT2x) can hold the bus while preparing the reply */
+        /* Some devices can hold the bus while preparing the reply */
```
keestux: The example of SHT2x was added by me. If we leave it out there could be a moment in the future when somebody wants to know why.

cpu/sam0_common/periph/i2c.c
keestux: BTW. The way to solve it is like this

           DEBUG("Written byte #%i to data reg, now waiting for DR"
                 " to be empty again\n", buffer_counter);

cpu/sam0_common/periph/i2c.c
dylad: I'll change that.

cpu/sam0_common/periph/i2c.c
dylad: Will fix

dylad:
@keestux I'm really really sorry I squashed the branch before seeing your comments...
I'll fix all your pending comments and wait for your ACK before re-squash.
Again I'm sorry...

aabadie:
> I need a little bit more time to test it...

no worry @keestux, if we merge this branch, it will be in the new_i2c_if feature branch. You'll still be able to test this feature branch and provide fixes if you find problems.

When all is adapted (cpu + drivers) and tested, we'll open a PR of the feature branch to master. We hope this will be done in the next 2 weeks.

dylad:
@aabadie We can wait for @keestux's test. I'm fine with that. I'll prepare the nRF52 during this time.

cpu/sam0_common/periph/i2c.c
```diff
@@ -45,365 +48,227 @@
...
-            return -1;
-    }
-
-    /* HACK: fixes cppcheck issue until #7588 is merged. */
-    if (I2CSercom == NULL) {
-        return -1;
-    }
-
+    assert(dev < I2C_NUMOF);
+    /* Initiliaze mutex */
```
aabadie: There's a typo on 'Initiliaze', should be 'Initialize'

dylad: Oops

keestux:
@dylad I'm not happy yet. My BME280 driver does not work (gives -EAGAIN). And when I enable debug the error goes away. This smells like a timing problem. And this is just a simple sensor. What other drivers were tested?

Oh, and don't worry about the resquashing

aabadie:
> What other drivers were tested?

I tested a hts221 (temperature/humidity) and lsm6dsl (accelero/gyro) from [ST mems extension](http://www.st.com/en/ecosystems/x-nucleo-iks01a2.html) directly plugged on a genuino-zero.
I modified the branch of this PR by merging #9298 and #9299 in it.
They were working well but I don't have a logic analyzer to track what is exchanged on the I2C bus.

dylad:
@keestux I've got one idea. I'll push a potential fix in a few minutes.

keestux:
@dylad Hmm, this is making it worse. Last debug lines are:

    Wait for device to be ready
    Generate start condition by sending address
    Wait for response.
    Looping through bytes
    Written byte #0 to data reg, now waiting for DR to be empty again

 Anyway, for me, time is up for today.

dylad:
@keestux the lastest introduced bug is solved by the last commit.
I think this should also solve your previous issue with your bme280 device.
Let me know if it's OK for you.

aabadie:
I tested the latest version of this branch on arduino-zero with bmp280 (using a fixed version of #9206) and it works.

keestux:
@dylad Great. It's working for me (bme280).
I'll continue to investigate, but it is good that you got this far. Thanks.

cpu/sam0_common/periph/i2c.c
```diff
@@ -248,8 +248,8 @@ int i2c_write_bytes(i2c_t dev, uint16_t addr, const void *data, size_t len,
         return ret;
     }
     /* Ensure all bytes has been sent */
-    while ((bus(dev)->STATUS.reg & SERCOM_I2CM_STATUS_BUSSTATE_Msk)
-           != BUSSTATE_IDLE) {}
+   /* while ((bus(dev)->STATUS.reg & SERCOM_I2CM_STATUS_BUSSTATE_Msk)
+           != BUSSTATE_IDLE) {}*/
```
keestux: This should be cleaned up.

cpu/sam0_common/periph/i2c.c
```diff
@@ -363,6 +355,10 @@ static inline int _write(SercomI2cm *dev, const uint8_t *data, int length,
             return -EIO;
         }
     }
+    if (stop) {
+        /* Prepare stop command before write last byte */
```
keestux: Is the comment correct? The last byte was already written.
Also, I get the feeling we should do something similar in _read. I mean, do the _stop after the while loop (if it is requested).

The comment probably belongs to the ACK/NACK, which you "prepare" it before doing the last read byte.

cpu/sam0_common/periph/i2c.c
dylad: Ooh a wild comment !
Will fix

dylad:
@keestux I've updated the comments

keestux:
Let's go for it
Do I need to ACK it somewhere? If not, here is my ACK.

dylad:
Squashed !

aabadie:
So let's go then!

PR #9202, opened 2018-05-25
cpu/stm32: unify i2c drivers and apply new API

aabadie:
### Contribution description

This PR unifies the I2C peripheral driver implementation for STM32 F0/F3/F7/L0/L4 families, since they are very similar and adapt the related board configurations.
This PR also changes the way i2c peripheral configurations are provided by the boards. Now each board must define a nice array of structs, like for other peripherals.

For F0, F7 and L4, I added a default configuration for the nucleo-f070rb, nucleo-f722ze, and b-l475e-iot01a boards.

I tested it on b-l072z-lrwan1 and b-l475e-iot01a with a hts221 sensor and it works. I couldn't really test the F3 driver, since the driver seems broken, even in master. I couldn't find a solution for this. Same for F0.

For F3, I also had to update the f303xe cmsis file, and also found an issue in the periph_common init. Will open a PR for this.

### Issues/PRs references

#6577


boards/b-l072z-lrwan1/include/periph_conf.h
```diff
@@ -159,6 +159,7 @@ static const spi_conf_t spi_config[] = {
 #define I2C_NUMOF           I2C_0_EN
 #define I2C_IRQ_PRIO        1
 #define I2C_APBCLK          (CLOCK_APB1)
+#define I2C_SPEED           I2C_SPEED_NORMAL
```
vincent-d: I think it would be great to take advantage of the refactoring to get rid of this "old-style" define way and use an array of struct as it's done for other periph.

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -323,6 +333,18 @@ typedef struct {
...
+ * @brief   I2C configuration data structure
+ */
+typedef struct {
+    I2C_TypeDef *dev;       /**< i2c device */
+    gpio_t scl;             /**< scl pin number */
+    gpio_t sda;             /**< sda pin number */
+    gpio_mode_t pin_mode;   /**< with or without pull resistor */
+    gpio_af_t af;           /**< I2C alternate function value */
+    uint8_t ev_irqn;        /**< event IRQ */
+} i2c_conf_t;
```
vincent-d: Good to have this, but it seems unused, isn't it?

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -323,6 +333,18 @@ typedef struct {
...
 
+/**
+ * @brief   I2C configuration data structure
+ */
+typedef struct {
+    I2C_TypeDef *dev;       /**< i2c device */
+    gpio_t scl;             /**< scl pin number */
+    gpio_t sda;             /**< sda pin number */
+    gpio_mode_t pin_mode;   /**< with or without pull resistor */
+    gpio_af_t af;           /**< I2C alternate function value */
```
vincent-d: Don't you need an af per pin (e.g. one for sda, and one for scl)?

aabadie:
Current status of this PR:
- L0: works, tested on b-l072z-lrwan
- L4: works, tested on b-l475e-iot01a
- F0 and F3 are broken. For both, the application hangs when setting up the peripheral for sending start condition. I couldn't find a reason why this happens and gdb didn't help either.

aabadie:
> F0 and F3 are broken

Just found the problem ! Now both are working, will push a fix. In fact, the I2C clock source has to be turned on in one RCC register (CFGR3).

aabadie:
This PR is ready for review again. Actual status: tested nucleo-f070/f303/l073/f722, b-l072z-lrwan1 and b-l475e-iot01a boards with success.
So now I2C works for all STM32 F0/F3/F7/L0/L4 cpu families.

aabadie:
There's a problem with Travis: this PR conflicts with current master (because of the DMA changes in the STM32 periph_common header file).
What should we do in this case ? Rebase `new_i2c_if` on top of master (with or without --preserve option) ? @miri64 did you have the same issue with the netif2 feature branch ? (how was this solved ?) Sorry, lots of questions.

vincent-d:
Quickly tested with nucleo-f091 (see https://github.com/aabadie/RIOT/pull/7). It seems OK so far.

cpu/stm32_common/periph/i2c.c
```diff
@@ -0,0 +1,463 @@
...
+ *               2017 we-sens.com
+ *               2018 Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     cpu_stm32l0
```
vincent-d: `cpu_stm32_common`

cpu/stm32_common/periph/i2c.c
```diff
@@ -0,0 +1,463 @@
...
+
+#define TICK_TIMEOUT    (0xFFFF)
+
+#if defined(CPU_FAM_STM32F0) || defined(CPU_FAM_STM32F3) || \
+    defined(CPU_FAM_STM32F7) || defined(CPU_FAM_STM32L0) || \
+    defined(CPU_FAM_STM32L4)
+
+#define I2C_IRQ_PRIO    (1)
+
+#define CLEAR_FLAG(dev)            dev->ICR |= I2C_ICR_NACKCF   | \
```
vincent-d: we usually avoid macro, I would suggest to use a constant to only define the value, 
i.e. `#define CLEAR_FLAG (I2C_ICR_NACKCF | I2C_ICR_ARLOCF | I2C_ICR_BERRCF)`
and then use it `dev->ISR |= CLEAR_FLAG`

cpu/stm32_common/periph/i2c.c
```diff
@@ -0,0 +1,463 @@
...
+    DEBUG("[i2c] init: initializing device\n");
+    mutex_init(&locks[dev]);
+
+    I2C_TypeDef *i2c = i2c_config[dev].dev;
+
+    periph_clk_en(i2c_config[dev].bus, i2c_config[dev].rcc_mask);
+
+#if defined(CPU_FAM_STM32F0) || defined(CPU_FAM_STM32F3)
+    /* Clear I2CSW bit */
+    RCC->CFGR3 &= ~(i2c_config[dev].rcc_sw_mask);
```
vincent-d: I think this is not the expected behavior, you should clear with `RCC_CFGR3_I2C1SW`

cpu/stm32_common/periph/i2c.c
```diff
@@ -0,0 +1,463 @@
...
+    /* Set I2CSW bits according */
+    RCC->CFGR3 |= i2c_config[dev].rcc_sw_mask;
+#endif
+
+    NVIC_SetPriority(i2c_config[dev].irqn, I2C_IRQ_PRIO);
+    NVIC_EnableIRQ(i2c_config[dev].irqn);
+
+    DEBUG("[i2c] init: configuring pins\n");
+    /* configure pins */
+    gpio_init(i2c_config[dev].scl_pin, GPIO_OD_PU);
```
vincent-d: Not sure if `GPIO_OD_PU` is the best here, same for sda. Why not using `GPIO_OUT`?

cpu/stm32_common/periph/i2c.c
```diff
@@ -0,0 +1,463 @@
...
+        length = 0;
+    }
+
+    /* end transmission */
+    _stop(i2c);
+
+    return length;
+}
+
+void i2c_poweron(i2c_t dev)
```
vincent-d: `i2c_poweron` and `i2c_poweroff` are not in the API anymore

cpu/stm32_common/periph/i2c.c
```diff
@@ -0,0 +1,463 @@
...
+    assert(dev < I2C_NUMOF);
+
+    uint16_t tick = TICK_TIMEOUT;
+
+    while ((i2c_config[dev].dev->ISR & I2C_ISR_BUSY) && tick--) {}
+
+    periph_clk_dis(i2c_config[dev].bus, i2c_config[dev].rcc_mask);
+}
+
+static inline void _start(I2C_TypeDef *i2c, uint8_t address,
```
vincent-d: Even though 10-bit address is not yet implemented, I think it would be better to use `uint16_t` for `address`. Maybe pass the flags to be able to enable this feature later on.

cpu/stm32_common/include/cpu_conf_stm32_common.h
```diff
@@ -0,0 +1,142 @@
...
+extern "C" {
+#endif
+
+#if defined(CPU_FAM_STM32F0) || defined(CPU_FAM_STM32F3) || \
+    defined(CPU_FAM_STM32F7) || defined(CPU_FAM_STM32L0) || \
+    defined(CPU_FAM_STM32L4)
+
+static const i2c_timing_param_t timing_params[] = {
+#if defined(CPU_FAM_STM32F0)
+    [ I2C_SPEED_NORMAL ]    = {
```
vincent-d: These timings are valid for an input clock wt 48MHz (same for L4 and F7, btw).

The clock requirements should be stated clearly somewhere (alongside the `i2c_conf_t` definition I guess).

cpu/stm32_common/periph/i2c.c
```diff
@@ -0,0 +1,463 @@
...
+    /* configure pins */
+    gpio_init(i2c_config[dev].scl_pin, GPIO_OD_PU);
+    gpio_init_af(i2c_config[dev].scl_pin, i2c_config[dev].scl_af);
+    gpio_init(i2c_config[dev].sda_pin, GPIO_OD_PU);
+    gpio_init_af(i2c_config[dev].sda_pin, i2c_config[dev].sda_af);
+
+    DEBUG("[i2c] init: configuring device\n");
+    /* set the timing register value from predefined values */
+    i2c_timing_param_t tp = timing_params[i2c_config[dev].speed];
+    uint32_t timing = (( (uint32_t)tp.presc << 28) |
```
vincent-d: You might use the constants from the vendor header for the shift values

aabadie:
Thanks for reviewing @vincent-d, I'll update the PR asap

cpu/stm32_common/periph/i2c.c
aabadie: This is what is done. But the macro used for setting the clock after is not correct. I'll update this but that will required another field in the configuration struct.

vincent-d: It works only when `rcc_sw_mask` is 1. This line will write 0, and the next one 1. If `rcc_sw_mask` is 0, it will leave the field as-is (so if this was the reset value it works, but then this line is useless).

aabadie:
@vincent-d, I think I addressed your comments. Regarding the pin configuration, even not a good reason, it kept what was set for f3 and l0. Looking at the reference manual didn't help.

cpu/stm32_common/include/cpu_conf_stm32_common.h
```diff
@@ -0,0 +1,139 @@
...
+        .scldel = 0xB,      /* t_SCLDEL = 500ns  */
+    },
+    [ I2C_SPEED_FAST_PLUS ] = {
+        .presc  = 0,
+        .scll   = 0x6,      /* t_SCLL   = 875ns */
+        .sclh   = 0x3,      /* t_SCLH   = 500ns */
+        .sdadel = 0x0,      /* t_SDADEL = 0ns   */
+        .scldel = 0x1,      /* t_SCLDEL = 250ns */
+    }
+#elif defined(CPU_FAM_STM32L4) || defined(CPU_FAM_STM32F7)
```
vincent-d: this could be factored in the `CPU_FAM_STM32F0` block

vincent-d: I'm also wondering if we should not group by frequency instead of by family. But this could be done later on if we think it makes more sense.

cpu/stm32_common/periph/i2c.c
```diff
@@ -0,0 +1,453 @@
...
+            break;
+        }
+    }
+
+    /* send STOP condition */
+    DEBUG("[i2c] stop: Generate stop condition\n");
+    i2c->CR2 |= I2C_CR2_STOP;
+}
+
+static inline void irq_handler(i2c_t dev)
```
vincent-d: It seems no interrupt is enable, is this function really needed or is this more for a future use?

vincent-d:
> Regarding the pin configuration, even not a good reason, it kept what was set for f3 and l0.

I'd say let's start with that and fix later if this is an issue. My only concern is when an external pull-up is already there.

cpu/stm32_common/periph/i2c.c
aabadie: One interrupt is enabled here: https://github.com/RIOT-OS/RIOT/pull/9202/files#diff-8703624eb1944c61de7e7320bf02f6e3R80

On certain families (F3, L4, F7) there are 2 per I2C (ev/er), I only took the `er` for the moment.

cpu/stm32_common/periph/i2c.c
gebart: Another method which will clear all w1c flags is
```c
dev->ISR = dev->ISR;
```

aabadie:
@vincent-d, comments addressed!

aabadie:
@vincent-d, just in case, I started to adapt the F4 within this PR. Will push an update soon.
The idea is to have one C file for each type and to build the right based on the type of CPU. It's way cleaner than merging both types in the same C (with an hell of ifdefs).

vincent-d:
I think this is the best approach, indeed.

I'd like to help, but I'll be out next week. I can still ask my colleagues if they can help to test and review.

aabadie:
I'm done with unification of STM32 I2C drivers ! F1 still needs to be tested but all others are working (I'm testing with an hts221 device).
When the API is finalized (return codes, etc), I'll update the driver internal implementation.

aabadie:
> F1 still needs to be tested

Done and works on iotlab-m3.

aabadie:
I adapted this PR with the new error codes scheme. Tested with #9298 and #9299 on nucleo-l073 and nucleo-f411 (2 types of ST drivers).

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -148,6 +148,16 @@ typedef uint32_t gpio_t;
...
 #define SPI_HWCS(x)         (SPI_HWCS_MASK | x)
 
+/**
+ * @name    Use the shared I2C functions
+ * @{
+ */
+/** Use read regs function from periph common */
+#define PERIPH_I2C_NEED_READ_REG
+/** Use write regs function from periph common */
+#define PERIPH_I2C_NEED_WRITE_REG
```
dylad: You're not using `PERIPH_I2C_NEED_READ_REGS` and `PERIPH_I2C_NEED_WRITE_REGS` ?

cpu/stm32_common/periph/i2c_1.c
```diff
@@ -0,0 +1,497 @@
...
+}
+
+int i2c_read_regs(i2c_t dev, uint16_t address, uint16_t reg, void *data,
+                  size_t length, uint8_t flags)
+{
+    assert(dev < I2C_NUMOF);
+
+    DEBUG("[i2c] read_regs: addr: %04X, reg: %04X\n", address, reg);
+
+    if (flags & I2C_REG16) {
```
dylad: add I2C_ADDR10 as an unsupported feature please. We still cannot test this feature without a specific sensor to test it.

cpu/stm32_common/include/periph_cpu_common.h
aabadie: because it's a special case of read/write_regs. It's not needed in this driver. Maybe we could also get rid of read/write_regs and use the periph_common implementation but I wasn't able the find a suitable solution.

dylad: This can be done later after the I2C refactoring then.

cpu/stm32_common/periph/i2c_1.c
aabadie: Agreed, fixed.

dylad:
@aabadie I cannot compile with your branch and BOARD=stm32f4discovery.
I guess there are some missing board periph_conf implementations .

aabadie:
> I guess there are some missing board periph_conf implementations .

Indeed, I forgot the stm32f3discovery and stm32f4discovery. This is done now but I had to update another F303 CMSIS file.

dylad:
I tested with a STM32F4discovery board and ADXL345 accel and this does not work.
When I tried to enable debug in i2c_2.c file, there is zero output on the serial port. But when debug is disabled on the driver I have the following : 
```
2018-06-11 09:50:27,871 - INFO # �main(): This is RIOT! (Version: 2018.07-devel-296-gcf6cf-dylan-K53SD-new_i2c_stm32l0_f3)
2018-06-11 09:50:27,873 - INFO # ADXL345 test application
2018-06-11 09:50:27,893 - INFO # Initializing ADXL345 accelerometer at I2C_DEV(0)... ret is :-116
2018-06-11 09:50:27,895 - INFO # [adxl345] init - error: invalid id value [0x12]
2018-06-11 09:50:27,895 - INFO # [Failed]
```
I've modified a bit the adxl345 to have some more output since the I2C driver don't want to talk to me.
Does it work for you with this board ? (these boards of mine were saved from the trash bin but I don't rely on them too much...)

aabadie:
> When I tried to enable debug in i2c_2.c file, there is zero output on the serial port

I already had this and it's a more general problem with uart initialization and that ends up in a system crash/blocked. I fixed this by calling `uart_stdio_init()` from `periph_common/init.c` and commenting it out in `sys/newlib_syscalls_default/syscalls.c`. There might be better solutions for this but it works.

> Does it work for you with this board ?

I can try. I have the hardware for that ;)

dylad:
> I already had this and it's a more general problem with uart initialization and that ends up in a system crash/blocked

Oh right @gebart already said something about that.

I'll re-try with some debug outputs then !

dylad:
@aabadie I still have an issue with my board : 
```
main(): This is RIOT! (Version: 2018.07-devel-297-g95bf1-dylan-K53SD-new_i2c_stm32l0_f3)
ADXL345 test application
Initializing ADXL345 accelerometer at I2C_DEV(0)... [i2c] wait_ready: Wait for device to be ready
[i2c] read_regs: Send slave address and clear ADDR flag
[Failed]
```

aabadie:
> I still have an issue with my board

I setup this board with the sensor and could reproduce. It seems the device is not responding to the start condition.
I don't have a logic analyzer (yet), can you have a look with yours ?

dylad:
> I setup this board with the sensor and could reproduce

Good news ! I'll not throw these boards back to the trash bin !
I'll test it tonight, my logic analyzer is at home.

aabadie:
@dylad, if unsure of your board, you can test with master, it works for me. So there's a regression introduced by the driver unification (for f1/f2/f4/l1).

aabadie:
@dylad: it works now, there a missing clk attribute in the board configuration... (a good way to lose time)

dylad:
@aabadie I confirm I2C now works on STM32F4DISCOVERY !
@vincent-d Do you think this PR is good to go ?

aabadie:
> Do you think this PR is good to go ?

I'd like to see if I can improve the error codes returned first. But if the @vincent-d is ok with the current state, I'm fine :)

cpu/stm32_common/periph/i2c_1.c
```diff
@@ -0,0 +1,497 @@
...
+int i2c_read_bytes(i2c_t dev, uint16_t address, void *data,
+                   size_t length, uint8_t flags)
+{
+    assert(dev < I2C_NUMOF);
+
+    I2C_TypeDef *i2c = i2c_config[dev].dev;
+
+    int ret = 0;
+
+    if (!(flags & I2C_NOSTART)) {
```
dylad: Please add unsupported flags here also since this function is usable for user

cpu/stm32_common/periph/i2c_1.c
```diff
@@ -0,0 +1,497 @@
...
+int i2c_write_bytes(i2c_t dev, uint16_t address, const void *data,
+                    size_t length, uint8_t flags)
+{
+    assert(dev < I2C_NUMOF);
+
+    I2C_TypeDef *i2c = i2c_config[dev].dev;
+
+    int ret = 0;
+
+    if (!(flags & I2C_NOSTART)) {
```
dylad: same here, add unsupported flags

cpu/stm32_common/periph/i2c_1.c
```diff
@@ -0,0 +1,497 @@
...
+    i2c->TXDR = reg;
+
+    /* write out data bytes */
+    ret = _write(i2c, data, length);
+    if (ret < 0) {
+        return ret;
+    }
+
+    /* end transmission */
+    ret = _stop(i2c);
```
dylad: No support for I2C_NOSTOP here ?

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -0,0 +1,535 @@
...
+
+    size_t n = length;
+    char *in = (char *)data;
+
+    I2C_TypeDef *i2c = i2c_config[dev].dev;
+
+    assert(i2c != NULL);
+
+    int ret = 0;
+
```
dylad: Please add unsupported flags here

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -0,0 +1,535 @@
...
+    I2C_TypeDef *i2c = i2c_config[dev].dev;
+    assert(i2c != NULL);
+
+    int ret = _wait_ready(i2c);
+    if (ret != 0) {
+        return ret;
+    }
+
+    if (!(flags & I2C_NOSTART)) {
```
dylad: also add unsupported flags

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /* write data to register */
+    ret = _write(i2c, data, length);
+    if (ret < 0) {
+        return ret;
+    }
+
+    /* finish transfer */
+    ret = _stop(i2c);
```
dylad: add I2C_NOSTOP here

dylad:
Also, Travis is reporting some weird errors

aabadie:
> Travis is reporting some weird errors

There's another conflict on top of master with new_i2c_if + this PR. I can have a look.

aabadie:
The rebase issue is fixed, some commits were not at the right place...

cpu/stm32_common/periph/i2c_1.c
dylad: You're right but it may have some use case with those weird I2C devices in the wild.

aabadie:
Comments addressed @dylad

dylad:
@aabadie Travis reported an error with cppcheck.
Squashed directly so we can merge this one.

aabadie:
@dylad, finally fixed.

dylad:
Great job @aabadie 
Here we go !

PR #9207, opened 2018-05-27
pkg: u8g2: adapt to new I2C api

basilfx:
### Contribution description

This PR updates the U8G2 package. Tested this one on the STK3600 (EFM32, adapted).

### Issues/PRs references

#6577

PR #9206, opened 2018-05-27
drivers: bmx280: adapt to new I2C api.

basilfx:
### Contribution description

This PR updates the BMx280 driver. Tested this one on the SLTB001A (EFM32, adapted).

It's probably better to wait for #8383, but it was easy to adapt for the time being.

### Issues/PRs references

#6577

PR #9205, opened 2018-05-27
drivers: si70xx: adapt to new I2C api.

basilfx:
### Contribution description

This PR updates the Si70xx drivers. Tested this one on the SLTB001A (EFM32, adapted).

### Issues/PRs references

#6577

PR #8383, opened 2018-01-19
drivers/bmx280: reworked driver and added SPI mode

haukepetersen:
### Issues/PRs references
For supporting  the `ruuvitag` it is needed to interface the `bmx280` sensor in SPI mode. While adding this mode, I found the driver was in a very bad state. This PR adds the possibility to use the driver in SPI mode and it significantly improves the drivers structure and style:

 bugs fixed (I wonder how these were able to go by previous reviews...):
- move global variables into device descriptor
- guard bus access (use acquire and release)

added functionality:
- enable SPI mode

structural improvements:
- reduce stack usage
- simplify the driver's structure
- centralize bus access code
- use assertions
- cleanup includes
- use shortcuts for bus access

style changes:
- fix line length
- cleanup and improve doxygen
- unify pointer notation (char *var over char* var)
- unify (error) return messages
- use `#ifdef MODULE_BME280` instead of `#if defined(BME..)`
- unify debug messages -> using `[bmx28] x: msg` scheme

This PR also adapts and improves the test application accordingly and also add the `bme280_spi` module as default SAUL module to the `ruuvitag`.

Note on commits: as it was impossible to cleanly split the rework and cleanup efforts from the SPI mode addition to the driver, I put all changes into a single commit...

### Testing:
The changes are tested and verified for the `bmx280_spi` and `bmp280_spi` on the `ruuvitag`. I don't have access to any board having these sensors connected via I2C, so the I2C mode still needs to be tested.

PR #9208, opened 2018-05-27
cpu: efm32: adapt to new I2C interface

basilfx:
### Contribution description

This PR updates the EFM32 port to work with the new I2C peripheral interface. I've tested this on the SLTB001A (which has Si7021, BMP280 + I2C power controller), SLSTK3401a (Si7021) and STK3600 (for #9207).

I don't have 10-bit I2C device, nor have I access to something to test 16 bit registers.

### Issues/PRs references

#6577, depends on ~#9205~ and ~#9206~ (or ~#8383~).

### DIFF
--- a/cpu/efm32/periph/i2c.c
+++ b/cpu/efm32/periph/i2c.c
@@ -18,6 +18,8 @@
  * @}
  */
 
+#include <errno.h>
+
 #include "cpu.h"
 #include "mutex.h"
 
@@ -25,30 +27,33 @@
 #include "periph/i2c.h"
 #include "periph/gpio.h"
 
-/* emlib uses the same flags, undefine fist */
-#undef I2C_FLAG_WRITE
-#undef I2C_FLAG_READ
-
 #include "em_cmu.h"
 #include "em_i2c.h"
 
+/**
+ * @brief   Large-enough value to have some timeout value for rogue I2C
+ *          transfers. Value based on kit driver (shipped with Simplicity
+ *          Studio).
+ */
+#define I2C_TIMEOUT (300000)
+
+/**
+ * @brief   Holds the I2C transfer progress.
+ */
 static volatile I2C_TransferReturn_TypeDef i2c_progress[I2C_NUMOF];
 
 /**
  * @brief   Initialized bus locks (we have a maximum of three devices)
  */
-static mutex_t i2c_lock[] = {
-    MUTEX_INIT,
-    MUTEX_INIT,
-    MUTEX_INIT
-};
+static mutex_t i2c_lock[I2C_NUMOF];
 
 /**
  * @brief   Start and track an I2C transfer.
  */
-static void _transfer(i2c_t dev, I2C_TransferSeq_TypeDef *transfer)
+static int _transfer(i2c_t dev, I2C_TransferSeq_TypeDef *transfer)
 {
     bool busy = true;
+    uint32_t timeout = I2C_TIMEOUT;
 
     /* start the i2c transaction */
     i2c_progress[dev] = I2C_TransferInit(i2c_config[dev].dev, transfer);
@@ -57,7 +62,7 @@ static void _transfer(i2c_t dev, I2C_TransferSeq_TypeDef *transfer)
     while (busy) {
         unsigned int cpsr = irq_disable();
 
-        if (i2c_progress[dev] == i2cTransferInProgress) {
+        if (i2c_progress[dev] == i2cTransferInProgress && timeout--) {
             cortexm_sleep_until_event();
         }
         else {
@@ -66,17 +71,34 @@ static void _transfer(i2c_t dev, I2C_TransferSeq_TypeDef *transfer)
 
         irq_restore(cpsr);
     }
+
+    /* check for timeout */
+    if (!timeout) {
+        return -ETIMEDOUT;
+    }
+
+    /* transfer finished, interpret the result */
+    switch (i2c_progress[dev]) {
+        case i2cTransferDone:
+            return 0;
+        case i2cTransferUsageFault:
+            return -EINVAL;
+        case i2cTransferAddrNack:
+            return -ENXIO;
+        case i2cTransferArbLost:
+            return -EAGAIN;
+        default:
+            return -EIO;
+    }
 }
 
-int i2c_init_master(i2c_t dev, i2c_speed_t speed)
+void i2c_init(i2c_t dev)
 {
-    /* assert number of locks */
-    assert(I2C_NUMOF <= (sizeof(i2c_lock) / sizeof(i2c_lock[0])));
-
     /* check if device is valid */
-    if (dev >= I2C_NUMOF) {
-        return -1;
-    }
+    assert(dev < I2C_NUMOF);
+
+    /* initialize lock */
+    mutex_init(&i2c_lock[dev]);
 
     /* enable clocks */
     CMU_ClockEnable(cmuClock_HFPER, true);
@@ -96,7 +118,7 @@ int i2c_init_master(i2c_t dev, i2c_speed_t speed)
     I2C_Init_TypeDef init = I2C_INIT_DEFAULT;
 
     init.enable = false;
-    init.freq = (uint32_t) speed;
+    init.freq = (uint32_t) i2c_config[dev].speed;
 
     I2C_Reset(i2c_config[dev].dev);
     I2C_Init(i2c_config[dev].dev, &init);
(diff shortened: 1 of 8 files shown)


cpu/efm32/periph/i2c.c
```diff
@@ -122,128 +126,102 @@ int i2c_init_master(i2c_t dev, i2c_speed_t speed)
...
-    if (i2c_progress[dev] != i2cTransferDone) {
-        return -2;
-    }
-
-    return length;
+    return _transfer(dev, &transfer);
 }
 
-int i2c_read_reg(i2c_t dev, uint8_t address, uint8_t reg, void *data)
+int i2c_read_reg(i2c_t dev, uint16_t address, uint16_t reg, void *data, uint8_t flags)
```
aabadie: `i2c_read_reg` and `i2c_write_reg` are now provided by the common driver in `drivers/periph_common`. Just define `PERIPH_I2C_NEED_READ_REG` and `PERIPH_I2C_NEED_WRITE_REG` in the `periph_cpu.h` file of your CPU type.
For example in #9202, I added the following in `cpu/stm32_common/periph_cpu_common.h`:

```c
/**
 * @name    Use the shared I2C functions
 * @{
 */
/** Use read regs function from periph common */
#define PERIPH_I2C_NEED_READ_REG
/** Use write regs function from periph common */
#define PERIPH_I2C_NEED_WRITE_REG
/** @} */
```

Other defines exist for `PERIPH_I2C_NEED_READ_REGS` and `PERIPH_I2C_NEED_WRITE_REGS` if your CPU driver implementation works with them.

cpu/efm32/periph/i2c.c
```diff
@@ -122,128 +126,102 @@ int i2c_init_master(i2c_t dev, i2c_speed_t speed)
...
-    if (i2c_progress[dev] != i2cTransferDone) {
-        return -2;
-    }
-
-    return length;
+    return _transfer(dev, &transfer);
 }
 
-int i2c_write_byte(i2c_t dev, uint8_t address, uint8_t data)
+int i2c_write_byte(i2c_t dev, uint16_t address, uint8_t data, uint8_t flags)
```
aabadie: You don't need this function, it's already provided by the common i2c driver (in drivers/periph_common/)

cpu/efm32/periph/i2c.c
```diff
@@ -122,128 +126,102 @@ int i2c_init_master(i2c_t dev, i2c_speed_t speed)
...
+    CMU_ClockEnable(i2c_config[dev].cmu, false);
+
+    /* release lock */
     mutex_unlock(&i2c_lock[dev]);
 
     return 0;
 }
 
-int i2c_read_byte(i2c_t dev, uint8_t address, void *data)
+int i2c_read_byte(i2c_t dev, uint16_t address, void *data, uint8_t flags)
```
aabadie: You also don't need this function

pkg/gecko_sdk/Makefile
```diff
@@ -1,6 +1,6 @@
 PKG_NAME=gecko_sdk
 PKG_URL=https://github.com/basilfx/RIOT-gecko-sdk
-PKG_VERSION=d381e526d68a2d0c951f37040c1c2e168ac66cd6
+PKG_VERSION=a5e7be18ef3d7947ab6f04f9cd74f8808da7aece
```
aabadie: Why do you need to update ?

basilfx: It includes an update for the I2C driver.

cpu/efm32/periph/i2c.c
basilfx: @aabadie I don't want to use the common implementations, because I then have to use the optional I2C flags, which I cannot use with `emlib`. Emlib has its own way to provide `I2C_START` and `I2C_STOP`.

cpu/efm32/periph/i2c.c
```diff
@@ -66,18 +58,30 @@ static void _transfer(i2c_t dev, I2C_TransferSeq_TypeDef *transfer)
 
         irq_restore(cpsr);
     }
+
+    /* transfer finished, interpret the result */
+    switch (i2c_progress[dev]) {
+        case i2cTransferDone:
+            return I2C_ACK;
```
dylad: We were wondering last week if it's better to return I2C_ACK when everything is fine or the number of bytes written on the bus.
What do you think ?

cpu/efm32/periph/i2c.c
aabadie: I see. Why not converting RIOT common flags to emlib one (using a switch) ? This way you can get rid of a these functions.

cpu/efm32/periph/i2c.c
basilfx: I don't think it really matters. `I2C_ACK` would imply all bytes are transferred (either direction), otherwise you would receive a different error (e.g. bus arbitration lost).

cpu/efm32/periph/i2c.c
basilfx: I can see if that works, but it would require me to briefly patch [`em_i2c.c`](https://github.com/basilfx/RIOT-gecko-sdk/blob/master/dist/emlib/src/em_i2c.c) (which I don't mind, since Gecko-SDK is already prepared for RIOT-OS).

cpu/efm32/periph/i2c.c
dylad: I agree but we need to get a consensus here so all CPUs will behave in the same way :)

dylad:
@basilfx Any progress on this ?

dylad:
ping @basilfx

basilfx:
@dylad Sorry, I haven't found the time to fix this yet.

dylad:
@basilfx no problem, do you think you can fix it before the new release ?

cpu/efm32/periph/i2c.c
```diff
@@ -66,18 +58,30 @@ static void _transfer(i2c_t dev, I2C_TransferSeq_TypeDef *transfer)
...
+            return I2C_ADDR_NACK;
+        case i2cTransferNack:
+            return I2C_DATA_NACK;
+        default:
+            return I2C_ERR;
+    }
 }
 
-int i2c_init_master(i2c_t dev, i2c_speed_t speed)
+int i2c_init(i2c_t dev)
```
aabadie: Should be `void i2c_init(i2c_t dev)`, according to the new i2c api

cpu/efm32/periph/i2c.c
```diff
@@ -66,18 +58,30 @@ static void _transfer(i2c_t dev, I2C_TransferSeq_TypeDef *transfer)
...
 {
-    /* assert number of locks */
-    assert(I2C_NUMOF <= (sizeof(i2c_lock) / sizeof(i2c_lock[0])));
-
     /* check if device is valid */
     if (dev >= I2C_NUMOF) {
```
aabadie: There's a consensus on using `assert`:
```c
assert(dev < I2C_NUMOF);
```

dylad:
I talked with @basilfx on IRC. He'll try to adapt the EFM32 this week-end if he find some time. 
We're almost done !

basilfx:
I have fixed all the comments. Sorry for the delay, as I have been very busy with other things. I also thought it would be more work, but I opted for not supporting `I2C_NOSTART`/`I2C_NOSTOP` instead.

This version won't compile, because it depends on current `master`, which updates `pkg/gecko-sdk` for a change I need. Once `new_i2c_if` gets rebased, it should work, in the mean time, simply change the hash in `pkg/gecko-sdk/Makefile` to `a5e7be18ef3d7947ab6f04f9cd74f8808da7aece`.

basilfx:
@smlng Please see [my comment](https://github.com/RIOT-OS/RIOT/pull/9208#issuecomment-403592258) :-)

smlng:
oops ... but yes with the new commit its works.

dylad:
@aabadie are you happy with the changes ?

smlng:
well then please squash to make travis happy

basilfx:
Travis is Happy!

smlng:
I trust @basilfx to have this tested well enough to get it merged, @dylad any objections?

smlng:
and ping @aabadie, let's get over with this.

dylad:
we can merge this one as soon as @aabadie dismiss his review if he think his comments were addressed !

dylad:
Here we go ! thanks @basilfx !

PR #9211, opened 2018-05-28
xtimer: timer & target overflow, hang resolved.

Josar:
This Problems encountered when testing for https://github.com/RIOT-OS/RIOT/pull/8990

1. The xtimer hung when 32 bit target overflowed, as the timer was not placed in the right list.
2. The xtimer hung when the hardware timer overflowed, as the comparison was wrong.

grafik

Tested with xtimer_hang on jiminy board.

### DIFF
--- a/sys/include/xtimer/implementation.h
+++ b/sys/include/xtimer/implementation.h
@@ -65,7 +65,21 @@ static inline uint32_t _xtimer_lltimer_mask(uint32_t val)
  * @brief xtimer internal stuff
  * @internal
  */
+
 uint64_t _xtimer_now64(void);
+
+/**
+ * @brief Sets the timer to the appropriate timer_list or list_head.
+ *
+ * @note    The target to set the timer to has to be at least bigger then the
+ *          ticks needed to jump into the function and calculate '_xtimer_now()'.
+ *          So that 'now' did not pass the target.
+ *          This is crucial when using low CPU frequencies and/or when the
+ *          '_xtimer_now()' call needs multiple xtimer ticks to evaluate.
+ *
+ * @param[in] timer   pointer to xtimer_t which is added to the list.
+ * @param[in] target  Absolute target value in ticks.
+ */
 int _xtimer_set_absolute(xtimer_t *timer, uint32_t target);
 void _xtimer_set(xtimer_t *timer, uint32_t offset);
 void _xtimer_set64(xtimer_t *timer, uint32_t offset, uint32_t long_offset);
--- a/sys/xtimer/xtimer_core.c
+++ b/sys/xtimer/xtimer_core.c
@@ -1,6 +1,7 @@
 /**
  * Copyright (C) 2015 Kaspar Schleiser <kaspar@schleiser.de>
- * Copyright (C) 2016 Eistec AB
+ *               2016 Eistec AB
+ *               2018 Josua Arndt
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -15,6 +16,7 @@
  * @brief xtimer core functionality
  * @author Kaspar Schleiser <kaspar@schleiser.de>
  * @author Joakim Nohlgård <joakim.nohlgard@eistec.se>
+ * @author Josua Arndt <jarndt@ias.rwth-aachen.de>
  * @}
  */
 
@@ -178,12 +180,27 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
     uint32_t now = _xtimer_now();
     int res = 0;
 
-    DEBUG("timer_set_absolute(): now=%" PRIu32 " target=%" PRIu32 "\n", now, target);
-
     timer->next = NULL;
-    if ((target >= now) && ((target - XTIMER_BACKOFF) < now)) {
+
+    /* Ensure that offset is bigger than 'XTIMER_BACKOFF',
+     * 'target - now' will allways be the offset no matter if target < or > now.
+     *
+     * This expects that target was not set too close to now and overrun now, so
+     * from setting target up until the call of '_xtimer_now()' above now has not
+     * become equal or bigger than target.
+     * This is crucial when using low CPU frequencies so reaching the '_xtimer_now()'
+     * call needs multiple xtimer ticks.
+     *
+     * '_xtimer_set()' and `_xtimer_periodic_wakeup()` ensure this by already
+     * backing off for small values. */
+    uint32_t offset = (target - now);
+
+    DEBUG("timer_set_absolute(): now=%" PRIu32 " target=%" PRIu32 " offset=%" PRIu32 "\n",
+          now, target, offset);
+
+    if (offset <= XTIMER_BACKOFF) {
         /* backoff */
-        xtimer_spin_until(target + XTIMER_BACKOFF);
+        xtimer_spin_until(target);
         _shoot(timer);
         return 0;
     }
@@ -195,6 +212,17 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
 
     timer->target = target;
     timer->long_target = _long_cnt;
+
+    /* Ensure timer is fired in right timer period.
+     * Backoff condition above ensures that 'target - XTIMER_OVERHEAD` is later
+     * than 'now', also for values when now will overflow and the value of target
+     * is smaller then now.
+     * If `target < XTIMER_OVERHEAD` the new target will be at the end of this
+     * 32bit period, as `target - XTIMER_OVERHEAD` is a big number instead of a
+     * small at the beginning of the next period. */
+    target = target - XTIMER_OVERHEAD;
+
+    /* 32 bit target overflow, target is in next 32bit period */
     if (target < now) {
         timer->long_target++;
     }
@@ -214,7 +242,7 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
 
             if (timer_list_head == timer) {
                 DEBUG("timer_set_absolute(): timer is new list head. updating lltimer.\n");
-                _lltimer_set(target - XTIMER_OVERHEAD);
+                _lltimer_set(target);
             }
         }
     }
@@ -493,10 +521,14 @@ overflow:
      * time to overflow.  In that case we advance to
      * next timer period and check again for expired
      * timers.*/
-    if (reference > _xtimer_lltimer_now()) {
+    /* check if the end of this period is very soon */
+    uint32_t now = _xtimer_lltimer_now() + XTIMER_ISR_BACKOFF;
+    if (now < reference) {
         DEBUG("_timer_callback: overflowed while executing callbacks. %i\n",
               timer_list_head != NULL);
         _next_period();
+        /* wait till overflow */
+        while( reference < _xtimer_lltimer_now()){}
         reference = 0;
         goto overflow;
     }
@@ -506,7 +538,7 @@ overflow:
         next_target = timer_list_head->target - XTIMER_OVERHEAD;
 
         /* make sure we're not setting a time in the past */
-        if (next_target < (_xtimer_lltimer_now() + XTIMER_ISR_BACKOFF)) {
+        if (next_target < (_xtimer_now() + XTIMER_ISR_BACKOFF)) {
             goto overflow;
         }
     }


kYc0o:
Looks legit, I'll run the tests on hardware. @kaspar030 can you take a look? It seems a straightforward change.

sys/xtimer/xtimer_core.c
```diff
@@ -190,6 +190,11 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
 
     timer->target = target;
     timer->long_target = _long_cnt;
+
+    /* Ensure timer is fired in right hardware timer periode. */
```
kaspar030: s/periode/period/

kYc0o:
@Josar can you provide your test example? The current xtimer_hang test succeeds both in master and this PR, though it seems more stable with your patch. Tested on arduino-mega2560.

kYc0o:
Much more stable with this PR in a z1 (msp430).

kaspar030:
@Josar Maybe make both commit message and PR title summarize that is changed instead of the intend of the change? This is not the first resolved hang in xtimer, and it won't be the last. ;)

kaspar030:
@gebart would you take a brief look if we're missing something obviuos?

kYc0o:
Murdock just agreed, however for some reason in my setup I can't compile xtimer_hang test (both in current master and this PR):

```console
make: Entering directory '/Users/facosta/git/RIOT-OS/RIOT/tests/xtimer_hang'
Building application "tests_xtimer_hang" for "samr21-xpro" with MCU "samd21".

In file included from /Users/facosta/git/RIOT-OS/RIOT/tests/xtimer_hang/main.c:28:0:
/Users/facosta/git/RIOT-OS/RIOT/sys/include/xtimer.h:34:10: fatal error: msg.h: No such file or directory
 #include "msg.h"
          ^~~~~~~
compilation terminated.
```

Josar:
The xtimer hang is obviously designed to find errors like this but in the Python Skript than allows a deviation of 10% hiding errors like that again. 

I tried to introduce probing in https://github.com/RIOT-OS/RIOT/pull/9025 and https://github.com/RIOT-OS/RIOT/pull/8865 so this errors could be be probed in a HIL setup.

Also the error can be visually seen in the output stream delay, but the test neglect this. 

I use the xtimer_hang test and added probing pins. As I saw the print delay.

kYc0o:
Well now the tests passed.

cladmi:
I successfully run the `xtimer_hang` test on
* iotlab-m3
* iotlab-a8-m3
* samr21-xpro
* arduino-zero
* wsn430-v1_4

I will re-run all automated tests on them.

Josar:
@cladmi how are the boards running with https://github.com/RIOT-OS/RIOT/pull/9037 ?
Before and after this patch?

kYc0o:
I think we're waiting for @gebart ACK. @cladmi has tested it and everything went fine.

Josar:
@gebart could you find some time to review this?

sys/xtimer/xtimer_core.c
```diff
@@ -190,11 +194,16 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
 
     timer->target = target;
     timer->long_target = _long_cnt;
+
+    /* Ensure timer is fired in right hardware timer period. */
+    target = target - XTIMER_OVERHEAD;
```
gebart: What happens when target is small (`target < XTIMER_OVERHEAD`)?

Josar: Then the target underflows, it will be bigger than now and no timer overflow will happen, no long target increase needed.
Unless now+overhead > target, which should be seldome. But, this leads me to a condition which schould have been adressed above when checking for backoff.
something like this:
```
    /* first comparison prevents overflowed target to trigger backoff, target in next period.
     * second comparison backoff condition.
     * Third condition check for overflowed target
     * Fourth condition backoff condition. */
    if ( (target >= now) && ((target - XTIMER_BACKOFF) < now) ||
         (target < now) && (target <= (now + XTIMER_BACKOFF))) {

...
    /* Ensure timer is fired in right hardware timer period.
     * Backoff condition check above ensures that taget - XTIMER_OVERHEAD > now
     * also for values when now will overflow soon and target is overflowed */
    target = target - XTIMER_OVERHEAD;
```

I will have a deeper look into this tomorrow and also make some tests.

Josar: Testing for `taget - XTIMER_OVERHEAD > now` is allready made in `_xtimer_set()` or `_xtimer_periodic_wakeup` by testing that `offset < XTIMER_BACKOFF`.

If the target for `_xtimer_set_absolute(...)` is set to close to `now` and `now` overleaps target this leads to a whole 32bit cicle delay. But this has to be ensure by the calling functions, which it is.

I added some coments to the code.

Josar:
Changed the Backoff condition as it did not address backing off when the target was small and now at the end of a period.

Josar:
@cladmi maybe this requires a second test on hardware?

Josar:
After rebase this new pkg seems to fail. Anyone an idea?
```
    failed:
    tests/pkg_monocypher/samr21-xpro
```

kYc0o:
Please rebase again, now that #9486 is merged.

Josar:
@kYc0o done.

kYc0o:
Tests passed again. @gebart and @kaspar030, any further comments?

cladmi:
I will re-run tests.

cladmi:
I re-ran tests and they all work on:

* iotlab-m3
* iotlab-a8-m3
* samr21-xpro
* arduino-zero (cortex cpu)
* wsn430-v1_4

However, they all work on master, after the first commit without the last commit fix and with the last commit.
So from testing, I can only see that it does not break these platforms, not that it fixes something.

Josar:
@cladmi can you trigger a error with https://github.com/RIOT-OS/RIOT/pull/9037 on master and the boards?

Josar:
I tried to make some tests clearly show the problem but could not get a satisfiying one working.
I think this can only be seen when using probes with an osciloscope or digitsal analyser as shown in the pictures above.

Josar:
@cladmi can you see a difference from mater to this with this test https://github.com/RIOT-OS/RIOT/pull/9596 ?

Josar:
@kaspar030 @gebart are ther any further objections?

Is this planned to be merged?

kYc0o:
As far as I understand, this PR would allow to reduce our tolerance of 10% on some xtimer tests, isn't it? If so, I'd say it's important to merge it.

I'll try to test on platforms with 16bit timers.

Ping @kaspar030 and @gebart.

Josar:
Any news?

kYc0o:
Can you split the uncrustify part? So we can merge that and leave only the fix here.

Josar:
https://github.com/RIOT-OS/RIOT/pull/9953

But could we nevertheless somehow continue here?

Josar:
Rebased and squashed to remove the uncrustify split.

kYc0o:
Thanks @Josar ! Maybe @A-Paul can help us here with the performance measurements?

kYc0o:
@A-Paul I'd say a comparison between current master and this PR with the oscilloscope, as @Josar did. 

@Josar can you add a bit more information on how did you test it?

Josar:
I added pin Toggles to The Timer Interrupt, the xTimer, the main and the worker threads 2/3. And then i compared the measured sequence with what i would expect. 
The timer has to apear before the xtimer and after that there comes one or more of the threads.
The main should apear in equal distance, IIRC after ten times of the threads.
There should be no gaps, hiccups or times when nothing happens, There should be a redundant pattern which at best does not change. I think the worker thread interrupts will adjust to be fired in the same timer interrupt after some time, but this is a matter of to much work for the cpu, imho.

A-Paul:
Hi @Josar,

> I used xtimer_hang and added pin Toggles to The Timer Interrupt, the xTimer, the main and the worker threads 2/3. And then i compared the measured sequence with what i would expect.

Are you using the GPIO APi for this? And, I can find those code changes in your Repo?

Josar:
@A-Paul the atmega timer debug pin schould be available. The others not and also not in my repo. The gpio API is quite slow for pin toggles, imho, so i used low level pin toggles.

A-Paul:
When I offered my assistance I thought that there is already a test routine that I can run and just report the results. I also don't know which platforms are required to test.

I'm working on my thesis and don't have enough time to dive into all the necessary hardware related GPIO stuff.

gebart:
This is one of the problems with xtimer; it is difficult to test because it is too integrated, too many facets to test at once

Josar:
@A-Paul i included the xtimer_hang pins and adapted the makefile.

What we wold expect is that the main pin is on for ca. 100ms and that in a regular interval
Also we would expect the worker to be after 1ms and 1.1ms. (As this falls in the XTIMER_ISR_BACKOFF the first timer is delayed untill the second is ready. Thus the time interval is longer.)

The second figure also shows cicled in red when there is a hardware interrupt before the second worker thread timer is set. This leads to a separation of the timer interrupts, As we only count 7 until they are in the same hardware interrupt again we lost one worker time 2, which is expected as it has a lower priority then worker timer 1. And thus in the moment when the hardware interrupt for the 2 worker is executet it has to wait for the 1 worker timer because of the XTIMER_ISR_BACKOFF and ten the first timer is executed first as it has the higher priority.

The third figure shows a hiccup in the main thread, which is an error, bug, problem you name it.

Github has some problems with uploading, figure will follow.

sys/xtimer/xtimer_core.c
```diff
@@ -178,12 +180,27 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
...
+    /* Ensure that offset is bigger than 'XTIMER_BACKOFF',
+     * 'target - now' will allways be the offset no matter if target < or > now.
+     *
+     * This expects that target was not set to close to now and overrun now, so
+     * from setting target up until the call of '_xtimer_now()' above now has not
+     * become equal and than bigger target.
+     * This is crucial when using low CPU frequencies so reaching the '_xtimer_now()'
+     * call needs multiple xtimer ticks.
+     *
+     * '_xtimer_set()' and `_xtimer_periodic_wakeup()` ensure this by allready
```
MrKevinWeiss: Spelling mistake allready->already

sys/xtimer/xtimer_core.c
```diff
@@ -178,12 +180,27 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
     uint32_t now = _xtimer_now();
```
jcarrano: I know it's not you who wrote this, but would you mind documenting what this function is supposed to do?

Without knowing the expected behaviour it is hard to see the fix.

sys/xtimer/xtimer_core.c
```diff
@@ -178,12 +180,27 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
...
 
-    DEBUG("timer_set_absolute(): now=%" PRIu32 " target=%" PRIu32 "\n", now, target);
-
     timer->next = NULL;
-    if ((target >= now) && ((target - XTIMER_BACKOFF) < now)) {
+
+    /* Ensure that offset is bigger than 'XTIMER_BACKOFF',
+     * 'target - now' will allways be the offset no matter if target < or > now.
+     *
+     * This expects that target was not set to close to now and overrun now, so
```
jcarrano: If there is a requirement in the input parameters, it should be stated in the docs for this function.

Josar:
@jcarrano 

> what was wrong

The timer where placed in the wrong list when they where set near the overflow. (line 201 - 247)

The next target (32 bit) was compared to a low lovel timer (unknown bit size) (line 536)

> What's the point of trying to correct for interrupt latency (e.g. XTIMER_OVERHEAD) when the tests allow such a big deviation?

IMHO, the test allows such deviation as the underlying problem was not found.
But this fix has nothing to do with correcting interrupt latency, as this was allready implemented.

sys/xtimer/xtimer_core.c
Josar: If you meant for `_xtimer_set_absolute` done.

sys/xtimer/xtimer_core.c
Josar: See above.

sys/include/xtimer/implementation.h
```diff
@@ -65,7 +65,21 @@ static inline uint32_t _xtimer_lltimer_mask(uint32_t val)
  * @brief xtimer internal stuff
  * @internal
  */
+
 uint64_t _xtimer_now64(void);
+
+/**
+ * @brief Sets the timer to the appropriate timer_list or list_head.
+ *
+ * @note    The target to set the timer to has to be at least bigger then then
```
MichelRottleuthner: (...) then the~n~

sys/include/xtimer/implementation.h
```diff
@@ -65,7 +65,21 @@ static inline uint32_t _xtimer_lltimer_mask(uint32_t val)
...
+ *          ticks needed to jump into the function and calculate '_xtimer_now()'.
+ *          So that 'now' did not pass the target.
+ *          This is crucial when using low CPU frequencies and/or when the
+ *          '_xtimer_now()' call needs multiple xtimer ticks to evaluate.
+ *
+ * @param[in] timer   pointer to xtimer_t which is added to the list.
+ * @param[in] target  Absolut target value in ticks.
```
MichelRottleuthner: Absolut**e**

sys/xtimer/xtimer_core.c
MichelRottleuthner: not to**o** close to

sys/xtimer/xtimer_core.c
```diff
@@ -178,12 +180,27 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
...
+     * from setting target up until the call of '_xtimer_now()' above now has not
+     * become equal and than bigger target.
```
MichelRottleuthner: equal **or bigger than** target?

sys/xtimer/xtimer_core.c
```diff
@@ -178,12 +180,27 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
...
+     * This is crucial when using low CPU frequencies so reaching the '_xtimer_now()'
+     * call needs multiple xtimer ticks.
+     *
+     * '_xtimer_set()' and `_xtimer_periodic_wakeup()` ensure this by already
+     * backing of for small values. */
```
MichelRottleuthner: backing of**f**

sys/xtimer/xtimer_core.c
```diff
@@ -195,6 +212,16 @@ int _xtimer_set_absolute(xtimer_t *timer, uint32_t target)
 
     timer->target = target;
     timer->long_target = _long_cnt;
+
+    /* Ensure timer is fired in right timer period.
+     * Backoff condition above ensures that 'target - XTIMER_OVERHEAD` is later
+     * than 'now', also for values when now will overflow and the value of target
+     * is smaller then now.
+     * If `target < XTIMER_OVERHEAD` the new target will be at the end of this
+     * 32bit period, instead at the beginning of the next.*/
```
MichelRottleuthner: new target will **not** be at the end of this 32bit period but instead at the beginning... ?

Josar:
@MichelRottleuthner 

>  I noticed that in some cases one of the workers isn't executed until a full wrap-around of the timer happens. This doesn't happen very often, once after the first ~2 seconds and then again after ~80 seconds.

As i could narrow it down this is caused by setting the timer to the long_list. 
https://github.com/RIOT-OS/RIOT/blob/67f4c19274f1bc52f96981448c2a69e9cd8990a3/sys/xtimer/xtimer_core.c#L232

cladmi:
@Josar I am trying to provide an unit-test with mock for this to show the issue (implementation dependent but still a start).

I have xtimer using my own mock functions of `timer` while still allowing to use timers in the low levels.
I currently copied `xtimer_hang` test as a base to just validate it's executing.

https://github.com/cladmi/RIOT/pull/new/pr/wip/write_test_xtimer_overflow_and_hang


Are you able to describe a minimal test that I could do to show this ?
With values that `timer_read` would return and would cause issues.

I can fake values of `timer_read` and execute the timer_callback with a delay.


I will also try to extract something from your description at the same time and do it myself but you may be faster than me.

cladmi:
Hmm my branch was based on your changes, I will revert your commits to be able to show an error.

sys/include/xtimer/implementation.h
```diff
@@ -65,7 +65,21 @@ static inline uint32_t _xtimer_lltimer_mask(uint32_t val)
...
  */
+
 uint64_t _xtimer_now64(void);
+
+/**
+ * @brief Sets the timer to the appropriate timer_list or list_head.
+ *
+ * @note    The target to set the timer to has to be at least bigger then the
+ *          ticks needed to jump into the function and calculate '_xtimer_now()'.
+ *          So that 'now' did not pass the target.
```
cladmi: That is hard to guarantee I think. The problem is that the calling task may be de-scheduled for a few ticks and then you miss it anyway.

As it is a private function, is it not possible to update the API to pass it the `now` and `offset` value to handle it ?

jcarrano: I really don't get this requirement. How can it be that for correct operation one needs to know the time it gets to get to a certain line? IMO this is very fragile.

cladmi: To not pass `now` and `offset`, another solution could be to just assume (and ensure) that `offset < UINT32_MAX - MAX_MARGIN` where `MAX_MARGIN` is the maximum time we allow to enter the function and read `now` again.

This way, you are sure that a timer loop can be detected if you took less than `MAX_MARGIN` to go to `now`.

A large enough solution could be to say it fits in an `int32_t` so uses `31bits`.
It is an easy thing to ensure over the code base.
With 1Mhz it would still allow working if you take 35 minutes between the first call and being in the function.

Josar: This is no problem to guarantee. `XTIMER_BACKOFF` has to be set to as much ticks as it needs to reach the backing-off case. 

I allready started an test which helps to extract the required values.
See: https://github.com/RIOT-OS/RIOT/pull/8990/files

This requirement is based on the fact that the time is ticking while you try to set a timer. If you set your deadline too short you will miss it. Thats it.

Josar:
@cladmi i think a low level timer mock is a good idea.

I think a test just needs to set the counter register of the lltimer to the end of the period before setting the xtimers. And also set the 32 bit low counter of the xtimer to the end of a period before setting the xtimer. 

Basically testing is needed for
1. First timer is at end of lltimer period and 2 timer is in next lltimer period.
2. First timer is at the end of the 32 bit period and 2 timer is at the beginning of the next 32 bit period.
3. A lltimer overflow happens while timer 1 is allready processed and waiting for timer 2 in the interrupt routine. But after line 496 ` if (reference > _xtimer_lltimer_now()) {`

sys/include/xtimer/implementation.h
cladmi: I will look at #8990 as a base.


When you start calling `_xtimer_set_absolute` you are still in your context, and you could be de-scheduled, for some time (maybe more than `XTIMER_BACKOFF`), so you do not know that the `target` is not elapsed, or not too close from your calling function.
Of course if you call the function with something that ends up being in the past, it should _shoot it directly, but you cannot do better as it was asked too late.

All the calling functions of `_xtimer_set_absolute` already have a `now` value (2 functions + 1 test).
By getting `now` and `offset` you remove the issues with "calling offset was too short" issue and handle this in the function directly.


However I see another issue also with the current implementation, it is that when `now` is read in the function, and `offset` is evaluated it is not with masked interrupts. So everything could also go wrong there too, be de-scheduled and get stuck with `xtimer_spin_until`.


The only hard limitation we cannot do anything about, is that between the first `now` and the `now` in the function, there should not be more than `UINT32_MAX / Frequency` ticks.

I wanted to say 2min23 for 32Mhz timers, but it looks like there are only sub MHz timers.
Even if there is this comment about cc2538 using a 16Mhz timer and code to handle them.

https://github.com/RIOT-OS/RIOT/blob/77dc923d7e96d8333a442d11e0e3e7d6563d3c12/sys/include/xtimer/tick_conversion.h#L41

``` bash
git grep XTIMER_HZ  . ':!sys'
boards/common/arduino-atmega/include/board_common.h:#define XTIMER_HZ                   (250000UL)
boards/frdm-kw41z/include/board.h:#define XTIMER_HZ                   (32768ul)
boards/hifive1/include/periph_conf.h:#define XTIMER_HZ                   (32768ul)
boards/ikea-tradfri/include/board.h:#define XTIMER_HZ           (250000UL)
boards/jiminy-mega256rfr2/include/board.h:#define XTIMER_HZ           (125000UL)
boards/mega-xplained/include/board.h:#define XTIMER_HZ                   (125000UL)
boards/mulle/include/board.h:#define XTIMER_HZ                   (32768ul)
boards/slstk3401a/include/board.h:#define XTIMER_HZ           (250000UL)
boards/slstk3402a/include/board.h:#define XTIMER_HZ           (250000UL)
boards/sltb001a/include/board.h:#define XTIMER_HZ           (250000UL)
boards/slwstk6000b/include/board.h:#define XTIMER_HZ           (250000UL)
boards/stk3600/include/board.h:#define XTIMER_HZ           (250000UL)
boards/stk3700/include/board.h:#define XTIMER_HZ           (250000UL)
boards/waspmote-pro/include/board.h:#define XTIMER_HZ                   (62500UL)
cpu/esp32/periph/timer.c:    CHECK_PARAM_RET (freq == XTIMER_HZ_BASE, -1);
cpu/esp32/periph/timer.c:    CHECK_PARAM_RET (freq == XTIMER_HZ_BASE, -1);
cpu/esp8266/periph/timer.c:    CHECK_PARAM_RET (freq == XTIMER_HZ_BASE, -1);
cpu/esp8266/periph/timer.c:    CHECK_PARAM_RET (freq == XTIMER_HZ_BASE, -1);
tests/bench_timers/Makefile:test-xtimer: CFLAGS+=-DTEST_XTIMER -DTIM_TEST_FREQ=XTIMER_HZ -DTIM_TEST_DEV=XTIMER_DEV
```


Looking at all this really needs a lot of concentration D:

cladmi: I could propose something with my ideas and see if it would help.

gebart: Did you see #7630 regarding passing a reference time as argument to _xtimer_set_absolute?
It needs a rebase but it is related to this discussion I think.

cladmi: I did not, I just jumped in the xtimer issue to provide a test, but it means that I had a similar reasoning. I will provide a minimal test to show an error.

cladmi:
> And also set the 32 bit low counter of the xtimer to the end of a period before setting the xtimer

You are talking about `_xtimer_high_cnt` right ?
If I am right, it is to handle boards where the low level timer is less than 32 bit.
As setting the lltimer close to overflow would not be enough to trigger issues.


Your test description is to show the fix in the interrupt handler right ?

I think I will try to show only the issue with `_xtimer_set_absolute` alone first.
This way I have only one timer :)

Josar:
> You are talking about _xtimer_high_cnt right ?

I talk about both. And the not 32 bit timer have more problems in the currrent implementation.

1. Placing the target in the right list affects both 32 and not 32 hardware timers.

If the target is at the beginning of the next low level timer period it will end in the long_list, regardless of the fact that `target - XTIMER_BACKOFF` is in this low level period. But when this period ends and it is set as new low level timer target it will be placed at the end of the new low level period.

2. Sheduling the next target in `_timer_callback` is a problem for not 32 bit low level timer.

3. The target overun problem is only stated here, but not new.

As the test supplied in https://github.com/RIOT-OS/RIOT/pull/10174 enables to see issues, this does not mean that this PR handles them all. It was a first step to even out the most trivial bugs which had a big impact.

De-sheduling is a seperate problem. And can happen at multiple times, this is a problem of not being interrupt safe. Basically a thread might be de-sheduled between reading the xtimer value untill it sets its new target. But this is not meant to be handled in this PR and requires a substantial rework.

jia200x:
I'm running all test with this branch on `samr21-xpro`. Post results later

leandrolanzieri:
@Josar please rebase now that #10174 has been merged.

Josar:
@leandrolanzieri done.

jia200x:
These are the results for [`compile-and-test-for-board.py`](https://github.com/RIOT-OS/Release-Specs/blob/master/02-tests/compile_and_test_for_board.py) with this PR on `samr21-xpro`. Only 2 errors, **completely unrelated to this PR**.
```
ERROR:samr21-xpro:Tests failed: 2
Failures during test:
- [tests/driver_my9221](tests/driver_my9221/test.failed)
- [tests/pkg_fatfs_vfs](tests/pkg_fatfs_vfs/test.failed)
```

sys/include/xtimer/implementation.h
cladmi: @gebart I think your fix still misses something, from what I get, the function should be calling `xtimer_now` from within a masked interrupt context, so that the time between this `now` and the `periph_timer_set` can be guaranteed to be fixed as you cannot be de-scheduled.

If not the comparison with `XTIMER_BACKOFF` does not guarantee anything.

cladmi: I will try to show these with tests.

leandrolanzieri:
Here are some measurements I did to compare results of master to this PR.
**Arduino Duemilanove (ATmega328)**
- Master
master_2009

- With PR
pr_applied_2009

- Conclusion
In master multiple gaps can be seen in workers and main pins. After the PR was applied that only happened 3 times (**always with the same gap time**, maybe caused by what's mentioned [here](https://github.com/RIOT-OS/RIOT/pull/9211#issuecomment-434540028))? but it was improved a lot.

**Arduino DUE**
- Master
master_due

- With PR
pr_applied_due

- Conclusion
Could not reproduce the gaps in master with this board. Workers and the main thread pin appear to be fine. After the PR all is the same.

**Note**: Capture files can be downloaded [here](https://github.com/RIOT-OS/RIOT/files/2542097/captures.zip)

Josar:
@leandrolanzieri with a 32bit hardware timer running with 1MHz this proplem **may** happen once in ca. 71 minutes, if i am not mistaken. To provoke this error you could let the timer run faster then 1MHz.

Besides that one of those fixes only affects not 32bit timers.

Issue #9214, opened 2018-05-28
make: compiling is broken for any ARM platform on OS X.


kYc0o:
#### Description

As per #9080, the newlib-nano directory is not included anymore in my local setup for
ARM platforms. Apparently the reason is that the commands used to add the directory
don't work properly on OS X. 
#### Steps to reproduce the issue

Compile anything on current master for any ARM platform on OS X.

#### Expected results

Compiles correctly.
#### Actual results

```console
arm-none-eabi-gcc \
	-DRIOT_FILE_RELATIVE=\"boards/samr21-xpro/board.c\" \
	-DRIOT_FILE_NOPATH=\"board.c\" \
	-Wall -Werror -Wextra -Wno-implicit-fallthrough -mno-thumb-interwork -mcpu=cortex-m0plus -mlittle-endian -mthumb -mfloat-abi=soft -march=armv6s-m -ffunction-sections -fdata-sections -fno-builtin -fshort-enums -ggdb -g3 -Os -std=c99 -fno-delete-null-pointer-checks -fdiagnostics-color -Wstrict-prototypes -Wold-style-definition -Werror=strict-prototypes -Werror=old-style-definition -fno-common -Wall -include '/Users/facosta/git/RIOT-OS/RIOT/examples/hello-world/bin/samr21-xpro/riotbuild/riotbuild.h'   -isystem  -I/Users/facosta/git/RIOT-OS/RIOT/core/include -I/Users/facosta/git/RIOT-OS/RIOT/drivers/include -I/Users/facosta/git/RIOT-OS/RIOT/sys/include -I/Users/facosta/git/RIOT-OS/RIOT/cpu/samd21/include -I/Users/facosta/git/RIOT-OS/RIOT/boards/samr21-xpro/include -I/Users/facosta/git/RIOT-OS/RIOT/cpu/sam0_common/include -I/Users/facosta/git/RIOT-OS/RIOT/cpu/cortexm_common/include -I/Users/facosta/git/RIOT-OS/RIOT/cpu/cortexm_common/include/vendor -I/Users/facosta/git/RIOT-OS/RIOT/sys/libc/include  -MD -MP -c -o /Users/facosta/git/RIOT-OS/RIOT/examples/hello-world/bin/samr21-xpro/board/board.o /Users/facosta/git/RIOT-OS/RIOT/boards/samr21-xpro/board.c
[...]

PR #9215, opened 2018-05-28
newlib-nano: Add workaround for systems without nano includes.

jcarrano:
### Contribution description

Some systems (OSX, ArchLinux) do not ship separate headers for newlib nano. All the standard system headers for nano should be the same with the exception of `newlib.h`.

The result for these systems is that the `NEWLIB_NANO_INCLUDE_DIR` is empty and the include line is wrongly formed.

This is a workaroround that reverts to the old behavior of using the non-nano include path if `NEWLIB_NANO_INCLUDE_DIR` is empty.

### Issues/PRs references

Fixes #9214, which was introduced by #9080 .

PR #9216, opened 2018-05-28
 newlib-nano: fix include directory

toonst:
### Contribution description
Adding the nano.specs file to CFLAGS make sure the correct include directory is found. This way we don't have to explicitly add the include directory.

### Issues/PRs references
Fixes https://github.com/RIOT-OS/RIOT/issues/9214
Fixes PR #9080 
Alternative to https://github.com/RIOT-OS/RIOT/pull/9215

### DIFF
--- a/makefiles/libc/newlib.mk
+++ b/makefiles/libc/newlib.mk
@@ -69,9 +69,7 @@ ifneq (1,$(USE_NANO_SPECS_FILE))
   # A cross GCC already knows the target libc include path (build-in at compile time)
   # but Clang, when cross-compiling, needs to be told where to find the headers
   # for the system being built.
-  # We also add -nostdinc to avoid including the host system headers by mistake
-  # in case some header is missing from the cross tool chain
-  NEWLIB_INCLUDES := -isystem $(NEWLIB_INCLUDE_DIR) -nostdinc
+  NEWLIB_INCLUDES := -isystem $(NEWLIB_INCLUDE_DIR)
   NEWLIB_INCLUDES += $(addprefix -isystem ,$(abspath $(wildcard $(dir $(NEWLIB_INCLUDE_DIR))/usr/include)))
   ifeq (1,$(USE_NEWLIB_NANO))
     # newlib-nano include directory is called either newlib-nano or nano. Use the one we find first.
--- a/makefiles/libc/newlib.mk
+++ b/makefiles/libc/newlib.mk
@@ -1,7 +1,15 @@
+ifeq ($(BOARD), native)
+  $(error Newlib is not supported on native target)
+endif
+
 ifneq (,$(filter newlib_nano,$(USEMODULE)))
+  USE_NEWLIB_NANO = 1
   # Test if nano.specs is available
   ifeq ($(shell $(LINK) -specs=nano.specs -E - 2>/dev/null >/dev/null </dev/null ; echo $$?),0)
-    USE_NEWLIB_NANO = 1
+    # Clang/llvm has no support for specs files
+    ifneq ($(TOOLCHAIN),llvm)
+      USE_NANO_SPECS_FILE = 1
+    endif
     ifeq ($(shell echo "int main(){} void _exit(int n) {(void)n;while(1);}" | LC_ALL=C $(CC) -xc - -o /dev/null -lc -specs=nano.specs -Wall -Wextra -pedantic 2>&1 | grep -q "use of wchar_t values across objects may fail" ; echo $$?),0)
         CFLAGS += -fshort-wchar
         LINKFLAGS += -Wl,--no-wchar-size-warning
@@ -13,8 +21,9 @@ ifneq (,$(filter newlib_gnu_source,$(USEMODULE)))
   CFLAGS += -D_GNU_SOURCE=1
 endif
 
-ifeq (1,$(USE_NEWLIB_NANO))
+ifeq (1,$(USE_NANO_SPECS_FILE))
   export LINKFLAGS += -specs=nano.specs
+  export CFLAGS += -specs=nano.specs
 endif
 
 export LINKFLAGS += -lc
@@ -45,15 +54,18 @@ NEWLIB_INCLUDE_PATTERNS ?= \
 # the patterns above. We use the -isystem gcc/clang argument to add the include
 # directories as system include directories, which means they will not be
 # searched until after all the project specific include directories (-I/path)
-NEWLIB_INCLUDE_DIR ?= $(firstword $(wildcard $(NEWLIB_INCLUDE_PATTERNS)))
+NEWLIB_INCLUDE_FILE = $(firstword $(wildcard $(addsuffix /newlib.h,$(NEWLIB_INCLUDE_PATTERNS))))
 
 # If nothing was found we will try to fall back to searching for a cross-gcc in
 # the current PATH and use a relative path for the includes
-ifeq (,$(NEWLIB_INCLUDE_DIR))
+ifeq (,$(NEWLIB_INCLUDE_FILE))
+  $(warning Using relative include dir)
   NEWLIB_INCLUDE_DIR := $(abspath $(wildcard $(dir $(shell command -v $(PREFIX)gcc 2>/dev/null))/../$(TARGET_ARCH)/include))
+else
+  NEWLIB_INCLUDE_DIR := $(dir $(NEWLIB_INCLUDE_FILE))
 endif
 
-ifeq ($(TOOLCHAIN),llvm)
+ifneq (1,$(USE_NANO_SPECS_FILE))
   # A cross GCC already knows the target libc include path (build-in at compile time)
   # but Clang, when cross-compiling, needs to be told where to find the headers
   # for the system being built.
@@ -61,13 +73,18 @@ ifeq ($(TOOLCHAIN),llvm)
   # in case some header is missing from the cross tool chain
   NEWLIB_INCLUDES := -isystem $(NEWLIB_INCLUDE_DIR) -nostdinc
   NEWLIB_INCLUDES += $(addprefix -isystem ,$(abspath $(wildcard $(dir $(NEWLIB_INCLUDE_DIR))/usr/include)))
-endif
-
-ifeq (1,$(USE_NEWLIB_NANO))
-  NEWLIB_NANO_INCLUDE_DIR ?= $(NEWLIB_INCLUDE_DIR)/newlib-nano
-  # newlib-nano overrides newlib.h and its include dir should therefore go before
-  # the regular system include dirs.
-  INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(INCLUDES)
+  ifeq (1,$(USE_NEWLIB_NANO))
+    # newlib-nano include directory is called either newlib-nano or nano. Use the one we find first.
+    NEWLIB_NANO_INCLUDE_DIR ?= $(firstword $(wildcard $(addprefix $(NEWLIB_INCLUDE_DIR)/, newlib-nano nano)))
+    # Workaround for systems where the nano header is not installed.
+    # For ArchLinux, see https://bugs.archlinux.org/task/50481
+    ifeq ($(NEWLIB_NANO_INCLUDE_DIR),)
+      NEWLIB_NANO_INCLUDE_DIR = NEWLIB_INCLUDE_DIR
+    endif
+    # newlib-nano overrides newlib.h and its include dir should therefore go before
+    # the regular system include dirs.
+    INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(INCLUDES)
+  endif
 endif
 
 # Newlib includes should go before GCC includes. This is especially important
--- a/makefiles/libc/newlib.mk
+++ b/makefiles/libc/newlib.mk
@@ -20,39 +20,42 @@ ifeq (1,$(USE_NANO_SPECS_FILE))
   export LINKFLAGS += -specs=nano.specs
   export CFLAGS += -specs=nano.specs
 else
-  # Search for Newlib include directories
+  # Try to search for newlib in the standard search path of the compiler for includes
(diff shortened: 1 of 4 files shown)


jcarrano:
Works for me on Arch. I'm closing #9215 as it is hacky.

toonst:
@gebart No I did not. And you're right, it needs some extra handling. I'm looking into it.

jcarrano:
@toonst Is is acceptable to fall back to what your previous PR (#9080 ) did if the toolchain is clang? And in that case, could we force a failure if nano's headers are not found, so that the error becomes more evident?

pwillem:
FYI I think this only works if your newlib version >= 2.8.0 (because of https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=commit;h=be5926babbfb23ded1de2181a7524b8c3eebe34c )

kYc0o:
CCN package is broken with this PR... I'd suggest to revert to the original state and then try another option for this?

smlng:
can we get this merged, need this to compile arm based boards on macOS!!

kYc0o:
@smlng have you seen Murdock's output?

smlng:
ah damn, but its only ccn-lite 😬 otherwise we should think about reverting #9080?

smlng:
so its a problem with all external packages?

toonst:
Sorry it took this long. 
I made some extra checks for the existence of a newlib.h file and error when trying to compile for native. I'm looking at getting newlib to work on linux, so if that is working we can remove the check. 
Could you test if this is working properly on OSX? Suggestions welcome!

smlng:
still works on macOS but also still get error when compiling ccn-lite-relay for any arm board.

jcarrano:
@smlng @kYc0o What happens with CCN is that the CFLAGS get passed twice to the compiler.

kYc0o:
I see, then it needs a check in the pkg makefile. @toonst do you mind to provide a patch for that also? Otherwise this will be broken and cannot be merged... It can also come in a separated PR, I can do it later if needed.

toonst:
@kYc0o I don't know cmake, but if I just remove the RIOT CFLAGS it seems to be fine. You think this is OK?

smlng:
also note new error reported by murdock https://ci.riot-os.org/RIOT-OS/RIOT/9216/20cf3fd8064cbcb7f7725999d5fe6cced27d7215/output/compile/tests/msg_avail/pic32-wifire.txt

jcarrano:
@smlng @toonst What if `-specs` is made part of the `CC` variable instead of a `CFLAG`? In a way it mitght even make more sense, because even if packages could override `CFLAGS`, the `specs` flag defines lower level compilation and linking details which are not supposed to change. It would also make it easier with clang, since specs is not a valid flag for clang.

kYc0o:
@toonst have you tried what @jcarrano suggests? The packages still failing 🙁

toonst:
@jcarrano @kYc0o adding the specs flag to CC doesn't work for cmake:
```
  The CMAKE_C_COMPILER:
    arm-none-eabi-gcc -specs=nano.specs
  is not a full path and was not found in the PATH.
```

jcarrano:
I just submitted an issue (#9254). We should fetch someone who knows about cmake.

tcschmidt:
@josephnoir cmake expert requested

josephnoir:
From the top of my head ... cmake adopts the `CFLAGS` into its own `CMAKE_C_FLAGS`, but only if the latter isn't already set ([docs](https://cmake.org/cmake/help/latest/envvar/CFLAGS.html)). Maybe they are added maynually added during configuration again? Since I have no clue how the build systems work together I'll have to take a look at whats going on to make a more educated guess.

josephnoir:
@smlng and I looked at the code a bit (since I'm not that familiar with the RIOT build system) and it does indeed look like the `C_FLAGS` are added twice for both ccn lite and jerryscript. Removing the extra definition added to `RIOT_CFLAGS` here and `EXT_CFLAGS` for jerryscript. See #9281

tcschmidt:
Do all tests complete now? Are we ready to finish the review?

makefiles/libc/newlib.mk
```diff
@@ -40,34 +42,34 @@ NEWLIB_INCLUDE_PATTERNS ?= \
...
-  NEWLIB_NANO_INCLUDE_DIR ?= $(NEWLIB_INCLUDE_DIR)/newlib-nano
-  # newlib-nano overrides newlib.h and its include dir should therefore go before
-  # the regular system include dirs.
-  INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(INCLUDES)
+  ifeq (1,$(USE_NEWLIB_NANO))
+    # newlib-nano include directory is called either newlib-nano or nano. Use the one we find first.
+    NEWLIB_NANO_INCLUDE_DIR ?= $(firstword $(wildcard $(addprefix $(NEWLIB_INCLUDE_DIR)/, newlib-nano nano)))
+    ifneq ($(NEWLIB_NANO_INCLUDE_DIR),)
+      # newlib-nano overrides newlib.h and its include dir should therefore go before
+      # the regular system include dirs, but after the other newlib includes.
```
gebart: Still not seeing why the nano include dir should go *after* the newlib include. newlib-nano overrides the newlib.h and therefore should go before newlib include or otherwise we will get the newlib.h from the regular newlib include directory.

pkg/ccn-lite/Makefile
```diff
@@ -5,7 +5,7 @@ PKG_LICENSE=ISC
 
 .PHONY: all
 
-export RIOT_CFLAGS = $(CFLAGS) $(INCLUDES)
+RIOT_CFLAGS := $(INCLUDES)
```
gebart: Does this work? Isn't `$(INCLUDES)` empty at this point? Should this be `=` instead of `:=`?

makefiles/libc/newlib.mk
toonst: You're right, I missed your first comment on it. Fixed!

tests/rmutex/Makefile
```diff
@@ -1,5 +1,8 @@
 include ../Makefile.tests_common
 
+# MIPS based boards are black listed due to some issue with newlib and _Atomic
+BOARD_BLACKLIST := mips-malta pic32-clicker pic32-wifire
```
smlng: note: this should be removed, if someone has solution for the MIPS toolchain. It seems that some include paths are mixed up and the mips-gcc cannot find some system header files?!

makefiles/libc/newlib.mk
```diff
@@ -13,61 +14,60 @@ ifneq (,$(filter newlib_gnu_source,$(USEMODULE)))
...
-# Ubuntu seem to put a copy of the newlib headers in the same place as
-# Gentoo crossdev, but we prefer to look at /etc/alternatives first.
-# On OSX, newlib includes are possibly located in
-# /usr/local/opt/arm-none-eabi*/arm-none-eabi/include or /usr/local/opt/gcc-arm/arm-none-eabi/include
-NEWLIB_INCLUDE_PATTERNS ?= \
-  /etc/alternatives/gcc-$(TARGET_ARCH)-include \
-  /usr/$(TARGET_ARCH)/include \
-  /usr/local/opt/$(TARGET_ARCH)*/$(TARGET_ARCH)/include \
-  /usr/local/opt/gcc-*/$(TARGET_ARCH)/include \
-  #
```
cladmi: The comment for the backslash terminated line was removed.

makefiles/libc/newlib.mk
```diff
@@ -13,61 +14,60 @@ ifneq (,$(filter newlib_gnu_source,$(USEMODULE)))
...
+  NEWLIB_INCLUDES := -isystem $(NEWLIB_INCLUDE_DIR)
+  NEWLIB_INCLUDES += $(addprefix -isystem ,$(abspath $(wildcard $(dir $(NEWLIB_INCLUDE_DIR))/usr/include)))
+  ifeq (1,$(USE_NEWLIB_NANO))
+    # newlib-nano include directory is called either newlib-nano or nano. Use the one we find first.
+    NEWLIB_NANO_INCLUDE_DIR ?= $(firstword $(wildcard $(addprefix $(NEWLIB_INCLUDE_DIR)/, newlib-nano nano)))
+    ifneq ($(NEWLIB_NANO_INCLUDE_DIR),)
+      # newlib-nano overrides newlib.h and its include dir
+      # should therefore go before the other newlib includes.
+      NEWLIB_INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(NEWLIB_INCLUDES)
+    endif
```
cladmi: What happens if there is no newlib-nano ?
Is it ok to ignore this as it was right now ?

jcarrano: In theory, code that depends on having newlib-nano  should test for the appropriate macros, so that compilation fails right away instead if causing weird errors.

cladmi: From reading around, the difference between newlib and newlib-nano, should, according to https://community.nxp.com/thread/389086, almost only be code size, so with how things are usually used, no API dependency.
So there should not be any "dependency on newlib-nano" that should be tested in the code.
However I would expect the build system to report it.

However, my knowledge on the topic is limited.

gebart: The size and layout of some internal newlib variables and structs are affected, for example sizeof(struct reent). Using the newlib.h from normal newlib when linking against newlib nano will probably work without any errors, but waste RAM. The other way around will result in buffer overflow bugs that will be difficult to debug.

tests/rmutex/Makefile
cladmi: The difference between master and this pr with `make info-build`:

``` diff
diff -u info-build-master  info-build-pr 
--- info-build-master   2018-06-19 13:47:58.983288813 +0200
+++ info-build-pr       2018-06-19 13:48:08.447320615 +0200
@@ -35,6 +35,8 @@
 FEATURES_CONFLICT_MSG: 
 
 INCLUDES: 
+       -isystem  
+       /opt/mips-mti-elf/2016.05-03/mips-mti-elf/include  
        -I/home/harter/work/git/RIOT/core/include  
        -I/home/harter/work/git/RIOT/drivers/include  
        -I/home/harter/work/git/RIOT/sys/include 
```

I also get warnings two times when doing info-build:

```
make -C tests/rmutex BOARD=pic32-wifire info-build > info-build-pr
/home/harter/work/git/RIOT/makefiles/libc/newlib.mk:51: newlib.h not found, guessing newlib include path from gcc path
/home/harter/work/git/RIOT/makefiles/libc/newlib.mk:51: newlib.h not found, guessing newlib include path from gcc path
```

gebart: Where is your newlib installation located? (newlib.h)
What distro?

cladmi: Found that the two times  warning comes from `$(MAKE) info-boards-supported` being called during `make info-build`.

-----

I use ubuntu 16.04 and I installed the mips-mti-elf as explained in the Wiki I think:

```
$ /opt/mips-mti-elf/2016.05-03/bin/mips-mti-elf-gcc --version
mips-mti-elf-gcc (Codescape GNU Tools 2016.05-03 for MIPS MTI Bare Metal) 4.9.2
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

And I have a only this `newlib.h` in mips-mti-elf directory, no `newlib-nano` directory:

`/opt/mips-mti-elf/2016.05-03/mips-mti-elf/include/newlib.h`

makefiles/libc/newlib.mk
```diff
@@ -13,61 +14,61 @@ ifneq (,$(filter newlib_gnu_source,$(USEMODULE)))
...
-endif
+  # Search for a newlib.h inside the above newlib include directory patterns
+  NEWLIB_INCLUDE_FILE = $(firstword $(wildcard $(addsuffix /newlib.h,$(NEWLIB_INCLUDE_PATTERNS))))
 
-ifeq ($(TOOLCHAIN),llvm)
-  # A cross GCC already knows the target libc include path (build-in at compile time)
-  # but Clang, when cross-compiling, needs to be told where to find the headers
-  # for the system being built.
-  # We also add -nostdinc to avoid including the host system headers by mistake
-  # in case some header is missing from the cross tool chain
```
cladmi: This `ifeq` section with the `gcc` case was removed and it adds a wrong `-isystem` directory for mips-gcc 4.9.2 on my computer.

It supplies `/opt/mips-mti-elf/2016.05-03/mips-mti-elf/include/` which replaces `/opt/mips-mti-elf/2016.05-03/lib/gcc/mips-mti-elf/4.9.2/include/`

Which breaks compilation for mips in the `rmutex` example.

However, re-adding the `ifeq` would only solve my case but maybe not `llvm` as it would be using the wrong directory.

makefiles/libc/newlib.mk
cladmi: @jcarrano Do you have some macros that could be checked to verify if newlib-nano is used ? I would like to verify that right now, some configuration are already broken and not using newlib-nano when requested.

pkg/ccn-lite/Makefile
cladmi: As it is called with `make -C pkg/PKGNAME` the variable was already set from environment variable.
However, there is not reasons to set it with `:=` either.

smlng:
this fixes (all) my issues on macOS with newlib, right now (2018-06-22) master is broken there

smlng:
however, I also agree with @cladmi that we should introduces changes/fixes step-by-step, ie. nano-spec, include-dir, and llvm support.

makefiles/libc/newlib.mk
toonst: Can you try if it works with the latest additions?

toonst:
Closed in favor of #9394 and #9415

PR #7588, opened 2017-09-08
cpu/sam0: re-work i2c driver

dylad:
This PR introduces a common i2c_config struct in `periph_conf.h` for all sam0 based boards. Additionnal I2C SERCOM can be add more easily.
SAML21 can know use SERCOM5 if desired.

I'm wondering if I can remove the init of GCLK_SLOW because it is only for some SMBus timeout (no used yet by RIOT). Should I remove it ? This should not impact the current driver state but I want the community's opinion first :)

I also tried to keep the 80 characters limit per lines.

PR #9219, opened 2018-05-29
drivers/periph_i2c: i2c_init should return void

aabadie:
### Contribution description

With the current design, `i2c_init` is called from `periph_init` and where return code is not used. This PR change this function to return `void `.

If accepted, this PR should be merged before #9202, #7588 and #9208. And these PRs updated after.

### Issues/PRs references

#6577


dylad:
SPI bus have the same behaviour. Well, I guess it's not harmful.

PR #9221, opened 2018-05-29
boards/msbiot: on-board buttons and saul integration

maribu:
This PR adds:
 - support for the on-board buttons
- SAUL integration of on-board buttons and on-board LEDs
- Added the MPU-9150 as default saul device (driver is loaded when module `saul_default` is used)

### DIFF
--- a/boards/msbiot/Makefile.dep
+++ b/boards/msbiot/Makefile.dep
@@ -2,3 +2,7 @@
 ifneq (,$(filter netdev_default gnrc_netdev_default,$(USEMODULE)))
   USEMODULE += cc110x
 endif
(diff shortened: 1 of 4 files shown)


aabadie:
To avoid confusion, I updated this PR's title since only it's only related to msbiot board

maribu:
Should I update the commit titles in the same way?

maribu:
I just also added saul integration of the on-board MPU-9150 to this PR, as this commit fits nicely into this PR

aabadie:
> Should I update the commit titles in the same way?

yes, they should start with `boards/msbiot: ...`

maribu:
Done :-)

Issue #9225, opened 2018-05-29
cc110x driver on STM32F4xxx MCUs not working reliable


maribu:
Outdated, see bottom of thread

maribu:
PR for Variant 1: https://github.com/RIOT-OS/RIOT/pull/9230

maribu:
PR for Variant 2: https://github.com/RIOT-OS/RIOT/pull/9231

I personally favor this version

maribu:
Please note that I added two more possible solutions to the first message. I really hate solution variant 3, as latency could be extremely relevant for some use cases, e.g. tactile Internet. (And, btw: I get a RTT of 0.1ms on the MSB IoT, and 0.3ms on the MSBA2 according to `ping6`. This is seems quite good to me.)

maribu:
My fixes stopped working on my board. Maybe I'm extremely unlucky and this is an hardware problem which somehow magically disappeared at the same time I changed my code?

maribu:
Tested with 8 devices, same issue as before. Either in all 4 pairs at least one device is faulty, or it is a software issue that I somehow fixed without noticing and commiting the changes :-( Shit :-(

maribu:
Maybe someone comes up with an idea if I provided all information I'm aware of:

When the cc110x worked on the MSB IoT, I got RTTs of 0.100ms to 0.101 ms. Now I get about 16ms RTTs in the rare case I get a corresponding pong. This was also the case before I started working on the cc110x driver.

In current state it takes from "cc110x_send()" to packet was send about 7ms. So something is wasting a *lot* of time between...

The MSBA2 has RTTs of 0.3ms, so I believe the 0.1ms is plausible.

I changed the SPI clock to 1MHz to see what happens. With that it worked reliable on the MSB IoT, but with RTTs of 17ms. So the actual SPI communication does not contribute much to the RTT.

 The cc110x_cs takes about 5µs (and I believe most of the time is the overhead of getting the current time and printing it over UART.) So that is not the cause of the slow RTT. I believe the slow sending and the high packet  loss rate is somehow related. And it is obviously SPI related, as packets get corrupted without being detected by CRC and changing the SPI clock solves the problem.

If you have any idea was could cause the problem, no matter how unlikely it seems, I'm more that happy to hear it.

I checked if in cc110x_cs GDO1 ever was high (chip not yet ready), but that never occured.

maribu:
[...]

PR #9231, opened 2018-05-29
cpu/stm32_common: Reset pin function in gpio_init()

maribu:
This PR implements the second variant of how to fix issue https://github.com/RIOT-OS/RIOT/issues/9225. Please follow the discussion in the issue before merging, as there is an alternative way to fix the issue.


maribu:
This does not solve the problem :-(

PR #9236, opened 2018-05-29
crypto: Poly1305 implementation

bergzand:
### Contribution description

Minimal implementation of the poly1305 one time message authentication code.

The code is strongly based on the public domain implementation from https://github.com/floodyberry/poly1305-donna

Tests based on the tests vectors as provided [rfc7539 Appendix A.3](https://tools.ietf.org/html/rfc7539#appendix-A.3).

### Issues/PRs references

None


bergzand:
Large rewrite of the algorithm to reduce the flash usage significantly. Algorithm itself eats around 1.1K of flash on a samr21-xpro.

MrKevinWeiss:
Are big endian systems a concern?

bergzand:
> Are big endian systems a concern?

Hmmm, let me check for a second. Do we have any systems to check this on?

bergzand:
I think it should also work on big endian systems, but I'm not 100% sure on this. The `u8to32` and `u32to8` calls should be endian-independent.

MrKevinWeiss:
Apparently MIPS boards are big endian...  #4710 talks a bit about it.

At first I was looking at the `u8to32` and `u32to8` however it also appears that you use a uint64_t in that call (so I don't know if it would truncate the useful 32 bits).
```
    const uint64_t uu0 = (u4 >> 2) * 5 + ctx->h[0] + ctx->pad[0];
    u32to8(mac, uu0);
```
My question was more, should we consider this?  I don't have any mips boards here to test on unfortunately.

Also the unittest doesn't really compiles for any boards (it is pretty big) but I will find a way around it and just run the single tests.

gebart:
AFAIK none of the boards supported in RIOT are big endian (including the MIPS and pic32 boards)

MrKevinWeiss:
Thanks, @gebart, makes development easy then.  Please disregard previous comments.

gebart:
Just to clarify: I am not opposed to making everything endian independent. I only stated that it will be difficult to test with the currently supported boards.

MrKevinWeiss:
Maybe we should continue with this?

bergzand:
> Maybe we should continue with this?

Mind if I rebase this first?

MrKevinWeiss:
That would be lovely :)

bergzand:
@MrKevinWeiss rebased!

sys/crypto/poly1305.c
```diff
@@ -0,0 +1,250 @@
...
+    ctx->final = 0;
+}
+
+static void poly1305_blocks(poly1305_ctx_t *ctx, const uint8_t *data, size_t len)
+{
+    const uint32_t hibit = (ctx->final) ? 0 : (1UL << 24); /* 1 << 128 */
+    uint32_t r0, r1, r2, r3, r4;
+    uint32_t s1, s2, s3, s4;
+    uint32_t h0, h1, h2, h3, h4;
+    uint64_t d0, d1, d2, d3, d4;
```
MrKevinWeiss: please move the d and c variables inside the scope of the while loop so the CI is happy.

bergzand:
> I ran the tests and everything seems fine. Please make the CI happy and squash.

It seems that I manage to lose a commit here with the rebased, added it again. :cry: 

Complaints should be fixed now.

MrKevinWeiss:
Hmm unrelated murdock error I will rerun.  
I retested and everything passes on native.
If your ready I think it can be squashed!

bergzand:
squashed!

MrKevinWeiss:
OK ACK!  I will wait an hour or two in case someone has a different opinion then merge!

bergzand:
@MrKevinWeiss Thanks for reviewing!

PR #9240, opened 2018-05-30
drivers/isrpipe: changed isrpipe_t content order

maxvankessel:
Wouldn't it be more handy to change order of content in `isrpipe_t`? 
So `tsrb` functions like `tsrb_full` and `tsrb_empty` can be accessed with an `isrpipe_t`.

`isrpipe_t` extends `tsrb_t` right?

### DIFF
--- a/sys/include/isrpipe.h
+++ b/sys/include/isrpipe.h
@@ -35,8 +35,8 @@ extern "C" {
  * @brief   Context structure for isrpipe
  */
 typedef struct {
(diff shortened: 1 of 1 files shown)


cgundogan:
@kaspar030 I am not quite familiar with the `isrpipe` module. To me the change looks quite unintrusive. Would you mind having a second look at it?

kaspar030:
Thanks @maxvankessel!

PR #9244, opened 2018-05-30
cpu/saml21: Fix possibly uninitialized variable in pm.c.

jcarrano:
### Contribution description

The code in pm.c for saml32 was missing a `default` case, this triggered a `-Wmaybe-unititialized`.

`-Wmaybe-unititialized` warnings are somewhat unreliable (see [this answer](https://stackoverflow.com/a/14132910)). Depending on optimization options, sometimes the compiler is able to detect it (sometimes it gives a false positive). Would it be a good idea to turn this warning back into a warning to avoid having unpredictable compilation errors?

### DIFF
--- a/cpu/saml21/periph/pm.c
+++ b/cpu/saml21/periph/pm.c
@@ -38,6 +38,7 @@ void pm_set(unsigned mode)
                 DEBUG("pm_set(): setting STANDBY mode.\n");
                 _mode = PM_SLEEPCFG_SLEEPMODE_STANDBY;
                 break;
+            default: /* Falls through */
             case 2:
                 DEBUG("pm_set(): setting IDLE mode.\n");
                 _mode = PM_SLEEPCFG_SLEEPMODE_IDLE2;


dylad:
Just wondering : Can't we just replace `case 2` by `default` ?

basilfx:
The default case is [better](https://github.com/RIOT-OS/RIOT/pull/8861#discussion_r179074575).

tcschmidt:
@josephnoir any advise here?

jcarrano:
@dylad I think that having
```
default:
case 2:
  code();
```
Makes it clear that "2" is an expected value, and that for other values we do the same as if the value was 2. Of course, if "case 2:" is omitted, the program is perfectly equivalent, but in my opinion less clear.

dylad:
Sure ! I was just wondering.
If we take this option, please add a comment `/* fall throught */` so we won't raise a new compiler warning later.

jcarrano:
@dylad Done.

jia200x:
Murdock is green.
@kaspar030 @dylad @basilfx is it ok for you to merge this PR?

jcarrano:
Should I rebase?

jia200x:
@jcarrano yup. And then we merge

jcarrano:
> Also, would it be a good idea to enable -Wswitch-default to warn about switch statements without default case?

What is the correct place to discuss such a proposal?

jia200x:
@jcarrano either a GH issue or devel list

jia200x:
Murdock green and go

gebart:
@jcarrano Create a new PR for it. It sounds like a good idea to enable that warning for all platforms which support it.

jia200x:
Murdock green -> GO

PR #9246, opened 2018-05-30
drivers/pppos: initial ppp over serial layer

maxvankessel:
Point to Point Protocol over Serial (PPPoS) works with HDLC framing, according to RFC1662.

This is the lower layer, and HDLC layer comes on top of this. (See  #9200)


jia200x:
@maxvankessel any news on this?

jia200x:
this depends on #9200

PR #9247, opened 2018-05-30
sys/ps: Use correct define for TLSF information.

jcarrano:
### Contribution description

I'm fixing my own bugs here.

The macro MODULE_TLSF_MALLOC indicates if tlsf is being used as the system-wide allocator. MODULE_TLSF only incates if TLSF is present.

PS should check for MODULE_TLSF_MALLOC to decide if heap information should be displayed.

### Issues/PRs references

The error was introduced in PR #9006.

### DIFF
--- a/sys/ps/ps.c
+++ b/sys/ps/ps.c
@@ -26,7 +26,7 @@
 #include "xtimer.h"
 #endif
 
-#ifdef MODULE_TLSF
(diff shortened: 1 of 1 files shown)


cgundogan:
again travis with
```
sys/ps/ps.c:84: error (cppcheckError): Analysis failed. If the code is valid then please report this failure.
```

ignoring, because it's a false positive.

PR #9251, opened 2018-05-30
Revert "newlib-nano: add include directory"

toonst:
This reverts commit 88aaa623641753c1b9eb164b9b1114e9c2c0700c.

As a temporary fix for #9214. Eventually  #9216 should be improved to properly handle this.


kYc0o:
@miri64 can we merge this as is? I mean, as a revert commit.

PR #8814, opened 2018-03-22
cpu/kinetis: Refactor LPTMR timer implementation

gebart:
### Contribution description

Refactor Kinetis LPTMR driver to use freerunning counter mode (TFC flag) and get rid of the RTT subtick counter dependency in the implementation.

Comparison between the bench_periph_timer results before and after reveal that the problem reported in #8532 is fixed, and the new implementation provides results which are close to optimal for a 32768 Hz timer.

### Issues/PRs references

Fixes #8532 
Uses #8531 for benchmarking
Depends on #8928 

#### Good results with this PR

```
------------- BEGIN STATISTICS --------------
Limits: mean: [-41, 41], variance: [58, 99]
Target error (actual trigger time - expected trigger time), in reference timer ticks
positive: timer is late, negative: timer is early
=== timer_set running ===
   interval    count       sum       sum_sq    min   max  mean  variance
   1 -    2:   25705    455646      2002373      2    35    17     77
   3 -    4:   25533    457326      1973191      2    35    17     77
   5 -    8:   25412    451046      1970014      2    35    17     77
   9 -   16:   25699    458563      2017198      2    36    17     78
  17 -   32:   25832    457768      2018909      1    35    17     78
  33 -   64:   25758    440454      1978830      1    35    17     76
  65 -  128:   25335    414320      1942196      0    34    16     76
 129 -  256:   25254    367689      1979616     -3    32    14     78
 257 -  512:   25677    285822      2020736     -7    31    11     78
      TOTAL   230205   3788634     18459378     -7    36    16     80
=== timer_set resched ===
   interval    count       sum       sum_sq    min   max  mean  variance
   1 -    2:   25369    484188      2000763      2    38    19     78
   3 -    4:   25188    509764      1958162      4    36    20     77
   5 -    8:   25546    513382      1994162      4    36    20     78
   9 -   16:   25491    515802      1980773      4    36    20     77
  17 -   32:   25565    513309      1976695      4    36    20     77
  33 -   64:   25536    497785      1975338      3    36    19     77
  65 -  128:   25226    469163      1978715      2    35    18     78
 129 -  256:   25580    433869      1984646      0    34    16     77
 257 -  512:   25562    344196      2051719     -5    32    13     80
      TOTAL   229063   4281458     20398742     -5    38    18     89
=== timer_set stopped ===
   interval    count       sum       sum_sq    min   max  mean  variance
   1 -    2:   25358    414569      1978321      0    32    16     78
[...]

PR #9262, opened 2018-06-01
kinetis: Update i2c driver to new API

gebart:
### Contribution description

Update the Kinetis I2C driver for the new API

Needs some more testing

### Issues/PRs references

#6577

### DIFF
--- a/boards/teensy31/Makefile.features
+++ b/boards/teensy31/Makefile.features
@@ -1,11 +1,13 @@
 # Put defined MCU peripherals here (in alphabetical order)
-FEATURES_PROVIDED += periph_cpuid
-FEATURES_PROVIDED += periph_gpio
 FEATURES_PROVIDED += periph_pwm
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
 FEATURES_PROVIDED += periph_timer
 FEATURES_PROVIDED += periph_uart
 
 # The board MPU family (used for grouping by the CI system)
 FEATURES_MCU_GROUP = cortex_m4_2
 
-include $(RIOTCPU)/cortexm_common/Makefile.features
+include $(RIOTCPU)/kinetis/Makefile.features
+# No HWRNG in MK20D7 devices
+FEATURES_PROVIDED := $(filter-out periph_hwrng,$(FEATURES_PROVIDED))
--- a/cpu/kinetis/periph/i2c.c
+++ b/cpu/kinetis/periph/i2c.c
@@ -30,6 +30,8 @@
 #include "irq.h"
 #include "bit.h"
 #include "mutex.h"
+#include "thread.h"
+#include "thread_flags.h"
 #include "periph_conf.h"
 #include "periph/i2c.h"
 
@@ -47,6 +49,11 @@
 #else
 #define TRACE(...)
 #endif
+
+/**
+ * @brief   Thread flag used internally for signaling between ISR and user thread
+ */
+#define THREAD_FLAG_KINETIS_I2C (1u << 8)
 
 /**
  * @brief   Array of I2C module clock dividers
@@ -68,19 +75,42 @@ static const uint16_t i2c_dividers[] = {
 };
 
 /**
+ * @brief   I2C IRQ reception state
+ */
+typedef struct {
+    uint8_t *datap; /**< pointer to output buffer */
+    size_t bytes_left; /**< how many bytes left to receive */
+} i2c_rx_t;
+
+/**
+ * @brief   I2C IRQ transmission state
+ */
+typedef struct {
+    const uint8_t *datap; /**< pointer to input buffer */
+    size_t bytes_left; /**< how many bytes left to transmit */
+} i2c_tx_t;
+/**
  * @brief   Driver internal state
  */
 typedef struct {
     mutex_t mtx; /**< Mutex preventing multiple users of the same bus */
-    unsigned active; /**< State variable to help catch user mistakes */
+    kernel_pid_t pid; /**< PID of thread waiting for a transfer to complete */
+    uint8_t active; /**< State variable to help catch user mistakes */
+    union {
+        i2c_tx_t tx; /**< State of ongoing transmission sequence */
+        i2c_rx_t rx; /**< State of ongoing reception sequence */
+    };
+    int retval; /**< return value from ISR */
 } i2c_state_t;
-static i2c_state_t i2c_state[I2C_NUMOF];
 
+
+static i2c_state_t i2c_state[I2C_NUMOF];
 
 int i2c_acquire(i2c_t dev)
 {
     assert((unsigned)dev < I2C_NUMOF);
     mutex_lock(&i2c_state[dev].mtx);
+    i2c_state[dev].pid = thread_getpid();
     return 0;
 }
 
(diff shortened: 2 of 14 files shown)


cpu/kinetis/periph/i2c.c
```diff
@@ -45,474 +48,322 @@
...
+      28,   32,   36,   40,   44,   48,   56,   68, /* 0x08 ~ 0x0f */
+      48,   56,   64,   72,   80,   88,  104,  128, /* 0x10 ~ 0x17 */
+      80,   96,  112,  128,  144,  160,  192,  240, /* 0x18 ~ 0x1f */
+     160,  192,  224,  256,  288,  320,  384,  480, /* 0x20 ~ 0x27 */
+     320,  384,  448,  512,  576,  640,  768,  960, /* 0x28 ~ 0x2f */
+     640,  768,  896, 1024, 1152, 1280, 1536, 1920, /* 0x30 ~ 0x37 */
+    1280, 1536, 1792, 2048, 2304, 2560, 3072, 3840, /* 0x38 ~ 0x3f */
 };
 
+static const i2c_conf_t i2c_config[I2C_NUMOF] = I2C_CONFIG;
```
aabadie: Why not directly defining this array of structs in the board periph_conf (like stm32, sam0, etc) instead of using the I2C_CONFIG macro ?

gebart: I wanted to avoid unintentionally instancing this local array inside any other files, so I decided to put it as a define instead.

aabadie: Ok but then it's not consistent with the other configurations arrays (UART, SPI, PWN, ADC, etc).

gebart:
Cherry picked in #8933 #8814 #8930 to avoid rebase problems later

cpu/kinetis/periph/i2c.c
gebart: alright, will change this for now

gebart:
Refactored to use IRQs when transferring the actual data bytes. This method lets the CPU perform other tasks while waiting for a slow I2C slave.

boards/teensy31/Makefile.features
```diff
@@ -1,11 +1,13 @@
 # Put defined MCU peripherals here (in alphabetical order)
-FEATURES_PROVIDED += periph_cpuid
-FEATURES_PROVIDED += periph_gpio
 FEATURES_PROVIDED += periph_pwm
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
```
dylad: Is this related ?

cpu/kinetis/periph/i2c.c
```diff
@@ -45,474 +51,502 @@
...
-    int n = 0;
-
-    while (length > 0) {
-        TRACE("i2c: tx: %02x\n", data[n]);
-        dev->D = data[n];
-
-        while (!(dev->S & I2C_S_IICIF_MASK));
-
-        dev->S = I2C_S_IICIF_MASK;
+    I2C_Type *i2c = i2c_config[dev].i2c;
```
dylad: Could you add a guard here please ? 
```
if (len == 0 || data == NULL) {
    return -EINVAL;
}
```

dylad: Could you also return in case of unsupported features ?

boards/teensy31/Makefile.features
gebart: I had to pull in some upstream merged PRs to avoid problems rebasing later.

cpu/kinetis/periph/i2c.c
gebart: `len == 0` is not an error, it can be used to send a start condition or stop condition, depending on flags.
I don't think there are any unsupported flags here (other than REG16, which I think we can just ignore in the _bytes functions)

dylad: > it can be used to send a start condition or stop condition, depending on flags

I never tought of it but is this really useful for the user ? I mean, we're suppose to read or write on the bus so why just sending a start out of the blue ?

gebart: The hih6130 is one weird device which uses empty writes to trigger measurements. The user sends a start condition for writing followed by a stop condition, so zero data bytes. Then when reading the device will return the most recent measurement.

dylad: > The hih6130 is one weird device which uses empty writes to trigger measurements

Indeed, I'll look at this one. Is this even I2C/TWI compliant ?

gebart: Dunno, but it's a device that we have a driver for

gebart:
It would be a good idea to test with the hih6130 as well because of the null write behaviour, but I don't remember where I put my breakout board for that sensor..

gebart:
Added some asserts for enabling interrupts and a check to ensure that data != NULL when len > 0

dylad:
@gebart I tend to ACK this PR but I have no hardware to test it :(
Do you know if someone have the required hardware for testing ?

gebart:
I have Mulle and frdm boards, and I know that there are some kinetis boards among the FU Berlin people

dylad:
I guess I'll resend an email on the devel mailing to require some help for testing

MichelRottleuthner:
Just tested this with pba-d-01-kw2x and the already merged srf08 driver. Works fine

cpu/kinetis/periph/i2c.c
```diff
@@ -45,474 +51,515 @@
...
-#if I2C_1_EN
-    [I2C_1] = MUTEX_INIT,
-#endif
-#if I2C_2_EN
-    [I2C_2] = MUTEX_INIT
-#endif
-#if I2C_3_EN
-    [I2C_3] = MUTEX_INIT
-#endif
+#define THREAD_FLAG_KINETIS_I2C (1u << 8)
```
gebart: Added this to avoid dependency on contested PR #9284. We can update the driver later if a consensus is reached regarding the reserved thread flags

dylad:
Just re-trigger Travis and we're good to go.

cpu/kinetis/periph/i2c.c
```diff
@@ -45,474 +51,515 @@
...
+        i2c_state[dev].retval = 0;
+        bit_set8(&i2c->C1, I2C_C1_IICIE_SHIFT);
+        /* Initiate master receive mode by reading the data register once when
+         * the C1[TX] bit is cleared and C1[MST] is set */
+        volatile uint8_t dummy;
+        dummy = i2c->D;
+        ++dummy;
+        /* Wait until the ISR signals back */
+        TRACE("i2c: read C1=%02x S=%02x\n", (unsigned)i2c->C1, (unsigned)i2c->S);
+        thread_flags_t tflg = thread_flags_wait_any(THREAD_FLAG_KINETIS_I2C | THREAD_FLAG_TIMEOUT);
```
gebart: We are checking for the TIMEOUT thread flag, but there is not yet anything setting it in the driver. I left this thread flag check here to make it simple in the future to add a timeout, but I did not yet find a satisfying solution which does not add extra dependencies to the driver. This part of the code can still be used if a user wants to eliminate the possibility of a bus hang blocking the application, they can use xtimer to provide a timer which sets the timeout flag.

dylad: I'm fine with this, switching thread while waiting here is an excellent idea and IMHO we should do it for every MCU/drivers :)

dylad:
Here we go ! thanks @gebart for your work !

PR #9263, opened 2018-06-01
tools/testrunner: fix long lines (E501)

yegorich:


dist/tools/testrunner/testrunner.py
```diff
@@ -37,7 +37,8 @@ def find_exc_origin(exc_info):
 
 def run(testfunc, timeout=10, echo=True, traceback=False):
     env = os.environ.copy()
-    child = pexpect.spawnu("make term", env=env, timeout=timeout, codec_errors='replace')
+    child = pexpect.spawnu("make term", env=env, timeout=timeout,
```
basilfx: This isn't compliant with PEP-8: https://www.python.org/dev/peps/pep-0008/#id17.

It should be aligned like this:

```python
child = pexpect.spawnu(
    "make term", env=env, timeout=timeout, codec_errors='replace')
```

yegorich:
@basilfx I see now. Thanks for reviewing. Then I'll drop the patch.

basilfx:
No problem! Still, thanks for the the PR ;-)

PR #9267, opened 2018-06-01
periph/i2c: Add 10 bit addressing magic number definition

gebart:
Special bit pattern indicating a 10 bit address is used for avoiding magic numbers in CPU implementations


dylad:
Why Murdock is now required for merging on this branche ??

aabadie:
@miri64 @kaspar030 is it possible to disable the PR merging blocker if one of the CI fails, just for this branch ? I think someone changed that recently in the project settings.

aabadie:
let's go her as well

PR #9272, opened 2018-06-02
examples/gnrc_border_router: fix path to uhcpd in README

Burdaa:
### Contribution description
examples/gnrc_border_router: fix path to uhcpd in README


aabadie:
Welcome and thanks for your contribution !
The fix is good but the commit message should be more precise on what is fixed. Can you amend it with the following message: `examples/gnrc_border_router: fix path to uhcpd in README` ?

aabadie:
Closing in favor of #9273

PR #9274, opened 2018-06-02
periph_common/I2C: introduce new return codes scheme

dylad:
### Contribution description
This PR changes the API return codes in order to use errno.

I don't master doxygen mysteries so If there is better way don't hesitate to tell me.

### Issues/PRs references
#6577

### DIFF
--- a/drivers/include/periph/i2c.h
+++ b/drivers/include/periph/i2c.h
@@ -193,38 +193,6 @@ typedef enum {
 /** @} */
 
 /**
- * @brief   I2C transfer results
- */
-enum {
-    /**
-     * @brief   All bytes were transferred successfully
-     */
-    I2C_ACK = 0,
-    /**
-     * @brief   NACK when transferring the address byte
-     *
-     * After the address + the read/write bit were send, we got an NACK as
-     * response. This means most probably, that there is no slave with the used
-     * address on the bus, or the slave did just not respond for some reason.
-     */
-    I2C_ADDR_NACK = -1,
-    /**
-     * @brief   NACK while writing data bytes
-     *
-     * The slave responded to a data byte written to it with a NACK.
-     */
-    I2C_DATA_NACK = -2,
-     /**
-      * @brief   Internal error
-      *
-      * This status code is returned, on any other internal error that might
-      * have occurred. Possible reasons are not supported modes (e.g. 10-bit
-      * addressing).
-      */
-    I2C_ERR = -3
-};
-
-/**
  * @brief   Initialize the given I2C bus
  *
  * The given I2C device will be initialized with the parameters as specified in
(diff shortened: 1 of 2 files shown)


drivers/periph_common/i2c.c
```diff
@@ -24,23 +24,22 @@
 
 #ifdef I2C_NUMOF
 
-#ifdef PERIPH_I2C_NEED_READ_REG
+#ifdef PERIPH_I2C_NEED_READ_REGS
```
aabadie: Seems unrelated ?

drivers/include/periph/i2c.h
```diff
@@ -272,9 +240,13 @@ int i2c_release(i2c_t dev);
  * @param[out] data         memory location to store received data
  * @param[in]  flags        optional flags (see @ref i2c_flags_t)
  *
- * @return                  I2C_ACK on successful transfer of @p len byte
- * @return                  I2C_ADDR_NACK if response to address byte was NACK
- * @return                  I2C_ERR for any other error
+ * @return                  0 Success
+ * @return                  -EIO Slave device doesn't ACK the byte
```
aabadie: maybe add a `when` between error code and the rest. `-EIO when slave device doesn't ACK the byte`.

drivers/periph_common/i2c.c
dylad: Oops, will fix

drivers/include/periph/i2c.h
```diff
@@ -240,13 +240,13 @@ int i2c_release(i2c_t dev);
...
- * @return                  -EIO Slave device doesn't ACK the byte
- * @return                  -ENXIO No devices respond on the address sent on the bus
- * @return                  -ETIMEDOUT  Timeout occurs before device's response
- * @return                  -EINVAL Invalid argument given
- * @return                  -EOPNOTSUPP MCU driver doesn't support the flag operation
- * @return                  -EAGAIN  Lost bus arbitration
+ * @return                  0 When success
+ * @return                  -EIO When slave device doesn't ACK the byte
+ * @return                  -ENXIO When no devices respond on the address sent on the bus
+ * @return                  -ETIMEDOUT  When timeout occurs before device's response
```
aabadie: There's an extra space before `When` here and after -EAGAIN also. Squash directly

dylad:
@aabadie I've addressed your comments

dylad:
Squashed.
I'm wondering what we can do about ret code for acquire/release ?

drivers/periph_common/i2c.c
```diff
@@ -37,7 +37,7 @@ int i2c_read_regs(i2c_t dev, uint16_t addr, uint16_t reg,
                   void *data, size_t len, uint8_t flags)
 {
     /* First set ADDR and register with no stop */
-    int8_t ret = i2c_write_bytes(dev, addr, &reg, (flags & I2C_REG16) ? 2 : 1,
+    int ret = i2c_write_bytes(dev, addr, &reg, (flags & I2C_REG16) ? 2 : 1,
```
aabadie: The alignment with the second line is incorrect now. Same below

aabadie:
> I'm wondering what we can do about ret code for acquire/release ?

There's an EBUSY error code that would be useful with i2c_require. But I think it's fine like it is now (0 on success, -1 on error), no need for more.

drivers/periph_common/i2c.c
dylad: Should be good now

aabadie:
Now it would be great if someone from the @RIOT-OS/admin team could remove Murdock from the list of required status checks in the `new_i2c_if` branch protection settings. Because in the actual status of this branch, Murdock will always fail...

aabadie:
The green button is back, let's merge

PR #9256, opened 2018-05-31
WIP: Lua tests

danpetry:
### Contribution description

This adds tests for the Lua pkg.
This PR will include a cut down version of the Lua test suite, written in Lua, and a basic assert test to test the "hello world" output in examples/lua, as a framework for tests for further additions such as API bindings.

### Issues/PRs references
Depends on https://github.com/RIOT-OS/RIOT/pull/9153, because without it, a lot of the tests are likely to fail.
Depends on #9282 to properly integrate Lua source code.

PR #9282, opened 2018-06-04
tools/mkconstfs: Add an improved tool.

jcarrano:
### Background

I need to generate a constant file system to run the Lua test suite, and rather than roll my own, I decided to enhance the one already in RIOT.

The problems I found with the current script:

- It outputs to `stdout` only and does not buffer the whole output, meaning it can fail but still generate a partial output. When this is used inside a makefile, the result is that the first call to `make` fails but on the second call there is already a (partial) file up to date and `make` ignores the target.
- It crawls the file system, recursively including everything it finds in a directory tree. This is not flexible enough.
- Filename mangling is not robust enough (it can generate invalid identifier names).

### Contribution description

In order not to break code that may use the older script, this PR adds a new script with a different name.

- Writing to a file is now possible without having to resort to shell redirection.
- The output is buffered and written only after all processing is done, so there is no partial files and no timestamp changes in case of errors.
- ~Complete name mangling. It will only generate valid C identifiers.~ No more name mangling: no more problems.
- Files to be included should be given in the command line - no more crawling. There are tools to [find](https://linux.die.net/man/1/find) and filter files better than any we may ever make ourselves.
- Command line parsing is done by `argparse`.

**At some point I will extend the tool to embed files in other formats different from constfs.**

### Issues/PRs references

I need this for #9256.


dist/tools/mkconstfs/mkconstfs2.py
```diff
@@ -0,0 +1,160 @@
...
+         mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ) as bfile:  # noqa: E127
+        yield from map(lambda x: x[1],
+            itertools.chain.from_iterable(     # noqa: E128
+                map(lambda l: itertools.chain(l[1], [(0, "\n")]),
+                    itertools.groupby(
+                        enumerate(map(byte2s, bfile)),
+                        lambda x: x[0]//16)
+                )  # noqa: E127
+            )
+        )
```
MichelRottleuthner: wooha "the python foo is strong in this one" Not sure if this is more "nice and efficient" than it is "ugly and hard to understand/maintain"^^ I'm also wondering why you are mixing `yield from` and `itertools.chain`. I'm certain this can be written in a cleaner way, which might also help with getting rid of all the `#noqa ...` markers ;)

dist/tools/mkconstfs/mkconstfs2.py
```diff
@@ -0,0 +1,160 @@
...
+                        help="Where to mount the resulting fs", default="/")
+
+    parser.add_argument("-o", '--output', metavar="output_file",
+                        help="Write the output to a file instead of stdout."
+                             "The file is only written if the command is successful"
+                             "(i.e. there is no partial output")
+
+    parser.add_argument("-r", '--root', metavar="root_base_path",
+                        help="Paths on the constf will be gerated for the real"
+                        "path of the files by considering this path to be the root"
```
MichelRottleuthner: insert whitespaces at the end of the multiline strings above (i.e. after _root_, _real_, ..)

dist/tools/mkconstfs/mkconstfs2.py
```diff
@@ -0,0 +1,160 @@
...
+    parser.add_argument("-m", '--mount', metavar="mountpoint",
+                        help="Where to mount the resulting fs", default="/")
+
+    parser.add_argument("-o", '--output', metavar="output_file",
+                        help="Write the output to a file instead of stdout."
+                             "The file is only written if the command is successful"
+                             "(i.e. there is no partial output")
+
+    parser.add_argument("-r", '--root', metavar="root_base_path",
+                        help="Paths on the constf will be gerated for the real"
```
MichelRottleuthner: ge**ne**rated?

jcarrano:
@MichelRottleuthner I just realized this script will generate wrong paths if ran in a system that doesn't use UNIX paths. Is that a problem?

dist/tools/mkconstfs/mkconstfs2.py
```diff
@@ -0,0 +1,160 @@
...
+def main():
+    parser = argparse.ArgumentParser(
+            description="Embed files into a constant file system")
+
+    parser.add_argument("-m", '--mount', metavar="mountpoint",
+                        help="Where to mount the resulting fs", default="/")
+
+    parser.add_argument("-o", '--output', metavar="output_file",
+                        help="Write the output to a file instead of stdout."
+                             "The file is only written if the command is successful"
```
bergzand: If you use `type=argparse.FileType` here, argparse checks the file and mode for you automatically [1].

[1]: https://docs.python.org/3.5/library/argparse.html#filetype-objects

jcarrano: For some reason, I didn't like FileType, but that was in the past and I think I forgot the reasons.

jcarrano: Now I remember, it opens the file right away. Not a good idea with mode "w+".

MichelRottleuthner:
> @MichelRottleuthner I just realized this script will generate wrong paths if ran in a system that doesn't use UNIX paths. Is that a problem?

Not a real problem but I'd prefer to make it as independent from the platform as possible. Its python so everyone will assume it works wherever python is available.

jcarrano:
> Its python so everyone will assume it works wherever python is available.

I think the same, but I don't have a windows PC now, so I won't be able to test it.

MichelRottleuthner:
I have access to a windows machine here so I can help with testing if needed

jcarrano:
@MichelRottleuthner Could you test it?

MichelRottleuthner:
Tried it on a windows machine but it only seems to work if used with files that are not prefixed by a path (backslashes are not converted to slashes). I'll try to find time to look into a solution later today. 
Another thing that I found: you implemented filename mangling to ensure valid identifiers - would it make sense to also do this for the struct names like vfs_mount_t?

jcarrano:
@MichelRottleuthner Ouch! I knew I won't get the windows thing right.

There is not chance of collision with things like `vfs_mount_t` because the mangling always prefixes an underscore (that's a cheap fix to ensure identifiers never start with a number). I don't think it's reasonable to check the mangled name against all identifiers defined in the file (especially because we are generating identifiers). The options are either leave it as is (what are the chances of a collision?) or drop mangling entirely and generate identifiers of the form _fileXXXXX where XXXXX is some code that does not depend on the filename.

dist/tools/mkconstfs/mkconstfs2.py
```diff
@@ -0,0 +1,176 @@
...
+                                     mangled_name=mangle_name(relp))
+                for relp in sorted(filemap.values()))
+
+    yield "};\n"
+
+    yield C_FOOTER.format(constfs_name=constfs_name, mount_pount=mount_point)
+
+
+def addroot(fname):
+    return "/" + fname if not fname.startswith("/") else fname
```
bergzand: Is it possible to rework this to something using the python `os.path` module or the `pathlib` module to keep it OS agnostic. It might just be what is causing issues for @MichelRottleuthner

MichelRottleuthner:
>There is not chance of collision with things like vfs_mount_t

sorry, my wording wasn't clear enough. What I meant was the name of the struct **instance** which is put into the constfs_name field in the C_FOOTER template. Arguably one coud blame the user for putting invalid names there but i think it would help if there is some kind of warning (or mangling).

>I'll try to find time to look into a solution later today

I have to shift that to tomorrow...

dist/tools/mkconstfs/mkconstfs2.py
MichelRottleuthner: I think this won't help (for everything) because I also got backslashes in the middle of the path

bergzand: Ah, so do I understand correctly that the problem is that a path provided as `path\to\file` on windows shows up in the constfs as `/path\to\file`?

MichelRottleuthner: Yes

jcarrano: Ourch, that is not supposed to happen, but it's not because of `addroot`.

bergzand: Yeah, I misunderstood the initial issue. @jcarrano I think it might be solved by doing a recursive os.path.split() and `PosixPath.join()` on the supplied file

jcarrano:
@MichelRottleuthner I think it's out of scope to check if the user tells the tool to create a struct with an invalid name. Almost every time this tool is used it will be from within a makefile and the name will be fixed. The Correct(TM) way of generating code and making sure it's correct is to create an abstract tree and dump it as code. Templating things we will always run into some kind of corner case, and it's not practical to check everything. As long as we don't generate wrong code that still compiles and fails silently and weirdly, it's ok.

jcarrano:
BTW, what is the opinion on dropping name mangling and going for serially generated identifiers? AFAIK, GCC allows for arbitrarily long identifiers, but then we could have a file that it a hairy mess of characters.

dist/tools/mkconstfs/mkconstfs2.py
MichelRottleuthner: Just had another look. The problem actually occurs only if relative paths are used. It then not only leaves backslashes there but also creates some crazy paths like `/../C:\path\to/C:/path/to/file`. I played around with various os.path / pathlib methods and found the following fix to make it working on linux and windows in the same way:
```diff
--- a/dist/tools/mkconstfs/mkconstfs2.py
+++ b/dist/tools/mkconstfs/mkconstfs2.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import os
 import sys
 import argparse
 import pathlib
@@ -50,7 +51,8 @@ static const uint8_t {varname}[] = {{
 
 
 def _relpath_p(path, start):
-    return posixpath.relpath(path.as_posix(), start.as_posix())
+    return posixpath.relpath(pathlib.Path(os.path.abspath(path)).as_posix(),
+                             pathlib.Path(os.path.abspath(start)).as_posix())
```

jcarrano: @MichelRottleuthner Great! Can you push it to my branch, please?

MichelRottleuthner: just tried that but somehow it doesn't work. Did you allow edits for maintainers?

MichelRottleuthner:
since the user/dev will not use the mangled names in most cases anyway you could also use short numbered identifiers. This could also save some memory ;)

dist/tools/mkconstfs/mkconstfs2.py
jcarrano: It was enabled. If it still does not work, I can apply a patch.
screenshot_2018-06-29_11-40-18

MichelRottleuthner: nope still doesn't work just add it yourself, please

jcarrano:
@MichelRottleuthner Squashed and edited the commit msg to give you some credit.

jcarrano:
Oops, accidentally hit "close"!!!

PR #7107, opened 2017-05-30
drivers: Add support for KW41Z builtin transceiver

gebart:
This is the radio found in NXP Kinetis KW41Z, KW21Z. Only 802.15.4 mode
is implemented (KW41Z also supports BLE on the same transceiver).

The driver uses vendor supplied initialization code for the low level
XCVR hardware, these files were imported from ~~Kinetis Connectivity
Software version 1.0.2 (framework_5.3.2)~~ mcuxpresso.nxp.com (KSDK 2.2.0, framework_5.3.5)

The reason for using the vendor code is that setting up the XCVR module requires a lot of precalculated values which I don't have time to recreate. The vendor code works and was imported with minimal modifications to support easy updates if a new version comes out. 

Tested with two FRDM-KW41Z boards running gnrc_networking example

Depends on: ~~#6994~~ ~~#6995~~ ~~#6916~~ ~~#6978~~ ~~#7362~~ #7897

Issue #7543, opened 2017-08-30
core: thread_flags: THREAD_FLAG_MUTEX_UNLOCKED, THREAD_FLAG_TIMEOUT are not used


gebart:
Follow up to #7533. I found that there are two other predefined constants which are not used in the tree.

- `THREAD_FLAG_MUTEX_UNLOCKED`
- `THREAD_FLAG_TIMEOUT`

I think the constants may be intended for `xtimer_msg_timeout`, which mentions core_thread_flags in the documentation, but does not in fact use thread flags for the implementation.

kaspar030:
#7558 implements a simple timeout function using THREAD_FLAG_TIMEOUT, which I think will be used in many timeout variants of existing functions.

When adding THREAD_FLAG_MUTEX_UNLOCKED, I thought it might be useful to wait for multiple mutexes at the same time. IMO supporting this would mess up the mutex code, so I propose to remove the define.

gebart:
#7558 was merged, waiting for #7717 then this will be fixed.

PR #9284, opened 2018-06-04
core: Add THREAD_FLAG_IRQ_SERVICED

gebart:
### Contribution description

Introduce a common thread flag which can be used as an alternative to the traditional mutex method where the user application locks a mutex and an interrupt handler unlocks it. The thread flags method may have a lower runtime overhead than using a mutex.

The Kinetis I2C driver in #9262 uses this for signalling when the transfer is complete. The kw41zrf driver in #7107 uses a thread flag for signalling, and will be updated to use this definition to avoid collisions with application specific thread flags.

### Issues/PRs references

#9262 #7107

### DIFF
--- a/core/include/thread_flags.h
+++ b/core/include/thread_flags.h
@@ -62,7 +62,7 @@
 #include "sched.h"  /* for thread_t typedef */
 
 #ifdef __cplusplus
- extern "C" {
+extern "C" {
 #endif
 
 /**
@@ -96,13 +96,13 @@
  *          ...
  *      }
  */
-#define THREAD_FLAG_MSG_WAITING      (0x1<<15)
+#define THREAD_FLAG_MSG_WAITING     (1u << 15)
 /**
  * @brief Set by @ref xtimer_set_timeout_flag() when the timer expires
  *
  * @see xtimer_set_timeout_flag
  */
-#define THREAD_FLAG_TIMEOUT          (0x1<<14)
+#define THREAD_FLAG_TIMEOUT         (1u << 14)
 /** @} */
 
 /**
--- a/core/include/thread_flags.h
+++ b/core/include/thread_flags.h
@@ -40,7 +40,9 @@
  * of them, thread flags should be considered.
  *
  * Note that some flags (currently the three most significant bits) are used by
- * core functions and should not be set by the user. They can be waited for.
+ * core functions, system libraries, and device drivers and should not be set by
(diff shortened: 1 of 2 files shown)


jcarrano:
The docs say:

> Note that some flags (currently the three most significant bits) are used by core functions and should not be set by the user. They can be waited for.

But then it goes to define only two reserved flags. For this reason I'm marking it as an API change.

gebart:
@jcarrano 
- Clarified doc text
- Move THREAD_FLAG_EVENT from event.h to thread_flags.h, for better visibility

gebart:
the three in the text for reserved flags is probably a leftover since #7543

jcarrano:
When you are woken up by THREAD_FLAG_IRQ_SERVICED, how do you know the flag was set by the i2c IRQ and not by any other IRQ?

gebart:
Good question! There is a potential for collisions here, so developer needs to ensure that the ISR only sets the flag when a thread is actually expecting it. In the i2c case we only enable the IRQ from the i2c module when we perform a transfer, so the thread will be expecting a thread flag to become set. The i2c module isr will not even know which thread to send the flag to unless there is a thread that has claimed the bus (i2c_acquire). For the network device, we know that each gnrc_netif thread is only managing a single device, so each network device can set the flag on its assigned thread without risk, as long as the netif thread does not do other things as well.
Do you think we should split this into two or more flags? THREAD_FLAG_PERIPH_IRQ, THREAD_FLAG_NETDEV_IRQ

jcarrano:
> Do you think we should split this into two or more flags? THREAD_FLAG_PERIPH_IRQ, THREAD_FLAG_NETDEV_IRQ

Where's the limit? What if, say a SPI or UART device also sets THREAD_FLAG_PERIPH_IRQ? I understand that this shouldn't happen if all peripheral calls are blocking because then a single thread can only use one peripheral al the same time, at least for protocols where only the host can start a transmission/reception. For things like UART or a I2C in slave mode, one would have to make sure that the flag is not set by a RX IRQ, because a RX may happen at any time and there might be some unrelated code waiting for that flag.

Another option would be to use the flag as is and then check the actual cause, similar to how a shared interrupt line works in hardware.

gebart:
@jcarrano we lose the benefit of the lightweight thread flags if we multiplex events like this. What you suggest is similar to what sys/event does.

PR #9288, opened 2018-06-04
drivers/at: make OK and ERROR replies configurable

aabadie:
### Contribution description

This PR provides 2 new macro to the AT command parser. These macros allow to change the expected OK and ERROR replies of an AT module.

I have one that doesn't answer OK or ERROR as the driver expects by default but rather +OK and +ERR.

### Issues/PRs references

#7084

PR #9289, opened 2018-06-04
drivers/mkrwan: add initial implementation

aabadie:
### Contribution description

This PR provides an initial implementation for the LoRa module available on the arduino-mkrwan1300 board. This module is controlled using AT commands so the driver is based on the generic AT command parser available in RIOT.

This PR also provides a basic test application for configuring the module, joining a network and sending data to the network. Reception of messages from the network is not supported in the current state.

### Issues/PRs references

PR based on #8354 and ~#9288~


jia200x:
unfortunately I don't have this module :/. I can review the code though

PR #8948, opened 2018-04-13
doc: doxygen: add command

miri64:
### Contribution description
This introduces the `@experimental` command to Doxygen via Doxygen's alias feature (credits goes to [Sebastian Pipping](http://doxygen.10944.n7.nabble.com/experimental-command-td2869.html) for the idea, though I used a slightly different approach and utilized our inclusion of bootstrap into the generated doc). Also, I provided an example by marking Skald as experimental.

### Issues/PRs references
None

PR #9290, opened 2018-06-05
Add SRAM based PRNG seeder

PeterKietzmann:
### Contribution description
This PR provides an initial implementation for a PUF random seed generator based on SRAM start-up patterns. For now it is only implemented/tested on three (cortex M) based platforms: remote-reva, samr21-xpro and nucleo-f411re which behave similar. The SRAM memory provides approx. 4-6% entropy between multiple boots on **one** board and no correlation of cells at the same address between **different** boards (of the same kind). We use the DEK hash mechanism to extract entropy. Further analysis showed that the min. entropy of hashed values converges to approx. 31 Bit (in a uint32_t) with 1kB SRAM at the input of the hashing mechanism. This value should be used as seed for PRNGs.

Note 1: Please note that you need to power-off the device in order to get "fresh" and high-entropy seeds
Note 2: A proper test and evaluation application will follow soon

### References
[1] [A. Van Herrewege, V. van der Leest, et al., “Secure prng seeding on commercial off-the-shelf microcontrollers”](https://www.researchgate.net/publication/262280454_Secure_PRNG_seeding_on_commercial_off-the-shelf_microcontrollers)
[2] [A. Schaller, “Lightweight Protocols and Applications for Memory-Based Intrinsic Physically Unclonable Functions Found on Commercial Off-The-Shelf Devices”](http://tuprints.ulb.tu-darmstadt.de/7014/)


cpu/cortexm_common/ldscripts/cortexm_base.ld
```diff
@@ -113,6 +113,9 @@ SECTIONS
         KEEP (*(.isr_stack))
         . = ALIGN(8);
         _estack = .;
+        KEEP (*(.puf_stack))
+        . = ALIGN(8);
```
cgundogan: you seem to pad 8 bytes here in the linker, but only use 4 bytes later. Shouldn't we use rather `ALIGN(4)` then?

cpu/stm32f4/vectors.c
```diff
@@ -22,6 +22,24 @@
...
+/* 1kB SRAM considered for seeding */
+#define SEED_RAM_LEN     (1024 / sizeof(uint32_t))
+
+/*  Allocation of the PUF seed variable */
+__attribute__((used,section(".puf_stack"))) uint32_t global_puf_seed;
+
+/* SRAM memory marker defined in the linker script */
+extern uint8_t _sram;
+
+void pre_startup(void)
```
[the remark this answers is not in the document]
cgundogan: After a second thought I am actually fine with using `pre_startup()` as it is almost nowhere used. We can improve on where to put the code in follow-up PRs.

cgundogan:
I had an offline discussion with @PeterKietzmann and we came to the following conclusions:
- do not use a pseudo module, but rather a board feature
- move the (for now simply) call to the hash function into a proper API function that can be called from `pre_setup()` or from somewhere else in the future
- such a simple API would need an input pointer for the location to read from (`pre_startup()` would use the sram location) and an output pointer (e.g. the dedicated `.puf_stack` allocated on the cpu stack)

these are all minor things, but improve modularization. Once they are implemented, we are good to go.

tcschmidt:
> What happens when the board is not power cycled? E.g. When you press the reset button or do a software reboot?

The desired behavior, I suppose, is to keep the previous seed in memory until memory relaxes.

cpu/cortexm_common/ldscripts/cortexm_base.ld
```diff
@@ -113,6 +113,9 @@ SECTIONS
         KEEP (*(.isr_stack))
         . = ALIGN(8);
         _estack = .;
+        KEEP (*(.puf_stack))
```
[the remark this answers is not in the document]
PeterKietzmann: Basically yes. I didn't consider the heap section but seems like a good candidate. I'll figure it out.

cpu/stm32f4/vectors.c
```diff
@@ -22,6 +22,24 @@
...
+{
+    global_puf_seed = dek_hash(&_sram + ISR_STACKSIZE, SEED_RAM_LEN);
```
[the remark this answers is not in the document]
PeterKietzmann: With the proposed "modularization" (which I didn't push so far) it's just a parameter anyway. But a position liek _bss_start  should work either. Also here, I'll do some tests an come back to you.

PeterKietzmann:
@gebart thanks for the review. I'll investigate about the comments and come back to you. 
>What happens when the board is not power cycled? E.g. When you press the reset button or do a software reboot?

Then previously written memory blocks will be used and entropy might be lost. There is no way around fully powering off the memory in order to get a fresh seed. Keeping the initial seed in memory over a soft-reset seems like a good plan. I'll think of a way to differentiate between power-off and soft-reset. But this might be an additional feature I think.

PeterKietzmann:
@gebart would you mind having an other look at the code?

- instead of allocating a separate section on the stack I tried the`oneway-malloc` to use the heap. Unfortunately malloc is not usable at that point (does this depend on the initialization of std-c library which happens later?). So the section exists as before. @cgundogan proposed to get rid of it by keeping the seed value local and writing to a global variable (allocated on the default stack) just *after* RAM initialization. Personally I think the proposed solution in this PR is somewhat cleaner. Furthermore, I could imagine the separate section might help when seed generation isn't done after a soft reset
What do you think is best?
- I did not progress with the power-off detection (-> don't re-seed after soft reset) but I'd like to see this as a separate feature anyway. Further, I think this topic relates to low power states

gebart:
Thanks for the detailed reply. I agree with your arguments for a separate linker section for the seed value.
Did you notice a difference in startup ram values when doing the full power cycle reboot compared to when just pressing the hardware reset button on any of your test platforms? Or is the hardware reset just as good as a power cycle?

PeterKietzmann:
>which leads to the same pattern with soft reset.

To be a bit more precise here: With the provided test application, a button or software reset always leads to the same seed. This is because the used SRAM (starting at `&_srelocate`) is initialized and doesn't change it's state. The state of the memory surely depends on the application that used this memory section before reset.

gebart:
@PeterKietzmann it seems like a potential vulnerability that ram keeps its content between resets. .bss and .data are cleared during early boot, but maybe we should clear the heap as well?

PeterKietzmann:
> but maybe we should clear the heap as well?

Might be a good idea, but let's address this separately.

@gebart what do you think about the following approach to detect a software reset: RIOT always creates a main and an idle thread. During creation, the thread stack is initialized and a [`STACK_MARKER`](https://github.com/RIOT-OS/RIOT/blob/master/cpu/cortexm_common/thread_arch.c#L139) is written. In the PUF startup we could check for this sequence. If it exists we assume a soft reset and use the old seed (which we read from the newly introduced `puf_stack` section). It the marker is not present, we perform the PUF mechanism.

gebart:
@PeterKietzmann that sounds like a good approach :+1:

MrKevinWeiss:
@PeterKietzmann I will run the tests throughout the day then post the results (it will take some time).  I will vary a few boards and a few parameters then would like someone to double check that the values seem reasonable.

MrKevinWeiss:
@PeterKietzmann I have finished testing, does [this](https://gist.github.com/MrKevinWeiss/7f6bdb4e45c40daef0cb88ecd2fecd11) seem OK.  I also don't know if you want any improvement ideas on the example test?  If you look at the test summary you should see what has been run.

tcschmidt:
Many thanks @MrKevinWeiss - it looks like 1024 bit input is about o.k., but 2048 is more reliable. Your choice @PeterKietzmann ?

PeterKietzmann:
I've increased the default RAM length and immediately squashed because it improves entropy of the 32-bit word slightly. This memory is reused anyway so it does not even cost resources (just some cycles during startup). Furthermore, the parameter is compile-time configurable anyway.

PeterKietzmann:
FYI: I fixed a couple of style issues that Travis/Murdock complained about, and immediately squashed.

MichelRottleuthner:
I ran the following tests with nucleo-l073rz and samr21-xpro:
https://gist.github.com/MichelRottleuthner/e0b28f6809ecd20d03f35fc8beb4253b
(will add results of another run with 2048 bytes but thats still running)

For the nucleo I just adapted the corresponding `vectors.c` analog to this PR.

Probably I'm missing something on how the linker and pre_startup code works together, but is there a reason behind using `_srelocate` instead of e.g. `_sram` ? 
As this happens before RAM init, so shouldn't the start address of RAM do?

>(...) Furthermore, the parameter is compile-time configurable anyway.

Checking against the actual RAM size (i.e. available PUF size) would be nice then (IMO).

>If it exists we assume a soft reset and use the old seed (...)

Wouldn't it make sense in this case to alter the seed in some way (hash or whatever)? Or somehow explicitly indicate that no new entropy can be pulled from the PUF at this moment? 
If it's done like you described above, you would indeed save the time to re-seed but you would still have the same seed over multiple (soft-)resets - which I suspect should be avoided. 

Just out of interest, have you tried if forcing the system to low-power mode with RAM-retention disabled will allow this PUF mechanism to work?

tcschmidt:
> Wouldn't it make sense in this case to alter the seed in some way (hash or whatever)? Or somehow explicitly indicate that no new entropy can be pulled from the PUF at this moment?
If it's done like you described above, you would indeed save the time to re-seed but you would still have the same seed over multiple (soft-)resets - which I suspect should be avoided.

Deterministic hashing (or some other deterministic function) does not add entropy. So applying this would just alter the seed in a fully predictable way.

What should be the meaning of it? (Could be a convenience function such as 'node reset produces a new sequence of a PRNG' - not sure whether this is helpful?)

MichelRottleuthner:
>Deterministic hashing (or some other deterministic function) does not add entropy. So applying this would just alter the seed in a fully predictable way.

Didn't want to imply that (see "...Or somehow explicitly indicate that no new entropy can be pulled from the PUF at this moment") -  just wanted to point out that it may be not a good thing to start with the same seed over an over again.

tcschmidt:
Understood. Still it seems unclear to me what the proper utility is: providing an entropy-input when available or some seed variation. Note if hashed, the original seed cannot be recovered. One could use an invertible function to enable both old and new seeding. (Probably difficult to explain, though).

Printing out a warning at boot time would be fine, I suppose.

sys/include/puf_sram.h
```diff
@@ -0,0 +1,57 @@
...
+#define SEED_RAM_LEN     (2048 / sizeof(uint32_t))
+#endif
+
+/**
+ * @brief Global seed variable, allocated on the stack in puf_sram.c
+ */
+extern uint32_t global_puf_seed;
+
+/**
+ * @brief builds hash from @p SEED_RAM_LEN bytes uninitialized SRAM, writs it
```
MichelRottleuthner: writ**e**s (side note: the doc isn't aligned as nicely as above here)

MichelRottleuthner:
>...use the latest seed state of the PNRG...

Sounds reasonable. IMO this should be done in a higher layer then ([random.c](https://github.com/PeterKietzmann/RIOT/blob/7ddc04aa143b6437ea67a422fa3310e090729a62/sys/random/random.c) in this case ?) because the PUF implementation won't (and shouldn't ?) have the state of whichever PRNG was used before. 
That's why I think it would make sense to also pass the info regarding 'truly fresh entropy' vs. 'reused' to the upper layer - so you know there what to expect from that value and decide what to do with it.

tcschmidt:
Agreed. So @PeterKietzmann has the problem of recovering that state at the right place.

tcschmidt:
> Re-using the last state of the PRNG is something that I also considered. Probably this would be relatively easy to get if we make the PRNG use the allocated puf_stack section for its state.

O.K., this should then move to the PRNG side.

> I agree that it might make sense to indicate (via API) how a seed was generated. However, the seed generation function is not meant for users. With the current infrastructure it would still be the random API that needs a parameter for indication. Yet, I don't want the change this API so I propose to go with a warning as well as proper documentation at first

Yes, I don't think that this is an API thing, since we are on the best effort base anyway. What reasonable answer could we give - a hidden mechanism did apply or not? 
We are not guarantying anything, we just try to act less stupid than is commonly the case.

Are we then ready to cleanup and finish?

PeterKietzmann:
Well, we've been talking about reset strategies but yet I did not even finish the reset detection. Either we go with this PR as is (including minor cleanups) or it takes some more time to implement (and test!) the reset detection.

tcschmidt:
> Well, we've been talking about reset strategies but yet I did not even finish the reset detection. Either we go with this PR as is (including minor cleanups) or it takes some more time to implement (and test!) the reset detection.

I guess, reset detection should be finished. This was checking the STACK_MARKER , right?

PeterKietzmann:
Yes, that was the basic idea.

PeterKietzmann:
I've added code that detects whether the board/chip starts from power down or reset (e.g., software or button) by looking for a known pattern in the RAM. Yet, the samd21 has been adapted and tested -> this works. If you reset via button, a warning is printed to the console. 

@MrKevinWeiss or @MichelRottleuthner could of you give it a try?

@gebart would you mind to review my latest commits please?  Especially I'd like to ask for your opinion in two concerns:
-  [Here](https://github.com/RIOT-OS/RIOT/pull/9290/files#diff-b11bb5d26de334dd34c10756ab3a00daR39) the idea is to call `puf_sram_softreset` with the location of `STACK_MARKER` on the idle thread stack. What should be done is basically the same as the thread implementation does: checking for misalignments and adjusting thread parameters accordingly. However, I did not want to duplicate too much. Do you have an idea how to better integrate this?
 -  [This change](https://github.com/RIOT-OS/RIOT/pull/9290/files#diff-6cb70140a7abc7aa72fd43bea7cf6bf8) was obviously needed to get the start address of the idle thread stack. Could you think of an alternative?

@tcschmidt I did **not** implement re-seeding with latest PRNG states because this requires changes in different PRNG implementations. IMO this PR is complex enough and I'd like to propose the alternative in a follow up PR, if we really want that. According to the first proposal, the initial "high entropy seed" is always used in case of soft resets.

tcschmidt:
> I did not implement re-seeding with latest PRNG states because this requires changes in different PRNG implementations. IMO this PR is complex enough and I'd like to propose the alternative in a follow up PR, if we really want that. According to the first proposal, the initial "high entropy seed" is always used in case of soft resets.

Agreed - I thought we already had consensus that this seed update is a matter of the PRNGs.

MichelRottleuthner:
did a quick test of the soft-reset detection on nucleo-l073rz and nucleo-l476rg: works as described on both

tcschmidt:
@gebart waiting for your feedback: can you allocate a few minutes?

sys/puf_sram/puf_sram.c
```diff
@@ -0,0 +1,48 @@
...
+     * SRAM pattern*/
+    global_puf_state = 0;
+    global_puf_seed = dek_hash(ram, SEED_RAM_LEN);
+    return global_puf_seed;
+}
+
+bool puf_sram_softreset(const uint8_t *marker_addr, const uint8_t *marker, size_t len)
+{
+    for (size_t i = 0; i < len; i++) {
+        if (marker_addr[i] != marker[i]) {
```
gebart: Suggestion, replace this loop by memcmp

cpu/cortexm_common/include/cpu_conf_common.h
```diff
@@ -68,6 +68,30 @@ extern "C" {
...
+ */
+#define STACK_MARKER                (0x77777777)
+/** @} */
+
+/**
+ * @brief   Initial program status register value for a newly created thread
+ *
+ * In the initial state, only the Thumb mode-bit is set
+ */
+#define INITIAL_XPSR                (0x01000000)
```
gebart: Does this need to be public?

cpu/cortexm_common/include/cpu_conf_common.h
```diff
@@ -68,6 +68,30 @@ extern "C" {
...
+/** @} */
+
+/**
+ * @brief   ARM Cortex-M specific exception return value, that triggers the
+ *          return to the task mode stack pointer
+ */
+#define EXCEPT_RET_TASK_MODE        (0xfffffffd)
```
gebart: Does this need to be public?

cpu/cortexm_common/ldscripts/cortexm_base.ld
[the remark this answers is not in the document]
gebart: If you place the puf section first there is less risk that it will be relocated when flashing new firmwares of an application. This reduces the risk of false negatives for the soft reset detection during OTA upgrades.

core/kernel_init.c
```diff
@@ -75,7 +75,7 @@ const char *main_name = "main";
 const char *idle_name = "idle";
 
 static char main_stack[THREAD_STACKSIZE_MAIN];
-static char idle_stack[THREAD_STACKSIZE_IDLE];
+char idle_stack[THREAD_STACKSIZE_IDLE];
```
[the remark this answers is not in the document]
PeterKietzmann: > Maybe we can solve the alignment issue by using gcc attributes, or a union between char array and uint64_t.

Can you elaborate?

> Do you have to use a marker at the end of the stack area or could you just place it at the beginning of the stack?

I just take what's there. For me it doesn't matter where it is located

sys/puf_sram/puf_sram.c
PeterKietzmann: This function is executed before libc has been initialized, that's why I took the simple loop

PeterKietzmann:
@gebart thanks for the review, again.

> Why do we need to use the idle stack at all to detect soft reset?
Can we insert another constant inside the puf section instead? 

Basically I intended to avoid further memory overhead. However, to be honest, I did not consider introducing my own marker. This could simplify several occasions of this PR. Will investigate over the day and report then.

PeterKietzmann:
@gebart please have a look at my latest commits. Among others, I implemented the "soft reset detection" by using a new constant inside the puf section. This removes ugly dependencies to previous core changes. 

We still have one (IMO minor) problem with the detection mechanism. After flashing software that includes `puf_sram`, for the first time after running any other app, we will not be able to detect the "soft reset" by looking at the concerned memory position. A new seed is generated based on shitty SRAM patters, without giving a warning. However, **in any case** we need **at least** one initial boot from power-down in order to generate a high entropy seed. That's why I decided for a simple compile-time warning [see here](https://github.com/RIOT-OS/RIOT/pull/9290/files#diff-4d60783d06bbf3b30a80e8cca8b25ca9R1). 

What do you think?

PeterKietzmann:
Alternatively I was thinking about introducing a 'generic' RIOT marker somewhere at the beginning of the RAM. Probably other components might benefit from that, although I have nothing specific in mind.

kaspar030:
> we will not be able to detect the "soft reset" by looking at the concerned memory position.
> What do you think?

How about leaving this question open for now? Most MCU's provide a way to figure out whether the last reboot was from PoR, wakeup from deep sleep, BOD. I'd suggest to find a simple method that works for maybe one board, abstract that away as much as possible (```bool boot_was_coldstart(void);```) so puf_sram is usable on that board, then find a more generic way later on.

PeterKietzmann:
@kaspar030 thanks for your feedback!

> How about leaving this question open for now?

Fine with me - if needed.

> Most MCU's provide a way to figure out whether the last reboot was from PoR, wakeup from deep sleep, BOD

I have no experience with that but I'll try somewhen today. However, this does not really improve the current situation for two reasons: (i) IMO it makes most sense to evaluate the memory in order to rate it's state. (ii) We will be able to detect **no** cold start after initial flashing **but** we need a power-down anyway.

> I'd suggest to find a simple method that works for maybe one board ... then find a more generic way later on

I would say the current approach is the most generic way. However, if it turns out that PoR detection or whatever is preferred, we cans till think about keeping this implementation hardware specific and  over-writable à la `HAVE_COLDSTART_DETECTION`

@gebart opinions?

tcschmidt:
The key point (and beauty) of the approach of @PeterKietzmann is that generic mechanisms are used as well as hardware properties that are widely available. We want to keep it that way.

PeterKietzmann:
The last commit presents how a specific, MCU-status based reboot detection could look like (implemented for the samd21). It's not really complicated but also not necessary. I propose to go without it and improve in follow-ups (if actually needed)

tcschmidt:
Yup @PeterKietzmann - keep the PUF modules independent of MCU add ons. If really needed, this function could be added by a simple MCU-wrapper lib.

cpu/samd21/Makefile.features
```diff
@@ -1 +1,4 @@
+FEATURES_PROVIDED += puf_sram
+CFLAGS += -DHAVE_REBOOT_DETECTION=1
```
gebart: This would be better located in some CPU header, e.g. periph_cpu.h or cpu_conf.h

cpu/samd21/vectors.c
```diff
@@ -22,6 +22,31 @@
...
+#ifdef MODULE_PUF_SRAM
+#include "puf_sram.h"
+
+/* SRAM memory marker defined in the linker script */
+extern uint32_t _srelocate;
+
+bool puf_sram_softreset(void)
+{
+    /*  read "Power On Reset" Bit in RCAUSE register */
+    if(PM->RCAUSE.reg & PM_RCAUSE_POR) {
```
gebart: Does this filter out hardware reset via reset button/signal and other non software originated resets?

sys/include/puf_sram.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+#define PUF_SRAM_MARKER  (0xad3021ff)
+
+/**
+ * @brief Global seed variable, allocated on the stack in puf_sram.c
+ */
+extern uint32_t puf_sram_seed;
+
+/**
+ * @brief Global seed state, allocated on the stack in puf_sram.c
```
gebart: Add some doc on how to interpret this variable. 0 means x, 1 means Y...

sys/include/puf_sram.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+extern uint32_t puf_sram_state;
+
+/**
+ * @brief builds hash from @p SEED_RAM_LEN bytes uninitialized SRAM, writs it
```
gebart: Typo writes

sys/include/puf_sram.h
```diff
@@ -0,0 +1,78 @@
...
+ *        to the global variable @p puf_sram_seed and returns the value
+ *
+ * @param[in] ram pointer to SRAM memory
+ *
+ * @return  a random number
+ */
+uint32_t puf_sram_uint32(const uint8_t *ram);
```
gebart: It seems unintuitive that this function takes a pointer, but not a length. I suggest making either both the pointer and the length into globals, or both as arguments.
The name is a bit misleading, when comparing with random_uint32 which returns a random uint32_t value. How about puf_sram_generate()?

sys/include/puf_sram.h
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * @brief checks for a memory marker determine whether memory contains old data.
+          Otherwise it assumes a reboot from power down mode
+ *
+ * @return    0 when preceding power down phase is expected
```
gebart: I don't understand this sentence
How about "0 when reset with power cycle was detected" to match the next line below

cpu/samd21/Makefile.features
PeterKietzmann: Well, I did not investigate much but defining in the header does not protect the fallback implementation in puf_sram.c (relates to build order?!)

cpu/samd21/vectors.c
PeterKietzmann: From SAMD21 datasheet sect. 7.4.1: *POR monitors VDDANA. It is always activated and monitors voltage at startup and also during all the sleep modes. If VDDANA goes below the threshold voltage, the entire chip is reset.*
The bit is set if a POR occurs.

PeterKietzmann:
@gebart thanks for your review, again. The last commit was meant as a showcase and after thinking about it, @tcschmidt and I decided to go with the previous solution. So I'd just remove that  commit. D'accord? I'll address the other comments asap

PeterKietzmann:
@gebart I addressed your comments and removed my previous commit including reset detection by SAMD21 MCU. If you agree I would start cleaning up commits, so we can converge with this PR somewhat soon.

sys/include/puf_sram.h
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * @brief SRAM marker to detect reboot without power-off
+ *
+ * Source: https://www.random.org/bytes/
+ */
+#define PUF_SRAM_MARKER  (0xad3021ff)
+
+/**
+ * @brief Global seed variable, allocated on the stack in puf_sram.c
```
gebart: Allocated on the stack is incorrect, it is a normal global variable, but allocated outside of the normal .data and .bss sections

sys/Makefile.include
```diff
@@ -89,3 +89,9 @@ endif
 ifneq (native,$(BOARD))
   INCLUDES += -I$(RIOTBASE)/sys/libc/include
 endif
+
+ifneq (,$(filter puf_sram,$(USEMODULE)))
+  # allocate space for 3 x 32 Bit integers: seed, state, marker
+  export PUF_SRAM_LEN ?= 12
+  export LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_puf_sram_len=$(PUF_SRAM_LEN)
```
gebart: Is this necessary? The linker should not need to be told how much space the puf section needs.

sys/random/random.c
```diff
@@ -23,16 +23,26 @@
...
     uint32_t seed;
-#ifdef MODULE_PERIPH_CPUID
+#ifdef MODULE_PUF_SRAM
+    /* TODO: hand state to application? */
+    if (puf_sram_state) {
+        LOG_WARNING("random: PUF SEED not fresh\n");
+    }
+    seed = puf_sram_seed;
+#endif
+#if !defined (MODULE_PUF_SRAM) && defined (MODULE_PERIPH_CPUID)
```
gebart:
```c
#elif defined(MODULE_PERIPH_CPUID)
```

sys/random/random.c
```diff
@@ -23,16 +23,26 @@
...
     luid_get(&seed, 4);
-#else
+#elif !defined (MODULE_PUF_SRAM) && !defined (MODULE_PERIPH_CPUID)
```
gebart:
```c
#else
```

PeterKietzmann:
@gebart addressed your comments. Ready to clean-up commits?

PeterKietzmann:
@gebart squashed and removed the `extern _pstack` in *vectors_cortexm.c*. I kept it in the linker because it looks cleaner to me and might be used some time in future.

cpu/stm32f4/vectors.c
```diff
@@ -22,6 +22,19 @@
 
 #include "vectors_cortexm.h"
 
+#ifdef MODULE_PUF_SRAM
+#include "puf_sram.h"
+
+/* SRAM memory marker defined in the linker script */
+extern uint32_t _srelocate;
+
+void pre_startup(void) {
```
kaspar030: please don't use ```pre_startup()```. Some platforms need that for pre-initialization (kinetis, nrf51).
For puf_sram, IMO it is valid to add an ```#ifdef MODULE_PUF_SRAM``` where needed, e.g., for cortexm, in cpu/cortexm/vectors_cortexm.c just before or after the call to ```pre_startup()```.

kaspar030: OK.

Then there is the issue of those twelve duplicated lines... Can you factor them out?

PeterKietzmann: I could make one function call from:

```
if (!puf_sram_softreset()) {
    puf_sram_generate((uint8_t *)&_srelocate, SEED_RAM_LEN);
}
```
but the rest won't change -> I don't see the benefit. Furthermore, in the current state the RAM start address and length are similar for the three MCUs , but this might vary for different boards.

kaspar030: How much do you expect the implementations differ?

If you move all of the current (three times duplicated) ```pre_startup()``` into ```sys/puf_sram/puf_sram.c``` and rename it to ```puf_sram_init()```, the duplication is removed. You can even move ```extern uint32_t _srelocate;``` into the function. Call that function from the cortexm reset_vector_default(). That would probably work on *all* cortexm boards. Just to be sure, it makes sense to add the feature to FEATURES_PROVIDED one-by-one, on a case-by-case basis.

cgundogan: >If you move all of the current (three times duplicated) pre_startup() into sys/puf_sram/puf_sram.c and rename it to puf_sram_init(), the duplication is removed. 

@kaspar030 can we postpone a proper refactoring to a follow-up? I am not very convinced of putting all above lines into a generic `puf_sram_init()` function and expect it to work for all cortex platforms. Besides that, there's no experience with other platforms (avr, mips, ...) yet and there might be other parameters to consider for `puf_sram_init()`. I propose that we first think about the API and which abstractions we need in order to make this module platform-independent and *then* do the proper refactoring. Would that be okay? We should also mark this module in the meantime as `@experimental` (see #8948)

kaspar030: > I am not very convinced of putting all above lines into a generic puf_sram_init() function and expect it to work for all cortex platforms. 

It doesn't have to work on more than the three mentioned platforms where it is made available through ```FEATURES_PROVIDED```.

> Besides that, there's no experience with other platforms (avr, mips, ...) yet and there might be other parameters to consider for puf_sram_init(). 

Irrelevant for now. If ```puf_sram_init()```  needs to be adapted for other architectures, we can introduce changes or multiple versions *then*.

> I propose that we first think about the API and which abstractions we need in order to make this module platform-independent and then do the proper refactoring. Would that be okay? 

IMO it is OK to merge half-experimental code into the main tree, given the promise that it gets fixed quickly. But the experimental part then has to be contained as much as possible.

Here that is not the case, 3*12 duplicate lines are introduced in cpu/, for *no reason*. Just copying the given ```pre_startup()``` into one file would lead to identical code without the duplication. That *cannot* be too much to ask to fulfill point 2.1. of the reviewing guidelines.

PeterKietzmann:
@kaspar030 I've addressed your comment and directly amended the changes to previous commits. Furthermore, I've re-tested on a samr21-xpro with similar results.

cpu/cortexm_common/ldscripts/cortexm_base.ld
```diff
@@ -108,6 +108,9 @@ SECTIONS
     /* stack section */
     .stack (NOLOAD):
     {
+        KEEP (*(.puf_stack))
+        . = ALIGN(8);
+        _pstack = .;
```
gebart: I don't see why we need to introduce this. There's no clear use case and the two consecutive aligns are redundant so this will always be equivalent to _sstack

PeterKietzmann:
@gebart okay. Removed it and immediately squashed.

cpu/cortexm_common/ldscripts/cortexm_base.ld
```diff
@@ -108,6 +108,8 @@ SECTIONS
     /* stack section */
     .stack (NOLOAD):
     {
+        KEEP (*(.puf_stack))
+        . = ALIGN(8);
```
gebart: No need to align twice

PeterKietzmann: Addressed. Damn, when you do things in a hurry ...

cpu/cortexm_common/ldscripts/cortexm_base.ld
```diff
@@ -108,6 +108,7 @@ SECTIONS
     /* stack section */
     .stack (NOLOAD):
     {
+        KEEP (*(.puf_stack))
```
kaspar030: any reason this is called "puf_stack" and not just "puf"?

sys/puf_sram/puf_sram.c
```diff
@@ -0,0 +1,55 @@
...
+    puf_sram_seed = dek_hash(ram, len);
+    /* write marker to a defined section for subsequent reset detection */
+    puf_sram_marker = PUF_SRAM_MARKER;
+    /* seting state to 0 means seed was generated from SRAM pattern */
+    puf_sram_state = 0;
+}
+
+bool puf_sram_softreset(void)
+{
+    if(puf_sram_marker != PUF_SRAM_MARKER){
```
kaspar030: two spaces missing here, feel free to fix.

cpu/stm32f4/Makefile.features
```diff
@@ -1,4 +1,5 @@
 FEATURES_PROVIDED += periph_hwrng
+FEATURES_PROVIDED += puf_sram
```
aabadie: Why only stm32f4 and not all STMs ? (same question for samd21, why not all sam0?)

PeterKietzmann: @aabadie in order to use start-up patterns of SRAM memory as entropy source, the properties need to be evaluated. Yet, I only did this with a samr21-xpro, nucleo-f411re and re-mote. The plan is to add more boards consecutively.

cgundogan:
Murdock also acknowledges. Ready to merge => GO

cpu/stm32f4/Makefile.features
aabadie: > The plan is to add more boards consecutively.

The problem with this approach is that f4 is not only nucleo-f411re but all boards with a stm32f4. Are they also evaluated ?
Maybe I'm wrong but I would move this in a more central place first, `cortexm_common/Makefile.features` sounds more appropriate, since it's a cortexm only feature.

PeterKietzmann: @aabadie I don't quite understand. 

>  f4 is not only nucleo-f411re but all boards with a stm32f4

You want to reduce the scope for that feature to nucleo-f411re?

> I would move this in a more central place first, cortexm_common/Makefile.features sounds more appropriate

You want to centralise and extend the feature to all Cortex-M platforms?

The assumption is, that the STM32F4 MCU has same SRAM properties on different boards. I just tried with the mentioned Nucleo as well as the STM32F4Disco board. Most likely, the other STM32FXYZ will behaves similar, but I did not yet evaluate this.

Currently this feature only works for Cortex-M platforms as you might have seen in the hook in *cortexm_common/vectors_cortexm.c*. But as mentioned, it is planned to extend this to other architectures.

aabadie:
> You want to reduce the scope for that feature to nucleo-f411re?

No :)

> Currently this feature only works for Cortex-M platforms as you might have seen in the hook in cortexm_common/vectors_cortexm.c.

Yes and that's what why I don't understand why only samd21, cc2538 and stm32f4 only provide the feature when all cortexm could benefit from it.

PeterKietzmann:
@aabadie because we don't know if the memory properties of all cortex boards suffice -> will do. By nature the PUF module is not cortex specific. It is just the code location that matters, which needs to be before startup (we implemented it in `pre_startup()` initially which lead to code duplication)

PR #9291, opened 2018-06-05
tools/ethos: adapt start_network.sh for macOS

smlng:
### Contribution description

This adapts the `start_network.sh` script to run on macOS by using only standard tools, and avoiding GNU tools such as `readlink`. Though ethos still needs to be fixed for OSX, anyways, see #6718.

### Issues/PRs references

- #4764 
- #6718

### DIFF
--- a/dist/tools/ethos/start_network.sh
+++ b/dist/tools/ethos/start_network.sh
@@ -1,6 +1,18 @@
 #!/bin/sh
 
-ETHOS_DIR="$(dirname $(readlink -f $0))"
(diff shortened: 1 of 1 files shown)


kYc0o:
Using the common tools, a.k.a. tuntap library for OSX, doesn't allow to use a tun/tap interface if it's not being opened at the user level e.g. by ethos.

Thus, IMHO, a separated script makes more sense to me, as I proposed in #6718. Can you verify this?

smlng:
RIOT native is broken for MacOS bc missing 64bit support for x86_64 CPUs in RIOT; hence, closing this one

PR #9292, opened 2018-06-05
tools/doccheck: adapt script to run on macOS

smlng:
### Contribution description

adapt `doccheck/check.sh` script to run on macOS by avoiding GNU tools, i.e., `readlink`, which are not present on non-Linux systems by default.

### DIFF
--- a/dist/tools/doccheck/check.sh
+++ b/dist/tools/doccheck/check.sh
@@ -8,7 +8,7 @@
 # General Public License v2.1. See the file LICENSE in the top level
 # directory for more details.
 
-RIOTBASE=$(readlink -f "$(dirname $(realpath $0))/../../..")
+RIOTBASE="$(cd $(dirname $0)/../../..; pwd)"
 
 if tput colors &> /dev/null && [ $(tput colors) -ge 8 ]; then
     CERROR="\e[1;31m"


dist/tools/doccheck/check.sh
```diff
@@ -6,17 +6,16 @@
 # General Public License v2.1. See the file LICENSE in the top level
 # directory for more details.
 
-RIOTBASE=$(readlink -f "$(dirname $(realpath $0))/../../..")
+RIOTBASE="$(cd $(dirname $0)/../../..; pwd)"
```
cladmi: Would just using `realpath -Le` work on mac ? and even remove the second `realpath`.
It would also resolve symlinks.

dist/tools/doccheck/check.sh
```diff
@@ -6,17 +6,16 @@
...
 
 ERRORS=$(make -C "${RIOTBASE}" doc 2>&1 | \
             grep '.*warning' | \
             sed "s#${PWD}/\([^:]*\)#\1#g")
 
-if [ -n "${ERRORS}" ]
-then
+[ -n "${ERRORS}" ] && {
```
cladmi: I personally prefer the `if` version.

smlng: mhm, I found many of this (non-explicit-if) variants in our scripts ...

But we agree that the `else` part is not needed because of the `exit 2` right?

dist/tools/doccheck/check.sh
smlng: `realpath` is a GNU tools as well, so not available by default on macOS (or FreeBSD). You're right it would resolve symlinks, but does it matter here and provide benefits over simply taking the resulting path?

dist/tools/doccheck/check.sh
cladmi: Yes

dist/tools/doccheck/check.sh
smlng: to me this is the most portable version

dist/tools/doccheck/check.sh
smlng: this rendered obsolete with recent changes to the script by other PR(s)

PR #9294, opened 2018-06-05
tools: allow static checks to be run from any directory

smlng:
### Contribution description

most static check script assume to be run from `RIOTBASE`, this PR changes this and allows
to execute the scripts from any (sub)directory.

### DIFF
--- a/dist/tools/cppcheck/check.sh
+++ b/dist/tools/cppcheck/check.sh
@@ -8,7 +8,10 @@
 # General Public License v2.1. See the file LICENSE in the top level
 # directory for more details.
 
-: "${RIOTTOOLS:=${PWD}/dist/tools}"
+: "${RIOTBASE:=$(cd $(dirname $0)/../../../; pwd)}"
+cd $RIOTBASE
+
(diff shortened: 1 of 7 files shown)


dist/tools/cppcheck/check.sh
```diff
@@ -8,7 +8,10 @@
 # General Public License v2.1. See the file LICENSE in the top level
 # directory for more details.
 
-: "${RIOTTOOLS:=${PWD}/dist/tools}"
+: "${RIOTBASE:=$(cd $(dirname $0)/../../../; pwd)}"
+cd $RIOTBASE
```
aabadie: Why do you need to change the current directory to RIOTBASE ?

smlng: because right now the script assumes to be run from RIOTBASE from the start. As per title, this PR allows the script to be called from anywhere, but that makes it necessary to change to RIOTBASE if not already run from there.

aabadie:
@cladmi do you have comments or are you ok to merge ?

PR #9296, opened 2018-06-05
cpu/lpc2387: Various fixes for GPIO driver

maribu:
- Fixed documentation
- Use bitwise operation instead of multiplication and addition in `GPIO_PIN()`
- Allow GPIOs to be configured as input via `gpio_init()`
- Fixed bugs in `gpio_init_mux`:
    - `0x01 << ((pin & 31) * 2)` was used before to generate the bitmask, but this would shift by 62 to the left. Correct is `0x01 << ((pin & 15) * 2)` (See [datasheet](https://www.nxp.com/docs/en/user-guide/UM10211.pdf) at pages 156ff)
    - Only one of the two bits was cleared previously
- Changed strategy to access GPIO pins:
    - Previous strategy:
        - Set all bits in FIOMASK except the one for the pin to control to disable access to them
        - Set/clear/read all pins in the target GPIO port (but access to all but the target pin is ignored because of the applied FIOMASK)
    - New strategy:
        - Set/clear/read only the target pin
    - Advantages:
        - Only one access to a GPIO register instead of two
        - Proven approach: Access to GPIOs on lpc2387 is mostly done by accessing the GPIO registers directy (e.g. see the sht11 driver). Those accesses never touch the FIOMASK register
        - No unwanted side effects: Disabling all but one pin in a GPIO port without undoing that seems not to be a good idea

### DIFF
--- a/cpu/lpc2387/include/periph_cpu.h
+++ b/cpu/lpc2387/include/periph_cpu.h
@@ -39,12 +39,22 @@ extern "C" {
  * @brief Fast GPIO register definition struct
  */
 typedef struct {
-    __IO uint32_t DIR;      /**< he */
-    uint32_t _reserved[3];  /**< really */
-    __IO uint32_t MASK;     /**< wants */
-    __IO uint32_t PIN;      /**< to */
-    __IO uint32_t SET;      /**< know */
-    __IO uint32_t CLR;      /**< everything */
+    /** @brief Direction: Output if corresponding bit is set, otherwise input */
+    __IO uint32_t DIR;
+    /** @brief 12 bytes of reseved memory we don't need to access */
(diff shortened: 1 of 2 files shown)


maribu:
Rebased against current RIOT/master and requested changed applied. @A-Paul: Thanks for your review. Out of curiosity: What board will you use for testing?

A-Paul:
@maribu, the only board with *lpc2387* I could find in our pile of hardware is the *MSB-A2*:
https://github.com/RIOT-OS/RIOT/wiki/Board%3A-MSBA2

maribu:
Thanks for reviewing. Nice to know that there are still other MSB-A2 boards out there :-)

PR #9298, opened 2018-06-06
drivers/hts221: adapt to i2c api return codes

aabadie:
### Contribution description

This PR readapts the hts221 driver to the I2C return codes (< 0 in case of an error, 0 on success).

### Issues/PRs references

Follow-up of #9195 and related to #6577

### DIFF
--- a/drivers/hts221/hts221.c
+++ b/drivers/hts221/hts221.c
@@ -50,7 +50,7 @@ int _humidity_calibration(hts221_t *dev)
     uint8_t buf[2];
     uint8_t reg = HTS221_REGS_H0_RH_X2 | 0x80;
     /* 1. read h0_rh and h1_rh coefficients */
-    if (i2c_read_regs(BUS, ADDR, reg, &buf[0], 2, 0) != 2) {
+    if (i2c_read_regs(BUS, ADDR, reg, &buf[0], 2, 0) < 0) {
         i2c_release(BUS);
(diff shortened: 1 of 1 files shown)


aabadie:
> why acquire/release are used several times in the init function ?

I looked at the code more in depth, and this is perfectly fine in term of workflow: it's just that some private functions are not acquiring the bus before performing their read/write operations.

> here is an output with a nucleo-l073:

```
2018-06-21 11:25:57,327 - INFO # Connect to serial port /dev/ttyACM0
Welcome to pyterm!
Type '/exit' to exit.
2018-06-21 11:25:58,801 - INFO # H: 56.4%, T: 24.8°C
2018-06-21 11:26:00,805 - INFO # H: 56.1%, T: 24.8°C
2018-06-21 11:26:02,803 - INFO # H: 56.0%, T: 24.8°C
2018-06-21 11:26:04,807 - INFO # H: 55.9%, T: 24.8°C
```

dylad:
> I looked at the code more in depth, and this is perfectly fine in term of workflow: it's just that some private functions are not acquiring the bus before performing their read/write operations.

Sure, but this is first time I saw this. Is there any gain from doing this ? I don't have strong opinion about this. If the driver was already in this state before, this is not our rights to change it with the I2C refactoring.
Anyway, it would be great if someone could test it, otherwise I can leave an untested ACK.

aabadie:
> it would be great if someone could test it

I think @smlng has the hardware, since he initially wrote this driver. Maybe he can comment on your other questions.

PR #9299, opened 2018-06-06
drivers/lsm6dsl: adapt the i2c api return codes

aabadie:
### Contribution description

This PR readapts the lsm6dsl driver to the I2C return codes (< 0 in case of an error, 0 on success).

### Issues/PRs references

Follow-up of #9193 and related to #6577


aabadie:
> What do you think ?

Each function is supposed to return 0 on success, so after adding all return codes, the result must be 0. If one of the function a negative error code, then the final result is negative as well, meaning there's an issue.
I'll check if the doxygen documentation is inline with that.

PR #9306, opened 2018-06-07
tests: test for prio inversion using msg_send_recv

haukepetersen:
### Contribution description
#7445 also introduces priority inheritance for `msg_send_receive` calls. This PR adds a test showcasing priority inversion for `msg_send_receive`, cutting it from #7445 for simpler reviewing...

### Issues/PRs references
Cut out from #7445 and similar to #7444

### DIFF
--- a/tests/thread_priority_inversion_msg/README.md
+++ b/tests/thread_priority_inversion_msg/README.md
@@ -1,0 +1,44 @@
+NOTE
+====
+THIS TEST WILL FAIL WHEN RUNNING RIOT USING THE DEFAULT CORE/KERNEL CONFIGURATION!
+
+TODO: merge the `core_priority_inheritance` module to enable priority
+      inheritance and with this successfully pass this test.
+
+Expected result
+===============
+When successful, you will see the 3 configured threads printing 6 different
+events in a defined order. If the test passes, the output should look like this:
+
+```
+main(): This is RIOT! (Version: xxx)
+Test for showing priority inversion when using msg_send_receive
+
+If this tests succeeds, you should see 6 events appearing in order.
+The expected output should look like this:
+Event  1:      t3 - waiting for incoming message
+Event  2:      t2 - starting infinite loop, potentially starving others
+Event  3:      t1 - sending msg to t3 (msg_send_receive)
+Event  4:      t3 - received message
+Event  5:      t3 - sending reply
+Event  6:      t1 - received reply
+
+TEST OUTPUT:
+Event  1:      t3 - waiting for incoming message
+Event  2:      t2 - starting infinite loop, potentially starving others
+Event  3:      t1 - sending msg to t3 (msg_send_receive)
+Event  4:      t3 - received message
+Event  5:      t3 - sending reply
+Event  6:      t1 - received reply
+
+   *** result: SUCCESS ***
+```
+
+Background
+==========
+Priority inversion is a known problem in real-time systems, leading in certain
+constellations to lower priority threads indirectly blocking higher priority
+threads. This test constructs a simple scenario, where a high priority thread
+(`t1`) is trying to `send_receive` a message to a low priority thread (`t3`),
+but never gets a reply as a medium priority thread (`t2`) is never letting `t3`
+to be scheduled.
--- a/tests/thread_priority_inversion_msg/main.c
+++ b/tests/thread_priority_inversion_msg/main.c
@@ -1,0 +1,143 @@
+/*
+ * Copyright (C) 2017 Technische Universität Berlin
+ *               2017,2018 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     tests
+ * @{
+ *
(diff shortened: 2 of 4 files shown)


tests/thread_priority_inversion_msg/README.md
```diff
@@ -0,0 +1,44 @@
...
+Test for showing priority inversion when using msg_send_receive
+
+If this tests succeeds, you should see 6 events appearing in order.
+The expected output should look like this:
+Event  1:      t3 - waiting for incoming message
+Event  2:      t2 - starting infinite loop, potentially starving others
+Event  3:      t1 - sending msg to t3 (msg_send_receive)
+Event  4:      t3 - received message
+Event  5:      t3 - sending reply
+Event  6:      t1 - received reply
```
danpetry: I reckon this is starvation, rather than priority inversion. Reason is that the medium priority thread prevents the low thread from running even before the message is sent, rather than pre-empting it while it's handling the message. Priority inversion would involve the second one - i.e. indirect pre-emption of a higher priority thread.
An example displaying priority inversion with messages would involve the high thread sending the message to the low thread, and the low thread receiving it and doing some action before sending. But while that action is happening, the medium thread comes along and pre-empts the low thread.

tests/thread_priority_inversion_msg/main.c
```diff
@@ -0,0 +1,143 @@
...
+         "Event  1:      t3 - waiting for incoming message\n"
+         "Event  2:      t2 - starting infinite loop, potentially starving others\n"
+         "Event  3:      t1 - sending msg to t3 (msg_send_receive)\n"
+         "Event  4:      t3 - received message\n"
+         "Event  5:      t3 - sending reply\n"
+         "Event  6:      t1 - received reply\n");
+    puts("TEST OUTPUT:");
+
+    /* create threads */
+    for (unsigned i = 0; i < T_NUMOF; i++) {
```
danpetry: IMO separating out three "high", "medium" and "low" named threads was nicer for readability.

tests/thread_priority_inversion_msg/main.c
```diff
@@ -0,0 +1,143 @@
...
+    msg_receive(&m);
+    event(4, "t3", "received message");
+
+    event(5, "t3", "sending reply");
+    msg_reply(&m, &rply);
+
+    return NULL;
+}
+
+static thread_task_func_t _handlers[] = { t1, t2, t3 };
```
danpetry: This might fit better at the top, next to _names? Or at least the two together?

danpetry:
NB I've differentiated between "review changes" and "suggestions". I.e. between saying, "this should be changed", and just giving feedback, to be changed if desired. Is this clear in the way I've done it?

tests/thread_priority_inversion_msg/main.c
haukepetersen: its a tradeoff, having it here means that I was able to skip the forward declaration of the thread functions...

tests/thread_priority_inversion_msg/main.c
haukepetersen: fixed.

tests/thread_priority_inversion_msg/README.md
haukepetersen: yes, will fix.

haukepetersen:
fixed style issues, looking into the starvation vs inversion problem next

haukepetersen:
changed the test slightly, should now actually test for prio inversion instead of prio starvation, right? In the end IMHO it does not really matter, as both versions break without any counter measures, and both versions are fixed with priority inheritance enabled, so they should actually test for very similar behavior.

tests/thread_priority_inversion_msg/main.c
danpetry: Ah. yep

danpetry:
Great, thanks very much. Would you please be able to change the README to match?
After that, I have no more comments. What do you think about keeping this and https://github.com/RIOT-OS/RIOT/pull/7444 open until https://github.com/RIOT-OS/RIOT/pull/7445 is merged? Then merge immediately after? Or some other related plan?

tests/thread_priority_inversion_msg/README.md
```diff
@@ -0,0 +1,44 @@
...
+===============
+When successful, you will see the 3 configured threads printing 6 different
+events in a defined order. If the test passes, the output should look like this:
+
+```
+main(): This is RIOT! (Version: xxx)
+Test for showing priority inversion when using msg_send_receive
+
+If this tests succeeds, you should see 6 events appearing in order.
+The expected output should look like this:
```
danpetry: Could you please change the readme to reflect the recent changes in the test?

haukepetersen: of course

haukepetersen:
fixed the README and squashed.

haukepetersen:
> What do you think about keeping this and #7444 open until #7445 is merged? Then merge immediately after? Or some other related plan?

Sounds like a plan.

haukepetersen:
fixed board blacklisting and squashed.

haukepetersen:
fixed one more little pep issue in the testrunner script

mguetschow:

PR #9307, opened 2018-06-07
WIP RFC Replace usage of USEPKG by USEMODULE.

cladmi:
### Contribution description

This addresses the inconsistent use of USEPKG mentioned in #6668.
Currently USEPKG is required to tell the build system to take `pkg/PKG_NAME` into account.

It was argued that it makes the user aware he is using a package as he is required to put USEPKG, but packages are being pulled as dependencies for some cpu. And also, some packages, like `lwip` are implemented by several modules and doing `USEMODULE = lwip_ipv6` triggers the inclusion of the package transparently.

This PR tries to implement the proposed solution

> I propose making USEMODULE += implicitly add to USEPKG, iff pkg/ exists.

The build system modification is done around these changes:

``` diff
diff --git a/Makefile.dep b/Makefile.dep
-# include package dependencies
--include $(USEPKG:%=$(RIOTPKG)/%/Makefile.dep)
+# process package dependencies
+-include $(ALL_PKGS:%=$(RIOTPKG)/%/Makefile.dep)

 # process dependencies
+# deduce packages from USEMODULE
+USEPKG := $(filter $(ALL_PKGS), $(USEMODULE))
```

and fixing packages so `pkg/*/Makefile.dep` can be included even if the module is not required. Like done in the normal Makefile.dep anyway.

This goes in the direction of making packages self contained like #9003 

## API CHANGE

It requires all packages `Makefile.dep` to be protected by the `ifneq` pattern.
It also require to remove all `USEPKG` in applications (which should be asserted right now).

## Implementation

The commits have been designed to be incremental and always be in a working state. But this was not tested.
I count on murdock to help finding if there are issues and fixing them in the right place.

## TODO

* [ ] Get maintainers feedback on the solution
* [ ] Check if the updated documentation is enough ! ! !
* [ ] Move all easy packages dependencies to `pkg/*/Makefile.dep` (this could be done in subsequent PRs also)
* [ ] check if if really all packages dependencies can be safely moved or document them

### Issues/PRs references

pkg: USEPKG vs USEMODULE #6668


cladmi:
The incremental changes also allow to possibly stop on a previous commit if preferred, or split this PR.

cladmi:
I also remember there are some packages pre-cleanup included that I can do in sub PRs.

Makefile.include
```diff
@@ -250,9 +250,22 @@ ifneq ($(GNRC_NETIF_NUMOF),1)
...
 endif
 
+# list of all RIOT packages
+ALL_PKGS := $(sort $(notdir $(patsubst %/Makefile,%,$(wildcard $(RIOTPKG)/*/Makefile))))
 # process dependencies
 include $(RIOTBASE)/Makefile.dep
 
+# Detect if an application or Makefile.dep defined USEPKG
+ifneq (,$(USEPKG))
+  ifeq ($(origin USEPKG), file)
```
cladmi: I currently need this because USEPKG is exported in `vars.inc.mk` and it seems already set before executing make in murdock.

I could remove the export in `vars.inc.mk` maybe, but I still do not get why it has a value when passing through here for the moment (I did not investigate).

cladmi:
I fixed the broken builds, remaining task is to move packages dependencies out of the main Makefile.dep.

cladmi:
Still remaining dependencies not migrated to `pkg/*/Makefile.dep`
They may be done later too.

* lwip%
  ```
  ifneq (,$(filter ieee802154 nrfmin,$(USEMODULE)))
    ...
    ifneq (,$(filter lwip%, $(USEMODULE)))
      USEMODULE += lwip_sixlowpan
    endif
  endif
  ```
* nordic_softdevice_ble removing modules
  ```
  ifneq (,$(filter nordic_softdevice_ble,$(USEMODULE)))
    # prevent application from being a router
    # TODO: maybe find a nicer solution in future build system update
    _ROUTER_MODULES = gnrc_ipv6_router% gnrc_rpl netstats_rpl auto_init_gnrc_rpl
    ifneq (,$(filter $(_ROUTER_MODULES),$(USEMODULE)))
      $(warning nordic_softdevice_ble: Disabling router modules:\
        $(filter $(_ROUTER_MODULES),$(USEMODULE)))
    endif
    USEMODULE := $(filter-out $(_ROUTER_MODULES),$(USEMODULE))
  endif
  ```
* openthread dependencies which are now all defined in the test Makefile

gebart:
There's a weird lwip failure in the Murdock logs. Maybe needs a rebase?

cladmi:
I moved the `lwip` dependencies to `pkg/lwip/Makefile.dep` and re-order them so it could also be my fault.

cladmi:
Looks still broken, I will investigate. I expected this as some of the dependencies definition are order specific.
If I do not manage easily, I will keep the problematic `lwip` ones in the main `Makefile.dep`

cladmi:
I moved the "default implementation" selection at the end of `pkg/lwip/Makefile.dep` and it should be fixed with this.

cladmi:
Will need to rebase due to #9884

smlng:
@cladmi how is this coming? I think this is a nice cleanup, if you need help on testing: I'm in.

cladmi:
Thanks for the update. It got somehow lost in my todo list, and was waiting for the "Get maintainers feedback on the solution" and "Check if the updated documentation is enough ! ! !" parts so did not put more work on it recently.

If you like this current solution, or tell me what I should change, I can start splitting parts.

Makefile.include
cladmi: There is currently a usage in `sys/shell/commands` that should be updated to use `USEMODULE`. I will try to go early in that direction.

cladmi:
I did a blind stupid rebase on master. May not be working right now.

From the first version I did, I would like to change the handling to do the dependencies first to allow deprecating incompatible things earlier without breaking anything.

I see some steps to go forward with this in parallel:

* [ ] cleanups
* [ ] Declare `ALL_PKGS`
* [ ] Always add `USEPKG` to `USEMODULE`
  * [ ] change the usages of `$(USEPKG)` to `$(USEMODULE)`.
    * [ ] do not export `USEPKG` anymore
  * [ ] Change the packages `Makefile.dep` to be always with `ifneq(,$(filter packagename,$(USEMODULE))` and add a sanity check for it (this will allow deprecating it and not break external code immediately).
    * [ ] With this, the packages `Makefile.dep` could always be parsed without side-effect, like done for all the others `Makefile.dep` right now.

PR #9311, opened 2018-06-08
tests/periph_i2c: python interface for shell

MrKevinWeiss:
### Contribution description
This is a python interface for the shell commands of the periph_i2c test.  It should make it easier to program tests that use the shell.  Eventually the test_shell_if.py can be move out and reused for other periphs.  This is part of the I2C embargo.

### Issues/PRs references
See #6577
Depends on #9168


MrKevinWeiss:
Closed in favor of #9409.

Issue #4881, opened 2016-02-23
Is -fno-builtin required?


gebart:
Currently we build cortexm_common, msba2, arduino-mega2560 and x86-multiboot-common with 
`CFLAGS += -fno-builtin`
which eliminates the compiler builtins for many common C library functions such as memcpy, memset, printf etc. Without -fno-builtin the compiler may optimize calls to standard library functions into more efficient calls, for example, `printf("hello\n")` (with no format arguments) could be optimized into `puts("hello")`.

https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html contains a list of some of the builtins.

I don't know the reason why the flag was added in the first place, possibly it may have interfered with some own implementation of C library functions.

I tried building without the flag on mulle and at first glance it seems to work. It even reduced the size of the gnrc_networking binary by about 2 kB!

See also: http://sjhan5.blogspot.se/2015/09/when-people-use-compile-option-fno.html

Kijewski:
For x86-multiboot-common GCC assumed that it could use some internal glibc functions when compiled without `-fno-builtin` that were not provided by Newlib (but it could be that my configuration was just wrong / less than optimal). Out of my head I don't recall which function that were.

I don't know if the other platforms you mentioned have (/ have had) the same problem.

gebart:
I would like to remove it, but I don't have time to work on it.

PR #9316, opened 2018-06-08
cortexm: Remove -fno-builtin from CFLAGS

gebart:
### Contribution description

Remove `-fno-builtin` from CFLAGS for Cortex-M targets. This allows the optimizer to optimize calls to printf, memcpy etc. printf is sometimes optimized into a call to puts or putchar. Loops may be optimized into calls to memcpy, memset and so on.

Testing: Unit tests pass on frdm-kw41z with this applied, examples/default works.
ROM usage for .text was reduced by 8 to 108 bytes for examples/default on the Cortex-M boards. 

### Issues/PRs references

#4881

### DIFF
--- a/makefiles/arch/cortexm.inc.mk
+++ b/makefiles/arch/cortexm.inc.mk
@@ -15,7 +15,7 @@ ifneq (,$(filter cortex-m0%,$(CPU_ARCH)))
 endif
 endif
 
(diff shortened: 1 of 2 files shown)


cladmi:
Anything I should do to test this ? (I will split the gnrc_netif fix)

I can do a size compare with a master build.

I could also set the reference Issue as a tracking issue to remove the option for the other boards.

gebart:
@keestux This may be interesting in relation to #10728 #10729

PR #9317, opened 2018-06-08
drivers/sht11: Major refactoring

maribu:
- Renamed driver to sht1x to reflect that fact SHT10, SHT11 and SHT15 are supported
- Removed sensor specific shell commands in favor of SAUL
- Major refactoring and new features:
    - Use RIOT's GPIO interface to access the sensor to increase portability
    - Changed API to allow more than one sensor per board
    - Added `sht1x_params.h` that specifies how the sensors is connected - each board can overwrite default settings by #defining SHT1X_PARAM_CLK and SHT1X_PARAM_DATA
    - Changed arithmetic to use integer calculations only instead of floating point arithmetic
    - Added support for checking the CRC sum
    - Allow optional skipping of the CRC check to speed up measuring
    - Added support for advanced features like reducing the resolution and skipping calibration to speed up measuring
    - Allow specifying the supply voltage of sensor which heavily influences the temperature result (and use that information to calculate the correct temperature)
    - Reset sensor on initialization to bring it in a well known state
    - Support for the obscure heater feature. (Can be useful to check the temperature sensor?)
- Confirmed new integer only arithmetic by unit tests
- Added SAUL integration

Hint: Before testing on the MSBA2 or other lpc2387 based boards, please apply [this PR](https://github.com/RIOT-OS/RIOT/pull/9296) containing various bugfixes for the gpio driver.


drivers/sht1x/sht1x.c
```diff
@@ -0,0 +1,665 @@
...
+    dev->conf &= SHT1X_CONF_MASK;
+    dev->conf |= conf & (~(SHT1X_CONF_MASK));
+
+    /* Send new status byte to sensor, if on-device config was changed */
+    if ((conf & SHT1X_CONF_MASK) != (dev->conf & SHT1X_CONF_MASK)) {
+        int retval = transmission_start(dev);
+        if (retval != 0)
+            return retval;
+
+        switch (write_byte(dev, SHT1X_STATUS_REG_W)){
```
aabadie: Maybe this file could be uncrustified ? (a space is missing before `{` here in a lot of other places).

drivers/sht1x/sht1x_saul.c
```diff
@@ -0,0 +1,157 @@
...
+    .type = SAUL_SENSE_TEMP
+};
+
+const saul_driver_t sht1x_saul_hum_driver = {
+    .read = read_hum,
+    .write = saul_notsup,
+    .type = SAUL_SENSE_HUM
+};
+
+const saul_driver_t sht1x_saul_conf_driver = {
```
aabadie: Why do you want to map the device configuration with SAUL ?

sys/auto_init/auto_init.c
```diff
@@ -20,8 +20,8 @@
 
 #include "auto_init.h"
 
-#ifdef MODULE_SHT11
-#include "sht11.h"
+#ifdef MODULE_SHT1X
```
aabadie: This can be removed in fact.

sys/auto_init/saul/auto_init_sht1x.c
```diff
@@ -0,0 +1,104 @@
...
+                sht1x_error(i, "?");
+                continue;
+        }
+
+        saul_entries[(i * 3)    ].dev = &(sht1x_devs[i]);
+        saul_entries[(i * 3) + 1].dev = &(sht1x_devs[i]);
+        saul_entries[(i * 3) + 2].dev = &(sht1x_devs[i]);
+        saul_entries[(i * 3)    ].name = "SHT1X temperature";
+        saul_entries[(i * 3) + 1].name = "SHT1X humidity";
+        saul_entries[(i * 3) + 2].name = "SHT1X config (Low resolution | "
```
aabadie: I think the calibration should be passed in the driver params.

sys/shell/commands/Makefile
```diff
@@ -8,9 +8,6 @@ endif
 ifneq (,$(filter ps,$(USEMODULE)))
   SRC += sc_ps.c
 endif
-ifneq (,$(filter sht11,$(USEMODULE)))
```
aabadie: I would keep the shell commands

tests/driver_sht1x/Makefile
```diff
@@ -0,0 +1,6 @@
+BOARD ?= native
+include ../Makefile.tests_common
+
+USEMODULE += sht1x
```
aabadie: Same comment about the version of sensor, better use the pseudomodule which will implicitly load the common concrete module.

tests/driver_sht1x/main.c
```diff
@@ -0,0 +1,140 @@
...
+        /* Testing for temperatures in -10.00°C and 65.00°C in steps of 0.13°C */
+        for (int16_t temp = -1000; temp < 6500; temp += 13){
+            for (uint16_t raw_hum = 0; raw_hum <= max_raw_hum; raw_hum++) {
+                int16_t exp_hum = expected_hum(&dev, raw_hum, temp);
+                if ((exp_hum < 0) || (exp_hum > 10000)){
+                    /* Result out of range, ignore it */
+                    continue;
+                }
+                int16_t got_hum = sht1x_humidity(&dev, raw_hum, temp);
+                if (
```
aabadie: This file needs to be uncrustified

tests/driver_sht1x/main.c
```diff
@@ -0,0 +1,140 @@
...
+        SHT1X_VDD_5_0V, SHT1X_VDD_4_0V, SHT1X_VDD_3_5V, SHT1X_VDD_3_0V,
+        SHT1X_VDD_2_5V,
+    };
+    const uint8_t confs[] = {SHT1X_CONF_LOW_RESOLUTION, 0};
+    const uint16_t max_raw_temps[] = {0xfff, 0x3fff};
+    const uint16_t max_raw_hums[] = {0xff, 0xfff};
+    const char *res_temp[] = {"12", "14"};
+    const char *res_hum[] = {"8", "12"};
+    const char *voltages[] = {"5.0", "4.0", "3.5", "3.0", "2.5"};
+    sht1x_dev_t dev = {.conf = 0};
```
aabadie: Why not use the `sht1x_init` with the default params ? This is the common usage for other sensors.

maribu: Because this would require an actual sensor to be present, but the unit tests are intended to run on "native" without any hardware. The unit tests verify that the integer arithmetic still gives the same results as the floating point arithmetic given in the data sheet. Because these tests are computationally expensive, it makes no sense to run them on embedded hardware.

Also, using default parameters would dramatically decrease test coverage. The unit tests iterate over all possible configuration states in the for loops and verify that the calculations are correct for all of them - including the configuration that was chosen to be the default one.

tests/driver_sht1x/Makefile
maribu: Sure the device specific pseudo modules should be used for unit tests that are not device specific? I would prefer to keep sht1x here, otherwise it is not obvious that is applies to all variants.

drivers/include/sht1x.h
```diff
@@ -0,0 +1,187 @@
...
+ * @defgroup    drivers_sht1x SHT10/SHT1X/SHT15 Humidity and Temperature Sensors
+ * @ingroup     drivers_sensors
+ * @brief       Driver for Sensirion SHT10/SHT1/SHT15 Humidity and Temperature
+                Sensors
+ * @{
+ *
+ * @file
+ * @brief       SHT1X Device Driver
+ *
+ * @author      Freie Universität Berlin, Computer Systems & Telematics
```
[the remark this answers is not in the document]
maribu: Yes, I could remove the author tag and keep them in the copyright declaration. The author tag does not provide additional information compared to the copyright declaration.

maribu:
@aabadie: Thanks for your review. Unless I have overlooked anything, I have addressed all issues you pointed out with the exception of the issues for the unit tests I commented on.

Please review again and give feedback if the two points with the unit tests can remain as they are for the reasons I stated in the comments above, or whether I should address them somehow.

Kind regards,
Marian

sys/auto_init/saul/auto_init_sht1x_saul.c
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @name    Import SAUL endpoints
+ * @{
+ */
+extern const saul_driver_t sht1x_saul_temp_driver;
+extern const saul_driver_t sht1x_saul_hum_driver;
+/** @} */
+
+void auto_init_sht1x_saul(void)
```
aabadie: No need for a saul specific function, you could add it to `auto_init_sht1x` but guard it with an `#ifdef` if the saul module is loaded. If you do this, `auto_init_sht1x_saul.c` can be removed. And the `auto_init_sht1x.c` could be with the other saul devices auto init files.

sys/shell/commands/sc_sht1x.c
```diff
@@ -0,0 +1,382 @@
...
+{
+    printf("[sht1x] Operation failed: %s\n", msg);
+}
+
+static int read_sensor(int16_t *temp, int16_t *hum, int argc, char **argv)
+{
+    const sht1x_dev_t *dev = get_dev(argc, argv);
+
+    if (!dev) {
+        return -1;
```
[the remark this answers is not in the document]
maribu: The error message is printed by `get_dev()` :-)

maribu:
@aabadie: Thanks for the review and for pointing me to the unit tests app. This is much better now :-)

drivers/include/sht1x.h
```diff
@@ -0,0 +1,186 @@
...
+ */
+int16_t sht1x_temperature(const sht1x_dev_t *dev, uint16_t raw);
+
+/**
+ * @brief             Calculate the relative humidity from the raw input
+ * @note              This internal funciton is exposed for unit tests
+ *
+ * @param dev         Device from which the raw value was received
+ * @param raw         The raw (unprocessed) temperature value
+ * @param temp        The temperate at which the humidity was measure in E-02 °C
```
aabadie: typo 'temperature'

sys/auto_init/auto_init.c
```diff
@@ -20,8 +20,8 @@
...
+void auto_init_sht1x(void);
```
aabadie: No need to declare this function here. Just use `extern` before the call to the function.

sys/auto_init/auto_init.c
```diff
@@ -101,9 +101,9 @@ void auto_init(void)
     DEBUG("Auto init xtimer module.\n");
     xtimer_init();
 #endif
-#ifdef MODULE_SHT11
-    DEBUG("Auto init SHT11 module.\n");
-    sht11_init();
+#ifdef MODULE_SHT1X
```
aabadie: I would move this just before the `#ifdef MODULE_AUTO_INIT_SAUL`.

maribu:
Unless I have overlooked anything, I have address all open issues and also added an on-device test for the driver. This test will read both temperature and humidity value three times for each configuration and then drop to the shell for manual testing. I added the MSBA2 as default board, because I tested on that. Please note that the driver does only work on the MSBA2 after the gpio driver bugs addressed by [this PR](https://github.com/RIOT-OS/RIOT/pull/9296) are fixed.

tests/driver_sht1x/main.c
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright (C) 2018 2018 Otto-von-Guericke-Universität Magdeburg
```
aabadie: 2018 appears twice

drivers/sht1x/sht1x.c
```diff
@@ -0,0 +1,694 @@
...
+        case 1:
+            return -EPROTO;
+    }
+
+    retval = read_byte(dev, status, SHT1X_ACK);
+    if (retval != 0) {
+        return retval;
+    }
+
+    {
```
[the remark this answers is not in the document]
maribu: This limits the validity of the variables `crc` and `expected` to that scope.

drivers/sht11/sht11.c
```diff
@@ -1,350 +0,0 @@
-/*
```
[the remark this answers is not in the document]
maribu: When looking at the [specific commit](https://github.com/RIOT-OS/RIOT/pull/9317/commits/c0903e509602395604635e5895020bf4c57e8dc1) the rename seemed to work, but in the overview of all changes it still looks like one file removed and another created

maribu:
In case you wonder why I put thumbs up on every comment: I in fact missed some and would like to tick them with a check mark. As there is no check mark, the thumbs  up is what comes nearest to it. So it is just for me to check that I actually applied every comment

sys/auto_init/saul/auto_init_sht1x.c
maribu: This was not actually calibrating the sensor - the calibration data is stored inside the sensor. But the sensor can be configured to not use them, which speeds up measurements about 10 ms. So this should be a runtime option, as this trade-off between speed and precision depends on the situation.

But the configuration was removed from SAUL anyway (as suggested), so that comment no longer applies.

maribu:
Done :-)

makefiles/pseudomodules.inc.mk
```diff
@@ -105,6 +106,10 @@ PSEUDOMODULES += si7013
 PSEUDOMODULES += si7020
 PSEUDOMODULES += si7021
 
+# include variants of SX127X drivers as pseudo modules
```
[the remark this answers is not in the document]
aabadie: Ok, I understand. This block should be moved to its original place (before the si1145 pseudo modules). Normally only changes related to sht11/10/15 should appear.

drivers/sht1x/include/sht1x_defines.h
```diff
@@ -0,0 +1,72 @@
...
+ * @name Possible values to pass as `ack` parameter to `write_byte`
+ * @see write_byte
+ * @{
+ */
+#define SHT1X_NO_ACK           (1)      /**< don't ack read in `read_byte` */
+#define SHT1X_ACK              (0)      /**< do acknowledge read in `read_byte` */
+/** @} */
+
+/**
+ * @name Commands that can be send to the SHT1X driver
```
aabadie: s/send/sen**t**/

aabadie:
Just triggered Murdock, maybe some other things will raise.

maribu:
Done. I believe I still have to add some boards to BOARD_INSUFFICIENT_MEMORY for the test, but lets wait for Murdock to complete

aabadie:
Murdock is still failing, I think you have to remove the `DIRS = drivers` in `boards/common/msb-430/Makefile`

maribu:
The default pin configuration is invalid on telosb, wsn430-v1_3b, wsn430-v1_4 and z1. How should I handle that?

aabadie:
> The default pin configuration is invalid on telosb, wsn430-v1_3b, wsn430-v1_4 and z1. How should I handle that?

No idea, this is related to 16bit platform but the error is weird. Maybe @cladmi has an idea ?

maribu:
Hi,

`cpu/msp430fxyz/include/periph_cpu.h` contains:
``` C
typedef uint16_t gpio_t;
[...]
#define GPIO_PIN(x, y)      ((gpio_t)(((x & 0xff) << 8) | (1 << (y & 0xff))))
```

so for `y > 15` the result of `(1 << (y & 0xff))` is (as far as I know) not strictly defined, as `a << b` is only defined for `0 <= b < BITWIDTH(a)`. Changing that line to 

```
#define GPIO_PIN(x, y)      ((gpio_t)(((x & 0xff) << 8) | (1 << (y & 0x0f))))
```

would make the result valid again. But maybe it is desired that the compilation fails for invalid pin specifications?

PS: I'm checking that right now with a dummy commit added on top of the PR. So please don't merge until I remove that again.

maribu:
@aabadie: I just rebased the PR against current master (and removed the testing commit). The issue with compilation on msp430fxyz boards [is now fixed](https://github.com/RIOT-OS/RIOT/pull/9367) :-)

PR #9318, opened 2018-06-08
core/runstate: track system running state

ZetaR60:
### Contribution description

This is a rough sketch of a system to track the running state of RIOT as it boots. Any code that must behave differently depending on where in the boot process the system is could benefit from this. This is just a proposal at this point; please let me know if this is acceptable as a core system.

One use case in particular is for a service that is started early in the boot process to depend on another service that is started later in the boot process. In general this is a bad idea, and currently it is impossible to do safely. In some cases though, it may be very difficult to get things to work otherwise (some uses of xtimer come in periph come to mind; currently things just spin). I propose that this might be acceptable if a number of conditions are met:
- There is no clear way of making the system work well otherwise
- There is a safe fallback if the later service is not available
- It is carefully checked for correctness, and tested with the service disabled

### DIFF
--- a/core/include/runstate.h
+++ b/core/include/runstate.h
@@ -1,0 +1,112 @@
+/*
+ * Copyright (C) 2018 Acutam Automation, LLC
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup   core_runstate System run state information
+ * @ingroup    core
+ * @brief      System run state information
+ * @details
+ *
+ * TODO
+ *
+ * @{
+ *
+ * @file
+ * @brief      System run state definitions
+ *
+ * @author     Matthew Blue <matthew.blue.neuro@gmail.com>
+ */
+
+#ifndef RUNSTATE_H
+#define RUNSTATE_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define RUNSTATE_MAJ     (5)
+#define RUNSTATE_MIN     (2)
+#define RUNSTATE_STEP    (0)
+
+#define RUNSTATE_MAJ_MASK     (0x7 << RUNSTATE_MAJ)
+#define RUNSTATE_MIN_MASK     (0x7 << RUNSTATE_MIN)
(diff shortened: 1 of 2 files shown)


tcschmidt:
@ZetaR60 could you be a bit more expressive on what this contribution actually does? Explain the key aspects in some more detail?

ZetaR60:
@tcschmidt It allows you to see what part of the running state the system is in by calling a function (e.g. the state could be just booted, periphs initialized, multithreading, basic drivers initialized, network initialized, sensors and actuators, main started, main ended, etc.). The labels in the runstate enum are the states the system goes through as it starts. The global runstate variable would  be updated as the system boots (not added yet, because this is just a rough sketch awaiting an okay to proceed). You can then query the global runstate variable to see what state the system is in. This allows you to write code that has conditions based on the state of the system, such as one module that initialized early safely calling another module that initialized later.

gebart:
The concept you describe seem slightly similar to Unix runlevel

tcschmidt:
Yes, but are these runlevels really well-defined for RIOT?

bergzand:
I think this could be a quite useful addition. I've noticed that a number of modules currently wait a predefined number of seconds and then assume that there is some network connectivity (coap_rdcli for example).

Having some notion of runlevels, these modules could simply trigger on a runstate transition.

tcschmidt:
@bergzand agreed - but to do this right, quite some conceptional pre-work seems needed.

gebart:
I don't think that the use case for this feature is quite clear. To me it seems like this might be better as an application specific state internal to the running application. There are very few decisions in the execution path before the main thread is started and coping with devices being initialized could be handled with proper dependencies in the auto init system. The xtimer example is something that may need special care anyway since a lot of device drivers require a timing source. Is there anything I did not see or misunderstood here?

ZetaR60:
@gebart I suppose this might be considered a simplified dependency system, based on the implication of a service (if enabled) being offered at a specific time during boot. If this PR is considered as such, how would it compare to what you were thinking of for a dependency system? This proposal could be changed to be a set of flags for various common features (or feature sets) being initialized, rather than an incrementing number.

I was considering calling this 'runlevel', except that Unix runlevels do not change during boot (e.g. setting runlevel to single user mode will make the system boot to single user mode).

PR #9319, opened 2018-06-09
drivers: add MH-Z19 CO2 sensor driver

moenoel:
### Contribution description

This commit adds a driver for the MH-Z19 CO2 sensor. Specifically the PWM interface (since it also has an UART). The PWM is interpreted by measuring the durations of high and low levels of a pulse and calculating the result according to a formular from the datasheet. It measures CO2 in parts per million up to 2000ppm.

See `tests/driver_mhz19/README.md` for further info and a link to the datasheet.

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -226,6 +226,11 @@ ifneq (,$(filter mag3110,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter mhz19,$(USEMODULE)))
+  USEMODULE += xtimer
+  FEATURES_REQUIRED += periph_gpio
+endif
+
 ifneq (,$(filter mma8x5x,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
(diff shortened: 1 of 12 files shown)


aabadie:
looks like a duplicate of #8648.

moenoel:
Whelp, I only looked whether there was a driver for this in master, not if there was already a PR. My bad.

Feel free to ignore this PR if the other implementation is better.

bergzand:
@moenoel I think both your and my approach are valid and certainly not mutual exclusive. My suggestion would be that we both continue working on our drivers and merge them together to give the end user the choice of reading over UART or over PWM.

What do you think?

aabadie:
> it might be easier to provide two modules (like mhz19_uart and mhz19_pwm).

You can provide 2 pseudo-modules (`mhz19_uart` and `mhz19_pwm`) that depends on a unique one `mhz19`. The common module will contain conditionally built code using `#ifdef MODULE_MHZ19_PWM` for example.

cgundogan:
>looks like a duplicate of #8648.

Then it probably makes sense to converge the discussion to only one PR (#8648). How about @moenoel  pull requesting this code onto @bergzand's PR and then we close this one?

moenoel:
> How about @moenoel pull requesting this code onto @bergzand's PR 

Will do once I find the time.

moenoel:
As suggested, this is now merged with the UART driver.

PR #9337, opened 2018-06-13
drivers*: add drivers with saul support in the doxygen drivers_saul group

aabadie:
### Contribution description

This PR is a follow-up of #9328 that put a doxygen dependency of drivers with saul support to the drivers_saul group.
This way all drivers with SAUL support can be reached directly when browsing the SAUL module documentation.

### Issues/PRs references

follow-up of #9328

### DIFF
--- a/drivers/include/lsm6dsl.h
+++ b/drivers/include/lsm6dsl.h
@@ -10,8 +10,10 @@
 /**
  * @defgroup    drivers_lsm6dsl LSM6DSL 3D accelerometer/gyroscope
  * @ingroup     drivers_sensors
+ * @ingroup     drivers_saul
  * @brief       Device driver for the LSM6DSL 3D accelerometer/gyroscope
  *
+ * This driver provides @ref drivers_saul capabilities.
  * @{
  *
  * @file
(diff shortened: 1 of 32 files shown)


gebart:
I like this idea!
When you use `@ingroup` twice the page is available from two places in the ToC? Nice feature

drivers/include/lsm6dsl.h
```diff
@@ -10,8 +10,10 @@
 /**
  * @defgroup    drivers_lsm6dsl LSM6DSL 3D accelerometer/gyroscope
  * @ingroup     drivers_sensors
- * @brief       Device driver for the LSM6DSL 3D accelerometer/gyroscope
+ * @ingroup     drivers_saul
+ * @brief       Device drive for the LSM6DSL 3D accelerometer/gyroscope
```
jia200x: s/drive/driver

jia200x:
looks good. There's only a minor issue I just posted

drivers/include/lsm6dsl.h
aabadie: Fixed and directly squashed

PR #9340, opened 2018-06-13
atmega_common/gpio.c Fixes GPIO_LOW interrupt

shr70:
fixes the GPIO_LOW interrupt on the atmega platform.

### Contribution description
This PR fixes a bug in the atmega_common gpio.c file. In order to set a GPIO_LOW interrupt, EICRA will need to set some bits to zero. The originial code did this by just shifting the `flank` variable. But 0 variables can not be shifted. Therefore the interrupts were not set correctly. Its fixed by checking first if GPIO_LOW is supposed to be set and then using `EICRA &= ~(0x3 << (int_num * 2));` to set the GPIO_LOW interrupt correctly.

### DIFF
--- a/cpu/atmega_common/periph/gpio.c
+++ b/cpu/atmega_common/periph/gpio.c
@@ -174,10 +174,12 @@ int gpio_init_int(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
 
     /* apply flank to interrupt number int_num */
     if (int_num < 4) {
+        EICRA &= ~(0x3 << (int_num * 2));
         EICRA |= (flank << (int_num * 2));
     }
 #if defined(EICRB)
     else {
+        EICRB &= ~(0x3 << ((int_num % 4) * 2));
         EICRB |= (flank << ((int_num % 4) * 2));
     }
 #endif


cpu/atmega_common/periph/gpio.c
```diff
@@ -174,11 +174,21 @@ int gpio_init_int(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
 
     /* apply flank to interrupt number int_num */
     if (int_num < 4) {
-        EICRA |= (flank << (int_num * 2));
+        if (flank == GPIO_LOW) {
```
ZetaR60: This issue also exists with all flanks that contain zeros (i.e. anything but GPIO_RISING).

I would suggest removing the check and instead always zero the relevant bits, followed by writing the flank onto them.

cpu/atmega_common/periph/gpio.c
```diff
@@ -174,11 +174,21 @@ int gpio_init_int(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
...
+            EICRA &= ~(0x3 << (int_num * 2));
+        }
+        else {
+            EICRA |= (flank << (int_num * 2));
+        }
     }
 #if defined(EICRB)
     else {
-        EICRB |= (flank << ((int_num % 4) * 2));
+        if (flank == GPIO_LOW) {
```
ZetaR60: See above.

cpu/atmega_common/periph/gpio.c
```diff
@@ -244,54 +254,46 @@ static inline void irq_handler(uint8_t int_num)
     __exit_isr();
 }
 
-ISR(INT0_vect, ISR_BLOCK)
-{
+ISR(INT0_vect, ISR_BLOCK){
```
ZetaR60: Why did you change the style of unrelated code? Several other instances below.

shr70: I missed this change. I ran uncrustify before pushing the branch. It does not seem to recognize ISR as a function and moved the curly brace therefore. I changed it back now.

ZetaR60: Looks like there is a fix being worked on for this issue: #9370

ZetaR60:
Unrelated failure in Murdock, but other than that it looks good. Please squash your fix into the first commit (git rebase then force push). For future use: Git provides a convenient way to label commits as fixes for automatic squashing (using ```--fixup``` for git commit).

PR #9347, opened 2018-06-14
cpu/cc2538: adapt periph to new I2C API

smlng:
### Contribution description

This adapt the CC2538 I2C to the new API.

Problem right now: it works with debug enabled but not without.

### Issues/PRs references

#6577

### DIFF
--- a/cpu/cc2538/periph/i2c.c
+++ b/cpu/cc2538/periph/i2c.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2015 Loci Controls Inc.
+ *               2017 HAW Hamburg
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -8,14 +9,14 @@
 
 /**
  * @ingroup     cpu_cc2538
- * @ingroup     drivers_periph
+ * @ingroup     drivers_periph_i2c
  * @{
  *
  * @file
  * @brief       Low-level I2C driver implementation
  *
  * @author      Ian Martin <ian@locicontrols.com>
- *
+ * @author      Sebastian Meiling <s@mlng.net>
  * @}
  */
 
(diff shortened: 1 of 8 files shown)


MrKevinWeiss:
Hmmm, it seems like the register is volatile which means it shouldn't be optimized out.  It could  just be the delay from the print of the debug.  I'll do a proper test (and might need some hardware from you) after I finish current task.

smlng:
Hey I fixed the issue by adding a small delay incrementing a variable, now it works with and without debug output.

smlng:
please note this also introduces a new gpio init function because the present cannot handle init of I2C pins but is used for UART. I'll will sort out things later on ...

dylad:
@smlng Can you tell us on which sensors did you test it ? It can help merging some devices faster at the same time.
 I'll take a look at it but if @MrKevinWeiss  or someone can test it. It would be awesome !

smlng:
I used the reworked `periph_i2c` by @MrKevinWeiss for testing, this is even better than just using some i2c sensor driver.

MrKevinWeiss:
Ya, so far it looks like there is a way to go...  I will play around with it for the remainder of the day but currently it seems like there is unexpected behavior in the remote-reva.
- read/write to wrong addr will result in data clk after failed addr
- a wrong addr write seems to block all reads until the correct write happens

I will spend the rest of the day trying to test/provide fix suggestions

MrKevinWeiss:
Man, I thought the sam_common0 had HW issues.  So I believe that the hardware will automatically send/clk a data byte regardless of an addr ack.  We probably have to accept it.  It seems like the write addr ack error does not get cleared properly.  If a read is issued after it will fail and will not be able to clear the byte (even though the scope looks fine).  It seems the only way to clear the error is with a successful write...

cpu/cc2538/periph/i2c.c
```diff
@@ -8,658 +9,331 @@
...
+
+        /* check master status */
+        uint_fast8_t stat = _i2c_master_status();
+        DEBUG ("%s: I2C master status (%u).\n", __FUNCTION__, stat);
+        if (stat & ANY_ERROR) {
+            _i2c_master_ctrl(STOP);
+            DEBUG("\tI2C master error: ");
+            if (stat &  DATACK) {
+                DEBUG("data ack lost!\n");
+                return -EIO;
```
MrKevinWeiss: If you can find a way to clear the error flags before exiting that would be nice... I ran out of time.

dylad:
@smlng Overall changes look ok but you need to add unsupported features check at the beginning of read_bytes and write_bytes.
```
if (flags & (I2C_REG16 || I2C_ADDR10) {
return -EOPNOTSUPP;
}
```

gebart:
@dylad I edited your comment to fix some typos in the code example

smlng:
@dylad good point, will do - but also need to sort out the write issue reported by @MrKevinWeiss

MrKevinWeiss:
I could be convinced to remove that from the test (or make it not critical).  This would mean that there **must** be a successful write before a read can happen.  **Ideally** there should be a reset of error state though.  I imagine the hardware isn't the most versatile.

cpu/cc2538/periph/i2c.c
MrKevinWeiss: The only way I could clear the flag is with a successful write.  Maybe try the standard solution?  (turn it off and on again)

smlng:
@MrKevinWeiss maybe the STOP command on error is not (always) needed, i.e., only put STOP when flags wasn't set in the original command. I need to tests this, because the data sheet doesn't say anything about reset error state (only IRQs), but it does say the STOP is either for successful write/read or error. 

I also need to add the flags handling as noted by @dylad, try to get this done in the next hours.

smlng:
hey @dylad  I added the flags check and I (believe) have fixed read/write issue with wrong addresses -> need to access the data register before checking for errors 😩  @MrKevinWeiss can you verify by (re)testing? THX!!

MrKevinWeiss:
So the problem of not clearing error flags before a read is still there.  The test was altered to specify a successful write must occur before a read.
Also an error was expected when sending a nostart read outside of a frame and did not occur.  The test was altered to allow for this.

It is up to you to decide if that should be acceptable behavior.  Everything else seems to work fine (ie 99% of use cases).

[Results](https://gist.github.com/MrKevinWeiss/8a6e58468b0f63fd74c565e6f1112556)

MrKevinWeiss:
If you would like to reproduce the two questionable error cases.
1. Send read_reg to a incorrect addr, then read_bytes to the correct one
2. Send a read_reg, then send a read_bytes with I2C_NOSTART flag active (should be an error since it no address should be sent after stop)

MrKevinWeiss:
Just to confirm if a NOSTART read is done it says it succeeds, nothing happens on the scope, it seems to just read back whatever data was last in the register...  This is improper use of the I2C frame, if it is not easy to catch in hardware though maybe we just leave if for the time being.

MrKevinWeiss:
I retested and it seems like the only issue left is the error flag not resetting when triggered with a NACK addr write.  To keep things simple I suggest we just specify that any read should have had a ACKed write before.  Eventually we will make this a warning for the test but not a test failure.  If that is the case just make Travis happy then we can merge.

MrKevinWeiss:
Also I don't think new_i2c_if branch goes through murdock... but I could be wrong.

dylad:
Here we go !

PR #8316, opened 2018-01-02
[rfc] file system benchmark

vincent-d:
### Contribution description

This is based on #8315 and #8239.
This provide an app to benchmark the file systems using vfs.

### Results

I've run the test on native with the emulated mtd device for spiffs and littlefs with the following results.

#### spiffs

Binary size:
```
   text	   data	    bss	    dec	    hex
  87874	    792	  49440	 138106	  21b7a
```

Timings:
```
benchmarking file system: spiffs
[BEGIN] Format test...
Test time: 69146us
[END] Format test
[BEGIN] Mount test...
Test time: 16786us
[END] Mount test
[BEGIN] Write test...
File #0, 400 bytes written in: 60569us
File #1, 400 bytes written in: 54963us
File #2, 400 bytes written in: 59451us
File #3, 400 bytes written in: 53766us
File #4, 400 bytes written in: 58093us
File #5, 400 bytes written in: 56973us
File #6, 400 bytes written in: 55487us
File #7, 400 bytes written in: 53980us
File #8, 400 bytes written in: 56722us
File #9, 400 bytes written in: 55598us
File #10, 400 bytes written in: 53592us
File #11, 400 bytes written in: 54107us
File #12, 400 bytes written in: 55462us
File #13, 400 bytes written in: 56342us
File #14, 400 bytes written in: 54442us
File #15, 400 bytes written in: 55705us
File #16, 400 bytes written in: 55280us
File #17, 400 bytes written in: 55397us
File #18, 400 bytes written in: 54621us
File #19, 400 bytes written in: 55357us
Mean time: 55795us
[END] Write test
[BEGIN] Read test...
File #0, 4000 bytes read in: 661us
File #1, 4000 bytes read in: 659us
File #2, 4000 bytes read in: 722us
File #3, 4000 bytes read in: 663us
File #4, 4000 bytes read in: 662us
File #5, 4000 bytes read in: 658us
File #6, 4000 bytes read in: 660us
File #7, 4000 bytes read in: 934us
File #8, 4000 bytes read in: 1039us
File #9, 4000 bytes read in: 1063us
File #10, 4000 bytes read in: 934us
File #11, 4000 bytes read in: 949us
File #12, 4000 bytes read in: 943us
File #13, 4000 bytes read in: 687us
File #14, 4000 bytes read in: 666us
File #15, 4000 bytes read in: 664us
File #16, 4000 bytes read in: 647us
File #17, 4000 bytes read in: 678us
File #18, 4000 bytes read in: 651us
[...]

PR #9349, opened 2018-06-14
mtd_spi_nor: remove sleep when waiting for write completed

vincent-d:
### Contribution description

When playing with #8316, I noticed that performances were a lot below theoretical expectations.
I tried changing the sleep time used when polling the 'write completed' status bit and found that simply yielding lead to the best performance.

It might lead to a lot of context switches if the erase time is really long, but in my benchmark, performances are 20 times better than before...

### Issues/PRs references

#8316

### DIFF
--- a/drivers/mtd_spi_nor/mtd_spi_nor.c
+++ b/drivers/mtd_spi_nor/mtd_spi_nor.c
@@ -44,14 +44,26 @@
 #define TRACE(...)
 #endif
 
-#ifndef MTD_SPI_NOR_WRITE_WAIT_US
-#define MTD_SPI_NOR_WRITE_WAIT_US (50 * US_PER_MS)
-#endif
-
 #define MTD_32K             (32768ul)
 #define MTD_32K_ADDR_MASK   (0x7FFF)
 #define MTD_4K              (4096ul)
 #define MTD_4K_ADDR_MASK    (0xFFF)
+
+#ifndef MTD_SPI_NOR_WAIT_C_ER
+#define MTD_SPI_NOR_WAIT_C_ER       (16 * US_PER_SEC)
+#endif
+
+#ifndef MTD_SPI_NOR_WAIT_S_ER
+#define MTD_SPI_NOR_WAIT_S_ER       (40 * US_PER_MS)
+#endif
+
+#ifndef MTD_SPI_NOR_WAIT_32K_ER
+#define MTD_SPI_NOR_WAIT_32K_ER     (20 * US_PER_MS)
+#endif
+
+#ifndef MTD_SPI_NOR_WAIT_4K_ER
+#define MTD_SPI_NOR_WAIT_4K_ER      (10 * US_PER_MS)
+#endif
 
 static int mtd_spi_nor_init(mtd_dev_t *mtd);
 static int mtd_spi_nor_read(mtd_dev_t *mtd, void *dest, uint32_t addr, uint32_t size);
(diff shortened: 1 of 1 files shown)


vincent-d:
@gebart any chance you could review this?

PeterKietzmann:
@vincent-d ping?!

vincent-d:
> Did you perform any experiments on how many iterations that this loop will spin after the change? Or how much time the flash write really takes?

No, I didn't. I'll try to find some time this week.

> but lower priority threads may be blocked more with this change than before

It's a drawback, but with the current code, the performance is really bad.

vincent-d:
> I'll try to find some time this week

It's been way more than a week but I finally ran some tests ;)

When erasing 4K, the loop used to iterate ~ 2k times. 
When writing, the loop iterates less than 100 times (between 27 and 87 times from my measurements).

So I changed a bit the code to pass a waiting estimate to the wait function and use it to wait, the estimate is then reduced by 1/n of its remaining value with n starting with 2 and incremented in each iteration. The lower bound of the waiting time is `2 * XTIMER_BACKOFF`.
I fine-tuned it with the flash I have to get close to the performances I had with "busy-yield-waiting".

Please take a look and let me know what you think.

vincent-d:
Ping @gebart ?

vincent-d:
I'd still like to have this merged ;)

benpicco:
The change looks reasonable, please squash.

drivers/mtd_spi_nor/mtd_spi_nor.c
```diff
@@ -44,15 +44,27 @@
...
+#define MTD_SPI_NOR_WAIT_S_ER       (40 * US_PER_MS)
+#endif
+
+#ifndef MTD_SPI_NOR_WAIT_32K_ER
+#define MTD_SPI_NOR_WAIT_32K_ER     (20 * US_PER_MS)
+#endif
+
+#ifndef MTD_SPI_NOR_WAIT_4K_ER
+#define MTD_SPI_NOR_WAIT_4K_ER      (10 * US_PER_MS)
+#endif
```
bergzand: Are all these numbers based on a standardized spec?

bergzand:
@vincent-d Tested on the PineTime with the SPI NOR flash chip. `tests/pkg_littlefs` went from approx. 15 sec to approx 18 sec execution time with the commit from this PR.

drivers/mtd_spi_nor/mtd_spi_nor.c
vincent-d: No, each flash might have different timings here. It might need some fine-tuning for your flash.

bergzand: > No, each flash might have different timings here. It might need some fine-tuning for your flash.

Does it make sense to move them to the flash params if they are model-specific?

bergzand:
> @vincent-d Tested on the PineTime with the SPI NOR flash chip. `tests/pkg_littlefs` went from approx. 15 sec to approx 18 sec execution time with the commit from this PR.

After some careful tuning, I was able to reduce the test run to 12 seconds, so I can confirm the performance increase (and probably together with a reduction of the number of thread wakeups).

bergzand:
@vincent-d I've submitted two PRs on top of this PR, one to modify the code with my suggestion [here](https://github.com/RIOT-OS/RIOT/pull/9349#discussion_r376982172) and one to add the actual erase timing in the debug output of the wait function.

Feel free to ignore them if you don't want to deal with those changes during this PR, I can also submit those two PRs after this PR is merged.

vincent-d:
@bergzand I'd prefer merging this one first and then yours if you don't mind

bergzand:
> @bergzand I'd prefer merging this one first and then yours if you don't mind

That's okay of course.

Issue #9342, opened 2018-06-13
OTA: tracking issue


bergzand:
# Tracking OTA progress

Tracking issue for the different parts required to enable over the air updates.

## Overview

We are aiming at a simple OTA prototype in a first phase, 
which can be extended in later phases.
The blueprint for the first prototype is the implementation we had working at the [SUIT Hackathon](https://github.com/suit-wg/Hackathon-Interim-Berlin) in June 2018 (see slide 3 of this [recap](https://github.com/suit-wg/Hackathon-Interim-Berlin/blob/master/SUIT-Berlin-Hackathon-recap.pdf)).
This combines a simplistic bootloader, 2 firmware slots and metadata that follows the SUIT specification. 
A basic update module, embarked in each image, manages firmware transfer (via CoAP), verification (complying with SUIT) and storage in Flash within the non-running image slot.

## Work breakdown
![ota work breakdown_v3](https://user-images.githubusercontent.com/11301549/47290920-f47dda00-d601-11e8-8d44-658a304ba713.jpg)

### Current state of development

@cladmi is addressing the tooling for the bootloader and metadata
@kYc0o is addressing the bootloader
@OYTIS is working on support for RIOT, including OTA, for KEA128LEDLIGHTRD board (https://github.com/RIOT-OS/RIOT/pull/9655)

See the [GitHub Software Updates project](https://github.com/RIOT-OS/RIOT/projects/11) for more details.

## System diagrams

Both of the diagrams reflect the content of [the working OTA update branch](https://github.com/RIOT-OS/RIOT/pull/9969). This contains working examples for both a metadata-based update, and a SUIT manifest-based update. The areas not relevant to the example addressed in each section are greyed out.

### Non-SUIT metadata based update mechanism

A description of the metadata is given [here](https://github.com/RIOT-OS/RIOT/issues/9342#issuecomment-413210688).

![ota component diagram - metadata based updating_v1](https://user-images.githubusercontent.com/11301549/45826950-c1d87d00-bcf5-11e8-854b-23ddd61c9fcb.jpg)

### SUIT metadata based update mechanism

![ota component diagram - SUIT-based updating_v1](https://user-images.githubusercontent.com/11301549/45826964-c735c780-bcf5-11e8-95d9-0f1cb9c82183.jpg)

The process being followed is to merge the functionality required for OTA updates piecewise, instead of as a large monolithic PR.

## Files for diagrams above

[...]

PR #9351, opened 2018-06-14
cortexm_common/ldscript: simplify generation of firmwares in section of the ROM

cladmi:
### Contribution description

This adds/modifies two build system configuration variables for `cortexm_common/ldscript/cortexm.ld`.

It is now possible to generate during a build, different elf files with different rom offset and firmware rom length by setting `ROM_OFFSET` and `FW_ROM_LEN` variables.

* Modify `ROM_OFFSET` so it can be set as a target specific variable to be used during linking.
* Add a new `FW_ROM_LEN` variable that can restrict the firmware length to use when linking instead of using the whole rom after `ROM_OFFSET`.
  * The previous solution was to change `ROM_LEN` to be `ROM_LEN = ROM_OFFSET + FW_ROM_LEN` which was not as straightforward to use.

### Use cases

The goal is firmwares that do not use the whole ROM for the over the air update scenario with a ROM split in 3 parts "Boot Loader - Firmware 1 - Firmware 2".

### Current state

In the test, I currently only set the supported board with a whitelist to `iotlab-xx` and `samr21-xpro` nodes. But it should be expandable to all boards using `cortexm_common/ldscript/cortexm.ld` as linking script.
I can do it depending on your feedback.

There are compilation tests for:

* Generating firmware with 2 different offsets
* Generating a firmware with a size of half the rom
* Overflow detection when setting a too big `FW_ROM_LEN`

### Issues/PRs references

Part of https://github.com/RIOT-OS/RIOT/issues/9342
~Depends on https://github.com/RIOT-OS/RIOT/pull/9451~

This is a modified version of things inspired from both following branches:

kYc0o https://github.com/kYc0o/RIOT/tree/ota_firmware_updates
kaspar030 https://github.com/kaspar030/RIOT/tree/ota_firmware_updates

I did not re-use the existing commits as it is a different implementation.
Tell me if you want to be mentioned in a commit.

### DIFF
--- a/cpu/cortexm_common/Makefile.include
+++ b/cpu/cortexm_common/Makefile.include
@@ -2,10 +2,6 @@
 INCLUDES += -I$(RIOTCPU)/cortexm_common/include
 INCLUDES += -I$(RIOTCPU)/cortexm_common/include/vendor
 
-ifneq (,$(ROM_OFFSET))
-  LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_rom_offset=$(ROM_OFFSET)
-endif
-
 # All variables must be defined in the CPU configuration when using the common
 # `ldscripts/cortexm.ld`
 ifneq (,$(ROM_START_ADDR)$(RAM_START_ADDR)$(ROM_LEN)$(RAM_LEN))
@@ -18,3 +14,10 @@ ifneq (,$(ROM_START_ADDR)$(RAM_START_ADDR)$(ROM_LEN)$(RAM_LEN))
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_rom_length=$(ROM_LEN)
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_ram_length=$(RAM_LEN)
 endif
+
+
+# Only define the linker symbol if the variable is set
+# The variable can be set using target specific variable thanks to lazy evaluation
+
+# ROM_OFFSET: offset in rom to start linking, allows supporting a bootloader
+LINKFLAGS += $(if $(ROM_OFFSET),$(LINKFLAGPREFIX)--defsym=_rom_offset=$(ROM_OFFSET))
--- a/cpu/cortexm_common/Makefile.include
+++ b/cpu/cortexm_common/Makefile.include
@@ -6,7 +6,13 @@ ifneq (,$(ROM_OFFSET))
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_rom_offset=$(ROM_OFFSET)
 endif
 
+# All variables must be defined in the CPU configuration when using the common
+# `ldscripts/cortexm.ld`
 ifneq (,$(ROM_START_ADDR)$(RAM_START_ADDR)$(ROM_LEN)$(RAM_LEN))
+  $(if $(ROM_START_ADDR),,$(error ROM_START_ADDR is not defined))
+  $(if $(RAM_START_ADDR),,$(error RAM_START_ADDR is not defined))
+  $(if $(ROM_LEN),,$(error ROM_LEN is not defined))
+  $(if $(RAM_LEN),,$(error RAM_LEN is not defined))
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_rom_start_addr=$(ROM_START_ADDR)
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_ram_start_addr=$(RAM_START_ADDR)
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_rom_length=$(ROM_LEN)
--- a/cpu/cortexm_common/Makefile.include
+++ b/cpu/cortexm_common/Makefile.include
@@ -21,3 +21,5 @@ endif
 
 # ROM_OFFSET: offset in rom to start linking, allows supporting a bootloader
 LINKFLAGS += $(if $(ROM_OFFSET),$(LINKFLAGPREFIX)--defsym=_rom_offset=$(ROM_OFFSET))
+# FW_ROM_LEN: rom length to use for firmware linking. Allows linking only in a section of the rom.
+LINKFLAGS += $(if $(FW_ROM_LEN),$(LINKFLAGPREFIX)--defsym=_fw_rom_length=$(FW_ROM_LEN))
--- a/cpu/cortexm_common/ldscripts/cortexm.ld
+++ b/cpu/cortexm_common/ldscripts/cortexm.ld
@@ -21,10 +21,11 @@
  */
 
 _rom_offset = DEFINED( _rom_offset ) ? _rom_offset : 0x0;
+_fw_rom_length = _rom_length - _rom_offset;
 
 MEMORY
 {
-    rom (rx)    : ORIGIN = _rom_start_addr + _rom_offset, LENGTH = _rom_length - _rom_offset
+    rom (rx)    : ORIGIN = _rom_start_addr + _rom_offset, LENGTH = _fw_rom_length
     ram (w!rx)  : ORIGIN = _ram_start_addr,               LENGTH = _ram_length
 }
 
--- a/cpu/cortexm_common/ldscripts/cortexm.ld
+++ b/cpu/cortexm_common/ldscripts/cortexm.ld
@@ -21,7 +21,9 @@
  */
 
 _rom_offset = DEFINED( _rom_offset ) ? _rom_offset : 0x0;
-_fw_rom_length = _rom_length - _rom_offset;
+_fw_rom_length = DEFINED( _fw_rom_length ) ? _fw_rom_length : _rom_length - _rom_offset;
+
+ASSERT((_fw_rom_length <= _rom_length - _rom_offset), "Specified firmware size does not fit in ROM");
 
 MEMORY
 {
--- a/cpu/cortexm_common/ldscripts/cortexm.ld
+++ b/cpu/cortexm_common/ldscripts/cortexm.ld
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2017 Inria
+ *               2018 Freie Universität Berlin
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -14,16 +15,17 @@
  * @brief           Memory definitions for the Cortex-M family
  *
  * @author          Francisco Acosta <francisco.acosta@inria.fr>
+ *                  Gaëtan Harter <gaetan.harter@inria.fr>
  *
  * @}
  */
 
-_boot_offset = DEFINED( _rom_offset ) ? _rom_offset : 0x0 ;
+_rom_offset = DEFINED( _rom_offset ) ? _rom_offset : 0x0;
 
 MEMORY
 {
-    rom (rx)    : ORIGIN = _rom_start_addr + _boot_offset, LENGTH = _rom_length - _boot_offset
-    ram (w!rx)  : ORIGIN = _ram_start_addr,                LENGTH = _ram_length
+    rom (rx)    : ORIGIN = _rom_start_addr + _rom_offset, LENGTH = _rom_length - _rom_offset
+    ram (w!rx)  : ORIGIN = _ram_start_addr,               LENGTH = _ram_length
 }
 
 INCLUDE cortexm_base.ld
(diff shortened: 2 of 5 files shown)


cladmi:
I re-added the main `if` to only test if all symbols are defined when one is defined.
It should fix boards that do not use the common `cortexm.ld`.

I tried doing it in `cortexm.ld` with `ASSERT` but the memory section complains about undefined symbol before the `ASSERT` so the message is `invalid length for memory region rom ` instead of the assert message.

cladmi:
Murdock broken build:

```
make: Entering directory '/tmp/dwq.0.9331483276232536/84bce6789fe2daca27b9d89f003b6097/tests/cortexm_common_ldscript'
make: *** No rule to make target '/tmp/dwq.0.9331483276232536/84bce6789fe2daca27b9d89f003b6097/build/auto_init.a', needed by '/tmp/dwq.0.9331483276232536/84bce6789fe2daca27b9d89f003b6097/build/tests_cortexm_common_ldscript_offset_0x1000.elf'.  Stop.
make: *** Waiting for unfinished jobs....
```
Is caused by https://github.com/RIOT-OS/RIOT/issues/8913 as the rule defining `BASELIBS` is defined before it has its real value.

I did not see it before as I compile without parallel build in a clean environment.

cladmi:
I rebased on commits that should handle`BASELIBS` definition.
It will do a separate PR for them next week.

I also added a debug test in Makefile.include to verify that BASELIBS is indeed stable in the file.

smlng:
@cladmi this depends on #9451 right?

cladmi:
Yes, my bad I forgot to update it.

cladmi:
I will add them to the list of tested boards in `tests/cortem_common_ldscript`.

MrKevinWeiss:
I tested on the bluepill (using swd flasher), tests/leds make LED0 flash.  I also ran the tests/cortem_common_ldscript and got
```Test compilation with offset 0x1000: [OK]
Test compilation with offset 0x2000: [OK]
Test compilation with half ROM length: [OK]
Test ROM overflow detection: [OK]
Test ROM overflow detection: [OK]
```
@kYc0o Let me know if you would like anything more extensive.

MrKevinWeiss:
> Did you change the ROM_OFFSET variable to do so?

Nope, I will reflash the bootloader (sometimes a pain) then attempt to vary the ROM_OFFSET using the tests/cortem_common_ldscript...  After the RIOT meeting.

kYc0o:
> then attempt to vary the ROM_OFFSET

The goal is not to modify it, but just test if RIOT flashed with the normal settings still works.

MrKevinWeiss:
> Actually the goal is to keep the bootloader
Hmmm seems like my hardware isn't happy (I cannot use the USB bootloader to program the bluepill).  I will try some other ones Monday.  I was able to flash with the st-link from the nucleo board or the UART setting it BOOT0 pin to 1.

MrKevinWeiss:
So here are the things I found...  Currently the bluepill, which uses the stm32f103c8 **works** with swd and RIOT.  Using the bootloader to flash is very outdated and full of problems.  The [bootloader firmware](https://github.com/rogerclarkmelbourne/STM32duino-bootloader/tree/master/binaries) generic_boot20_pc13.bin changed the bootmode pin to B2 (which means the boot1 jumper must be connected to 1).  The dfu-util also seems to show 3 instances instead of 1, requiring a serial number to specify the bootloader.
```
dfu-util: Invalid DFU suffix signature
dfu-util: A valid DFU suffix will be required in a future dfu-util release!!!
dfu-util: More than one DFU capable USB device found! Try `--list' and specify the serial number or disconnect all but one device
```
Here are the instances with the most recent bootloader
```
Found DFU: [1eaf:0003] ver=0201, devnum=110, cfg=1, intf=0, alt=2, name="STM32duino bootloader v1.0  Upload to Flash 0x8002000", serial="LLM 003"
Found DFU: [1eaf:0003] ver=0201, devnum=110, cfg=1, intf=0, alt=1, name="STM32duino bootloader v1.0  Upload to Flash 0x8005000", serial="LLM 003"
Found DFU: [1eaf:0003] ver=0201, devnum=110, cfg=1, intf=0, alt=0, name="STM32duino bootloader v1.0  ERROR. Upload to RAM not supported.", serial="LLM 003"
```
The third thing is it appears to not produce a .bin file which the bootloader requires.

Finally the last problems (though it is more likely my problem) requires it to be run with sudo or the udev rules updated.

Note that I used the dfu-util 0.8 as this is the most recent version for Ubuntu 16.04 LTS

cladmi:
> The third thing is it appears to not produce a .bin file which the bootloader requires.

If you do `make BOARD=bluepill PROGRAMMER=dfu-util flash` it should create the binfile and use it for flashing.

https://github.com/RIOT-OS/RIOT/blob/9f26d374ae0a572a48a1fa299e679cbc1ad4680f/boards/common/stm32f103c8/Makefile.include#L19-L26

kYc0o:
I had a arduino-mkr1000 on my drawer, so I could test. Everything works and the bootloader is kept. I'd say then it's safe to merge.

kYc0o:
However,  discovered a bug specific of that target (sam0). It works just by chance but the Makefiles need to be adapted. Can you take a look @cladmi?

kYc0o:
Basically, the `--defsym=_rom_offset=0x2000` variable is correctly set, however there's no logic to subtract that amount from the total size of the ROM. This can lead to no error while compiling large images which wouldn't fit the available ROM.

cladmi:
Which board is it ?

Because this should be handled by

    _fw_rom_length = DEFINED( _fw_rom_length ) ? _fw_rom_length : _rom_length - _rom_offset;

I just moved what was done before in the `MEMORY` section to setting a variable.

kYc0o:
It doesn't work in master too. The board is the arduino-mkr1000 which embeds a samd21.

cladmi:
What is your test to say "it doesn't work" because for me it does.

I test in `examples/hello-world`:

```
$ rm -r bin/
$ make BOARD=arduino-mkr1000
$ cat bin/arduino-mkr1000/hello-world.bin | wc -c
8504

$ make BOARD=arduino-mkr1000  elffile ROM_LEN=8503+0x2000
.... region `rom' overflowed by 1 byte.
$ make BOARD=arduino-mkr1000  elffile ROM_LEN=8504+0x2000
OK
```

Where 0x2000 is `ROM_OFFSET` taken from `boards/common/arduino-mkr/Makefile.include`.


I will try to add this as an automated test too.

cladmi:
It also checked when setting `FW_ROM_LEN` directly to `8503` or `8504`.

kYc0o:
I just typed `BOARD=arduino-mkr1000 QUIET=0 make -C examples/default`. However, I was confused because the output is from make *before* any computation is done:

```console
 -Wl,--defsym=_rom_offset=0x2000 -Wl,--defsym=_rom_start_addr=0x00000000 -Wl,--defsym=_ram_start_addr=0x20000000 -Wl,--defsym=_rom_length=0x40000 -Wl,--defsym=_ram_length=0x8000
```

It might be useful to know (print) if the calculation is done properly, but not mandatory.

cladmi:
Yes it is done in the linkerscript so not visible with QUIET=0.

This currently only tested setting FW_ROM_LEN with the `test-fw_len_half_rom` target that checks the output of:

```
arm-none-eabi-readelf --symbols bin/arduino-mkr1000/tests_cortexm_common_ldscript.elf  | grep _fw_rom_length
   437: 0003e000     0 NOTYPE  GLOBAL DEFAULT  ABS _fw_rom_length
```

I can add another one that verifies the main elf file.

I can also add test for an exact match of the size.

cladmi:
I am going to rebase/re-organize commits to get correct changes and separate tests that could have worked before.

cladmi:
I rewrote the git history and re-tested each commit.

I still did variable name changes early to move toward the final version as introducing `_fw_rom_length` was required to test using `readelf`.

kYc0o:
Everything looks ok now, however for the bluepill and stm32mindev boards the test of ROM_OFFSET is "failing", since the test for that variable is skipped and it should be performed. Can you check?

cladmi:
However I did a fixup to put an `if` in a newline and github is not showing it…

kYc0o:
Ok great, you might squash your last fixup.

bergzand:
Anything specific you need my opinion on?

cladmi:
@bergzand Mainly a second look as it is part of #9342 and that it would be great to a have someone else than @kYc0o who also contributed to writing the changes.

As there is not much feedback if boards are still working https://github.com/RIOT-OS/RIOT/pull/9351#issuecomment-402779172 I also checked that the `.text` start address is set with the same value as master using:

    arm-none-eabi-readelf --section-headers ELFFILE.elf 2>/dev/null | awk '/.text/{printf "0x%s\n", $5}'

I checked both with and without `PROGRAMMER=dfu-util` for `BOARDS="arduino-mkr1000 bluepill feather-m0 opencm904 spark-core"`

cpu/cortexm_common/ldscripts/cortexm.ld
```diff
@@ -14,16 +15,20 @@
...
+ *                  Gaëtan Harter <gaetan.harter@inria.fr>
  *
  * @}
  */
 
-_boot_offset = DEFINED( _rom_offset ) ? _rom_offset : 0x0 ;
+_rom_offset = DEFINED( _rom_offset ) ? _rom_offset : 0x0 ;
+_fw_rom_length = DEFINED( _fw_rom_length ) ? _fw_rom_length : _rom_length - _rom_offset;
+
+ASSERT((_fw_rom_length <= _rom_length - _rom_offset), "Specified _fw_rom_length does not fit in ROM");
```
bergzand: Maybe reword the assert message to something like
`Specified firmware size does not fit in ROM`
to keep it understandable for end users.

cpu/cortexm_common/Makefile.include
```diff
@@ -2,13 +2,24 @@
...
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_ram_start_addr=$(RAM_START_ADDR)
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_rom_length=$(ROM_LEN)
   LINKFLAGS += $(LINKFLAGPREFIX)--defsym=_ram_length=$(RAM_LEN)
 endif
+
+
+# Only define the linker symbol if the variable is set
+# The variable can be set using target specific variable thanks to lazy evaluation
+
+# ROM_OFFSET: offset in rom to start linking, allows using bootloaders
```
bergzand: Is the "bootloaders" in this comment about RIOT specific bootloaders, about board supplied bootloaders, or both?

cpu/cortexm_common/ldscripts/cortexm.ld
```diff
@@ -14,16 +15,20 @@
  * @brief           Memory definitions for the Cortex-M family
  *
  * @author          Francisco Acosta <francisco.acosta@inria.fr>
+ *                  Gaëtan Harter <gaetan.harter@inria.fr>
  *
  * @}
  */
 
-_boot_offset = DEFINED( _rom_offset ) ? _rom_offset : 0x0 ;
+_rom_offset = DEFINED( _rom_offset ) ? _rom_offset : 0x0 ;
```
bergzand: extra whitespace between number and semicolon

bergzand:
> @bergzand I will add a README with the expected output.

Thanks, I'm afraid that mainly the negative tests might confuse people

cladmi:
I added a `compile-test` sub target to mention it in the documentation and a README.md.

cpu/cortexm_common/Makefile.include
cladmi: Done

bergzand:
ACK!

@cladmi Thank you for adding the readme. Please squash

cladmi:
> Thank you for adding the readme.

You were right to request having one in the first place :) I just tend to forget if it is not minor changes.

I am re-running the test in our docker image for verification (but pushed before testing of course…)

cladmi:
Compilation and compile tests works on docker image and arm-none-eabi-gcc 8.1.0.

bergzand:
@cladmi Your commits show up as unverified in the github webinterface. Are you sure that those are your commits? Should we trust these changes? :laughing:

cladmi:
Oh it's because I am from my laptop and did not commit with my fu-berlin address. Will rebase.

cladmi:
Oh just misconfiguration of the commit ID I think, will rebase and see.

kYc0o:
Congrats! One step forward for easy OTA updates 😃

PR #9354, opened 2018-06-14
Add support for the SODAQ ONE board

keestux:
### Add support for the SODAQ ONE board

This is another SODAQ board similar to the Autonomo and Explorer boards.  It is compatible
with Arduino SAMD.

This commit is based on Explorer and Autonomo and it adds all the specifics of the SODAQ ONE. Perhaps @aabadie can do the review since he added sodaq-explorer


boards/sodaq-one/include/board.h
```diff
@@ -0,0 +1,119 @@
...
+ * @{
+ */
+#define XTIMER_WIDTH        (16)
+/** @} */
+
+/**
+ * @name    LED pin definitions and handlers
+ * @{
+ */
+#define LED_GREEN_PIN       GPIO_PIN(PB, 10)
```
aabadie: The LED macro should be called LED_0_XX/LED_1_XX, etc to be consistent with other boards. This way those macros can be used on any platform that provides them.

boards/sodaq-one/board.c
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#include "cpu.h"
+#include "board.h"
+#include "periph/gpio.h"
+
+void board_init(void)
+{
+    /* initialize the on-board LEDs, switch them off to start with */
+    LED_GREEN_OFF;
```
aabadie: I don't understand why you need to call this macro before initializing the pin

keestux: Well, the reason is to have a defined (low) value before enabling it as output. This trick is mostly used to avoid a spike when enabling some output port. For a LED it doesn't really matter.

keestux:
Example is removed
LEDs are renamed

I don't see much that can be refactored into a common boards/sodaq. At least, it's not obvious to me.

About I2C refactoring #6577, yes please. The sooner the better.

PeterKietzmann:
@aabadie?

aabadie:
@PeterKietzmann, this one is waiting for the I2C refactoring embargo.

PeterKietzmann:
@aabadie not anymore

aabadie:
Thanks @keestux. Now Murdock is reporting errors, can you fix them ? After, it will be good to go.

PR #9357, opened 2018-06-14
 make: Clean up WERROR handling 

gebart:
### Contribution description

Clean up of the WERROR handling in Makefile.include to let the WERROR variable only control the actual -Werror compiler flag. All other warning related flags are set regardless of WERROR=1 or not.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -129,13 +129,13 @@ endif
 WERROR ?= 1
 export WERROR
 ifeq ($(WERROR),1)
-  CFLAGS += -Wall -Werror -Wextra
+  CFLAGS += -Werror
 endif
 
 WPEDANTIC ?= 0
 export WPEDANTIC
 ifeq ($(WPEDANTIC),1)
-    CFLAGS += -Wpedantic -pedantic-errors
+  CFLAGS += -Wpedantic
 endif
 
 # remove this once codebase is adapted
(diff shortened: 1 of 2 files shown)


Makefile.include
```diff
@@ -129,13 +129,16 @@ endif
...
+  CFLAGS += -Werror
 endif
 
 WPEDANTIC ?= 0
 export WPEDANTIC
 ifeq ($(WPEDANTIC),1)
-    CFLAGS += -Wpedantic -pedantic-errors
+  CFLAGS += -Wpedantic
+  ifeq ($(WERROR),1)
+    CFLAGS += -pedantic-errors
```
cladmi: From some testing, `-pedantic-errors` looks redundant with `-Werror -Wpedantic` and so could be removed.
I get the same behavior with and without `-pedantic-errors` for the following:

Compiling unittests for samr21-xpro with

```
make -C tests/unittests/ tests-spiffs BOARD=samr21-xpro WPEDANTIC=1 WERROR=0
```
and with `WERROR=1`.

While having the following diff to have a warning:

``` diff
diff --git a/pkg/spiffs/Makefile.spiffs b/pkg/spiffs/Makefile.spiffs
index 1cb6ab8..bd7256c 100644
--- a/pkg/spiffs/Makefile.spiffs
+++ b/pkg/spiffs/Makefile.spiffs
@@ -1,6 +1,6 @@
 MODULE := spiffs
 
 # Disable 'ISO C99 doesn’t support unnamed structs/unions [-Werror=pedantic]'
-CFLAGS += -Wno-pedantic
+#CFLAGS += -Wno-pedantic
 
 include $(RIOTBASE)/Makefile.base
```

gebart: deleted this block

gebart:
squashed

PR #9360, opened 2018-06-14
embunit: Const test case names

gebart:
### Contribution description

There is no reason to let embunit think test case names are writable. Fixes `-Wwrite-strings`


gebart:
I am lazy today. It fixes warnings when using -Wwrite-strings which is not turned on by default, but it is just common sense that these strings should be const because they should never be modified at run time.

PR #9363, opened 2018-06-16
core/pthread: make them compilable with g++

gschorcht:
If ```pthread``` is used in source files that have to be compiled with g++, ```atomic``` has to be included instead of ```stdatomic.h``` in ```pthread_spin.h```. 

The compilation with g++ may also require the declaration of function  ```sched_yield```. This should be declared not only for the native architecture, but in any case. It can then be implemented in ```thread_arch.c``` if necessary.

### DIFF
--- a/sys/posix/pthread/include/pthread_spin.h
+++ b/sys/posix/pthread/include/pthread_spin.h
@@ -20,10 +20,11 @@
 #ifndef PTHREAD_SPIN_H
 #define PTHREAD_SPIN_H
 
-#include <stdatomic.h>
-
 #ifdef __cplusplus
-extern "C" {
+#include <atomic>
+using std::atomic_flag;
+#else
+#include <stdatomic.h>
 #endif
 
 /**
@@ -35,6 +36,10 @@ extern "C" {
 typedef struct {
     atomic_flag flag; /**< Current lock state */
 } pthread_spinlock_t;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 /**
  * @brief           Intializes a spinlock.
--- a/core/include/native_sched.h
+++ b/core/include/native_sched.h
@@ -48,6 +48,14 @@ inline int sched_yield(void)
     puts("[ERROR] sched_yield called (defined in sched.h)\n");
     return 0;
 }
+#else
+/**
+ * @brief Compilation with g++ may require the declaration of this function.
+ *
+ * If implementation of this function is required, it can be realized in
+ * thread_arch.c.
+ */
+extern int sched_yield(void);
 #endif /* BOARD_NATIVE */
 
 #ifdef __cplusplus


sys/posix/pthread/include/pthread_spin.h
```diff
@@ -20,12 +20,17 @@
...
 
-#include <stdatomic.h>
-
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#ifdef __cplusplus
+#include <atomic>
+using namespace std;
```
josephnoir: Not a good idea to do this in a header as the `using` statement leaks into the including file.

gschorcht:
@josephnoir Thank you for the review. Of course, there are a lot C files that include ```stdatomic.h``` without any problems. As stated in the title the problem raises, when you try to compile a C++ file that includes ```stdatomic.h```. In this case, you have to include ```<atomic>``` instead.

You can reproduce the problem as following:

- rename ```tests/pthread/main.c``` to ```tests/pthread/main.cpp```
- execute ```make BOARD=native -C tests/pthread QUIET=0 CXXEXFLAGS=-std=c++11```

Unfortunatly, including ```<atomic>``` instead of ```stdatomic.h``` was not enough. It requires to declare ```std``` as used namespace. It's clear to me, that declaring used namespaces in a header file is definitly the wrong place. It was just a workaround as described here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60932

I have changed it in following way:

```
typedef struct {
#ifdef __cplusplus
    std::atomic_flag flag; /**< Current lock state */
#else
    atomic_flag flag; /**< Current lock state */
#endif
} pthread_spinlock_t;
```

Furthermore, when I tried to reproduce the problem on native platform, I had to realize  that C binding has to be declared after the include of the header files. With my Xtensa g++ suite, I didn't see this problem. So, I have changed the order too.

sys/posix/pthread/include/pthread_spin.h
gschorcht: It's clear to me and you are definitely right. It was just a workaround as described here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60932. I changed it already.

josephnoir:
It looks like you added the include block twice now.  I can reproduce the error using the `arm-none-eabi` version 7.2.1 20170904 building for the samr21 board.

However, I'm still not convinced that this is a good solution. If you try to build the rmutex test in the same way, it fails as well. The error messages I see are loads of these:

```
$HOME/Tools/gcc/gcc-arm-none-eabi-7-2017-q4-major/lib/gcc/arm-none-eabi/7.2.1/include/stdatomic.h:43:9: error: '_Atomic' does not name a type
 typedef _Atomic unsigned char atomic_uchar;
         ^~~~~~~
```

Reading online the only solution seems to be indeed to switch to the C++ header. But I'm not a fan of the using statement and otherwise we need to write a lot of `ifdefs` which leads to a lot of clutter. As far as I an tell, the following headers include the stdatomic header: 

```
core/include/rmutex.h
sys/include/vfs.h
sys/include/can/pkt.h
pkg/emb6/include/sock_types.h
sys/posix/pthread/include/pthread_spin.h
```

IIRC @gebart worked on atomics in RIOT before, any helpful insights?

gschorcht:
@josephnoir Oh sorry, obviously I overlooked the duplication of the include when rebasing. Thanks, I fixed it.

Concerning the C++ compatible header files: You are completely right, obviously there are more header files that include ```stdatomic.h``` are therefore not compilable with g++. I just realized it when porting RIOT to ESP32 because I tried to use a vendor library which uses pthread and is only available in c++. All files I found were

./core/include/rmutex.h
./cpu/msp430_common/include/stdatomic.h
./pkg/emb6/include/sock_types.h
./sys/include/can/pkt.h
./sys/include/vfs.h
./sys/posix/pthread/include/pthread_spin.h

but I don't know how many there could be in packages.

From my point of view, it is more basic question with two possible options:

1. We change the other files, which are not so many to make them compilable with C++.

2. We define a new coding rule which states that RIOT cannot use C++ source code and C++ must not be used. In that case the coding rule, that C header files should be always wrapped by extern "C" statement is no longer needed.

From my point of view option 1 would be better way. I could change it in all the files.

josephnoir:
Since RIOT officially claims C++ support on its homepage I think just abandoning support is not a viable option. But if we just change the existing files, someone will write a new module that includes the stdatomic header and we'll have the same problem again.

Another approach that comes to mind is a `riot/atomic.h` header which does nothing but include either `stdatomic.h` for C or `atomic` for C++ and includes a using statement to import all the related functions and types into `::` (e.g. through`using std::atomic_flag`). The header should then be used in modules that require atomics.

This won't affect developers building applications besides importing all the atomic stuff from `std`. But they probably know their language and can include the respective header directly if they want to use atomics.

Hm, do you have another idea? 🤔

gschorcht:
@josephnoir Your option 3 is very good idea. It would be perfect for me.

josephnoir:
The header should probably be located in core/include. Will you create it or should I?

josephnoir:
@gschorcht What is the status here?

gschorcht:
@josephnoir sorry, I'm quite busy with my business and my RIOT ESP ports at the moment.

josephnoir:
What I'm worrying about is that we'll run into the same problem again in the future because most code is just tested in C.

If we want to keep this in the header files, we should use the same syntax everywhere, e.g., the one already used in `./sys/include/vfs.h`.

gschorcht:
@josephnoir 

-  ```./sys/posix/pthread/include/pthread_spin.h``` changed according to the style as used in ```./sys/include/vfs.h```.
- ```./core/include/rmutex.h``` is not compilable with C99+
- ```tests/can_trx``` compiles after some C99+ changes in ```main.cpp``` without changes in ```./sys/include/can/pkt.h``` (obviously only parts written in C use this header)

josephnoir:
K, let's see what Murdock has to say.

gschorcht:
Uff, doxygen pitfall ...

josephnoir:
Sorry this takes so long. Since this is a change to core we need a second review, let's ping @gebart again.

josephnoir:
@gebart Thanks!

Issue #9359, opened 2018-06-14
thread: possibility to add architecture specific data to thread control block required


gschorcht:
During the ports of RIOT for ESP8266 and ESP32 MCUs, I had to learn that it would be really helpful if it would be possible to extend the thread control block by architecture specific data that belong to the thread context, .e.g., the saved interrupt level for nested interrupts with task switching.

At the moment I'm doing that by managing additional structures in additional tables for all threads in parallel to the thread table.

From my point of view it would be very easy to realize it as following:

The definition of ```struct thread``` is by a simple preprocessor definition, for example ```THREAD_ARCH_EXT```. Since ```thread.h``` includes ```cpu_conf.h```, additional members for specific data could be added by the architecture:

```
struct _thread {
    char *sp;                       /**< thread's stack pointer         */
    uint8_t status;                 /**< thread's status                */
    uint8_t priority;               /**< thread's priority              */
    ...
    THREAD_ARCH_EXT                 /**< architecture specific extensions */
    ...
}
```
If the definition of ```THREAD_ARCH_EXT``` in ```cpu_conf.h``` would be
```
#define THREAD_ARCH_EXT   uint32_t saved_int_level;
```
```saved_int_level``` would become member of the thread control block.

**Comment by kaspar030** [comment]
Yeah, we've had this issue before (#1234), but dismissed it as there was no need anymore.
Feel free to pick this up again!

I'd suggest using a struct, e.g.,

```
struct _thread {
    char *sp;                       /**< thread's stack pointer         */
    uint8_t status;                 /**< thread's status                */
    uint8_t priority;               /**< thread's priority              */
    ...
#ifdef CPU_HAS_THREAD_ARCH_T
  thread_arch_t arch;     /** architecture dependent thread state */
#endif
    ...
}
```

In ```cpu/$arch/include/cpu_conf.h```:
```
typedef struct {
  uint32_t saved_int_level;
} thread_arch_t;
#define CPU_HAS_THREAD_ARCH_T
```

kYc0o:
Can we close this since #9364 adds this functionality?

kaspar030:
Yes!

PR #9364, opened 2018-06-16
core: add architecture dependent thread information

gschorcht:
By adding a member of type ```thread_arch_t``` to the ```thread_t``` structure, depending on the existence of definition ```HAVE_THREAD_ARCH_T```, it is possible to add architecture-dependent information to the thread control block that are required to restore the context. This could be used for example for references to the co-processor save area.

Since ```thread.h``` includes ```$(ROITCPU)/include/cpu_conf.h```, the architecture dependent definition of additional context information can be placed there. For example, the following code add additional members to the thread context as required for the EPS32 architecture.
```
/**
 * @brief   Architecture depended part of thread control block for ESP32 CPU
 */
#define HAVE_THREAD_ARCH_T
typedef struct {
    uint32_t* saved_int_level;    /**< pointer to saved interrupt level */
    uint32_t* cp_save_area;       /**< pointer to co-processor save area */
} thread_arch_t;
```
This PR is related to issue #9359

### DIFF
--- a/core/include/thread.h
+++ b/core/include/thread.h
@@ -209,7 +209,10 @@ struct _thread {
     const char *name;               /**< thread's name                  */
     int stack_size;                 /**< thread's stack size            */
 #endif
-};
+#ifdef HAVE_THREAD_ARCH_EXT_T 
+    thread_arch_ext_t arch;         /**< architecture dependent exten- 
+                                         sions                          */ 
+#endif };
 
 /**
  * @def THREAD_STACKSIZE_DEFAULT
--- a/core/include/thread.h
+++ b/core/include/thread.h
@@ -209,10 +209,10 @@ struct _thread {
     const char *name;               /**< thread's name                  */
     int stack_size;                 /**< thread's stack size            */
 #endif
-#ifdef HAVE_THREAD_ARCH_EXT_T 
-    thread_arch_ext_t arch;         /**< architecture dependent exten- 
-                                         sions                          */ 
-#endif };
+#ifdef HAVE_THREAD_ARCH_T
+    thread_arch_t arch;             /**< architecture dependent part    */
+#endif
+};
 
 /**
  * @def THREAD_STACKSIZE_DEFAULT


core/include/thread.h
```diff
@@ -209,6 +209,9 @@ struct _thread {
     const char *name;               /**< thread's name                  */
     int stack_size;                 /**< thread's stack size            */
 #endif
+#ifdef HAVE_THREAD_ARCH_STATE_T
```
kaspar030: may sound nitpicky, but do you insist on the "state" part of the variable? it is not part of ```thread_t```, and as this is a data type, implicitly it contains "(state) data", so IMO it can be shortened.
What do you think?

kaspar030: Sorry, I was not clear. I was just suggesting renaming ```thread_arch_state_t``` to ```thread_arch_t```. The feature is very much wanted!

Sorry again!

gschorcht: Changed and rebased. Description updated.

haukepetersen:
+1 for the change. For the sake of being nit-picky here:  I am with @kaspar030 that `thread_arch_t` is the more fitting name in this case, as `thread_arch_ext_t` IMHO implies that this is an extension of `thread_arch`, which it is not, right? 

I only care so much for this 'stupic wording', as this effects the 'deep core' of RIOT...

gschorcht:
@haukepetersen @kaspar030
Ok, should I change it again or not? I can live with ```thread_arch_ext_t``` (which stands for architecture dependent extension of ```thread_t```) as well as ```thread_arch_t```. I have to commit it again anyway because of a trailing whitespace :-(

kaspar030:
> Ok, should I change it again or not?

yes please!

gschorcht:
Ok, I will do tomorrow since I'm not at my desk today.

gschorcht:
@kaspar030 Changed as requested

kaspar030:
ACK from my side. It is a small change, but very central, so we need another look. @gebart @haukepetersen could you take another look?

gschorcht:
When can I expect that this change is merged. There is no progress since July, 6th.

kYc0o:
I'd say if we have no further comments in 24h we can merge it. It's a kernel feature so it needs two ACKs, thus I'd give mine after that time.

kaspar030:
> I'd say if we have no further comments in 24h we can merge it. It's a kernel feature so it needs two ACKs, thus I'd give mine after that time.

This needs a proper second look, not a free ACK after 24hrs.

kYc0o:
> This needs a proper second look, not a free ACK after 24hrs.

I'm giving that proper second look, that's why I proposed 24h to have time to look at it if someone who you asked the proper second look is not available. I surely take more time than they.

gschorcht:
Thanks you all.

kYc0o:
I let @kaspar030 merge this in case of he waits for @haukepetersen and @gebart opinions.

kaspar030:
Thanks @kYc0o!

gschorcht:
Thanks

PR #9370, opened 2018-06-18
uncrustify: special function ISR

Josar:
This PR sets uncrustify to handle AVR ISR macro like functions and wrap them.

Uncrustify would change them from wrapped to unwrapped like this.

```
ISR(TIMER_1_ISRC, ISR_BLOCK){
    _isr(1, 2);
}
```

With this PR the ISR stay or will be changed to be formatted as follows 
```
ISR(TIMER_1_ISRC, ISR_BLOCK)
{
    _isr(1, 2);
}
```

### DIFF
--- a/uncrustify-riot.cfg
+++ b/uncrustify-riot.cfg
@@ -74,3 +74,9 @@ align_on_tabstop       = TRUE      # align on tabstops
 align_enum_equ_span    = 4         # '=' in enum definition
 align_struct_init_span = 0         # align stuff in a structure init '= { }'
 align_right_cmt_span   = 3         #
+
+#
+# Special cases
+#
+
+set PROTO_WRAP ISR   # Wrap ISR macros like functions


uncrustify-riot.cfg
```diff
@@ -74,3 +74,9 @@ align_on_tabstop       = TRUE      # align on tabstops
 align_enum_equ_span    = 4         # '=' in enum definition
 align_struct_init_span = 0         # align stuff in a structure init '= { }'
 align_right_cmt_span   = 3         #
+
+#
+# Special cases
+#
+
+set PROTO_WRAP ISR   # Wrap AVR ISR macros like a functions
```
aabadie: either remove `a` or the `s` at the end of function. I would do the former.

aabadie: also git grep shows that this macro is also used in other CPUs (msp430, nrf5xx, ) so the comment should not mention  AVR only.
Do you think the special case could be extended using some kind of regexp like `set PROTO_WRAP *ISR` ? This way it will also cover `UART_X_ISR`, `TIMER_X_ISR` used in other cpus.

Josar: The wrapping problem does not occur with the other isr as they are redefinition of functions and are then used as functions. So uncrustify recognizes them as functions.

they look like follows.
```
void I2C_2_ISR(void)
void TIMER_1_ISR(void)
void UART_0_ISR_RX(void)
```

PR #9373, opened 2018-06-19
boards/microbit: set openocd as default programmer

aabadie:
### Contribution description

This PR provide a working openocd configuration for the micro:bit in replacement to the initial "fscopy" programmer.
I was surprised that the change is so 'simple'.

Since I don't have a JLink, I couldn't use gdb with this board before. Now this is possible with openocd.

Note that I'm using a recent version of openocd (master version, less than 6 months) so maybe older versions (0.9) don't provide support for nrf51x cpu.

### Issues/PRs references

None


haukepetersen:
not sure this works with all microbits per default, at least the older models do not come pre-flashed with the segger firmware. Also did you check that the virtual serial port is still working? Last time I tried the Segger firmware for the flash chip it did not...

aabadie:
The microbit I have provide by default a DAPLink interface and are recognized as such by my system. There are now reference to jlink required with openocd. Except if you tweaked the flash chip with the Segger firmware.

PeterKietzmann:
@smlng interested?

PeterKietzmann:
Ups forget it. I confused microbit and calliope again...

aabadie:
No one interested by this PR ? FYI, Microbit are supported by the IoT-LAB testbed and this configuration is used there with openocd 0.10 for flashing/resetting/debugging.
Regarding @haukepetersen's previous comment, I couldn't any reference of versions of microbit preflashed with a segger firmware. The flashing/debugging is done via DAPLink, which is what this PR uses with OpenOCD.

aabadie:
anyone with a micro:bit for testing this PR ?

aabadie:
To prove that this PR works, I put the full output of 
```sh
make BOARD=microbit -C examples/default flash term
```
in this [gist](https://gist.github.com/aabadie/12bceacae77ab65195cd99b1b62f5496)

aabadie:
Any chance to get an ACK an before the release here ?

haukepetersen:
I just tested it with my `microbit`, and it seems so work. Though my openocd output looks kind of broken in parts, the firmware is still flashed ok:
```
### Flashing Target ###                                                                                                                  │
Open On-Chip Debugger 0.10.0+dev-00207-g4109263 (2017-12-01-12:49)                                                                       │
Licensed under GNU GPL v2                                                                                                                │
For bug reports, read                                                                                                                    │
        http://openocd.org/doc/doxygen/bugs.html                                                                                         │
cortex_m reset_config sysresetreq                                                                                                        │
adapter speed: 1000 kHz                                                                                                                  │
Info : CMSIS-DAP: SWD  Supported                                                                                                         │
Info : CMSIS-DAP: Interface Initialised (SWD)                                                                                            │
Info : CMSIS-DAP: FW Version = 1.0                                                                                                       │
Info : SWCLK/TCK = 1 SWDIO/TMS = 1 TDI = 0 TDO = 0 nTRST = 0 nRESET = 1                                                                  │
Info : CMSIS-DAP: Interface ready                                                                                                        │
Info : clock speed 1000 kHz                                                                                                              │
Info : SWD DPIDR 0x0bb11477                                                                                                              │
Info : nrf51.cpu: hardware has 4 breakpoints, 2 watchpoints                                                                              │
    TargetName         Type       Endian TapName            State                                                                        │
--  ------------------ ---------- ------ ------------------ ------------                                                                 │
 0* nrf51.cpu          cortex_m   little nrf51.cpu          unknown                                                                      │
target halted due to debug-request, current mode: Thread                                                                                 │
xPSR: 0xc1000000 pc: 0x00000440 msp: 0x20000200                                                                                          │
auto erase enabled                                                                                                                       │
Warn : Unknown device (HWID 0x0000008f)                                                                                                  │
Warn : using fast async flash loader. This is currently supported                                                                        │
Warn : only with ST-Link and CMSIS-DAP. If you have issues, add                                                                          │
Warn : "set WORKAREASIZE 0" before sourcing nrf51.cfg/nrf52.cfg to disable it                                                            │
wrote 30720 bytes from file /home/hauke/dev/riot/RIOT/examples/default/bin/microbit/default.elf in 3.073043s (9.762 KiB/s)               │
verified 30168 bytes in 0.348019s (84.653 KiB/s)                                                                                         │
shutdown command invoked                                                                                                                 │
Done flashing 
```

boards/microbit/dist/openocd.cfg
```diff
@@ -0,0 +1,3 @@
+transport select swd
```
haukepetersen: this file is basically a duplication of `airfy-beacon/dist/openocd.cfg`, `yunjia-nrf51822/dist/openocd.cfg`, and possibly more. Cant we better put it in some shared location for all of them?!

aabadie: > Cant we better put it in some shared location for all of them?!

See #10243, this is a first step.

haukepetersen: nice one. So would you mind to get #10243 in first, adapt this PR, and then merge this PR? I will actively help to get this all done tomorrow :-)

smlng:
I would agree with @haukepetersen to keep the `fscopy` variant still in, but put `OpenOCD` as default. As pointed out it is unclear if older OpenOCD version support nrf51 boards.

aabadie:
Note that this PR is now superseded by #10243.

>  keep the fscopy variant still in, but put OpenOCD as default

#10243 keeps fscopy and just add OpenOCD as an alternative. This can still be changed but if we want to keep it as simple as possible for newcomers (nothing to install for flashing the board), then having fscopy by default makes more sense. If we have openocd by default, newcomers will have to know how to override the PROGRAMMER variable, which is not documented anywhere.

> it is unclear if older OpenOCD version support nrf51 boards

It is a least supported by openocd 0.10 which is the current version installed by Ubuntu.

PR #9374, opened 2018-06-19
sys/random: add SHA1PRNG

PeterKietzmann:
### Contribution description

This PR adds a SHA1 based pseudo random number generator referring to the ~~[Java implementation](http://www.docjar.net/html/api/gnu/java/security/provider/SHA1PRNG.java.html)~~
Edit: [This Java Implementation](https://github.com/dmlloyd/openjdk/blob/342a565a2da8abd69c4ab85e285bb5f03b48b2c9/src/java.base/share/classes/sun/security/provider/SecureRandom.java)

In future work I'd like to enhance the random [auto-initialization](https://github.com/RIOT-OS/RIOT/blob/master/sys/random/random.c#L30) in order to easily seed with longer sequences than 32-bit as well as getting random arrays from the PRNG itself instead of using the common [`random_bytes`](https://github.com/RIOT-OS/RIOT/blob/master/sys/random/random.c#L43) function.

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -631,6 +631,10 @@ ifneq (,$(filter random,$(USEMODULE)))
     USEMODULE += tinymt32
   endif
 
+  ifneq (,$(filter prng_sha1prng,$(USEMODULE)))
+    USEMODULE += hashes
+  endif
+
   USEMODULE += luid
 endif
 
--- a/sys/random/sha1prng.c
+++ b/sys/random/sha1prng.c
@@ -1,0 +1,143 @@
+/*
+ * Copyright (C) 2018 HAW Hamburg
+ *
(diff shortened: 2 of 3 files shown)


tcschmidt:
@cgundogan just for the sake of correctness: should test the SHA-1 implementation against an official testside.

bergzand:
@tcschmidt The The SHA1 module from RIOT is also used in the [UUID module](https://github.com/RIOT-OS/RIOT/pull/9046) [1]. That one generates results identical to the UUID module from the python standard library.

tcschmidt:
@PeterKietzmann you are talking about a test suite for random number generators? That exists from NIST and should probably just be used.
However, in most cases we don't test the algorithm, but only the implementation - meaning correctness, robustness, and efficiency.

PR #9243, opened 2018-05-30
makefiles/cflags.inc.mk: add -Wmissing-include-dirs flag 

cladmi:
### Contribution description

Warn if a user-supplied include directory does not exist.

### Issues/PRs references

None

PR #9384, opened 2018-06-20
newlib.mk: replace the error by a warning on nano header not found

cladmi:
### Contribution description

Because of https://github.com/RIOT-OS/RIOT/pull/9243, failing to find the proper newlib-nano include directory triggered an error as the given newlib-nano include directory does not exist.

This PR does:

 * only add the NEWLIB_NANO_INCLUDE_DIR to INCLUDES if it exists
 * show a warning if the directory does not exist but still keeps going
   as it was the previous behavior.

### Warning

It does not solve any issue with `newlib-nano/newlib.h` not being included but just prints a warning and keeps building without nano headers as it was the behavior before #9243

### Issues/PRs references

Mentionned in:
* https://github.com/RIOT-OS/RIOT/issues/9381
* https://github.com/RIOT-OS/RIOT/pull/9216
* https://github.com/RIOT-OS/RIOT/pull/9243 post merge feedback

### DIFF
--- a/makefiles/libc/newlib.mk
+++ b/makefiles/libc/newlib.mk
@@ -78,7 +78,7 @@ ifeq (1,$(USE_NEWLIB_NANO))
   # newlib-nano overrides newlib.h and its include dir should therefore go before
   # the regular system include dirs.
   ifeq (,$(wildcard $(NEWLIB_NANO_INCLUDE_DIR)))
-    # HACK until this is really fixed. It keeps the old build system behaviour
+    # TODO: HACK until this is really fixed. It keeps the old build system behaviour
     # but removes the error on non existing directory
     $(warning $(lastword $(MAKEFILE_LIST)) could not correctly find include path to newlib-nano headers.)
     $(warning It was set to $(NEWLIB_NANO_INCLUDE_DIR) which does not exist on your system.)
--- a/makefiles/libc/newlib.mk
+++ b/makefiles/libc/newlib.mk
@@ -77,7 +77,15 @@ ifeq (1,$(USE_NEWLIB_NANO))
                                                     $(NEWLIB_INCLUDE_DIR)nano))
   # newlib-nano overrides newlib.h and its include dir should therefore go before
   # the regular system include dirs.
-  INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(INCLUDES)
+  ifeq (0,$(words $(wildcard $(NEWLIB_NANO_INCLUDE_DIR))))
+    # HACK until this is really fixed. It keeps the old build system behaviour
+    # but removes the error on non existing directory
+    $(warning $(lastword $(MAKEFILE_LIST)) could not correctly find include path to newlib-nano headers.)
+    $(warning It was set to $(NEWLIB_NANO_INCLUDE_DIR) which does not exist on your system.)
+    $(warning The non newlib-nano newlib.h may be used)
+  else
+    INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(INCLUDES)
+  endif
 endif
 
 # Newlib includes should go before GCC includes. This is especially important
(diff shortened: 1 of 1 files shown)


makefiles/libc/newlib.mk
```diff
@@ -67,7 +67,15 @@ ifeq (1,$(USE_NEWLIB_NANO))
   NEWLIB_NANO_INCLUDE_DIR ?= $(NEWLIB_INCLUDE_DIR)/newlib-nano
   # newlib-nano overrides newlib.h and its include dir should therefore go before
   # the regular system include dirs.
-  INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(INCLUDES)
+  ifeq (0,$(words $(wildcard $(NEWLIB_NANO_INCLUDE_DIR))))
```
gebart:
```make
ifeq (,$(wildcard $(NEWLIB_NANO_INCLUDE_DIR)))
```
should be enough, is it not?

makefiles/libc/newlib.mk
```diff
@@ -67,7 +67,15 @@ ifeq (1,$(USE_NEWLIB_NANO))
...
+    # HACK until this is really fixed. It keeps the old build system behaviour
+    # but removes the error on non existing directory
+    $(warning $(lastword $(MAKEFILE_LIST)) could not correctly find include path to newlib-nano headers.)
+    $(warning It was set to $(NEWLIB_NANO_INCLUDE_DIR) which does not exist on your system.)
+    $(warning The non newlib-nano newlib.h may be used)
```
gebart: Suggestion "The regular non-nano newlib.h configuration header may be used instead, with unknown side effects"

makefiles/libc/newlib.mk
```diff
@@ -67,7 +67,15 @@ ifeq (1,$(USE_NEWLIB_NANO))
   NEWLIB_NANO_INCLUDE_DIR ?= $(NEWLIB_INCLUDE_DIR)/newlib-nano
   # newlib-nano overrides newlib.h and its include dir should therefore go before
   # the regular system include dirs.
-  INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(INCLUDES)
+  ifeq (0,$(words $(wildcard $(NEWLIB_NANO_INCLUDE_DIR))))
+    # HACK until this is really fixed. It keeps the old build system behaviour
+    # but removes the error on non existing directory
+    $(warning $(lastword $(MAKEFILE_LIST)) could not correctly find include path to newlib-nano headers.)
+    $(warning It was set to $(NEWLIB_NANO_INCLUDE_DIR) which does not exist on your system.)
```
gebart: add quotes around `$(NEWLIB_NANO_INCLUDE_DIR)` so that it is obvious if it is empty as well

makefiles/libc/newlib.mk
```diff
@@ -67,7 +67,15 @@ ifeq (1,$(USE_NEWLIB_NANO))
   NEWLIB_NANO_INCLUDE_DIR ?= $(NEWLIB_INCLUDE_DIR)/newlib-nano
   # newlib-nano overrides newlib.h and its include dir should therefore go before
   # the regular system include dirs.
-  INCLUDES := -isystem $(NEWLIB_NANO_INCLUDE_DIR) $(INCLUDES)
+  ifeq (0,$(words $(wildcard $(NEWLIB_NANO_INCLUDE_DIR))))
+    # HACK until this is really fixed. It keeps the old build system behaviour
```
cgundogan: could you add a `TODO:` here? It is used at several other places and easy to grep for.

makefiles/libc/newlib.mk
cladmi: I agree.

I used the `0, $(words` without thinking because in this PR https://github.com/RIOT-OS/RIOT/pull/8475 the last change was not working for @aabadie with `wildcard` (it was working on my computer).

cladmi:
Thanks for your feedback, I addressed each comment in a separate fixup commit.

cladmi:
An idea could be to tell in the warning to set `NEWLIB_NANO_INCLUDE_DIR` to the right directory for their setup.

cgundogan:
@cladmi do we still need this? I like to have the build system fail for the case that newlib-nano couldn't be found. It certainly is an error. With the current fixes upstream (and with #9415) I don't think, that decent systems will fail in finding the newlib.h for nano. Except when the system is wrongly configured (no nano installed, etc.)

PR #9392, opened 2018-06-21
drivers/isl29125: apply unified params definition scheme and new i2c API

MichelRottleuthner:
### Contribution description
Analog to #8688 this updates the parameter scheme (#7519) and applies the new i2c API to the isl29125 driver.

tested with sparkfun isl29125 with nucleo-l073

### Issues/PRs references
#6577

### DIFF
--- a/drivers/isl29125/include/isl29125_params.h
+++ b/drivers/isl29125/include/isl29125_params.h
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2017 Freie Universität Berlin
+ * Copyright (C) 2018 HAW-Hamburg
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_isl29125
+ *
+ * @{
+ * @file
+ * @brief       Default configuration for ISL29125 devices
+ *
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Michel Rottleuthner <michel.rottleuthner@haw-hamburg.de>
+ */
+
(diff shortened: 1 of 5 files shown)


dylad:
This looks good. I'm supposed to receive custom board with this sensor very soon so I'll give it a try asap unless someone is faster than me.

drivers/isl29125/include/isl29125_params.h
```diff
@@ -0,0 +1,73 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Set default configuration parameters
+ * @{
+ */
+#ifndef ISL29125_PARAM_I2C
+#define ISL29125_PARAM_I2C              I2C_DEV(0)
```
MrKevinWeiss: Probably should == TEST_ISL29125_I2C

drivers/isl29125/include/isl29125_params.h
```diff
@@ -0,0 +1,73 @@
...
+#endif
+#ifndef ISL29125_PARAM_GPIO
+#define ISL29125_PARAM_GPIO             (GPIO_PIN(0, 0))
```
MrKevinWeiss: Same here TEST_ISL29125_IRQ_PIN

drivers/isl29125/include/isl29125_params.h
MichelRottleuthner: I removed the parameters from the tests Makefile so there is only the one definition in this file, which can still be overwritten.

drivers/isl29125/include/isl29125_params.h
MichelRottleuthner: addressed: same as above

MrKevinWeiss:
ACK, seems good to me and appears to work!

PR #9396, opened 2018-06-22
tests/unittests, cn_cbor: remove invisible unicode char from Makefile

smlng:
### Contribution description

fixes/removes an invisible uincode char in the Makefile of `tests-cn_cbor` which causes build failure on macOS

```
-<U+FEFF>include $(RIOTBASE)/Makefile.base
```


smlng:
thx @bergzand for acting swiftly 👍

cladmi:
Is there another check that could be added to `dist/tools/whitespacecheck/check.sh` to prevent these ?

smlng:
@cladmi: I used `pcregrep --color='auto' -n "[\x80-\xFF]" <file>`, maybe add that to a check - also see #9397

Issue #9401, opened 2018-06-22
ARM: Non-upstreamed openocd configs need to adopt new dap support


pyropeter:
OpenOCD has changed their dap support and that requires all configs for ARM targets to be modified. More information here: http://openocd.zylin.com/#/c/4468/

The change was merged, but OpenOCD was not released since then, so there is still time to address this.

This only affects a few boards, as most boards use the configs contained in OpenOCD. Maybe it would be a good idea to push upstreaming efforts for the remaining board configs?

Affected boards:
* [ ] cc2650stk
* [ ] cc2650-launchpad
* [x] pba-d-01-kw2x, #9406
* [ ] lpc2387
* [x] All frdm boards with `USE_OLD_OPENOCD` (right?)

gebart:
> - All frdm boards with USE_OLD_OPENOCD (right?)

If you have this new version of openocd you have, by definition, not an old openocd, so you should set USE_OLD_OPENOCD=0. We could improve it and let it be determined by a check of the openocd version. Help wanted.

PR #9406, opened 2018-06-25
boards/pba-d-01-kw2x: use OpenOCD-supplied config

pyropeter:
This partially addresses #9401.

I tested this with OpenOCD `0.10.0.r439.gc809f024` (from GIT), which is really cutting-edge. The referenced file `kx.cfg` exists in the OpenOCD GIT since 2015-01-10 and should be contained in OpenOCD `>= v0.9.0`.

I'm unsure if this will break the board support for people with older OpenOCD, because I don't know which OpenOCD releases even support this board.


A-Paul:
Hi @pyropeter, you mentioned that you already made tests. Can you give a short overview what you have tested (flash, debug)?

pyropeter:
> Hi @pyropeter, you mentioned that you already made tests. Can you give a short overview what you have tested (flash, debug)?

I tested `make flash` and `make debug`.

pyropeter:
Yes. The Kinetis support in OpenOCD got fixed, refactored and expanded a lot over the last few years and I didn't manage to find out if OpenOCD < 0.9.0 even supports this CPU. But I only spent about ten minutes researching that. The OpenOCD commit messages do not clearly state the impact of their changes, so it is quite hard to do.

PR #9408, opened 2018-06-25
cpu/stm_common: fix i2c when address is NACKed

MrKevinWeiss:
### Contribution description

Fixes the fact that the stop bit does not get sent if address is NACKed.  As well as fixes flag issues when using read bytes.  There is still extra reads that go on but from the software it seems to work.  

### Issues/PRs references
See #6577

### DIFF
--- a/cpu/stm32_common/periph/i2c_2.c
+++ b/cpu/stm32_common/periph/i2c_2.c
@@ -220,24 +220,31 @@ int i2c_read_bytes(i2c_t dev, uint16_t address, void *data, size_t length,
     assert(i2c != NULL);
 
     int ret = 0;
-
     if (!(flags & I2C_NOSTART)) {
         DEBUG("[i2c] read_bytes: Send Slave address and wait for ADDR == 1\n");
         ret = _start(i2c, address, I2C_FLAG_READ, flags);
         if (ret < 0) {
             return ret;
         }
+        _clear_addr(i2c);
+        if (length == 1 && !(flags & I2C_NOSTOP)) {
+            DEBUG("[i2c] read_bytes: Set ACK = 0\n");
+            i2c->CR1 &= ~(I2C_CR1_ACK);
+        }
+        else {
+            i2c->CR1 |= I2C_CR1_ACK;
+        }
+        _clear_addr(i2c);
     }
-
-    if (length == 1) {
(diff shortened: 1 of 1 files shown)


cpu/stm32_common/periph/i2c_2.c
```diff
@@ -220,25 +220,32 @@ int i2c_read_bytes(i2c_t dev, uint16_t address, void *data, size_t length,
...
-        i2c->CR1 &= ~(I2C_CR1_ACK);
+        _clear_addr(i2c);
+        if (length == 1 && !(flags & I2C_NOSTOP)) {
+            DEBUG("[i2c] read_bytes: Set ACK = 0\n");
+            i2c->CR1 &= ~(I2C_CR1_ACK);
+        }
+        else {
+            i2c->CR1 |= I2C_CR1_ACK;
+        }
+        _clear_addr(i2c);
```
smlng: is it necessary to do this twice, i.e., before and after this `if`?

MrKevinWeiss: Yup, good catch.  I got rid of the first one since it reflects the old code closer.

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -430,6 +435,7 @@ static int _start(I2C_TypeDef *i2c, uint8_t address, uint8_t rw_flag, uint8_t fl
     /* Wait for ADDR flag to be set */
     while (!(i2c->SR1 & I2C_SR1_ADDR) && tick--) {
         if ((i2c->SR1 & ERROR_FLAG) || !tick) {
+            i2c->CR1 |= I2C_CR1_STOP;
```
aabadie: According to the API, the return code below should be -EIO I think.

MrKevinWeiss:
So maybe we can merge, it isn't perfect but it is an improvement.

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -430,6 +435,7 @@ static int _start(I2C_TypeDef *i2c, uint8_t address, uint8_t rw_flag, uint8_t fl
...
             return -ETIMEDOUT;
```
aabadie: I was talking of this line for the return code.

PR #9409, opened 2018-06-25
tests/periph_i2c: Add automated tests

MrKevinWeiss:
### Contribution description

This is all the components needed to run automated i2c tests.  It is not great but so far it catches quite a few issues.

### Issues/PRs references
#6577


tcschmidt:
ping @smlng

tests/periph_i2c/main.c
```diff
@@ -448,27 +404,24 @@ int cmd_i2c_write_regs(int argc, char **argv)
...
+    int dev;
 
-    if (check_dev() == 1) {
-        return 1;
-    }
-
-    if (argc < 5) {
-        INVALID_ARGS;
-        printf("Usage:\n%s ADDR REG FLAG BYTE0 [BYTE1 ...]\n", argv[0]);
+    dev = _check_param(argc, argv, 5, 5 + BUFSIZE,
```
smlng: shouldn't this `_check_param(argc, argv, 6, (5 + BUFSIZE),` because the `6` param contains the first byte, ie. the minimum. At least the help text suggests that the first byte is not optional, or not?

smlng: ah no it should be `_check_param(argc, argv, 5, (5 + BUFSIZE -1)`, bc the last param is already the first byte hence buffer must be one less, right?

tests/periph_i2c/main.c
```diff
@@ -334,60 +301,53 @@ int cmd_i2c_write_byte(int argc, char **argv)
...
+    int dev;
 
-    if (check_dev() == 1) {
-        return 1;
-    }
-
-    if (argc < 4) {
-        INVALID_ARGS;
-        printf("Usage:\n%s: ADDR FLAG BYTE0 [BYTE1 [BYTE_n [...]]]\n", argv[0]);
+    dev = _check_param(argc, argv, 4, 4 + BUFSIZE,
```
smlng: same here (see below) should be `_check_param(argc, argv, 4, (4 + BUFSIZE -1,)`

tests/periph_i2c/tests/base_if.py
```diff
@@ -0,0 +1,173 @@
+import serial
+import logging
+import errno
+import os
+import time
+import serial.tools.list_ports
+
+
+class IfParams:
```
smlng: as discussed, use dictionary instead

tests/periph_i2c/tests/base_if.py
```diff
@@ -0,0 +1,173 @@
...
+    __SUCCESS = '0'
+    __RESULT_SUCCESS = 'Success'
+    __RESULT_ERROR = 'Error'
+    __RESULT_TIMEOUT = 'Timeout'
+
+    def __init__(self, port='/dev/ttyACM0', baud=115200):
+        self.connect(port, baud)
+
+    def get_port(self):
+        if (self.__dev.isOpen()):
```
smlng: here (and below) `.isOpen()` is deprecated according to documentation, use `.is_open` instead

tests/periph_i2c/tests/base_if.py
```diff
@@ -0,0 +1,173 @@
...
+        logging.debug("Autoconnecting")
+        for element in comlist:
+            if (element.device not in used_com):
+                connected.append(element.device)
+        for port in connected:
+            logging.debug("Port: " + port)
+            self.connect(port)
+            ret = fxn().data
+            try:
+                logging.debug("ID rx: %s" % ret)
```
smlng: the *new* python way for string formatting would look like this:

```
 logging.debug("ID rx: {}".format(ret))
```

simply replace every `%s` or `%d` and such by `{}`,
that way you don't need to care about variable types (which is not very pythonic anyway)

tests/periph_i2c/tests/base_if.py
```diff
@@ -0,0 +1,173 @@
...
+                            d_len = len(data[1]) - 1
+                            cmd_info.data = bytearray.fromhex(data[1][2:d_len])
+                    except ValueError:
+                        self.data = data[1:]
+                cmd_info.msg = "EOK-command success [0]"
+                logging.debug(self.__RESULT_SUCCESS)
+                cmd_info.result = self.__RESULT_SUCCESS
+            elif (isinstance(data[0], int)):
+                cmd_info.data = int(data[0], 0)
+                cmd_info.msg = "%s-%s [%d]" % (errno.errorcode[cmd_info.data],
```
smlng: and with multiple variables 

```
cmd_info.msg = "{}-{} [{}]".format(errno.errorcode[cmd_info.data], os.strerror(cmd_info.data), cmd_info.data)
```

tests/periph_i2c/tests/bpt_if.py
```diff
@@ -0,0 +1,503 @@
...
+    def set_tmr_lo_us(self, data=0):
+        return self.write_bytes(148, data, 4)
+
+    def get_res104(self):
+        return self.read_bytes(152, 104)
+
+    def set_res104(self, data=0):
+        return self.write_bytes(152, data, 104)
+
+    def get_command_list(self):
```
smlng: what do you need this for?

tests/periph_i2c/tests/periph_i2c_if.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self, port=None, baud=115200, used_com=[]):
+
+        if (port is None):
+            self.autoconnect('periph_i2c', self.get_id, used_com)
+        else:
+            self.connect(port, baud)
+
+    def set_param(self, dev=None, addr=None, reg=None, len=None, data=None):
+        if (dev is not None):
```
smlng: why the `if` can't I set a param to `None` if it was set before?

tests/periph_i2c/tests/periph_i2c_if.py
```diff
@@ -0,0 +1,107 @@
...
+            self.__data = data
+
+    def acquire(self, dev=__dev,):
+        return self.send_cmd('i2c_acquire %d' % dev)
+
+    def release(self, dev=__dev,):
+        return self.send_cmd('i2c_release %d' % dev)
+
+    def read_reg(self, dev=__dev, addr=__addr, reg=__reg, flag=None):
+        if (flag is None):
```
smlng: Better `if not flag: ` this matches `None` and `0` (and even `False`)

tests/periph_i2c/tests/periph_i2c_if.py
```diff
@@ -0,0 +1,107 @@
...
+    def write_byte(self, dev=__dev, addr=__addr, data=__data, flag=None):
+        if (flag is None):
+            return self.send_cmd('i2c_write_byte %d %d %d' % (dev, addr, data))
+        else:
+            return self.send_cmd('i2c_write_byte \
+            %d %d %d %d' % (dev, addr, flag, data))
+
+    def write_bytes(self, dev=__dev, addr=__addr, data=__data, flag=None):
+        str = ''
+        for val in data:
```
smlng: there is a short hand for that:
```
str = ' '.join(str(x) for x in data)
```

smlng: which you could also place directly into the commands below

tests/periph_i2c/tests/test.py
```diff
@@ -0,0 +1,328 @@
...
+__BPT_USER_REG = 152
+
+
+def kwexpect(val1, val2, level="WARN"):
+    res = level
+    try:
+        if (isinstance(val1, str) and isinstance(val2, str)):
+            if (val1.lower() == val2.lower()):
+                res = "PASS"
+        elif (isinstance(val1, list)):
```
smlng: you assume val2 is a list, too?

tests/periph_i2c/tests/test.py
```diff
@@ -0,0 +1,328 @@
...
+                return [level, val1, val2]
+        elif ((val1 == val2) or (val1 is val2)):
+            res = "PASS"
+    except TypeError:
+        res = 'TypeError'
+
+    return [res, val1, val2]
+
+
+class TestParam:
```
smlng: might also wanna use dictionary here?

tests/periph_i2c/tests/test.py
```diff
@@ -0,0 +1,328 @@
...
+
+def check_write_regs(t, d_num, addr, reg, val):
+    t.run_test(i2c.write_regs(d_num, addr, reg, val), "Success")
+    t.run_test(i2c.read_regs(d_num, addr, reg, len(val)), "Success", val)
+
+
+def write_test():
+    cmd_log = list()
+    t = Test('write test', 'Tests the write functionality for default i2c bus \
+(0) with the bpt', cmd_log)
```
smlng: indent please

tests/periph_i2c/tests/test.py
```diff
@@ -0,0 +1,328 @@
...
+    t.skip_test("i2c.read_bytes(d_num, __BPT_ADDR, 1, 4), Error, [0]")
+    t.skip_test("RESET DUT")
+    t.run_test(i2c.read_bytes(d_num, __BPT_ADDR, 1, 1))
+
+    return t
+
+
+def print_full_result(test):
+    print('===================================================================\
+=============')
```
smlng: minor, but this looks a bit ugly - can't we live we no line break or less `====` to fit 80 chars?

smlng:
@MrKevinWeiss my comment on the python stuff are more informational, hence non blocking. But please have a look at the C stuff specifically the numbers for the argv checker - might be a possible buffer overflow?

tests/periph_i2c/tests/test.py
MrKevinWeiss: Yup but why check, if it isn't it will just get caught in the exception.  Also this is a throwaway test, only as an example.

tests/periph_i2c/tests/test.py
MrKevinWeiss: Yup but I don't want to spend more time on this.

tests/periph_i2c/tests/test.py
MrKevinWeiss: Then the output includes the indents I think...

tests/periph_i2c/tests/test.py
MrKevinWeiss: Again throw away so it doesn't matter.

MrKevinWeiss:
I disagree with the Travis errors, the training white-spaces are a feature of the .md (so that it can have a newline) and the duplicated if else are not touched by me (as well as serve an actual purpose if certain flags are set).

tests/periph_i2c/tests/README.md
```diff
@@ -0,0 +1,15 @@
+How To Run
+==========
+Connect the DUT I2C0 to the BPT.  Information for BPT setup can be found [here](https://github.com/MrKevinWeiss/Testing)
```
aabadie: DUT and BPT acronyms should be described first here.

aabadie:
> the training white-spaces are a feature of the .md

Indeed, but I think (personal preference) that it's sometime better to add an extra newline in this case. If you strongly disagree, you can try to patch the trailing white spaces checker script to avoid this particular case in .md files.

cladmi:
As we did some python reviews IRL yesterday and found that the API would still need some adapting.
I will add the WIP label, this way you can rebase/squash if you want to until it stabilize.

MrKevinWeiss:
Thanks, I will separate out the connection soon...

MrKevinWeiss:
@smlng Ready for review then I will squash and rebase.

smlng:
@dylad if you agree, we should get this merged swiftly. Besides adapting the test to the new API it also adds an automated test script which is very helpful.

dylad:
@smlng If this PR is mature enough feeling free to ACK/merge

MrKevinWeiss:
@smlng what a tempting big green button...  Don't you just want to see what happens when you press it?

Issue #8778, opened 2018-03-14
sys/random: random_uint32_range is only uniform if range is a power of 2


gebart:
#### Description

It does not say so explicitly in the documentation, but most users would assume that this function will return a uniform distributed random number in the interval [a,b).

```c
static inline uint32_t random_uint32_range(uint32_t a, uint32_t b)
{
    return (random_uint32() % (b - a)) + a;
}
```

However, the assumption is only true if `b - a` is a power of two, because of the discrete values (integer) involved. If `b - a` is _not_ a power of two, then the output will be weighted towards the lower end of the range, depending on the distance between `b - a` and the closest power of two less than `b - a`.

#### Steps to reproduce the issue

```c
uint32_t val = random_uint32_range(0, 0xC0000000);
```

#### Expected results

`val` above should have a uniform distribution

#### Actual results

The distribution of `val` above will have more weight in the first quartile.

#### Versions

Current RIOT master branch

gebart:
I will fix and add a unit test when I find some time

kaspar030:
how did you catch this? :) any news?

gebart:
Caught it on paper :) I did some calculations because the expression in the code looked non-uniform. This is really only an issue when the reduced range of numbers is large, otherwise the difference is pretty much impossible to detect from samples.

kaspar030:
I'll untag for the release, as I don't think we'd block the release for this.

tcschmidt:
Fixed with #9410.

PR #9410, opened 2018-06-25
sys/random: fix distribution of random_uint32_range() 

PeterKietzmann:
### Contribution description
#8778 depicts that  [`random_uint32_range()`](https://github.com/RIOT-OS/RIOT/blob/master/sys/include/random.h#L95) does not return uniformly distributed numbers in the desired interval. This PR adds a fix that provides uniform distributions (as displayed below) but comes with a certain performance regression.

### Issues/PRs references
 #8778

#### Native with `prng_xorshift`, [0,0xC0000000)-interval, 50k numbers

Old:
old

With fix:
new

#### samr21-xpro with `prng_xorshift`, [0,0xC0000000)-interval, 10seconds
Old: 244 KiB/s
With fix: 213 KiB/s

#### samr21-xpro with `prng_xorshift`, [0,0x80000001)-interval, 10seconds
Old: 244 KiB/s
With fix: 206 KiB/s

#### samr21-xpro with `prng_xorshift`, [0,0x80000000)-interval, 10seconds
Old: 244 KiB/s
With fix: 217 KiB/s

#### samr21-xpro with `prng_xorshift`, [0x18000)-interval, 10seconds
Old: 244 KiB/s
With fix: ~~168 KiB/s~~ 202 KiB/s

#### samr21-xpro with `prng_xorshift`, [0x10000)-interval, 10seconds
Old: 244 KiB/s
With fix: ~~186 KiB/s~~ 217 KiB/s

#### samr21-xpro with `prng_tinymt32`, [0,0xC0000000)-interval, 10seconds
Old: 229 KiB/s
With fix: 198 KiB/s

#### samr21-xpro with `prng_tinymt32`, [0,0x18000)-interval, 10seconds
Old: 229 KiB/s
With fix: ~~151 KiB/s~~ 180 KiB/s

#### samr21-xpro with `prng_tinymt32`, [0,0x10000)-interval, 10seconds
Old: 229 KiB/s
With fix: ~~177 KiB/s~~ 205 KiB/s

### DIFF
--- a/sys/random/random.c
+++ b/sys/random/random.c
@@ -54,3 +54,30 @@ void random_bytes(uint8_t *target, size_t n)
         *target++ = *random_pos++;
     }
 }
+
+uint32_t random_uint32_range(uint32_t a, uint32_t b)
+{
+    assert(a < b);
+
+    uint32_t divisor, rand_val, range = b - a;
+    uint8_t range_msb = bitarithm_msb(range);
+
+    /* check if range is a power of two */
+    if (!(range & (range - 1))) {
+        divisor = (1 << range_msb) - 1;
+    }
+    else if (range_msb < 31) {
+        /* leftshift for next power of two interval */
+        divisor = (1 << (range_msb + 1)) -1;
+    }
+    else {
+        /* disable modulo operation in loop below */
+        divisor = UINT32_MAX;
+    }
+    /* get random numbers until value is smaller than range */
+    do {
+        rand_val = (random_uint32() & divisor);
+    } while (rand_val >= range);
+    /* return random in range [a,b] */
+    return (rand_val + a);
+}
--- a/tests/rng/main.c
+++ b/tests/rng/main.c
@@ -90,12 +90,26 @@ static int cmd_dump(int argc, char **argv)
 {
     uint32_t samples = 100;
 
-    if (argc > 1) {
+    if (argc < 2) {
+        /* run the test */
+        test_dump(samples);
+    }
+    else if (argc == 2) {
         samples = strtoul(argv[1], NULL, 0);
+        /* run the test */
+        test_dump(samples);
+    }
+    else if (argc == 4) {
+        samples = strtoul(argv[1], NULL, 0);
+        uint32_t low_thresh = strtoul(argv[2], NULL, 0);
+        uint32_t high_thresh = strtoul(argv[3], NULL, 0);
+        /* run the test */
+        test_dump_range(samples, low_thresh, high_thresh);
+    }
+    else {
+        printf("usage: %s [samples] [lower-bound upper-bound]\n", argv[0]);
     }
 
-    /* run the test */
-    test_dump(samples);
 
     return 0;
 }
@@ -220,12 +234,25 @@ static int cmd_speed(int argc, char **argv)
 {
     uint32_t duration = 10;
 
-    if (argc > 1) {
-        duration = strtoul(argv[1], NULL, 0);
+    if (argc < 2) {
+        /* run the test */
+        test_speed(duration);
     }
-
-    /* run the test */
-    test_speed(duration);
+    else if (argc == 2) {
+        duration = strtoul(argv[1], NULL, 0);
+        /* run the test */
+        test_speed(duration);
+    }
+    else if (argc == 4) {
+        duration = strtoul(argv[1], NULL, 0);
+        uint32_t low_thresh = strtoul(argv[2], NULL, 0);
+        uint32_t high_thresh = strtoul(argv[3], NULL, 0);
+        /* run the test */
+        test_speed_range(duration, low_thresh, high_thresh);
+    }
+    else {
+        printf("usage: %s [duration] [lower-bound upper-bound]\n", argv[0]);
+    }
 
     return 0;
 }
--- a/tests/rng/test.c
+++ b/tests/rng/test.c
@@ -104,6 +104,27 @@ static inline uint32_t test_get_uint32(void)
 }
 
 /**
+ * @brief   Retrieve a 32-bit number of the source RNG on [a,b)-interval
+ */
+static inline uint32_t test_get_uint32_range(uint32_t a, uint32_t b)
+{
+    if (source == RNG_PRNG) {
+        return random_uint32_range(a, b);
+    }
+#ifdef MODULE_PERIPH_HWRNG
+    else if (source == RNG_HWRNG) {
+        puts("Range feature not supported by HWRNG");
+    }
+#endif
+    else if (source == RNG_CONSTANT) {
+        /* use the seed as the constant value */
+        return seed;
+    }
+
+    return 0;
+}
+
+/**
  * @brief   Helper for printing `passed` or `failed` depending on condition
  *
  * @param[in] condition     The test condition.
@@ -189,6 +210,15 @@ void test_dump(uint32_t samples)
 
     while (samples--) {
         printf("%" PRIu32 "\n", test_get_uint32());
+    }
+}
+
+void test_dump_range(uint32_t samples, uint32_t low_thresh, uint32_t high_thresh)
+{
+    test_init("dump range");
+
+    while (samples--) {
+        printf("%" PRIu32 "\n", test_get_uint32_range(low_thresh, high_thresh));
     }
 }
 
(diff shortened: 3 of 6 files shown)


sys/random/random.c
```diff
@@ -54,3 +54,28 @@ void random_bytes(uint8_t *target, size_t n)
...
+    else {
+        /* get random numbers until value is smaller than range */
+        while (random >= range) {
+            random = random_uint32();
+            /* get new value from next power of two interval
+             * else, get new random number */
+            if (highest_pow2 << 1) {
+                random = (random % (highest_pow2 << 1) );
+            }
+            /* get new value in same interval if max. already reached */
```
basilfx: Is there a line missing, or should this comment move to line 71?

tests/rng/main.c
```diff
@@ -90,12 +90,26 @@ static int cmd_dump(int argc, char **argv)
...
+    }
+    else if (argc == 4) {
         samples = strtoul(argv[1], NULL, 0);
+        uint32_t low_thresh = strtoul(argv[2], NULL, 0);
+        uint32_t high_thresh = strtoul(argv[3], NULL, 0);
+        /* run the test */
+        test_dump_range(samples, low_thresh, high_thresh);
+    }
+    else {
+        printf("usage: %s [samples] [lower-bound upper-bound]", argv[0]);
```
basilfx: Missing newline.

tests/rng/main.c
```diff
@@ -220,12 +234,25 @@ static int cmd_speed(int argc, char **argv)
...
+    }
+    else if (argc == 4) {
+        duration = strtoul(argv[1], NULL, 0);
+        uint32_t low_thresh = strtoul(argv[2], NULL, 0);
+        uint32_t high_thresh = strtoul(argv[3], NULL, 0);
+        /* run the test */
+        test_speed_range(duration, low_thresh, high_thresh);
+    }
+    else {
+        printf("usage: %s [duration] [lower-bound upper-bound]", argv[0]);
```
basilfx: Missing newline.

sys/random/random.c
PeterKietzmann: leftover

basilfx:
My comments have been addressed. I have tested this, and looks good.

sys/random/random.c
```diff
@@ -54,3 +54,27 @@ void random_bytes(uint8_t *target, size_t n)
...
 }
+
+uint32_t random_uint32_range(uint32_t a, uint32_t b)
+{
+    uint32_t range = b - a;
+    uint32_t highest_pow2 = (1 << bitarithm_msb(range));
+    uint32_t random = random_uint32();
+
+    /* check if range is a power of two */
+    if (!(range % highest_pow2)) {
```
cgundogan: the compiler won't be able to optimize the modulo operation. To check whether `range` is a power of two, you could use
```c
range && !( range & ( range - 1))
```

sys/random/random.c
```diff
@@ -54,3 +54,27 @@ void random_bytes(uint8_t *target, size_t n)
...
+        return (random % range) + a;
+    }
+    else {
+        /* get random numbers until value is smaller than range */
+        while (random >= range) {
+            random = random_uint32();
+            /* get new value from next power of two interval
+             * else, get new random number */
+            if (highest_pow2 << 1) {
```
cgundogan: no need to calculate `highest_pow2 << 1` with each loop cycle. You could shift `highest_pow2` once, before entering the loop.

sys/random/random.c
```diff
@@ -54,3 +54,27 @@ void random_bytes(uint8_t *target, size_t n)
...
+                random = (random % (highest_pow2 << 1) );
```
cgundogan: since `highest_pow2 << 1` is again a power of two, you can use binary `and` instead of an expensive modulo operation

tcschmidt: ... and then you can move the if clause out of the loop, by simply setting highest_pow2 to zero 
if (!(highest_pow2 << 1))

PeterKietzmann:
@cgundogan, @tcschmidt thanks for your reviews. Please check my latest commit. Results seem ok at first sight: distributions look uniform and with xorshift, [0x18000)-interval on a samr21-xpro we get 202 KiB/s (before: 169KiB/s). Will report some further results tomorrow.

gebart:
The change looks OK. What is the worst case performance like? 
range = 0x80000001 should be particularly bad, but any power of two + a small number should be bad. We should see approximately 50 % of the throughput compared to the performance of a power of two range.

PeterKietzmann:
> range = 0x80000001 should be particularly bad

I've added the numbers above. Basically this is 206 vs 244 KiB/s.  @gebart why do you expect 50%?

gebart:
@PeterKietzmann it was a wrong assumption. But what is happening at 0x80000001 is that approximately half of the generated numbers will be discarded

PeterKietzmann:
Added the assert. @smlng, @cgundogan please re-check the code and remove your change requests (if addressed)

sys/random/random.c
```diff
@@ -54,3 +54,36 @@ void random_bytes(uint8_t *target, size_t n)
...
 }
+
+uint32_t random_uint32_range(uint32_t a, uint32_t b)
+{
+    uint32_t range = b - a;
+    uint32_t highest_pow2 = (1 << bitarithm_msb(range));
+    uint32_t random = random_uint32();
+
+    /* check if range is a power of two */
+    if (range && !(range & (range - 1))) {
```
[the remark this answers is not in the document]
PeterKietzmann: > [edit2]: but if range is pow2 wouldn't that mean range == highest_pow2 then? If so it might be even more efficient to compare those instead of the bit operation?!

Tried that, but results are same (well, "my" solution produced even two more numbers in 20 seconds, but I guess this is below any optimization influence anyway)

sys/random/random.c
```diff
@@ -54,3 +54,36 @@ void random_bytes(uint8_t *target, size_t n)
...
+uint32_t random_uint32_range(uint32_t a, uint32_t b)
+{
+    assert(a < b);
+
+    uint32_t range = b - a;
+    uint32_t highest_pow2 = (1 << bitarithm_msb(range));
+    uint32_t random = random_uint32();
+
+    /* check if range is a power of two */
+    if (!(range & (range - 1))) {
```
smlng: I will not block this, but IMHO the code could gain readability by improving its structure. What I mean is that it essentially comes to:

1. set a proper `highest_pow2` 
2. generate a random number (fitting into range), and 
3. return it `+ a`. 

For instance something like:

```
uint32_t random_uint32_range(uint32_t a, uint32_t b)
{
    assert(a < b);

    uint32_t highest_pow2, range = b - a;
    uint8_t range_msb = bitarithm_msb(range);

    /* check if range is a power of two */
    if (!(range & (range - 1))) {
        highest_pow2 = (1 << range_msb) - 1;
    }
    else if (range_msb < 31) {
        /* leftshift for next power of two interval */
        highest_pow2 = (1 << (range_msb + 1)) -1;
    }
    else {
        /* disable modulo operation in loop below */
        highest_pow2 = UINT32_MAX;
    }
    /* get random numbers until value is smaller than range */
    do {
        random = (random_uint32() & highest_pow2);
    } while (random >= range);
    /* return random in range [a,b] */
    return (random + a);
}
```

smlng: hope I didn't changed the logic ...

sys/random/random.c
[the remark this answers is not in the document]
smlng: yeah I think nowadays its hard to outsmart the compiler 😉 So best we can do is go for well documented and structured code.

sys/random/random.c
smlng: oops forgot to define random, add change:

```
uint32_t highest_pow2, random, range = b - a;
```

PeterKietzmann:
@smlng even though it's just a matter of taste, I've adopted your proposal regarding code structuring to have you on my side. Distribution as well as performance tests  didn't (really) change. Can I squash?

smlng:
> to have you on my side

no need for that, but appreciated 😄 please squash

sys/random/random.c
```diff
@@ -54,3 +54,30 @@ void random_bytes(uint8_t *target, size_t n)
...
+uint32_t random_uint32_range(uint32_t a, uint32_t b)
+{
+    assert(a < b);
+
+    uint32_t highest_pow2, random, range = b - a;
+    uint8_t range_msb = bitarithm_msb(range);
+
+    /* check if range is a power of two */
+    if (!(range & (range - 1))) {
+        highest_pow2 = (1 << range_msb) - 1;
```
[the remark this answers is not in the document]
PeterKietzmann: Will check the effect but if not noticeable I tend to leave as is.

sys/random/random.c
```diff
@@ -54,3 +54,30 @@ void random_bytes(uint8_t *target, size_t n)
...
+    }
+    else if (range_msb < 31) {
```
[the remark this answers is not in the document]
PeterKietzmann: Tried that before, without effect. So I decided for keeping it this way which seems more self-explaining to me.

sys/random/random.c
```diff
@@ -54,3 +54,30 @@ void random_bytes(uint8_t *target, size_t n)
         *target++ = *random_pos++;
     }
 }
+
+uint32_t random_uint32_range(uint32_t a, uint32_t b)
+{
+    assert(a < b);
+
+    uint32_t highest_pow2, random, range = b - a;
```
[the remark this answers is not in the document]
PeterKietzmann: Will change

tests/rng/test.c
```diff
@@ -378,3 +408,27 @@ void test_speed(uint32_t duration)
...
+    uint64_t timeout = 0;
+    uint64_t samples = 0;
+
+    /* initialize test */
+    test_init("speed range");
+
+    /* collect samples as long as timer has not expired */
+    timeout = xtimer_now_usec64() + (duration * US_PER_SEC);
+
+    while (xtimer_now_usec64() < timeout) {
```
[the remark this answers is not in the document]
PeterKietzmann: Agree. Maybe start/stop via msg or callback in future? IMO this is also out of scope for this PR.

PeterKietzmann:
@gebart I've updated variable names and squashed immediately. Did not update the bit-logic stuff as I couldn't measure any difference. Further improvement of the test app should be done in a separate PR.

tests/rng/test.c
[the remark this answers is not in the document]
gebart: Yes, xtimer_set is a good proposal. We can do it in a different PR

PR #9411, opened 2018-06-26
cpu/stm32f4: add support for stm32f405rg

gdoffe:
Add support for STM32F405RG

### Contribution description

I cannot share the board description but here is the output on terminal:
```
> main(): This is RIOT! (Version: 2018.07-devel-648-g97e22-gdo-sfl-laptop-stm32f405rg)
> ps
        pid | state    Q | pri 
          1 | pending  Q |  15
          2 | running  Q |   7
> 
```
- Import last header file (stm32f405xx.h) from ST and include it in cpu/stm32f4/include/cpu_conf.h
- Update vectors.c with STM32F405RG vectors
- Set ADC_DEVS to 3 in cpu/stm32f4/include/periph_cpu.h


gdoffe:
Updated with use of CPU_LINE.

vincent-d:
And go

PR #9415, opened 2018-06-26
newlib: support nano specs

toonst:
* When using newlib_nano and a nano.specs file is present and the compiler supports it, we use this file for letting the linker find the right include paths. If not, we manually try to find the right newlib (nano) paths.
* When testing for the nano.specs file, we add -Werror so compilers which don't support .specs files give the correct output.
* When none of the searched paths contain newlib.h, compilation fails.

I'm still unsure about the following points:
* When deriving the newlib include path from predefined directories or from the compiler include path, we print a warning. --> should be silent maybe?
* When using newlib_nano is specified, and we can not find a newlib nano directory, we print a warning that the regular newlib is used instead. --> should fail instead maybe?

### Issues/PRs references

Based on @cgundogan 's  #9394 
Cleanup of #9216

### DIFF
--- a/makefiles/libc/newlib.mk
+++ b/makefiles/libc/newlib.mk
@@ -1,7 +1,8 @@
 ifneq (,$(filter newlib_nano,$(USEMODULE)))
+  USE_NEWLIB_NANO = 1
   # Test if nano.specs is available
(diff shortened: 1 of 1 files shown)


cgundogan:
we probably can get rid of the predefined search paths (in the near future). I don't think they are necessary anymore..

smlng:
FYI: works on macOS

cladmi:
I added the "API Change" label for the side effects.

PR #9422, opened 2018-06-27
drivers/srf08: i2c update

MrKevinWeiss:
### Contribution description

Update the srf08 driver to new API.

### Issues/PRs references

#6577


MrKevinWeiss:
Tested with the samr21 and the srf08, it has the same results as the master.

PR #9423, opened 2018-06-27
drivers/srf02: Update to new i2c API

MrKevinWeiss:
### Contribution description

Update the srf02 sensor driver to the new I2C API.

### Issues/PRs references
#6577

### DIFF
--- a/drivers/srf02/srf02.c
+++ b/drivers/srf02/srf02.c
@@ -17,6 +17,7 @@
  * @author      Zakaria Kasmi <zkasmi@inf.fu-berlin.de>
  * @author      Peter Kietzmann <peter.kietzmann@haw-hamburg.de>
  * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Kevin Weiss <kevin.weiss@haw-hamburg.de>
  *
  * @}
  */
@@ -30,11 +31,6 @@
 
 #define ENABLE_DEBUG        (0)
 #include "debug.h"
-
-/**
- * @brief   Per default use normal speed on the I2C bus
- */
-#define BUS_SPEED           (I2C_SPEED_NORMAL)
 
 /**
  * @brief   SRF02 register addresses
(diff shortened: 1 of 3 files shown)


MrKevinWeiss:
My srf02 may have some issues, it seems to work... just not reliably.  This behavior is in the master as well.  On the scope everything looks fine. (It basically will intermittently respond to a correct address).  If someone can test it on different hardware that isn't possibly broken that would be great!

MrKevinWeiss:
It may be that or it may be this from the [datasheet](http://www.pishrobot.com/files/products/datasheets/srf02.pdf) (although it shouldn't keep doing this)

> Checking for Completion of Ranging
You do not have to use a timer on your own controller to wait for ranging to finish. You can take advantage of the fact that the SRF02 will not respond to any I2C activity whilst ranging. Therefore, if you try to read from the SRF02 (we use the software revision number a location 0) then you will get 255 (0xFF) whilst ranging. This is because the I2C data line (SDA) is pulled high if nothing is driving it. As soon as the ranging is complete the SRF02 will again respond to the I2C bus, so just keep reading the register until its not 255 (0xFF) anymore. You can then read the sonar data. Your controller can take advantage of this to perform other tasks while the SRF02 is ranging. The SRF02 will always be ready 70mS after initiating the ranging.

drivers/srf02/srf02.c
```diff
@@ -86,22 +78,24 @@ int srf02_init(srf02_t *dev, i2c_t i2c, uint8_t addr)
...
 
 uint16_t srf02_read(const srf02_t *dev)
 {
-    uint8_t res[2];
+    uint8_t res[2] = {0xFF, 0xFF};
 
     /* read the results */
     i2c_acquire(dev->i2c);
-    i2c_read_regs(dev->i2c, dev->addr, REG_HIGH, res, 2);
+    i2c_read_regs(dev->i2c, dev->addr, REG_HIGH, res, 2, 0);
```
smlng: not your fault and I know this is (more) about adapting to the new I2c iface ...

but what if read_regs fails? This function will return garbage then, right? To me it would make more sense to change the function to `int srf02_read(const srf02_t *dev, uint16_t *val)` write result to `val` and return 0 in success and errors accordingly.

However, that would at least be a separate commit, if not a PR -

drivers/srf02/srf02.c
```diff
@@ -86,22 +78,24 @@ int srf02_init(srf02_t *dev, i2c_t i2c, uint8_t addr)
...
-    i2c_write_reg(dev->i2c, dev->addr, REG_CMD, mode);
+    status = i2c_write_reg(dev->i2c, dev->addr, REG_CMD, mode, 0);
     i2c_release(dev->i2c);
+    return status;
 }
 
 uint16_t srf02_read(const srf02_t *dev)
 {
-    uint8_t res[2];
+    uint8_t res[2] = {0xFF, 0xFF};
```
smlng: why `0xff` instead of `0x0`, just curious 😄

smlng: ah, you use that below to get invalid/failed reads, but then see my other comment on how this (IMHO) should be fixed instead.

smlng:
I can try, if @MrKevinWeiss provides the sensor 😄

smlng:
Tested on remote-revb with `tests/driver_srf02/` app, works like charm - but didn't check in depth with scope or logic analyser. RE-ACK, would merge.

dylad:
Here we go !

PR #9427, opened 2018-06-27
tests/rng: Use xtimer for speed test timeout

gebart:
### Contribution description

Significantly reduces the benchmark overhead and makes the test results more accurate by using an xtimer timeout instead of polling the time.
This measurement shows that the overhead of the range sampling function is greater than originally estimated. It will be difficult to do anything to improve this performance in the general case, but the applications using this functionality may be optimized by choosing other ranges, if possible. The intent of the test is to provide estimates of the overhead.

#### Measurements from frdm-kw41z
master:
speed: 485 KiB/s (1.000)
speed 10 0x1000 0x9000: 416 KiB/s (0.858)
speed 10 0x1000 0x9001: 353 KiB/s (0.728)
This PR:
speed: 1784 KiB/s (1.000)
speed 10 0x1000 0x9000: 1083 KiB/s (0.607)
speed 10 0x1000 0x9001: 737 KiB/s (0.413)

### Issues/PRs references

Idea from discussion in #9410

### DIFF
--- a/tests/rng/test.c
+++ b/tests/rng/test.c
@@ -385,50 +385,73 @@ void test_entropy(uint32_t samples)
     printf("Calculated %02f bits of entropy from %" PRIu32 " samples.\n", (double) entropy, samples);
 }
 
+void cb_speed_timeout(void *arg)
+{
+    unsigned *running = arg;
+    *running = 0;
+}
+
 void test_speed(uint32_t duration)
 {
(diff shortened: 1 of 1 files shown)


PeterKietzmann:
(I'm quite surprised how big the overhead actually was)

gebart:
The Samr21 has a really slow clock domain synchronization for at least one of the timers, I remember from testing the tests/bench_timers application on it.

PR #9431, opened 2018-06-27
tests: Fix size_t vs int issues

neiljay:
size_t can be any unsigned type but is never a signed integer.


neiljay:
@miri64 thanks for the quick merge!

PR #9437, opened 2018-06-27
boards: add samr30-xpro board and samr30 cpu

biboc:
This PR adds samr30-xpro board and samr30 cpu

### DIFF
--- a/boards/samr30-xpro/Makefile
+++ b/boards/samr30-xpro/Makefile
@@ -1,0 +1,3 @@
+MODULE = board
+
+include $(RIOTBASE)/Makefile.base
--- a/boards/samr30-xpro/Makefile.dep
+++ b/boards/samr30-xpro/Makefile.dep
@@ -1,0 +1,7 @@
+ifneq (,$(filter gnrc_netdev_default netdev_default,$(USEMODULE)))
+  USEMODULE += at86rf212b
+endif
+
+ifneq (,$(filter saul_default,$(USEMODULE)))
+  USEMODULE += saul_gpio
+endif
--- a/boards/samr30-xpro/Makefile.features
+++ b/boards/samr30-xpro/Makefile.features
@@ -1,0 +1,15 @@
+# Put defined MCU peripherals here (in alphabetical order)
+FEATURES_PROVIDED += periph_adc
+FEATURES_PROVIDED += periph_gpio
+FEATURES_PROVIDED += periph_i2c
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
+FEATURES_PROVIDED += periph_spi
+FEATURES_PROVIDED += periph_timer
+FEATURES_PROVIDED += periph_uart
+
+# The board MPU family (used for grouping by the CI system)
+FEATURES_MCU_GROUP = cortex_m0_2
+
+# samr30 is a specific flavor of saml21
+-include $(RIOTCPU)/saml21/Makefile.features
--- a/boards/samr30-xpro/Makefile.include
+++ b/boards/samr30-xpro/Makefile.include
@@ -1,0 +1,8 @@
+# define the cpu used by the samr30-xpro board (based on saml21)
+export CPU = saml21
+export CPU_MODEL = samr30g18a
+
+# set edbg device type
+EDBG_DEVICE_TYPE = atmel_cm0p
+
+include $(RIOTMAKE)/boards/sam0.inc.mk
--- a/boards/samr30-xpro/board.c
+++ b/boards/samr30-xpro/board.c
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2017 Baptiste Clenet <bapclenet@gmail.com>
+ *               2018 Inria
+ *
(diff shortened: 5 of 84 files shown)


dylad:
@biboc I'll do a proper review right after the I2C embargo is over. It should be fast since this MCU is based on SAML21.
First thing I can say, you should adapt doxygen + comments. You left a lot of reference to SAML21, it should be replaced by SAMR30

biboc:
Ok for the reference, I'll edit my PR

dylad:
@biboc the I2C embargo is now over so I guess you can update your periph_conf.h to the new interface.
I'll be on holiday and offline most of the time in August, but I hope to see this PR merged in September :)

aabadie:
@biboc @dylad, FYI I received a couple of these boards today and started to test this PR a bit. I found some issues that I fixed and pushed in the PR.
I also rebased on the latest master and updated the I2C configuration. I also updated all doxygen comments, fixed LED macros and saul support for the board GPIOs.
Note that I split the initial commit in 2: one for the CPU addition and one for the board addition.

What has been tested with success:
- timer/xtimer/saul
- I2C with one sensor plugged on PA16/PA17
- radio configuration using gnrc_networking, but not communication between 2 nodes
- flashing works either with edbg or openocd

I don't really like all this the copy-pasted code but maybe some factorization can be done in follow-up PRs.

As is, I would say that this PR is in a mergeable state (after squashing and green CIs) but maybe @dylad you want to have a look first.

dylad:
> I don't really like all this the copy-pasted code but maybe some factorization can be done in follow-up PRs.

Can you elaborate a bit more on this point ? I think the only copy-paste code come from timer/rtc/rtt driver.

I'll take a look asap but I don't have this board to test.

aabadie:
> Can you elaborate a bit more on this point ? I think the only copy-paste code come from timer/rtc/rtt driver.

Yes, some periph implementation could have been shared. But the rest is ok.

dylad:
> Yes, some periph implementation could have been shared

Since SAMR30 is a SAML21 yes, we can merged all cpu/samr30/periph to cpu/saml21/periph right now.

cladmi:
@aabadie The problem is that `saml21` is already a `CPU` so a CPU should not really depend on another. Can't you declare the board with a `saml21` cpu and do the specific handling based on `CPU_MODEL` in `saml21` ?

aabadie:
> Can't you declare the board with a saml21 cpu and do the specific handling based on CPU_MODEL in saml21 ?

This sounds like a good idea. I'll try it and let you know how it goes.

dylad:
How does samr21 acts ? I mean a samr21 is a samd21. you copy the same behaviour for samr30 and saml21.

aabadie:
@dylad, I dropped the cpu/samr30 directory and slightly adapted the saml21 pm.c. It builds fine now and still works (tested applications examples/default, tests/periph_rtc and tests/periph_rtt).

I can try to re-add authorship of initial commits to @biboc if we want (because my git rebases/edit/amends dropped him).

aabadie:
Just fixed errors reported by Murdock. Let me know when I can squash :)

aabadie:
@dylad, CI is all green. Ok to merge ?

dylad:
Here we go.

biboc:
Have you tried your change on a board?

keestux:
@biboc @dylad
Sorry to complain about the PR that is already merged.
Why was an older ASF version used, 3.34.2 instead of 3.35.1?
Why was it needed to modify the vendor files? The new method (since c5cbc428b, 2017-09-20) is to verbatim copy from ASF.

keestux:
@biboc @dylad
Well, the thing is, the method to copy vendor files was changed, back in 2017. Before c5cbc42 it was as it is described in the README of samr30. It's something I described for samd21 and the samr30 README is basically a copy of that.
Since c5cbc42 we simply copy the files verbatim (i.e. unchanged) from ASF into RIOT. In all cases we describe what ASF version was used, or whatever source was used.

Furthermore I am unhappy of the fact that the samr30 files are from ASF 3.34.2. I believe that was totally unnecessary.

You may think: who cares, but I see great value in documenting. When we want to pull in updates from newer ASF we need to know what changed. Each ASF can have potential bug fixes, which is also the case for 3.34.2 => 3.35.1
```
-typedef volatile       uint32_t WoReg8;  /**< Write only  8-bit register (volatile unsigned int) */
+typedef volatile       uint8_t  WoReg8;  /**< Write only  8-bit register (volatile unsigned int) */
```

dylad:
@keestux Do you want to update the SAMR30 vendor files to 3.35.1 ?

I don't know why 3.34.2 what pick up but we can still improve this :)

keestux:
@dylad Yes, I already have it prepared, but I want to do it on top of #10914

dylad:
Ok I'll have a look at it this weekend for sure.

biboc:
@keestux, you're right, I did pay attention to ASF version and how files verbatim are copied. Feel free to update at best

keestux:
@biboc I guess you wanted to say: "... I did **not** pay attention ...".
In any case, can you have a look at #10941, and confirm that it is working for you?

PR #9438, opened 2018-06-27
kinetis: Clean up SPI module power handling

gebart:
### Contribution description

Fixes a problem with hardware CS when using slow SPI speeds, mentioned in https://github.com/RIOT-OS/RIOT/issues/6437#issuecomment-283114795

The problem was identified as the module was being disabled before the CS line has been deasserted when using fast core clocks in combination with slow SPI speeds. By setting the HALT bit and waiting for the TXRXS status flag to become unset we allow the hardware to acknowledge that the transfer is completely finished and we can finally disable the module clocks.

### Issues/PRs references

#6437

PR #9440, opened 2018-06-27
drivers/lis3dh: Add missing .clk initializer

gebart:
### Contribution description

There was no initializer for the .clk member of lis3dh_params_t, so it was always defaulted to SPI_CLK_100KHZ for platforms which use the default SPI speed definitions. The device is rated for up to 10 MHz SPI speeds, according to the data sheet, so a default setting of 5 MHz should be a safe choice without wasting time on transfers.

### Issues/PRs references

discovered while testing #9438


aabadie:
Maybe have a look at  #8697, which also provide this change, among other things.

PeterKietzmann:
@gebart thanks for the fix. If you don't see reasons against it, I'd rather go with #8697 because it's been out longer.

gebart:
Agreed, closing in favor of #8697

PR #9268, opened 2018-06-01
drivers/fxos8700: adapt to new i2c API

gebart:
### Contribution description

Adapt drivers/fxos8700 to new i2c API

### Issues/PRs references

#6577

PR #9442, opened 2018-06-27
boards/frdm: Add bus configuration for onboard FXOS8700CQ sensor 

gebart:
### Contribution description

Adds missing bus and slave address configuration for the supported FRDM Kinetis boards. These boards each have different configurations for the FXOS8700 sensor, so it is necessary to specify the address for each board.
This should not be affected by the i2c embargo because it applies to the device driver configuration and it works exactly the same both in master and after the changes in #9268.

### Issues/PRs references

#9268

### DIFF
--- a/boards/frdm-k22f/include/board.h
+++ b/boards/frdm-k22f/include/board.h
@@ -55,6 +55,14 @@ extern "C"
 /** @} */
 
 /**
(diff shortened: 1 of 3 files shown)


gebart:
tested on frdm-kw41z

PR #9451, opened 2018-06-28
Makefile.include: fully define BASELIBS before using its value

cladmi:
### Contribution description

As described in https://github.com/RIOT-OS/RIOT/issues/8913 some places are using the immediate value of `BASELIBS` before it is actually set (== use of non initialized variables…). The problem was mitigated because ELFFILE is always rebuild even if files in `$(BASELIBS)` did not change.

This PR addresses this by moving the definition of `BASELIBS` sooner in `Makefile.include`.

This required:

* processing `modules.inc.mk` earlier
* process all `pkg/*/Makefile.include` as `libcose` defines `PSEUDOMODULES` which modifies `BASELIBS` https://github.com/RIOT-OS/RIOT/issues/8984 and https://github.com/RIOT-OS/RIOT/pull/9003

I also added a debug commit that asserts that `BASELIBS` was not changed during build. It can be removed before merging.

### Issues/PRs references

This was problematic when trying to reference `$(BASELIBS)` as a target dependency in the application Makefile: https://github.com/RIOT-OS/RIOT/pull/9351

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -362,6 +362,9 @@ LINKFLAGPREFIX ?= -Wl,
 
 DIRS += $(EXTERNAL_MODULE_DIRS)
 
+# Save value to verify it is not modified later
+_BASELIBS_VALUE_BEFORE_USAGE := $(BASELIBS)
+
 # Linker rule
 $(ELFFILE): FORCE
 ifeq ($(BUILDOSXNATIVE),1)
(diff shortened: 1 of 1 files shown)


cladmi:
I moved the 'test' commit earlier in the history to help demonstrate the fixes:

You can execute these tests on each commit and see the fixes applied:

```
make -C examples/hello-world
make -C tests/unittests
make -C examples/bindist
```

Each commit currently fixes one problem so it is easy to verify.

cladmi:
Is there a reason to keep the BASELIBS change detection or should I remove as I planned to ?

smlng:
> Is there a reason to keep the BASELIBS change detection or should I remove as I planned to ?

mhm, maybe its good to keep this (and maybe add other) checks in the build system - but could be moved into a distinct file like `makefiles/checks.inc.mk` and maybe add a switch to dis/enable these checks? I mean it would be nice to verify that other changes to the make/build stuff isn't introducing any problems related to BASELIBS and such.

PR #9456, opened 2018-06-28
cpu/atmega_common: misc gpio interrupt fixes

ZetaR60:
This fixes two misc problems with ATmega GPIO interrupts:
- if the flank is not supported, it currently partially initializes the interrupt before returning
- if an interrupt condition occurs prior to the interrupt being enabled, the interrupt triggers immediately (because the flag wasn't cleared first)


PeterKietzmann:
Code-wise ACK, change makes sense. Will run some basic tests in a moment

PR #9432, opened 2018-06-27
drivers/pn532: adapt to new i2c API

MichelRottleuthner:
### Contribution description
Aapplies the new i2c API to the pn532. 
Tested with adafruit PN532 RFID/NFC shield v1.3 and nucleo144-f446
### Issues/PRs references
#6577

PR #9458, opened 2018-06-29
drivers/pn532: adapt to new i2c API (recreated)

MichelRottleuthner:
### Contribution description
clone of #9432 - with all the changes addressed. Accidentally dleted the old branch and wasn't able to reopen the PR.

### Issues/PRs references
#6577

### DIFF
--- a/drivers/pn532/pn532.c
+++ b/drivers/pn532/pn532.c
@@ -120,15 +120,7 @@ int pn532_init(pn532_t *dev, const pn532_params_t *params, pn532_mode_t mode)
     gpio_init(dev->conf->reset, GPIO_OUT);
     gpio_set(dev->conf->reset);
     dev->mode = mode;
-    if (mode == PN532_I2C) {
-#ifdef PN532_SUPPORT_I2C
-        if (i2c_init_master(dev->conf->i2c, I2C_SPEED_FAST) != 0) {
-            DEBUG("pn532: initialization of I2C bus failed\n");
-            return -1;
-        }
-#endif
-    }
-    else {
+    if (mode == PN532_SPI) {
 #ifdef PN532_SUPPORT_SPI
         /* we handle the CS line manually... */
         gpio_init(dev->conf->nss, GPIO_OUT);
@@ -175,7 +167,10 @@ static int _write(const pn532_t *dev, uint8_t *buff, unsigned len)
     if (dev->mode == PN532_I2C) {
 #ifdef PN532_SUPPORT_I2C
         i2c_acquire(dev->conf->i2c);
-        ret = i2c_write_bytes(dev->conf->i2c, PN532_I2C_ADDRESS, buff, len);
+        ret = i2c_write_bytes(dev->conf->i2c, PN532_I2C_ADDRESS, buff, len, 0);
+        if (ret == 0) {
+            ret = (int)len;
+        }
         i2c_release(dev->conf->i2c);
 #endif
     }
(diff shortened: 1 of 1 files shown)


dylad:
@MichelRottleuthner Travis is complaining, there are some styles issues.
You can fix and squash directly.

dylad:
@MichelRottleuthner I have no idea... maybe @kaspar030 has an idea ?
btw, Travis is complaining again with whitespace issue.

kaspar030:
> why are we using such an old version of cppcheck?

Murdock uses the ubuntu 16.04 version, 1.72. Travis seems to be even older. Maybe that can be updated? @aabadie?

dylad:
Yeah, it would be great to update it so we don't have to add voodoo things here.

aabadie:
> Maybe that can be updated?

I prefer dropping Travis and switch to CircleCI like I did in #8729. The advantage is that it uses the official docker image riotdocker so the tools versions will be same has the ones on Murdock.

dylad:
> I prefer dropping Travis and switch to CircleCI like I did in #8729

How do it works for this PR ?

aabadie:
> How do it works for this PR ?

Revert the changes related to cppcheck and only keep the changes related to I2C. If Travis fails, it won't block the merge button anyway.

MichelRottleuthner:
reverted cpp changes. Is the upgrade to the CI planned before the i2c rework branch will be merged?

dylad:
@aabadie you mean we can merge it right now without taking care of Travis ?

aabadie:
>  you mean we can merge it right now without taking care of Travis ?

Yes

dylad:
Let's go then !

drivers/pn532/pn532.c
```diff
@@ -80,7 +80,7 @@
 
 #if ENABLE_DEBUG
 #define PRINTBUFF printbuff
-static void printbuff(char *buff, unsigned len)
+static void printbuff(uint8_t *buff, unsigned len)
```
aabadie: Is this a leftover of changes related to cppcheck ?

drivers/pn532/pn532.c
```diff
@@ -120,15 +120,7 @@ int pn532_init(pn532_t *dev, const pn532_params_t *params, pn532_mode_t mode)
...
-    if (mode == PN532_I2C) {
-#ifdef PN532_SUPPORT_I2C
-        if (i2c_init_master(dev->conf->i2c, I2C_SPEED_FAST) != 0) {
-            DEBUG("pn532: initialization of I2C bus failed\n");
-            return -1;
-        }
-#endif
-    }
-    else {
+    if (mode == PN532_SPI) {
```
aabadie: It should be `PN532_I2C`, no ?

aabadie: Ok, I see

drivers/pn532/pn532.c
aabadie: I would also revert this and open a separate PR. A least, it should not be part of the same commit containing I2C changes

dylad:
@aabadie I'll let you push the button then :)

drivers/pn532/pn532.c
MichelRottleuthner: reverted. shall I open the other PR against new_i2c_if too? (master should work though)

aabadie: I think it should be fine against master.

PR #4360, opened 2015-11-28
Add a RELIC package and unit test that shows how to use it

OlegHahm:
Add install of cmake to .travis.yml.

Test-Information:

This has been tested on OS X and on Ubuntu for both the native target
and STM32F4Discovery.

Adopted #2213 from @tfar.

PR #9463, opened 2018-06-29
Monocypher: Initial import of package

bergzand:
### Contribution description

Add support for the [Monocypher] crypto library as external package. Compared to TweetNaCl, the flash usage is twice the size (around 10K), but signature generation and verification is 5 - 8 times as fast as TweetNaCl on the different ARM cortex versions.

### Issues/PRs references

None

[Monocypher]: https://monocypher.org

### DIFF
--- a/pkg/monocypher/Makefile
+++ b/pkg/monocypher/Makefile
@@ -1,0 +1,11 @@
+PKG_NAME=monocypher
+PKG_URL=https://github.com/LoupVaillant/Monocypher
+PKG_VERSION=6c2edbb48a61f43452c48c5f1e3ddabb19af3111
+PKG_LICENSE=CC-0
+
+.PHONY: all
+
+all: git-download
+	$(Q)"$(MAKE)" -C $(PKG_BUILDDIR)/src
+
+include $(RIOTBASE)/pkg/pkg.mk
--- a/pkg/monocypher/Makefile.dep
+++ b/pkg/monocypher/Makefile.dep
@@ -1,0 +1,1 @@
+USEMODULE += monocypher_sha512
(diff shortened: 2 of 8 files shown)


tcschmidt:
How does compare to the Relic library already available in RIOT?

bergzand:
> How does compare to the Relic library already available in RIOT?

@tcschmidt Do you have an example on how to use ed25519 based signatures with RELIC? I'd be happy to compare it, but I'd have to know how to actually use RELIC first. For me the interesting metrics to compare on are flash size, stack space usage and speed on the Cortex-M{0+,3,4F}

Monocypher doesn't compile for 8 bit and 16 bit architectures unfortunately.

tcschmidt:
Maybe a look at #4360 helps - or contact @tfar .

bergzand:
@tcschmidt #4360 also was the first thing I thought of for an example, but it only implements ecdh. Comparing goals between RELIC and Monocypher, it seems that RELIC is a library for the low level cryptographic operations where Monocypher presents the user with the high level cryptographic primitives.

Assuming you still want a direct performance comparison between RELIC and Monocypher for public key signature operations, would you mind asking @tfar if he has an example somewhere that is using RELIC for ed25519 based signatures?

jcarrano:
@bergzand Could you add a simple doc.txt so that it shows up in the API docs. Otherwise no one will even know we support this lib?

> Monocypher doesn't compile for 8 bit and 16 bit architectures unfortunately

Put that in the doc.txt

bergzand:
@jcarrano Added a doc.txt based on the tweetnacl one. Also fixed the whitespace issues reported by travis.

jcarrano:
@bergzand Squash and we are ready to go. BTW, its so annoying that the CI fails when you don't squash!

bergzand:
@jcarrano Squashed!

bergzand:
Thanks for the quick review @jcarrano!

PR #7897, opened 2017-10-28
kinetis: Implement low power modes

gebart:
~~Based on #7882~~
Depends on ~#10320~, ~~#9930~~ 

Low power modes using pm_layered with automatic blocking of modes depending on what peripherals are in use.
The UART peripheral driver has been enhanced with checking of error indication flags in the RX interrupt.

TODO
 - [x] board configuration updates for LLWU
 - [ ] Measure power consumption
 - [ ] Test peripherals
 - [x] LPUART LLS operation similar to UART
 - [x] Make pm_layered usage conditional on USEMODULE

Tested on FRDM-K22F, both with PIT and with LPTMR for xtimer.

The UART on the FRDM-K22F seem to collect some garbage during boot when using low power modes, resulting in the first one or two commands entered over pyterm to return error messages about command not found. Typing the same command again (by hitting enter again in pyterm) always seem to work. It does not seem to be dependent on how long we wait after boot, but rather there's some junk being collected in the internal stdio_uart RX buffer and the buffer being flushed when the first newline is received.
Tested with pyterm, picocom, and minicom to ensure that the UART is working properly both with long RX sequences (pyterm) and with single keystroke transmissions (minicom).

Example terminal output with boot garbage after flashing:
```
2017-10-28 15:46:08,467 - INFO # �����main(): This is RIOT! (Version: 2018.01-devel-149-g6d3d3-cheesecake-pr/kinetis-pm)
2017-10-28 15:46:08,469 - INFO # Welcome to RIOT!
> ps

2017-10-28 15:46:10,825 - INFO #  �.�ps
2017-10-28 15:46:10,828 - INFO # shell: command not found: �.�ps
> 
2017-10-28 15:46:11,411 - INFO #  ps
2017-10-28 15:46:11,420 - INFO # 	pid | name                 | state    Q | pri | stack  ( used) | base addr  | current     
2017-10-28 15:46:11,427 - INFO # 	  - | isr_stack            | -        - |   - |   1024 (  116) | 0x1fff0000 | 0x1fff03c8
2017-10-28 15:46:11,436 - INFO # 	  1 | idle                 | pending  Q |  15 |    768 (  128) | 0x1fff0714 | 0x1fff0994 
2017-10-28 15:46:11,444 - INFO # 	  2 | main                 | running  Q |   7 |   1536 (  632) | 0x1fff0a14 | 0x1fff0e8c 
2017-10-28 15:46:11,450 - INFO # 	    | SUM                  |            |     |   3328 (  876)
```

PR #9468, opened 2018-06-30
boards/mulle: Fix mixed up RX TX pins on UART0

gebart:
### Contribution description

The TX and RX definitions were accidentally switched compared to the actual pin functions in the CPU.

### Issues/PRs references

Discovered when testing #7897


gebart:
From the CPU manual, chapter on pin mux functions:
image
image

PR #9472, opened 2018-06-30
kinetis: Conditionally enable MCG

gebart:
### Contribution description

Small cleanup that allows a build without the Kinetis multipurpose clock generator (MCG) driver. Might become useful in a bootloader build where ROM space is very constrained.

Saves 788 bytes of ROM on examples/hello-world when building with 
```
BOARD=mulle DISABLE_MODULE=periph_mcg make
```

The clock speed definition must be updated to match the power on reset default for the CPU in order to get proper output on UART etc, but it is out of scope for this PR.

### Issues/PRs references

Split from #7897

### DIFF
--- a/cpu/kinetis/Makefile.features
+++ b/cpu/kinetis/Makefile.features
@@ -1,5 +1,6 @@
 FEATURES_PROVIDED += periph_cpuid
 FEATURES_PROVIDED += periph_hwrng
 FEATURES_PROVIDED += periph_gpio
+FEATURES_PROVIDED += periph_mcg
 
 include $(RIOTCPU)/cortexm_common/Makefile.features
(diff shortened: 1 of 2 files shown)


kaspar030:
So now it is turned off by default?

gebart:
It is supposed to be on by default unless MODULE_DISABLE=periph_mcg
If it is not then I did something wrong.

kYc0o:
> cpu/kinetis/Makefile.include has a line which adds USEMODULE+=periph_mcg

That should be enough. I'll test on my kinetis boards.

kaspar030:
> cpu/kinetis/Makefile.include has a line which adds USEMODULE+=periph_mcg

:+1:

PR #9476, opened 2018-07-02
cpu/stm32_common: various improvements in i2c_1 driver

### LINKED ISSUE #9461: stm32common i2c_1.c cannot write multiple registers (new i2c api)
#### Description
Cannot write multiple registers, it seems it gets locked up.  I think @aabadie has a solution already that works.  (This may be fixed, I won't have HW to fix until Monday, I just wanted it documented somewhere)

Write multiple registers to a sensor with any stm32f0/3/7 and l0/4. I used stm32f303 to test

#### Expected results
For it not to hang

#### Actual results
[automated test](
https://gist.github.com/MrKevinWeiss/e68873abc21e0d4703ebba7d94f54874)
It hangs

#### Versions
new_i2c_if branch

aabadie:
### Contribution description

This PR improves the error codes and the different bus checks for the first family of STM32 I2C drivers (i2c_1). So it's related to L0/L4 and F0/F3/F7 STM32 families.

It may potentially fix #9461 but not sure.

Here is the list of changes provided by this PR:
- dont send stop if bus is busy and before any action is done on bus
- add bus check that looks for bus error, arbitration and nack events
- add a small delay before performing bus checks

I'm not sure how this can be tested with

### Issues/PRs references

related to #6577 and attempt to fix #9461


cpu/stm32_common/periph/i2c_1.c
```diff
@@ -428,14 +443,28 @@ static inline int _write(I2C_TypeDef *i2c, const uint8_t *data, size_t length)
...
+        int ret = _check_bus(i2c);
+        if (ret < 0) {
+            return ret;
+        }
+    }
+
+    DEBUG("[i2c] write: Waiting for write to complete\n");
+    uint16_t tick = TICK_TIMEOUT;
+    while (!(i2c->ISR & I2C_ISR_TC) && tick--)  {
+        if (i2c->ISR & ERROR_FLAG || !tick) {
```
smlng: the `!tick` will never happen due to the `while` condition, hence I would suggest to move the `if` after the `while`

cpu/stm32_common/periph/i2c_1.c
```diff
@@ -451,17 +480,53 @@ static inline int _stop(I2C_TypeDef *i2c)
...
     }
 
     /* send STOP condition */
     DEBUG("[i2c] stop: Generate stop condition\n");
     i2c->CR2 |= I2C_CR2_STOP;
 
+    /* Wait for the stop to complete */
+    tick = TICK_TIMEOUT;
+    while (!(i2c->CR2 & I2C_CR2_STOP) && tick--) {
+        if (!tick) {
```
smlng: move `if` after the while

cpu/stm32_common/periph/i2c_1.c
smlng: okay, actually it will happen, but only on the last iteration, so I would still move it after the while, looks way cleaner (IMHO)

cpu/stm32_common/periph/i2c_1.c
aabadie: I agree and I addressed this kind of comments in the last fixup commit of this PR

MrKevinWeiss:
@dylad Should we merge?

PR #9480, opened 2018-07-03
Docs: add "allow time to ACK before dismissing review" to maintainer guidelines

danpetry:
### Contribution description

This PR includes a comment about allowing time for other maintainers to ACK before dismissing a review.

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -128,6 +128,11 @@ This could include:
     review
 -   Usage of GitHub functionality, for example "Reviewers" and "Assignees" lists
 
+#### Github etiquette
+
+-   If there are multiple maintainers reviewing a PR, always give the other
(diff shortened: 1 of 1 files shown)


danpetry:
Cool thanks. Re the preciseness of the language, I had a conversation with @tcschmidt on friday about this. So, it would be good to maybe start talking about and establish a precedent for how much these points nail things down. In general I would say:
- Should allow maintainers the amount of room for interpretation which reflects the amount of experience that we expect maintainers to have at the point when they become maintainers. In other words, they should not cover things that we expect maintainers to know at that point
- Otherwise, they should be as precise as possible
- Nevertheless, they should always be phrased as "guidelines" rather than "rules" because of the inherent voluntary nature of an open source project
What do you think?

PR #9489, opened 2018-07-03
sys/xtimer: remove redundant callback schedule for overflow

Hyungsin:
### Contribution description
xtimer schedules a callback on next overflow (_lltimer_set(0xFFFFFFFF)) when there is no timer in the current timer period.

However, if there is no xtimer EVEN in the future timer period (i.e., literally zero on-going xtimer), this next overflow callback does not have to be scheduled. 

This PR removes such redundant overflow callback schedules, which reduces energy/time resource consumption by scheduling overflow callbacks only when necessary.

### Issues/PRs references
#


Hyungsin:
closed in favor of #9530

PR #9490, opened 2018-07-03
cpu/atmega_common: UART sleep support

ZetaR60:
### Contribution description
This adds support for ```uart_poweron``` and ```uart_poweroff``` to cpu/atmega_common. This is in preparation for more complete power management on the ATmega.

### Issues/PRs references
Contains code salvaged from #9420

### DIFF
--- a/cpu/atmega_common/periph/uart.c
+++ b/cpu/atmega_common/periph/uart.c
@@ -241,12 +241,6 @@ void uart_poweron(uart_t uart)
 
 void uart_poweroff(uart_t uart)
 {
-    unsigned state = irq_disable();
-
-    /* disable and reset UART */
-    dev[uart]->CSRB = 0;
-    dev[uart]->CSRA = 0;
-
     switch (uart) {
         case 0: power_usart0_disable();
 #if (UART_NUMOF == 2)
@@ -261,8 +255,6 @@ void uart_poweroff(uart_t uart)
 #endif
         default: return;
     }
-
-    irq_restore(state);
 }
 
 static inline void isr_handler(int num)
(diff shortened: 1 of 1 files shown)


cpu/atmega_common/periph/uart.c
```diff
@@ -74,94 +78,193 @@ static mega_uart_t *dev[] = {
...
+#if defined(CPU_ATMEGA328P)
+    (void)uart;
+    return ((PRR >> PRUSART0) & 0x1);
+#else
+
+    if (uart == 0) {
+        return ((PRR0 >> PRUSART0) & 0x1);
+    }
+
+#if defined(CPU_ATMEGA1284P)
```
MichelRottleuthner: by moving this `#if` inside `else if (uart == 1) {` two lines can be saved and readability is improved a bit

cpu/atmega_common/periph/uart.c
```diff
@@ -74,94 +78,193 @@ static mega_uart_t *dev[] = {
...
 {
-    uint16_t brr;
+    dev[uart]->BRR = uart_brr[uart];
 
+#if defined(UART_DOUBLE_SPEED)
+    dev[uart]->CSRA |= (1 << U2X0);
+#endif
+}
+
+static inline void _calc_brr(uart_t uart, uint32_t baudrate)
```
MichelRottleuthner: is an extra method for this necessary? it's only used a single time

cpu/atmega_common/periph/uart.c
ZetaR60: Done.

cpu/atmega_common/periph/uart.c
ZetaR60: It improves readability, and is directly adapted from prior behavior. It should produce the same instructions, because of inline.

PR #9495, opened 2018-07-04
driver: fix cppcheck error in pn532

smlng:
cppcheck reported 'duplicateBranch' error in 2 if-else
    statements, b/c they are empty when running static-tests.

### Contribution description

should fix cppcheck error: https://travis-ci.org/RIOT-OS/RIOT/builds/399003316?utm_source=github_status&utm_medium=notification

### DIFF
--- a/drivers/pn532/pn532.c
+++ b/drivers/pn532/pn532.c
@@ -164,18 +164,19 @@ static int _write(const pn532_t *dev, uint8_t *buff, unsigned len)
     (void)buff;
     (void)len;
 
-    if (dev->mode == PN532_I2C) {
+    switch (dev->mode) {
 #ifdef PN532_SUPPORT_I2C
+    case PN532_I2C:
         i2c_acquire(dev->conf->i2c);
         ret = i2c_write_bytes(dev->conf->i2c, PN532_I2C_ADDRESS, buff, len, 0);
         if (ret == 0) {
             ret = (int)len;
         }
         i2c_release(dev->conf->i2c);
+        break;
 #endif
-    }
-    else {
 #ifdef PN532_SUPPORT_SPI
+    case PN532_SPI:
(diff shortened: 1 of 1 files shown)


drivers/pn532/pn532.c
```diff
@@ -222,10 +227,15 @@ static int _read(const pn532_t *dev, uint8_t *buff, unsigned len)
         buff[0] = 0x80;
         reverse(buff, len);
         ret = (int)len + 1;
+        break;
 #endif
+    default:
+        DEBUG("pn532: invalid mode (%i)!\n", dev->mode);
+    }
+    if (ret > 0) {
```
aabadie: This different from the previous behavior. I would just skip it if not using `ENABLE_DEBUG` (I think it should be optimized by the compiler):
```c
if (ENABLE_DEBUG && (ret > 0)) {
...
}
```

smlng: the compile will optimise this away even without `ENABLE_DEBUG` as extra condition, thus its not needed here.

smlng: and yes the behaviour is different, but safer because before ifs there was an error but debugging enabled it would print undefined buffer content.

smlng:
btw. the changes save 4B on binary size (ROM) without changes to RAM.

smlng:
@aabadie I would merge this into the feature branch, b/c latest for the merge into master we need to fix this anyways, or it will pop up in the feature PR.

aabadie:
>latest for the merge into master we need to fix this anyways, or it will pop up in the feature PR.

Only in Travis, with Murdock it will pass.

smlng:
@aabadie ah different cppcheck version in Murdock and Travis, right? Anyway, isn't annoying to have Travis fail on every PR in the I2C rework now - with this fix its happy again 😉

PR #9503, opened 2018-07-05
sys/stdio: cleanup of STDIO abstraction

haukepetersen:
### Contribution description
Currently, the design of our STDIO mapping is partly a mess - e.g. why does the Segger RTT STDIO need to call `uart_stdio_x` functions?!

So this PR is a first step towards cleaning up the situation: I simply factored out the `init/read/write` functions in a more generic header (`stdio_base.h`), while mapping the actual implementation to this interface during compile-time base on the selected modules. The 'users' of these backend functions now don't need to care about anything implementation specific (hence no `uart` left in the name)...

P.S.: my driver for this cleanup is to run the RIOT shell over BLE...

### Issues/PRs references
none


sys/stdio_uart/stdio_uart.c
```diff
@@ -49,73 +51,73 @@ extern ethos_t ethos;
...
 
-static ssize_t uart_stdio_vfs_read(vfs_file_t *filp, void *dest, size_t nbytes)
+static ssize_t stdio_uart_vfs_read(vfs_file_t *filp, void *dest, size_t nbytes)
 {
     int fd = filp->private_data.value;
     if (fd != STDIN_FILENO) {
         return -EBADF;
     }
-    return uart_stdio_read(dest, nbytes);
+    return stdio_uart_read(dest, nbytes);
```
vincent-d: `stdio_read`?

sys/stdio_uart/stdio_uart.c
```diff
@@ -49,73 +51,73 @@ extern ethos_t ethos;
...
 
-static ssize_t uart_stdio_vfs_write(vfs_file_t *filp, const void *src, size_t nbytes)
+static ssize_t stdio_uart_vfs_write(vfs_file_t *filp, const void *src, size_t nbytes)
 {
     int fd = filp->private_data.value;
     if (fd == STDIN_FILENO) {
         return -EBADF;
     }
-    return uart_stdio_write(src, nbytes);
+    return stdio_uart_write(src, nbytes);
```
vincent-d: `stdio_write`?

haukepetersen:
> Apart from the minor renaming issues which have already been spotted by Murdock, I'm wondering if the vfs initialization code should not be moved out of stdio_uart to a generic module, as this code should be shared for all stdio_* implementations.

yes, that makes certainly sense. I would however like to do it in a follow up, so this PR doesn't get blown up by too many changes. Are you ok with that?

vincent-d:
> Are you ok with that?

I am

haukepetersen:
prefect. See #9790

vincent-d:
You can squash ;)

haukepetersen:
squashed.

haukepetersen:
and fixed a minor doxygen misshap

basilfx:
Maybe it's good to add the API change label, because this changes `USEMODULE += rtt_stdio` into `USEMODULE += stdio_rtt`?

haukepetersen:
done.

haukepetersen:
@gebart @bergzand @kaspar030 are you all good with this change? If so, please feel free to press the green button...

gebart:
I agree with the intent and general idea of this PR, but I did not have time to look at the code. It's good to see that the stdio code is being cleaned up :+1:

sys/stdio_rtt/stdio_rtt.c
```diff
@@ -319,7 +317,7 @@ static ssize_t rtt_stdio_vfs_write(vfs_file_t *filp, const void *src, size_t nby
 
 #endif
 
-void uart_stdio_init(void) {
+void stdio_init(void) {
     #ifndef RTT_STDIO_DISABLE_STDIN
```
basilfx: Similar to `STDIO_UART_DEV`, shouldn't this then be named `STDIO_RTT_DISABLE_STDIN`?

haukepetersen:
@basilfx good catch, fixed and squashed.

basilfx:
Then I also ACK :-)

PR #9516, opened 2018-07-07
Drivers for MPU-9250 9-axis motion sensor.

VilleHiltunen:
### Contribution description

I was encouraged to share the driver for mpu-9250 that was a by-product of a finished project. The driver is almost completely [Fabian Nack's mpu-9150 driver](https://github.com/RIOT-OS/RIOT/tree/master/drivers/mpu9150), since the sensors are very similar. The main difference is that mpu-9250 accelerometer configuration register is split into two. I also added a rudimentary interrupt enable function to the driver.

The fork also uses the old mpu-9150 test application and saul code.

To test the driver, wire the sensor to your board using the board defined I<sup>2</sup>C pins, and of course the ground and 3.3v power. Then flash the test application onto the board, and it should output measurements via UART.

Testing the interrupt is a bit trickier, because there was no easy way to include a board agnostic interrupt pin. Cheapest way of testing the interrupts is to amend the test application by declaring a pin for interrupts, which is then wired to the mpu interrupt. Also, the interrupt must be enabled with the driver function after the initialization is done:
```
gpio_init(your_pin_here, GPIO_IN_PD);

if (mpu9250_set_interrupt(&dev, 1)) {
		puts("[Error] Couldn't configure MPU interrupt.");
		return 1;
	}
```
And then in the while loop poll the pin every loop. Reading the sensor should clear the interrupt.
```
while (1) {
    if (gpio_read(your_pin_here)){
        ...
        ...
```

Sadly, as my project ended, I lost access to a MPU-9250 myself, so I can't verify the code with the latest RIOT head. Also, I have no experience with the SAUL implementation which I merely copied. It is completely untested and needs to be verified.

### Issues/PRs references

The module is not linked to any issues or other pull requests.


smlng:
I tested an mpu9250 using the existing mpu9150 driver in RIOT, it works - so (IMHO) no need to duplicate this. If this improves or adds functionality please consider adapting the existing driver accordingly.

I mark this as duplicate for now.

smlng:
> I tested mpu9250 with mpu9150 driver too and couldn't see any problems.

thanks for verifying, that matches my observation when I tested the 9250 with the existing driver.

So I'll vote for closing this and if new features are required would suggest to add them to the existing one.

gschorcht:
I have checked the register maps of both sensors. What I figured out at a first look:

- As mentioned by @VilleHiltunen, the ```ACCEL_CONFIG``` register is splitted into two registers (register 28 an 29). The additional register ```ACCEL_CONFIG2``` only contains the settings ```ACCEL_FCHOICE_B``` and  ```A_DLPF_CFG``` that seem to be new in MPU9250.
- Additional new configuration ```FCHOICE_B``` in ```GYRO_CONFIG``` register 26
- In MPU9250, the FIFO mode can be configured by by an additional bit in register 26.

Without having deep knowledge about the meaning of ```FCHOICE```, it seems that only additional features are added to the extra registers/bits. If they are not used, MPU9250 works fine with MPU9150 driver.

What if we would define a pseudomodule ```mpu9250``` and merge the specifics of both sensors into ```mpu9150```?

One advantage of the new driver I see is the interrupt handling. But there is to be done more for it. At the moment it can simply enable the interrupt. The ISR of the application has to check for the interrupt source. There should be at least a function like ```mpu9250_get_int_source```. Otherwise the application would have to access the sensor direclty using I2C.

smlng:
closing, see discussion above

gschorcht:
@smlng I would prefere if drivers that are working for multiple sensor are called `abc98xx` and the user enable it by pseudmodules `abc9812` and `abc9834`.

Therefore, I think we should either rename the driver to `mpu9x50` and should define pseudomodules `mpu9150` and `mpu9250` or we should at least define a pseudomodule `mpu9250`. In both cases the driver could take into account the very small differences if needed.

PR #9517, opened 2018-07-09
boards/stm32l0538-disco: add initial support

aabadie:
### Contribution description

This PR adds an initial support for the [stm32l0538-disco](https://www.st.com/en/evaluation-tools/32l0538discovery.html) board.

This board embeds an epaper ink display and I could use it with the U8G2 package.

The board can be flashed via st-link but no UART is directly usable via st-link, so a serial to USB adapter is required for stdio.

### Issues/PRs references

None

### DIFF
--- a/boards/stm32l0538-disco/include/periph_conf.h
+++ b/boards/stm32l0538-disco/include/periph_conf.h
@@ -1,0 +1,150 @@
+/*
+ * Copyright (C) 2018 Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser General
+ * Public License v2.1. See the file LICENSE in the top level directory for more
+ * details.
+ */
+
+/**
+ * @ingroup     boards_stm32l0538-disco
+ * @{
+ *
+ * @file
+ * @brief       Peripheral MCU configuration for the STM32L0538-DISCO board
+ *
+ * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
+ */
+
+#ifndef PERIPH_CONF_H
+#define PERIPH_CONF_H
+
+#include "periph_cpu.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @name    Clock system configuration
+ * @{
+ */
+#define CLOCK_HSI           (16000000U)         /* internal oscillator */
+#define CLOCK_CORECLOCK     (32000000U)         /* desired core clock frequency */
+#define CLOCK_LSE           (0)                 /* enable low speed external oscillator */
+
+/* configuration of PLL prescaler and multiply values */
+/* CORECLOCK := HSI / CLOCK_PLL_DIV * CLOCK_PLL_MUL */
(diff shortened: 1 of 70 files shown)


fjmolinas:
@aabadie I have this board will test next week.

aabadie:
I had a look at this one. For periph_rtc and rtt, actually the board doesn't provide an external LSE, so I had to use a custom clock configuration and no the shared L0 one.

For the other tests, for some I had to blacklist the board, for others there are working (e.g. shell, periph_gpio, rng, struct_tm_utility, periph_eeprom, periph_flashpage). I don't know what happened for those ones.

aabadie:
Thanks @fjmolinas, squashed and rebased.

boards/stm32l0538-disco/include/periph_conf.h
```diff
@@ -0,0 +1,171 @@
...
+ * @brief       Peripheral MCU configuration for the STM32L0538-DISCO board
+ *
+ * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
+ */
+
+#ifndef PERIPH_CONF_H
+#define PERIPH_CONF_H
+
+#include "periph_cpu.h"
+#include "cfg_rtt_default.h"
```
fjmolinas: This is not valid, it is calculated based on the CLK beeing 32Khz, but here it is 37Khz. I created a branch to patch this https://github.com/fjmolinas/RIOT/tree/pr_stm32_rtt_lsi.

The problem is the test still fails because LSI isn't very precise it can have a pretty big deviation around 37KhZ.

fjmolinas:
Also the documentation needs improving.

boards/stm32l0538-disco/include/periph_conf.h
aabadie: You are right and your branch is useful and should be PRed I think.

aabadie:
> I would still add the feature but document that when using LSI manual calibration must be done. The other option is just dropping rtt. What do you think?

I would drop it for simplicity and until LSI based RTT is correctly supported. What do you mean with manual calibration ?

fjmolinas:
> I would drop it for simplicity and until LSI based RTT is correctly supported. What do you mean with manual calibration ?

You would need to measure the actual frequency and change it in the code.

fjmolinas:
> I would drop it for simplicity and until LSI based RTT is correctly supported.

Ok go ahead and drop it, I didn't PR the RTT changes because I still need to improve some corner cases. Squash right away.

aabadie:
Removed the RTT feature and added some documentation. Squashed and rebased.

fjmolinas:
@aabadie when you fix the whitespace squash directly.

aabadie:
>  when you fix the whitespace squash directly.

done

fjmolinas:
@aabadie Do you have time for one last commit? there is still a test with oveflowing RAM

fjmolinas:
@aabadie Thanks! All green with this one, GO!

PR #9519, opened 2018-07-09
boards/arduino-644p: Add initial support | cpu/atmega644p : Add initial support

Aang-3:
### Contribution description

This PR adds support for the Atmega644p MCU which is a less powerful 1284p. It also adds a board (arduino-644p), which relies on atmega commons files. It is named "arduino-644p" because it was designed to use the arduino bootlader (https://github.com/MCUdude/MightyCore), but on an unofficial card which does not come with onboard leds. I used the mega-xplained files as a base.
The MCUdude github page also shows an Opensource card they made that supports the 644p.


aabadie:
Thanks for your contribution and Welcome @Aang-3 !

aabadie:
@Aang-3, if you are still there, I think I better understand your approach now (it took me some time, sorry ;) )

My feeling is that the MightyCore development board looks very nice and could be used as a base for different specific boards using specific AVR versions.
So why not call your board `mightycore-644p` ? Afterwards, other boards could be provided for the other AVR, `mightycore-328p`, etc. All `mightycore-xxx` based boards, could inherit from a common one in `boards/common` (which itself could inherit from `common/boards/atmega` or something).
What do you think?

I also noticed that your commit messages are not fully compliant with RIOT 'standards', please split them by modules/rough parts: `cpu: add atmega-644p support`, `boards/common: add common folder for mightycore based boards`, `boards: add support for migthycore-644p`, etc.

Hope to see you soon!

PR #9521, opened 2018-07-09
cpu/stm32f0: add periph_pm support

vincent-d:
### Contribution description

Add stm32f0 support in stm32_common pm driver.
Tested with a nucleo-f091rc. Didn't check the actual power consumption yet.

### Issues/PRs references

None

### DIFF
--- a/cpu/stm32_common/periph/pm.c
+++ b/cpu/stm32_common/periph/pm.c
@@ -38,7 +38,9 @@
  *
  * Available values can be found in reference manual, PWR section, register CR.
  */
-#if defined(CPU_FAM_STM32L0) || defined(CPU_FAM_STM32L1)
+#if defined(CPU_FAM_STM32F0)
+#define PM_STOP_CONFIG  (PWR_CR_LPDS)
+#elif defined(CPU_FAM_STM32L0) || defined(CPU_FAM_STM32L1)
 #define PM_STOP_CONFIG  (PWR_CR_LPSDSR | PWR_CR_ULP)
 #else
 #define PM_STOP_CONFIG  (PWR_CR_LPDS | PWR_CR_FPDS)
--- a/cpu/stm32_common/periph/pm.c
+++ b/cpu/stm32_common/periph/pm.c
@@ -2,6 +2,7 @@
  * Copyright (C) 2016 Kaspar Schleiser <kaspar@schleiser.de>
  *               2015 Freie Universität Berlin
  *               2015 Engineering-Spirit
+ *               2017-2019 OTA keys S.A.
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -19,17 +20,14 @@
  * @author      Nick v. IJzendoorn <nijzndoorn@engineering-spirit.nl>
  * @author      Kaspar Schleiser <kaspar@schleiser.de>
  * @author      Fabian Nack <nack@inf.fu-berlin.de>
+ * @author      Vincent Dupont <vincent@otakeys.com>
  *
  * @}
  */
 
 #include "irq.h"
 #include "periph/pm.h"
-#if defined(CPU_FAM_STM32F1) || defined(CPU_FAM_STM32F2) || \
-    defined(CPU_FAM_STM32F4) || defined(CPU_FAM_STM32L0) || \
-    defined(CPU_FAM_STM32L1)
 #include "stmclk.h"
-#endif
 
 #define ENABLE_DEBUG (0)
 #include "debug.h"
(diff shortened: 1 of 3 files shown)


vincent-d:
Tested on a custom device (stm32f091 based) and works!

cpu/stm32_common/periph/pm.c
```diff
@@ -39,31 +42,56 @@
...
+/**
+ * @brief   Define EWUP config flags
+ *
+ * Available values can be found in reference manual, PWR section, register CSR.
+ */
+#if defined(CPU_FAM_STM32F1) || defined(CPU_FAM_STM32F2) || \
+    defined(CPU_FAM_STM32F4)
+#define PM_EWUP_CONFIG  (PWR_CSR_EWUP)
+#elif defined(PWR_CSR_EWUP8)
+#define PM_EWUP_CONFIG  (PWR_CSR_EWUP8 | PWR_CSR_EWUP7 | PWR_CSR_EWUP6 | \
```
kaspar030: Maybe readability can be improved by using a static inline?

```
static const unsigned _pwr_ewup_config(void)
{
unsigned tmp = 0;
#if defined(PWR_CSR_EWUP8)
tmp |= PWR_CSR_EWUP8;
#endif
#if defined(PWR_CSR_EWUP7)
tmp |= PWR_CSR_EWUP7;
#endif
...
#if defined(PWR_CSR_EWUP1)
tmp |= PWR_CSR_EWUP1;
#endif

return tmp;
}
```

... then use ```PWR->CSR = _pwr_ewup_config()``` below.

This *should* optimize to a single value even when assigning a ```volatile``` register.

vincent-d: I'll give a try

vincent-d: Done

vincent-d:
Ping

cpu/stm32_common/periph/pm.c
```diff
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2016 Kaspar Schleiser <kaspar@schleiser.de>
+ * Copyright (C) 2018 OTA keys S.A.
```
aabadie: According to the authors list order, I think the copyright order should be reversed. Otherwise this gives the impression that OTA Keys initially contributed this file, which is not true.

aabadie:
Apart from the copyright comment, the changes look good and work.

cpu/stm32_common/periph/pm.c
vincent-d: Fixed

vincent-d:
Squashed

aabadie:
Thanks @vincent-d. Now let's wait for @kaspar030's approval (and merge ? :) )

vincent-d:
Ping @kaspar030

smlng:
please rebase, I'll to find a device for testing

smlng:
if good, we can dismiss @kaspar030 review

vincent-d:
Rebased and adapted stm32l0 implementation

vincent-d:
rebased

aabadie:
Ah yes, sorry ;) I'll have a look later.

cpu/stm32_common/periph/pm.c
```diff
@@ -60,35 +115,16 @@ void pm_set(unsigned mode)
...
-
-            PWR->CSR |= PWR_CSR_EWUP1;
-#if !defined(CPU_LINE_STM32L053xx)
-            /* STM32L053 only have 2 wake pins */
-            PWR->CSR |= PWR_CSR_EWUP3;
-#endif
-#else
-            PWR->CSR |= PWR_CSR_EWUP;
 #endif
+            PWR->CSR |= _ewup_config();
```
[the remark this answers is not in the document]
vincent-d: it's actually how it's implemented. One can define `PM_EWUP_CONFIG` to enable the pins depending on their needs.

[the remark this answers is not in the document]
fjmolinas: Your are right, sorry I didn't read the code properly.

aabadie:
I'll give a last try to this PR tomorrow and will merge it. After I'll adapt #11173.

cpu/stm32_common/periph/pm.c
```diff
@@ -40,18 +42,71 @@
...
+#endif
+
+static inline uint32_t _ewup_config(void)
+{
+    uint32_t tmp = 0;
+#ifdef PM_EWUP_CONFIG
+    tmp |= PM_EWUP_CONFIG;
+#elif defined(PWR_CSR_EWUP)
+    tmp |= PWR_CSR_EWUP;
+#else
```
aabadie: I think this should be done differently.
In the current state, if the user doesn't provide a define for `PW_EWUP_CONFIG`, then this function will activate all wake-up pins.
This results in an invalid default STANDBY configuration for most of the nucleo boards in this case. For example, on nucleo-f091rc, with this default configuration, the STANDBY mode doesn't work unless one builds the application with `CFLAGS=-DPM_EWUP_CONFIG=0`.

What about having the possibility to provide the `PM_EWUP_CONFIG` in the board `periph_conf.h` and include this file in `pm.c` ?
Then there's no need for this function, just use the following snippet in the STANDBY case, below:
```c
PWR->CSR |= PM_EWUP_CONFIG
```
And provide a default PM_EWUP_CONFIG to 0 somewhere in the `periph_cpu_common.h` file.

vincent-d: I think this has been inherited from the previous versions which used to enable all wake-up pins.

I don't mind changing to use `PM_EWUP_CONFIG` only.

vincent-d:
I changed the wake-up pins config, and refactored even more.

aabadie:
Thanks for updating @vincent-d, it looks much better like this :)

I think the history of commits should be reworked: one commit introducing the pm code refactoring and one adding the support for F0. That would make history cleaner.

I'll give a try asap.

vincent-d:
@aabadie done!

PR #9533, opened 2018-07-10
make:fix dedup USEMODULE_INCLUDES in Makefile.include

smlng:
Fixes a regression introduces #9479 which break building on macOS.

### Issues/PRs references

#9479


smlng:
see also https://github.com/RIOT-OS/RIOT/commit/202e73dcdf8dd69dcf3291e95580df96cfb4d15a

and https://coderwall.com/p/cezf6g/define-your-own-function-in-a-makefile

cgundogan:
re-triggered murdock, because `tests_xtimer_usleep.elf` was failing due to a timeout ...

cgundogan:
just for reference: using the equals sign is actually a correct approach [1]. It's just `Make` for macOS that again breaks the rules.

[1] https://www.gnu.org/software/make/manual/html_node/Multi_002dLine.html

smlng:
that's for _GNU_ `make` <- big difference, very bad ...

cgundogan:
>that's for GNU make <- big difference, very bad ...

Isn't OSX using GNU make?

smlng:
it surely doesn't use any GNU stuff - at least not by default, though you can install all of those nasty GNU versions via HomeBrew or MacPorts 😉

cgundogan:
what then do you use? BSD make?

josephnoir:
Works on my Mac.

cgundogan:
@josephnoir what is your output of `make --version`? Just to be sure you don't use GNU make.

smlng:
mhm, its GNU make on macOS anyways (oops, 😬)

```
make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

PR #9524, opened 2018-07-09
pkg/semtech-loramac: use internal eeprom to store lorawan config (deveui, appeui, etc)

aabadie:
### Contribution description

This PRs use the new `periph_eeprom` in the semtech-loramac package to automatically read lorawan configuration from the internal eeprom, if the cpu provides this feature.

For example, using the `tests/pkg_semtech-loramac` application, one just has to set the required EUI and Keys, then type `loramac save` in the shell. After a reboot, no need to reenter the keys, they are automatically re-read from eeprom.

The automatic read uses an internal magic number that is stored at a certain position of the EEPROM. These are defines and can be changed at build time.

It's also possible to erase previously store configuration parameters.

Finally, the frame counters can also be stored and reused after a reboot. This is specific to ABP activation that needs to keep this value after a reboot, otherwise this needs to be cleared manually in the network backend.

Note that I add to modify the module dependency of `periph_eeprom` when it's loaded as an optional dependency on STM32.

### Issues/PRs references

None

PR #9537, opened 2018-07-10
sys/eepreg: EEPROM registration support

ZetaR60:
### Contribution description
This adds a specific method for keeping track of which things have EEPROM access, and the location and size of each thing using the EEPROM.

Currently WIP.

### Issues/PRs references
#9524

### DIFF
--- a/sys/eepreg/eepreg.c
+++ b/sys/eepreg/eepreg.c
@@ -17,24 +17,92 @@
  * @}
  */
 
+#include <stdint.h>
+
+#include "assert.h"
 #include "eepreg.h"
 #include "eepreg_conf.h"
 #include "periph/eeprom.h"
 
-eepreg_t eepreg_addr(unsigned num)
+/* basic compile-time sanity check for eepreg_conf.h */
+#define EEPREG_S_TEST    (EEPREG_MAX_ENTRY_NUM == \
+                          sizeof(eepreg_sizes) / sizeof(eepreg_sizes[0]))
+#define EEPREG_N_TEST    (EEPREG_MAX_ENTRY_NUM == \
+                          sizeof(eepreg_names) / sizeof(eepreg_names[0]))
+static_assert(EEPREG_S_TEST, "[eepreg] wrong num of entries in eepreg_sizes");
+static_assert(EEPREG_N_TEST, "[eepreg] wrong num of entries in eepreg_names");
+
+static uint32_t _find_entry(unsigned num)
 {
-    /* TODO */
+    uint32_t addr = 0;
+
+    for (unsigned i = 0; i <= num; i++) {
+        /* check if correct entry is found */
+        if (i == num) {
+            return addr;
+        }
+
+        /* skip entry name */
+        addr += sizeof(eepreg_names[i]);
+
+        /* skip entry content */
+        addr += eepreg_sizes[i];
+    }
+
+    return (uint32_t)(-1);
+}
+
+static int _check_name_at_addr(unsigned num, uint32_t addr)
+{
+    size_t len = sizeof(eepreg_names[num]);
+    uint8_t data[len];
+
+    eeprom_read(addr, data, len);
+
+    for (unsigned i = 0; i <= len; i++) {
+        if (eepreg_names[num][i] != data[i]) {
+            return -1;
+        }
+    }
+
     return 0;
+}
+
+uint32_t eepreg_addr(unsigned num)
+{
+    uint32_t addr;
+
+    if (num >= EEPREG_MAX_ENTRY_NUM) {
+        return (uint32_t)(-1);
+    }
+
+    addr = _find_entry(num);
+
+    if (_check_name_at_addr(num, addr) < 0) {
+        return (uint32_t)(-1);
+    }
+
+    /* skip name */
+    addr += sizeof(eepreg_names[num]);
+
+    return addr;
 }
 
 int eepreg_check_name(unsigned num)
 {
-    /* TODO */
-    return 0;
+    uint32_t addr = _find_entry(num);
+
+    return _check_name_at_addr(num, addr);
 }
 
 int eepreg_write_name(unsigned num)
 {
-    /* TODO */
+    size_t len = sizeof(eepreg_names[num]);
+    uint32_t addr = _find_entry(num);
+
+    if (eeprom_write(addr, eepreg_names[num], len) != len) {
+        return -1;
+    }
+
     return 0;
 }
--- a/sys/eepreg/include/eepreg_conf.h
+++ b/sys/eepreg/include/eepreg_conf.h
@@ -22,6 +22,8 @@
 #ifndef EEPREG_CONF_H
 #define EEPREG_CONF_H
 
+#include <stdint.h>
+
 #include "eepreg.h"
 
 #ifdef __cplusplus
@@ -42,12 +44,13 @@ extern "C" {
 enum {
     FOO_EEPREG_NUM,
     BAR_EEPREG_NUM,
+    EEPREG_MAX_ENTRY_NUM,
 };
 
 /**
  * @brief   Example sizes array
  */
-static const eepreg_t eepreg_sizes[] =
+static const uint32_t eepreg_sizes[] =
 {
     FOO_EEPREG_SIZE,
     BAR_EEPREG_SIZE,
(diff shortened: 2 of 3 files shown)


ZetaR60:
Note to reviewer(s): removing @aabadie review request. From #9524:

> I won't have the time to look in detail at #9537 but I like the idea in general and will be happy to reuse it in this PR when you think it's ready.

aabadie:
@ZetaR60, I looked at your proposition and I see some room for improvement:
- first I don't think the current design will simplify usage from different modules/applications. Because the content of the registry is statically defined in the `eepreg_conf.h` file
- It would be more handy to be able to add entries dynamically to the registry (removing them is another story and I don't think we want this)
- Having 2 different arrays (one for names, one for sizes) is not nice. I would rather add a struct `eepreg_entry_t` with 2 attributes, `name` and `size` and have a unique static array of entries (maybe a linked list would be better, e.g. similar to `saul_reg`).

That would give this:
- in `eepreg.h`:
```c

/* This can be easily configured at build time with CFLAGS */
#ifndef EEPREG_MAX_ENTRY_NUM
#define EEPREG_MAX_ENTRY_NUM   (8U)
#endif

typedef struct {
    uint8_t *name;
    uint32_t size;
} eepreg_entry_t;
```

- in `eepreg.c`:
```c
#include "eepreg.h"

static eepreg_entry_t eepreg_entries[EEPREG_MAX_ENTRY_NUM];

int eepreg_add_entry(eepreg_entry_t entry)
{
    /* retrieve last valid entry index (with size > 0 ?) */

    /* check if entry index after last index is less than EEPREG_MAX_ENTRY_NUM */

   /* copy the input entry attributes to the corresponding one in the array */
}

/* here is one of your function adapted with the new eepreg_entries array  */
static uint32_t _find_entry(unsigned num)
{
    uint32_t addr = 0;
    for (unsigned i = 0; i <= num; i++) {
        /* check if correct entry is found */
        if (i == num) {
            return addr;
        }
        /* skip entry name */
        addr += sizeof(eepreg_entries[i].name); /* <= here  */
        /* skip entry content */
        addr += eepreg_entries[i].size; /* <= and here  */
    }
    return (uint32_t)(-1);
}
```

The `eepreg_conf.h` would become useless.

What do you think ?

ZetaR60:
@aabadie Thanks for the feedback. I do agree that it needs some improvement.

I was originally considering to do a dynamically built singly-linked list, sort of like SAUL. However, it would be easy to build something that works 95% of the time, and then corrupts your EEPROM when someone re-factors relevant files and you do a firmware update. Another difficulty with a dynamic list might be removing a section of used EEPROM when updating the firmware (i.e. when a feature is deprecated by a product manufacturer); this would require either leaving an unused dummy section in the EEPROM, or moving data around in EEPROM during migration. I agree that we probably do not want a "normal" operation to remove registry entries, but I do think some sort of special mechanism is required for handling changes during firmware updates.

I think that any solution should have the these properties:
- incidental changes to RIOT source code should never re-order how the list is built
- the list is optionally manipulable, in order to facilitate manual manipulation of the EEPROM by the user and/or code that migrates the EEPROM from one state to another.

The question is how to preserve these properties while also having other desirable things, like requiring as little manual configuration from the user as possible. There are a few possible ways forward on this. What are your thoughts on how to proceed?

aabadie:
> it would be easy to build something that works 95% of the time, and then corrupts your EEPROM when someone re-factors relevant files and you do a firmware update

If someone refactors files and breaks compatibility between version, he will be responsible of providing EEPROM migration code that will be run during the update. I don't think the eepreg should be responsible of that.

> The question is how to preserve these properties while also having other desirable things, like requiring as little manual configuration from the user as possible. There are a few possible ways forward on this. What are your thoughts on how to proceed?

As I said previously, I don't see how I can use eepreg in #9524 in its current state. Should I write an `eepreg_conf.h` in the package code that will included first ? But then how could someone else reuse/extend the registry ?
Do you think my proposition above would be a good starting point ? It provides a way to configure the size of the registry at build time and to add/set new registry entries at run time. This should cover my use case in #9524 but maybe it's not enough.

ZetaR60:
> Do you think my proposition above would be a good starting point ? It provides a way to configure the size of the registry at build time and to add/set new registry entries at run time. This should cover my use case in #9524 but maybe it's not enough.

I think it is a good starting point, but I think this warrants a bit more planning with respect to detecting changes and migrating when they occur. If you want you could merge #9524 without eepreg, and add support for the registry in a follow-up PR.

> If someone refactors files and breaks compatibility between version, he will be responsible of providing EEPROM migration code that will be run during the update. I don't think the eepreg should be responsible of that.

I disagree here. If code changes were the only way the registry could get changed then that would be adequate, but there are a number of other situations where the user may want to migrate. I am anticipating that the user will want to get from point A to point B, but lacks the knowledge required to create new migration code. The ideal situation would be if detection of changes and migration could happen automatically. I think for now, we should just ensure that the information required to automatically detect and migrate is easy to get, so that migration can be added in the future. My thought with the current setup of eepreg is that the user could provide an old and new version of ```eepreg_conf.h``` to a migration utility, but there are probably better ways of doing this.

I am considering writing the registry information at the beginning of the EEPROM, and then validating that against a dynamically built registry (similar to your recommendation). This would replace the way magic numbers are currently used. The registry information would then be available to be used during the firmware update to automatically migrate the EEPROM (via code to be written in a future PR when it is actually needed). For now the user could be given the option to: either have ```eepreg_addr``` fail when the registry has changed, or to ignore the registry when it has changed and overwrite it with new information. What do you think?

> As I said previously, I don't see how I can use eepreg in #9524 in its current state. Should I write an ```eepreg_conf.h``` in the package code that will included first ? But then how could someone else reuse/extend the registry ?

Sorry for the lack of clarity. The intention is that ```eepreg_conf.h``` is provided by the application, and is controlled by the user. Code using eepreg does the following:
- Provides the define ```FOO_EEPREG_SIZE``` in its header file. This is included in ```eepreg_conf.h``` so that it knows how much EEPROM is being used.
- Includes ```eepreg_conf.h``` in its source files, so that it knows ```FOO_EEPREG_NUM```, which is used as an identifier, e.g. ```eepreg_addr(FOO_EEPREG_NUM)```.

This will probably change though, because as you pointed out, the registry should really be built dynamically.

aabadie:
> If you want you could merge #9524 without eepreg, and add support for the registry in a follow-up PR.

That sounds reasonable and will help moving faster with lorawan support and I think the way EEPROM is used in #9524 allows an application developer to control where things should go on the EEPROM (using CFLAGS, etc).

> I am considering writing the registry information at the beginning of the EEPROM, and then validating that against a dynamically built registry (similar to your recommendation). This would replace the way magic numbers are currently used

Having the registry information at the beginning of the EEPROM is interesting. But I think the magic numbers should still be used by the calling code (a library or an application) and by the registry itself. This will help the registry to find the start address of a given entry. I'm saying this because the index of an entry in the registry might change after a software update while the stored information is still valid. And using a fixed index (some kind of static const variable in a module's code) might cause trouble if the same index is used by 2 different modules.

> This will probably change though, because as you pointed out, the registry should really be built dynamically.

At least we agree here ;)

ZetaR60:
> Having the registry information at the beginning of the EEPROM is interesting. But I think the magic numbers should still be used by the calling code (a library or an application) and by the registry itself. This will help the registry to find the start address of a given entry. I'm saying this because the index of an entry in the registry might change after a software update while the stored information is still valid. And using a fixed index (some kind of static const variable in a module's code) might cause trouble if the same index is used by 2 different modules.

I don't think a magic number should ever be used to actually find an entry, only to confirm that an entry is at an address. Consider a device that has wireless and a remote login password. The password is stored in the EEPROM along with a magic number like this: ```PASSWD\0FOOPASS\0```. If an attacker can inject something into the EEPROM and trigger a search for this magic number, then they can change this password. There is also the possibility that something will be written to the EEPROM that accidentally contains the magic number.

This is what I am thinking should be written to the beginning of the EEPROM:
- A magic number for the registry itself
- The length in bytes of the registry information
- The name for the first section of EEPROM used
- The length of the first section in bytes
- The name for the second section
- The length of the second section
- etc.

The sections of EEPROM used would have no magic number directly in front of them with this method. Having this information would allow you to confirm that each section is in the correct position, and move them during a firmware change if they are not. It also securely prevents mixing of data and metadata.

aabadie:
> This is what I am thinking should be written to the beginning of the EEPROM

This is ok for me and is also what I had in mind.

Issue #3355, opened 2015-07-10
Provide getopt for platforms that don't support it


miri64:
newlib provides `getopt.h` which would implementing commands like `ping6` far easier. Sadly msp430-libc and avr-libc do not provide this header. Conclusively, a lightweight implementation for those platforms would be much appreciated.

N8Fear:
Did you consider to take the getopt stuff from musl? I think musl is a very neatly written libc implementation. It's also relatively modular and doesn't assume too much. It's also MIT-licensed so license shouldn't be a problem either.

MohmadAyman:
I've wrote the lib for the avr, I used some of musl's code and some public domain code plus my own code, so about the licence thing, is there a problem with that -using MIT and public domain code-, do I mention who's code did I use, even thought I did modify like half of it ?
Forgive my so many questions =D, I'm new to riot and to open source development in general.
Thanks.

authmillenon:
About MIT: there are several variants of the MIT license (most notably the Expat and the X11 license, both are compatible to (LGPLv2.1)[http://www.gnu.org/licenses/license-list.html]). If it is compatible to LGPLv2.1 it should be no problem to include them. Public Domain should be no problem either.

MohmadAyman:
thanks for the info about the licensing thing.
I've started a pull request #4106 . but i'm still having some doxgyen errors right now, i'll try to fix them asap.

sudarshansrinivasan:
Hi , 
I am complete newbie to RIOT & open source . Currently  looking at the PRs labeled "Newbie-Task-Candidate".
I wanted to know if this PR# is still open  to work on  and suggestions how proceed with it?
I see some changes were submitted for review in https://github.com/RIOT-OS/RIOT/pull/4106.

Thanks

miri64:
@sudarshansrinivasan for avr-libc have a look at the [comment by @N8Fear above](https://github.com/RIOT-OS/RIOT/issues/3355#issuecomment-120600215). The implementation (the c-file) should go to `cpu/atmega_common` and the header to `cpu/atmega_common/include`. The idea is to implement a feature interchangable with POSIX's getopt functionality: https://docs.python.org/3/library/getopt.html.

@gebart how is the status for this on msp430 (+newlib?)?

miri64:
(`getopt` suffices for now, no need for `getopt_long`)

gebart:
[...]

PR #9538, opened 2018-07-10
sys/optparse: Add command line parser.

jcarrano:
### Contribution description

**UPDATE April 2019**: I completely reworked the API and the code and this description too.

This PR adds a generic command line parser that can be used by shell commands. See #11400 for an example.

#### Features

- Supports short (-c) and long (--color) options and switches in addition to positional arguments (and also optional positional arguments).
- Handling of arguments beginning with "-" by using "--" to mark the end of options.
- **Configuration for parsers can be stored in static structures and arrays (in flash.)**
- Generates friendly error messages and help strings.
- Common functionality included: parse ints, floats, set flags, count occurrences, etc.
- Extensible via user callbacks.

There are two components to this module: the data structures (i.e. the header) and the code itself (i.e. what is run). I'll explain why I consider both separately:

#### Code.

Currently shell commands implement their own parsing leading to:

* Duplicated code and duplicated effort.
* Duplicated bugs.
* Possible inconsistencies in command line argument syntax.
* The parsers end up being pretty basic.

The reason is that parsing anything is usually very easy to get wrong.

#### Data structures

Beyond de-duplicating code, a major motivation for this module is enabling a high-level interface to RIOT commands.

Interaction with RIOT commands is currently mediated by the shell. This is a text interface that is essentially free-form both in its input as well as the output. PR #10624 tries to solve this for the _output_ side. This PR can provide a mechanism by which command _input_ can be delivered. This would involve commands declaring their signature via "opt_conf_t" but, instead of using the code in "optparse.c" to convert lists of strings to values, a special module can be devised specifically for machine-to-machine interaction.

It is out of scope for this PR to explore such a module. The goal of this is to introduce a way to declare command line arguments that decouples the specification from the parsing.

### Testing

I included unit tests. They are not yet complete.

### Issues/PRs references

See: #11400.
For an example in which parsing took more time than it should, see #9523 . It was not a huge deal, but more than it could have been.

Fixes https://github.com/RIOT-OS/RIOT/issues/3355

### DIFF
--- a/sys/Makefile
+++ b/sys/Makefile
@@ -145,6 +145,9 @@ endif
 ifneq (,$(filter bluetil_%,$(USEMODULE)))
   DIRS += net/ble/bluetil
 endif
+ifneq (,$(filter optparse,$(USEMODULE)))
+  DIRS += optparse
+endif
 
 DIRS += $(dir $(wildcard $(addsuffix /Makefile, $(USEMODULE))))
 
--- a/sys/include/optparse.h
+++ b/sys/include/optparse.h
@@ -1,0 +1,260 @@
+/*
+ *      optparse.h
+ *
+ *      Copyright 2010 Juan I Carrano <juan@carrano.com.ar>
+ *      Copyright 2018 Freie Universität Berlin
+ *
+ *      This program is free software; you can redistribute it and/or modify
+ *      it under the terms of the GNU General Public License as published by
+ *      the Free Software Foundation; either version 2 of the License, or
+ *      (at your option) any later version.
+ */
+
+/**
+ * @{
+ * @defgroup sys_optparse   Command line option parser.
+ * @ingroup sys
+ * @file
+ * @brief  Command-line parser
+ * @author Juan I Carrano <j.carrano@fu-berlin.de>
+ *
+ * Terminology:
+ *
+ * option: Optional value. Consists of a key an possibly a value. Each type of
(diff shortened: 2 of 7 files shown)


jcarrano:
@kaspar030 The macros are to help define the setters. However, I'm thinking of removing the setters altogether. The reason I put the setters is to prevent the error where one could forget to set certain member of the structure.

Also, what are your thoughts on splitting the configuration structure so that the parser results are placed somewhere else? That would allow the configuration to be defined in a global const variable.
I'm thinking of making another struct like this:
```c
typedef  union { /**< this used to be in opt_rule */
        int *d_int;
        bool *d_bool;
        double *d_double;
        float *d_float;
        char **d_str;       /**< Pointer to a user-defined pointer */
        void *cb_data;
} opt_ data_t;

typedef struct opt_result {
     void *arg_parser_data; /**< Callback data for arg_parser. moved from opt_conf */
     opt_data_t *rules_result; /*<  */
} opt_result_t;
```

kaspar030:
BTW, did you check out existing code, like https://github.com/cofyc/argparse or https://github.com/docopt/docopt.c?

jcarrano:
@kaspar030 docopt generates code (ugh!). argparse is quite similar to this, and seems good, but it does not support as many option types as this module.
In addition, as I mentioned before, I'm thinking it may be a good idea to modify this module so that the configuration is strictly read-only, but I wanted to know what other people thought.

Issue #9371, opened 2018-06-19
assert: c99 static_assert macro doesn't function for multiple static_asserts in the same scope


pwillem:
#### Description

c99 static assert introduced in 2bd1d7f045 doesn't allow multiple static_asserts in the same scope. GCC throws `/RIOT/core/include/assert.h:122:8: error: redeclaration of enumerator ‘static_assert_failed_on_div_by_0’`

#### Steps to reproduce the issue

1. add to examples/hello-world/main.c
``` C
 #include <stdio.h>
+#include <assert.h>

 int main(void)
 {
+    static_assert(sizeof(int) == sizeof(unsigned int), "int vs unsigned int");
+    static_assert(sizeof(int) != sizeof(long long int), "int vs long long int");
+
     puts("Hello World!");

     printf("You are running RIOT on a(n) %s board.\n", RIOT_BOARD);

```
2. make -C examples/hello-world
#### Expected results

example/hello-world should compile

#### Actual results

redeclaration error during compilation:
``` C
~/src/RIOT/examples/hello-world/main.c: In function ‘main’:
~/RIOT/core/include/assert.h:122:12: error: redeclaration of enumerator ‘static_assert_failed_on_div_by_0’
     enum { static_assert_failed_on_div_by_0 = 1 / (!!(cond)) }
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~/RIOT/examples/hello-world/main.c:28:5: note: in expansion of macro ‘static_assert’
     static_assert(sizeof(int) != sizeof(long long int), "int vs long long int");
     ^~~~~~~~~~~~~
~/RIOT/core/include/assert.h:122:12: note: previous definition of ‘static_assert_failed_on_div_by_0’ was here
     enum { static_assert_failed_on_div_by_0 = 1 / (!!(cond)) }
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~/RIOT/examples/hello-world/main.c:27:5: note: in expansion of macro ‘static_assert’
     static_assert(sizeof(int) == sizeof(unsigned int), "int vs unsigned int");
     ^~~~~~~~~~~~~
```
#### Versions

```
Installed compiler toolchains
-----------------------------
             native gcc: gcc (GCC) 8.1.1 20180531
      arm-none-eabi-gcc: arm-none-eabi-gcc (Arch Repository) 8.1.0
                avr-gcc: avr-gcc (GCC) 8.1.0
[...]

PR #5824, opened 2016-09-05
make: use -std=c99 as default

kaspar030:
Previously, the fallback was "gnu99", but we don't allow GNU extensions by convention.

PR #9544, opened 2018-07-11
cflags: use gnu99 only if nothing else is specified (mips, msp430, native)

vincent-d:
### Contribution description

On mips, msp430 and native, the `-std` option was forced to `gnu99` even though the user changed it.
This propose to change it only if nothing else is passed in the `CFLAGS` by adding a makefile variable and use it in the central cflags.inc.mk

### Issues/PRs references

Workaround #9371 by allowing a user to switch to c11.
First gnu99 changes were introduced in #5824

### DIFF
--- a/boards/native/Makefile.include
+++ b/boards/native/Makefile.include
@@ -47,7 +47,11 @@ export CGANNOTATE ?= cg_annotate
 export GPROF ?= gprof
 
 # basic cflags:
-export CFLAGS += -Wall -Wextra -pedantic -std=gnu99
+export CFLAGS += -Wall -Wextra -pedantic
+# default std set to gnu99 of not overwritten by user
+ifeq (,$(filter -std=%, $(CFLAGS)))
+  export CFLAGS += -std=gnu99
+endif
 
 ifeq ($(shell uname -m),x86_64)
   export CFLAGS += -m32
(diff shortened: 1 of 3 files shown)


kaspar030:
How about adding ```CFLAGS_STD```? If evaluated using ```CFLAGS_STD ?= foo``` at the right spots, we might save some of the logic.

vincent-d:
> How about adding CFLAGS_STD?

While I also believe it could save some of the logic, it will break any existing app already using `CFLAGS += -std=...` so I'm not sure if it's the best choice.

vincent-d:
Renamed the variable to `STD_DEFAULT_GNU_99` and added to `vars.inc.mk`. I don't know if it should be exported or not, though.

vincent-d:
Ping @gebart @kaspar030

gebart:
After some consideration, I am reluctant to add yet another environment variable to the build system. What do you think about using this instead? 
```make
ifeq (, filter(-std=%, $(CFLAGS)))
CFLAGS += -std=gnu99
endif
```

vincent-d:
@gebart you're right it's actually easier. See the last update

gebart:
Initial Murdock failure was unrelated, retry fixed it. -> go

PR #9555, opened 2018-07-12
openthread: Enable RX complete netdev event

bergzand:
Fixes "issue" where openthread assumed that this flag was enabled by
default with netdev devices. Most, if not all devices enable this flag by default so the problem did not show itself.

### Issues/PRs references

None


jia200x:
@bergzand good catch! Thanks

PR #9557, opened 2018-07-12
Arduino: fix several boards with C++ issues due to new I2C params

smlng:
### Contribution description

Fix C++ related issues for some boards, ie. struct attributes need to be init in order.

### DIFF
--- a/boards/nucleo-f103/Makefile.features
+++ b/boards/nucleo-f103/Makefile.features
@@ -6,9 +6,6 @@ FEATURES_PROVIDED += periph_spi
 FEATURES_PROVIDED += periph_timer
 FEATURES_PROVIDED += periph_uart
 
-# load the common Makefile.features for Nucleo boards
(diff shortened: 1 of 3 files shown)


boards/nucleo-f103/Makefile.features
```diff
@@ -6,9 +6,6 @@ FEATURES_PROVIDED += periph_spi
 FEATURES_PROVIDED += periph_timer
 FEATURES_PROVIDED += periph_uart
 
-# load the common Makefile.features for Nucleo boards
-include $(RIOTBOARD)/common/nucleo64/Makefile.features
```
dylad: Why removing this ?

smlng: because then this board will not have the Arduino feature anymore and won't be used for the Arduino hello world examples :) other boards (like the f410/f411) don't have this feature either - I guess it was a copy paste for the f103 anyway, but never clashed before.

PR #8769, opened 2018-03-12
boards: add Ublox C030-U201

vincent-d:
### Contribution description

This adds the support of the [Ublox C030-U201](https://www.u-blox.com/en/product/c030-app-board) board. This is a stm32f4-based board (the cpu support is also added here) with a Ublox 3G modem (SARA-U201) and GPS (MAX-M8C).
Most of the peripheral are added from the datasheet/schematics. I tested the stdio UART, LEDs and button through saul and xtimer.

### Issues/PRs references

Can be a good base to work on #8602

PR #9558, opened 2018-07-12
boards/ublox-c030-u201: adapt to new I2C interface

dylad:
### Contribution description
This PR intends to repair a break from the ongoing I2C embargo.
I cherry-picked the Ublox commit from #8769 then I adapt the periph_conf to the new i2C interface.
Ublox board will still not compiles on this branch because I did not cherry-pick the commit that add support for the STM32 CPU (cause a merge conflict) but it will compiles once the branch get rebased into master (tested locally)

### Issues/PRs references
see #8769 which breaks the embargo

### DIFF
--- a/boards/ublox-c030-u201/include/periph_conf.h
+++ b/boards/ublox-c030-u201/include/periph_conf.h
@@ -198,47 +198,37 @@ static const spi_conf_t spi_config[] = {
  * @name I2C configuration
  * @{
  */
-#define I2C_0_EN            1
-#define I2C_1_EN            1
-#define I2C_NUMOF           2
-#define I2C_IRQ_PRIO        1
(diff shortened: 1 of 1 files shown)


smlng:
@dylad you missed the commit with the cpu from #8769 - also it would be better to cherry-pick the commit into the new_i2c_if branch first, or not?

dylad:
@smlng I did intentionately, I cannot cherry-pick this commit because of a merge conflict, but il will runs fine once the branch get rebased (tested locally)

> also it would be better to cherry-pick the commit into the new_i2c_if branch first, or not?

I don't know if there is any difference TBH.

dylad:
I guess it's better to add the cherry-picked commit here so we will keep trace of it in the PR history of github.

PR #9562, opened 2018-07-12
mrf24j40: Fix netdev recv implementation

bergzand:
The mrf24j40 driver should return the frame length when both buf is NULL
and len is zero and drop the packet when len is nonzero and buf is NULL.
This PR fixes that behaviour

### Issues/PRs references

None

### DIFF
--- a/drivers/mrf24j40/mrf24j40_netdev.c
+++ b/drivers/mrf24j40/mrf24j40_netdev.c
@@ -110,46 +110,45 @@ static int _recv(netdev_t *netdev, void *buf, size_t len, void *info)
     mrf24j40_t *dev = (mrf24j40_t *)netdev;
     uint8_t phr;
     size_t pkt_len;
+    int res = -ENOBUFS;
 
     /* Disable receiving while reading the RX fifo (datasheet sec. 3.11.4) */
     mrf24j40_reg_write_short(dev, MRF24J40_REG_BBREG1, MRF24J40_BBREG1_RXDECINV );
 
-    /* get the size of the received packet */
+    /* get the size of the received packet, this resets the receive FIFO to be
+     * ready for a new frame, so we have to disable RX first (see above) */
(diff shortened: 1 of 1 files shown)


drivers/mrf24j40/mrf24j40_netdev.c
```diff
@@ -110,46 +110,45 @@ static int _recv(netdev_t *netdev, void *buf, size_t len, void *info)
...
-        mrf24j40_rx_fifo_read(dev, phr + 1, &(radio_info->lqi), 1);
-        mrf24j40_rx_fifo_read(dev, phr + 2, &(rssi_scalar), 1);
-        radio_info->rssi = mrf24j40_dbm_from_reg(rssi_scalar);
+        res = pkt_len;
     }
-
     /* Turn on reception of packets off the air */
     mrf24j40_reg_write_short(dev, MRF24J40_REG_BBREG1, 0x00);
-    return pkt_len;
+    /* Return -ENOBUFS if a too small buffer is supplied. Return packet size
```
PeterKietzmann: Why don't you need to flush the rx buffer anymore?
`mrf24j40_reg_write_short(dev, MRF24J40_REG_RXFLUSH, MRF24J40_RXFLUSH_RXFLUSH);`

bergzand: Page 109 of the mrf24j40 datasheet describes the ways to reset the RX FIFO pointer. writing the RXFLUSH bit is one way, but it is also reset by reading the first byte of the received frame. As we always [read the first bytes of the frame](https://github.com/RIOT-OS/RIOT/pull/9562/files#diff-c64be8fae4f7b67b2428da522851b68eR120) this extra flush is not necessary.

PeterKietzmann: Nice!

bergzand:
Thanks for the review and testing!

PR #9563, opened 2018-07-12
gnrc_netif_ieee802154: drop frame on buffer error

bergzand:
### Contribution description

This adds a netdev recv call to indicate that the received frame should be
dropped when there is no buffer space available to store the frame.

`gnrc_netif_ethernet.c` already has this call [here](https://github.com/RIOT-OS/RIOT/blob/master/sys/net/gnrc/netif/gnrc_netif_ethernet.c#L175)

### Issues/PRs references

None


bergzand:
Just checked, `gnrc_netif_raw.c` also already handles this case.

PR #9568, opened 2018-07-12
nrf52xxxdk: match i2c pins with arduino layout

kYc0o:
### Contribution description
While experimenting with the X-NUCLEO-IKS01A1, I realised that the I2C pins were not mapped to the Arduino shield. I don't know if this has a reason but for me it seems legit to make it match, since it allows the use of Arduino shields using I2C.

I think this doesn't touch any of the files for the I2C refactoring.

### Issues/PRs references
None.

### DIFF
--- a/boards/common/nrf52xxxdk/include/periph_conf.h
+++ b/boards/common/nrf52xxxdk/include/periph_conf.h
@@ -100,8 +100,8 @@ static const spi_conf_t spi_config[] = {
 static const i2c_conf_t i2c_config[] = {
     {
         .dev = NRF_TWIM1,
-        .scl = 28,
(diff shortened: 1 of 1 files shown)


dylad:
Sorry @kYc0o but it does :p 
Hopefully, the I2C refactoring is really really close to an end.

dylad:
@kYc0o We can merge this PR once #9548 is merged ;)

aabadie:
IIRC, this is already done in new_i2c_if (I also faced the same issue with nrf52dk)

kYc0o:
Yes, it's done there but I didn't find any description of it. This PR is then useless.

PR #9570, opened 2018-07-13
backport: board stm32mindev

smlng:
### Contribution description

This board was merged in between and has I2C support, hence it needs adaption which is only possible by backporting it first. Followup PR will do the required changes.


smlng:
not needed

PR #9571, opened 2018-07-13
cpp: fix board I2C configs for C++ support

smlng:
### Contribution description

The I2C configuration structs of several boars causes problems when uses with C++, e.g., attributes were missing or not initialised in correct order.


smlng:
wait for murdock!

dylad:
Let's merge it !!!

boards/common/iotlab/include/periph_conf_common.h
```diff
@@ -149,6 +149,8 @@ static const i2c_conf_t i2c_config[] = {
         .speed          = I2C_SPEED_NORMAL,
         .scl_pin        = GPIO_PIN(PORT_B, 6),
         .sda_pin        = GPIO_PIN(PORT_B, 7),
+        .scl_af         = GPIO_AF_OUT_OD,
```
aabadie: all those changes are useless since af cannot be changed on STM32 F1.
Normally, the scl_af and sda_af attributes are not used when the CPU is stm32f1.

dylad: This change is needed because of CPP check, @smlng also modified the driver accordingly.

smlng: @aabadie as @dylad wrote, I changed the driver accordingly omitting the `#ifdef` in there. However, I agree with you as this is kind of non-dynamic and fixed for stm32f1 CPUs it should not be exposed in the configuration of the board, which suggests it might be dynamic. But this is a general problem in many drivers and their config: i.e, that RIOT exposes params in configs that are actually not configurable at all. 

Nevertheless, C++ requires to init all struct attributes properly and in order, omitting one or more or changing the order will result in compile errors. Hence I had to expose the scl/sda_af in the config.

aabadie:
> Hence I had to expose the scl/sda_af in the config

and this is bad. look at uart and spi structs in periph_common header, the af attributes are skipped for F1, the same should be done for I2C. I think something went wrong there during the rebase.

dylad:
We don't have the choice here. If we don't add scl_af sda_af to the periph_conf, CPP apps won't compile so Murdock will never be happy.

aabadie:
> CPP apps won't compile so Murdock will never be happy.

I know but this can be done globally in the CPU code, not in each board peripheral config.

smlng:
> this can be done globally in the CPU code
No it can't @aabadie , at least not as long as all STM CPUs share the same `i2c_config_t` with `scl_af` and `sda_af` attributes, then C++ requires them to be properly initialised (and in order). The only working alternative would be to have a distinct `i2c_config_t` for STM32F1 CPUs w/o the `_af` attributes.

aabadie:
I'm pretty sure it can. The same struct is used by 2 different driver implementation. Just use `#ifndef CPU_FAM_STM32F1/#endif` around the af attribute. Did you try that ?

aabadie:
> all those #ifdefs are also kind of bad.

with unified driver implementations, I don't see a better way. For me, code duplication is even worse

dylad:
Why is it better to drop af params ? IMHO its better to have the struct for every family

aabadie:
I'm not saying that we don't want the struct to configure i2c F1 boards.
I'm saying that the af attributes are useless for this family because it's something that cannot be changed/configured. Look at the code in master (for iotlab boards for example), there's nothing for af modes.
The af attributes must be skipped for F1 with an #ifndef in the i2c struct definition (in cpu/stm32_common/include). This avoids code duplication in the struct initializers and save a couple of bytes. Note that the same strategy is used for uart and spi. 
As I said already, the 'same' i2c struct is alreasy used for different families (2 types of drivers) and some don't use all attributes already (bus clock, etc).
@vincent-d, if you are around can you have a look and give your opinion ?

vincent-d:
I agree with @aabadie here, we should do as for uart and spi, see: https://github.com/RIOT-OS/RIOT/blob/446ff9f282875e5867b746c984821d4032ddbb7e/cpu/stm32_common/include/periph_cpu_common.h#L333 with uart driver for instance. Same struct but some fields are out for F1.

vincent-d:
Forget it, I missed #9574 ;)

PR #9574, opened 2018-07-14
i2c: cpp check fixes

dylad:
### Contribution description
This PR should fix all CPP issues for new_i2c_if feature branch, I cherry-picked some commits from @smlng and #9571. I also include the required changes from @aabadie (see #9571 discussion)

Once this PR get merged, I'll rebase new_i2c_if3 to see if Murdock is happy.
This way it should ease comparison between original new_i2c_if branch and new_i2c_if3 which will be rebased onto master.

### Issues/PRs references
#9571

### DIFF
--- a/boards/acd52832/include/periph_conf.h
+++ b/boards/acd52832/include/periph_conf.h
@@ -104,7 +104,8 @@ static const i2c_conf_t i2c_config[] = {
     {
         .dev = NRF_TWIM1,
         .scl = 28,
-        .sda = 29
+        .sda = 29,
+        .speed = I2C_SPEED_NORMAL
     }
 };
 
--- a/boards/feather-m0/include/periph_conf.h
+++ b/boards/feather-m0/include/periph_conf.h
@@ -212,8 +212,8 @@ static const i2c_conf_t i2c_config[] = {
     {
         .dev      = &(SERCOM3->I2CM),
         .speed    = I2C_SPEED_NORMAL,
-        .sda_pin  = GPIO_PIN(PA, 22),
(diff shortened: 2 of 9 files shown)


dylad:
Murdock still fails because ublox and EFM32 boards need to be rebased onto master but I prefer rebase a cloned branch rather than this branch to keep tracks of the changes.

aabadie:
Thanks @dylad, regarding STM32F1, this is ok now.

boards/acd52832/include/periph_conf.h
```diff
@@ -104,7 +104,8 @@ static const i2c_conf_t i2c_config[] = {
     {
         .dev = NRF_TWIM1,
         .scl = 28,
-        .sda = 29
+        .sda = 29,
+        .speed = I2C_SPEED_NORMAL
```
cladmi: As you said, this one is actually more than a warning as I2C_SPEED_NORMAL is actually not 0. Thanks cpp check :)

cladmi:
> Murdock still fails because ublox and EFM32 boards need to be rebased onto master but I prefer rebase a cloned branch rather than this branch to keep tracks of the changes.

You mean that it is just Murdock that fails, and would fail without it anyway and that it can be merged even with the error ?

The changes look good to me.

dylad:
This PR fixes a lot of issues with Murdock (all CPPcheck in fact) but with or without this PR, Murdock fails with EFM32 based board and the ublox board. Theses failures are not related to this PR. 
And the remaining Murdock failures will be fix with the rebase.

cladmi:
Ok that what I was thinking but was not sure.

PR #9579, opened 2018-07-16
stm32_common: fix ram length for MCUs with non-contiguous ram regions

### LINKED ISSUE #9534: linkerscripts for stm32 incorrect for MCUs with non-contiguous SRAM1 & SRAM2
The common linker script for stm32 based platforms doesn't take into account that on some platforms the available RAM is split up in different RAM sections e.g. SRAM1 & SRAM2. 
On some MCUs there is a hardware feature to remap the address space of SRAM2 consecutively after SRAM1, there it doesn't actually impose a problem. This feature is not available on all MCUs though.
On stm32l476rg for example SRAM1 is at 0x20000000 while SRAM2 resides at 0x10000000. After SRAM1 unmapped address space is located (see [RM0351 reference manual page 73](https://www.st.com/content/ccc/resource/technical/document/reference_manual/02/35/09/0c/4f/f7/40/03/DM00083560.pdf/files/DM00083560.pdf/jcr:content/translations/en.DM00083560.pdf#page=73)).

I stumbled across this when testing low power RAM retention which is only available on SRAM2 on the named MCU. 

As a minimum fix the RAM_LEN value in [stm32_mem_lengths.mk](https://github.com/RIOT-OS/RIOT/blob/master/cpu/stm32_common/stm32_mem_lengths.mk) for affected platforms should be reduced to the minimum consecutively addressable amount (96K in the above case).

A better fix would need to add another ram section to [cortexm.ld](https://github.com/RIOT-OS/RIOT/blob/master/cpu/cortexm_common/ldscripts/cortexm.ld) and [cortexm_base.ld](https://github.com/RIOT-OS/RIOT/blob/master/cpu/cortexm_common/ldscripts/cortexm_base.ld) and respectively additional parameters to [stm32_mem_lengths.mk](https://github.com/RIOT-OS/RIOT/blob/master/cpu/stm32_common/stm32_mem_lengths.mk). 

I'd like to use this Issue to collect which other MCUs are affected by this (if any) and discussion on how to properly integrate the split RAM to the linker scripts before opening up a PR. Feel free to help, of course ;)

_edit_
Just checked the RIOT-supported L4 MCUs.

no automatic remap (problem exists):
stm32l475vg (b-l475e-iot01a)
stm32l476rg (nucleo-l476rg)
stm32l476vg (stm32l476g-disco)

automatic remap (no problem):
stm32l432kc
stm32l433rc
stm32l452re
stm32l496zg

MichelRottleuthner:
### Contribution description
On some MCUs the ram sizes specified in stm32_mem_lengths.mk are based on the sum of available ram, while it is physically not mapped to a contiguous ram region. This will make the linker place data/code in unmapped address space if enough ram is used. See #9534 for details.
This PR fixes the problem by decreasing the ram sizes to the size of the "primary" ram region.

For testing the following snippet may be helpful:

```c
extern uint32_t _sram;
extern uint32_t _eram;

int main(void)
{
    printf("ram start @0x%p\n", &_sram);
    printf("ram end   @0x%p\n", &_eram);

    printf("last value in ram @0x%p: 0x%08lX\n", &_eram - 1, *(&_eram - 1));

    return 0;
}
```

On stm32-l476rg without the fix the above code crashes:
```
ram start @0x0x20000000
ram end   @0x0x20020000

Context before hardfault:
   r0: 0x00000018
   r1: 0x20000d18
   r2: 0xa0000000
   r3: 0x12000000
  r12: 0x08002102
   lr: 0x08000cd1
   pc: 0x08000cd0
  psr: 0x21000000

FSR/FAR:
 CFSR: 0x00008200
 HFSR: 0x40000000
 DFSR: 0x00000008
 AFSR: 0x00000000
 BFAR: 0x2001fffc
Misc
EXC_RET: 0xfffffffd
In GDB:
  set $pc=0x8000cd0
  frame 0
  bt

ISR stack overflowed by at least 16 bytes.
```

with the fix it works as expected:

```
ram start @0x0x20000000
ram end   @0x0x20018000
last value in ram @0x0x20017ffc: 0x99E5B210
```

A thing for future PRs: the currently unused ram regions (i.e. SRAM2 on L4 and CCMRAM on F3) should be properly added to the linker to allow using them in applications. Also a test to verify ram sizes may be a good thing.

### Issues/PRs references
Fixes #9534

### DIFF
--- a/cpu/stm32_common/stm32_mem_lengths.mk
+++ b/cpu/stm32_common/stm32_mem_lengths.mk
@@ -235,6 +235,7 @@ else ifeq ($(STM32_TYPE), L)
       RAM_LEN = 64K
     else ifeq ($(STM32_MODEL2), 7)
       RAM_LEN = 96K
+      RAM2_LEN = 32K
     else ifeq ($(STM32_MODEL2), 5)
       RAM_LEN = 160K
     else ifeq ($(STM32_MODEL2), 9)
--- a/cpu/stm32_common/stm32_mem_lengths.mk
+++ b/cpu/stm32_common/stm32_mem_lengths.mk
@@ -120,19 +120,20 @@ ifeq ($(STM32_TYPE), F)
       endif
     else ifeq ($(STM32_MODEL), 303)
       ifneq (, $(filter $(STM32_ROMSIZE), 6 8))
-        RAM_LEN = 16K
+        RAM_LEN = 12K
         CCMRAM_LEN = 4K
       else ifeq ($(STM32_ROMSIZE), B)
         RAM_LEN = 40K
         CCMRAM_LEN = 8K
       else ifeq ($(STM32_ROMSIZE), C)
-        RAM_LEN = 48K
+        RAM_LEN = 40K
         CCMRAM_LEN = 8K
       else ifneq (, $(filter $(STM32_ROMSIZE), D E))
-        RAM_LEN = 80K
+        RAM_LEN = 64K
+        CCMRAM_LEN = 16K
       endif
     else ifeq ($(STM32_MODEL3), 4)
-      RAM_LEN = 16K
+      RAM_LEN = 12K
       CCMRAM_LEN = 4K
     else ifeq ($(STM32_MODEL), 373)
       RAM_LEN = 32K
@@ -233,7 +234,7 @@ else ifeq ($(STM32_TYPE), L)
     ifeq ($(STM32_MODEL2), 3)
       RAM_LEN = 64K
     else ifeq ($(STM32_MODEL2), 7)
-      RAM_LEN = 128K
+      RAM_LEN = 96K
     else ifeq ($(STM32_MODEL2), 5)
       RAM_LEN = 160K
     else ifeq ($(STM32_MODEL2), 9)
--- a/examples/gcoap/Makefile
+++ b/examples/gcoap/Makefile
@@ -11,7 +11,8 @@ RIOTBASE ?= $(CURDIR)/../..
 
 BOARD_INSUFFICIENT_MEMORY := chronos msb-430 msb-430h nucleo-f031k6 nucleo-f042k6 \
                              nucleo-l031k6 nucleo-f030r8 nucleo-f334r8 nucleo-l053r8 \
-                             stm32f0discovery telosb wsn430-v1_3b wsn430-v1_4 z1
+                             stm32f0discovery telosb wsn430-v1_3b wsn430-v1_4 z1 \
+                             nucleo-f303k8
 
 ## Uncomment to redefine port, for example use 61616 for RFC 6282 UDP compression.
 #GCOAP_PORT = 5683
--- a/examples/nanocoap_server/Makefile
+++ b/examples/nanocoap_server/Makefile
@@ -9,7 +9,7 @@ RIOTBASE ?= $(CURDIR)/../..
 
 BOARD_INSUFFICIENT_MEMORY := chronos msb-430 msb-430h nucleo-f031k6 nucleo-f042k6 \
                              nucleo-l031k6 nucleo-f030r8 nucleo-l053r8 stm32f0discovery \
-                             telosb z1
+                             telosb z1 nucleo-f303k8
 
 # Include packages that pull up and auto-init the link layer.
 # NOTE: 6LoWPAN will be included if IEEE802.15.4 devices are present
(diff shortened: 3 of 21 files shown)


kYc0o:
If I understand correctly, the separated RAM (outside 0x20000000) is not currently usable?

For the CCM I understand that has an special usage, but I don't get then how the second part (apparently at 0x10000000) can be used.

I'd ACK your change regarding the correct mapping, however I'd like to check how difficult it could be to actually use the ram at that address space.

MichelRottleuthner:
>I'd ACK your change regarding the correct mapping, however I'd like to check how difficult it could be to >actually use the ram at that address space.

depends on how you define "using it".
I think easiest from the developer perspective would be to enable the linker to make use of all available ram "in the most reasonable way". There is an [application report by TI](http://www.ti.com/lit/an/spraa46a/spraa46a.pdf#page=4) that describes quite nicely the necessary splitting feature of the linker to achieve that. According to [this thread](https://answers.launchpad.net/gcc-arm-embedded/+question/223976) the gnu toolchain doesn't have this feature though and I can't think of a workaround.

To enable the extra ram space we need something like I did [here](https://github.com/MichelRottleuthner/RIOT/commit/f93f399babf79e67555754178cccc0ef9e41753c#diff-9dfccc95a25f8fcd8e88384e18cd94b9). Using the additional ram can then be done manually like [this](https://github.com/RIOT-OS/RIOT/commit/b210f117d7d0a4302208a65b5f7a6174d1399ed0#diff-6a756b820c59b3a309af47469bf94bb2R38). Doing it manually may be preferable to let the developer decide what should be put to the additional ram, because of the additional features it provides (e.g. parity check, low-power retention, write protection, higher performance).
To get this independent of the toolchain, we could introduce macros like PLACE_IN_RAM_XY which maps to whatever is needed to explicitly place stuff in the approriate section (e.g. `__attribute__((section("name")))`).

In case of CCM we should actually prefer to place code segments there to achieve better performance due to ["optimum Harvard configuration"](https://www.st.com/content/ccc/resource/technical/document/application_note/bb/09/ca/83/14/e9/44/c5/DM00083249.pdf/files/DM00083249.pdf/jcr:content/translations/en.DM00083249.pdf#page=5).

I'm not sure this all should be in the same PR. But I'm pretty sure this needs some discussion ;)

cpu/stm32_common/stm32_mem_lengths.mk
```diff
@@ -233,7 +234,7 @@ else ifeq ($(STM32_TYPE), L)
     ifeq ($(STM32_MODEL2), 3)
       RAM_LEN = 64K
     else ifeq ($(STM32_MODEL2), 7)
-      RAM_LEN = 128K
+      RAM_LEN = 96K
```
vincent-d: I think a comment or another variable for SRAM2 would be worth

MichelRottleuthner: hmm wouldn't hurt but I tried to follow "don't add stuff that may be used later, but isn't now". I have no strong opinion on that though.

MichelRottleuthner:
I hoped this won't happen, but it looks like some applications won't fit to the smaller ram anymore. I wonder how those apps actually worked before o.O

kYc0o:
> Doing it manually may be preferable to let the developer decide what should be put to the additional ram, because of the additional features it provides (e.g. parity check, low-power retention, write protection, higher performance).

Then I'd say to have this as the main RAM source and the we can discuss what to do with the "normal" RAM. Is this RAM big enough to fit all the applications that were enabled before?

kYc0o:
> I hoped this won't happen, but it looks like some applications won't fit to the smaller ram anymore. I wonder how those apps actually worked before o.O

These applications were always compiled but AFAIK never tested on hardware, which of course would lead to hard faults.

MichelRottleuthner:
As you can see in [stm32_mem_lengths.mk](https://github.com/RIOT-OS/RIOT/pull/9579/files#diff-64b80cd5a8fbc742325d74ed3eeb648aR137) the CCMRAM (and also SRAM2) are way smaller than the SRAM1 (e.g. 4K vs. 16K). So making it the default will probably create more problems than it solves. I haven't checked yet how much of the code are actually instructions, but I think for a lot of (most?) applications it's to small to put all instructions there. Maybe mapping some selected OS functionality there would make sense (periph drivers, msg, xtimer, scheduler etc.). In that case we would still need to think of a solution that doesn't break anything for other platforms and is compatible to the various toolchains used.

kYc0o:
Ok, then is unusable. Let's just mark the boards with insufficient memory on the applications that can't fit and move on.

kYc0o:
Please use a shorter comment for the first commit, add more lines below explaining the change and if possible reference the datasheet, although I'm fine with just naming the affected CPUs.

MichelRottleuthner:
@kYc0o fine with the commit message like this?

kaspar030:
@cladmi IMO this needs a backport, what do you think? if yes, @bergzand would you like to try your magic on it?

kYc0o:
> @kYc0o fine with the commit message like this?

Perfect! Thank you so much for your contribution!

cladmi:
@kaspar030 I can do the backport but cannot test it

cladmi:
Backport provided in #9598

PR #9581, opened 2018-07-17
at86rf2xx: Move flags from netdev to radio

bergzand:
### Contribution description

The flags from the ieee802154 struct are only used in the radio code, there is no advantage of having make sense to have them in the netdev struct if their meaning is different per radio driver.

Also fixes a bug where (un)setting the `AUTOACK` flag would (un)set the `ACK_REQ` flag.

### Issues/PRs references

Minor cleanup preparing for #7736 

Similar PRs are required for the other radio drivers using flags.

### DIFF
--- a/drivers/at86rf2xx/at86rf2xx.c
+++ b/drivers/at86rf2xx/at86rf2xx.c
@@ -161,7 +161,7 @@ void at86rf2xx_tx_exec(const at86rf2xx_t *dev)
     at86rf2xx_reg_write(dev, AT86RF2XX_REG__TRX_STATE,
                         AT86RF2XX_TRX_STATE__TX_START);
     if (netdev->event_callback &&
-        (dev->netdev.flags & AT86RF2XX_OPT_TELL_TX_START)) {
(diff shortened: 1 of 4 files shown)


bergzand:
This changes the default flags enabled for the radio (`NETOPT_AUTOACK` is no longer enabled by default).

smlng:
because AUTOACK is enable on `at86rf2xx_reset` function anyways.

PeterKietzmann:
Green lights @smlng!

smlng:
@bergzand question: what was `AT86RF2XX_OPT_RAWDUMP` for, any idea? It wasn't used before as far as I see, but you the define here.

bergzand:
@smlng Looking at changes from [a long time ago](https://github.com/RIOT-OS/RIOT/blob/8f9e1656e1a00e496e4b13411aa5191ed083a479/drivers/at86rf2xx/at86rf2xx_netdev.c), this was from when part of what is now handled in netif. I would say similar to the current `NETOPT_RAWMODE` flags.

PR #9583, opened 2018-07-17
mrf24j40: Move flags from netdev to radio

bergzand:
### Contribution description

The flags from the ieee802154 struct are only used in the radio code, there is no advantage of having make sense to have them in the netdev struct if their meaning is different per radio driver.

Also fixes a bug where the `NETOPT_AUTOACK` doesn't work, but is dis/enabled with the `NETOPT_ACK_REQ` setting.

### Issues/PRs references

Minor cleanup preparing for #7736

Similar PRs are required for the other radio drivers using flags.

### DIFF
--- a/drivers/include/mrf24j40.h
+++ b/drivers/include/mrf24j40.h
@@ -104,8 +104,7 @@ extern "C" {
                                                      *   start */
 #define MRF24J40_OPT_TELL_RX_END        (0x4000)    /**< notify MAC layer on RX
                                                      *   finished */
-#define MRF24J40_OPT_REQ_AUTO_ACK       (0x8000)    /**< notify MAC layer on RX
-                                                     *   finished */
+#define MRF24J40_OPT_AUTOACK            (0x8000)    /**< Auto ack RX frames */
 /** @} */
 
 
--- a/drivers/include/mrf24j40.h
+++ b/drivers/include/mrf24j40.h
@@ -92,19 +92,19 @@ extern "C" {
  *
  * @{
  */
-#define MRF24J40_OPT_CSMA               (0x0100)    /**< CSMA active */
-#define MRF24J40_OPT_PROMISCUOUS        (0x0200)    /**< promiscuous mode
+#define MRF24J40_OPT_CSMA               (0x01)      /**< CSMA active */
+#define MRF24J40_OPT_PROMISCUOUS        (0x02)      /**< promiscuous mode
                                                      *   active */
-#define MRF24J40_OPT_PRELOADING         (0x0400)    /**< preloading enabled */
-#define MRF24J40_OPT_TELL_TX_START      (0x0800)    /**< notify MAC layer on TX
+#define MRF24J40_OPT_PRELOADING         (0x04)      /**< preloading enabled */
+#define MRF24J40_OPT_TELL_TX_START      (0x08)      /**< notify MAC layer on TX
                                                      *   start */
-#define MRF24J40_OPT_TELL_TX_END        (0x1000)    /**< notify MAC layer on TX
+#define MRF24J40_OPT_TELL_TX_END        (0x10)      /**< notify MAC layer on TX
                                                      *   finished */
-#define MRF24J40_OPT_TELL_RX_START      (0x2000)    /**< notify MAC layer on RX
+#define MRF24J40_OPT_TELL_RX_START      (0x20)      /**< notify MAC layer on RX
                                                      *   start */
-#define MRF24J40_OPT_TELL_RX_END        (0x4000)    /**< notify MAC layer on RX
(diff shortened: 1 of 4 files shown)


bergzand:
This changes the default flags enabled for the radio (`NETOPT_ACK_REQ` is no longer enabled by default), I have to take a look at how to solve this.

drivers/include/mrf24j40.h
```diff
@@ -104,8 +104,7 @@ extern "C" {
                                                      *   start */
 #define MRF24J40_OPT_TELL_RX_END        (0x4000)    /**< notify MAC layer on RX
                                                      *   finished */
-#define MRF24J40_OPT_REQ_AUTO_ACK       (0x8000)    /**< notify MAC layer on RX
-                                                     *   finished */
+#define MRF24J40_OPT_AUTOACK            (0x8000)    /**< Auto ack RX frames */
```
smlng: why not use `0x01` to `0x80` here

drivers/include/mrf24j40.h
```diff
@@ -133,6 +132,7 @@ typedef struct {
     netdev_ieee802154_t netdev;             /**< netdev parent struct */
     /*  device specific fields  */
     mrf24j40_params_t params;               /**< parameters for initialization */
+    uint16_t flags;                         /**< Internal device flags */
```
smlng: and make this an `uint8_t` which would be enough for the flags above.

bergzand:
> question: is it possible/worth to safe a byte here?

Should be possible, but there are still some random "flags" set here and there in the driver. I have to investigate whether they actually trigger something or that they were useless from the start.

PeterKietzmann:
@bergzand any progress?

bergzand:
> Should be possible, but there are still some random "flags" set here and there in the driver. I have to investigate whether they actually trigger something or that they were useless from the start.

@smlng: Looked around and had to remove two calls that set obsolete flags. The driver should still work (as far as I've tested). Not sure if it was the case before, but the ack_req setting is not set at boot with this PR.

PR is also rebased to latest master.

drivers/include/mrf24j40.h
bergzand: Agreed, Changed

drivers/include/mrf24j40.h
bergzand: also changed

PeterKietzmann:
> but the ack_req setting is not set at boot with this PR

But this is not intended, right?

bergzand:
> But this is not intended, right?

That depends on whether you consider the old implementation/behaviour a bug or a feature :)

PeterKietzmann:
I just wanted to agree that ACKREQ should be enabled by default, no matter about bugs, features or whatever. We agreed in #9957.
@smlng do you see your comments addressed? Then, please dismiss your change request!

PeterKietzmann:
@smlng thanks. Would you mind setting the "Reviewed:.." labels for what you did? I'll run tests a bit later

PeterKietzmann:
I've tested on a nucleo-l073rf + mrf24j40 (node A) and between that board and a samr21-xpro (node B).

1. 
```
ping6 1000 <ll node B> 1000 0 0`
1000 packets transmitted, 286 received, 72% packet loss, time 804.0634942 s
```
-> bad but 'know' issue

2.
```
ifconfig 7 ack_req (on both)
ping6 1000 <ll node B> 1000 0 0`
1000 packets transmitted, 998 received, 1% packet loss, time 159.06552569 s
```
-> fine

3.
```
ifconfig 7 -autoack (on node A)
ifconfig 7 ack_req (on node B)
```
-> as expexted: B requests ACKs, A doesn't send ACKs, B retransmits

4.
```
ifconfig 7 set nid 0x12 (on both)
```
... and running 2. again -> as expected

I'm fine with this PR **but** it feels wrong to merge it, knowing that we loos default ACK_REQ. Let's wait with merging until the fix is around.

PeterKietzmann:
Could someone update me about the current status of the 802.15.4/netdev cleanups?

>  ack_req setting is not set at boot with this PR

Is this still the case?

PeterKietzmann:
>  knowing that we lose default ACK_REQ.

@bergzand is the above still the case?

bergzand:
@jia200x Does this PR still make sense with your netdev plans?

fjmolinas:
> @jia200x Does this PR still make sense with your netdev plans?

ping! @jia200x

MrKevinWeiss:
Any update, should this be rebased or is it out of date now?

jia200x:
@jia200x Does this PR still make sense with your netdev plans?

I missed the comments! Sorry

This doesn't contradict the netdev plans and I'm not aware of anyone working on porting this radio to the radio HAL. So I would say it's fine to proceed.

benpicco:
`mrf24j40` has been ported to the Radio HAL, I think we can close this now.

PR #9598, opened 2018-07-18
stm32_common: fix ram length for MCUs with non-contiguous ram regions [backport 2018.07]

### LINKED ISSUE #9534: linkerscripts for stm32 incorrect for MCUs with non-contiguous SRAM1 & SRAM2

cladmi:
# Backport of #9579

### Contribution description
On some MCUs the ram sizes specified in stm32_mem_lengths.mk are based on the sum of available ram, while it is physically not mapped to a contiguous ram region. This will make the linker place data/code in unmapped address space if enough ram is used. See #9534 for details.
This PR fixes the problem by decreasing the ram sizes to the size of the "primary" ram region.

For testing the following snippet may be helpful:

```c
extern uint32_t _sram;
extern uint32_t _eram;

int main(void)
{
    printf("ram start @0x%p\n", &_sram);
    printf("ram end   @0x%p\n", &_eram);

    printf("last value in ram @0x%p: 0x%08lX\n", &_eram - 1, *(&_eram - 1));

    return 0;
}
```

On stm32-l476rg without the fix the above code crashes:
```
ram start @0x0x20000000
ram end   @0x0x20020000

Context before hardfault:
   r0: 0x00000018
   r1: 0x20000d18
   r2: 0xa0000000
   r3: 0x12000000
  r12: 0x08002102
   lr: 0x08000cd1
   pc: 0x08000cd0
  psr: 0x21000000

FSR/FAR:
 CFSR: 0x00008200
 HFSR: 0x40000000
 DFSR: 0x00000008
 AFSR: 0x00000000
 BFAR: 0x2001fffc
Misc
EXC_RET: 0xfffffffd
In GDB:
  set $pc=0x8000cd0
  frame 0
  bt

ISR stack overflowed by at least 16 bytes.
```

with the fix it works as expected:

```
ram start @0x0x20000000
ram end   @0x0x20018000
last value in ram @0x0x20017ffc: 0x99E5B210
```

A thing for future PRs: the currently unused ram regions (i.e. SRAM2 on L4 and CCMRAM on F3) should be properly added to the linker to allow using them in applications. Also a test to verify ram sizes may be a good thing.

### Issues/PRs references
Fixes #9534

### DIFF
--- a/cpu/stm32_common/stm32_mem_lengths.mk
+++ b/cpu/stm32_common/stm32_mem_lengths.mk
@@ -235,6 +235,7 @@ else ifeq ($(STM32_TYPE), L)
       RAM_LEN = 64K
     else ifeq ($(STM32_MODEL2), 7)
       RAM_LEN = 96K
+      RAM2_LEN = 32K
     else ifeq ($(STM32_MODEL2), 5)
       RAM_LEN = 160K
     else ifeq ($(STM32_MODEL2), 9)
--- a/cpu/stm32_common/stm32_mem_lengths.mk
+++ b/cpu/stm32_common/stm32_mem_lengths.mk
@@ -120,19 +120,20 @@ ifeq ($(STM32_TYPE), F)
       endif
     else ifeq ($(STM32_MODEL), 303)
       ifneq (, $(filter $(STM32_ROMSIZE), 6 8))
-        RAM_LEN = 16K
+        RAM_LEN = 12K
         CCMRAM_LEN = 4K
       else ifeq ($(STM32_ROMSIZE), B)
         RAM_LEN = 40K
         CCMRAM_LEN = 8K
       else ifeq ($(STM32_ROMSIZE), C)
-        RAM_LEN = 48K
+        RAM_LEN = 40K
         CCMRAM_LEN = 8K
       else ifneq (, $(filter $(STM32_ROMSIZE), D E))
-        RAM_LEN = 80K
+        RAM_LEN = 64K
+        CCMRAM_LEN = 16K
       endif
     else ifeq ($(STM32_MODEL3), 4)
-      RAM_LEN = 16K
+      RAM_LEN = 12K
       CCMRAM_LEN = 4K
     else ifeq ($(STM32_MODEL), 373)
       RAM_LEN = 32K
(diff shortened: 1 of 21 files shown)


cladmi:
@MichelRottleuthner can you test the backport ?

PR #9616, opened 2018-07-20
at86rf2xx: Always set channel on device

### LINKED ISSUE #8118: at86rf2xx: netdev reset

bergzand:
### Contribution description

This removes the check if the current configured channel equals the new
channel. This check prevents the at86rf2xx channel to be configured
after a reset which causes the radio to be non-functional after a
NETOPT_STATE_RESET.

### Issues/PRs references

fixes #8118

### Testing

Branch is intentionally not based on the latest master to prevent #9577/#9606 from causing issues.

The issue that causes gnrc_netif to switch to short source MAC addresses is not resolved here.

### DIFF
--- a/drivers/at86rf2xx/at86rf2xx_getset.c
+++ b/drivers/at86rf2xx/at86rf2xx_getset.c
@@ -140,11 +140,11 @@ uint8_t at86rf2xx_get_chan(const at86rf2xx_t *dev)
 
 void at86rf2xx_set_chan(at86rf2xx_t *dev, uint8_t channel)
 {
-    if ((channel > AT86RF2XX_MAX_CHANNEL) ||
+    if ((channel > AT86RF2XX_MAX_CHANNEL)
 #if AT86RF2XX_MIN_CHANNEL /* is zero for sub-GHz */
-        (channel < AT86RF2XX_MIN_CHANNEL) ||
+       || (channel < AT86RF2XX_MIN_CHANNEL)
 #endif
-        (dev->netdev.chan == channel)) {
+        ) {
         return;
     }
 


PeterKietzmann:
This PR fixes the described issue after resetting the interface for the first time (`ifconfig 7 set state reset`). However, after resetting it a second time I can't ping anymore. Doing a button reset, everything wors fine again. I did not further investigate yet. Ideas?

bergzand:
I'd have to investigate a bit more. For the moment I think we can conclude that this is only a partial fix.

smlng:
I can confirm the issue. Tested with `samr21-xpro` and `examples/gnrc_networking`, after two `ifconfig 7 set state reset` the board does not receive anymore - *but* it can send, i.e., I see pings from the board to another node and even the replies by the other node in the sniffer, but the samr21 still reports timeouts for the pings.

Fun fact: I tested on *master*, so this PR is not the problem.

smlng:
hence, my ACK holds - now a tested one. @PeterKietzmann you wanna file an issue, as you found it.

bergzand:
I suspect that the issue fixed here is only one of the issues that causes problems after a reset. (One of) the other issues is that the flag set by NIB to use long addresses is cleared by the reset and not reapplied.

smlng:
interesting: after two resets the node does not respond to direct pings anymore but it does to pings on `ff02::1` aka multicast. I first thought it might be something messed up with the RX state or the respective interrupts, but when multicast works this cannot be the case. So it seems that the MAC address matching does is messed up?

smlng:
same issue on PhyNode (pba-d-01-kw2x) so its more general and also (very likely) not driver related but in a higher layer  (or conceptual how radio drivers interact with higher layers).

PeterKietzmann:
See #9656 and extend, if needed.

cladmi:
When migrating the last Pull Requests for the release, I noticed this `backporting-needed` PR was not backported to the release.

Due to the tests being now all run I would find it hard to backport it now.
I will note it as a known issue.

This also show that this should be integrated to one of the automated tests.

PR #8864, opened 2018-04-02
pkg/semtech-loramac: update to upstream v4.4.1 + cleanup

aabadie:
### Contribution description

This PR updates the target version of the Semtech [loramac-node](https://github.com/Lora-net/LoRaMac-node) to the latest release (4.4.1).

For simplicity, the  branch is based on #8639.

While working on it, I realized that some code in the driver was useless (rx timeout timer) so I removed it.

I tested the port against TheThingsNetwork but for the moment it doesn't as reliably as expected.

### Issues/PRs references

Based on ~#8639~

PR #9618, opened 2018-07-20
semtech-loramac: fix wrong behavior with RX windows

jia200x:
### Contribution description

Sometimes when receiving ACK (confirmed messages) with the Semtech LoRaMAC pkg, the RX_TIMEOUT event is called twice (one from [this timer](https://github.com/RIOT-OS/RIOT/blob/master/drivers/sx127x/sx127x.c#L251) and also from [this interrupt](https://github.com/RIOT-OS/RIOT/blob/master/drivers/sx127x/sx127x.c#L195)). 

The reason is because the timer is used as an escape route if the longest possible LoRaWAN packet is not received  in the whole process (around 3 seconds in EU868), and not for RX window timeout. This is not intended by the original LoRaMAC implementation (as seen [here](https://github.com/Lora-net/LoRaMac-node/blob/develop/src/mac/region/RegionEU868.h#L140)) and sometimes produces a wrong behavior in the MAC internal states. 

This PR fixes that issue. 

~Also, another bug appeared when using the proper RX Symbol Timeout mechanism. The `randr` function from `0003-adapt-utilities-functions.patch` had a wrong behavior due to int32_t casting and gave several numbers out of the given range, which caused unexpected behaviors in LoRaMAC. This is also addressed here~ (already fixed in #8864 )

### Issues/PRs references
None so far

### DIFF
--- a/pkg/semtech-loramac/contrib/semtech_loramac_radio.c
+++ b/pkg/semtech-loramac/contrib/semtech_loramac_radio.c
@@ -96,7 +96,11 @@ void SX127XSetRxConfig(RadioModems_t modem, uint32_t bandwidth,
     sx127x_set_freq_hop(&sx127x, freqHopOn);
     sx127x_set_hop_period(&sx127x, hopPeriod);
     sx127x_set_iq_invert(&sx127x, iqInverted);
-    sx127x_set_symbol_timeout(&sx127x, 2 * symbTimeout);
+
+    /* for dealing with timing issues, we set twice 
+     * the symbol timeout */
+    sx127x_set_symbol_timeout(&sx127x, symbTimeout*2);
+
     sx127x_set_rx_single(&sx127x, !rxContinuous);
 }
 


pkg/semtech-loramac/patches/0003-adapt-utilities-functions.patch
```diff
@@ -30,15 +30,15 @@ index 8861235..e339ba5 100644
...
++    (void) seed;
++};
  
  int32_t randr( int32_t min, int32_t max )
  {
 -    return ( int32_t )rand1( ) % ( max - min + 1 ) + min;
 -}
-+    return (int32_t)random_uint32() % (max - min + 1) + min;
-+};
++    return (int32_t) (random_uint32_range(0,max-min) + min);
```
aabadie: put a space after the comma and around arithmetic operator:
```c
(random_uint32_range(0, max - min) + min);
```

smlng:
@jia200x please address comments, so we can merge this - or is this outdated? Then close.

jia200x:
@smlng this PR is still valid. However, it will conflict with #8864 (they remove the timers from the driver, but this PR still fixes our current LoRaWAN integration)

dylad:
@jia200x I thought #8864 was based on this PR. This is not the case ?

jia200x:
@dylad it was not based, but as far as I understand it already removes the timeout in the driver. I will test #8864 and merge it, and then check if the issue is still present

dylad:
@jia200x Alright, let me know if #8864 doesn't solve your issue, I have some LoRa hardware to play here so I can test and merge too :)

aabadie:
> but as far as I understand it already removes the timeout in the driver

#8864 doesn't remove the timeout in the driver but only in the package adaption. I think I did all the cleanup initially but removed it when I saw this PR. I can reapplied the change in #8864 if you want. I also think that #8864 is more priority since it improves quite some things and is opened for a long time now (~ 6 months).

smlng:
@jia200x and @aabadie what's the state here, is this now outdated (and hence closable) or needs adaption?

jia200x:
@smlng I've still seen some issues regarded this. I prefer to keep it open

smlng:
and I prefer to get it merged 😉 especially if the issue is still present and this is (kind of) a fix ... which might need some adaption, due to #8864 being merged

aabadie:
> I prefer to keep it open

Do you plan to update this PR soon ? Some changes were applied in the last release but not all. Maybe it's worth having them merged ?

jia200x:
> Do you plan to update this PR soon ? Some changes were applied in the last release but not all. Maybe it's worth having them merged ?

Yes, I agree. I will try to do it during the next days

jia200x:
I confirm the original issue is not present anymore in master. The "RX Timeout" is set as expected from the Semtech LoRaMAC pkg.

[Here's a picture](https://gist.githubusercontent.com/jia200x/daac639c022d5b085e0019b63eadbdb0/raw/0456c2b865c7d01f68821ab44b49ff8d0c734712/file.png) of how the class A cycle look like on `b-l072z-lrwan1`. The first pulse is the transmission. The second and third pulses are the reception windows (RX2 is bigger than RX1, this comes from the MAC layer).

There are still some timing issues (related to #10545) which will be fixed as soon as that issue is solved.

jia200x:
I will then close this PR ;)

PR #9623, opened 2018-07-23
Makefile.include: Replace PHONY by FORCE

jcarrano:
### Contribution description

Makefile.include abuses PHONY targets. These should be used for non-file targets, but they are currently being used to force re-running of the recipe.

This PR replaces PHONY targets by FORCE and clarifies when they should be used.

### References

See the GNU Make manual:

* https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html
* https://www.gnu.org/software/make/manual/html_node/Force-Targets.html

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -323,7 +323,7 @@ include $(RIOTBASE)/drivers/Makefile.include
 $(RIOTPKG)/%/Makefile.include::
 	$(Q)"$(MAKE)" -C $(RIOTPKG)/$* Makefile.include
 
-.PHONY: $(USEPKG:%=$(RIOTPKG)/%/Makefile.include)
+$(USEPKG:%=$(RIOTPKG)/%/Makefile.include): FORCE
 -include $(USEPKG:%=$(RIOTPKG)/%/Makefile.include)
 
 # Deduplicate includes without sorting them
--- a/Makefile.include
+++ b/Makefile.include
@@ -349,9 +349,13 @@ include $(RIOTMAKE)/modules.inc.mk
 .PHONY: all link clean flash flash-only term doc debug debug-server reset objdump help info-modules
 .PHONY: print-size elffile binfile hexfile
 .PHONY: ..in-docker-container
-# Target can depend on FORCE to always rebuild but still let make use file
-# modification timestamp (contrary to .PHONY).
-# Use it for goals that may keep outputs unchanged when executed.
+# Targets that depend on FORCE will always be rebuilt. Use it for recipes that
+# may change the output even if prerequisites don't change (for example, if
+# it depends on configuration variables.)
+# For targets that are not a file, and only for that, use .PHONY. For more
+# information, see:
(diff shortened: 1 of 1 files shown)


Makefile.include
```diff
@@ -349,9 +349,13 @@ include $(RIOTMAKE)/modules.inc.mk
...
-# Target can depend on FORCE to always rebuild but still let make use file
-# modification timestamp (contrary to .PHONY).
-# Use it for goals that may keep outputs unchanged when executed.
+# Targets that depend on FORCE will always be rebuilt. Use it for recipes that
+# may change the output even if prerequisites don't change (for example, if
+# it depends on configuration variables.)
+# For targets that are not a file, and only for that, use .PHONY. For more
+# information, see:
+# https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html
+# https://www.gnu.org/software/make/manual/html_node/Force-Targets.html
```
cladmi: The documentation does not unfortunately say that `FORCE` should be use for file targets. I already found that when adding `FORCE` in this file: https://github.com/RIOT-OS/RIOT/pull/8844

Even if it was not that great, my previous documentation was mentioning the justification with the file timestamp being taken into account with FORCE contrary to .PHONY.
I find it important because it is not clear in the documentation and it explain why you should use `FORCE`.

cladmi: To be more precise on my thought, you changed to "how it should be used", I was more writing the "why it should be used", and unfortunately the official documentation does not say why. And I think both should be kept so I would prefer if you had a merge of both.

cladmi: > If the file is updated each time the rule runs, it should not change anything. What FORCE gives us is the possibility of choosing not to update the file and then it will not trigger an update in it's dependents.

That what I explained with the previous documentation. So just keep it I guess and maybe clarify it.

No need to explain everything, keep it simple. Adding your part with "file targets -> FORCE" and "non file -> .PHONY" would be enough.

Makefile.include
```diff
@@ -349,11 +349,17 @@ include $(RIOTMAKE)/modules.inc.mk
...
-# it depends on configuration variables.)
-# For targets that are not a file, and only for that, use .PHONY. For more
-# information, see:
+
+# Targets that depend on FORCE will always be rebuilt. Contrary to a .PHONY
+# target, they are considered real files and the modification timestamp is taken
+# into account.
+#
+# FORCE is useful for goals that may keep outputs unchanged (for example, if it
+# depends on environment or configuration variables). If the goal were PHONY, it
```
cladmi: Nitpick, replace `PHONY` by `.PHONY` but otherwise I like it this way.

Makefile.include
jcarrano: > Some other considerations: you cannot make a pattern rule phony, but you can make it FORCE.

I correct myself, what the manual says is "The implicit rule search (see Implicit Rules) is skipped for .PHONY targets."

Issue #9460, opened 2018-06-29
docker: make commands for docker require insecure setup.


jcarrano:
#### Description

Communicating withe the docker daemon requires elevated privileges. This can be accomplished by either using `sudo` or adding a user to the `docker` group. The latter option is advocated by the [RIOT wiki](https://github.com/RIOT-OS/RIOT/wiki/Use-Docker-to-build-RIOT), and is expected by the build system:

> Depending on your operating system, you might need to add yourself to the docker group:

Using the `docker` group, however, is not as secure as it seems, and in effectively the same as granting the user root permissions (without password). See:

* Docker manual: [Manage Docker as a non-root user](https://docs.docker.com/install/linux/linux-postinstall/), [Docker daemon attack surface](https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface)
* This [blog post](https://www.projectatomic.io/blog/2015/08/why-we-dont-let-non-root-users-run-docker-in-centos-fedora-or-rhel/)
* This [other one](https://fosterelli.co/privilege-escalation-via-docker.html)

#### Steps to reproduce the issue

Try to follow the instructions at https://github.com/RIOT-OS/RIOT/wiki/Use-Docker-to-build-RIOT with a user that is not in the 'docker' group, or on a system where there is no such group.

#### Expected results

`make BUILD_IN_DOCKER=1` should work anyways.

#### Actual results

`make BUILD_IN_DOCKER=1` fails with insufficient permissions.

cladmi:
As you raised the issue in the first place, do you mind updating the documentation now that https://github.com/RIOT-OS/RIOT/pull/9635 is merged ?

jcarrano:
Fixed, closing.

PR #9635, opened 2018-07-25
makefiles/docker.inc.mk: allow using sudo for docker

cladmi:
When testing `llvm` support I often try running with murdock `docker` image to get the same environment but I do not have my user in the `docker` group.
This add a solution for this.

### Contribution description

~Add `DOCKER_WITH_SUDO` variable that runs docker using `sudo` when set to 1.~
Allow overwriting `DOCKER` command with for example `DOCKER="sudo docker"`.

Not all users register their user in the docker group and should not run the
whole make process as root to use docker.
Creating files as a user is correctly handled as `id -u` is still run as the
original user.

The variable name was chosen in 5 seconds and can be changed.

### Testing

```
make -C examples/hello-world BUILD_IN_DOCKER=1 DOCKER="sudo docker"
make: Entering directory '/home/harter/work/git/RIOT/examples/hello-world'
sudo docker run --rm -t -u "$(id -u)" \
...

make -C examples/hello-world/ BUILD_IN_DOCKER=1
make: Entering directory '/home/harter/work/git/RIOT/examples/hello-world'
docker run --rm -t -u "$(id -u)" \
...
```

### Issues/PRs references

https://github.com/RIOT-OS/RIOT/issues/9460
Testing llvm with docker https://github.com/RIOT-OS/RIOT/pull/9398/

### DIFF
--- a/makefiles/docker.inc.mk
+++ b/makefiles/docker.inc.mk
@@ -79,6 +79,9 @@ DOCKER_OVERRIDE_CMDLINE := $(foreach varname,$(DOCKER_ENV_VARS), \
     ))
 DOCKER_OVERRIDE_CMDLINE := $(strip $(DOCKER_OVERRIDE_CMDLINE))
 
+# Overwrite if you want to use `docker` with sudo
+DOCKER ?= docker
+
 # This will execute `make $(DOCKER_MAKECMDGOALS)` inside a Docker container.
 # We do not push the regular $(MAKECMDGOALS) to the container's make command in
 # order to only perform building inside the container and defer executing any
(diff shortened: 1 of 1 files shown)


cladmi:
The documentation would need to be updated accordingly after this is merged of course.

jcarrano:
Could we instead test if the user is in the `docker` group?

jcarrano:
Or maybe check the permissions at `/var/run/docker.sock` or try to run `docker ps`.

cladmi:
If you can test this on mac maybe.

cladmi:
I made the change simpler not introducing a special option, just exposing a `DOCKER` variable that can be overwritten.

Doing an auto-detection would be another PR that I am not really interested in doing.

kaspar030:
please no auto-sudo.

jcarrano:
@cladmi Murdock is whining about "Commit message is longer than 50 characters"

cladmi:
There is a 50 characters "warning" but the hard limit is 72 or something.
I will squash and murdock should be green.

PR #9639, opened 2018-07-26
Move testing infrastructure files to dist/tests

yegorich:
Create if_lib package containing all the modules and
adapt the *.py files to import each other using the intra-package
references.

The idea behind a package is to invoke test.py either by permanently
modifying PYTHONPATH in user profile via adding path to dist/tests or
make temporary PYTHONPATH changes during the invocation:

PYTHONPATH=$PYTHONPATH:$RIOTBASEdist/tests python3 test.py

Leave periph_i2c_if.py in the same folder as test.py as this file is
just a Python wrapper around periph specific main.c.

Signed-off-by: Yegor Yefremov <yegorslists@googlemail.com>


yegorich:
This PR is an RFC for reusing the BPT interface for other peripherals like UART, SPI etc. The main design issue here is how to best manage the importing of BPT interface files. This approach suggests modifying PYTHONPATH env variable instead of making changes to sys.path inside the test.py itself.

yegorich:
@MrKevinWeiss I'm for `test_if`. @smlng what about you?

MrKevinWeiss:
So after talking with @smlng we would like the .py files to be in dists/test/if_lib/* not dist/tools/testsimulatorlib/* Also add a readme or manifest to quickly explain what the package is doing (a one liner is fine for now).  Since the I2C will be merged soon, just a rebase/squash to master should be fine.

MrKevinWeiss:
ACK!

PR #9647, opened 2018-07-28
Mango Virtual Machine as package with example

janmoehl:
### Contribution description

Adds the [Mango Virtual Machine](https://github.com/ektrah/mango) as package. Contains also a small example, on how to use it.

### Issues/PRs references
No References


jcarrano:
Hello @janmoehl and welcome to RIOT.

While the VM looks quite capable, I'm concerned that the mango interpreter may be somewhat immature, at least in terms of documentation. That being said, I'm very interested in running interpreted languages on top of RIOT. I want to make sure, though, that it is useful for a broader audience.

Another issue is that the mango assembler is needed, so it should be added as a tool (`dist/tools`). Seeing that it is written in Haskell, I think we may have dependency issues - Haskell support varies across platforms. We could require the user to have the `stack` tool installed, that would save some trouble.

Is there any high level frontend for mango? If not, is it planned?

janmoehl:
Thank you for the friendly welcome.

Yes, the mango interpreter is as far as I do not know in productive use or strongly tested.

The use of `Stack` could save a lot of pain. I'll see if I get the compiler to work with `Stack` in the next few days, but I'm not that familiar with Haskell.

The creator of Mango is not planning a high level frontend at the moment. But the mango bytecode is heavily inspired by common intermediate language (CIL). Many operations are not available, e.g. object orientation. But it is possible to translate the basic instructions of the CIL bytecode almost directly into the mango bytecode. I wrote a rudimentary prototype for that, which I have not published yet. It lacks testing, but I was able to write small programs in C #, use the C # compiler to get the CIL bytecode, and compile that into mango bytecode.

jcarrano:
It would be interesting to see the performance and safeness of the VM. I'm working on Lua support and one of my objectives is to have a proper sandboxed interpreter. The problem with Lua is that the bytecode is not 100% portable and cannot live in flash, so maybe Mango could fill that gap.

A feature of mango I liked is that for "system calls" the interpreter returns control to its caller instead of nesting C calls.

> The creator of Mango is not planning a high level frontend at the moment. 

I believe there are valid use cases for a bytecode-only interpreter, where the compilation is the responsibility of the user. In other words, RIOT itself could include the VM only and leave the choice of frontend to the user/developer.

PR #6178, opened 2016-12-06
can stm32: add a driver for STM32 bxCAN peripheral 

vincent-d:
This driver is compliant with the candev interface. It has been tested
with STM32F0 and STM32F2 only at this time but should be compliant with
other STM32Fx devices.

This is based on #5793.

PR #9655, opened 2018-07-31
boards: Add support for KEA128LEDLIGHTRD

OYTIS:
The PR includes support for S9KEA128 MCU with a limited set of peripherals (LEDs, GPIOs, PIT timer and CAN) and support for the board.
I had to add some conditional compilation to kinetis codebase, because KEA family has significant differences from what was supported so far (different clock generation/distribution, different pin multiplexing).
I've also run into some issues with CAN subsystem and added some hacks to my driver as a workaround, but it's worth debugging these problems to the cause.

### DIFF
--- a/boards/kea128ledlightrd/Makefile
+++ b/boards/kea128ledlightrd/Makefile
@@ -1,0 +1,3 @@
+MODULE = board
+
+include $(RIOTBASE)/Makefile.base
--- a/boards/kea128ledlightrd/Makefile.dep
+++ b/boards/kea128ledlightrd/Makefile.dep
@@ -1,0 +1,6 @@
+USEMODULE += stdio_rtt
+
+# setup clock to run at 40MHz
+CFLAGS += -DCLOCK_SETUP=1
+
+include $(RIOTCPU)/kinetis/Makefile.dep
--- a/boards/kea128ledlightrd/Makefile.features
+++ b/boards/kea128ledlightrd/Makefile.features
@@ -1,0 +1,7 @@
+# Put defined MCU peripherals here (in alphabetical order)
+FEATURES_PROVIDED += periph_timer
+
+# The board MPU family (used for grouping by the CI system)
+FEATURES_MCU_GROUP = cortex_m0+
+
+include $(RIOTCPU)/kinetis/Makefile.features
--- a/boards/kea128ledlightrd/Makefile.include
+++ b/boards/kea128ledlightrd/Makefile.include
@@ -1,0 +1,8 @@
+# define the cpu used by the KEA128LEDLIGHTRD board
+export CPU = kinetis
+export CPU_MODEL = s9keaz128aclh48
+
+DEBUG_ADAPTER ?= arm-jtag-swd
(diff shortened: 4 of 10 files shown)


PeterKietzmann:
I guess it is [this board](https://www.mouser.de/datasheet/2/302/KEA128LEDLIGHTQSG-782665.pdf)?  Nice! @smlng might be interested as well.

smlng:
looks really nice, @PeterKietzmann can you order the board 😀 For the CPU (kinetis) I'd like to have @gebart opinions, too.

smlng:
Just from a quick view on this, it might be best to sort and separate this out a bit. I see 3 major contributions in this PR:

1. add KEA CPU support
2. add KEAxyz board support
3. add NXP/Kinetis CAN periph driver

which I would pack into distinct PRs based on one-another in that order. Though the first cannot be used without the second, it still would ensure that the changes can be tested with existing kinetis CPUs. For the CAN driver I would suggest to rename it a bit and put it into `periph/can.c` see #6178, too.

OYTIS:
Sure, I can break the PR into three (or four with Olimex debugger support) if you find it more convenient for review. I'm not sure how to deal with dependencies though, the PR for the board will apparently not compile until MCU support is merged.
Note that CAN driver is KEA-specific, K series uses a different IP core (FlexCAN vs MSCAN).

OYTIS:
By the way, what's the purpose of `typedef int dont_be_pedantic` declarations?

smlng:
@OYTIS its more about the `#ifdef #else` where the file would be empty in some cases (ie. the ifdef does not hold) and with certain compiler flags we would get errors then. So the compiler should not be pedantic here 😁

OYTIS:
@smlng Thanks! The first PR is [there](https://github.com/RIOT-OS/RIOT/pull/9666), I'll move `mscan.c` under `periph` later and make a second one.

OYTIS:
@smlng Split this PR into three. Previous two can be found at https://github.com/RIOT-OS/RIOT/pull/9666 and https://github.com/RIOT-OS/RIOT/pull/9711. MSCAN is now under `periph/`.

smlng:
@OYTIS thanks!

smlng:
requires #9666

smlng:
@OYTIS this seems to include #9711, too. Can you rebase on master and remove the MSCAN dependency here. Then this can move independently.

OYTIS:
@smlng Sure. Then MSCAN support on KEA128LEDLIGHTRD should come as a separate PR, right?

smlng:
yes, what I mean is: leave #9711 as is. But remove any MSCAN related config and code stuff from this PR. This might require a followup PR to enable the periph_can feature for this board later.

OYTIS:
@smlng Done. Also fixed a bug in https://github.com/RIOT-OS/RIOT/pull/9711.

OYTIS:
@smlng @gebart Oh this one. To be honest, I don't have access to the KEA128LEDLIGHTRD board any more. So unless someone wants to take over, or test and merge, that can only be closed I suppose.

PR #9666, opened 2018-08-01
kinetis: add support for S9KEAZ128

OYTIS:
Add supports for Kinetis EA series and S9KEAZ129 MCU.
This series is significantly different from K, for instance it uses different clock generation/distribution module and different pin multiplexing scheme. One might think it starts getting too far with conditional compilation already, maybe it's worth refactoring cpu/kinetis at some point to be more modular.

### DIFF
--- a/cpu/kinetis/Makefile.features
+++ b/cpu/kinetis/Makefile.features
@@ -2,6 +2,10 @@ FEATURES_PROVIDED += periph_cpuid
 FEATURES_PROVIDED += periph_hwrng
 FEATURES_PROVIDED += periph_gpio
 FEATURES_PROVIDED += periph_gpio_irq
+ifeq (EA,$(KINETIS_SERIES))
+FEATURES_PROVIDED += periph_ics
+else
 FEATURES_PROVIDED += periph_mcg
+endif
 
 include $(RIOTCPU)/cortexm_common/Makefile.features
--- a/cpu/kinetis/include/periph_cpu.h
+++ b/cpu/kinetis/include/periph_cpu.h
@@ -30,6 +30,26 @@
 extern "C" {
 #endif
 
+#ifdef PORT_PCR_MUX
+#  define KINETIS_HAVE_PCR
+#endif
+
+#ifdef SIM_PINSEL_REG
+#  define KINETIS_HAVE_PINSEL
+#endif
+
+#ifdef ADC_CFG1_MODE_MASK
+#  define KINETIS_HAVE_ADC_K
+#endif
+
+#ifdef SPI_CTAR_CPHA_MASK
+#  define KINETIS_HAVE_MK_SPI
+#endif
+
+#ifdef LPTMR_CSR_TEN_MASK
+#  define KINETIS_HAVE_LPTMR
+#endif
+
 /**
  * @name    CPU specific gpio_t type definition
  * @{
@@ -145,6 +165,7 @@ typedef enum {
 /** @} */
 #endif /* ndef DOXYGEN */
 
+#ifdef KINETIS_HAVE_PCR
 /**
  * @brief   PORT control register bitmasks
  *
@@ -165,18 +186,21 @@ typedef enum {
     GPIO_PCR_PD    = (PORT_PCR_PE_MASK),    /**< enable pull-down */
     GPIO_PCR_PU    = (PORT_PCR_PE_MASK | PORT_PCR_PS_MASK)  /**< enable PU */
 } gpio_pcr_t;
+#endif /* KINETIS_HAVE_PCR */
 
 #ifndef DOXYGEN
 /**
  * @name    GPIO flank configuration values
  * @{
  */
+#ifdef KINETIS_HAVE_PCR
 #define HAVE_GPIO_FLANK_T
 typedef enum {
     GPIO_RISING  = PORT_PCR_IRQC(0x9),  /**< emit interrupt on rising flank */
     GPIO_FALLING = PORT_PCR_IRQC(0xa),  /**< emit interrupt on falling flank */
     GPIO_BOTH    = PORT_PCR_IRQC(0xb),  /**< emit interrupt on both flanks */
 } gpio_flank_t;
+#endif /* KINETIS_HAVE_PCR */
 /** @} */
 #endif /* ndef DOXYGEN */
 
@@ -202,6 +226,7 @@ enum {
  * @{
  */
 #define HAVE_ADC_RES_T
+#ifdef KINETIS_HAVE_ADC_K
 typedef enum {
     ADC_RES_6BIT  = (0xfe),             /**< not supported */
     ADC_RES_8BIT  = ADC_CFG1_MODE(0),   /**< ADC resolution: 8 bit */
@@ -210,6 +235,7 @@ typedef enum {
     ADC_RES_14BIT = (0xff),             /**< ADC resolution: 14 bit */
     ADC_RES_16BIT = ADC_CFG1_MODE(3)    /**< ADC resolution: 16 bit */
 } adc_res_t;
+#endif /* KINETIS_HAVE_ADC_K */
 /** @} */
 
 #if defined(FTM_CnSC_MSB_MASK)
@@ -260,6 +286,8 @@ typedef enum {
  * @name    SPI mode bitmasks
  * @{
  */
+
+#ifdef KINETIS_HAVE_MK_SPI
 #define HAVE_SPI_MODE_T
 typedef enum {
     SPI_MODE_0 = 0,                                         /**< CPOL=0, CPHA=0 */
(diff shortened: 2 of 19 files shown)


cpu/kinetis/periph/ics.c
```diff
@@ -0,0 +1,197 @@
...
+ * @date 2015-07-24
+ * @brief Device specific configuration file for SKEAZ1284 (implementation file)
+ *
+ * Provides a system configuration function and a global variable that contains
+ * the system frequency. It configures the device and initializes the oscillator
+ * (PLL) that is part of the microcontroller device.
+ */
+
+#include <stdint.h>
+#include "cpu.h"
```
smlng: This file is taken from the reference library (or similar), right? You just added this `#include`, we may need to clarify and document this documentation, so if someone wants to update this file with a future release from the ref-lib, they know what to adapt for RIOT.

cpu/kinetis/periph/timer.c
```diff
@@ -184,6 +201,7 @@ static inline int pit_set(uint8_t dev, uint32_t timeout)
     const uint8_t ch = pit_config[dev].count_ch;
     /* Disable IRQs to minimize the number of lost ticks */
     unsigned int mask = irq_disable();
+
```
smlng: unrelated style changes

cpu/kinetis/periph/timer.c
```diff
@@ -306,13 +331,15 @@ static inline void _lptmr_set_cb_config(uint8_t dev, timer_cb_t cb, void *arg)
 /**
  * @brief  Compute the LPTMR prescaler setting, see reference manual for details
  */
-static inline int32_t _lptmr_compute_prescaler(uint8_t dev, uint32_t freq) {
+static inline int32_t _lptmr_compute_prescaler(uint8_t dev, uint32_t freq)
+{
```
smlng: on the other hand this one is valid, and according to coding conventions 😄

cpu/kinetis/vectors.c
```diff
@@ -187,31 +187,70 @@ ISR_VECTOR(1) const isr_t vector_cpu[CPU_IRQ_NUMOF] = {
     [I2S0_Rx_IRQn    ] = isr_i2s0_rx,         /* I2S0 receive interrupt */
 #endif
 #ifdef UART0
-#ifdef UART_RPL_RPL_MASK
+#  ifdef KINETIS_SINGLE_UART_IRQ
```
smlng: we might need to discuss and clarify (in coding conventions) how to indent pre-processor `#if`s

cpu/kinetis/periph/gpio.c
```diff
@@ -226,15 +248,17 @@ int gpio_init_int(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
...
+#ifdef KINETIS_HAVE_PCR
 void gpio_init_port(gpio_t pin, uint32_t pcr)
 {
     /* enable PORT clock in case it was not active before */
     clk_en(pin);
 
     /* if the given interrupt was previously configured as interrupt source, we
-     * need to free its interrupt context. We to this only after we
-     * re-configured the pin in case an event is happening just in between... */
+    * need to free its interrupt context. We to this only after we
```
smlng: unrelated change and (now) wrong indention, ie. off by one space

cpu/kinetis/include/vectors_kinetis.h
```diff
@@ -49,6 +49,8 @@ void isr_can1_ored_message_buffer(void); /**< CAN1 OR'd message buffers interrup
 void isr_can1_rx_warning(void); /**< CAN1 Rx warning interrupt handler */
 void isr_can1_tx_warning(void); /**< CAN1 Tx warning interrupt handler */
 void isr_can1_wake_up(void); /**< CAN1 wake up interrupt handler */
+void isr_mscan_rx(void); /**< MSCAN RX interrupt handler */
```
smlng: indention of doc-comment

cpu/kinetis/include/periph_cpu.h
```diff
@@ -202,6 +226,7 @@ enum {
  * @{
  */
 #define HAVE_ADC_RES_T
+#ifdef KINETIS_HAVE_ADC_16BIT
```
smlng: why?

OYTIS: Ow, sorry, that might be related to changes to adc, which I didn't test and decided not to include into this PR. I'll remove that.

cpu/kinetis/vectors.c
OYTIS: What would be your preferred indentation. Most of the code doesn't indent nested `#if`s, but sometimes it gets too hard to read.

cpu/kinetis/periph/ics.c
OYTIS: Yes, I've just taken this from vendor's examples. Probably we should rewrite it at some point, but I just can't see an immediate value of such work.
Can it be documented in-line in `ics.c` or is there some dedicated place for that?

cpu/kinetis/periph/timer.c
OYTIS: I've run the code through uncrustify at some point, which annoyingly doesn't always leave already committed code unchanged. I'll proof-read again.

cpu/kinetis/include/periph_cpu.h
OYTIS: @smlng Actually this `#ifdef` still makes sense. `ADC_CFG1_MODE` etc. are not defined for all CPUs. What's important here, it's not defined for S9KEA.

cpu/kinetis/periph/ics.c
OYTIS: @smlng  Added some documentation to ics.c

OYTIS:
@smlng @gebart Fixed issues and rebased.

cpu/kinetis/Makefile.features
```diff
@@ -2,6 +2,11 @@ FEATURES_PROVIDED += periph_cpuid
 FEATURES_PROVIDED += periph_hwrng
 FEATURES_PROVIDED += periph_gpio
 FEATURES_PROVIDED += periph_gpio_irq
+ifneq (EA,$(KINETIS_SERIES))
```
smlng: can you please explain why the features `mcg` and `ics` are exclusive for the `EA` series? I mean `ics` is new, but you change the usage of `mcg` here.

cpu/kinetis/include/periph_cpu.h
smlng: mhm, the name of the macro suggests that the `ifdef` should only change the definition of `ADC_RES_16BIT` and not including or excluding the whole ADC config.

OYTIS: @smlng It _might_ be a misnomer, yes. The point is that KEA ADC (not in scope of this PR) has a different set of registers than the one already implemented here. Since they don't have any special names and only differ in resolution, I've called the defines `KINETIS_HAVE_ADC_16BIT` and `KINETIS_HAVE_ADC_12BIT` (the latter is, again, not included here). I could rename it to something more meaningful (like e.g. `KINETIS_HAVE_KEA_ADC` and `KINETIS_HAVE_K_ADC`).

OYTIS:
@smlng @gebart 
Fixed unmatched endif, rebased and renamed `HAVE_ADC_16BIT` to `HAVE_ADC_K` (as in Kinetis K series).

cpu/kinetis/include/periph_cpu.h
```diff
@@ -268,6 +296,7 @@ typedef enum {
     SPI_MODE_3 = (SPI_CTAR_CPOL_MASK | SPI_CTAR_CPHA_MASK)  /**< CPOL=1, CPHA=1 */
 } spi_mode_t;
 /** @} */
+#endif // KINETIS_HAVE_MK_SPI
```
smlng: please use C comments only, i.e. `/* ... */`

cpu/kinetis/include/periph_cpu.h
```diff
@@ -165,18 +186,21 @@ typedef enum {
     GPIO_PCR_PD    = (PORT_PCR_PE_MASK),    /**< enable pull-down */
     GPIO_PCR_PU    = (PORT_PCR_PE_MASK | PORT_PCR_PS_MASK)  /**< enable PU */
 } gpio_pcr_t;
+#endif //KINETIS_HAVE_PCR
```
smlng: please change to `/* ... */` here and below, I counted 8 "wrong" comment over all files

smlng:
CI is not happy, and for me it also fails with an error in periph/timer when compiling for PhyNode (pba-d-01-kw2x)

smlng:
ping @OYTIS, this needs to be in first, before we can get the KEA boards merged.

OYTIS:
@smlng 
Sorry for the delay. All the failures have been hunted, I believe it's good now.

OYTIS:
@smlng @gebart 
Anything else needed before merge? Or should the two dependent PRs (https://github.com/RIOT-OS/RIOT/pull/9711, https://github.com/RIOT-OS/RIOT/pull/9655) be reviewed first?

smlng:
I successfully tested with the PhyNode, just to make sure it doesn't break stuff for existing nodes. To me this looks good to go, so ACK & GO!

OYTIS:
@smlng Thanks a lot!

I've rebased the next two PRs to make them more observable.

PR #9691, opened 2018-08-03
cpu/cc2538: add TI vendor headers

smlng:
Currently the cc2538 is based on from-scratch adaption which is
    not feature complete and thus lacks defines etc. Introducing the
    official vendor header will ease future extension and adaptions
    of the CPU and its features.

### Contribution description

This PR adds official TI vendor headers for the CC2538. In follow-up PRs I'll adapt existing
implementation to use vendor header defines where ever useful. Hence, this only a first step only adding files not touching existing code.


PeterKietzmann:
ACK in general. Are these CC2538-only headers? I didn't find an indicator in the headers on first sight.

smlng:
you're right no mentioning in the header doc or license what they are good for, but they are taken from the official `cc2538_foundation_firmware_1_0_1_0` - which one can download from TI.

PR #9698, opened 2018-08-03
cpu/cc2538: refine periph/spi implementation

smlng:
### Contribution description

Adapt the periph SPI implementation to make use of vendor header defines and also utilise proper
GPIO API functions instead of manually configure through registers.

### DIFF
--- a/boards/remote-pa/include/periph_conf.h
+++ b/boards/remote-pa/include/periph_conf.h
@@ -51,17 +51,17 @@ static const i2c_conf_t i2c_config[] = {
  */
 static const spi_conf_t spi_config[] = {
     {
-        .dev      = SSI0,
-        .mosi_pin = GPIO_PD0,
-        .miso_pin = GPIO_PC4,
-        .sck_pin  = GPIO_PD1,
-        .cs_pin   = GPIO_PD3
+        .num      = 0,
+        .mosi_pin = GPIO_PIN(3, 0),
+        .miso_pin = GPIO_PIN(2, 4),
+        .sck_pin  = GPIO_PIN(3, 1),
+        .cs_pin   = GPIO_PIN(3, 3)
     },
     {
-        .dev      = SSI1,
-        .mosi_pin = GPIO_PC7,
-        .miso_pin = GPIO_PA4,
-        .sck_pin  = GPIO_PB5,
+        .num      = 1,
+        .mosi_pin = GPIO_PIN(2, 7),
+        .miso_pin = GPIO_PIN(0, 4),
+        .sck_pin  = GPIO_PIN(1 ,5),
         .cs_pin   = GPIO_UNDEF
     }
 };
--- a/cpu/cc2538/include/periph_cpu.h
+++ b/cpu/cc2538/include/periph_cpu.h
@@ -25,6 +25,7 @@
 #include <stdio.h>
 
 #include "cpu.h"
+#include "vendor/hw_ssi.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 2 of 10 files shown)


boards/remote-pa/include/periph_conf.h
```diff
@@ -66,17 +66,17 @@ static const spi_clk_conf_t spi_clk_config[] = {
...
     },
     {
-        .dev      = SSI1,
-        .mosi_pin = GPIO_PC7,
-        .miso_pin = GPIO_PA4,
-        .sck_pin  = GPIO_PB5,
+        .num      = 1,
+        .mosi_pin = GPIO_PIN(2, 7),
+        .miso_pin = GPIO_PIN(0, 4),
+        .sck_pin  = GPIO_PIN(2 ,5),
```
PeterKietzmann: `GPIO_PB5` -> `GPIO_PIN(1, 5)`

cpu/cc2538/include/periph_cpu.h
```diff
@@ -200,7 +205,7 @@ typedef struct {
  * @{
  */
 typedef struct {
-    cc2538_ssi_t *dev;      /**< SSI device */
+    uint8_t num;            /**< SSI device */
```
PeterKietzmann: The semantics of this parameter changed, but the documentation didn't. Can you extend the comment? For example, mention that this is the device **number** 1...x or so

cpu/cc2538/periph/spi.c
```diff
@@ -69,33 +73,16 @@ void spi_init(spi_t bus)
...
-
-    IOC_PXX_OVER[spi_config[bus].mosi_pin] = IOC_OVERRIDE_OE;
-    IOC_PXX_OVER[spi_config[bus].miso_pin] = IOC_OVERRIDE_DIS;
-    IOC_PXX_OVER[spi_config[bus].sck_pin ] = IOC_OVERRIDE_OE;
-    IOC_PXX_OVER[spi_config[bus].cs_pin  ] = IOC_OVERRIDE_OE;
-
-    gpio_hardware_control(spi_config[bus].mosi_pin);
-    gpio_hardware_control(spi_config[bus].miso_pin);
-    gpio_hardware_control(spi_config[bus].sck_pin);
-    gpio_hardware_control(spi_config[bus].cs_pin);
```
PeterKietzmann: `gpio_hardware_control` isn't used anywhere else. How about we remove its definition? Or do you plan to cleanup cc2538/gpio anyway?

cpu/cc2538/periph/spi.c
```diff
@@ -26,40 +26,44 @@
...
+
 /**
  * @brief   Array holding one pre-initialized mutex for each SPI device
  */
 static mutex_t locks[SPI_NUMOF];
 
 static inline cc2538_ssi_t *dev(spi_t bus)
 {
-    return spi_config[bus].dev;
+    return (cc2538_ssi_t *)(SSI0_BASE + (spi_config[bus].num * 0x1000));
```
PeterKietzmann: This function is called in every `acquire`, `transfer`, `release` function. Do you have any estimation how this change effects performance?

boards/remote-pa/include/periph_conf.h
smlng: oops, will fix

cpu/cc2538/periph/spi.c
smlng: then it will be constant lookup just as before.

smlng:
Hi @PeterKietzmann thx for the quick feedback and testing:

* I like the `.num` device index better than actually providing the hardware address in `periph_conf.h` because the developer typically cares about which device is enabled and not its address. The `.num` is required if you want to use SPI1 only, that is not using/configuring SPI0. Contrary to the timer implementation this is way simpler here because there are no IRQ handlers that need to be matched to the SPI device used.
* yes the CS stuff I saw, too - but to me that's definitely a separate PR, also to keep this one simple.
* clock settings: I'll look into, this should be fixed here.

smlng:
@PeterKietzmann are we good here? As said clock issue and CS hwpins will be extra PRs

PeterKietzmann:
The new openmote needs adaption as well

smlng:
done

PeterKietzmann:
please squash

PR #9701, opened 2018-08-03
Fmt: add fmt hex byte

toonst:
### Contribution description

Create fmt function for converting a single hex string byte (two characters, eg. 5E) to an actual byte.
This diverges a bit from the other functions of fmt in that it does not return a size_t, but to me this seems justified based on the use case of this API function.

### DIFF
--- a/sys/fmt/fmt.c
+++ b/sys/fmt/fmt.c
@@ -142,7 +142,6 @@ size_t fmt_hex_bytes(uint8_t *out, const char *hex)
     size_t len = fmt_strlen(hex);
 
     if (len & 1) {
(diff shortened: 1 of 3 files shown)


aabadie:
@miri64, I don't think this is an API change since it just extends the existing one. I would rather use the 'new feature' label.
Otherwise, I'm ok with this addition. @toonst, can you add an unit test for this function. Just extend the [existing unit tests of the fmt module](https://github.com/RIOT-OS/RIOT/tree/master/tests/unittests/tests-fmt).

toonst:
added unit tests

PR #9711, opened 2018-08-05
Add support for MSCAN module

OYTIS:
Add support for MSCAN module (KEA-specific).

Based on https://github.com/RIOT-OS/RIOT/pull/9666


OYTIS:
@smlng Same as here: https://github.com/RIOT-OS/RIOT/pull/9655, don't have the board at hand any more.

PR #9716, opened 2018-08-06
boards/b-l475e-iot01a: configure second I2C and on-board sensors (lis3mdl, lsm6dsl and hts221)

aabadie:
### Contribution description

Now that the I2C refactoring is in and since it provides I2C support for STM32 L4, this PR configures a second I2C peripheral and defines default parameters for the on-board HTS221, LIS3MDL and LSM6DSL sensors of the ST B-L475E-IOT01A board.

### Issues/PRs references

#7585


aabadie:
I think @kYc0o has this board, so he might be able to test.

PR #9723, opened 2018-08-07
boards/nucleo-f070: add missing periph_i2c feature

aabadie:
### Contribution description

This PR reverts commit f16272506c59310e0256e237ba39adfcf50ed1c5 that removed the I2C feature from nucleo-f070rb during the I2C rework.

The I2C configuration was still there (and still works) so it's just a matter of re-adding it to the build system.

Instead of manually creating the commit, I use `git revert <hash>` to keep a reference of the original commit.

### Issues/PRs references

Introduced in I2C rework #6577

### DIFF
--- a/boards/nucleo-f070rb/Makefile.features
+++ b/boards/nucleo-f070rb/Makefile.features
@@ -1,6 +1,7 @@
 # Put defined MCU peripherals here (in alphabetical order)
 FEATURES_PROVIDED += periph_adc
 FEATURES_PROVIDED += periph_gpio
+FEATURES_PROVIDED += periph_i2c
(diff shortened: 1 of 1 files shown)


aabadie:
Interesting, in fact f162725 was modifying the nucleo-f030r8 whereas the commit comment was mentioning nucleo-f070rb... I removed the revert commit because the initial commit was correct. So this PR is now just about adding the missing `periph_i2c` feature in the build system (I2C is already configured for this board).

aabadie:
Thanks @smlng, let's merge this. At least, there's one STM32F0 configured for checking I2C driver implementation.

PR #9726, opened 2018-08-07
pkg/lua: Make the module searchers conform to the API.

jcarrano:
#### Bug 1

I messed up the specification for the Lua loader module.

The module searchers in the `require` package should return a string
if the module is not found, and not raise an error.

See: https://www.lua.org/manual/5.3/manual.html#pdf-package.searchers

If the searchers raised an error, then the search would stop after the first loader fails to find (which is the current- buggy- behaviour.) The result is that only the pure-lua module table get searched, and it is impossible to load a C extension module.

#### Bug 2

I violated my own specification, resulting in a crash where a module was requested but no modules were supplied by the user.

This used to work, but then some time ago I remove the definition of the weak symbols used for the tables, leaving them pointing to NULL by default.

### Testing procedure

I included a test script in the third commit.

### DIFF
--- a/tests/lua_loader/main.c
+++ b/tests/lua_loader/main.c
@@ -1,0 +1,105 @@
+/*
+ * Copyright (C) 2019 Freie Universität Berlin.
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @{
+ *
+ * @file
+ * @brief       Test custom lua loader ("require" module)
+ *
+ * @author      Juan Carrano <j.carrano@fu-berlin.de>
+ *
+ * This application defines some pure-lua and some C-extension modules.
+ * First it tests running a module as a script (lua_riot_do_module).
+ *
+ * Then it goes into a loop reading single lines of input from stdin and
+ * executing them as lua source. The test script can use that to test module
+ * loading.
+ *
+ * @}
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+
+#include "lua_run.h"
+#include "lua_builtin.h"
+
+static const uint8_t pure_module_1[] = "\n\
+return {a='Quando uma lua'}\n\
+";
+
+static const uint8_t pure_module_2[] = "\n\
+return {a='chega de repente'}\n\
+";
+
+static const uint8_t test_script[] = "\n\
+print'I am a module, hi!'\n\
+";
+
+/* The -1 is because the final '\0' is not part of the source code that should
+ * be read by lua.
+ */
+const struct lua_riot_builtin_lua _lua_riot_builtin_lua_table[] = {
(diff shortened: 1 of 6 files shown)


jcarrano:
I labeled it as a bug because I was not respecting the API and that qualifies as a bug.

@danpetry Do you mind giving me a hand with this one?

jcarrano:
@danpetry Ping.

tcschmidt:
@jcarrano ping.

jcarrano:
Well, sorry for the delay. Testing this required creating some lua C-extension modules, which I was lazy to write and also did not have time to. I finally got around to doing it, and uncovered another bug in the process.

I added a test application.

tests/lua_loader/main.c
```diff
@@ -0,0 +1,103 @@
...
+static char lua_mem[LUA_MEM_SIZE] __attribute__ ((aligned(__BIGGEST_ALIGNMENT__)));
+
+#define LINEBUF_SZ (32)
+static char linebuf[LINEBUF_SZ];
+
+int main(void)
+{
+    int status;
+    status = lua_riot_do_module("test", lua_mem, LUA_MEM_SIZE,
+                                        LUAR_LOAD_BASE, NULL);
```
leandrolanzieri: Please align properly.

jcarrano:
@leandrolanzieri 

> I noticed a change of behaviour compared to master, when trying to require a module that is not defined.

D'oh! good catch. I made a copypasta-error (it already felt like copypasta when I was writing the code, I should't have ignored that hunch).

I removed the ready-for build. I'll add it back once I have squashed.

leandrolanzieri:
@jcarrano that seems to fix it. Maybe you could add '\n' to the error strings, currently the output looks like:

```
Runtime error   [string "return require "something""]:1: module 'something' not found:
        no field package.preload['something']Module not found in Lua-builtinsModule not found in C-builtins
```

jcarrano:
I added some newline and tabs, it should be prettier now.

tests/lua_loader/tests/01-run.py
```diff
@@ -0,0 +1,36 @@
...
+# directory for more details.
+
+# Tell the lua interpreter running in riot to load some modules and print
+# the value of a variable inside that module.
+
+import os
+import sys
+
+MODULE_QUERIES = [
+    ("m1", "a", "1"),
```
leandrolanzieri: I just realized, if the expected value matches the module name, then the `expect_exact` will pass no matter the returned value, right?

jcarrano: Dang! you are right! For some reason I thought expect_exact meant "expect exactly this". It seems I'm not be smart enough to use pexpect.

But wait, there's even more! Say I changed the 1 with a 9, then thanks to glorious pyterm the
"2019-05-31 ....!" at the beginning of a line would also be matched. It seems I'm not smart enough to use pyterm either.

leandrolanzieri:
@danpetry any other comments?

PR #9733, opened 2018-08-07
sys/shell: Add function to call the shell with a list of strings.

jcarrano:
### Contribution description

The new function makes it easier to call command line tools from within code.

I'm using it to provide a sort of system()/execl() call from within Lua.

### Related PRs

Depends on #10788 to integrate the tests into the main shell tests.

### DIFF
--- a/sys/include/shell.h
+++ b/sys/include/shell.h
@@ -72,6 +72,19 @@ typedef struct shell_command_t {
  */
 void shell_run(const shell_command_t *commands, char *line_buf, int len);
 
+/**
+ * @brief           Call a command-line tool.
+ *
+ * @param[in]       shell_commands    Array of user-defined commands.
+ * @param           argc    Number if string in argv. Must be greater or equal to 1.
+ * @param           argv    Array of pointers to string. argv[0] is the name of
+ *                          the command.
+ *
+ * @return          -1 if the command was not found or if argc < 1.
+ * @return          If the command is executed, it's exit code.
+ */
+int shell_call(const shell_command_t *shell_commands, int argc, char **argv);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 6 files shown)


danpetry:
Re the high level aspects of the PR:
1. Could you please provide tests?
2. Could you please explain briefly why you want to provide a Lua binding to the shell rather than binding to the modules individually, and accessing their commands directly? (Please correct me if my understanding here is inaccurate)

jcarrano:
@danpetry I added the test. The rebase was unintended (our awesome static-test script did it.)

jcarrano:
> Could you please explain briefly why you want to provide a Lua binding to the shell rather than binding to the modules individually, and accessing their commands directly? (Please correct me if my understanding here is inaccurate)

Because I won't write individual bindings to modules (other than as an example). This will allow me to provide a [system()](http://pubs.opengroup.org/onlinepubs/9699919799/functions/system.html)-like (lua has [os.execute](https://www.lua.org/manual/5.3/manual.html#pdf-os.execute) but that takes a single string.) 

The reason for providing such a function from lua is convenience: having an easy to use example application without having to add a lot of code.

kaspar030:
LGTM.

How about merging the test with tests/shell?

tests/shell_call/Makefile
```diff
@@ -0,0 +1,13 @@
...
+USEMODULE += shell
+
+DISABLE_MODULE += auto_init
+
+# chronos is missing a getchar implementation
+BOARD_BLACKLIST += chronos
+
+include $(RIOTBASE)/Makefile.include
+
+test:
```
kaspar030: This target is not needed anymore.

kaspar030: anyhow, please merge the test into tests/shell. They're testing the same module, and tests are kinda expensive for the CI...

jcarrano:
I rebased on top of #10788 so that tests can be more easily integrated.

jcarrano:
I'm not rebasing this (yet) again. Closing.

PR #9741, opened 2018-08-08
makefiles/buildtests.inc.mk: allow overwriting make output redirection

cladmi:
### Contribution description

When running tests, I often needed to modify `buildtests.inc.mk` to see the
actual errors/standard output.
This allows overwriting the default redirection of both to `/dev/null`.
It is low level and directly given to the command execution.

As it is interpreted by make, it can even be overwritten using build system
variables:

    'BUILDTEST_MAKE_REDIRECT=>/tmp/buildtest.$${board}.out 2>&1'

### Issues/PRs references

I needed something like this when running `buildtest` for the release tests.

### DIFF
--- a/makefiles/buildtests.inc.mk
+++ b/makefiles/buildtests.inc.mk
@@ -1,5 +1,7 @@
 .PHONY: buildtest
 
+BUILDTEST_MAKE_REDIRECT ?= >/dev/null 2>&1
+
 ifeq ($(BUILD_IN_DOCKER),1)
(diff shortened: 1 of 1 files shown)


smlng:
ah when I do 

```
BUILDTEST_MAKE_REDIRECT='>/tmp/buildtest.$${board}.out 2>&1' make -C tests/minimal/ buildtest
```

it works

cladmi:
Thanks for testing.

PR #9756, opened 2018-08-10
pkg: update ccn-lite and adapt shell commands

smlng:
This updates the ccn-lite package version which brings in the
    latest upstream fixes for some compiler issues found on macOS
    with clang and newer GCC versions.

    A minor adaption of the RIOT shell commands is also included.

### Issues/PRs references

https://github.com/cn-uofbasel/ccn-lite/pull/298

### DIFF
--- a/pkg/ccn-lite/Makefile
+++ b/pkg/ccn-lite/Makefile
@@ -1,6 +1,6 @@
 PKG_NAME=ccn-lite
 PKG_URL=https://github.com/cn-uofbasel/ccn-lite/
-PKG_VERSION=48b17ebee7600b2e79b3acf0728217473d7a4ee8
+PKG_VERSION=f85a2a055db92b5978f2d4f92b830b5b6b9acb42
 PKG_LICENSE=ISC
 
 .PHONY: all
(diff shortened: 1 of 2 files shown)


smlng:
huhu @cgundogan you sure want this, right 😉

cgundogan:
Why again did you add a 4. parameter?
https://github.com/cn-uofbasel/ccn-lite/blob/master/src/ccnl-core/src/ccnl-prefix.c#L267
https://github.com/cn-uofbasel/ccn-lite/blob/master/src/ccnl-core/include/ccnl-prefix.h#L155

according to the CCN.-lite master, there are only 3 parameters for `ccnl_URItoPrefix`

smlng:
> Why again did you add a 4. parameter?
 why me? In this PR I remove the 4th parameter to make it work.

smlng:
and according to `git blame` Oleg introduced the original shell command using 4 params:

```
5db537fbab (Oleg Hahm         2016-02-03 16:46:28 +0100 170)     struct ccnl_prefix_s *prefix = ccnl_URItoPrefix(pfx, suite, NULL, 0);
```

cgundogan:
Hm, that's true (:

PR #9759, opened 2018-08-10
 dist: move testrunner to pythonlibs as package 

smlng:
### Contribution description

As testrunner is moved to dist/pythonlibs which is exported
via PYTHONPATH, testrunner is found by all test scripts.

### Issues/PRs references

#9758


aabadie:
Nice one. I like it.

tests/bench_msg_pingpong/tests/01-run.py
```diff
@@ -16,6 +15,5 @@ def testfunc(child):
 
 
 if __name__ == "__main__":
-    sys.path.append(os.path.join(os.environ['RIOTTOOLS'], 'testrunner'))
```
aabadie: The line below can now be moved with the other imports at the top of the script.

smlng:
@aabadie please note this is based on #9758, maybe you can also have a look at that one, too ... THX!

tests/bench_msg_pingpong/tests/01-run.py
smlng: yes will do

aabadie:
@smlng, I don't think removing `sys` is a good idea. See [this](https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python)

smlng:
> I don't think removing sys I

was suggested by @cladmi, but reading the linked blog-entry it sound reasonable to use `sys.exit` in out context - so should I drop the change?

cladmi:
> was suggested by @cladmi, but reading the linked blog-entry it sound reasonable to use sys.exit in out context - so should I drop the change?

My bad, I did not know about this https://docs.python.org/3/library/constants.html#exit and never had issues before so was wondering why.

smlng:
dropped the removal of `import sys` and fixed flake8 errors reported by Murdock, good to review this one again - needs squashing now. @cladmi and @aabadie?

smlng:
squashed, I'd still like to have an ACK by @cladmi on this one, too. So please don't merge right away, but leave him do it, thx.

cladmi:
I will take care of reviewing after the release.

cladmi:
Changes look good, I will re-run tests and see if something complains.

cladmi:
Some gardening may need to be done after PRs still using `sys.append.path` are merged.
The tests may be found using:

    git grep sys.path.append

PR #9760, opened 2018-08-10
boards/stm32f103c8: fix openocd config

MrKevinWeiss:
### Contribution description

Update the common/stm32f103c8 openocd configuration file to fix occasional flash problems that was noticed when running multiple tests with [compile_and_test_for_board.py](https://github.com/RIOT-OS/Release-Specs/pull/68).

I used the default configuration from ST for the board and reran all the tests.  I did not receive any flashing errors after that.  I added modifications for RIOT and to allow debug.

### Issues/PRs references
[Release-Specs/pull/68](https://github.com/RIOT-OS/Release-Specs/pull/68).

### DIFF
--- a/boards/common/stm32f103c8/dist/openocd.cfg
+++ b/boards/common/stm32f103c8/dist/openocd.cfg
@@ -1,4 +1,25 @@
+source [find interface/stlink.cfg]
+
+set WORKAREASIZE 0x5000
+
+transport select "hla_swd"
+
+set CHIPNAME STM32F103C8Tx
+
+# Enable debug when in low power modes
+set ENABLE_LOW_POWER 1
+
+# Stop Watchdog counters when halt
+set STOP_WATCHDOG 1
+
+# STlink Debug clock frequency
+set CLOCK_FREQ 4000
+
+# use hardware reset, connect under reset
+# connect_assert_srst needed if low power mode application running (WFI...)
+reset_config srst_only srst_nogate connect_assert_srst
+set CONNECT_UNDER_RESET 1
+
 source [find target/stm32f1x.cfg]
-reset_config none separate
 
 $_TARGETNAME configure -rtos auto
--- a/boards/common/stm32f103c8/Makefile.include
+++ b/boards/common/stm32f103c8/Makefile.include
@@ -30,5 +30,9 @@ else
   export DEBUG_ADAPTER ?= stlink
   export STLINK_VERSION ?= 2
 
+  # call a 'reset halt' command before starting the debugger
+  # it is required as `connect_assert_srst` is set
+  export OPENOCD_DBG_START_CMD = -c 'reset halt'
+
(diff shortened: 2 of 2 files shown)


boards/common/stm32f103c8/dist/openocd.cfg
```diff
@@ -1,4 +1,25 @@
+source [find interface/stlink.cfg]
```
aabadie: I don't think this is required. This is because DEBUG_ADAPTER is set in Makefile.include. Then openocd make will automatically include `stlink.inc.mk`, which will add the right command.

MrKevinWeiss:
I still need to test again and try to see if debug works.

MrKevinWeiss:
@smlng Hmmm, OK I'll have to also try a bit later.  A quick search didn't reveal anything.  First I should confirm debug though.

cladmi:
It should be fixed on mac now that you removed the `source [find interface/stlink.cfg]` line.
Merging all `stlink` in one file is quite recent in `openocd`.

MrKevinWeiss:
@smlng can you retry?

smlng:
jep, works on macOS

MrKevinWeiss:
Anyone know why travis can't build?

boards/common/stm32f103c8/dist/openocd.cfg
```diff
@@ -1,4 +1,21 @@
...
+
+# Stop Watchdog counters when halt
+set STOP_WATCHDOG 1
+
+# STlink Debug clock frequency
+set CLOCK_FREQ 4000
+
+# use hardware reset, connect under reset
+# connect_assert_srst needed if low power mode application running (WFI...)
+reset_config srst_only srst_nogate connect_assert_srst
```
maribu: @MrKevinWeiss: I think that this change is kind of a regression when having the common Blue Pill and Black Pill user in mind: Most users likely just connect SWDIO, SWDCLK, GND and 3.3V to the device. In that case the user will now have to press the reset button manually to trigger flashing and once again after flashing to boot the new firmware.

Would it not make more sense to keep the default reset config, unless e.g. an environment variable like `STM32F103_USE_HW_RESET` is set to `1`?

Alternatively giving a hint in the documentation of the Blue Pill in the flash section that the debugger needs to be connected not only to SWDIO and SWDCLK, but also to the reset pin of the board would solve the issue to me. That way users won't start wondering why flashing using this OpenOCD config won't work without the debugger connected to the reset pin, while it does work fine with the default config of OpenOCD.

cladmi: Something that could be nice, would be to update the `openocd` handling so you could say `OPENOCD_FILE=openocd_hw_reset.cfg` or something. This way, no need for a board specific variable but it would be another task for the future.
Openocd has `--filename` for the config file and `--search` for the search path which would work nicely in that case.

PR #9779, opened 2018-08-15
boards: Harmonized UART config of Nucleo-64 boards

maribu:
# Motivation
The Nucleo-64 boards are a convenient platform to evaluate the different STM32 MCUs, as they are pretty identical. However, the configuration in RIOT creates differences between the boards, e.g. the UARTs are added in different order. This PR changes the order so that USARTs are added in ascending order with the exception of USART2, which is always the first. (USART2 is the one connected to the integrated ST-LINKv2, for this special role it was chosen as the first.) In fact, most Nucleo-64 boards did already comply with this order.

# Changes in Detail

- nucleo-f334r8:
  - Added USART1 and USART3 config
- nucleo-f401re:
  - Changed order of UARTs to match the other boards
- nucleo-f410rb:
  - Added USART6 configuration
  - Fixed incorrect DMA settings (according to reference manual of the MCU)
- nucleo-f446re:
  - Changed order of UARTs to match the other boards
- nucleo-l152re:
  - Changed order of UARTs to match the other boards

### DIFF
--- a/boards/nucleo-f334r8/include/periph_conf.h
+++ b/boards/nucleo-f334r8/include/periph_conf.h
@@ -90,10 +90,32 @@ static const uart_conf_t uart_config[] = {
         .tx_af    = GPIO_AF7,
         .bus      = APB1,
         .irqn     = USART2_IRQn
+    },
+    {
+        .dev        = USART1,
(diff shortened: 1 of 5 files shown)


PeterKietzmann:
@MrKevinWeiss would you review this PR?!

Issue #7734, opened 2017-10-13
stm32: approach for common peripheral definitions


nmeum:
While porting RIOT to the bluepill port #7727 we noticed that there are many boards supported by RIOT which use the stm32f103 MCU. Since many definitions we added and tested are similar among the boards which use this MCU we wonder if it would make sense to implement a common `periph_conf.h` header file for those boards.

For instance if you compare the `periph_conf.h` header file of the nucleo-f103 board with the one from the maple-mini they look eye to eye identical. The problem with this code duplication is the following: When porting RIOT to the bluepilll board we found that some peripherals were not configured properly, for instance the general purpose timer four wasn't configured. Instead of backporting the changes made by us for the bluepill board we would suggest adding some kind of common `periph_conf.h` for these boards.

Any suggestion on how to do this? We noticed the `-common` board folders in `boards/`, however, they don't seem to fit our purpose sense these are common folders for boards made by one manufacturer.

temmihoo:
In the MCU world most "board" peripherals are actually integrated within the MCU itself by the silicon vendor.

Often the only board-specific part of configuration are the board pin names. Some boards do have external-to-MCU hardware features such as early ELL-i boards having ENC28J60 and quite many boards having a number of LEDs and switches not to mention simple sensors and the like. Supporting these easily without necessitating MCU integrated peripheral support repetition would be ideal.

I think it would make sense to have a commons system that would allow (not exhaustive list):
- things common to a specific MCU model such as STM32F334R8
- things common to a MCU model such as STM32F334*
- things common to a MCU family such as STM32F3* or even STM32F*
- things common to a MCU core such as Cortex M3
- things common to a board family such as ST Nucleo64-*

Personally I tend to think of the board specific pin names as mostly useless abstraction and am always thinking from point of view of the MCU model specific reference manual. This is so because my aim is to help in designing products out of the chips instead of developing endlessly on the push pin wire level. Also, when reading of the hardware peripherals I am reading the MCU model specific reference manual instead of the board vendor slideware.

Achieving this type of orthogonality might be technically difficult using Makefiles and #include directives so some sort of preprocessing might be needed.

temmihoo:
#7661 is related

haukepetersen:
[...]

PR #9780, opened 2018-08-15
boards: Extended/fixed SPI config for STM32 Nucleo-64 boards

maribu:
- nucleo-f030r8:
    - Added SPI config
    - Tested both SPI interfaces successfully
- nucleo-f072rb:
    - Added config for SPI2
    - needs to be tested (I don't have that board (yet))
- nucleo-f303re:
    - Fixed typo in SPI config
    - Extended SPI config
    - Tested new SPI2 interface successfully

### DIFF
--- a/boards/nucleo-f030r8/include/periph_conf.h
+++ b/boards/nucleo-f030r8/include/periph_conf.h
@@ -140,6 +140,56 @@ static const pwm_conf_t pwm_config[] = {
 /** @} */
 
 /**
+ * @name   SPI configuration
+ *
+ * @note    The spi_divtable is auto-generated from
+ *          `cpu/stm32_common/dist/spi_divtable/spi_divtable.c`
+ * @{
+ */
+static const uint8_t spi_divtable[2][5] = {
+    {       /* for APB1 @ 48000000Hz */
+        7,  /* -> 187500Hz */
+        6,  /* -> 375000Hz */
+        5,  /* -> 750000Hz */
+        2,  /* -> 6000000Hz */
+        1   /* -> 12000000Hz */
+    },
+    {       /* for APB2 @ 48000000Hz */
+        7,  /* -> 187500Hz */
+        6,  /* -> 375000Hz */
+        5,  /* -> 750000Hz */
+        2,  /* -> 6000000Hz */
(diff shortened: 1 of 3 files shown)


PeterKietzmann:
Code changes look ok – didn't check the configurations yet and neither tested so far. Seeing this PR reminds me that something like #7734 could make a lot of sense.

maribu:
@PeterKietzmann: Yes, totally.

But there are some gotchas to keep in mind:
- Not every pin is available on every board. E.g. an SPI configuration would not make much sense, if the corresponding pins are not available on the specific board
- Some MCUs allow supplying external oscillators. So, e.g. the prescaler configs for the SPI clock could be different on two boards with the exact same MCU
- Some buses (e.g. SPI or I2C on STM32) can be used on different pins. Maybe one board provides some pins as a special connector (e.g. similar to the I2C connector on the Calliope Mini), but another board uses the same interface on alternative pins

I think per CPU header files (e.g. name `periph_spi.h`, `periph_i2c.h`, ...) with some `#ifdef`s where needed could solve the problem:

```
static const spi_conf_t spi_config[] = {
#ifdef PROVIDE_SPI1
    {
        .dev      = SPI1,
#ifdef SPI1_VARIANT_2
        .mosi_pin = GPIO_PIN(PORT_B, 5),
        .miso_pin = GPIO_PIN(PORT_B, 4),
        .sclk_pin = GPIO_PIN(PORT_B, 3),
        .cs_pin   = GPIO_PIN(PORT_A, 15),
#else
        .mosi_pin = GPIO_PIN(PORT_A, 7),
        .miso_pin = GPIO_PIN(PORT_A, 6),
        .sclk_pin = GPIO_PIN(PORT_A, 5),
        .cs_pin   = GPIO_PIN(PORT_A, 4),
#endif
        .af       = GPIO_AF5,
        .rccmask  = RCC_APB2ENR_SPI1EN,
        .apbbus   = APB2
    },
#endif
};
```

The `#ifdef` for the different pinout variants could be added when needed, so that the code remains simple of available hardware is not to diverse.

PeterKietzmann:
No complaints for more than 4 days -> ACK and go

PR #9781, opened 2018-08-15
atmega_common: allow defining rom and ram length for link.

cladmi:
### Contribution description

Allow configuring __TEXT_REGION_LENGTH__ and __DATA_REGION_LENGTH__ linkerscript
variables using ROM_LEN and RAM_LEN makefile variables.

This also add a compatibility for ubuntu xenial / current murdock by using newer linker scripts.

### TODO

Update all atmega CPUs ROM and RAM LENGTH and `BOARD_NOT_ENOUGH_MEMORY` in murdock accordingly

* [x] Find info on bootloader sizes
  * Missing some listed inline
* [x]  cpu/atmega1281
* [x] cpu/atmega1284p
* [x] cpu/atmega2560
* [x] cpu/atmega256rfr2
* [x] cpu/atmega328p
* [ ] Update Not enough memory
* BOOTLOADER_SIZE
   * [x] jiminy-mega25rfr2
   * [x] mega-xplained
   * [x] waspmote-pro

### Issues/PRs references

Found in discussion https://github.com/RIOT-OS/RIOT/pull/9628#issuecomment-412926167

### DIFF
--- a/boards/mega-xplained/Makefile.include
+++ b/boards/mega-xplained/Makefile.include
@@ -1,6 +1,11 @@
 # define the cpu used by the Mega Xplained board
 export CPU = atmega1284p
 
+# Found by checking fuse settings (2048 words so 4KB)
+# https://www.microchip.com/DevelopmentTools/ProductDetails/atmega1284p-xpld
+BOOTLOADER_SIZE ?= 4K
+ROM_RESERVED ?= $(BOOTLOADER_SIZE)
+
 # configure the terminal program
 export PORT_LINUX  ?= /dev/ttyACM0
 export PORT_DARWIN ?= $(firstword $(sort $(wildcard /dev/tty.usbserial*)))
--- a/boards/waspmote-pro/Makefile.include
+++ b/boards/waspmote-pro/Makefile.include
@@ -1,6 +1,12 @@
 # define the cpu used by the waspmote pro board
 export CPU = atmega1281
 
+# Bootloader uses stk500v1 protocol, which usually is implemented in
+# bootloaders of 2K size.
+# http://www.libelium.com/products/waspmote/hardware/
+BOOTLOADER_SIZE ?= 2K
+ROM_RESERVED ?= $(BOOTLOADER_SIZE)
+
 # configure the terminal program
 PORT_LINUX  ?= /dev/ttyACM0
 PORT_DARWIN ?= $(firstword $(sort $(wildcard /dev/tty.usbserial*)))
--- a/boards/arduino-duemilanove/Makefile.include
+++ b/boards/arduino-duemilanove/Makefile.include
@@ -14,4 +14,7 @@ export FFLAGS += -p m328p
 # overridden for debugging (which requires changes that require to use an ISP)
 export PROGRAMMER ?= arduino
 
+BOOTLOADER_SIZE ?= 2K
+ROM_RESERVED ?= $(BOOTLOADER_SIZE)
+
 include $(RIOTBOARD)/common/arduino-atmega/Makefile.include
--- a/boards/arduino-mega2560/Makefile.include
+++ b/boards/arduino-mega2560/Makefile.include
@@ -14,4 +14,7 @@ export FFLAGS += -p m2560
 # overridden for debugging (which requires changes that require to use an ISP)
 export PROGRAMMER ?= stk500v2
 
+BOOTLOADER_SIZE ?= 8K
+ROM_RESERVED ?= $(BOOTLOADER_SIZE)
+
 include $(RIOTBOARD)/common/arduino-atmega/Makefile.include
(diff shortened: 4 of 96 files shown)


cladmi:
I talked with @kYc0o regarding how is the bootloader handled.
As the linker does not know about it, the firmware is always linked at address 0 so we suppose it is stored at the end of the ROM.

This means, that the `ROM_OFFSET` variable from `cortexm_common` should not be used here and maybe another `ATMEGA_BOOLOADER_SIZE` variable could be defined and used for the `__TEXT__REGION_LENGTH__` symbol definition.

cladmi:
@maribu This could interest you too

cladmi:
I am missing the bootloader size for the following boards (if there is one ?). Maybe you can know this information as you added the board support:

* jiminy-mega25rfr2 @Josar 
* mega-xplained @ZetaR60 
* waspmote-pro @kYc0o

cladmi:
Looks like setting the bootloader size did not required blacklisting any applications.
But the bootloader size is taken into account:


```
make -C examples/hello-world/ BOARD=arduino-mega2560 QUIET=0
... -Wl,--defsym=__TEXT_REGION_LENGTH__=256K-8K   ...

avr-readelf --symbols examples/hello-world/bin/arduino-mega2560/hello-world.elf  | grep __TEXT_REGION_LENGTH__
   369: 0003e000     0 NOTYPE  GLOBAL DEFAULT  ABS __TEXT_REGION_LENGTH__
```

cladmi:
To handle the bootloader that I think is located at the end of the ROM I added a `ROM_RESERVED` variable, which is basically the opposite of `ROM_OFFSET` that was used in `cortexm_common`. Not sure about the naming. @kYc0o maybe you have an opinion about this.

boards/mega-xplained/Makefile.include
```diff
@@ -1,6 +1,11 @@
 # define the cpu used by the Mega Xplained board
 export CPU = atmega1284p
 
+# TODO is there a bootloader on the flash ? I do not understand...
+# https://www.microchip.com/DevelopmentTools/ProductDetails/atmega1284p-xpld
+# ARDUINO_BOOTLOADER_SIZE = 2K
+ROM_RESERVED = $(ARDUINO_BOOTLOADER_SIZE)
```
ZetaR60: I could not find the bootloader size in the official docs. I did check the fuse settings for my mega-xplained and it is set to a boot size of 2048 words, which corresponds to 4KB (I assume that 4K is 4096 here and not 4000).

Currently the bootloader on mega-xplained is being clobbered rather than being used, due to a few things not being properly supported. Does this affect reserving ROM?

Also, this and a few of the other boards are not Arduinos, so you might want to change the name of ```ARDUINO_BOOTLOADER_SIZE``` to avoid confusion.

kYc0o: Yes, on the mega-xplained the bootloader is not being used since it requires a very old programmer which is not available for Linux and afaik is not free. It is only included in some (old) versions of Atmel Studio.

I'd vote in favor to not taking it into account, how ever we should put a big warning somewhere indicating that the bootloader is being removed.

boards/waspmote-pro/Makefile.include
```diff
@@ -1,6 +1,11 @@
 # define the cpu used by the waspmote pro board
 export CPU = atmega1281
 
+# TODO is there a bootloader on the flash ? I do not understand...
+# http://www.libelium.com/products/waspmote/hardware/
+# ARDUINO_BOOTLOADER_SIZE = 2K
+ROM_RESERVED = $(ARDUINO_BOOTLOADER_SIZE)
```
kYc0o: There's a bootloader and it takes 2K, so please change.

kYc0o:
Another small question, why did you call the linker folder "ldscripts_compat"?

boards/waspmote-pro/Makefile.include
cladmi: Any source to put as reference ?

boards/mega-xplained/Makefile.include
cladmi: I would rather say take the bootloader into account even if we are not using it.
So if a user is using his board both on windows and linux it works, if he uses only on Linux, he wastes the last 4k that are not used as he is not using the BL.

boards/waspmote-pro/Makefile.include
kYc0o: Well, I know it uses stk500v1 protocol, which usually is implemented in bootloaders of 2K size. I tried to find exactly the reference but no luck. Maybe someone at @Libelium might know.

boards/mega-xplained/Makefile.include
ZetaR60: That sounds okay, or even just ```BOOTLOADER_SIZE```.

cladmi:
I did not yet, but I can replace the `=` to `?=` to allow overwriting the bootloader size / rom reserved from environment variable.

kYc0o:
> I did not yet, but I can replace the = to ?= to allow overwriting the bootloader size / rom reserved from environment variable.

Please do it and squash, this PR is ready.

PR #9782, opened 2018-08-15
boards/stm32f103c8: update flash flags for DFU

smlng:
### Contribution description

This PR updates the flash flags for the alternative DFU programmer, this is necessary as the boot loader was updated with new USB device ID. This also adds a short documentation on how to flash the boot loader and use DFU to flash RIOT binaries.


smlng:
lets leave the udev rules to be another PR as discussed f2f, okay?

PR #9783, opened 2018-08-15
[RFC] drivers/cc110x: Extended cc110x_params_t

maribu:
# Motivation
Currently the cc110x driver allows to change the base frequency manually by `#define`-ing the configuration register values. Arbitrary frequencies could be set in software, but based on the circuit/antenna design only a narrow bandwidth will actually work. In practice, only 433MHz, 868MHz and 900MHz are relevant. Even though the actual frequency also depends on the oscillator used, only a 26 MHz quartz (as used in the reference design) is used in practice.

Channel configuration in those ISM band is quite harmonized (and sometimes even enforced by legislation). As a result, three possible base frequencies should be sufficient (~433MHz, ~868MHz, ~900MHz). This PR allows choosing the base frequency in `cc110x_params_t` using an enum of the three possible base frequencies.

# Changes in detail
- Added two entries to cc110x_params_t:
    - Allows choosing base frequency (433 MHz, 868 MHz, 900 MHz)
    - Allows choosing SPI frequency
- Reduced SPI clock frequency on the MSB-IoT
    - RTT (ping6) increases from ~16.7ms to ~17.3ms
    - packet loss rate decreases from ~90% to ~0%

# To be discussed
What are the "best" configuration for the three base frequencies? How about the bandwidth of each channel? Currently 49.99kHz channel width is configured, which matches the input filter of width of 58.04kHz quite nicely. However, 25kHz channel width matches e.g. LPD433 channel spacing.

Before this commit two different base frequency configurations were used: 869.5250 MHz in  `const char cc110x_default_conf[]` and 869.5269 MHz in `const char cc110x_default_base_freq[3]`. While the former is just in the middle of an 25kHz wide 868 MHz channel, the latter is just between two 25kHz channels.

Choosing a frequency just between of two 25kHz channels seems to make sense when using 50kHz channels, so that those two channels are actually "combined". But is this legal? If not, I suggest to stick with the 25kHz channels spacing and use a base frequency in the lowest 25kHz channel and reduce the channel bandwidth two 25kHz. (However, the input filter cannot be smaller than 58.04 kHz.)

(Currently this commit uses the center of 25kHz channels as base frequency with 50 kHz channel width. This would interfere with 3 channels and is the worst possible configuration. So this PR should not be accepted as it is.)

# Testing
- [x] Test with an 868 MHz circuit (done on MSB-IoT and MSB-A2)
- [x] Test with an 433 MHz circuit (done with an CC1101 breakout board on nucleo-f303re boards)
- [ ] Test with an 900 MHz circuit (not legal in E.U. as far as I know, so I'm unable to test. Also, I do not have any CC1101 with appropriate antenna design)

### DIFF
--- a/drivers/cc110x/cc110x.c
+++ b/drivers/cc110x/cc110x.c
@@ -38,6 +38,8 @@
 #define ENABLE_DEBUG    (0)
 #include "debug.h"
 
+static const char *cc110x_band_name[] = { "433", "868", "900" };
+
 /* Internal function prototypes */
 #ifndef CC110X_DONT_RESET
 static void _reset(cc110x_t *dev);
@@ -75,7 +77,7 @@ int cc110x_setup(cc110x_t *dev, const cc110x_params_t *params)
     cc110x_writeburst_reg(dev, CC110X_PATABLE, CC110X_DEFAULT_PATABLE, 8);
 
     /* set base frequency */
-    cc110x_set_base_freq_raw(dev, CC110X_DEFAULT_FREQ);
+    cc110x_set_band(dev, dev->params.band);
 
     /* Set default channel number */
     cc110x_set_channel(dev, CC110X_DEFAULT_CHANNEL);
@@ -85,9 +87,10 @@ int cc110x_setup(cc110x_t *dev, const cc110x_params_t *params)
     luid_get(&addr, 1);
     cc110x_set_address(dev, addr);
 
-    LOG_INFO("cc110x: initialized with address=%u and channel=%i\n",
+    LOG_INFO("cc110x: initialized with address=%u and channel=%i @ %sMHz\n",
             (unsigned)dev->radio_address,
-            dev->radio_channel);
+            dev->radio_channel,
+            cc110x_band_name[dev->params.band]);
 
     return 0;
 }
@@ -107,16 +110,13 @@ uint8_t cc110x_set_address(cc110x_t *dev, uint8_t address)
     return 0;
 }
 
-void cc110x_set_base_freq_raw(cc110x_t *dev, const char* freq_array)
+void cc110x_set_band(cc110x_t *dev, cc110x_band_t band)
 {
 #if ENABLE_DEBUG == 1
-    uint8_t _tmp[] = { freq_array[2], freq_array[1], freq_array[0], 0x00};
-    uint32_t *FREQ = (uint32_t*) _tmp;
 
-    DEBUG("cc110x_set_base_freq_raw(): setting base frequency to %uHz\n",
-            (26000000>>16) * (unsigned)(*FREQ));
+    DEBUG("cc110x_set_band(): Setting band to %sMHz\n", cc110x_band_name[band]);
 #endif
-    cc110x_writeburst_reg(dev, CC110X_FREQ2, freq_array, 3);
+    cc110x_writeburst_reg(dev, CC110X_FREQ2, cc110x_base_freqs[band], 3);
 }
 
 void cc110x_set_monitor(cc110x_t *dev, uint8_t mode)
--- a/drivers/cc110x/cc110x.c
+++ b/drivers/cc110x/cc110x.c
@@ -38,7 +38,9 @@
 #define ENABLE_DEBUG    (0)
 #include "debug.h"
 
+#if ENABLE_DEBUG == 1
 static const char *cc110x_band_name[] = { "433", "868", "900" };
+#endif
 
 /* Internal function prototypes */
 #ifndef CC110X_DONT_RESET
@@ -113,7 +115,6 @@ uint8_t cc110x_set_address(cc110x_t *dev, uint8_t address)
 void cc110x_set_band(cc110x_t *dev, cc110x_band_t band)
 {
 #if ENABLE_DEBUG == 1
-
     DEBUG("cc110x_set_band(): Setting band to %sMHz\n", cc110x_band_name[band]);
 #endif
     cc110x_writeburst_reg(dev, CC110X_FREQ2, cc110x_base_freqs[band], 3);
--- a/drivers/cc110x/cc110x.c
+++ b/drivers/cc110x/cc110x.c
@@ -89,10 +89,9 @@ int cc110x_setup(cc110x_t *dev, const cc110x_params_t *params)
     luid_get(&addr, 1);
     cc110x_set_address(dev, addr);
 
-    LOG_INFO("cc110x: initialized with address=%u and channel=%i @ %sMHz\n",
+    LOG_INFO("cc110x: initialized with address=%u and channel=%i\n",
             (unsigned)dev->radio_address,
-            dev->radio_channel,
-            cc110x_band_name[dev->params.band]);
+            dev->radio_channel);
 
     return 0;
 }
--- a/drivers/cc110x/include/cc110x-defaultsettings.h
+++ b/drivers/cc110x/include/cc110x-defaultsettings.h
@@ -34,7 +34,6 @@ extern const char cc110x_default_pa_table[8];
 #endif
 
 extern const char cc110x_base_freqs[3][3];
-extern const char cc110x_channel_widths[3][2];
 
 #ifdef __cplusplus
 }
--- a/drivers/cc110x/include/cc110x-defaultsettings.h
+++ b/drivers/cc110x/include/cc110x-defaultsettings.h
@@ -33,10 +33,8 @@ extern "C" {
 extern const char cc110x_default_pa_table[8];
 #endif
 
-#ifndef CC110X_DEFAULT_FREQ
-#define CC110X_DEFAULT_FREQ cc110x_default_base_freq
-extern const char cc110x_default_base_freq[3];
-#endif
+extern const char cc110x_base_freqs[3][3];
+extern const char cc110x_channel_widths[3][2];
 
 #ifdef __cplusplus
 }
--- a/boards/msbiot/include/board.h
+++ b/boards/msbiot/include/board.h
@@ -30,11 +30,12 @@ extern "C" {
  * @name    Configure connected CC1101 (radio) device
  * @{
  */
-#define CC110X_PARAM_SPI    SPI_DEV(0)
-#define CC110X_PARAM_CS     GPIO_PIN(PORT_B, 12)
-#define CC110X_PARAM_GDO0   GPIO_PIN(PORT_C, 4)
-#define CC110X_PARAM_GDO1   GPIO_PIN(PORT_A, 6)
-#define CC110X_PARAM_GDO2   GPIO_PIN(PORT_C, 5)
+#define CC110X_PARAM_SPI      SPI_DEV(0)
+#define CC110X_PARAM_CS       GPIO_PIN(PORT_B, 12)
+#define CC110X_PARAM_GDO0     GPIO_PIN(PORT_C, 4)
+#define CC110X_PARAM_GDO1     GPIO_PIN(PORT_A, 6)
+#define CC110X_PARAM_GDO2     GPIO_PIN(PORT_C, 5)
+#define CC110X_PARAM_SPI_CLK  SPI_CLK_1MHZ
 /** @} */
 
 /**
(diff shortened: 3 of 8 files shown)


maribu:
Choosing the SPI clock frequency btw. solves https://github.com/RIOT-OS/RIOT/issues/9225. (Or better provides a work around. Actually the CC1101 should be compatible with up to 10 MHz - so 5 MHz should be fine.)

PeterKietzmann:
@OlegHahm, @kaspar030 interested?

drivers/cc110x/include/cc110x-defaultsettings.h
```diff
@@ -33,10 +33,8 @@ extern "C" {
 extern const char cc110x_default_pa_table[8];
 #endif
 
-#ifndef CC110X_DEFAULT_FREQ
-#define CC110X_DEFAULT_FREQ cc110x_default_base_freq
-extern const char cc110x_default_base_freq[3];
-#endif
+extern const char cc110x_base_freqs[3][3];
+extern const char cc110x_channel_widths[3][2];
```
PeterKietzmann: Where is this used?

drivers/cc110x/cc110x.c
```diff
@@ -38,6 +38,8 @@
 #define ENABLE_DEBUG    (0)
 #include "debug.h"
 
+static const char *cc110x_band_name[] = { "433", "868", "900" };
```
PeterKietzmann: Do we really need/want this?

drivers/cc110x/include/cc110x-defaultsettings.h
maribu: `const char cc110x_channel_widths[3][3];` is a left over I forgot to delete. (I was considering to use different channel width depending on the base band as e.g. the 900MHz LoRa channels have a spacing of 2.16 MHz instead of the 50kHz currently used. But I dropped that idea.)

`const char cc110x_base_freqs[3][3];` is used [here](https://github.com/maribu/RIOT/blob/bcaf49d037103ce758cba60945ff0cc9f6c5f5b9/drivers/cc110x/cc110x.c#L113-L120):
``` C
void cc110x_set_band(cc110x_t *dev, cc110x_band_t band)
{
#if ENABLE_DEBUG == 1


    DEBUG("cc110x_set_band(): Setting band to %sMHz\n", cc110x_band_name[band]);
#endif
    cc110x_writeburst_reg(dev, CC110X_FREQ2, cc110x_base_freqs[band], 3);
}
```

maribu:
I mis-calculated the channel filter bandwidth. It currently is 232.14kHz (MDMCFG5 = 0x7c), not 58,04kHz (which would be the result for MDMCFG5 = 0xfc).

The default value for the input filter bandwidth is 203kHz, the default for the channel bandwidth is 199.951kHz. While I'm not an expert at the physical layer, it intuitively seems to make sense to me to set the input filter bandwidth to the same value as the channel bandwidth  - as done in the default settings. Thus should make communication less prone to interference.

So I will update both input filter bandwidth (set it to 58kHz - the smallest possible value) and set the channel bandwidth to 25 kHz. Rigorous evaluation of the configuration will not be possible for now. But hopefully, I can provide fine tuned configuration values some time next year, when I hopefully have access to some larger set of IoT nodes :-)

PeterKietzmann:
> I suggest to reduce the bandwidth to 25kHz (the input filter remains at 50kHz) and check, if this has a significant impact on performance

Please do that and report the outcome.

> following the 25kHz channel spacing seems to be the best option

To me this seems like the most reasonable solution. But I haven't worked with sub-GHz or this specific device so far, which is why I triggered @Citrullin. However, you can at least provide a single commit for this change and we decide later.

>  it intuitively seems to make sense to me to set the input filter bandwidth to the same value as the channel bandwidth

Isn't there a note in the reference manual? I don't know details about filters on that device but if the input filter is placed before any carrier or phase correction, it might make sense to have a slightly higher bandwidth for that signal.

Please ping me once you think this PR is ready for testing.

maribu:
OK, apparently this is more involved than I thought it to be. If I reduce the input filter bandwidth to 56kHz, it stops working. I even set modulation to OOK, but even that won't work with an input filter of 56 kHz width. However, setting the channel width to 25 kHz does work. I set the input filter to the default in the data sheet (203kHz instead of 232kHz), which also works fine.

Citrullin:
@PeterKietzmann I am currently on a different topic, but I am happy to help. Generally speaking: I don't have that much experience in the area of wireless devices. But I will take a look into this in the near future.

maribu:
I think it is best if I first collect further information on channel layout. The choice of the base frequency and the channel width I implemented still feel quite arbitrary and I have no reasons to justify them. I will push this PR forward once I have collected this information.

Here are the references I found so far:
- [LPD433](https://en.wikipedia.org/wiki/LPD433)
- [SRD860](https://en.wikipedia.org/wiki/Short-range_device#SRD860)
- [LoRa Channels](http://www.rfwireless-world.com/Tutorials/LoRa-channels-list.html)

maribu:
I noticed that setting the input filter to 58kHz bandwidth does work on the MSB-A2. I previously missed the deviation parameter, which specifies the shift amount by which the frequency is shifted. (It is shifted from the base frequency to the left for a zero and to the right for a one, so the difference between the symbol for one and zero is twice as much.)

Setting the parameters as specified in Figure 4 of the cc1101 data sheet (except for the baud rate) does work fine for the MSB-A2. The CC1101 on the MSB-IoT however seems to perform much worse. I guess that the PCB antenna on the MSB-IoT is much less sensitive. As a result, the input filter has to be wider and the deviation parameter bigger. I believe successfully communication within a 25kHz channel is not possible on the MSB-IoT. However, in [SRD860](https://en.wikipedia.org/wiki/Short-range_device#SRD860) the frequency range form 863.0MHz to 868.0MHz can be used for 1MHz wide channels with a transmission power of up to 25mW ERP (at least this is how I understood that).

I believe a reasonable choice would be to use the [LoRa channels](http://www.rfwireless-world.com/Tutorials/LoRa-channels-list.html) CH_10_868 (865.20MHz) to CH_16_868 (867.00MHz) with the channel spacing of 300kHz. This would allow using the default input filter (203kHz bandwidth) and the default deviation (+/- 47.607kHz) with little to no interference between adjacent channels. Maybe the increased deviation (compared to the settings now) would allow higher symbol rates or longer ranges.

For the 433 MHz band using the LPD433 channels seems to be the best choice. I will check if the cheap break out boards for the CC1101 operating in the 433MHz band work reliable with the limited bandwidth. Maybe the symbol rate needs to be reduced in that band. The code will need a bit of modification to choose channel spacing, deviation and base frequency (and maybe even symbol rate) by setting the base band.

PeterKietzmann:
@maribu just ping me once you're ready.

maribu:
This will take longer than expected. I'm just checking the configuration registers against the documentation and there are some things strictly wrong. E.g. forward error correction (FEC) is enabled, but it can only be used with fixed sized packets. And the driver configures packets to be of variable length. I assume the device just ignores the FEC setting (hopefully) in this case. Also there are some values different from the default without any reason given. I personally believe that sticking with the default (and most widely used and tested) values is the best, unless a good reason exists to not do so. And this reason should be documented in the code.

Also, quite a few nice features are not used. The CC1100/CC1101 have carrier sense support, but it is turned off.

So: I would like to do more clean up on the driver and give better documentation.

Also, there are other odd things. With the current master branch version the device sends with 150 kBaud (which translates to 150 kBit/s on 2-FSK, as FEC is broken by invalid settings and manchester code is disabled). To send 8 preamble bytes, 2 sync word bytes, 1 packet length byte, 1 address byte and 2 CRC bytes (14 Bytes packet header/trailer overhead in total) the device needs 14 Byte * 8bit/Byte / 150kBit/s ~= 0.75 ms. ping6 on the MSB-A2 states round trip times of 0.3ms, which is less than sending an empty packet in *one* direction... (I believe this might be platform specific and unrelated to the driver. Maybe getting the system time on the MSB-A2 doesn't work correctly?)

maribu:
@kaspar030, @OlegHahm: You are listed as authors in `cc110x-defaultsettings.c`. The configuration register values differ quite significantly from the default values:

https://github.com/RIOT-OS/RIOT/blob/d4ba12959631d1b0947d9827ea53e0a8dbc7a584/drivers/cc110x/cc110x-defaultsettings.c#L48-L95

There is zero documentation on why those values are chosen the way they are, so I have to guess what the reasons were for not sticking with the default values here. Some of those values are also obviously wrong (e.g. FEC conflicts with variable packet length, but both are enabled). Some other could make sense. E.g. the LNA maximum gain is limited, which might be required for legal operation on some boards. (E.g. the emitted signal power (ERP) should be at most 25mW in the 868 MHz band.)

I'm considering to set all those values back to the defaults suggested in the data sheet, unless I can find a good reason for a particular configuration to differ from the default. So, do you recall any reasons for setting the configuration the way it is now?

PeterKietzmann:
@maribu as far as I know this driver hasn't been used by anyone for a long time. Furthermore, it seems to me that you (and @Citrullin) are one of the few people interested in this device and my gut feeling tells me that neither @kaspar030 nor @OlegHahm will spend time on this -> this should give you the freedom to correct, enhance, optimize, ... the driver until it is stable and efficient. I can provide you with some testing on MSBA2 boards and review, but I won't find time to discuss details about the implementation.

And in general, I'm also a fan of default values :-).

maribu:
@PeterKietzmann: Thanks for the hint. So when I do not need to fear breaking existing setups when working on the cc110x driver, I think I will do some bigger modifications to it. Features like carrier sense are not exposed and the driver was written before the GNRC network stack was present, rewriting some parts with GNRC in mind could make the code a lot easier.

Btw: The device is surprisingly capable for such an ancient transceiver. It should be able to compete with 802.15.4 devices and breakout boards are available for less than 2€. Combined with e.g. the bluepill board (again less than 2€) it seems to be the cheapest way to get started with RIOT.

Citrullin:
> Btw: The device is surprisingly capable for such an ancient transceiver. It should be able to compete with 802.15.4 devices and breakout boards are available for less than 2€. Combined with e.g. the bluepill board (again less than 2€) it seems to be the cheapest way to get started with RIOT

The reason why I bought the bluepill and the CC110x :) But Lora is of course a bit more resilient, because of the modulation. But from a getting started point of view way better. Just 5 bucks. Cheap enough to get it into a classroom ;) A topic I would like to do in the future :) The modulation is, compared to lora, simpler to explain. Add some sensor for a few bucks and you have a sensor node for under 10 Euro.

maribu:
I'm closing this PR in favor of a rewrite of the driver

maribu:
@Citrullin: Do you mind to check my [cc110x reimplementation](https://github.com/maribu/RIOT/tree/cc110x)? I assume you want to check it with your Blue Pill and an 433 MHz CC1101 break out board, right? If so: I added the correct configuration to the Blue Pill already, you just need to wire it up an add  `USEMODULE += cc1101` (and maybe also the module `cc110x_shell_command`, `cc110x_temperature` and `cc110x_csma` might be interesting.) There should be quite usable documentation, so you might want to run `make doc` and open `doc/doxygen/html/index.html` in your browser.

With the Blue Pill the following pins have to be wired together (left site CC1101, right Blue Pill):

```
VCC  <---> 3.3V
GND  <---> GND
CSN  <---> A4
SCK  <---> A5
MISO <---> A6
MOSI <---> A7
GDO0 <---> B0
GDO2 <---> B1
```

I suggest to use either `examples/gcoap` or `examples/gnrc_networking` to test the networking. If you add `cc110x_temperature` you may also want to add `saul_default`. The temperature sensor included in the CC1101 can be easily used via "saul read 1". (Saul read 0 will read the current RSSI in dBm.) On my break out boards the temperature is about 8°C to hot - this might very well be because the characteristics in the data sheet work given for 3.0V, but the Blue Pill uses 3.3V. (Just scaling the numbers up by 3.3/3.0 will result in numbers about 20°C too cold, so I kept the numbers in the data sheet.)

Thanks :-)

Citrullin:
@maribu I take a look into it at the weekend. Can I contract you directly, if I have issues with it? eMail, Chat or something? :)

maribu:
Sure. You find both my email address and my PGP key on the homepage referenced in my GitHub profile. (I think putting my email address on as few public places as possible might be a good idea, e.g. in case I will change it then I won't need to updated all outdated references...)

PR #9792, opened 2018-08-17
Makefile.include: configure '--output-sync=target'

cladmi:
### Contribution description

This should allow better output with parallel build.

### Issues/PRs references

For example the `cortexm_common_ldscript` test and also it looks like it led to misunderstanding of issues in https://github.com/RIOT-OS/RIOT/pull/9398

### DIFF
--- a/.murdock
+++ b/.murdock
@@ -155,7 +155,7 @@ compile() {
 
     # compile
     CCACHE_BASEDIR="$(pwd)" BOARD=$board TOOLCHAIN=$toolchain RIOT_CI_BUILD=1 \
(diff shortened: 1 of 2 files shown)


cladmi:
This breaks tests…

But it correctly prints this test: https://ci.riot-os.org/RIOT-OS/RIOT/9792/be74623728c229e31e3f6c68bc4ed5eb6bfdf01c/output/compile/tests/cortexm_common_ldscript/iotlab-m3.txt

I will check other outputs sync options.

cladmi:
This has been changed to only configure murdock to not use parallel output.
Some build issues were revealed by it and should be fixed before it is possible to merge this.

PR #9793, opened 2018-08-17
net stats: move layer 2 netstats from netdev driver to gnrc_netif

smlng:
### Contribution description

Currently handling and collection of layer 2 netstats is rather distributed, i.e. some is handled in the low level netdev drivers other parts are in GNRC. This PR move handling of layer 2 netstats to gnrc_netif. This is done step by step so each commit should be one after another.

### Testing

use `tests/gnrc_netif` or `examples/gnrc_networking` and transmit something, then lookup `ifconfig`.

### DIFF
--- a/cpu/esp32/esp-wifi/esp_wifi_netdev.c
+++ b/cpu/esp32/esp-wifi/esp_wifi_netdev.c
@@ -271,10 +271,6 @@ static int _esp_wifi_init(netdev_t *netdev)
 {
     DEBUG("%s: %p\n", __func__, netdev);
 
-    #ifdef MODULE_NETSTATS_L2
-    memset(&netdev->stats, 0x00, sizeof(netstats_t));
-    #endif
-
     return 0;
 }
 
@@ -315,17 +311,10 @@ static int _esp_wifi_send(netdev_t *netdev, const iolist_t *iolist)
 
     /* send the the packet to the peer(s) mac address */
     if (esp_wifi_internal_tx(ESP_IF_WIFI_STA, dev->tx_buf, dev->tx_len) == ESP_OK) {
-        #ifdef MODULE_NETSTATS_L2
-        netdev->stats.tx_success++;
-        netdev->stats.tx_bytes += dev->tx_len;
-        #endif
         netdev->event_callback(netdev, NETDEV_EVENT_TX_COMPLETE);
     }
     else {
         DEBUG("%s: sending WiFi packet failed\n", __func__);
-        #ifdef MODULE_NETSTATS_L2
-        netdev->stats.tx_failed++;
-        #endif
         ret = -EIO;
     }
 
@@ -373,11 +362,6 @@ static int _esp_wifi_recv(netdev_t *netdev, void *buf, size_t len, void *info)
         /* copy received date and reset the receive length */
         memcpy(buf, dev->rx_buf, dev->rx_len);
         dev->rx_len = 0;
-
-        #ifdef MODULE_NETSTATS_L2
-        netdev->stats.rx_count++;
-        netdev->stats.rx_bytes += size;
-        #endif
 
         mutex_unlock(&dev->dev_lock);
         return size;
--- a/sys/net/gnrc/link_layer/gomach/gomach_internal.c
+++ b/sys/net/gnrc/link_layer/gomach/gomach_internal.c
@@ -103,10 +103,10 @@ int _gnrc_gomach_transmit(gnrc_netif_t *netif, gnrc_pktsnip_t *pkt)
 #ifdef MODULE_NETSTATS_L2
     if (netif_hdr->flags &
             (GNRC_NETIF_HDR_FLAGS_BROADCAST | GNRC_NETIF_HDR_FLAGS_MULTICAST)) {
-        netif->dev->stats.tx_mcast_count++;
+        netif->stats.tx_mcast_count++;
     }
     else {
-        netif->dev->stats.tx_unicast_count++;
+        netif->stats.tx_unicast_count++;
     }
 #endif
 #ifdef MODULE_GNRC_MAC
--- a/sys/net/gnrc/link_layer/gomach/gomach_internal.c
+++ b/sys/net/gnrc/link_layer/gomach/gomach_internal.c
@@ -343,8 +343,8 @@ int gnrc_gomach_send_preamble_ack(gnrc_netif_t *netif, gnrc_gomach_packet_info_t
     assert(netif != NULL);
     assert(info != NULL);
 
+    gnrc_pktsnip_t *pkt;
     gnrc_pktsnip_t *gomach_pkt = NULL;
-    gnrc_pktsnip_t *pkt = NULL;
     gnrc_netif_hdr_t *nethdr_preamble_ack = NULL;
 
     /* Start assemble the preamble-ACK packet according to preamble packet info. */
--- a/sys/net/gnrc/netif/gnrc_netif.c
+++ b/sys/net/gnrc/netif/gnrc_netif.c
@@ -25,7 +25,7 @@
 #include "net/gnrc/ipv6/nib.h"
 #include "net/gnrc/ipv6.h"
 #endif /* MODULE_GNRC_IPV6_NIB */
-#ifdef MODULE_NETSTATS_IPV6
+#ifdef MODULE_NETSTATS
 #include "net/netstats.h"
 #endif
 #include "fmt.h"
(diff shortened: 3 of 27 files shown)


miri64:
Mh this somewhat collides with the efforts @bergzand is doing in disentangling netdev into a layered system. Can you two please sync on that.

smlng:
@miri64 we did ... via IRC, that's (also) why I asked for his review here - and yours bc you know GNRC best 😄

bergzand:
Mostly that this is the easiest solution at the moment. AFAIK there are no other network stacks using or exposing netstats, so I don't see problems with moving the implementation to GNRC specific code.

IMHO it can be easily moved to a separate netdev layer if necessary, but for now I'm happy with this as it removes another cleanup step I would have to do before being able to implement a layered netdev structure

sys/net/gnrc/netif/gnrc_netif.c
```diff
@@ -1392,25 +1407,28 @@ static void _event_cb(netdev_t *dev, netdev_event_t event)
...
-                    if (pkt) {
-                        _pass_on_packet(pkt);
-                    }
+            case NETDEV_EVENT_RX_COMPLETE:
+                pkt = netif->ops->recv(netif);
+                if (pkt) {
+#ifdef MODULE_NETSTATS_L2
+                    netif->stats.rx_count++;
+                    netif->stats.rx_bytes += pkt->size;
+#endif
```
bergzand: The current gnrc L2 specific code removes the L2 header from the `pkt`. Using the size of `pkt` here measures 14 bytes too few for Ethernet for each received packet.

I think this would have to be moved to the layer specific `recv` implementations.

smlng:
@bergzand comments addressed, maybe today HnA provides a merge window?

bergzand:
@smlng What do you think, do you want to have every device tested individually? I'm willing to trust the code as long as it compiles and with the three gnrc_netif variants tested.

sys/net/gnrc/netif/gnrc_netif.c
```diff
@@ -1356,6 +1366,11 @@ static void *_gnrc_netif_thread(void *args)
                     DEBUG("gnrc_netif: error sending packet %p (code: %u)\n",
                           msg.content.ptr, res);
                 }
+#ifdef MODULE_NETSTATS_L2
+                else {
+                    netif->stats.tx_bytes += res;
```
gebart: Was this subtle difference in behavior intentional?
I understand that it is not desirable to "waste" bytes in the netif struct to just pass the frame size from here to the TX feedback event handler, but this does change the current behavior.
number of bytes sent are increased when preloading the frame before sending, while frame count is increased upon TX_COMPLETE feedback event. This means that when using NETOPT_PRELOADING, we will get the byte count of all preloaded frames, while the number of transmissions are the number of frames actually sent. This can be both more or less than the number of preloaded frames since it is possible to transmit the same preloaded frame multiple times, as well as preloading another frame without sending the first at all.

smlng: mhm, a change in behaviour was not intended as such. And looking for instance at the cc2538rf implementation there is no difference, at lease not to me. In the `_send()` function of that driver it was

```
netdev->stats.tx_bytes += pkt_len;
[...]
return pkt_len;
```

and here 

```
res = netif->ops->send()
[...]
netdev->stats.tx_bytes += res;
```

so for this driver it should be the same, right? I'll check the others, too - where do you see the diff exactly?

gebart: Sorry for the noise, I had somehow misread the code and I thought that the rx_bytes increment used to be in the tx_exec function..

There is a time delay between the increment of stats.tx_bytes until the increment of stats.tx_complete, but it is the same in the original code as in this PR.

gebart:
esp_eth_netdev needs adaptation:

```
make: Entering directory '/tmp/dwq.0.012982471714853783/906cdea00a9ea8b7aef8f4696ae664b9/examples/gnrc_networking'
ESP32_SDK_DIR should be defined as /path/to/esp-idf directory
ESP32_SDK_DIR is set by default to /opt/esp/esp-idf
Building application "gnrc_networking" for "esp32-olimex-evb" with MCU "esp32".

esp_eth_netdev.c: In function '_esp_eth_init':
esp_eth_netdev.c:173:19: error: 'netdev_t {aka struct netdev}' has no member named 'stats'
     memset(&netdev->stats, 0, sizeof(netstats_t));
                   ^
esp_eth_netdev.c: In function '_esp_eth_send':
esp_eth_netdev.c:219:15: error: 'netdev_t {aka struct netdev}' has no member named 'stats'
         netdev->stats.tx_success++;
               ^
esp_eth_netdev.c:220:15: error: 'netdev_t {aka struct netdev}' has no member named 'stats'
         netdev->stats.tx_bytes += dev->tx_len;
               ^
esp_eth_netdev.c:226:15: error: 'netdev_t {aka struct netdev}' has no member named 'stats'
         netdev->stats.tx_failed++;
               ^
esp_eth_netdev.c: In function '_esp_eth_recv':
esp_eth_netdev.c:280:15: error: 'netdev_t {aka struct netdev}' has no member named 'stats'
         netdev->stats.rx_count++;
               ^
esp_eth_netdev.c:281:15: error: 'netdev_t {aka struct netdev}' has no member named 'stats'
         netdev->stats.rx_bytes += size;
               ^
```

smlng:
fixed esp32 and rebased

smlng:
@bergzand good to go here?

bergzand:
@smlng Minor remark above, otherwise good to go I'd say

sys/net/gnrc/link_layer/gomach/gomach_internal.c
```diff
@@ -356,7 +355,7 @@ int gnrc_gomach_send_preamble_ack(gnrc_netif_t *netif, gnrc_gomach_packet_info_t
      * This is to allow the preamble sender to deduce the exact phase of the receiver. */
     gomach_preamble_ack_hdr.phase_in_us = gnrc_gomach_phase_now(netif);
 
-    pkt = gnrc_pktbuf_add(NULL, &gomach_preamble_ack_hdr, sizeof(gomach_preamble_ack_hdr),
+    gnrc_pktsnip_t *pkt = gnrc_pktbuf_add(NULL, &gomach_preamble_ack_hdr, sizeof(gomach_preamble_ack_hdr),
```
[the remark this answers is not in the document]
smlng: In theory yes, as it is not related to netstats. However, as I touched the file the static code analyser found (that is: complained) that the previous value of `pkt` (which is `NULL` see above) was not used before being overwritten. To me this seems a bit overly correct, but in the end I need to please the CI to get the green button.

TL;DR: change is necessary to make CI happy 😦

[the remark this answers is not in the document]
bergzand: > TL;DR: change is necessary to make CI happy frowning

Makes sense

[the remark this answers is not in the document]
miri64: I prefer variable definitions to be at the top of a block so one can have an overview and consider alignment (or rather more important: resulting padding i.e. waste from misalignment) on the stack, so I would prefer to just remove the `= NULL` above and keep this line as it was. E.g. I don't know the size of `gnrc_gomach_frame_preamble_ack_t` so I would prefer it to be defined / allocated last, so the padding "hole" is at the end.

smlng:
@miri64 (re)adapted as suggested, lets see what Murdock thinks

smlng:
@miri64 can I squash?

miri64:
Yes please :-)

smlng:
done

cpu/esp32/esp-wifi/esp_wifi_netdev.c
```diff
@@ -315,17 +311,12 @@ static int _esp_wifi_send(netdev_t *netdev, const iolist_t *iolist)
 
     /* send the the packet to the peer(s) mac address */
     if (esp_wifi_internal_tx(ESP_IF_WIFI_STA, dev->tx_buf, dev->tx_len) == ESP_OK) {
-        #ifdef MODULE_NETSTATS_L2
-        netdev->stats.tx_success++;
-        netdev->stats.tx_bytes += dev->tx_len;
-        #endif
+#ifdef MODULE_NETSTATS_L2
```
miri64: Wait... This changes behavior. Before this was not in the `#ifdef` block.

smlng: mhm, right, will remove ... can I amend directly?

miri64: Yepp

gschorcht:
BTW, `esp-wifi` isn't compilable anymore with this PR. I will provide a fix.

PR #9794, opened 2018-08-17
Add limitations to doc.txt.

danpetry:
This describes the limitations of the interrupt pin being multiplexed.

### Contribution description

The pin from the LoRa Bee is multiplexed which may result in difficulties for people adding functionality later.
Not sure how likely people are to care about this, but felt like it should be recorded somewhere...


jcarrano:
@leandrolanzieri Is this accurate? Can we merge it?

leandrolanzieri:
> @leandrolanzieri Is this accurate? Can we merge it?

The comment about the FSK mode seems fine. According to the documentation DIO2 pin is used as data while in FSK Continuous mode and `RxReady` in FSK Packet mode.

It might be worth mentioning that **only** DIO0 and DIO1 are connected to the OR diodes gate.

This issue is being addressed here #9815.

boards/sensebox_samd21/doc.txt
```diff
@@ -53,4 +53,12 @@ The pin configuration for the mSD-Bee module radio on the
 *boards/sensebox/include/sdcard_spi_params.h** file assumes that it will be
 connected to the XBEE2 port. If a different configuraion is needed please
 modify this file.
- */
\ No newline at end of file
+
+#### Limitations
+There is only one interrupt pin on the XBEE headers. This means that the
+interrupt pins from the LoRa Bee are multiplexed. This has the following
```
leandrolanzieri: Maybe specify that DIO0 and DIO1 are the pins connected to an OR gate?

danpetry: I've pushed a change. Are you ok with this?

If no more comments, I'll squash and rebase.

danpetry:
@leandrolanzieri we ok to go on this? If so, I'll squash and rebase

leandrolanzieri:
Yes It's ok.
ACK.

danpetry:
Squashed and rebased

PR #9800, opened 2018-08-20
Add stm32loader tool

leandrolanzieri:
### Contribution description
This adds the stm32loader python script to flash boards with the stm32 bootloader, through the serial port. This is a modified version of [this one](https://github.com/espruino/Espruino/blob/master/scripts/stm32loader.py) that supports using the 'extended erase' command without the global erase. This is done because not all stm32 ROM bootloaders support special functions (like global erase). Instead you can specify the amount of sectors that you want to erase like this:

`./stm32loader.py -e -S -u -w -v -l 0x1ff example/main.bin`

One more chip ID has been added to the known ones.

### DIFF
--- a/dist/tools/flake8/check.sh
+++ b/dist/tools/flake8/check.sh
@@ -23,7 +23,7 @@ cd $RIOTBASE
 : "${RIOTTOOLS:=${RIOTBASE}/dist/tools}"
 . "${RIOTTOOLS}"/ci/changed_files.sh
 
-EXCLUDE='^(.+/vendor/|dist/tools/cc2538-bsl|dist/tools/mcuboot|dist/tools/uhcpd)'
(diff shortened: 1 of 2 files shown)


MrKevinWeiss:
Looks like a useful tool.  If anyone thinks it should reside somewhere else (ie, a patch vs the full code) speak now otherwise I will merge.

MrKevinWeiss:
@leandrolanzieri maybe make it clear in the contribution description it is needed for #9816.

leandrolanzieri:
Yes sorry. @kYc0o was introduced to flash boards that use the stm32loader like the Lobraro Lorabox here #9816. It is used as the flasher tool

leandrolanzieri:
Closing this as stm32loader is now included in #9816.

PR #9804, opened 2018-08-20
stm32_common/i2c: fix wrong speed parameters for stm32l4 and stm32f7

MichelRottleuthner:
### Contribution description

The common i2c driver shares speed configuration data (scll / sclh, etc.) among multiple MCUs. The correct values for these parameters depend on the i2c clock that can be configured to various sources. For the L4 and F7 the configuration was based on the examples for 48 Mhz clock but that isn't correct. The clock source is changed as follows (# means old config * means new):
```
stm32-f722ze
HSI     16 MHz *
SYSCLK 216 MHz
PCLK    54 Mhz #

stm32-l475vg
HSI     16 MHz *
SYSCLK  80 MHz
PCLK    20 Mhz #
```

I used the HSI because that doesn't depend on other configurations on these MCUs, and this way the configuration for L4 and F7 can still be shared.

I haven't checked yet if the other MCUs (F3 / L0) are also affected.

### Issues/PRs references
none

### DIFF
--- a/boards/b-l475e-iot01a/include/periph_conf.h
+++ b/boards/b-l475e-iot01a/include/periph_conf.h
@@ -203,6 +203,7 @@ static const i2c_conf_t i2c_config[] = {
         .sda_af         = GPIO_AF4,
         .bus            = APB1,
         .rcc_mask       = RCC_APB1ENR1_I2C1EN,
+        .rcc_sw_mask    = RCC_CCIPR_I2C1SEL_1, /* HSI (16 MHz) */
         .irqn           = I2C1_ER_IRQn,
     },
     {
@@ -214,6 +215,7 @@ static const i2c_conf_t i2c_config[] = {
         .sda_af         = GPIO_AF4,
         .bus            = APB1,
         .rcc_mask       = RCC_APB1ENR1_I2C2EN,
+        .rcc_sw_mask    = RCC_CCIPR_I2C2SEL_1, /* HSI (16 MHz) */
         .irqn           = I2C2_ER_IRQn,
     },
 };
--- a/boards/nucleo-f722ze/include/periph_conf.h
+++ b/boards/nucleo-f722ze/include/periph_conf.h
@@ -150,6 +150,7 @@ static const i2c_conf_t i2c_config[] = {
         .sda_af         = GPIO_AF4,
         .bus            = APB1,
         .rcc_mask       = RCC_APB1ENR_I2C1EN,
+        .rcc_sw_mask    = RCC_DCKCFGR2_I2C1SEL_1, /* HSI (16 MHz) */
         .irqn           = I2C1_ER_IRQn,
     }
 };
--- a/cpu/stm32_common/include/cpu_conf_stm32_common.h
+++ b/cpu/stm32_common/include/cpu_conf_stm32_common.h
@@ -31,6 +31,14 @@ extern "C" {
     defined(CPU_FAM_STM32F7) || defined(CPU_FAM_STM32L0) || \
     defined(CPU_FAM_STM32L4)
 
+#if defined(CPU_FAM_STM32F0) || defined(CPU_FAM_STM32F3)
+#define I2C_CLOCK_SRC_REG (RCC->CFGR3)
+#elif defined(CPU_FAM_STM32L4)
+#define I2C_CLOCK_SRC_REG (RCC->CCIPR)
+#elif defined(CPU_FAM_STM32F7)
+#define I2C_CLOCK_SRC_REG (RCC->DCKCFGR2)
+#endif
+
 /**
  * @brief   Timing register settings
  *
--- a/cpu/stm32_common/include/cpu_conf_stm32_common.h
+++ b/cpu/stm32_common/include/cpu_conf_stm32_common.h
@@ -31,14 +31,6 @@ extern "C" {
     defined(CPU_FAM_STM32F7) || defined(CPU_FAM_STM32L0) || \
     defined(CPU_FAM_STM32L4)
 
-#if defined(CPU_FAM_STM32F0) || defined(CPU_FAM_STM32F3)
-#define I2C_CLOCK_SRC_REG (RCC->CFGR3)
-#elif defined(CPU_FAM_STM32L4)
-#define I2C_CLOCK_SRC_REG (RCC->CCIPR)
-#elif defined(CPU_FAM_STM32F7)
-#define I2C_CLOCK_SRC_REG (RCC->DCKCFGR2)
-#endif
-
 /**
  * @brief   Timing register settings
  *
--- a/cpu/stm32_common/include/cpu_conf_stm32_common.h
+++ b/cpu/stm32_common/include/cpu_conf_stm32_common.h
@@ -45,8 +45,7 @@ extern "C" {
  * @ref i2c_timing_param_t
  */
 static const i2c_timing_param_t timing_params[] = {
-#if defined(CPU_FAM_STM32F0) || defined(CPU_FAM_STM32L4) || \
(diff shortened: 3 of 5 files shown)


MichelRottleuthner:
@MrKevinWeiss would be nice to have a feature for checking i2c clocks on the raspi/bluepill tester in the future :)

cpu/stm32_common/include/cpu_conf_stm32_common.h
```diff
@@ -31,14 +31,21 @@ extern "C" {
     defined(CPU_FAM_STM32F7) || defined(CPU_FAM_STM32L0) || \
     defined(CPU_FAM_STM32L4)
 
+#if defined(CPU_FAM_STM32F0) || defined(CPU_FAM_STM32F3)
```
aabadie: Those defines should go in `i2c_1.c`

MichelRottleuthner:
Thanks for testing @aabadie. I moved the define as requested and enabled HSI clock. But now that I think about it, I'm not sure if we should do it that way. Which clock source is used is defined by the 'rcc_sw_mask' value (btw. this name should probably adapted) and thus isn't hardcoded for all L4 and F7 MCUs. Adding a runtime condition for the L4/F7, shouldn't hurt too much because it is only done on init, what do you think?

MichelRottleuthner:
Looks like for the F3 the timing config is based on the 8 Mhz example.
See Table 147 (Examples of timings settings for f_I2CCLK = 8 MHz) in
[RM0316 rev 8, page 849](https://www.st.com/resource/en/reference_manual/dm00043574.pdf#page=849).
But actually SYSCLK (72 MHz) [is used as clock source](https://github.com/RIOT-OS/RIOT/blob/master/boards/nucleo-f303re/include/periph_conf.h#L218), see section 9.4.13 (RCC_CFGR3) in
[RM0316 rev 8, page 164](https://www.st.com/resource/en/reference_manual/dm00043574.pdf#page=164).
I wonder why this configuration actually works at all o.O - am I overlooking something?

Based on the reference manual the L0 configuration also seems to be a bit off. At least the `FAST_PLUS` mode looks [like it's configured to fast](https://github.com/RIOT-OS/RIOT/blob/master/cpu/stm32_common/include/cpu_conf_stm32_common.h#L102) (~1.8MHz). See [t_SCL formula in reference manual](https://www.st.com/resource/en/reference_manual/dm00095744.pdf#page=708) I didn't do measurements to confirm this though.

aabadie:
> Adding a runtime condition for the L4/F7, shouldn't hurt too much because it is only done on init, what do you think?

That would be my preferred solution too

aabadie:
@MichelRottleuthner, now I don't really like using hsi as clock source for L4/F7.
My main question is: what part of the code is responsible of enabling it ? Maybe it shouldn't be done by the i2c peripheral driver or, on the contrary, it should be enabled when acquire the bus and disable it when releasing the bus. But this second approach may break other peripherals that also use it.

Another solution could be that the clock source should always be sysclk, and the board configuration should provide the timing values based on sysclk (defined at board level). I'm thinking of a tool to generate these timing values (for each supported I2C speeds) similar to what is done for SPI.

MichelRottleuthner:
After thinking about this again I don't really like the idea of using sysclock as source.
Using another (sysclock) frequency will then effectively change the i2c clock if i2c config is not adapted.
With HSI this won't be a problem because it's fixed.
Sure we can say that the developer is responsible to make sure that nothing breaks. But if we can avoid dependencies between configurations where changing one breaks another, IMO we should prefer that.
In the future this should also ease integrating DVFS functions which would be really nice to lower power consumption.

aabadie:
Ok, HSI is probably simpler and more correct (in terms of timings configuration). Normally this PR is working but I'd like to test it again before merging.

MrKevinWeiss:
I can test Monday with a scope if you can't do it by then.

MrKevinWeiss:
BOARD  | SELECTED_SPEED  | MEASURED_SCK 
--|---|--
 nucleo-f410rb | I2C_SPEED_NORMAL  |  44.6kHz
 nucleo-f410rb | I2C_SPEED_FAST  |  119kHz
 nucleo-l073rz | I2C_SPEED_NORMAL  |  80kHz
 nucleo-l073rz | I2C_SPEED_FAST  |  179kHz
 nucleo-l073rz | I2C_SPEED_FAST_PLUS  |  226kHz

aabadie:
@MrKevinWeiss, it would be great if you could also test on f0, f3, f7 and l4. F4 is not concerned by this change.

MrKevinWeiss:
Ya, @MichelRottleuthner pointed that out.  I need to find the boards (or order them)

MrKevinWeiss:
Those were the only ones that I had on hand.  I will do a sweep tomorrow.

MichelRottleuthner:
>F4 is not concerned by this change

right, but the measurements show that we should take a look at F4 too

boards/b-l475e-iot01a/include/periph_conf.h
```diff
@@ -203,6 +203,7 @@ static const i2c_conf_t i2c_config[] = {
         .sda_af         = GPIO_AF4,
         .bus            = APB1,
         .rcc_mask       = RCC_APB1ENR1_I2C1EN,
+        .rcc_sw_mask    = RCC_CCIPR_I2C1SEL_1, /* HSI (16 MHz) */
```
aabadie: With #9728 being merged, this change also needs to be applied to `boards/common/stm32/cfg_i2c1_pb8_pb9.h`

boards/nucleo-f722ze/include/periph_conf.h
```diff
@@ -150,6 +150,7 @@ static const i2c_conf_t i2c_config[] = {
         .sda_af         = GPIO_AF4,
         .bus            = APB1,
         .rcc_mask       = RCC_APB1ENR_I2C1EN,
+        .rcc_sw_mask    = RCC_DCKCFGR2_I2C1SEL_1, /* HSI (16 MHz) */
```
aabadie: This conflicts now that #9728 is merged, but this change can also just be applied to boards/common/stm32/cfg_i2c1_pb8_pb9.h.

tcschmidt:
Punch this into the next release @MrKevinWeiss ?

MrKevinWeiss:
Hmm, PHiLIP now supports i2c timing information... I could write a test that could be used.  Maybe we can get it done for the 2019.07 release.  I'll ask @MichelRottleuthner .

tcschmidt:
Ready to finish this @MichelRottleuthner @MrKevinWeiss @aabadie ?

aabadie:
Still needs a rebase.

MrKevinWeiss:
> @MrKevinWeiss would be nice to have a feature for checking i2c clocks on the raspi/bluepill tester in the future :)

I agree, in fact I believe philip supports clock estimates... just a matter of writing a test.  Oh also exposing the clock speed in RIOT.

MrKevinWeiss:
Oh jeez, really forgot about this one 😞

fjmolinas:
> Oh jeez, really forgot about this one disappointed

@MrKevinWeiss @aabadie do you think you could take over this one?

MrKevinWeiss:
/remind me to check this PR on November 24 1, 2021

aabadie:
With the clock rework, that would mean enabling HSI16 when I2C is used. There's some logic in the clock configuration to enable that.
Another option is to compute the parameters based on the speed of the clock source.

aabadie:
> /remind me to check this PR on November 24 1, 2021

too late...

I opened #17324 to take over that work.

MrKevinWeiss:
I wish I we had a remind me bot... I remember that day, I barely had time to breathe. I should have brought the board back with me.

PR #9815, opened 2018-08-22
drivers/sx127x: Add initial support for multi interrupt pins

leandrolanzieri:
### Contribution description
[Some boards](https://learn.watterott.com/bee-modules/lora/) that breakout the sx127x radios unify the interrupt pins into a single one. This is an initial support for those cases. When the *multi dio* pin triggers an interrupt the registers on the sx127x are checked to identify the interrupt cause, and the appropriate isr is executed. 

Currently this fix is needed by the 'SenseBox with SAMD21' board.

### Usage
To enable the "multi dio" pin the macro `SX127X_USE_DIO_MULTI` must be defined. This can be done by adding `CFLAGS += -DSX127X_USE_DIO_MULTI` to the makefile of the board.

### DIFF
--- a/boards/sensebox_samd21/include/board.h
+++ b/boards/sensebox_samd21/include/board.h
@@ -129,18 +129,17 @@ extern "C" {
 
 #define SX127X_PARAM_RESET                  GPIO_UNDEF
 
-#define SX127X_PARAM_DIOMULTI               XBEE1_INT_PIN       /* D24 */
+#define SX127X_PARAM_DIO0                   GPIO_UNDEF
+
+#define SX127X_PARAM_DIO1                   GPIO_UNDEF
+
+#define SX127X_PARAM_DIO2                   GPIO_UNDEF
+
+#define SX127X_PARAM_DIO3                   GPIO_UNDEF
+
+#define SX127X_PARAM_DIO_MULTI              XBEE1_INT_PIN       /* D24 */
 
 #define SX127X_PARAM_PASELECT               (SX127X_PA_BOOST)
-
-#define SX127X_PARAMS                     { .spi       = SX127X_PARAM_SPI,     \
-                                            .nss_pin   = SX127X_PARAM_SPI_NSS, \
-                                            .reset_pin = SX127X_PARAM_RESET,   \
-                                            .dio0_pin  = SX127X_PARAM_DIOMULTI,\
-                                            .dio1_pin  = SX127X_PARAM_DIO1,    \
-                                            .dio2_pin  = SX127X_PARAM_DIO2,    \
-                                            .dio3_pin  = SX127X_PARAM_DIO3,    \
-                                            .paselect  = SX127X_PARAM_PASELECT }
 /** @} */
 
 /**
--- a/drivers/include/sx127x.h
+++ b/drivers/include/sx127x.h
@@ -92,6 +92,9 @@ extern "C" {
 #define SX127X_IRQ_DIO3                  (1<<3)  /**< DIO3 IRQ */
 #define SX127X_IRQ_DIO4                  (1<<4)  /**< DIO4 IRQ */
 #define SX127X_IRQ_DIO5                  (1<<5)  /**< DIO5 IRQ */
+#ifdef SX127X_USE_DIO_MULTI
+#define SX127X_IRQ_DIO_MULTI             (1<<6)  /**< DIO MULTI IRQ */
+#endif
 /** @} */
 
 /**
@@ -208,6 +211,9 @@ typedef struct {
     gpio_t dio3_pin;                   /**< Interrupt line DIO3 (CAD done) */
     gpio_t dio4_pin;                   /**< Interrupt line DIO4 (not used) */
     gpio_t dio5_pin;                   /**< Interrupt line DIO5 (not used) */
+#ifdef SX127X_USE_DIO_MULTI
+    gpio_t dio_multi_pin;              /**< Interrupt line for multiple IRQs */
+#endif
     uint8_t paselect;                  /**< Power amplifier mode (RFO or PABOOST) */
 } sx127x_params_t;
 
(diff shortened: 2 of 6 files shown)


boards/sensebox_samd21/include/board.h
```diff
@@ -123,24 +123,33 @@ extern "C" {
...
-                                            .paselect  = SX127X_PARAM_PASELECT }
+#define SX127X_PARAM_DIO2           GPIO_UNDEF
+
+#define SX127X_PARAM_DIO3           GPIO_UNDEF
+
+#define SX127X_PARAM_DIOMULTI       XBEE1_INT_PIN       /* D24 */
+
+#define SX127X_PARAM_PASELECT       (SX127X_PA_RFO)
+
+#define SX127X_PARAMS               { .spi       = SX127X_PARAM_SPI,           \
```
aabadie: This define is not needed since all `SX127X_PARAMS_*` are already defined and will be used in the `sx127x_params.h` file.

drivers/sx127x/include/sx127x_params.h
```diff
@@ -33,46 +33,62 @@ extern "C" {
  * @{
  */
 #ifndef SX127X_PARAM_SPI
-#define SX127X_PARAM_SPI                    (SPI_DEV(0))
+#define SX127X_PARAM_SPI          (SPI_DEV(0))
```
aabadie: unrelated alignment change, please revert (here and below)

drivers/include/sx127x.h
```diff
@@ -208,6 +211,9 @@ typedef struct {
     gpio_t dio3_pin;                   /**< Interrupt line DIO3 (CAD done) */
     gpio_t dio4_pin;                   /**< Interrupt line DIO4 (not used) */
     gpio_t dio5_pin;                   /**< Interrupt line DIO5 (not used) */
+#ifdef SX127X_USE_DIO_MULTI
+    gpio_t dio_multi_pin;              /**< Interrupt line for muliple IRQs */
```
aabadie: s/muliple/multiple/

drivers/sx127x/include/sx127x_params.h
```diff
@@ -33,46 +33,62 @@ extern "C" {
...
-#define SX127X_PARAMS                       { .spi       = SX127X_PARAM_SPI,     \
-                                              .nss_pin   = SX127X_PARAM_SPI_NSS, \
-                                              .reset_pin = SX127X_PARAM_RESET,   \
-                                              .dio0_pin  = SX127X_PARAM_DIO0,    \
-                                              .dio1_pin  = SX127X_PARAM_DIO1,    \
-                                              .dio2_pin  = SX127X_PARAM_DIO2,    \
-                                              .dio3_pin  = SX127X_PARAM_DIO3,    \
-                                              .paselect  = SX127X_PARAM_PASELECT }
+#ifdef SX127X_USE_DIO_MULTI
+    #define SX127X_PARAMS             { .spi       = SX127X_PARAM_SPI,          \
```
aabadie: no need to indent this `#define`, same below

drivers/sx127x/sx127x.c
```diff
@@ -207,35 +219,81 @@ static void sx127x_on_dio3_isr(void *arg)
...
+    if (dev->params.dio0_pin != GPIO_UNDEF) {
+        res = gpio_init_int(dev->params.dio0_pin, GPIO_IN, GPIO_RISING,
+                                sx127x_on_dio0_isr, dev);
+        if (res < 0) {
+            DEBUG("[sx127x] error: failed to initialize DIO0 pin\n");
+            return res;
+        }
+    }
+    else {
+        DEBUG("[sx127x] error: no DIO0 pin defined\n");
```
aabadie: is it still an error ?

aabadie: I would do this differently:
```c
#ifndef SX127X_USE_DIO_MULTI
   /* Check if DIO0 pin is defined */
    if (dev->params.dio0_pin != GPIO_UNDEF) {
        res = gpio_init_int(dev->params.dio0_pin, GPIO_IN, GPIO_RISING,
                            sx127x_on_dio0_isr, dev);
        if (res < 0) {
            DEBUG("[sx127x] error: failed to initialize DIO0 pin\n");
            return res;
        }
    }

    /* check if DIO1 pin is defined */
    if (dev->params.dio1_pin != GPIO_UNDEF) {
        res = gpio_init_int(dev->params.dio1_pin, GPIO_IN, GPIO_RISING,
                            sx127x_on_dio1_isr, dev);
        if (res < 0) {
            DEBUG("[sx127x] error: failed to initialize DIO1 pin\n");
            return res;
        }
    }
#else
/* your new multi dio stuff here */
#endif
```

drivers/sx127x/sx127x_netdev.c
```diff
@@ -240,6 +240,36 @@ static void _isr(netdev_t *netdev)
     sx127x_t *dev = (sx127x_t *) netdev;
 
     uint8_t irq = dev->irq;
+
+    #ifdef SX127X_USE_DIO_MULTI
```
aabadie: No need to indent this `#ifdef`

drivers/sx127x/sx127x.c
leandrolanzieri: I did it that way because with this radio I think you at least need the DIO0 defined. So in the case that DIO0 is not defined and you are not using the "multi" mode it should fail. If you are using the "multi" mode and no DIO0 is defined, it would check if the MULTI_DIO pin is defined.

leandrolanzieri: @aabadie Did you have a chance to look at this?

aabadie: > Did you have a chance to look at this?

Yes, and I still want this to be changed since it can be done simpler. I edited my code snippet above with what I have in mind.
I also think the `sx127x_on_dioX_isr` static functions could be excluded if `SX127X_USE_DIO_MULTI` is defined. Can you check if that save a few bytes ?

leandrolanzieri: Modified so it would only initialize the Multi dio pin if `SX127X_USE_DIO_MULTI` is defined. Also excluded the `sx127x_on_diox_isr` functions.

leandrolanzieri:
@aabadie ping?

leandrolanzieri:
Rebased and squashed.

aabadie:
Thanks @leandrolanzieri !

PR #9819, opened 2018-08-22
tools/doccheck: Detect when `make doc` fails to run.

jcarrano:
The previous doccheck would give a false negative when doxygen does not even run (because of misconfiguration). Guess what? Doxygen was never running on Travis! That means our docchecks in the static tests were a lie.

### Issues/PRs references

For an example of why this is a problem, see #9818.

~Depends on #10237 (if we enable the check before fixing travis then everything will fail,)~
Depends on ~#10251~ (the old version of Doxygen currently in travis reports a lot of errors).

### DIFF
--- a/dist/tools/doccheck/check.sh
+++ b/dist/tools/doccheck/check.sh
@@ -20,15 +20,20 @@ else
     CRESET=
 fi
 
-ERRORS=$(make -C "${RIOTBASE}" doc 2>&1 | \
-            grep '.*warning' | \
-            sed "s#${PWD}/\([^:]*\)#\1#g")
+DOXY_OUTPUT=$(make -C "${RIOTBASE}" doc 2>&1)
+DOXY_ERRCODE=$?
 
-if [ -n "${ERRORS}" ]
-then
-    echo -e "${CERROR}ERROR: Doxygen generates the following warnings:${CRESET}"
-    echo "${ERRORS}"
+if [ "${DOXY_ERRCODE}" -ne 0 ] ; then
+    echo "'make doc' exited with non-zero code (${DOXY_ERRCODE})"
+    echo "${DOXY_OUTPUT}"
     exit 2
+else
+    ERRORS=$(echo "${DOXY_OUTPUT}" | grep '.*warning' | sed "s#${PWD}/\([^:]*\)#\1#g")
+    if [ -n "${ERRORS}" ] ; then
+        echo -e "${CERROR}ERROR: Doxygen generates the following warnings:${CRESET}"
+        echo "${ERRORS}"
+        exit 2
+    fi
 fi
 
 exclude_filter() {


dist/tools/doccheck/check.sh
```diff
@@ -10,6 +10,11 @@
 
 RIOTBASE="$(cd $(dirname $0)/../../..; pwd)"
 
+die() {
+   echo "warning: Failed to build run doxygen"
```
jia200x: these error messages can come from the Doxygen makefile. E.g there might be a "doxygen not found" or a "git-lfs" not installed, they are 2 different errors

jia200x:
besides the comment I just made, the script is running as expected. Great!

jia200x:
(Travis fails, the error seems to be triggered there)

jcarrano:
> (Travis fails, the error seems to be triggered there)

Weird. It works on my machine. Could it be that `make doc` was failing all this time and we never detected it?

jia200x:
> Weird. It works on my machine. Could it be that make doc was failing (in travis) all this time and we never detected it?

If I remember correctly I have seen Doxygen errors, not sure where (Travis CI or Murdock) . I will try the Docker image to see if it works.

dist/tools/doccheck/check.sh
jcarrano: The idea of this PR was to catch any non-zero exit code from `make doc`. Maybe I can change the message to "make doc exited with non-zero code" to make it explicit.
Afterwards we could add filters to point at the cause of the problem.

jcarrano:
@jia200x Is this PR still relevant or has it been solved in another one?

jcarrano:
Update: the problem was Travis and it's outdated lesscpy command. I did another PR to fix it.

aabadie:
How about moving travis configuration to xenial (instead of trusty) ?

cladmi:
@aabadie why not updating it directly to the new bionic docker image after the release when we use it in `murdock` ?

aabadie:
> why not updating it directly to the new bionic docker image after the release when we use it in murdock ?

Easy, have a look at #8729. But it requires to move to CircleCI => which provides a nice container environment.

jcarrano:
Can we first update and then move to CircleCI, I'm sure it won't be without inconvenients. Then we would have the problems that come from updating ubuntu and changing CI all at the same time.

cladmi:
Can we not use our image in `travis` ?

But sure using the latest version on `travis` would be better anyway. What is missing to go for it ?

aabadie:
> Can we not use our image in travis ?

Apparently yes, see https://docs.travis-ci.com/user/docker/
But I find it less handy than the circle ci config. Note that on CircleCI there's the possibility to store builds artifacts and this is what is used to have an overview of the generated doc in #8729

cladmi:
@aabadie It is still two different tasks, use travis with an updated image even if `ugly` and migrating should be justified by real benefits and could be a separate thing. Fix first, migrating is another step.

Now there is a real issue with `travis` with the `doxygen` version as shown in https://github.com/RIOT-OS/RIOT/pull/10237

jcarrano:
@cladmi 

> use travis with an updated image even

d'oh! It seems that is not possible. @aabadie is right, we need to move out of travis, see this:

https://github.com/travis-ci/travis-ci/issues/9928

Travis is stuck with Trusty. Let them rot!!!!

cladmi:
@jcarrano please rebase now that https://github.com/RIOT-OS/RIOT/pull/10251 is merged.
You can also directly squash.

cladmi:
@jia200x It's good for me, I let you have the final word.

PR #9824, opened 2018-08-22
 pkg/jerryscript: build into '$(BINDIR)' and cleanups

cladmi:
### Contribution description

Build into the BINDIR directory instead of the source repository.
This makes 'clean' work as expected without other intervention.

Also it goes in the direction of having the package source repository
board independent.

I also included cleanup for `jerryscript` makefiles self described in the commits messages.

### Testing procedure

Clean actually now cleans.
```
rm -rf bin; make; make clean
# Then check
ls bin; git -C bin/pkg/native/jerryscript/ status --porcelain
pkg
?? .git-downloaded
```

In master we had
```
ls bin; git -C bin/pkg/native/jerryscript/ status --porcelain
pkg
?? .git-downloaded
?? riot/
```

Where riot contained all the build files.

### Issues/PRs references

Part of https://github.com/RIOT-OS/RIOT/pull/9820

### DIFF
--- a/pkg/jerryscript/Makefile.jerryscript
+++ b/pkg/jerryscript/Makefile.jerryscript
@@ -1,4 +1,4 @@
-BUILD_DIR  ?= $(CURDIR)/riot
+BUILD_DIR  ?= $(BINDIR)/jerryscript
 
 JERRYHEAP  ?= 16
 
--- a/pkg/jerryscript/Makefile.jerryscript
+++ b/pkg/jerryscript/Makefile.jerryscript
@@ -4,9 +4,9 @@ JERRYHEAP  ?= 16
 
 EXT_CFLAGS :=-D__TARGET_RIOT
 
-.PHONY: libjerry riot-jerry flash clean
+.PHONY: libjerry
 
-# all: libjerry riot-jerry
+all: libjerry
 
 libjerry:
 	mkdir -p $(BUILD_DIR)
--- a/pkg/jerryscript/Makefile
+++ b/pkg/jerryscript/Makefile
@@ -9,7 +9,7 @@ CFLAGS += -Wno-implicit-fallthrough
 
 # disable warnings when compiling with LLVM for board native
 ifeq ($(TOOLCHAIN)_$(BOARD),llvm_native)
-export CFLAGS += -Wno-macro-redefined -Wno-gnu-folding-constant
+  CFLAGS += -Wno-macro-redefined -Wno-gnu-folding-constant
 endif
 
 all: git-download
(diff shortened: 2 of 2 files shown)


cladmi:
@cgundogan any idea why changing this `-H` to an absolute directory is required ?

cladmi:
It does not fail on my computer and with docker, so will retry without.

cladmi:
Non parallel build works

```
BOARDS="slstk3402a" make -C examples/javascript/ TOOLCHAIN=llvm buildtest 'BUILDTEST_MAKE_REDIRECT=>/dev/null'  NPROC=1
```
Parallel does not
```
BOARDS="slstk3402a" make -C examples/javascript/ TOOLCHAIN=llvm buildtest 'BUILDTEST_MAKE_REDIRECT=>/dev/null'  NPROC=
```

Now it is reproducible I can investigate.

cladmi:
Found that it is because I replaced the default target by `all` to correctly use what was in `Makefile.base`. But that is actually causing problems as it overwrites `jerryscript.a`.

I think it should not even include `Makefile.base` as it is not used.

pkg/jerryscript/Makefile.jerryscript
```diff
@@ -1,15 +1,14 @@
-BUILD_DIR  ?= $(CURDIR)/riot
+BUILD_DIR  ?= $(BINDIR)/jerryscript
```
smlng: why not use `BUILD_DIR  ?= $(PKG_BUILDDIR)` instead?

cladmi: First it is not defined there, and also it is the directory where the package is cloned.

As described in the commit message and the title of this PR. It is the goal on this one to move toward out of source build, and all normal modules in RIOT build in `$(BINDIR)/module_name`.

With out of source build, the package could be cloned only once for all platforms (20M * 120 boards == a lot).

smlng: ok, then we definitely need to work on the naming of RIOT directories. Because if `PKG_BUILDDIR` is the directory where the package is cloned and *not* build, its very confusing - the name is why I suggested to use that for `BUILD_DIR`.

cladmi: Some of the packages are still build into there when they use their own build system.

aabadie:
@cladmi, what is the status of this PR ? It would at least require a rebase

cladmi:
I squashed to remove the commits that were more on the finding what to do path.

There are now the 5 first commits (check the order locally and not from github as it uses `author date`) that are for cleaning up and style changes.
And the 6th commit that changes the build directory to `BINDIR/jerryscript`.

I could split if wanted.

The test command using `llvm` does not work anymore as llvm is blacklisted with llvm, so just remove the toolchain for now.

PR #9838, opened 2018-08-24
ds323x: Minimal driver for DS323x extremely accurate RTC

gebart:
### Contribution description

A simple device driver for enabling the square wave output on the DS3234 SPI RTC from Maxim.
This driver currently only supports the SPI connected DS3234, and only for initializing 1 Hz square wave output on the SQW pin, nothing else.

The intended use case is to use this for calibrating internal oscillators on different boards by connecting the square wave output to a GPIO interrupt pin, and through those measurements achieve better accuracy from cheaper timers.

The register map is the same for all DS323x, but only the DS3234 is implemented here.

### Testing procedure

External hardware: https://www.sparkfun.com/products/10160
Required tools: Oscilloscope or logic analyzer

Configure the device parameters in ds3234_params.h, build and flash tests/driver_ds3234. Examine the SQW pin on the RTC using an oscilloscope or logic analyzer, see a 1 Hz square wave.

### Issues/PRs references

nothing yet.

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -124,6 +124,10 @@ ifneq (,$(filter ds1307,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter ds3234,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_spi
+endif
+
 ifneq (,$(filter dsp0401,$(USEMODULE)))
   USEMODULE += xtimer
   FEATURES_REQUIRED += periph_gpio
--- a/drivers/include/ds3234.h
+++ b/drivers/include/ds3234.h
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2018 SKF AB
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_ds3234  DS3234 Extremely Accurate SPI RTC
+
+ * @brief       Driver for Maxim DS3234 Extremely Accurate SPI Bus RTC with
+ *              Integrated Crystal and SRAM
+
+ * @{
+ *
+ * @file
+ * @brief       DS3234 device driver
+ *
+ * @author      Joakim Nohlgård <joakim.nohlgard@eistec.se>
+ */
+
+#ifndef DS3234_H
+#define DS3234_H
+
+#include <periph/gpio.h>
+#include <periph/spi.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Parameters for the DS3234 device driver
+ */
+typedef struct {
+    spi_t spi;              /**< SPI bus the sensor is connected to */
+    spi_clk_t clk;          /**< SPI bus clock speed */
+    gpio_t cs;              /**< CS pin GPIO handle */
+} ds3234_params_t;
+
+/**
+ * @brief   Return codes for the DS3234 device driver
+ */
+enum ds3234_return_codes {
+    DS3234_OK = 0,
+    DS3234_NO_DEV = 1,
+    DS3234_NO_SPI = 2
+};
+
+/**
+ * @brief   Initialize the DS3234 RTC as a PPS device
+ *
+ * This will enable square wave output on the SQW pin at 1 Hz
+ *
+ * @param  params     DS3234 circuit parameters
+ *
+ * @return DS3234_OK        on success
+ * @return DS3234_NO_SPI    if cannot initiate SPI
+ * @return DS3234_NO_DEV    if the device is not found on the bus
+ */
+int ds3234_pps_init(const ds3234_params_t *params);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* DS3234_H */
+/** @} */
--- a/drivers/Makefile.include
+++ b/drivers/Makefile.include
@@ -66,6 +66,10 @@ ifneq (,$(filter ds1307,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/drivers/ds1307/include
 endif
 
+ifneq (,$(filter ds3234,$(USEMODULE)))
+  USEMODULE_INCLUDES += $(RIOTBASE)/drivers/ds3234/include
+endif
+
 ifneq (,$(filter dsp0401,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/drivers/dsp0401/include
 endif
--- a/drivers/ds3234/Makefile
+++ b/drivers/ds3234/Makefile
@@ -1,0 +1,1 @@
+include $(RIOTBASE)/Makefile.base
--- a/drivers/ds3234/ds3234.c
+++ b/drivers/ds3234/ds3234.c
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2018 Eistec AB
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ *
+ */
+
+/**
+ * @ingroup     drivers_ds3234
+ * @{
+ *
+ * @file
+ * @brief       Driver for the DS3234 Extremely Accurate SPI Bus RTC with
+ *              Integrated Crystal and SRAM, from Maxim
+ *
+ * @author      Joakim Nohlgård <joakim.nohlgard@eistec.se>
+ * @}
+ */
+
+#include <stdint.h>
+#include <errno.h>
+#include "ds3234.h"
+#include "ds3234_regs.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+/* SPI command byte parameters */
+#define DS3234_CMD_READ         (0x00)
+#define DS3234_CMD_WRITE        (0x80)
+
+
+/**
+ * @brief Read one or more registers from the sensor
+ *
+ * @param[in]  dev    device descriptor
+ * @param[in]  addr   register address
+ * @param[in]  len    number of bytes to read
+ * @param[out] buf    destination buffer
+ */
+static void ds3234_read_reg(const ds3234_params_t *dev, uint8_t addr, size_t len, uint8_t *buf)
+{
+    uint8_t command = DS3234_CMD_READ | addr;
+    /* Acquire exclusive access to the bus. */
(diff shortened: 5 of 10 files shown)


drivers/Makefile.dep
```diff
@@ -100,6 +100,12 @@ ifneq (,$(filter ds1307,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter ds3234,$(USEMODULE)))
+  USEMODULE += ds323x
+  USEMODULE += ds323x_ds3234
```
PeterKietzmann: This name seems a bit weird to me. Is it even generated or used somewhere?

drivers/ds323x/Makefile
```diff
@@ -0,0 +1,4 @@
+# enable submodules
+SUBMODULES := 1
```
PeterKietzmann: So far you just implemented *ds3234.c*. Do we actually need the SUBMODULES now?

drivers/ds323x/ds3234.c
```diff
@@ -0,0 +1,99 @@
...
+
+int ds3234_pps_init(const ds3234_params_t *dev)
+{
+    /* initialize CS pin */
+    int res = spi_init_cs(dev->spi, dev->cs);
+    if (res < 0) {
+        return -EIO;
+    }
+
+    if (ENABLE_DEBUG) {
```
PeterKietzmann: Why not `#ifdef`?

drivers/ds323x/ds3234.c
```diff
@@ -0,0 +1,99 @@
...
+            ds3234_read_reg(dev, k, 1, &dbg_reg);
+            print_byte_hex(k);
+            print_str(": ");
+            print_byte_hex(dbg_reg);
+            print_str("\n");
+        }
+    }
+    uint8_t reg = 0;
+    ds3234_read_reg(dev, DS323X_REG_CONTROL, 1, &reg);
+    reg &= ~(DS323X_REG_CONTROL_EOSC_MASK | DS323X_REG_CONTROL_INTCN_MASK |
```
PeterKietzmann: Would you mind to spend a note which options you (un)set here? Is this a form of device reset?

drivers/ds323x/include/ds3234_params.h
```diff
@@ -0,0 +1,65 @@
+/*
```
PeterKietzmann: Do you think it is necessary to have separate *params* files?

drivers/Makefile.dep
gebart: It's just to get the right file compiled without special tricks to get it recognized as a file from the ds323x directory

drivers/ds323x/Makefile
gebart: Implemented this way makes it easy to add a ds3231.c later on

drivers/ds323x/ds3234.c
gebart: I'll add a comment. The bits rs1 rs2 select which frequency the sqw pin will output. The intcn bit selects whether the sqw pin is an interrupt output or a square wave signal output. The eosc bit enables the oscillator when it is cleared.

drivers/ds323x/include/ds3234_params.h
gebart: Since ds3234 is spi connected and the others are i2c, I figured the configuration should be separate

PeterKietzmann:
@gebart thanks for the immediate feedback. It wasn't clear to me that the other device is I2C based. Then I'm fine with the file-wise isolation. Let's wait until the device arrives.

gebart:
Sounds good :+1:

leandrolanzieri:
I just tested this on a  Nucleo F030R8 and on a SAM R21 (after rebasing to apply #9885), and can confirm it works as intended.

Here you can se the 1Hz square wave:
sqw_ds3234

Going through the device datasheet I saw there is no ID register or something like that to evaluate if device is actually present on the bus or not during initialization. Would it be worth it to add a function that writes and reads a byte from the SRAM present on the device? Maybe something like:
```C
bool ds3234_is_present(const ds3234_params_t *dev)
{
    uint8_t value = 0x55, address = 0x0A, reg = 0;
    ds3234_write_reg(dev, DS323X_REG_SRAM_ADDR, 1, &address);
    ds3234_write_reg(dev, DS323X_REG_SRAM_DATA, 1, &value);
    ds3234_write_reg(dev, DS323X_REG_SRAM_ADDR, 1, &address);
    ds3234_read_reg(dev, DS323X_REG_SRAM_DATA, 1, &reg);
    
    if (ENABLE_DEBUG) {
        print_str("Expected: 0x");
        print_byte_hex(value);
        print_str(" - Read: 0x");
        print_byte_hex(reg);
        print_str("\n");
    }

    return value == reg;
}

```

gebart:
@leandrolanzieri I would prefer to not touch the SRAM portion of the chip. What we could do instead is read back the written register and see that it matches what we expect, if you would like some more feedback to the application.

leandrolanzieri:
@gebart yes, we could do that. The problem is that, with the current configuration, the control register is always 0.

leandrolanzieri:
Maybe we could go without checking if the device is present for now?

PeterKietzmann:
Can't you just write ones to RS1 and RS2 in the Control register, read it out, check, and reset again afterwards? We could leave this feature out, sure, but now that it's on the table I think it would be nice to implement it at once.

leandrolanzieri:
> Can't you just write ones to RS1 and RS2 in the Control register, read it out, check, and reset again afterwards? We could leave this feature out, sure, but now that it's on the table I think it would be nice to implement it at once.

Yes, I think this could be easily done.

drivers/ds323x/ds3234.c
```diff
@@ -94,6 +95,12 @@ int ds3234_pps_init(const ds3234_params_t *dev)
     reg &= ~(DS323X_REG_CONTROL_EOSC_MASK | DS323X_REG_CONTROL_INTCN_MASK |
         DS323X_REG_CONTROL_RS1_MASK | DS323X_REG_CONTROL_RS2_MASK);
     ds3234_write_reg(dev, DS323X_REG_CONTROL, 1, &reg);
+    uint8_t readback = 0;
```
leandrolanzieri: I tested this and still does not solve the issue of not having the sensor connected. During the tests I disconnected the sensor and sometimes the reading of the control register returned 0, so the check would pass. I suggest to write the control register first with two known bits and then checking. Something like:
```C
    ds3234_read_reg(dev, DS323X_REG_CONTROL, 1, &reg);

    reg |= DS323X_REG_CONTROL_RS1_MASK;
    reg &= ~(DS323X_REG_CONTROL_RS2_MASK);

    ds3234_write_reg(dev, DS323X_REG_CONTROL, 1, &reg);

    uint8_t readback = 0;
    ds3234_read_reg(dev, DS323X_REG_CONTROL, 1, &readback);
    if (reg != readback) {
        DEBUG("ds3234: readback mismatch: expected %u, actual %u\n", (unsigned)reg, (unsigned)readback);
        return -EIO;
    }

    reg &= ~(DS323X_REG_CONTROL_EOSC_MASK | DS323X_REG_CONTROL_INTCN_MASK |
        DS323X_REG_CONTROL_RS1_MASK | DS323X_REG_CONTROL_RS2_MASK);
    ds3234_write_reg(dev, DS323X_REG_CONTROL, 1, &reg);
```

PeterKietzmann:
ping @gebart.

PeterKietzmann:
@leandrolanzieri how about you take over this PR? You base your changes on @gebart's commit and I'll test/review then. @gebart do you agree?

gebart:
@PeterKietzmann sounds good! I will not be able to work any more on this PR

drivers/ds323x/ds3234.c
leandrolanzieri: Added a comment

leandrolanzieri:
@PeterKietzmann I rebased and applied the suggested changes.

drivers/ds323x/include/ds3234_params.h
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * @name    Set default configuration parameters for the DS3234 devices
+ * @{
+ */
+#ifndef DS3234_PARAM_SPI
+#define DS3234_PARAM_SPI                (SPI_DEV(0))
+#endif
+#ifndef DS3234_PARAM_CS
+#define DS3234_PARAM_CS                 (SPI_HWCS(0))
```
PeterKietzmann: I think `GPIO_PIN(0, 0)` would be the more common way here.

drivers/ds323x/ds3234.c
```diff
@@ -0,0 +1,117 @@
...
+ * @}
+ */
+
+#include <stdint.h>
+#include <errno.h>
+#include "ds3234.h"
+#include "ds323x_regs.h"
+#include "fmt.h"
+
+#define ENABLE_DEBUG    (1)
```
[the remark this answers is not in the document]
leandrolanzieri: Done

drivers/ds323x/ds3234.c
```diff
@@ -0,0 +1,117 @@
...
+    if (res < 0) {
+        return DS3234_NO_SPI;
+    }
+    DEBUG("ds3234: init on SPI_DEV(%u)\n", dev->spi);
+
+    if (ENABLE_DEBUG) {
+        for (int k = 0; k <= 0x19; ++k) {
+            uint8_t dbg_reg = 0;
+            ds3234_read_reg(dev, k, 1, &dbg_reg);
+            print_byte_hex(k);
```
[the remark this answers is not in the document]
leandrolanzieri: Changed that

drivers/ds323x/include/ds3234_params.h
leandrolanzieri: Changed

leandrolanzieri:
> Furthermore, what is your opinion about the pseudo-module preparation in this PR?

@PeterKietzmann I think it might be confusing to see that the DS323X family is there but actually only one device is supported. I believe for now we can skip that and just keep this minimal driver as DS3234 until an implementation of another device of the family is implemented.

drivers/include/ds3234.h
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2018 SKF AB
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_ds3234  DS3234 Extremely Accurate SPI RTC
```
maribu: @gebart: I think the `@ingroup` is missing here. As a documentation is shown [directly in modues](http://api.riot-os.org/modules.html) in the doxygen doc. Care to make a follow up PR to address this?

PR #9845, opened 2018-08-26
make: Introduce periph_gpio_irq feature

gebart:
### Contribution description

Introduce a new feature flag `periph_gpio_irq` for when a driver/application requires GPIO interrupts. This allows the GPIO drivers to avoid allocating space in RAM for interrupt context. See the included change to the msp430fxyz gpio driver.

### Testing procedure

No code is modified, only the header and makefiles (except for the example in msp430fxyz). If the compilation is OK, then everything should be fine.

### Issues/PRs references

This could fix the build failures in #8711 
Based on #9844

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -784,6 +784,11 @@ ifneq (,$(filter uuid,$(USEMODULE)))
   USEMODULE += random
 endif
 
+# Enable periph_gpio when periph_gpio_irq is required
+ifneq (,$(filter periph_gpio_irq,$(USEMODULE)))
+  USEMODULE += periph_gpio
+endif
+
 # always select gpio (until explicit dependencies are sorted out)
 FEATURES_OPTIONAL += periph_gpio
 
--- a/cpu/kinetis/Makefile.features
+++ b/cpu/kinetis/Makefile.features
@@ -1,6 +1,7 @@
 FEATURES_PROVIDED += periph_cpuid
 FEATURES_PROVIDED += periph_hwrng
 FEATURES_PROVIDED += periph_gpio
+FEATURES_PROVIDED += periph_gpio_irq
 FEATURES_PROVIDED += periph_mcg
 
 include $(RIOTCPU)/cortexm_common/Makefile.features
--- a/boards/airfy-beacon/Makefile.features
+++ b/boards/airfy-beacon/Makefile.features
@@ -1,6 +1,6 @@
 # Put defined MCU peripherals here (in alphabetical order)
 FEATURES_PROVIDED += periph_adc
-FEATURES_PROVIDED += periph_gpio
+FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += periph_i2c
 FEATURES_PROVIDED += periph_rtt
 FEATURES_PROVIDED += periph_spi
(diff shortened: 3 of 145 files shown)


Makefile.dep
```diff
@@ -784,6 +784,11 @@ ifneq (,$(filter uuid,$(USEMODULE)))
   USEMODULE += random
 endif
 
+# Enable periph_gpio when periph_gpio_irq is required
+ifneq (,$(filter periph_gpio_irq,$(USEMODULE)))
+  USEMODULE += periph_gpio
```
cladmi: In that case, should it not be `FEATURES_REQUIRED +=` ?
It has the same result of having the name put in `USEMODULE` but also triggers the `FEATURES` dependency checks.

cpu/kinetis/Makefile.features
```diff
@@ -1,6 +1,7 @@
 FEATURES_PROVIDED += periph_cpuid
 FEATURES_PROVIDED += periph_hwrng
 FEATURES_PROVIDED += periph_gpio
+FEATURES_PROVIDED += periph_gpio_irq
```
cladmi: This one is on a different line while other are put on the same one.

cladmi:
> It was mainly to save memory on wsn430. Other CPUs may benefit as well, but I think it is better to do follow up PRs for those.

Ok, that is what I was wondering. So would be a following tracking PR/issue.

cpu/kinetis/Makefile.features
cladmi: Maybe put it on the same line to be coherent, people like to copy paste afterwards.

cladmi:
Any test I could run just to validate the wsn430 ?

kaspar030:
I like this a lot. tests/minimal and the coming OTA bootloader will loose RAM weight.

gebart:
@kaspar030 and also ROM size, the IRQ handlers for GPIO interrupts will not be linked in the binary.

gebart:
@cladmi do you have the board? if so you could try tests/periph_gpio and fiddle with some wires on the GPIO pins

gebart:
But anyway, if the linking is OK for wsn430, then the functions are still there, so it should be fine even without on hardware testing if it was working before.

Is the name periph_gpio_irq fine or should it be periph_gpio_int? (since the init function is named gpio_init_int)

A follow up to this PR is to add `FEATURES_REQUIRED += periph_gpio` to all boards which use gpio_set/clear/read etc in the board.c, alternatively, conditionally compile parts of the board_init, such as `gpio_init(LED0_PIN)`.

cladmi:
I prefer `irq` than `int`.

I only have the board on IoT-LAB so no direct access… :/

cladmi:
We tried with @kYc0o on `z1`.

We successfully ran `init_int 2 5 0` which sets an interrupt for the button both on master and this PR.

kaspar030:
Now we have "periph_gpio_irq" as module, but "gpio_init_int" as function. I suggest we rename the latter to "gpio_irq_init()", but provide a define (or static inline wrapper) for the old name.

gebart:
@kaspar030 
>Now we have "periph_gpio_irq" as module, but "gpio_init_int" as function. I suggest we rename the latter to "gpio_irq_init()", but provide a define (or static inline wrapper) for the old name.

Good idea! :+1:

PR #9848, opened 2018-08-27
sys/eepreg: EEPROM registration support (version 2)

ZetaR60:
### Contribution description

This provides an extensive facility to easily manage the locations of data stored in EEPROM via a meta-data registry. Entries are addressed via a string name and can be dynamically added, and the position easily retrieved. Additional facilities are provided to allow migration utilities to be built, permitting legacy data to be moved to new versions of RIOT.

### Testing procedure
Tests provided in ```tests/eepreg```. Currently passes with no problems on mega-xplained.

### Issues/PRs references
Previous version: #9537

### DIFF
--- a/sys/eepreg/eepreg.c
+++ b/sys/eepreg/eepreg.c
@@ -1,0 +1,517 @@
+/*
+ * Copyright (C) 2018 Acutam Automation, LLC
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup   sys_eepreg
+ * @{
+ *
+ * @file
+ * @brief   eepreg implementation
+ *
+ * @author  Matthew Blue <matthew.blue.neuro@gmail.com>
+ * @}
+ */
+
+#include <errno.h>
+#include <limits.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "eepreg.h"
+#include "irq.h"
+#include "periph/eeprom.h"
+
+#define ENABLE_DEBUG 0
+#include "debug.h"
+
+/* EEPREG magic number */
+static const char eepreg_magic[] = "RIOTREG";
+
+/* constant lengths */
+#define MAGIC_SIZE      (sizeof(eepreg_magic))
+
+/* constant locations */
+#define REG_START        (EEPROM_RESERV_CPU_LOW + EEPROM_RESERV_BOARD_LOW)
+#define REG_MAGIC_LOC    (REG_START)
+#define REG_END_LOC      (REG_MAGIC_LOC + MAGIC_SIZE)
+#define REG_ENT1_LOC     (REG_END_LOC + EEPREG_LOC_LEN)
+#define DAT_START        (EEPROM_SIZE - EEPROM_RESERV_CPU_HI \
+                          - EEPROM_RESERV_BOARD_HI - 1)
+
+static inline uint32_t _read_meta_uint(uint32_t loc)
+{
+    uint8_t data[4];
+    uint32_t ret;
+
+    eeprom_read(loc, data, EEPREG_LOC_LEN);
+
+    /* unused array members will be discarded */
+    ret = ((uint32_t)data[0] << 24)
+          | ((uint32_t)data[1] << 16)
+          | ((uint32_t)data[2] << 8)
+          | ((uint32_t)data[3]);
+
+    /* bit shift to discard unused array members */
+    ret >>= 8 * (4 - EEPREG_LOC_LEN);
+
+    return ret;
+}
+
+static inline void _write_meta_uint(uint32_t loc, uint32_t val)
+{
+    uint8_t data[4];
+
+    val <<= 8 * (4 - EEPREG_LOC_LEN);
+
+    data[0] = (uint8_t)(val >> 24);
+    data[1] = (uint8_t)(val >> 16);
+    data[2] = (uint8_t)(val >> 8);
+    data[3] = (uint8_t)val;
+
+    eeprom_write(loc, data, EEPREG_LOC_LEN);
+}
+
+static inline uint32_t _get_reg_end(void)
+{
+    return _read_meta_uint(REG_END_LOC);
+}
+
+static inline void _set_reg_end(uint32_t loc)
+{
+    _write_meta_uint(REG_END_LOC, loc);
+}
+
+static inline uint32_t _get_free_loc(void)
+{
+    /* free location is stored at the end of the registry */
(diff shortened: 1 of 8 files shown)


ZetaR60:
Currently eepreg is performing many more reads than it actually needs to. I am rewriting some of the internal functions to correct this.

ZetaR60:
Changes:
- Registry structure has been better optimized. It is now similar to a linked list (it is no longer necessary to read a whole entry to get to the next one), data pointers have been moved around to make it easier to get the last entry's data location, and the registry is a couple bytes shorter.
- Many redundant EEPROM reads have been replaced by passing arguments to internal functions, and reads have been reduced in a few other places.

Registry structure (from the header documentation):
```
 *    EEPROM_RESERV_CPU_LOW
 *    EEPROM_RESERV_BOARD_LOW
 *    Registry magic number ("RIOTREG\0")
 *    Registry end pointer
 *    Registry entry 1 meta-data length (1 byte)
 *    Registry entry 1 name (unterminated)
 *    Registry entry 1 data pointer
 *    Registry entry 2 meta-data length
 *    Registry entry 2 name
 *    Registry entry 2 data pointer
 *    ... (new registry meta-data may be added in ascending order)
 *    unused space
 *    ... (new data locations may be added in descending order)
 *    Entry 2 data
 *    Entry 1 data
 *    EEPROM_RESERV_BOARD_HI
 *    EEPROM_RESERV_CPU_HI
```

ZetaR60:
Note: This PR is ready for review.

sys/eepreg/eepreg.c
```diff
@@ -0,0 +1,518 @@
...
+
+    return 0;
+}
+
+int eepreg_check(void)
+{
+    char magic[MAGIC_SIZE + 1];
+
+    /* make sure string is always terminated */
+    magic[MAGIC_SIZE] = '\0';
```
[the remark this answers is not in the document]
ZetaR60: Fixed.

ZetaR60:
Thanks for taking a look.

> I still don't get why you need EEPROM_RESERV_CPU_LOW and EEPROM_RESERV_BOARD_LOW to determine the registry start position.
> Same with EEPROM_RESERV_CPU_HI and EEPROM_RESERV_LOW_HI, we could just use a REG_SIZE instead.

This is intended to give the CPU and board configuration a safe method to store things in EEPROM if it needs a hard-coded location not using eepreg. Otherwise there is no safe location to store things outside of eepreg, because eepreg uses space from both the beginning and end of the EEPROM. It is functionally the same as having REG_SIZE and REG_OFFSET definitions, except that anyone reading the definitions in the config files will immediately see how much is being reserved instead of having to calculate it from the EEPROM size.

jia200x:
@ZetaR60 soft freeze is planned for the next week. What else is missing here in order to proceed?

ZetaR60:
@jia200x Nothing; everything is finished. The only changes to the code that I anticipate are due to feedback from reviewers.

leandrolanzieri:
@ZetaR60 I'm trying to test this on a nucleo-l152re. While running the test app I found that the `_get_meta_loc` function hangs, when it is called from `eepreg_add`.
I tried to debug it and inside that function the meta_len is always 0. Am I messing with some configuration?

ZetaR60:
@leandrolanzieri Looks like there was a bug where ```_get_meta_loc``` tries to read an entry when the registry is empty. The issue should now be fixed. Could you re-run the test app, and if it fails again could you post the output with ```#define ENABLE_DEBUG 1``` set in ```sys/eepreg/eepreg.c```?

ZetaR60:
@leandrolanzieri Thanks for the tests and suggestions.

I have rebased to pick up changes to how the Python tests are referenced, and updated ```tests/eepreg/Makefile``` to support this.

ZetaR60:
@aabadie Everything is ready to merge, subject to your approval. I have gone ahead and squashed; I hope you don't mind. There was only one significant change since the fixes for your review comments. The diff for it is reproduced below. A final double-check run of ```tests/eepreg``` works fine.

```
diff --git a/sys/eepreg/eepreg.c b/sys/eepreg/eepreg.c
index 874bb35..1e89e46 100644
--- a/sys/eepreg/eepreg.c
+++ b/sys/eepreg/eepreg.c
@@ -172,15 +172,15 @@ static inline uint32_t _get_meta_loc(const char *name)
 {
     uint32_t meta_loc = REG_ENT1_LOC;
     uint32_t reg_end = _get_reg_end();
-    uint8_t meta_len = _get_meta_len(meta_loc);
 
-    while (meta_loc + meta_len <= reg_end) {
+    while (meta_loc < reg_end) {
+        uint8_t meta_len = _get_meta_len(meta_loc);
+
         if (_cmp_name(meta_loc, name, meta_len)) {
             return meta_loc;
         }
 
         meta_loc += meta_len;
-        meta_len = _get_meta_len(meta_loc);
     }
 
     /* no meta-data found */
```

sys/eepreg/eepreg.c
```diff
@@ -0,0 +1,515 @@
...
+    uint32_t tot_meta_len = reg_end - meta_loc;
+
+    uint32_t data_loc = _get_data_loc(meta_loc, meta_len);
+    uint32_t data_len = _get_data_len(meta_loc, data_loc);
+
+    /* data_loc is above last_loc due to descending order */
+    uint32_t tot_data_len = data_loc - last_loc;
+
+    /* don't allow interrupts when editing registry */
+    unsigned int irq_state = irq_disable();
```
aabadie: I think this could be moved in a more central place: in the EEPROM peripheral driver, `eeprom_write` function. At least for AVR. For STM32s, this is more complex.

ZetaR60: I have removed the IRQ calls. It was probably not necessary anyway, since you would have to be doing weird things in an interrupt to mess it up.

jia200x:
let's try to get this before the hard feature freeze! (Oct 23)

sys/eepreg/eepreg.c
ZetaR60: I meant that there should be no problems in eepreg itself with being interrupted, so long as the reads/writes eventually complete. "Weird" things in this case would be manipulating the registry inside of interrupt code.

ZetaR60:
@aabadie What else is needed regarding reviewing this PR?

aabadie:
Also, please squash. I let @leandrolanzieri push the green button.

ZetaR60:
Squashed.

@aabadie I think @leandrolanzieri does not have the ability to merge (he does not appear as a member for me). We can wait for his statement of final approval though.

leandrolanzieri:
All good for me, @aabadie you will have to merge. Thanks @ZetaR60 !

ZetaR60:
@aabadie @leandrolanzieri Thank you both for your help!

PR #9852, opened 2018-08-28
gnrc_rpl: remove DEBUG from header files

bergzand:
### Contribution description
These DEBUG defines generate conflicts with the DEBUG setting in C files that include these
headers making it impossible to enable the DEBUG setting in these C
files.

### Testing procedure

enable DEBUG in `sys/net/gnrc/routing/rpl/gnrc_rpl_control_messages.c`

### Issues/PRs references

None

### DIFF
--- a/sys/net/gnrc/routing/rpl/gnrc_rpl_internal/validation.h
+++ b/sys/net/gnrc/routing/rpl/gnrc_rpl_internal/validation.h
@@ -26,31 +26,6 @@ extern "C" {
 #include "net/gnrc/rpl/structs.h"
 #include "net/gnrc/icmpv6.h"
 
-#define ENABLE_DEBUG    (0)
-#include "debug.h"
-
-/**
- * @brief   Checks validity of DIS control messages
- *
- * @param[in]   dis     The DIS control message
- * @param[in]   len     Length of the DIS control message
- *
- * @return  true, if @p dis is valid
- * @return  false, otherwise
- */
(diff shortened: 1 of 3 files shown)


sys/net/gnrc/routing/rpl/gnrc_rpl_internal/validation.h
```diff
@@ -46,8 +43,6 @@ static inline bool gnrc_rpl_validation_DIS(gnrc_rpl_dis_t *dis, uint16_t len)
         return true;
     }
 
-    DEBUG("RPL: wrong DIS len: %d, expected: %d\n", len, expected_len);
```
smlng: I would suggest to add a DEBUG message where these functions are used (its only in one file), i.e. in `gnrc_rpl_control_messages.c`

bergzand: As an alternative solution, I could also move these functions as to `gnrc_rpl_control_messages.c` and keep the `DEBUG` messages in.

smlng: I agree DEBUG output shouldn't be in header files, but also we need debug output - hence my point: don't remove but move them. I don't have a preference how, so you can proceed with your idea.

bergzand: Thanks for the reply!

bergzand:
@smlng The functions are moved to the C-file.

smlng:
please squash

bergzand:
@smlng Squashed!

PR #9857, opened 2018-08-29
net/gcoap: add user ptr to response handler functions

haukepetersen:
### Contribution description
This PR changes the `gcoap` API slightly to allow passing user defined pointers to response handler functions (`gcoap_resp_handler_t`). This is analog to the user context pointers that were added to the resource handler functions in nanocoap a while ago (`coap_resource_t`).

Adding these user context pointers is quite useful, when handling concurrent requests of a similar type (i.e. registration to multiple CoRE RD's at the same time in my case).

I actually can't remember if this change has been discussed before, or if this change does even already exist. A quick search through the open PRs and issue did not yield any results, so I guess not?!

### Testing procedure
So far, this is rather a request for comments, so all changes to the code are untested...

### Issues/PRs references
none

### DIFF
--- a/examples/gcoap/gcoap_cli.c
+++ b/examples/gcoap/gcoap_cli.c
@@ -14,6 +14,7 @@
  * @brief       gcoap CLI support
  *
  * @author      Ken Bannister <kb2ma@runbox.com>
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
  *
  * @}
  */
@@ -31,8 +32,8 @@
 
 static ssize_t _encode_link(const coap_resource_t *resource, char *buf,
                             size_t maxlen, coap_link_encoder_ctx_t *context);
-static void _resp_handler(unsigned req_state, coap_pkt_t* pdu,
-                          sock_udp_ep_t *remote);
+static void _resp_handler(const gcoap_request_memo_t *memo, coap_pkt_t* pdu,
+                          const sock_udp_ep_t *remote);
 static ssize_t _stats_handler(coap_pkt_t* pdu, uint8_t *buf, size_t len, void *ctx);
 static ssize_t _riot_board_handler(coap_pkt_t* pdu, uint8_t *buf, size_t len, void *ctx);
 
@@ -84,16 +85,16 @@ static ssize_t _encode_link(const coap_resource_t *resource, char *buf,
 /*
  * Response callback.
  */
-static void _resp_handler(unsigned req_state, coap_pkt_t* pdu,
-                          sock_udp_ep_t *remote)
+static void _resp_handler(const gcoap_request_memo_t *memo, coap_pkt_t* pdu,
+                          const sock_udp_ep_t *remote)
 {
     (void)remote;       /* not interested in the source currently */
 
-    if (req_state == GCOAP_MEMO_TIMEOUT) {
+    if (memo->state == GCOAP_MEMO_TIMEOUT) {
         printf("gcoap: timeout for msg ID %02u\n", coap_get_id(pdu));
         return;
     }
-    else if (req_state == GCOAP_MEMO_ERR) {
+    else if (memo->state == GCOAP_MEMO_ERR) {
         printf("gcoap: error in response\n");
         return;
     }
--- a/examples/gcoap/gcoap_cli.c
+++ b/examples/gcoap/gcoap_cli.c
@@ -145,7 +145,8 @@ static void _resp_handler(const gcoap_request_memo_t *memo, coap_pkt_t* pdu,
             block.blknum++;
             coap_opt_add_block2_control(pdu, &block);
             int len = coap_opt_finish(pdu, COAP_OPT_FINISH_NONE);
-            gcoap_req_send((uint8_t *)pdu->hdr, len, remote, _resp_handler);
+            gcoap_req_send((uint8_t *)pdu->hdr, len, remote,
+                           _resp_handler, memo->context);
         }
         else {
             puts("--- blockwise complete ---");
@@ -258,7 +259,7 @@ static size_t _send(uint8_t *buf, size_t len, char *addr_str, char *port_str)
         return 0;
     }
 
-    bytes_sent = gcoap_req_send(buf, len, &remote, _resp_handler);
+    bytes_sent = gcoap_req_send(buf, len, &remote, _resp_handler, NULL);
     if (bytes_sent > 0) {
         req_count++;
     }
--- a/sys/include/net/gcoap.h
+++ b/sys/include/net/gcoap.h
@@ -141,8 +141,8 @@
  *
  * Here is the expected sequence for handling a response in the callback.
  *
- * -# Test for a server response or timeout in the _req_state_ callback
- *    parameter. See the GCOAP_MEMO... constants.
+ * -# Test for a server response or timeout in the `state` field of the `memo`
+ *    callback parameter (`memo->state`). See the GCOAP_MEMO... constants.
  * -# Test the response with coap_get_code_class() and coap_get_code_detail().
  * -# Test the response payload with the coap_pkt_t _payload_len_ and
  *    _content_type_ attributes.
@@ -646,13 +646,20 @@ typedef struct gcoap_listener {
 } gcoap_listener_t;
 
 /**
+ * @brief   Forward declaration of the request memo type
+ *
+ */
+typedef struct gcoap_request_memo gcoap_request_memo_t;
+
+/**
  * @brief   Handler function for a server response, including the state for the
  *          originating request
  *
  * If request timed out, the packet header is for the request.
  */
-typedef void (*gcoap_resp_handler_t)(unsigned req_state, coap_pkt_t* pdu,
-                                     sock_udp_ep_t *remote);
+typedef void (*gcoap_resp_handler_t)(const gcoap_request_memo_t *memo,
+                                     coap_pkt_t* pdu,
+                                     const sock_udp_ep_t *remote);
 
 /**
  * @brief  Extends request memo for resending a confirmable request.
@@ -665,7 +672,7 @@ typedef struct {
 /**
  * @brief   Memo to handle a response for a request
  */
-typedef struct {
+struct gcoap_request_memo {
     unsigned state;                     /**< State of this memo, a GCOAP_MEMO... */
     int send_limit;                     /**< Remaining resends, 0 if none;
                                              GCOAP_SEND_LIMIT_NON if non-confirmable */
@@ -679,7 +686,7 @@ typedef struct {
     gcoap_resp_handler_t resp_handler;  /**< Callback for the response */
     xtimer_t response_timer;            /**< Limits wait for response */
     msg_t timeout_msg;                  /**< For response timer */
-} gcoap_request_memo_t;
+};
 
 /**
  * @brief   Memo for Observe registration and notifications
--- a/sys/include/net/gcoap.h
+++ b/sys/include/net/gcoap.h
@@ -647,7 +647,6 @@ typedef struct gcoap_listener {
 
 /**
  * @brief   Forward declaration of the request memo type
- *
  */
 typedef struct gcoap_request_memo gcoap_request_memo_t;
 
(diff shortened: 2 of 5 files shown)


kb2ma:
I can see the value in a context for a request, and passing the context to the response handler would be simpler than looking up the context in the handler. The example that comes to my mind is the state associated with a guaranteed delivery requirement. In other words, even a Confirmable request can fail. What then?

I have only inspected the code so far, but it looks good. I will test as soon as I get the chance.

My only regret is that this change will break all existing request code, but at least the fix is pretty mechanical.

kb2ma:
To future-proof the change, we can include the request memo in the callback rather than the context itself.
```
typedef void (*gcoap_resp_handler_t)(unsigned req_state, coap_pkt_t* pdu,
                                     gcoap_request_memo_t *memo);
```
With this declaration, we can make more parameters available in the future via the memo without requiring that users change existing code. I replaced the remote endpoint parameter because I'm not sure how widely it is used, and of course the endpoint is available in the request memo.

We could specify in the documentation that the user can count on only specific attributes -- remote_ep and context.

haukepetersen:
In general, using the `memo` as parameter sounds sensible. I am not sure about dropping the `remote` parameter however, since the `remote` parameter in the callback and the `remote_ep` field in the memo struct are not necessarily the same to my understanding -> request to multicast address vs reply from a unicast address.

But then I just noticed, that I might have a large gap in understanding when it comes to CoAP and multicast traffic. I did never read into the group communication documents and it seems that I have misunderstood the RFC so far - it seems to me that the scenario above is not allowed in the first place?! 

In that case dropping the `remote` parameter in favor of the `memo` parameter sounds good to me. But we keep the change of adding a `context` parameter to the send function, right?

kb2ma:
I totally blanked that the original purpose of the remote host in the handler was for the multicast case. I have essentially no multicast experience, so I read up on 7252 sec. 8 as well as 7390. What did you mean specifically about your gap in understanding from the groupcomm documents relative to the need for the remote parameter in the handler?

As I read those documents, particularly 7390, sec 2.5, I agree it makes sense to retain the remote reference in the handler. For example, the client might make a multicast request for which it *expects* individual responses, like "Are you functioning nominally?" And now the client will know which servers responded.

So, how about this declaration for the response handler? I re-added the remote server pointer and made the request memo const. 

```
typedef void (*gcoap_resp_handler_t)(unsigned req_state, coap_pkt_t* pdu,
              sock_udp_ep_t *remote, const gcoap_request_memo_t *memo);
```

As I mentioned earlier in this thread, please document which memo attributes the implementer can count on. Let's limit the exposure of the internals.

And yes, I completely agree that the context pointer must be included in gcoap_req_send2() and gcoap_req_send(), in order to place it in the request memo.

haukepetersen:
> What did you mean specifically about your gap in understanding from the groupcomm documents relative to the need for the remote parameter in the handler?

Mainly that I simply never opened/read RFC7390 :-)

I am still thinking about the best solution for selecting paramters to the resp_handler callback, my current thoughts are: I don't quite like the idea to pass the `memo` struct, and then telling the user which fields are ok to use, and which should not be used. This seems not to be the cleanest/most resilient approach to API design. Further having this mixture of passing some `memo` fields directly (-> `state`) while others can be accessed indirectly (e.g. `context`, `remote_ep`).

So let's see what our options can be (at least what I can think of right now):
- go back to pass the `context` parameter instead of `memo`: not very future prove and not flexible (e.g. user may not access the original `remote_ep` parameter) -> IMO not a good idea
- live with the fact that users have to make sure to read the doc any only access/use the allowed fields in the memo struct -> might indeed be a sensible trade off between flexibility and API robustness
- we could do a simple cosmetic restructuring of the `memo` struct: move 'non-public' fields into something like an `internal' (e.g. `memo->internal.PRIVATE_FIELD`) sub-struct -> overkill?

On top when passing the `memo` field, we might at least think about dropping the `req_state` paramter then, right?

haukepetersen:
just to get a feel: I just pushed a commit that drops the `req_state` parameter in favor of passing a `memo` pointer. I think I tend to like it for its cleanliness.

kb2ma:
Thanks for summarizing the options for the handler. I like your latest resolution, like below.

```
typedef void (*gcoap_resp_handler_t)(const gcoap_request_memo_t *memo,
                                     coap_pkt_t* pdu,
                                     const sock_udp_ep_t *remote);
```

I was in favor of keeping the `req_state` parameter in the declaration because it was like a return value for the handler. The user definitely needs to inspect this parameter first. However, your change basically keeps this order of parameters. So now, the user must inspect `memo->state` first, which is just a small change. It's nice with this declaration that we also get access to the context parameter at no extra cost.

I am not concerned about including the memo in the declaration. The memo as a whole really is a context object, qualifying the response PDU. Regarding the need to specify which fields in the PDU that the user should access:

* We mark the struct 'const', so the user knows not to mess with it.
* We have a similar use model for the coap_pkt_t in a request handler. The user must access the payload pointer and length via the coap_pkt_t when building the response.
* It's worth it not to break applications in the future if we need to retain/pass another parameter.
  
I also agree that it's overkill to add some 'private' marking to the struct, at least for now. Maybe later when the APIs are stable.

So, please document and update examples, and I'll test it out. When you update documentation, remember the _Handling the response_ section in the module documentation.

haukepetersen:
I fully agree, and looking at the function today I also still like the solution in passing the memo as first parameter. I will adapt the PR accordingly later today.

haukepetersen:
Adapted this PR according to the results of our discussion:
- the resp_handler() callback now passes a const pointer to the memo struct as first parameter
- the memo holds an additional user defined `context` field
- the `gcoap_req_send()` and `gcoap_req_send2()` functions take an additional `void *context` parameter

kb2ma:
Sorry to take so long to get back on this one, @haukepetersen. Let's get this merged! The code looks basically sound and matches our discussion above.

I think the first step is to rebase and add a commit to update cord_ep's use of gcoap_req_send2() and gcoap_resp_handler_t callback. We also need some instructions on how to test.

haukepetersen:
no problem, I am not less responsible for delays in this PR...

I will try to do the requested, but can't promise anything before next Wednesday... Will try my best.

kb2ma:
As we discussed, I rebased and updated the cord_ep* examples in my [gcoap/req_user_context](https://github.com/kb2ma/RIOT/tree/gcoap/req_user_context) branch. Please rebase your opt_gcoap_contextptrtoresphandler branch so I can create a PR against it.

Also, would you please confirm that the cord_epsim example fails with the default RD address of ff02::1. I use a tap interface between aiocoap-rd on my Linux workstation and the cord_epsim example for native. aiocoap complains with:

> WARNING:coap-server:Dropping pktinfo from ancdata because it indicates a multicast address

and eventually the cord_epsim example errors in the CLI with:

> [cord_epsim] error: unable to send registration

The example works when I compile with RD_ADDR set to the link local unicast address.

haukepetersen:
>Also, would you please confirm that the cord_epsim example fails with the default RD address of ff02::1

Confirmed! Even if `aiocoap` would not reject the request, the example would fail with a multi-cast address since gcoap is not able to handle group communication (yet). I don't remember the exact location in the code, but somewhere when receiving, we check if the dst addr of the outgoing request is equal to the src addr of the response (as stated in the RFC) -> and this will of course fail for multicast messages...

I will change the `epsim` example in a separate PR, so that at least the is no broken default...

haukepetersen:
also rebased.

kb2ma:
> the example would fail with a multi-cast address since gcoap is not able to handle group communication (yet)

Right, and this PR provides the data for the client to know when a response is from a multicast request. I see a folllow-on PR on the horizon. :-)

haukepetersen:
Fixed the issue for now in #10464

But the group communication stuff would indeed be nice to implement at some point :-)

kb2ma:
In the description you say: 

> Adding these user context pointers is quite useful, when handling concurrent requests of a similar type (i.e. registration to multiple CoRE RD's at the same time in my case)

I'm starting to understand CoRE RD. What exactly would you want to retrieve from the context?

kb2ma:
Now that the response handler has access to the coap_request_memo_t, we may want to streamline the implementation of cord_ep in a follow on PR. Presently cord_ep has separate handlers for on_update and on_remove. However, the coap_request_memo_t has access to the request header. So, a single handler could read the request code from the header instead of hard coding the code in the separate functions.

haukepetersen:
yes, that might actually be a nice cleanup, though I doubt it will make much difference in terms of code-size.

> What exactly would you want to retrieve from the context?

For a better design, I would like to remove the restriction of only being able to connect to a single RD at any given point of time. This would mean, to keep all the state needed for a single connection in a struct, and pass a pointer to this as argument to each API function (so simply the standard approach we see all over the place in RIOT). But a pointer to the actual state struct would also be needed to pass to the request handler functions, hence the context argument...

haukepetersen:
I left this PR sit there idle for some time now, maybe its a good moment to revisit the changes. I rebased it and also adapted the `cord_ep` and `cord_epsim` implementations, so all occurrences of `gcoap_req_sendX()` in the RIOT code base are adapted.

I did however not (yet) look into optimizing the `cord_ep` implementation using the new possibilities as described above. That would be meat for another PR.

@kb2ma @bergzand @kaspar030: what do you guys think, does it make sense to proceed with this PR?

kb2ma:
Thanks, @haukepetersen! I agree completely to proceed with this PR. The ability to pass state from request to response, as well as from response to a follow-on request is useful and supports constrained resource use. Block extension handling is another good example for this need. I intend to review again ASAP.

haukepetersen:
Hej @kb2ma: lost this PR get out of view once more :-)

I rebased it and adapted it to the latest changes in gcoap. Would you mind to have another look?! Thanks!

kb2ma:
Thanks for rebasing this PR back to life. I will look again after the release. It will be a good time to add infrastructure.

haukepetersen:
sounds good to me!

chrysn:
Having this would help applications using OSCORE (their response handler is basically "decrypt and verify with these nonces, then pass on to the actual response handler").

Is this PR in an overall state where we expect no further obstacles? If so, I'd make this branch a prerequisite for the OSCORE demos and build on it (reviewing the usability as I go), otherwise I'll need to work around (and can still review, but from a more theoretical perspective).

haukepetersen:
go ahead! I hope its just the final review/verification that this PR is waiting for :-)

sys/include/net/gcoap.h
```diff
@@ -626,14 +626,21 @@ typedef struct gcoap_listener {
     struct gcoap_listener *next;        /**< Next listener in list */
 } gcoap_listener_t;
 
+/**
+ * @brief   Forward declaration of the request memo type
+ *
```
kb2ma: typo: remove extra line

examples/gcoap/gcoap_cli.c
```diff
@@ -144,7 +145,8 @@ static void _resp_handler(unsigned req_state, coap_pkt_t* pdu,
             block.blknum++;
             coap_opt_add_block2_control(pdu, &block);
             int len = coap_opt_finish(pdu, COAP_OPT_FINISH_NONE);
-            gcoap_req_send((uint8_t *)pdu->hdr, len, remote, _resp_handler);
+            gcoap_req_send((uint8_t *)pdu->hdr, len, remote,
+                           _resp_handler, NULL);
```
kb2ma: I would set the context for the new request with the context from the received response. I realize it's NULL at present, but this code should not make that assumption.

chrysn:
Works very well for my OSCORE demo use case. (In its current example, the caller allocates a struct that contains the request ID necessary for correlation, and a mutex used to tell whoever allocated that struct that he may return thus freeing the struct).

examples/gcoap/gcoap_cli.c
haukepetersen: of course, makes a lot of sense!

haukepetersen:
addressed issues and rebased. Let me know when you want me to squash.

kb2ma:
Looks good. Tested all the clients. Please squash!

haukepetersen:
squashed

PR #9858, opened 2018-08-29
 doc: refactor `hashes` group

jia200x:
### Contribution description
Following #9842, this PR groups all hashes into the `hashes` group. I also define/modify some other groups in order to create the categories Checksums, Keyed cryptographic hash functions, Unkeyed cryptographic hash functions and Non cryptographic hash functions (all of them under `sys_hashes` group).

This PR does the following:

- Checksums moved from `sys`group to `sys_hashes` group
- Define groups for `sha256` and `sha3`
- Add CMAC to `sys_hashes_keyed`
- Add MD5, SHA1, SHA256 and SHA3 to `sys_hashes_unkeyed`
- Add all hashes from `hashes.h` files to `sys_hashes_non_crypto`

I also deprecated the section based documentation from the `doc.txt`, because it was outdated and it had the same information as the Doxygen group list.

### Testing procedure
`make doc`. Go to `System>Hashes` in the generated documentation and see how these components are grouped.

### Issues/PRs references
#8479

### DIFF
--- a/sys/checksum/doc.txt
+++ b/sys/checksum/doc.txt
@@ -9,7 +9,7 @@
 
 /**
  * @defgroup    sys_checksum Checksum
- * @ingroup     sys
+ * @ingroup     sys_hashes
  * @brief       Checksum function libraries
  *
  * This module provides a number of checksum functions. Most notably is the
--- a/sys/hashes/doc.txt
+++ b/sys/hashes/doc.txt
@@ -36,3 +36,21 @@
  * * SHA-256
  *
  */
+
+/**
+ * @defgroup    sys_hashes_non_crypto Non-cryptographic hash functions
+ * @ingroup     sys_hashes
+ * @brief       A collection of non-cryptographic hash algorithms.
+ */
+
+/**
+ * @defgroup    sys_hashes_unkeyed Unkeyed cryptographic hash functions
+ * @ingroup     sys_hashes
+ * @brief       A collection of unkeyed cryptographic hash algorithms.
+ */
+
+/**
+ * @defgroup    sys_hashes_keyed Keyed cryptographic hash functions
+ * @ingroup     sys_hashes
+ * @brief       A collection of keyed cryptographic hash algorithms.
+ */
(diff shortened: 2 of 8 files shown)


sys/checksum/doc.txt
```diff
@@ -9,7 +9,7 @@
 
 /**
  * @defgroup    sys_checksum Checksum
- * @ingroup     sys
+ * @ingroup     sys_hashes
```
PeterKietzmann: Why do you put checksum in `sys_hashes`? I would guess that (as long as we keep both checksum/ and hashes/ in separate folders) we would keep both `@ingroup sys` ?

jia200x: As I have read, strictly speaking checksums are hashes. See for instance [this entry](https://en.wikipedia.org/wiki/List_of_hash_functions).
Although they are in different folder, I'm trying to organize them by topic :)

jia200x:
> Nice! Did you also have a look at the header files in sys/include/checksum/?

Yes, since I move the `sys_checksum` group to `sys_hashes`, these headers are not affected by the change (the `@ingroup` directive is in charge of moving the group)

sys/checksum/doc.txt
PeterKietzmann: I was referring to the mismatch between folder structure and documentation. But I agree that it looks way nicer this way. So the conclusion might be to consider file shifting which is not the purpose of this PR and maybe not wanted anyway.

PeterKietzmann:
>  these headers are not affected by the change

Sure, I was more thinking about the need for further cleanup. Otherwise I'm fine with this PR.

sys/checksum/doc.txt
jia200x: Ah, I see. Yes, could be an option. I would be in for moving the folder in another PR

jia200x:
> Sure, I was more thinking about the need for further cleanup. Otherwise I'm fine with this PR.

Good! :)

PeterKietzmann:
PS: I'm wondering if we need that many commits for this kind of change

jia200x:
> PS: I'm wondering if we need that many commits for this kind of change

Maybe it doesn't make that much sense. I will try to commit by group rather than by feature.

PR #9859, opened 2018-08-29
doc: set SUBGROUPING to NO in riot.doxyfile

### LINKED ISSUE #9849: doxygen: formatting problems in member groups
```@name <title> @{ ... @}``` constructs (so-called member groups) are very often used in the documentation to group related definitions or declarations within a separate section with its own title. Examples are LED pin configuration, I2C interface configurations, a.s.o.

Unfortunatly, member groupings by the ```@name``` command lead to different output styles dependent on whether all members are of the same type or not. For example, in ```boards/common/arduino-mkr/periph_conf.h``` you can find the following doxygen documentation for timer and UART configuration.

```
/**
 * @name Timer peripheral configuration
 * @{
 */
#define TIMER_NUMOF         (2U)
#define TIMER_0_EN          1
#define TIMER_1_EN          1
...
/** @} */
...
/**
 * @name UART configuration
 * @{
 */
static const uart_conf_t uart_config[] = {
    {
        ...
    }
};

/* interrupt function name mapping */
#define UART_0_ISR          isr_sercom5

#define UART_NUMOF          (sizeof(uart_config) / sizeof(uart_config[0]))
/** @} */
```
Even though bothof them use ```@name``` to group all related definitions/declaration within a separate section, their outputs look different:

doxygen_01

doxygen_02

The reason is that doxygen's subgrouping is activated in ```doc/doxygen/riot.doxyfile``` by ```SUBGROUPING=YES```. When subgrouping is activated, the ```@name``` command defines a subgroup below the current section if all group members are of the same type. If group members are of different type, a top level group is created instead.

Thus, in the example, the ```UART configuration``` is defined as a top-level group and looks as expected, but the ```Timer peripheral configuration`` is defined as a subgroup in the section ```Macros``` and doesn't look like expected.

If there are no important reasons for subgrouping, I would suggest to disable the subgrouping to improve the quality of the documentation through a consistent look and feel.

In C ++, for which this sub-grouping was originally introduced, sub-grouping can be disabled class-wise by command ```@nosubgroup```.

jia200x:
### Contribution description
This PR implements the idea in #9849. The SUBGROUPING option is set to NO, so the `@name` directive doesn't define a subgroup if all members have the same type. So, all `@name` groups are shown with the same format.

### Testing procedure
Run  `make doc` with and without this PR and see the difference. In #9849 there's a picture describing how does it look without this PR.

### Issues/PRs references
Fixes #9849

### DIFF
--- a/doc/doxygen/riot.doxyfile
+++ b/doc/doxygen/riot.doxyfile
@@ -350,7 +350,7 @@ DISTRIBUTE_GROUP_DOC   = NO
 # \nosubgrouping command.
 # The default value is: YES.
 
-SUBGROUPING            = YES
+SUBGROUPING            = NO
 
 # When the INLINE_GROUPED_CLASSES tag is set to YES, classes, structs and unions
 # are shown inside the group in which they are included (e.g. using \ingroup)


aabadie:
I can confirm that the rendering is more consistent. I couldn't find any obvious regressions while browsing briefly the rest of the documentation.
Maybe @miri64 will find something ?

jcarrano:
I'm for this change. I hate when tools automagically decide to do stuff they were not told to and getting doxygen to spit out what you want is already hard enough.

Issue #9690, opened 2018-08-02
periph/gpio: enable API usage for expanders


kYc0o:
This issue aims to track the progress on the definitions for the GPIO API extensions, and its implementations on the corresponding CPUs.

  - [ ] API proposal
  - [ ] Tests
  - [ ] CPU implementations

kYc0o:
@ZetaR60 I think #9190 should be splitted into the proposed steps to ease the merging. Please feel free to propose new steps or modify the current ones if you feel so.

ZetaR60:
@kYc0o Updated the TODO list with #9860.

ZetaR60:
@kYc0o Updated the TODO list with #9958.

gschorcht:
Just to inform, I have used the GPIO extension API with two real expanders PCF8575 and PCF8574A.  
Thanks to @ZetaR60, it works like a charm. If the driver defines an interface that corresponds to the peripheral GPIO interface, it is very easy to integrate the driver as GPIO extension. I was really impressed how easy it was.

ZetaR60:
Thank you for your kind words @gschorcht !

gschorcht:
@kYc0o Sorry, I missed that this tracking issue is explicity for GPIO when I was extending references with the PRs for ADC, DAC and PWM extension. Should we rename it to perhiph/* to track all types of peripheral extensions or should we open tracking issues for each type of extension API?

kaspar030:
I'd like to take another look at the redirection method.

I haven't taken an in-depth look at the API changes other than for gpio, so this applies mostly to gpio.

The reason why we want to change all the API's in the first place is that we need to support multiple "providers" of the functionality (I2C gpio extenders, bitbanging soft bus implementations, UART-over-USB, external ADC/DAC, ..). Ideally we manage to do that while

- keeping the current user-facing API (so we don't need to change all drivers)
- integrate it into the current API (so drivers can seamlessly use the new "providers")
- don't break existing applications. This includes the fact that the current implementation, when compiled with LTO, optimizes "gpio_set/get/..." to a single memory write if called with constant PIN parameters
- blah blah maintainability, code size, ram usage, ...

If I understand the current method correctly:

- the current periph code is renamed to "gpio_*_cpu". IMO this is good. (although I'd prefer "gpio_cpu_*", but let's talk concept first).

- the interception happens in wrapper functions named like the legacy "gpio_*" API (where else? so good.)

[...]

Issue #9582, opened 2018-07-17
periph_*: Allow extensions for external periph adapters


kYc0o:
Many times, internal peripherals in a MCU are enough to handle basic operations, like GPIOs, ADCs, DACs, etc. However, for complex applications, extensions for such peripherals are needed, and actually widely used in industry.

Since RIOT aim is to keep the periph interfaces as slim and portable as possible, making extensions to them is indeed a complicated task. The aim of this issue is to explore the possible solutions to this, and to work together towards a common definition on how to write APIs and driver extensions for this use-case.

I'll summarize here the interfaces that need changes, and the PRs if there's one.

## GPIO
  - [ ] #9860 (partially replaces #9190) (PR)
  - [ ] #9958 (partially replaces #9190) (PR)

## ADC
  - [ ] #10527  (PR)

## DAC
  - [ ] #10512 (PR)

## PWM
  - [ ] #10533 (PR)

ZetaR60:
My proposal for implementing this. Most of this can been seen in #9190, but there are a few small differences.

1. Any periph interface that is addressed with some form of ID reserves the leading bit of that ID for use by the extension interfaces. This includes of course, GPIO, ADC, DAC, but also perhaps even I2C, SPI, UART, etc. Making this a general policy for all of the addressed interfaces would make it easy to add things like bridges for buses, for example. If the leading bit is a zero, the ID is intended to be used by the CPU, if it is a one then it is intended to be used by an extension interface.
2. If the identifier indicates that the resource being requested is an extension device _and_ extensions are being compiled for that interface, then the API call will be redirected to the relevant extension interface driver.
3. Redirections are implemented by splitting the periph API in two: one that is identical to the current list of functions to be called by users of the API, and one that has ```_ll``` (short for low-level) appended to it (for use by CPU implementations). The low-level functions are called by an inline function from the header that also contains the redirection code. When the extension interface for that periph API is not enabled, the redirection code is not included and so the inline should compile down to be exactly the same as if the extension interface did not exist.
4. The redirect code consists of three operations per API call: check to see if the leading bit is one, call a function ```foocall_redir```, and return (which may return the value from ```foocall_redir```, if the original API call uses a return value). When using CPU resources with the extension interface enabled, this only adds a single check to the overhead.
[...]

PR #9860, opened 2018-08-29
periph/gpio: support for extension API

ZetaR60:
### Contribution description
This is an implementation of the intercept method outlined by my proposal in [#9582](https://github.com/RIOT-OS/RIOT/issues/9582#issuecomment-410094702) (parts 1-4) for the periph/gpio interface. In short, it allows other drivers (such as GPIO expanders) to be addressed using gpio_t so that all existent code can easily gain support through those drivers. For non-implementation detail and discussion, please see #9582

At @kYc0o 's suggestion, I am breaking the implementation of #9582 / #9690 into several pieces to make it easier to review. This PR only contains the API intercept code.

### Testing procedure
A test case is provided for detecting collisions between use of ```GPIO_PIN``` and ```GPIO_EXT_PIN```. Otherwise the existing GPIO tests may be used to confirm CPU based pins work with the API change. This PR only contains API definitions for the redirection functions, so implementation of them will be tested in a follow-up PR.

### Issues/PRs references
Partially replaces and closes #9190 
Implements parts 1-4 of my proposal in #9582 
Part of work tracked by #9690

### DIFF
--- a/drivers/include/periph/gpio.h
+++ b/drivers/include/periph/gpio.h
@@ -173,7 +173,7 @@ void gpio_toggle_cpu(gpio_t pin);
 void gpio_write_cpu(gpio_t pin, int value);
 /** @} */
 
-#if MODULE_EXTEND_ADC || DOXYGEN
+#if MODULE_EXTEND_GPIO || DOXYGEN
 /**
  * @brief   Redirecting versions of the GPIO functions
  *
@@ -191,7 +191,7 @@ void gpio_clear_redir(gpio_t pin);
 void gpio_toggle_redir(gpio_t pin);
 void gpio_write_redir(gpio_t pin, int value);
 /** @} */
-#endif /* MODULE_EXTEND_ADC || DOXYGEN */
+#endif /* MODULE_EXTEND_GPIO || DOXYGEN */
 
 /**
  * @brief   Initialize the given pin as general purpose input or output
--- a/drivers/include/periph/gpio.h
+++ b/drivers/include/periph/gpio.h
@@ -77,11 +77,33 @@ typedef unsigned int gpio_t;
 #define GPIO_PIN(x,y)       ((gpio_t)((x & 0) | y))
 #endif
 
+#ifndef GPIO_EXT_DEV_LOC
+/**
+ * @brief   Set device ID location in gpio_t when using extensions
+ */
+#define GPIO_EXT_DEV_LOC    (8*sizeof(gpio_t) - 8)
+#endif
+
+#ifndef GPIO_EXT_THRESH
+/**
+ * @brief   Pin numbers greater than GPIO_EXT_THRESH use extensions
+ */
+#define GPIO_EXT_THRESH    ((gpio_t)(UINT_MAX) >> 1)
+#endif
+
+#ifndef GPIO_EXT_PIN
+/**
+ * @brief   Convert (device, pin) tuple to @c gpio_t value
+ */
+#define GPIO_EXT_PIN(x, y) \
+    (gpio_t)(~GPIO_EXT_THRESH | (x << GPIO_EXT_DEV_LOC) | GPIO_PIN(0,y))
+#endif
+
 #ifndef GPIO_UNDEF
 /**
  * @brief   GPIO pin not defined
  */
-#define GPIO_UNDEF          ((gpio_t)(UINT_MAX))
+#define GPIO_UNDEF          (GPIO_EXT_THRESH)
 #endif
 
 /**
@@ -134,6 +156,42 @@ typedef struct {
 #endif
 
 /**
+ * @brief   Low-level versions of the GPIO functions
+ *
+ * These are for cpu gpio.c implementation and should not be called directly.
+ * @{
+ */
+int gpio_init_ll(gpio_t pin, gpio_mode_t mode);
+int gpio_init_int_ll(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
+                     gpio_cb_t cb, void *arg);
+void gpio_irq_enable_ll(gpio_t pin);
+void gpio_irq_disable_ll(gpio_t pin);
+int gpio_read_ll(gpio_t pin);
+void gpio_set_ll(gpio_t pin);
+void gpio_clear_ll(gpio_t pin);
+void gpio_toggle_ll(gpio_t pin);
+void gpio_write_ll(gpio_t pin, int value);
+/** @} */
+
+/**
+ * @brief   Redirecting versions of the GPIO functions
+ *
+ * These are for the extension interface and should not be called directly.
+ * @{
+ */
+int gpio_init_redir(gpio_t pin, gpio_mode_t mode);
+int gpio_init_int_redir(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
+                        gpio_cb_t cb, void *arg);
+void gpio_irq_enable_redir(gpio_t pin);
+void gpio_irq_disable_redir(gpio_t pin);
+int gpio_read_redir(gpio_t pin);
+void gpio_set_redir(gpio_t pin);
+void gpio_clear_redir(gpio_t pin);
+void gpio_toggle_redir(gpio_t pin);
+void gpio_write_redir(gpio_t pin, int value);
+/** @} */
+
+/**
  * @brief   Initialize the given pin as general purpose input or output
  *
  * When configured as output, the pin state after initialization is undefined.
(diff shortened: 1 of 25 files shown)


gschorcht:
@ZetaR60 Do you think that you will be able to rebase this PR soon? The ESP32 CPU is not in the master on which your branch is based. I need it to test the `periph/gpio.c` changes for the ESP32.

ZetaR60:
Rebased and ```cpu/esp*``` supported added.

I basically had to throw out all of the ```cpu/*/periph/gpio.c``` changes and start over because all the gpio.c files got reorganized with some recent changes. I have already done this same task several times in #9190, so it would be nice to prioritize this PR amongst the extension PRs for merging before it gets outdated again.

gschorcht:
Great, thanks. It has to be frustrating to do the same again and again.

I would prefer that someone from the core team takes a look again (@kaspar030 @gebart @kYc0o). From my point of view, this part of the GPIO extension is OK.

gschorcht:
@ZetaR60 I'm wondering whether it would possible to define a macro that is set by the CPU peripheral configuration which defines the number of ports provided by the MCU, for example:
```
#define GPIO_PORT_NUMOF    (4)
```
In this case it would become possible to handle GPIO expansion devices based on a port ID starting with `GPIO_PORT_NUMOF`. In addition, the `GPIO_PIN(x, y)` macro could be redefined to handle MCU ports and extension ports in the same way. The extension interface would be completely invisible to the application. It always uses `GPIO_PIN(port, pin)`, regardless of whether the GPIO is provided by the MCU or by the expansion device.
```C
#define GPIO_PIN(x,y)       (x < GPIO_PORT_NUMOF) \
                             ? GPIO_PIN_LL(x, y) \
                             : GPIO_EXT_PIN(x - GPIO_PORT_NUMOF, y))
```
or even better with more consistent naming as proposed in https://github.com/RIOT-OS/RIOT/issues/9582#issuecomment-447766752 :wink:
```C
#define GPIO_PIN(x,y)       (x < GPIO_PORT_NUMOF) \
                             ? GPIO_PIN_MCU(x, y) \
                             : GPIO_PIN_EXT(x - GPIO_PORT_NUMOF, y))
```
The only change to your PR would be to rename `GPIO_PIN` to `GPIO_PIN_LL (MCU)` in all cpu definitions, which shouldn't be a big thing, and to define `GPIO_PORT_NUMOF` for MCUs with multiple ports. There could be a default definition of 1 port.

For using drivers or application there would change nothing.

drivers/include/periph/gpio.h
```diff
@@ -77,11 +77,33 @@ typedef unsigned int gpio_t;
 #define GPIO_PIN(x,y)       ((gpio_t)((x & 0) | y))
 #endif
 
+#ifndef GPIO_EXT_DEV_LOC
+/**
+ * @brief   Set device ID location in gpio_t when using extensions
+ */
+#define GPIO_EXT_DEV_LOC    (8*sizeof(gpio_t) - 8)
```
gschorcht: Here, we will get a problem for `cpu/atmega_common/include/periph_cpu_common.h`. The following was added with merge of PR #10498:
https://github.com/RIOT-OS/RIOT/blob/2ebb187ca9b5413ef2ef5ab64e53d2e953d22482/cpu/atmega_common/include/periph_cpu_common.h#L37-L40

We should ask @lebrush why he changed the type of `gpio_t` with the PWM implementation. Either this definition has to removed from here or we also need to add a `GPIO_EXT_DEV_LOC` definition in `cpu/atmega_common/include/periph_cpu_common.h`

gschorcht:
@ZetaR60 Any thoughts to my last idea to introduce a macro `GPIO_PORT_NUMOF` and to rename `GPIO_PIN` to `GPIO_PIN_LL` in low level drivers to be able to define a new `GPIO_PIN` macro which allows the seamless integration of GPIO extension devices as ports.
```cpu/*/include/periph_cpu.h
#define GPIO_PIN_LL(x, y) ...
#define GPIO_PORT_NUMOF (x)
```
```peripg/gpio.h
#define GPIO_PIN(x,y)       (x < GPIO_PORT_NUMOF) \
                             ? GPIO_PIN_LL(x, y) \
                             : GPIO_EXT_PIN(x - GPIO_PORT_NUMOF, y))
```

ZetaR60:
@gschorcht

> Any thoughts to my last idea

Sorry for the late reply. I have not been able to be as prompt as I would like to be lately.

> In this case it would become possible to handle GPIO expansion devices based on a port ID starting with GPIO_PORT_NUMOF. In addition, the GPIO_PIN(x, y) macro could be redefined to handle MCU ports and extension ports in the same way. The extension interface would be completely invisible to the application. It always uses GPIO_PIN(port, pin), regardless of whether the GPIO is provided by the MCU or by the expansion device.

The extension interface is already completely invisible, because the compiler does not "see" the macros. I don't think that getting rid of `GPIO_EXT_PIN` would give any advantages in increasing compatibility. I also recall that when working out how to write `GPIO_EXT_PIN` I had major problems on some systems because of weird `gpio_t` encoding methods. Not everything uses ascending port numbers, so I don't think the port encoding idea will be practical. Using the leading bit to indicate an extension device made the encoding agnostic to any MCU weirdness (as apparently no system collided with the leading bit, as determined by `tests/periph_gpio_coll`).

> I'm wondering whether it would possible to define a macro that is set by the CPU peripheral configuration which defines the number of ports provided by the MCU

This would be a useful thing to add to the extension API follow-up macro changes that we have been discussing in #9582.

drivers/include/periph/gpio.h
ZetaR60: Argh. Those changes should have been made in a separate PR. Also note the incompatible `GPIO_UNDEF` definition. This PR needs to be finalized and merged soon to reduce the accumulation of incompatible changes we need to undo.

Setting it back to at least `uint16_t` would be better, but we could get by with using the alternative encoding from `cpu/lpc1768/include/periph_cpu.h`, though this limits the user to 8 extension devices with 16 pins each. The `GPIO_UNDEF` definition needs to be removed.

gschorcht:
> The extension interface is already completely invisible, because the compiler does not "see" the macros.

Yes, for the compiler it is clear. I meant seamless integrated or invisible from the application developer's perspective.

> I don't think that getting rid of GPIO_EXT_PIN would give any advantages in increasing compatibility.

Maybe, we have different perspectives in mind when we are talking about extensions. The application developer uses the `GPIO_PIN` macro either hard-coded or in configuration parameters, such as for drivers. Using `GPIO_EXT_PIN` instead of `GPIO_PIN` requires that the application developer has to know which GPIOs are provided by the MCU and which GPIOs are provided by the extension. I'm wondering, is it always necessary that the application developer has to aware that.

Let's suppose the board developer creates a board where 4 GPIO ports are used from MCU and 2 GPIO ports are provided by an GPIO extensions. He could tell the application developer in the board documentation that the board provides 6 GPIO ports with 8 pins each. The application developer would then use them from `GPIO_PIN(0, x)` ... `GPIO_PIN(5, x)`. This is what I meant when I said seamless integration.

So the question is:
Do you think that from the perspective of application developers, it is always necessary to know what an MCU GPIO is and what an expansion GPIO is? If so, everything is fine the way it is. If not, we should hide `GPIO_EXT_PIN` a bit more from the application developer's point of view, at least optionally.

For me, as an application developer, it would be really cool not to know which GPIOs are provided by the MCU and which GPIOs by extensions.

ZetaR60:
Rebased, changed `*_ll` to `*_cpu` as discussed in #9582, and backported a few of the improvements from @gschorcht 's PRs.

ZetaR60:
@gschorcht When you get a chance, can you update your PRs to change `*_ll` to `*_cpu` ?

gschorcht:
@ZetaR60 Sure, will do it tomorrow.

gschorcht:
@ZetaR60 ADC, DAC and PWM extensions APIs are changed too.

drivers/include/periph/gpio.h
```diff
@@ -133,6 +155,44 @@ typedef struct {
...
+void gpio_irq_enable_cpu(gpio_t pin);
+void gpio_irq_disable_cpu(gpio_t pin);
+int gpio_read_cpu(gpio_t pin);
+void gpio_set_cpu(gpio_t pin);
+void gpio_clear_cpu(gpio_t pin);
+void gpio_toggle_cpu(gpio_t pin);
+void gpio_write_cpu(gpio_t pin, int value);
+/** @} */
+
+#if MODULE_EXTEND_ADC || DOXYGEN
```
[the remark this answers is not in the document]
ZetaR60: D'oh!

gschorcht:
@kaspar030 Sorry I simply approved that my change requests have been considered. As long as it is not squashed it should not happen that someone merges accidentally.

PR #9862, opened 2018-08-30
examples/opensensemap: Add example of TTN and openSenseMap integration

leandrolanzieri:
### Contribution description
This is an example of how to integrate [The Things Network](https://thethingsnetwork.org) with [openSenseMap](https://opensensemap.org) using lora serialization helper to encode the information sent to the gateway.

### Testing procedure
Follow the instructions provided in the README.md file.

### Issues/PRs references
Depends on ~~#9863~~

### DIFF
--- a/examples/opensensemap/README.md
+++ b/examples/opensensemap/README.md
@@ -102,8 +102,8 @@ SenseBox configuration!**
 
     /* Write data to serialization. Replace with your sensors measurements. */
     lora_serialization_write_temperature(&serialization, 25.3);
-    lora_serialization_write_humidity(&serialization, 50.2);
     lora_serialization_write_uint8(&serialization, 82);
+    lora_serialization_write_humidity(&serialization, 50.2);
 ```
 
 ### Build and Flash
--- a/examples/opensensemap/README.md
+++ b/examples/opensensemap/README.md
@@ -1,0 +1,127 @@
+Integration of The Things Network and openSenseMap
+=======
+
+Description
+-----------
+
+This application show how to send data to the openSenseMap platform by
+integrating it with The Things Network.
+
+This application is using the Over-The-Air Activation procedure.
+
+Preparation
+-----------
+### The Things Network
+Go to [the website](https://www.thethingsnetwork.org/) and create an account.
+On the [console section](https://console.thethingsnetwork.org/) you will be
+able to create a new application.Once the application has been created register
+a new device.
+
+Finally on the _Integrations_ section of your application add a new
+'HTTP Integration'. **Make sure that URL points to
+https://ttn.opensensemap.org/v1.1 and that the Method is POST**.
+
+### openSenseMap
+First, you need to register a new box. For that go to the
+[openSenseMap website](https://opensensemap.org/), register and create a new
+senseBox.
+
+You will need to choose the station name, the exposure, location and type of
+hardware (you can choose either of them). Under _Manual configuration_ you can
+add sensors to your box. For this application we added 3 sensors:
+
+1. **Temperature**
+    - Phenomenom: 'Temperature'
+    - Unit: 'C'
+    - Type: 'temperature'
+2. **Battery**
+    - Phenomenon: 'Battery'
+    - Unit: '%'
+    - Type: 'uint8'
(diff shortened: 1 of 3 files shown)


examples/opensensemap/README.md
```diff
@@ -0,0 +1,127 @@
...
+SenseBox configuration!**
+
+```C
+    /* Reset serialization descriptor */
+    lora_serialization_reset(&serialization);
+
+    /* Write data to serialization. Replace with your sensors measurements. */
+    lora_serialization_write_temperature(&serialization, 25.3);
+    lora_serialization_write_humidity(&serialization, 50.2);
+    lora_serialization_write_uint8(&serialization, 82);
```
danpetry: humidity and uint8 should be swapped

leandrolanzieri: Thanks! I swapped it.

examples/opensensemap/README.md
```diff
@@ -0,0 +1,127 @@
...
+First, you need to register a new box. For that go to the
+[openSenseMap website](https://opensensemap.org/), register and create a new
+senseBox.
+
+You will need to choose the station name, the exposure, location and type of
+hardware (you can choose either of them). Under _Manual configuration_ you can
+add sensors to your box. For this application we added 3 sensors:
+
+1. **Temperature**
+    - Phenomenom: 'Temperature'
```
danpetry: Phenomenon

examples/opensensemap/README.md
```diff
@@ -0,0 +1,127 @@
...
+accepted types
+[documented here](https://github.com/sensebox/ttn-osem-integration).
+
+Once the box is created go to the dashboard and click the 'Edit' button. On the
+'TheThingsNetwork' section choose:
+
+- **Decoding Profile**: LoRa serialization
+- **TTN application ID**: (found on the 'Application Overview' section)
+- **TTN device ID**: (found on the 'Device Overview' section as Device ID)
+- **Port**: (LoRa port where the data is being sent to The Things Network)
```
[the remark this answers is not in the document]
leandrolanzieri: Added this to description.

examples/opensensemap/README.md
leandrolanzieri: Fixed.

leandrolanzieri:
Done

examples/opensensemap/main.c
```diff
@@ -0,0 +1,151 @@
...
+    /* Reset serialization descriptor */
+    lora_serialization_reset(&serialization);
+
+    /* Write data to serialization. Replace with your sensors measurements.
+     * Keep in mind that the order in which the data is written into the
+     * serialization needs to match the decoders order.
+     */
+    lora_serialization_write_temperature(&serialization, 25.3);
+    lora_serialization_write_uint8(&serialization, 82);
+    lora_serialization_write_humidity(&serialization, 50.2);
```
danpetry: I see that this is very similar to the lorawan example. My thoughts are however that this example could be designed more towards presenting a clean, extensible interface to the user. Personally I like the structure of e.g. the gcoap and dtls-echo examples - they put a simple interface to the program in main.c. Then, if people want to change other aspects of how the program responds, they can go into the other .c files.

I would suggest finding a program structure so that in main.c they can do something like:
- add measurements to be sent in a declarative manner in a simple macro or function (for Sensebox this could also add the drivers at compile time, not sure whether this would require a separate macro/function or whether it could be done automatically). They could perhaps specify that the application sends dummy data (as above), or direct from sensor, or from a custom function
- change the frequency of data sending

Then in another file (loramac_send.c?) 
- asyncronous data sending stuff incl thread definition perhaps
- arrays storing keys

Basically the philosophy I'm suggesting is to present a really simple API to a user in main.c which allows them to quickly, easily and flexibly write data sending applications according to their own hardware setup. As easy as possible. E.g. as little as possible above just plugging together their sensebox and loading this application.

leandrolanzieri:
Discussed offline with @danpetry and it would be better if this takes the form of an application on the [Applications repo](https://github.com/RIOT-OS/applications) adding some sensor reading with SAUL.

PR #9864, opened 2018-08-30
drivers/netdev_ieee802154: cleanup NETOPT_MAX_PACKET_SIZE

SemjonKerner:
This PR relieves code duplications in netdev interfaces and unifies them in the shared, generic ieee802154 netdev driver.
I focused on NETOPT_MAX_PACKET_SIZE by intent, if there are more obvious duplications, I will fix those separately.

### DIFF
--- a/cpu/native/socket_zep/socket_zep.c
+++ b/cpu/native/socket_zep/socket_zep.c
@@ -304,21 +304,8 @@ static int _init(netdev_t *netdev)
 
 static int _get(netdev_t *netdev, netopt_t opt, void *value, size_t max_len)
 {
-    socket_zep_t *dev = (socket_zep_t *)netdev;
-    uint16_t *v = value;
-
-    assert((dev != NULL));
-    switch (opt) {
-        case NETOPT_MAX_PACKET_SIZE:
-            assert(value != NULL);
-            if (max_len != sizeof(uint16_t)) {
-                return -EOVERFLOW;
-            }
-            *v = SOCKET_ZEP_FRAME_PAYLOAD_LEN;
-            return sizeof(uint16_t);
-        default:
-            return netdev_ieee802154_get(&dev->netdev, opt, value, max_len);
-    }
+    assert(netdev != NULL);
+    return netdev_ieee802154_get((netdev_ieee802154_t *)netdev, opt, value, max_len);
 }
 
 static int _set(netdev_t *netdev, netopt_t opt, const void *value,
--- a/drivers/netdev_ieee802154/netdev_ieee802154.c
+++ b/drivers/netdev_ieee802154/netdev_ieee802154.c
@@ -151,6 +151,13 @@ int netdev_ieee802154_get(netdev_ieee802154_t *dev, netopt_t opt, void *value,
             res = sizeof(l2filter_t **);
             break;
 #endif
+        case NETOPT_MAX_PACKET_SIZE:
+            assert(max_len >= sizeof(int16_t));
+            *((uint16_t *)value) = (IEEE802154_FRAME_LEN_MAX -
+                                  IEEE802154_MAX_HDR_LEN) -
+                                  IEEE802154_FCS_LEN;
+            res = sizeof(uint16_t);
+            break;
         default:
             break;
     }
--- a/cpu/cc2538/radio/cc2538_rf_netdev.c
+++ b/cpu/cc2538/radio/cc2538_rf_netdev.c
@@ -31,8 +31,6 @@
 #define ENABLE_DEBUG        (0)
 #include "debug.h"
 
-#define _MAX_MHR_OVERHEAD   (25)
-
 /* Reference pointer for the IRQ handler */
 static netdev_t *_dev;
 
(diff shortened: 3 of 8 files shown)


SemjonKerner:
hey @bergzand are you still looking into this?

drivers/netdev_ieee802154/netdev_ieee802154.c
```diff
@@ -151,6 +151,13 @@ int netdev_ieee802154_get(netdev_ieee802154_t *dev, netopt_t opt, void *value,
             res = sizeof(l2filter_t **);
             break;
 #endif
+        case NETOPT_MAX_PACKET_SIZE:
+            assert(max_len >= sizeof(int16_t));
+            *((uint16_t *)val) = (IEEE802154_FRAME_LEN_MAX -
```
bergzand: `/home/hydrazine/dev/RIOT/drivers/netdev_ieee802154/netdev_ieee802154.c:156:27: error: 'val' undeclared (first use in this function)`

I guess this should be `value` instead ;)

SemjonKerner: Oh oh. I fixed this right after PR, but i forgot to push it, sorry.

bergzand: No worries.

In the future please use a fixup commit instead of force pushing. Using fixup commits makes it easy for the reviewer to track changes between revisions  and they can be automatically squashed before merging.

SemjonKerner: You are right. This was meant to go into the PR and i just missed the push, this was allready fixed before and i just pushed it now without touching it, thats why i commented it. Sorry for the inconvenience :)

bergzand: It's fine, luckily for me this PR is not that big :)

SemjonKerner:
@bergzand good to go?

cpu/native/socket_zep/socket_zep.c
```diff
@@ -304,21 +304,8 @@ static int _init(netdev_t *netdev)
...
-            if (max_len != sizeof(uint16_t)) {
-                return -EOVERFLOW;
-            }
-            *v = SOCKET_ZEP_FRAME_PAYLOAD_LEN;
-            return sizeof(uint16_t);
-        default:
-            return netdev_ieee802154_get(&dev->netdev, opt, value, max_len);
-    }
+    assert(netdev != NULL);
+    return netdev_ieee802154_get(netdev, opt, value, max_len);
```
bergzand: Why not remove this function alltogether and set the `.get` [here] to `netdev_ieee802154_get`?

[here]: https://github.com/RIOT-OS/RIOT/blob/9b1d91202fd5d5d77cb0e386b80395a51ad114ea/cpu/native/socket_zep/socket_zep.c#L324

SemjonKerner: I'd say for two reasons, because i kept it similar to the _set function but mainly because the `assert(netdev != NULL);` is not done in `netdev_ieee802154_get`.

bergzand: Good enough for me, I can also imagine that we're going to need this function in the future again.

bergzand: I just want to test if everything still works at runtime. Changes still look good to me.

PeterKietzmann:
To track some `ping6` based tests that I'm gonna provide next:
- [ ] at86rf233 - kw2xrf
- [ ] mrf24j40 - cc2538
- [ ] at86rf233 - cc2420

SemjonKerner:
Oh, you can proceed testing, but maybe i can get rid of some more defines. On first glance it seems that I may also delete `KW2XRF_MAX_PKT_LENGTH`, `CC2420_PKT_MAXLEN`, `AT86RF2XX_MAX_PKT_LENGTH`, `CC2538_RF_MAX_DATA_LEN`. I'll have to get into this later today.

PeterKietzmann:
Then please add the WIP label next time!!! I will stop testing now. Faced other problems when putting my hands on the cc2420.

SemjonKerner:
> Then please add the WIP label next time!!! I will stop testing now. Faced other problems when putting my hands on the cc2420.

@PeterKietzmann Can you please specify the problems?
This PR was not supposed to be WIP, opposing to that I expect it to work properly.

Also it was not obvious to me that the named defines are redundant because, even though they are just the same as `IEEE802154_FRAME_LEN_MAX`, they all are used and hence substituting them did not appeal to me in the first place.
When i thought about it earlier today, I had another view on the code and changed my mind.
If this is a concern for anyone i'd prefer to provide another cleanup PR for that so this will not interfere with anything presented here.

PeterKietzmann:
Unrelated to your changes, see #10043

bergzand:
@PeterKietzmann Am I correct when I conclude that this PR doesn't break/impact connectivity and packet loss? To prevent this PR from getting stuck, how about we merge this and assume that @SemjonKerner is going to provide a cleanup in the near future to remove the rest of the obsolete defines.

@SemjonKerner Does this make sense? Is this okay with you?

SemjonKerner:
@bergzand I don't expect this PR to be the reason for the observed problems with packet loss, and i think the tests suggest the same. So yes, this makes sense to me.

bergzand:
@SemjonKerner Murdock is pointing out a few build errors with `socket_zep`

SemjonKerner:
@bergzand fixed. Just a missing typecast.

PeterKietzmann:
> Oh, you can proceed testing, but maybe i can get rid of some more defines

This let me conclude you were about to continue working on this PR. The problems I pointed out don't relate to this PR. As one can see, results w/ and w/o are "the same". I won't block this PR.

bergzand:
> This let me conclude you were about to continue working on this PR. The problems I pointed out don't relate to this PR. As one can see, results w/ and w/o are "the same". I won't block this PR.

Thanks, and especially thanks for the thorough test you did on this!

@SemjonKerner Please squash

bergzand:
Thanks for the cleanup @SemjonKerner!

PR #9867, opened 2018-08-30
make: unify app folder search (examples/*, tests/*, ...)

kaspar030:
### Contribution description

There were at least three variants of getting a list of examples and tests application folders at even more places in the code.

This PR tries to unify this.

### Testing procedure

1. murdock total job count shouldn't change
2. complile_test output looks as before
3. main folder make clean / distclean works as before

### DIFF
--- a/dist/tools/compile_test/compile_test.py
+++ b/dist/tools/compile_test/compile_test.py
@@ -21,11 +21,14 @@
 
 from __future__ import print_function
 
+from collections import defaultdict
 from itertools import groupby
-from os import devnull, environ, listdir
+from os import devnull, environ
 from os.path import abspath, dirname, isfile, join
 from subprocess import CalledProcessError, check_call, check_output, PIPE, Popen
-from sys import exit, stdout, argv
+from sys import argv, exit, stdout
+
+
 try:
     # Python 2.x
     from StringIO import StringIO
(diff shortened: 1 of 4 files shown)


kaspar030:
@cladmi didn't you have a PR for this somewhere already?

dist/tools/compile_test/compile_test.py
```diff
@@ -83,12 +85,31 @@ def get_results_and_output_from(fd):
         elif read_more_output:
             output.write(line)
 
+def get_app_dirs():
+    yield from check_output(["make", "-f", "makefiles/app_dirs.inc.mk", "info-apps"]) \
```
aabadie: Any reason for using `yield from` ?

aabadie:
@kaspar030, this needs rebase now that #9868 is merged.

dist/tools/compile_test/compile_test.py
```diff
@@ -86,12 +89,33 @@ def get_results_and_output_from(fd):
...
+    { "examples": ["hello_world", "gnrc_networking" ],
+      "tests": ["minimal", "fmt_print" ]
+      }
+    """
+    res = defaultdict(list)
+    for app_dir in app_dirs:
+        folder, app = app_dir.split("/", 1)
+        res[folder].append(app)
+
+    return res
```
cladmi: I think it is better to return a normal dict `return dict(res)` as using a default dict without knowing it is one can lead to trouble.

cladmi: I do not find any reference about it tough.

cladmi:
The other things I had in my branch and that I can propose with upcoming PRs is:

* having a `APPLICATION_DIRS_REGEX` variable
* `APPLICATION_DIRS_EXCLUDE` to filter-out something. For testing, I often want to run everything, except one I know is broken or that works but takes too long.

Something I am thinking about but not drafted yet, is move the `examples/*` and `tests/*` definition to files in the `examples` and `tests` directories. With the goal that each package could list his tests.

makefiles/app_dirs.inc.mk
```diff
@@ -0,0 +1,14 @@
...
+# 1. use wildcard to find Makefiles
+# 2. use patsubst to drop trailing "/"
+# 3. use patsubst to drop possible leading "./"
+# 4. sort
+APP_DIRS := $(sort $(patsubst ./%,%,$(patsubst %/,%,$(dir $(wildcard \
+	$(RIOTBASE)/examples/*/Makefile \
+	$(RIOTBASE)/tests/*/Makefile    \
+	)))))
+
+info-apps:
```
cladmi: Same here, I would use `applications`.

dist/tools/compile_test/compile_test.py
kaspar030: then let's keep it.

makefiles/app_dirs.inc.mk
kaspar030: ok

dist/tools/compile_test/compile_test.py
```diff
@@ -86,12 +89,33 @@ def get_results_and_output_from(fd):
             output.write(line)
 
 
+def get_app_dirs():
+    yield from check_output(["make", "-f", "makefiles/app_dirs.inc.mk", "info-applications"]) \
```
cladmi: The script is currently supposed to be python2 compatible and `python2` is used by default on ubuntu because of the shebang. `yield from` breaks this compatibility.

Also here, as `split` returns a list and not a generator, `yield from` adds nothing.
Even `check_output` and `decode` are not generator.

kaspar030: ok, it is just returning now.

kaspar030:
> @aabadie please tell when you agree to squash.

As @aabadie's comment was addressed, I already squashed. Let's still give @aabadie a chance to take a last look.

aabadie:
Nothing else to add from my side. Let's merge this one.

PR #9870, opened 2018-08-31
ethos: Add drop frame case to recv function

bergzand:
### Contribution description

The ethos driver does not drop the received frame if the recv function
is called with NULL buffer and with a length. This PR fixes that.

### Issues/PRs references

depends on #9869


bergzand:
@kaspar030 ping?

bergzand:
Anybody willing to ACK this?

bergzand:
Thanks!

PR #9872, opened 2018-08-31
net/nanocoap: allow empty uri/location path option

haukepetersen:
### Contribution description
For CoAP, there is actually a difference between `/some/path` and `/some/path/`. This needs to be reflected when parsing the URI and location path options from a given string.

This PR fixes `nanocoap` to allow appending the trailing `/`, which is for CoAP an empty option field.

I am unsure what to do when parsing query strings, as for that case IMHO an empty option does not really make sense. That is why I only skip the length check in case the separation character is a `/`...

Background: I encountered this problem with interfacing the the `aiocoap` resource directory, where it expects enpoints to update their entry on a resource that has to end with a `/`...

### Testing procedure
For example:
- run the `examples/gcoap` application on native
- make wireshark listen on the tap bridge, possibly filter for coap message
- make a simple GET request for some random resource, e.g.
```
coap get ff02::1 5683 /foo/bar
```
- look at the request in wireshark, it should have to uri-path option fields set
- make the request again, this time with a trailing `/`:
```
coap get ff02::1 5683 /foo/bar/
```
- now wireshark should show the request with a third, empty uri-path option field.

### Issues/PRs references
none

### DIFF
--- a/sys/net/application_layer/nanocoap/nanocoap.c
+++ b/sys/net/application_layer/nanocoap/nanocoap.c
@@ -603,7 +603,9 @@ size_t coap_opt_put_string(uint8_t *buf, uint16_t lastonum, uint16_t optnum,
 
         part_len = (uint8_t *)uripos - part_start;
 
-        if (part_len) {
+        /* Creates empty option if part for Uri-Path or Uri-Location contains only *
+         * a trailing slash, except for root path ("/"). */
+        if (part_len || ((separator == '/') && (lastonum == optnum))) {
             bufpos += coap_put_option(bufpos, lastonum, optnum, part_start, part_len);
             lastonum = optnum;
         }
@@ -658,7 +660,9 @@ ssize_t coap_opt_add_string(coap_pkt_t *pkt, uint16_t optnum, const char *string
 
         part_len = (uint8_t *)uripos - part_start;
 
-        if (part_len) {
+        /* Creates empty option if part for Uri-Path or Uri-Location contains
+         * only a trailing slash, except for root path ("/"). */
+        if (part_len || ((separator == '/') && write_len)) {
             if (pkt->options_len == NANOCOAP_NOPTS_MAX) {
                 return -ENOSPC;
             }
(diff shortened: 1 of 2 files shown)


kb2ma:
Thanks for addressing this problem. [This thread](https://mailarchive.ietf.org/arch/msg/core/YyuHQGZ4ALWGSNOUkCzOjheVq_c) on the core ML confirms the resolution. Now on to reviewing and testing the solution.

kb2ma:
@haukepetersen, in your resource directory example, what is the semantic difference between two URIs, where the only difference in the URIs is the addition of the trailing slash?

haukepetersen:
The RD (`aiocoap-rd`) expects the client to POST an empty message to a given registration resource (e.g. `rd/1/`) when updating its entry. And here the RD is very strict with the semantics, as it does not accept a POST message to `/rd/1`. The author did this on purpose, to make sure clients handle the trailing slash correctly as apparently specified by the CoAP RFC...

kb2ma:
I just realized that the base for this PR predates the merge of #8920. Please rebase.

I've read through the URI RFC (3986) and reviewed 7252. I think this PR needs a couple of additions, and I'll point those out in the code after the rebase.

haukepetersen:
rebased

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -603,7 +603,7 @@ size_t coap_opt_put_string(uint8_t *buf, uint16_t lastonum, uint16_t optnum,
 
         part_len = (uint8_t *)uripos - part_start;
 
-        if (part_len) {
```
kb2ma: Need to accommodate a request for root path, "/". RFC 7252 sec. 6.4 step 8 says there should not be an option for this path.

kb2ma:
I also experimented with how nanocoap as a server handles a request with a trailing slash. It makes a distinction. So if the server defines `/mypath` but not `/mypath/`, the server will reject a request for `/mypath/` with a 4.04 response. I'm OK with this, because really I think it's an error in the request, and allows us to make this distinction between paths if necessary.

haukepetersen:
>  I'm OK with this, because really I think it's an error in the request, and allows us to make this distinction between paths if necessary.

+1!

haukepetersen:
Alright, merged the additional fix + unit test into this PR (thanks to @kb2ma) and squashed the commits accordingly.

Issue #9874, opened 2018-09-02
i2c SAML21 using sercom5 gets stuck at poweron


fedepell:
#### Description
On SAML21 based boards, for example saml21-xpro, if I2C is configured to use sercom5 it will get stuck forever during initialization in _i2c_poweron at:

```
 _i2c_poweron (dev=0)
     at xxx/RIOT/cpu/sam0_common/periph/i2c.c:262
 262	    while (bus(dev)->SYNCBUSY.bit.ENABLE) {}
``` 

The condition is never met and code is stuck there forever.

I see that sercom5 handling is slightly different from the others (has a separate clock) so maybe some detail is missing there. Using a different sercomX works as expected, but with specific pinout usage sercom5 needs to be used.

#### Steps to reproduce the issue
Configure the I2C to use sercom5, that is:
```
static const i2c_conf_t i2c_config[] = {
    {
        .dev      = &(SERCOM5->I2CM),
        .speed    = I2C_SPEED_NORMAL,
        .scl_pin  = GPIO_PIN(PB, 31),
        .sda_pin  = GPIO_PIN(PB, 30),
        .mux      = GPIO_MUX_D,
        .gclk_src = GCLK_PCHCTRL_GEN_GCLK0,
        .flags    = I2C_FLAG_NONE
    }
};
```
And add the i2c module to your application:
`USEMODULE += periph_i2c`

#### Expected results
I2C gets correctly initialized.

#### Actual results
I2C initialization stuck forever blocking whole RIOT OS.

#### Versions
arm-none-eabi-gcc: arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2018-q2-update) 7.3.1 20180622 (release) [ARM/embedded-7-branch revision 261907]

dylad:
feel free to close your issues when your PRs get merged :)
Thanks for spotting and fixing these bugs !

fedepell:
Now that #9876 and #9877 have been merged the issue can be closed. I also checked that I2C reads correctly from a I2C configured on SERCOM5.

PR #9877, opened 2018-09-02
saml21: fix sercom_set_gen for sercom5

fedepell:
### Contribution description
While debugging #9874 on the SAML21 architecture I noticed that the SERCOM5 gen clock was not correctly activated. Analyzing the code I noticed that sercom_set_gen assumes that SERCOM5_GCLK_ID_CORE is in sequence with the other SERCOMx in this code:
```
static inline void sercom_set_gen(void *sercom, uint32_t gclk)
{
#if defined(CPU_FAM_SAMD21)
    GCLK->CLKCTRL.reg = (GCLK_CLKCTRL_CLKEN | gclk |
                         (SERCOM0_GCLK_ID_CORE + sercom_id(sercom)));
    while (GCLK->STATUS.reg & GCLK_STATUS_SYNCBUSY) {}
#elif defined(CPU_FAM_SAML21)
    GCLK->PCHCTRL[SERCOM0_GCLK_ID_CORE + sercom_id(sercom)].reg =
#endif
}
```

 but actually it is not:

```
saml21/include/instance/sercom0.h:#define SERCOM0_GCLK_ID_CORE        18      
saml21/include/instance/sercom1.h:#define SERCOM1_GCLK_ID_CORE        19      
saml21/include/instance/sercom2.h:#define SERCOM2_GCLK_ID_CORE        20      
saml21/include/instance/sercom3.h:#define SERCOM3_GCLK_ID_CORE        21      
saml21/include/instance/sercom4.h:#define SERCOM4_GCLK_ID_CORE        22      
saml21/include/instance/sercom5.h:#define SERCOM5_GCLK_ID_CORE        24   
```
And therefore the function touches the wrong clock and the device will not work.

### Testing procedure
Use a device on SERCOM5 and see that it will not work correctly, see for example #9874.

### Issues/PRs references
See also #9874.

### DIFF
--- a/cpu/sam0_common/include/periph_cpu_common.h
+++ b/cpu/sam0_common/include/periph_cpu_common.h
@@ -352,8 +352,13 @@ static inline void sercom_set_gen(void *sercom, uint32_t gclk)
                          (SERCOM0_GCLK_ID_CORE + sercom_id(sercom)));
     while (GCLK->STATUS.reg & GCLK_STATUS_SYNCBUSY) {}
 #elif defined(CPU_FAM_SAML21)
-    GCLK->PCHCTRL[SERCOM0_GCLK_ID_CORE + sercom_id(sercom)].reg =
+    if (sercom_id(sercom) < 5) {
+        GCLK->PCHCTRL[SERCOM0_GCLK_ID_CORE + sercom_id(sercom)].reg =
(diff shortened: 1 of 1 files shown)


fedepell:
Done, thanks! Sorry for the long title, got it from the git commit ;)

PR #9878, opened 2018-09-02
cpu: efm32: add support for 32-bit timers

basilfx:
### Contribution description

Newer EFM32 chips have dedicated 32-bit timers (they are register compatible). This PR allows one to make use of the 32-bit timers, which are already supported by the vendor library.

### Testing procedure
Take the SLSTK3402a board, and change the timer configuration as follows (originally, it was 250000 Hz and 16 bit):

```c
#define XTIMER_HZ           (1000000UL)
#define XTIMER_WIDTH        (32)
```

Then change the timer configuration to

```c
static const timer_conf_t timer_config[] = {
    {
        {
            .dev = WTIMER0,
            .cmu = cmuClock_WTIMER0
        },
        {
            .dev = WTIMER1,
            .cmu = cmuClock_WTIMER1
        },
        .irq = WTIMER1_IRQn
    }
};

#define TIMER_NUMOF         PERIPH_NUMOF(timer_config)
#define TIMER_0_ISR         isr_wtimer1
```

Observe that any timer test will still compile and work as intended.

### Issues/PRs references
None

### DIFF
--- a/cpu/efm32/periph/timer.c
+++ b/cpu/efm32/periph/timer.c
@@ -38,6 +38,20 @@
  */
 static timer_isr_ctx_t isr_ctx[TIMER_NUMOF];
 
+/**
+ * @brief   Check whether device is a using a WTIMER device (32-bit)
+ */
+static inline bool _is_wtimer(timer_t dev)
+{
+#if defined(WTIMER_COUNT) && WTIMER_COUNT > 0
+    return ((uint32_t) timer_config[dev].timer.dev) >= WTIMER0_BASE;
+#else
+    (void) dev;
+
+    return false;
+#endif
+}
+
 int timer_init(tim_t dev, unsigned long freq, timer_cb_t callback, void *arg)
 {
     TIMER_TypeDef *pre, *tim;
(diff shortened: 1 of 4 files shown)


cpu/efm32/periph/timer.c
```diff
@@ -106,12 +120,12 @@ int timer_set_absolute(tim_t dev, int channel, unsigned int value)
         return -1;
     }
 
-    if (value > 0xffff) {
+    if (value > TIMER_TopGet(timer_config[dev].timer.dev)) {
```
kYc0o: Why not `0xFFFFFFFF` like before?

kYc0o:
Compile-wise it works for slstk3402a with the provided instructions, I even found that the ROM size is reduced by ~200B and RAM ~4B.

Do you mind to enable it by default on the supported boards?

Can you paste test results from xtimer_* tests? (not all of them but a sensible subset, e.g. 4 tests).

cpu/efm32/periph/timer.c
kYc0o: Thanks for the explanation @chrysn, I'd prefer then the change as is it right now.

kYc0o:
Excellent! Thanks @basilfx! You might squash and wait for Murdock's approval.

kYc0o:
ACK.

basilfx:
Thanks @kYc0o!

PR #9879, opened 2018-09-02
cpu: efm32: add support for no DC-DC converter

basilfx:
### Contribution description
EFM32 CPUs have an integrated DC-DC converter for optimizing power consumption for the IO part of the chip. This DC-DC converter must be disabled if it is not used (e.g. chip doesn't use battery power).

Current EFM32 board with DC-DC convert have this feature enabled, but for my own custom board, I need to be able to disable it.

### Testing procedure
Take any EFM32 board with a MCU that support DC-DC converter (e.g. the SLTB001a) and change the DC-DC configuration in `periph_conf.h` to:

```c
/**
 * @name    DC-DC configuration
 * @{
 */
#define EMU_DCDCINIT_OFF
/** @} */
```

It should still compile, and the board should still work (but uses more power).

### Issues/PRs references
None

### DIFF
--- a/boards/slstk3401a/include/periph_conf.h
+++ b/boards/slstk3401a/include/periph_conf.h
@@ -55,6 +55,15 @@ extern "C" {
 /** @} */
 
 /**
+ * @name    DC-DC configuration
+ * @{
+ */
+#ifdef EMU_DCDCINIT_OFF
+#error "This option will soft-brick your board. Please remove it."
+#endif
+/** @} */
+
+/**
  * @name    ADC configuration
  * @{
  */
--- a/boards/sltb001a/include/periph_conf.h
+++ b/boards/sltb001a/include/periph_conf.h
@@ -55,6 +55,15 @@ extern "C" {
 /** @} */
 
 /**
+ * @name    DC-DC configuration
+ * @{
+ */
+#ifdef EMU_DCDCINIT_OFF
+#error "This option will soft-brick your board. Please remove it."
+#endif
+/** @} */
+
+/**
  * @name    ADC configuration
  * @{
  */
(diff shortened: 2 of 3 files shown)


kYc0o:
Just tested in sltb001a and it doesn't work... However I notice a decrease on ROM usage of ~1KB.

kYc0o:
And... I think by disconnecting the dc-dc converter I bricked my board 😅

kYc0o:
If ever happens to someone else, I unbricked it with this tutorial: https://www.silabs.com/community/mcu/32-bit/knowledge-base.entry.html/2014/05/21/unlock_a_brickede-xkAm

basilfx:
@kYc0o It seems that the `sltb001a` and the `slstk3401a` rely on it's DC-DC converter, so it cannot be turned off. Just tested this with the `slstk3402a` and that one works fine.

It seems that in the EFM32PG12B (on the `slstk3402a`) they have [changed the DC-DC](https://www.silabs.com/documents/public/application-notes/an0948-power-configurations-and-dcdc.pdf) a bit. See also [here](https://devtools.silabs.com/dl/documentation/doxygen/5.5/efr32bg13/html/group__EMU.html#ga5625f85c3d89aba15e151ed8d1b701af)

basilfx:
@kYc0o are you OK with the explanation?

kYc0o:
Yes, thanks @basilfx ! However I need to test this somehow... Do you know if I can use my STK3700?

basilfx:
No, unfortunately the STK3700 does not have a DC-DC integrated.

kYc0o:
Can we `#ifdef` then the unsupported boards? This would avoid bricking them by accident.

kYc0o:
Yes, that's perfect for me.

basilfx:
Added this to the PR.

PR #9880, opened 2018-09-02
boards: sltb001a: add docs

basilfx:
### Contribution description
This PR adds docs for the SLTB001a.

### Issues/PRs references
#9791

PR #9881, opened 2018-09-02
boards: slstk3401a: add docs

basilfx:
### Contribution description
This PR adds docs for the SLSTK3401a.

### Issues/PRs references
#9791

### DIFF
--- a/boards/slstk3401a/doc.txt
+++ b/boards/slstk3401a/doc.txt
@@ -1,0 +1,214 @@
+/**
+@defgroup    boards_slstk3401a Silicon Labs SLSTK3401A starter kit
+@ingroup     boards
+@brief       Support for Silicon Labs SLSTK3401A starter kit
+
+## Overview
+Silicon Labs EFM32 Pearl Gecko PG1 Starter Kit is equipped with the EFM32 microcontroller. It is specifically designed for low-power applications, having energy-saving peripherals, different energy modes and short wake-up times.
+
+The starter kit is equipped with an Advanced Energy Monitor. This allows you to actively measure the power consumption of your hardware and code, in real-time.
+
+## Hardware
+
+### MCU
+| MCU             | EFM32PG1B200F256GM48                                 |
+|-----------------|------------------------------------------------------|
+| Family          | ARM Cortex-M4F                                       |
+| Vendor          | Silicon Labs                                         |
+| Vendor Family   | EFM32 Pearl Gecko 1B                                 |
+| RAM             | 32.0KB                                               |
+| Flash           | 256.0KB                                              |
+| EEPROM          | no                                                   |
+| Frequency       | up to 40 MHz                                         |
+| FPU             | yes                                                  |
+| MPU             | yes                                                  |
+| DMA             | 8 channels                                           |
+| Timers          | 2x 32-bit + 2x 16-bit + 1x 16-bit (low power)        |
+| ADCs            | 12-bit ADC                                           |
+| UARTs           | 4x USART, 1x UART, 1x LEUART                         |
+| SPIs            | 4x USART                                             |
+| I2Cs            | 2x                                                   |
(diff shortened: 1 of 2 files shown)


boards/slstk3401a/doc.txt
```diff
@@ -0,0 +1,214 @@
+/**
+@defgroup    boards_slstk3401a Silicon Labs SLSTK3401A starter kit
+@ingroup     boards
+@brief       Support for Silicon Labs SLSTK3401A starter kit
+
+## Overview
+Silicon Labs EFM32 Pearl Gecko PG1 Starter Kit is equipped with the EFM32 microcontroller. It is specifically designed for low-power applications, having energy-saving peripherals, different energy modes and short wake-up times.
```
aabadie: It would be better to split the textual lines to have them fit within 80 characters width. For the large tables below, this rule can be skipped I think.

basilfx: I don't mind to change this, but #9791 and #9880 that have the same problem are already merged :-)

@jia200x What do you think?

jia200x: I personally think writing documentation is different to write code. In general, documentation is meant to be written in Markdown and read in the web browser or PDF.

The 80 characters width makes sense to me in actual code where hard-to-read code is error prone and it's based on instructions, but writing documentation should be closer IMO to write any kind of document.

What do you think?

kaspar030: > The 80 characters width makes sense to me in actual code where hard-to-read code is error prone and it's based on instructions, but writing documentation should be closer IMO to write any kind of document.

Longer lines usually are also harder to edit, and they need to be read while editing, so keeping the lines short makes sense for the same reasons as for code.

jia200x: > Longer lines usually are also harder to edit, and they need to be read while editing, so keeping the lines short makes sense for the same reasons as for code.

Yes, it's true those are harder to edit, but it's the same for every standard documents opened in a text editor. To me the benefits of keeping it 80 chars long doesn't pay off the overhead.

jia200x: (well, it seems other projects are using 80 chars limit for documentation. I'm OK with enforcing 80 lines then besides my thoughts ;)

gschorcht: I would like to reopen the discussion since I discussed exactly this issue with @aabadie  for another PR.

I completely agree with Jose. Breaking lines after 80 characters for source code is a must and it makes sense because you want to see the structure of the code. But  IMHO it makes absolutely  no sense for documentation where you just have floating text and no structure issues. The documentation is something more like text document in an office program.

It is a quite frustrating and time-consuming task to redo all the line breaks of a complete paragraph because of a small change in only one line, e.g., a typo. You all know how hard it is to change the documentation of functions in header files when something changes.

I have experienced it when I wrote the comprehensive board documentations for the ESP32 boards in the new doxygen approach. In the old wiki approach, nobody would have the idea to limit floating text lines to 80 characters.

Finally, there is always the possibility to reduce the width of the editor window, or let the editor interrupt the lines temporarily.

I'm afraid that it will reduce my motivation to write documentation further, if the 80 characters limit will also be applied to doc.txt. The result will be short and unusable documents.

jia200x: could we open an issue with this discussion? There are several on-going documentations that would benefit if there's consensus here.

jia200x:
@aabadie can we merge this one and adapt later when the 80 chars discussion gets resolved? I think it will be beneficial for the users to have this in the documentation.

jia200x:
great. Thank you!

jia200x:
& GO.

boards/slstk3401a/doc.txt
PeterKietzmann: @jia200x, @gschorcht did you ever open an issue for that?

PR #9883, opened 2018-09-03
boards/sensebox_samd21: Use PA_BOOST option on SX127X radios

leandrolanzieri:
### Contribution description
This modifies the board configuration of the SX127X driver. Due to hardware limitations the PA_BOOST option should be used to get a decent SNR on the packages and reduce the considerably high amount of package loses.


jia200x:
I confirm this fixes the issue in the Sense Box SAMD21. Otherwise, packet reception was below SNR -6 and always got CRC FAIL 100% in the gateway. Let's wait Murdock

PR #9887, opened 2018-09-05
driver/enc28j60: cleanup

smlng:
### Contribution description

Some cleanup and optimisation of the ethernet driver. Safes overall 12 bytes, too.

### DIFF
--- a/drivers/enc28j60/enc28j60.c
+++ b/drivers/enc28j60/enc28j60.c
@@ -34,6 +34,10 @@
 #define ENABLE_DEBUG    (0)
 #include "debug.h"
 
+#define SPI_BUS         (dev->p.spi)
+#define CS_PIN          (dev->p.cs_pin)
+#define INT_PIN         (dev->p.int_pin)
+#define RST_PIN         (dev->p.rst_pin)
 /**
  * @brief   Amount of time to hold the reset pin low on reset
  */
@@ -90,9 +94,9 @@ static void switch_bank(enc28j60_t *dev, int8_t bank)
         return;
     }
     /* clear old value */
-    spi_transfer_reg(dev->spi, dev->cs_pin, (CMD_BFC | REG_ECON1), 0x03);
+    spi_transfer_reg(SPI_BUS, CS_PIN, (CMD_BFC | REG_ECON1), 0x03);
     /* set new value */
-    spi_transfer_reg(dev->spi, dev->cs_pin, (CMD_BFS | REG_ECON1), bank);
+    spi_transfer_reg(SPI_BUS, CS_PIN, (CMD_BFS | REG_ECON1), bank);
     /* remember active bank */
     dev->bank = bank;
 }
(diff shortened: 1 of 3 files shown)


drivers/enc28j60/enc28j60.c
```diff
@@ -34,6 +34,10 @@
 #define ENABLE_DEBUG    (0)
 #include "debug.h"
 
+#define SPI_BUS         (dev->p.spi)
```
kaspar030: I don't think this is any cleaner. It also makes it much harder to change code using tools, compared to not using a define.

In order to reduce duplication, you could use per-function variables.

drivers/include/enc28j60.h
```diff
@@ -39,21 +39,17 @@ typedef struct {
...
-    netdev_t netdev;        /**< pull in the netdev fields */
-    spi_t spi;              /**< SPI bus the transceiver is connected to */
-    gpio_t cs_pin;          /**< pin connected to the CHIP SELECT line */
-    gpio_t int_pin;         /**< pin connected to the INT line */
-    gpio_t reset_pin;       /**< pin connected to the RESET line */
-    mutex_t devlock;        /**< lock the device on access */
-    int8_t bank;            /**< remember the active register bank */
-    uint32_t tx_time;       /**< last transmission time for timeout handling */
+    netdev_t netdev;            /**< pull in the netdev fields */
+    enc28j60_params_t p;        /**< SPI and pin confiuration */
```
kaspar030: could this be made a const pointer?

smlng: mhm good point, but currently its done this way in all drivers. Nevertheless, it should safe bytes and time to simply use a (const) pointer here. But when we don't copy here anymore, is this safe memory-wise? I mean if we use the params headers with arrays it sure is but one could pass a params struct and though it away afterwards, or not?

drivers/enc28j60/enc28j60.c
smlng: thats how its mostly done right now for many drivers, which doesn't make it (more) right but at least consistent 😉

kaspar030: Well, it is done for about a third of all drivers... I see this is a more general discussion, so ok.

drivers/include/enc28j60.h
kaspar030: true, nevermind.

drivers/enc28j60/enc28j60.c
```diff
@@ -85,30 +89,29 @@
 
 static void switch_bank(enc28j60_t *dev, int8_t bank)
 {
-    /* only switch bank if needed */
-    if ((bank < 0) || (dev->bank == bank)) {
+    assert(bank < 0x04);
+
+    if (bank < 0) {
```
kaspar030: I'd prefer this semantic change split out (into another PR).

drivers/include/enc28j60.h
smlng: mhm, it also increases ROM/RAM usage by some 30 Bytes, which I don't understand as I would've guessed that using a pointer instead of copying the struct should safe some bits and bytes

drivers/enc28j60/enc28j60.c
smlng: I made it separate commit

smlng:
ping @kaspar030, I put changes into distinct/independent commits, IMHO separate PRs is a bit overkill here.

PeterKietzmann:
> I'd prefer this semantic change split out (into another PR).

@kaspar030 the bank change was split out into a separate commit which I think is totally fine. Not to bother you too much with this PR and to get it in the upcoming release I consider dismissing your change request tomorrow.

PR #9896, opened 2018-09-06
jlink.sh: remove duplicated check

kYc0o:
### Contribution description

Currently, jlink.sh makes double check for the existance of an `HEXFILE`, one of this checks  is not necessary, as it already exists a function for checking it.

### Testing procedure

Flash a board using the jlink debugger/programmer.

### DIFF
--- a/dist/tools/jlink/jlink.sh
+++ b/dist/tools/jlink/jlink.sh
@@ -66,12 +66,6 @@ _JLINK_TERMFLAGS="-ts 19021"
 # a couple of tests for certain configuration options
 #
 test_config() {
-    if [ -z "${HEXFILE}" ]; then
-        echo "no hexfile"
-    else
-        echo "HEXFILE found"
-    fi
-
     if [ -z "${JLINK}" ]; then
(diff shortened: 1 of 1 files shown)


dist/tools/jlink/jlink.sh
```diff
@@ -99,6 +93,8 @@ test_hexfile() {
         echo "Error: Unable to locate HEXFILE"
         echo "       (${HEXFILE})"
         exit 1
+    else
```
basilfx: Do we need to print on success? If so, then `test_elffile` should also be adapted.

basilfx:
Tested this on SLSTK3402a that uses JLink. Works fine.

dist/tools/jlink/jlink.sh
kYc0o: True, I'd say we remove that part, is not really needed. I'll remove it and squash directly.

kYc0o: Done.

kYc0o:
Thanks @basilfx !

PR #9900, opened 2018-09-07
periph/timer: extend API with timer_diff call

smlng:
### Contribution description

This PR proposes an extension of the periph/timer API, namely to provide a `timer_diff` function that allows to calculate the difference of two time values (acquired by running `timer_read`) and also account for the timers configuration. Currently its not possible to generically calculate the difference without explicitly knowing the timers configuration, the problem arises when running a timer in 16-Bit mode on a 32-Bit platform.

If a platform only uses timers with same width as the platform, e.g. Kinetis, a common helper function is provided in `drivers/periph_common/timer.c`. Otherwise a platform may implement its own variant and also define MACRO `PERIPH_TIMER_PROVIDES_DIFF`. The latter is done for CPUs `stm32_common`, `cc2538`, `cc26x0`, and `samd21`.

Though this is actually an API extension I still set the label *API CHANGE* just in case.

### Testing procedure

This PR provides a test application see `tests/periph_timer_diff`.

### Issues/PRs references

Problem was discovered when reviewing #9283

### DIFF
--- a/cpu/cc2538/include/periph_cpu.h
+++ b/cpu/cc2538/include/periph_cpu.h
@@ -242,6 +242,11 @@ typedef struct {
 } timer_conf_t;
 
 /**
+ * @brief Timer implementation provides custom timer_diff function
+ */
+#define PERIPH_TIMER_PROVIDES_DIFF
+
+/**
  * @name   Override resolution options
  * @{
  */
--- a/cpu/cc2538/periph/timer.c
+++ b/cpu/cc2538/periph/timer.c
@@ -213,6 +213,18 @@ unsigned int timer_read(tim_t tim)
     }
 }
 
+unsigned int timer_diff(tim_t tim, unsigned int begin, unsigned int until)
+{
+    unsigned int diff = 0;
+    if (tim < TIMER_NUMOF) {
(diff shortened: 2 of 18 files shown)


kaspar030:
Wouldn't it make more sense to introduce a function that returns the width of a periph timer? Using that, one single 'timer_diff()' could be provided.

smlng:
mhm, implementation wise its seems like the same number of CPU specific`timer_width` functions instead of `timer_diff`. But I see you point in that it might be more useful to get the timer width for other use cases, too. will adapt - but there are some other bugs that popped along the line :wink see #9902

gebart:
I agree that there are use cases which would benefit from being able to handle timer rollover, but I am not convinced that this is the best way to do it. 
A timer width or timer max function would also be able to solve this, as well as some other situations, like if you would want to compute a target into the future, with rollover (timer_add).

smlng:
well then thanks for the input, I'd say we have rough consensus, so I'll provide a timer_width (or timer_mask) function in a separate PR and will update and simplify this one afterwards.

smlng:
superseded by #10493

PR #9901, opened 2018-09-07
net/gcoap: use sock_udp_ep_equal()

haukepetersen:
### Contribution description
Removing some duplicated code: `sock_udp_ep_equal()` was recently added as shared code to `sock_util`. So `gcoap` might as well use it.

### Testing procedure
Apply this patch an run the `exmaples/gcoap` example. You should still be able to send out requests as before...

### Issues/PRs references
none


kb2ma:
I agree completely with this PR. I noticed the sock util commit as well at the time.

kaspar030:
@kb2ma I'll let you have a proper look!

PR #9906, opened 2018-09-07
periph_common: add as dependency to periph drivers 

smlng:
### Contribution description

 Rational: the periph_common module is included by each and every CPU/MCU
    via one way or the other, hence it can be activated by default. With this
    PR its added to the list of default modules.

Actually, the periph_common was missing for CPUs `esp8266` and `fe310` which would result in I2C or SPI not working for instance, because they rely on periph auto init by periph_common. This is fixed with this PR implicitly -> but requires #9905.

### Issues/PRs references

~~based on #9905~~

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -877,6 +877,11 @@ USEMODULE += $(filter periph_%,$(FEATURES_USED))
 # select cpu_check_address pseudomodule if the corresponding feature is used
 USEMODULE += $(filter cpu_check_address, $(FEATURES_USED))
 
+# include periph_common if any periph_* driver is used
+ifneq (,$(filter periph_%, $(USEMODULE)))
+  USEMODULE += periph_common
+endif
+
 # recursively catch transitive dependencies
 USEMODULE := $(sort $(USEMODULE))
 USEPKG := $(sort $(USEPKG))
--- a/cpu/cc2538/Makefile.include
+++ b/cpu/cc2538/Makefile.include
@@ -1,6 +1,3 @@
 export CPU_ARCH := cortex-m3
 
-# include common SPI functions
-USEMODULE += periph_common
-
 include $(RIOTMAKE)/arch/cortexm.inc.mk
--- a/cpu/cc26x0/Makefile.include
+++ b/cpu/cc26x0/Makefile.include
@@ -1,5 +1,3 @@
 export CPU_ARCH := cortex-m3
 
-USEMODULE += periph_common
-
 include $(RIOTMAKE)/arch/cortexm.inc.mk
--- a/cpu/cc430/Makefile.include
+++ b/cpu/cc430/Makefile.include
@@ -1,3 +1,3 @@
 include $(RIOTCPU)/msp430_common/Makefile.include
 
-export USEMODULE += periph periph_common
+export USEMODULE += periph
--- a/cpu/efm32/Makefile.include
+++ b/cpu/efm32/Makefile.include
@@ -20,9 +20,6 @@ USEMODULE += cpu_$(EFM32_FAMILY)
 # vectors.o is provided by 'cpu_$(EFM32_FAMILY)' and not by 'cpu'
 VECTORS_O := $(BINDIR)/cpu_$(EFM32_FAMILY)/vectors.o
 
-# include common periph module
-USEMODULE += periph_common
-
 # include layered power management
 USEMODULE += pm_layered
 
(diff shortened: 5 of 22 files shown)


smlng:
rebased, bc #9905 was merged

smlng:
ola, any comments on this generalisation?

PeterKietzmann:
Mabye from @haukepetersen?

cladmi:
By default it only always adds the definition of `periph_init` as all the other modules are protected by `#ifdef` on things that should be defined by the periph config.

In general it makes sense as the dependencies have been implemented using headers macros instead of the build system.
I would add a comment on top of the `DEFAULT_MODULE += periph_common` about that.


> Actually, the periph_common was missing for CPUs esp8266

It is there https://github.com/RIOT-OS/RIOT/blob/e4ca897661460a2eea68d407afea8faa0fbf0254/cpu/esp8266/Makefile.include#L122 and it is using `periph_init`

> Actually, the periph_common was missing for CPUs (…) fe310

It is not using `periph_init` function so is currently not relying on it.
If `periph_init` is now included by default, it should be addressed before.

smlng:
@cladmi rebase and added comment as suggested

smlng:
ping!

smlng:
ping again

smlng:
@cladmi what was the problem here, i.e. are any changes required. Do you agree or disagree with making `periph_common` a default module?

kaspar030:
It might make more sense to make periph_* depend on periph_common.
Or even be explicit. As far as I can see, only hwrng, i2c, rtc, rtt, spi, timer, pm, cpuid, flashpage, eeprom would make use of this module.

smlng:
> It might make more sense to make periph_* depend on periph_common.
> Or even be explicit. As far as I can see, only hwrng, i2c, rtc, rtt, spi, timer, pm, cpuid, flashpage, eeprom would make use of this module.

mhm, sounds reasonable, too - my observation/point was that is doesn't make sense to have `USEMODULE += periph_common` explicitly added by all CPUs independently and rather move it to a central location. However, you're right instead of making it a default module we could add a rule as you stated above more `periph_%` to pull in `periph_common`.

smlng:
@kaspar030 add a fixup with your suggestion, need to reword the commit message when squashing

kaspar030:
Now there might be a problem when no periph_ is actually used, as then theres no periph_init(), but that is hard-called in all architectures. Unfortunately the call cannot easily be moved into e.g., kernel_init(), as some platforms don't call it last in their startup sequence...

Maybe that is not a problem. It is not with the current codebase.

smlng:
well that hard-coded call of `periph_init` (and other stuff) is a problem in itself I'd say and should be fixed/changed in the long run, too. But for now, RIOT always pulls in some `periph_` module (eg `timer`) and hence also `periph_common`. 

So I would go for this solution, also (thanks to you suggestion to make it a dependency) its much cleaner and more appropriate way to add `USEMODULE += periph_common` in `Makefile.dep` compared to adding it by each and every CPU.

kaspar030:
> So I would go for this solution

+1, one step at a time. :)

smlng:
can/shall I squash?

kaspar030:
yes please. @cladmi wanna take another look?

smlng:
done, ready to merge ...

cladmi:
Currently it makes that `periph_common` depends on `periph_common` but should not break anything :D 
This could be changed in another PR by renaming it to `periph-common` instead this way it could differentiate from `periph_` pseudomodules. It will be easier as there should be only one place using the name `periph_common` after this PR.

I agree with he given justification in the commit message and the current changes done.

However, I think that now there are other places that might need the change too:

```
git grep 'USEMODULE += periph_common'
Makefile.dep:  USEMODULE += periph_common
cpu/esp8266/Makefile.include:USEMODULE += periph_common
cpu/fe310/Makefile.include:USEMODULE += periph_common
makefiles/arch/atmega.inc.mk:USEMODULE += periph_common
```

smlng:
@cladmi comments addressed, squash?

smlng:
@cladmi how to proceed? Can I squash?

cladmi:
Please squash, it got lost in my mailbox.

smlng:
done!

cladmi:
Forgot in the Approve comment, I checked that the branch rebased on master only had this dependency definition to `periph_common`:

```
git grep 'USEMODULE += periph_common'
Makefile.dep:  USEMODULE += periph_common
```

PR #9914, opened 2018-09-10
isotp: advance protocol FSM when no confirmation is used

OYTIS:
The user can specify CAN_ISOTP_TX_DONT_WAIT to ignore transmit
confirmations, _isotp_conf_tx_tx() should be called in this case
as if confirmation was received immediately.

Signed-off-by: Anton Gerasimov <tossel@gmail.com>

UPD:
by the way, sending with confirmation doesn't seem to work in general case.

### DIFF
--- a/sys/can/isotp/isotp.c
+++ b/sys/can/isotp/isotp.c
@@ -646,6 +646,10 @@ static int _isotp_tx_send(struct isotp *isotp, struct can_frame *frame)
         return _isotp_dispatch_tx(isotp, isotp->tx.tx_handle);
     }
 
+    if (isotp->opt.flags & CAN_ISOTP_TX_DONT_WAIT) {
+      _isotp_tx_tx_conf(isotp);
+    }
+
     return 0;
 }
 


OYTIS:
@vincent-d 
Thanks for the comment, I've misunderstood the purpose of `CAN_ISOTP_TX_DONT_WAIT`. Yes, I'm experiencing a problem with TX confirmations (low-level ones, `CAN_MSG_TX_CONFIRMATION`), namely that the confirmation is a message passed from one thread of another and the time it takes to reach the destination is unpredictable (could be up to several seconds under debug).

My failure scenario was:
1. RIOT ISO/TP sends first frame.
2. CAN driver immediately confirms it.
3. Remote device receives the first frame and sends flow control frame.
4. Flow control frame is received on RIOT and dispatched to ISO/TP. As transimission has not yet been confirmed ISO/TP ignores it.
5. Finally transmit confirmation reaches ISO/TP (but it's too late).

I might be doing something wrong of course (like e.g. wrong thread priorities).

OYTIS:
@vincent-d Probably losing/reordering events is the real problem. I'll keep the PR open until I find the real solution. Thanks again!

OYTIS:
@vincent-d 
I somehow can't reproduce it any more. Sorry for the hassle. Closing.

Issue #9913, opened 2018-09-10
Build dependencies - processing order issues


cladmi:
There currently are concurrency issues in the order dependencies are defined and used, and cpu dependencies cannot be easily parsed.

I plan to address this by moving the processing of `Makefile.dep` before including `$(BOARD)/Makefile.include`

Dependencies resolution uses the following variables
* USEMODULE
* USEPKG
* FEATURES_
* BOARD
* CPU
* CPU_FAM (which depend on files in CPU)
* CPU_ARCH (which depend on files in CPU)
* CPU_MODEL

### Steps

- [ ] Fix CPU dependencies not being parsed easily as reported first in https://github.com/RIOT-OS/RIOT/pull/9892
    - [ ] ~Add a new file in board that include the common files too and currently only defines `CPU`, `Makefile.board`.~ Reuse `Makefile.features` to define `CPU`
    - [ ] Migration of CPU/CPU_MODEL https://github.com/RIOT-OS/RIOT/pull/11477
    - [ ] Move CPU dependencies to `$(CPU)/Makefile.dep`
    - [x] Fix the `stm32_common` issue described in https://github.com/RIOT-OS/RIOT/pull/9892
         - [x] Remove them being defined as dependencies ~https://github.com/RIOT-OS/RIOT/pull/10153~
- [ ] At that point, Makefile.features can be split to directly do `include $(RIOTCPU)/$(CPU)/Makefile.features` instead of hardwriting it.
- [ ] Fix `CPU_FAM` and `CPU_ARCH` defined in `Makefile.include` when they should be already available for `Makefile.dep`. Define them in ~`Makefile.cpu`~ `Makefile.features` files.
- [ ] Move all BOARD/CPU dependencies to `$(BOARD)/Makefile.dep` and `$(CPU)/Makefile.dep`
- [ ] Fix other USEMODULE defined in `Makefile.include` issues (can be done in parallel)
- [ ] Move `Makefile.dep` before doing `Makefile.include`.
- [ ] Reorganize files inclusion to be done only when needed

Prerequisites cleanup

- [x] mips-malta including common cpu features https://github.com/RIOT-OS/RIOT/pull/10064
- [x] boards: always include cpu features #10078
    - [x] cpu/esp8266 re-defining CPU https://github.com/RIOT-OS/RIOT/pull/10061
    - [x] hifive1 not processing CPU features https://github.com/RIOT-OS/RIOT/pull/10063
[...]

PR #9915, opened 2018-09-10
Makefile.include: always include board Makefile.features

cladmi:
### Contribution description

makefiles/info-global.inc.mk already includes this file unconditionally.
So it is misleading to say it is optional.

### Testing procedure

Murdock compiling everything means it is working

### Issues/PRs references

Found while working on https://github.com/RIOT-OS/RIOT/issues/9913


cladmi:
I tried to also make all include of `Makefile.features` unconditional and it is working. I can push this here too or do a separate PR as it is a different cleanup.

OlegHahm:
Seems reasonable for me. ACK.

PR #9916, opened 2018-09-10
dist/tools/jlink.sh: fix testrunner on stdio_rtt based devices

kYc0o:
### Contribution description

This PR enables to run `make test` on devices using jlink as a debugger/programmer, along with stdio_rtt.
Currently this is not possible due to two things:

  - testrunner executes `make term` + `make reset` while running the test scripts, triggering two executions of `JLinkExe`. `make term` executes `JLinkExe` in background to then connect through `pyterm` to a TCP port. `JLinkExe` sends a prompt expecting for commands, but nothing is send and it seems to block other executions of itself. This is problematic since `make reset` also needs `JLinkExe` but it's locked by the previous command, thus `make test` fails. Sending the output to /dev/null unlocks `JLinkExe` and other instances of it can be launched.

  - testrunner kills the processes needed for `make term` when it exits, sending the KILL signal through `os.killpg` which should kill the processes group. However, for stdio_rtt, the term process needs a connection to the target through `JLinkExe`, which is executed in background but using `setsid`. This leads to `JLinkExe` not being part of the group and thus is not killed. `make test` would succeed but every time it's executed an instance of `JLinkExe` is left running. By removing `setsid` from the call the problem is solved. It seems to me that is a copy/paste leftover, otherwise I'd like to know why `setsid` is needed in this case.

### Testing procedure

Do `make test` for any supported test on a stdio_rtt based device.

### Issues/PRs references

This helps to solve `make test` in Hamilton boards, from #9013.

### DIFF
--- a/dist/tools/jlink/jlink.sh
+++ b/dist/tools/jlink/jlink.sh
@@ -245,7 +245,7 @@ do_term() {
             -speed '${JLINK_SPEED}' \
             -if '${JLINK_IF}' \
             -jtagconf -1,-1 \
-            -commandfile '${RIOTTOOLS}/jlink/term.seg' & \
+            -commandfile '${RIOTTOOLS}/jlink/term.seg' >/dev/null & \
             echo  \$! > $JLINK_PIDFILE" &
 
     sh -c "${JLINK_TERMPROG} ${JLINK_TERMFLAGS}"
--- a/dist/tools/jlink/jlink.sh
+++ b/dist/tools/jlink/jlink.sh
@@ -240,7 +240,7 @@ do_term() {
     # don't trapon Ctrl+C, because JLink keeps running
     trap '' INT
     # start Jlink as RTT server
-    setsid sh -c "${JLINK} ${JLINK_SERIAL} \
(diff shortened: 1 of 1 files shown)


PeterKietzmann:
@leandrolanzieri interested?

leandrolanzieri:
@kYc0o is it correct to apply this to #9851 in order to test?

kYc0o:
No, actually the title is a little misleading, this is only to perform `make test` on boards using stdio_rtt. I'll update accordingly.

basilfx:
@kYc0o As discussed last week, RTT works better with this patch, so I think we can merge it now?

kYc0o:
> @kYc0o As discussed last week, RTT works better with this patch, so I think we can merge it now?

I'd love it, but I need an ACK ;-)

Hyungsin:
@haukepetersen, please review this PR. Hamilton cannot be merged due to this PR.

kYc0o:
@basilfx already tested it, so maybe he can give his ACK.

kYc0o:
Thanks! 😄

Issue #9888, opened 2018-09-05
gcoap: cannot get resources with paths containing multiple digit numerics


pokgak:
#### Description
gcoap states that resources must be ordered alphabetically and checks for it with `strcmp` but that actually compares it ASCII-ordered (`/0, /1, /10, /11, /2, /3, ...`) and not alphabetically (`/0, /1, /2, ..., /10, /11`)

Note: this doesn't affect paths such as `/a/10` and `/b/2` because it stops comparing the paths at the alphabets and doesn't reach the numbers

#### Steps to reproduce the issue

1. declare coap resources which contains multidigit numeric (`/0, /1, ..., /10, /11`)
2. `coap get [address] [port] /10`

#### Expected results
Gets value for /10

#### Actual results
`gcoap: response Error, code 4.04, empty payload` because gcoap stops looking for /10 after comparing /10 and /2 because it is not 'alphabetically ordered' and returns GCOAP_RESOURCE_NO_PATH instead.

pokgak:
As possible solutions, 
1. change the alphabetically ordered statement in gcoap.h to ASCII-ordered OR
2. use something like the [alphanum algorithm](http://www.davekoelle.com/alphanum.html) and not `strcmp` to look for the path

kb2ma:
Thanks for the report, @pokgak! I agree that the documentation is wrong on the requirement for sorting the entries. The list is indeed ASCII-ordered rather than alphabetic . So, in addition to your examples, `/B` sorts before `/a`. However, the issue you describe is that numbers are not sorted numerically.

The reason for the sort requirement is just to make the lookup simple and fast. For this purpose in an embedded system, I would not go farther than that. So, I definitely support an update to the documentation to say that entries must be in ASCII order, and would include an example for clarity.

kaspar030:
+1 for fixing the documentation.

PR #7580, opened 2017-09-07
gcoap: DEVELHELP checks resources are sorted

cladmi:
gcoap requires resources to be sorted alphabetically, but currently nothing verifies it.
I added a runtime check that warns if not the case, only enable when DEVELHELP is.

Also, in tests-gcoap, the test resources where not sorted.

To make it visible, without adding many tests, I needed to enable DEVELHELP in unittests makefile.

I also fixed an unused variable enclosing ifdef, as its NDEBUG that controls assert.
NDEBUG should be enabled when DEVELHELP is, but if you enable it with -DDEVELHELP=1, NDEBUG is not and it did not compile anymore.

PR #9919, opened 2018-09-11
nanocoap/example: fix resource order

kb2ma:
### Contribution description

Fix order of resources in nanocoap server example. Must be in ASCII order. Also updated inline comment based on #9888.

### Testing procedure

Build/run nanocoap_server example.

Send a POST for /riot/value, like the following in libcoap CLI:
```
$ echo 1 |./coap-client -m post -T 5a -U coap://[fe80::200:bbff:febb:2%tap0]/riot/value -f -
```

We do not expect any response from coap-client, but instead we see `4.04` because the nanocoap server was unable to find the resource. Also, verify the fix with a 'get' request to be sure the value was set.

```
$ ./coap-client -m get -T 5a -U coap://[fe80::200:bbff:febb:2%tap0]/riot/value
1
```
### Issues/PRs references

Related to #9888 for gcoap, but not strictly a fix for it. I just checked nanocoap to be sure no similar issues.

### DIFF
--- a/examples/nanocoap_server/coap_handler.c
+++ b/examples/nanocoap_server/coap_handler.c
@@ -104,12 +104,12 @@ ssize_t _sha256_handler(coap_pkt_t* pkt, uint8_t *buf, size_t len, void *context
     return pkt_pos - (uint8_t*)pkt->hdr;
 }
 
-/* must be sorted by path (alphabetically) */
+/* must be sorted by path (ASCII order) */
 const coap_resource_t coap_resources[] = {
     COAP_WELL_KNOWN_CORE_DEFAULT_HANDLER,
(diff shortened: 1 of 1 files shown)


cladmi:
That reminds me of a PR I did a long time ago https://github.com/RIOT-OS/RIOT/pull/7580 to add warnings if they are not sorted (in gcoap as nanocoap was not merged yet). Would it make sense to re-open one ?
It is the second error like this I see in our examples, and one more with the mentioned issue.

cladmi:
I reviewed and agree with the changes but cannot test right now.

kb2ma:
Thanks for the review and merges @bergzand!

PR #9933, opened 2018-09-14
tests: hierarchical structuring of the test directory

OlegHahm:
### Contribution description

The list of directories inside the tests folder has become really, really long and difficult to oversee. Hence, I propose to introduce some sub-directories to logically order the tests into groups represented by directories.


cladmi:
I fixed the other `flake8` complains.

cladmi:
A first note, is that the new tests directories should be added to https://github.com/RIOT-OS/RIOT/blob/bbc16cced2c994ec57e0e7808e788932407c8885/makefiles/app_dirs.inc.mk to be detected by murdock and `compile_test`.
To make this more flexible, I am thinking about adding sub-files to declare this applications patterns. Like tests could have a file saying include the files in this directories, and each directory says, there are applications in these ones.

I like the idea of re-organizing the tests but with the current state I have a bit mixed feeling as there are still a lot of tests directories and directories with tests inside so an in between situation. So maybe go even further with the directory re-organization.

For packages I am also thinking of putting the tests in each package directory to move toward self contained packages. But it would require the modular definition of `APPLICATION_DIRS` I think.

MrKevinWeiss:
I like nesting.  I think if we can get consensus from enough people then schedule a time to get this done because I can imagine hitting a moving target is difficult.

PR #9934, opened 2018-09-14
boards: Add two KW41Z-based boards

pyropeter:
### Contribution description

This PR refactors the FRDM-KW41Z boards support to allow for code sharing with other KW41Z-based boards and then adds two of them:

* kw41z-mini: https://openlabs.co/store/kw41z-mini (Basically a KW41Z breakout board with quarzes, reset button, one LED, and a 1.27mm 2x5 ARM Cortex-M debug connector)
* Beduino Pro Mini: Our kw41z-mini clone with integrated DAPLink programmer and a second LED (To be published)

The first board is commercially available (for now, at least) and the second one is completely unpublished at the moment. We finished producing the first 15 prototype boards by hand yesterday and we will now look into commercial manufacture (and acquiring funds). We are aiming to publish our KICAD source files under a permissive license. I would not be insulted if the limited availability of the boards was reason to reject (parts of) this pull request.

This is the first time I introduced a `boards/common` directory, so I probably made some errors.

### TODO

* [ ] Add documentation for both boards

### DIFF
--- a/boards/beduino-pro-mini/include/board.h
+++ b/boards/beduino-pro-mini/include/board.h
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2018 Tristan Bruns
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser General
+ * Public License v2.1. See the file LICENSE in the top level directory for more
+ * details.
+ */
+
+/**
+ * @ingroup     boards_beduino-pro-mini
+ * @{
+ *
+ * @file
+ * @brief       Board specific definitions for the Beduino Pro Mini
+ *
+ * @author      Joakim Nohlgård <joakim.nohlgard@eistec.se>
+ * @author      Maximilian Luenert <malu@uni-bremen.de>
+ * @author      Tristan Bruns <tbruns@uni-bremen.de>
+ */
+
+#ifndef BOARD_H
+#define BOARD_H
+
+#include "cpu.h"
+#include "periph_conf.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/**
+ * @name    LED pin definitions and handlers
+ * @{
+ */
+#define LED0_PIN            GPIO_PIN(PORT_B,  0)
+#define LED0_MASK           (1 << 0)
+#define LED0_ON             (GPIOB->PCOR = LED0_MASK)
+#define LED0_OFF            (GPIOB->PSOR = LED0_MASK)
+#define LED0_TOGGLE         (GPIOB->PTOR = LED0_MASK)
+#define LED1_PIN            GPIO_PIN(PORT_C,  5)
+#define LED1_MASK           (1 << 5)
+#define LED1_ON             (GPIOC->PCOR = LED1_MASK)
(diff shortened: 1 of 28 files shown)


boards/beduino-pro-mini/include/board.h
```diff
@@ -0,0 +1,80 @@
...
+#define XTIMER_DEV                  (TIMER_LPTMR_DEV(0))
+#define XTIMER_CHAN                 (0)
+/* LPTMR is 16 bits wide and runs at 32768 Hz (clocked by the RTC) */
+#define XTIMER_WIDTH                (16)
+#define XTIMER_BACKOFF              (5)
+#define XTIMER_ISR_BACKOFF          (5)
+#define XTIMER_OVERHEAD             (4)
+#define XTIMER_HZ                   (32768ul)
+#endif
+/** @} */
```
pyropeter: @gebart If I understand the KW41Z and xtimer correctly, the `XTIMER_*` defines should be the same for all KW41Z-based boards, right? If true, this should probably also be moved to `common/kw41z`, right? Is there some convention on how to do this?

boards/kw41z-mini/Makefile.dep
```diff
@@ -0,0 +1,6 @@
+ifneq (,$(filter saul_default,$(USEMODULE)))
+  USEMODULE += saul_gpio
+  USEMODULE += saul_adc
+endif
+
+include $(RIOTCPU)/kinetis/Makefile.dep
```
pyropeter: I forgot to refactor `Makefile.dep`

gebart:
It does not make sense to share the periph_conf.h and adc_params.h between different boards unless they are completely pin compatible. But if they are so similar it doesn't make any sense to have a separate configuration at all. Are they identical to the frdm-kw41z in the schematic? If they are, why do you need a separate board config to begin with?

x3ro:
I just tried out the `kw41z-mini` and the basics seem to be working fine, that is hello-world, default, lua_basics. Didn't have time to test any more right now, except `gnrc_networking` which currently panics in `gnrc_rpl_auto_init.c:32`:

~~~
#if (GNRC_NETIF_NUMOF == 1)
    gnrc_netif_t *netif = gnrc_netif_iter(NULL);
    assert(netif != NULL);  // <------
~~~

Did networking work well for you?

x3ro:
Oh and I also had to add this to `Makefile.include` be able to flash with a JLink:

~~~
ifeq (jlink,$(PROGRAMMER))
  export JLINK_DEVICE := MKW41Z256XXX4
  include $(RIOTMAKE)/tools/jlink.inc.mk
endif
~~~

To be honest I'm not sure where would best put this, though. Mostly putting it here so I don't forget :D

gebart:
@x3ro you need #7107 for a radio driver for the kw41z.

aabadie:
Looks like this PR was missed when usb-kw41z and phynode-kw41z were recently added. @pyropeter would mind split this PR: one introducing a common place for kw41z based boards (the most important!), one adding support for the kw41z-mini ?
I'm with @smlng to not provide a board support for a board that is not (yet) available on the market.

PR #9935, opened 2018-09-14
 Makefile.include: replace $(USEPKG:%=$(BINDIR)/%.a) target 

cladmi:
### Contribution description

Replace the per package archive file target by phony targets.

This allows at the same time to correctly describe PSEUDOMODULES packages as they should not be file targets and prepares for build system changes where some packages could produce objects
instead of archives for the 'ld -r' PR.

Having an explicit file target here was not adding anything.

### Testing procedure

If all applications build without problem even in parallel mode and the test application commit also build, everything should be ok.

### Issues/PRs references

* Mentioned the problem of packages generating objects in a [comment](https://github.com/RIOT-OS/RIOT/pull/8711/files#r192727184)  in ld -r PR https://github.com/RIOT-OS/RIOT/pull/8711
* Noticed the PSEUDOMODULES issue when working on wolfssl https://github.com/RIOT-OS/RIOT/pull/9894

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -475,9 +475,9 @@ endif
 pkg-prepare:
 	-@for i in $(USEPKG) ; do "$(MAKE)" -C $(RIOTPKG)/$$i prepare ; done
 
-$(USEPKG:%=$(BINDIR)/%.a): $(BUILDDEPS) FORCE
+$(USEPKG:%=$(BINDIR)/%.a): $(BINDIR)/%.a: $(BUILDDEPS) FORCE
 	@mkdir -p $(BINDIR)
-	$(QQ)"$(MAKE)" -C $(RIOTPKG)/$(patsubst $(BINDIR)/%.a,%,$@)
+	$(QQ)"$(MAKE)" -C $(RIOTPKG)/$*
 
 clean:
 	-@for i in $(USEPKG) ; do "$(MAKE)" -C $(RIOTPKG)/$$i clean ; done
--- a/Makefile.include
+++ b/Makefile.include
@@ -476,7 +476,6 @@ pkg-prepare:
 	-@for i in $(USEPKG) ; do "$(MAKE)" -C $(RIOTPKG)/$$i prepare ; done
 
 $(USEPKG:%=$(BINDIR)/%.a): $(BINDIR)/%.a: $(BUILDDEPS) FORCE
-	@mkdir -p $(BINDIR)
 	$(QQ)"$(MAKE)" -C $(RIOTPKG)/$*
 
 clean:
--- a/Makefile.include
+++ b/Makefile.include
@@ -430,13 +430,13 @@ endif # RIOTNOLINK
 $(ELFFILE): $(BASELIBS)
 	$(Q)$(_LINK) -o $@
 
-$(BINDIR)/$(APPLICATION_MODULE).a: $(USEPKG:%=$(BINDIR)/%.a) $(BUILDDEPS)
+$(BINDIR)/$(APPLICATION_MODULE).a: pkg-build $(BUILDDEPS)
 	$(Q)DIRS="$(DIRS)" "$(MAKE)" -C $(APPDIR) -f $(RIOTMAKE)/application.inc.mk
 $(BINDIR)/$(APPLICATION_MODULE).a: FORCE
 
 # Other modules are built by application.inc.mk and packages building
-_SUBMAKE_LIBS = $(filter-out $(BINDIR)/$(APPLICATION_MODULE).a $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS), $(BASELIBS))
-$(_SUBMAKE_LIBS): $(BINDIR)/$(APPLICATION_MODULE).a $(USEPKG:%=$(BINDIR)/%.a)
+_SUBMAKE_LIBS = $(filter-out $(BINDIR)/$(APPLICATION_MODULE).a $(APPDEPS), $(BASELIBS))
+$(_SUBMAKE_LIBS): $(BINDIR)/$(APPLICATION_MODULE).a pkg-build
 
 # 'print-size' triggers a rebuild. Use 'info-buildsize' if you do not need to rebuild.
 print-size: $(ELFFILE)
@@ -471,11 +471,12 @@ ifneq (, $(filter clean, $(MAKECMDGOALS)))
     all $(BASELIBS) $(USEPKG:%=$(RIOTPKG)/%/Makefile.include) $(BUILDDEPS): clean
 endif
 
-.PHONY: pkg-prepare
+.PHONY: pkg-prepare pkg-build pkg-build-%
 pkg-prepare:
 	-@for i in $(USEPKG) ; do "$(MAKE)" -C $(RIOTPKG)/$$i prepare ; done
 
-$(USEPKG:%=$(BINDIR)/%.a): $(BINDIR)/%.a: $(BUILDDEPS) FORCE
+pkg-build: $(USEPKG:%=pkg-build-%)
+pkg-build-%: $(BUILDDEPS)
 	$(QQ)"$(MAKE)" -C $(RIOTPKG)/$*
 
 clean:


cladmi:
The test application also works in master, as nothing uses the file, but the file does not exist after building in `master` despite being a file target:

```
$ make -C examples/hello-world/  RIOT_CI_BUILD=1
make: Entering directory '/home/cladmi/git/work/RIOT/examples/hello-world'
Building application "hello-world" for "native" with MCU "native".

make[1]: Nothing to be done for 'prepare'.
make[1]: Nothing to be done for 'all'.
  22949     568   47716   71233   11641 /home/cladmi/git/work/RIOT/examples/hello-world/bin/native/hello-world.elf
make: Leaving directory '/home/cladmi/git/work/RIOT/examples/hello-world'
```

```
$ ls examples/hello-world/bin/native/fake_pkg.a
ls: cannot access 'examples/hello-world/bin/native/fake_pkg.a': No such file or directory
```

Makefile.include
```diff
@@ -430,13 +430,13 @@ endif # RIOTNOLINK
...
-$(BINDIR)/$(APPLICATION_MODULE).a: $(USEPKG:%=$(BINDIR)/%.a) $(BUILDDEPS)
+$(BINDIR)/$(APPLICATION_MODULE).a: $(BUILDDEPS) | pkg-build
 	$(Q)DIRS="$(DIRS)" "$(MAKE)" -C $(APPDIR) -f $(RIOTMAKE)/application.inc.mk
 $(BINDIR)/$(APPLICATION_MODULE).a: FORCE
 
 # Other modules are built by application.inc.mk and packages building
-_SUBMAKE_LIBS = $(filter-out $(BINDIR)/$(APPLICATION_MODULE).a $(USEPKG:%=$(BINDIR)/%.a) $(APPDEPS), $(BASELIBS))
-$(_SUBMAKE_LIBS): $(BINDIR)/$(APPLICATION_MODULE).a $(USEPKG:%=$(BINDIR)/%.a)
+_SUBMAKE_LIBS = $(filter-out $(BINDIR)/$(APPLICATION_MODULE).a $(APPDEPS), $(BASELIBS))
+$(_SUBMAKE_LIBS): $(BINDIR)/$(APPLICATION_MODULE).a | pkg-build
```
gebart: @cladmi Does this correctly detect when a file is modified in the pkg?
This is an order-only prerequisite which is also a .PHONY, which is unfamiliar territory for me. We need to be certain that this does not cause make to assume that the target is up to date when the pkg files have been modified.

cladmi: @gebart For me `order-only` `.PHONY` prerequisite is something you execute everytime but you still check the file timestamp.
I used an 'order-only' because building it should not invalidate the other files timestamp, it is not its role here.

My train of thoughts here is:

* ELFFILE is rebuilt if files in `$(BASELIBS)` changed (but also depend on FORCE, so it is rebuilt everytime to ensure it is rebuilt if we change one of the variables)
* These lines say that "Files in `$(BASELIBS)` are generated by executing the `$(BINDIR)/$(APPLICATION_MODULE).a` and `pkg-build` targets.
* The target `$(BINDIR)/$(APPLICATION_MODULE).a`, that in fact generate most of the targets in `$(BASELIBS)`, currently depends on FORCE to rebuild it everytime
* The target `pkg-build` is `.PHONY` so rebuilt everytime

So now, they are all always executed.

I just put it `order-only` as for me, it is the same behavior as before, execute it every time (there was `$(USEPKG:%=$(BINDIR)/%.a): $(BUILDDEPS) FORCE` before) but still consider the generated files timestamps.

Currently it also does not really change anything as `$(BUILDDEPS)` contains `.PHONY` targets so forces rebuilding anyway. I may be able to change it to an order-only target, just a bit worried of the `make clean all -j` case.

gebart: What I am confused about is how make treats the indirect dependencies via the order only prereq. Does make treat the .a file as needing a rebuild, or do we introduce a race here?
I know that the elf is always relinked in the current implementation of the build system, but that is not really what we want to do when nothing has changed, so it is necessary that the dependencies are correct or we are likely to introduce hidden problems that will bite us in the future.

cladmi:
@gebart I know, that is in my TODO pipeline to remove the FORCE and I started the process some time ago. It just takes many steps. Before they were all `.PHONY` targets and not even separated.

I explain below why I go with this only.

It's not perfect and would gladly adapt or take yours if you have a better solution.

----

The problems with the current rule, is that it is wrong that each `$(BINDIR)/pkgname.a` is built by going in the package directory:
* One counter example is `nordic_softdevice_ble` which is built by the `$(APPLICATION_MODULE)/.a` target. And because of this, make would not reconsider the file timestamp even if updated as it is not by the target that said it changes it.
* it also prevents having packages that are pseudomodules which may be required for `wolfssl` if you only want to use the crypto libraries and not the ssl stack.

Keeping this wrong rule is for me worst than having it defined as a .PHONY target.
One is false, the other "only" forces rebuilding.

I replaced the `order-only` target by a normal `.PHONY` target because you advised to do it and that indeed in the current state it was not adding anything except confusion.

For me this current change is required for the `ld -r` PR because even if separating between REALPKGS LIBPKGS for this rule it would still be wrong that all packages archives and modules are all build in there, so I would not go in this direction of making it more complex but still wrong.

-----

To give some history context, in march there were no file target to build the libraries, the elffile, the binfile, all were just a "bash script" block https://github.com/RIOT-OS/RIOT/commit/f64de068b37e0b2c7d4ce04d82de8e72e869f143#diff-ebbb3165b3276a049759e5cb4e04d6a4
So make could never consider that a file may be up to date and should not be rebuilt.

cladmi:
My real way to go later that I thing about, is to define a variable for each module `MODULE_SRC_DIR_module_name` that can then be used directly in `Makefile.include` to go in the correct directory to build instead of doing the recursive `application_dir` -> `sys` -> `net/gnrc` -> `netif` make hierarchy. But I did not push it yet as there were higher priority blocking issues.

Even add some `MODULE_MAKEFILE_PATH_module_name` to handle packages where the makefile is stored somewhere else than in the source directory.

And to remove the `FORCE`, do somehow the same thing as the `CFLAGS` file and lazy save all the variables that should trigger a rebuild https://github.com/RIOT-OS/RIOT/pull/9634.

gebart:
please squash

cladmi:
Thanks for the thorough review, I really appreciate that you challenged it.
It showed the limits and that I did not explained correctly all  the justifications.

I will try to summarize in the commit message.

cladmi:
I tried to summarize the explanation in the last commit, please tell if you want me to adapt it.

gebart:
Looks fine, please squash

cladmi:
For me they are split in independent commits and do some cleanup before doing the final "big" change, so would rather keep them, they all correctly compile `tests/lwip`.

However I can remove the "NEEDS SQUASHING" label that I set because I had a "TEST REMOVE ME" commit.

cladmi:
I will just push them on the `ld -r` PR and see if it works. Please wait for my input before merging.

cladmi:
I can also change `pkg-build` to be on the left of `$(BUILDDEPS)` now that it is not anymore an order-only dependency.

cladmi:
I amended the last commit with the following change to be more close to what was before.

``` diff
index 9f0352f86..dc336ef57 100644
--- a/Makefile.include
+++ b/Makefile.include
@@ -430,7 +430,7 @@ endif # RIOTNOLINK
 $(ELFFILE): $(BASELIBS)
        $(Q)$(_LINK) -o $@

-$(BINDIR)/$(APPLICATION_MODULE).a: $(BUILDDEPS) pkg-build
+$(BINDIR)/$(APPLICATION_MODULE).a: pkg-build $(BUILDDEPS)
        $(Q)DIRS="$(DIRS)" "$(MAKE)" -C $(APPDIR) -f $(RIOTMAKE)/application.inc.mk
 $(BINDIR)/$(APPLICATION_MODULE).a: FORCE
```

gebart:
:+1:

cladmi:
Thanks for reviewing.

Issue #9920, opened 2018-09-11
drivers/tsl2561: The i2c bus is not released


jcarrano:
#### Description

The driver for the TSL2561 acquires the bus and does not release it. This means that it is not possible to use other devices connected to the same bus.

See https://github.com/RIOT-OS/RIOT/blob/6265da2894ed5cb7c414a44cd2856be918c3a06b/drivers/tsl2561/tsl2561.c#L57

#### Expected results

- Every call to `i2c_acquire()` should be matched by a call to `i2c_release()`
- The bus should not be be held if it is not being used.

#### Actual results

- There is only one call to `i2c_acquire()` when the driver is initialized.
- There are no calls to  `i2c_release()`.

dylad:
Good catch ! This bug must be there for a long time as it has not been introduced by the I2C rework.
Anyway, @jcarrano I can provide a PR for fixing this if you have the hardware to test it.

jcarrano:
@dylad I don't have this hardware. I found the bug because I was implementing a "similar" sensor (TSL4531, #9911 ) which turned out not to be similar at all.

dylad:
@jcarrano Alright, I'll provide a fix soon then and we'll see if someone can test it on real hardware.

dylad:
@jcarrano I'm closing this, since #9937 got merged.

PR #9937, opened 2018-09-14
drivers/tsl2561: add missing acquire∕release functions

dylad:
### Contribution description
This PR adds missing i2c_acquire/i2c_release function to the tsl2561 driver.
The init function acquired the I2C dev but it was never released, so it prevent the system from reading other sensors on the same bus.

### Issues/PRs references
Thanks @jcarrano for spotting this in #9920


dylad:
Thanks @aabadie !

PR #9948, opened 2018-09-17
ili9341: Initial import of ili9341 LCD driver

bergzand:
### Contribution description

Simple driver and test application of the ili9341

### Testing procedure

Test application included

### Review notes

I've wrote this driver quite a while ago, it might be missing some of the newer coding practices.

### Issues/PRs references

None


aabadie:
Yeah! Thanks @bergzand :)
I'll test this one asap.

bergzand:
@aabadie Which board had this display on it? And what is the most useful default pinout for it?

aabadie:
> Which board had this display on it?

stm32-f429i-disc1

> And what is the most useful default pinout for it?

Look at [this branch](https://github.com/aabadie/RIOT/tree/pr/ili9341_aa) and in particular, [this commit](f93c3c0772ddfcf04bf2c605d01f4a807f8c0ed2), [the second commit](944bf14a5151784c55cedca274fd5adaf61ec581) also provides some cleanup

drivers/ili9341/Makefile
```diff
@@ -0,0 +1,2 @@
+MODULE = ili9341
```
aabadie: This line is not needed

drivers/include/ili9341.h
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+  * @brief   Device descriptor for a ili9341
+  */
+typedef struct {
+    ili9341_params_t params;          /**< I2C device that sensor is connected to */
+} ili9341_t;
+
+
+int ili9341_init(ili9341_t* dev, ili9341_params_t* prms);
```
aabadie: Missing doxygen documentation

bergzand: Hey, I checked for that, will fix :smile:

drivers/ili9341/Makefile
bergzand: Will remove

bergzand:
>  And finally, one question: the driver doesn't allow to display images. Do you have an idea how this could be done ?

The driver works by selecting an area on the display and update all pixels in that area. This closely match the lower level hardware. So yes, an image is possible, but I'd have to generate something from a bitmap. 

Say we have an image of the RIOT logo of X by Y pixels, we can use the `ili9341_map` function with `x2-x1==X` and `y2-y1==Y` to stream the pixels to build the logo to the device. `*colors` just needs to be an array of pixels, `X*Y` long, to build the logo.

If I have some time I can try to build it, otherwise feel free to open an PR on my branch :)

aabadie:
By the way, I tested this PR (with my 2 extra commits applied) and it works like a charm.

aabadie:
@x3ro, can you try [this branch](https://github.com/aabadie/RIOT/tree/pr/ili9341_aa) ? I tested it on the disc1 version. It uses the right way of passing driver parameters and I'd like @bergzand to include my changes in his PR (I don't know why I can't open a PR on top of this one)

bergzand:
@aabadie I have some time this weekend to work on the driver, just have to build a small setup to test the device.

x3ro:
Hey @aabadie, yeah your branch works fine. I'm just wondering whether the pin information should really go in the driver, given that it's quite board specific. Also, why are you using magic numbers for the ports instead of e.g. `PORT_A`?

aabadie:
Any news here @bergzand ?

> I'm just wondering whether the pin information should really go in the driver, given that it's quite board specific

I would say that this is not required. Maybe a comment in the driver would be enough. It just that this screen is provided by this board, which is already available in RIOT.

> why are you using magic numbers for the ports instead of e.g. PORT_A?

If we use `PORT_A` the driver won't build with some architectures like sam0 which uses PA. This could be fixed by the way.

x3ro:
@aabadie That would also be fixed by removing the pin info from the driver and putting it in the board, right? Or in other words, not define `ILI9341_PARAM_*` in the driver at all.

aabadie:
> Or in other words, not define ILI9341_PARAM_* in the driver at all.

What do you mean exactly ? using GPIO_UNDEF ?

bergzand:
> Any news here @bergzand ?

Let's start with a rebase here to have all the new board names.

bergzand:
>     * For the pin defines, you use a different wording than the ST SDK uses in the case of `DC` (yours) vs `WRX` (SDK and source I found). I have no idea which one is "the better" name, just wanted to point it out. `DC` confused me a bit because I immediately expanded it to "direct current" :D

The name of the pin is indeed `WRX`, but it only has the `WRX` function when using the parallel interface. When using the 4-line system (AKA SPI), it is called `D/CX`.

> Apparently the reset pin is not connected on the `stm32f429-disco` according to [this](https://stm32f4-discovery.net/2014/06/library-18-ili9341-ltdc-stm32f429-discovery/). I set it to an arbitrary pin and it seems to work. What is this used for? Do I need it?

I've made the reset pin optional, `GPIO_UNDEF` is now a valid value for the pin and will result in the hardware reset not being used. For my setup this doesn't work as the device requires a hardware reset before initialization, but if the reset of the board is attached to the reset of the display it might be good enough.

bergzand:
@aabadie I think I've covered all your issues.

I've added an little endian mode in case the data is in little endian format. this slows down the driver a bit, so big endian is preferred, but it is useful to have when the source data/image is in little endian format.

bergzand:
There is one remaining issue and that is that I can't seem to read data from the display. While not necessarily an issue when using the device, it would be nice to have.

@aabadie Could you please test if reading data from the display works for you? I'm not sure if it is because of my hardware or that I'm doing something wrong in the driver.

I'd rather not block this PR on this bug since it is not really critical for using the display.

drivers/ili9341/ili9341.c
```diff
@@ -0,0 +1,283 @@
...
+                dev->params.spi_clk);
+    ili9341_cmd_start(dev, cmd, true);
+    /* Dummy transfer */
+    spi_transfer_byte(dev->params.spi, dev->params.cs_pin, true, 0x00);
+    spi_transfer_bytes(dev->params.spi, dev->params.cs_pin, false, NULL,
+                       data, len);
+    spi_release(dev->params.spi);
+}
+
+static void ili9341_set_area(ili9341_t *dev, uint16_t x1, uint16_t x2,
```
silkeh: Maybe it is an idea to add this to the public API. Together with a write function that allows one to set `n` bytes in memory this would allow the writing of an area in multiple chunked writes. I have no use case for this, so feel free to ignore.

drivers/include/ili9341.h
```diff
@@ -0,0 +1,174 @@
...
+void ili9341_map(ili9341_t *dev, uint16_t x1, uint16_t x2, uint16_t y1,
+                 uint16_t y2, const uint16_t *color);
+
+/**
+ * @brief   Raw write command
+ *
+ * @param[in]   dev     device descriptor
+ * @param[in]   cmd     command
+ * @param[in]   data    data from the device
+ * @param[in]   len     length of the returned data
```
silkeh: Seems to me that `data` is data *to* the device and `len` is the length of that data.

drivers/include/ili9341.h
```diff
@@ -0,0 +1,174 @@
...
+ * @brief   Fill a rectangular area with an array of colors
+ *
+ * @param[in]   dev     device descriptor
+ * @param[in]   x1      x coordinate of the first corner
+ * @param[in]   x2      x coordinate of the opposite corner
+ * @param[in]   y1      y coordinate of the first corner
+ * @param[in]   y2      y coordinate of the opposite corner
+ * @param[in]   color   array of colors to fill the area with
+ */
+void ili9341_map(ili9341_t *dev, uint16_t x1, uint16_t x2, uint16_t y1,
```
silkeh: I don't think the function name is clear. Maybe `set_area` is clearer?

drivers/include/ili9341.h
```diff
@@ -0,0 +1,174 @@
...
+ * @param[in]   x1      x coordinate of the first corner
+ * @param[in]   x2      x coordinate of the opposite corner
+ * @param[in]   y1      y coordinate of the first corner
+ * @param[in]   y2      y coordinate of the opposite corner
+ * @param[in]   color   single color to fill the area with
+ */
+void ili9341_fill(ili9341_t *dev, uint16_t x1, uint16_t x2, uint16_t y1, uint16_t y2, uint16_t color);
+
+/**
+ * @brief   Fill a rectangular area with an array of colors
```
silkeh: I think 'pixels' would be clearer than 'colors'. A note to remind users that the input data should exactly match the size of the area might also be useful.

drivers/include/ili9341.h
silkeh: Additional observation: `x2/y2` seem to be the the x/y-coordinate of the pixel in the opposite corner, not the opposite corner itself.

drivers/include/ili9341.h
bergzand: Ack and ack

drivers/ili9341/ili9341.c
bergzand: We can always extend the API in a follow up PR :)

drivers/include/ili9341.h
bergzand: Sssssht

drivers/include/ili9341.h
bergzand: > I don't think the function name is clear. Maybe `fill_pixels` is clearer?

Ack

> Additional observation: `x2/y2` seem to be the the x/y-coordinate of the pixel in the opposite corner, not the opposite corner itself (same for all other x2/y2's).

So a documentation fix?

silkeh: If explained clearly a documentation fix is fine to me, but a code change would enable users to do things like: `fill_pixels(x, y, x + picture_width, y + picture_height, picture)`.

drivers/ili9341/ili9341.c
```diff
@@ -0,0 +1,283 @@
...
+                      sizeof(params));
+    params[0] = byteorder_htons(y1);
+    params[1] = byteorder_htons(y2);
+    ili9341_write_cmd(dev, ILI9341_CMD_PASET, (uint8_t *)params,
+                      sizeof(params));
+}
+
+int ili9341_init(ili9341_t *dev, const ili9341_params_t *prms)
+{
+    memcpy(&dev->params, prms, sizeof(ili9341_params_t));
```
aabadie:
```suggestion
    dev->params = *prms; /* maybe also rename prms to params ? */
```

tests/driver_ili9341/riot_logo.h
```diff
@@ -0,0 +1,71 @@
+const uint16_t picture[69][128] = {
```
aabadie: Instead of `69`, I had to use `70` with an extra empty line at the end of array to avoid having an alone green pixel at the bottom right of the logo.
I have no idea why it was there. One could think of a division by 2 issue but there is no such thing in `ili9341_map` implementation.

aabadie:
@bergzand, ping! There are still unaddressed comments and this PR would useful for #12552.

kaspar030:
The ST7789V used on the PineTime needs spi mode 3. I git a commit adding that (https://github.com/kaspar030/RIOT/tree/ili9341_spi_mode). Maybe cherry-pick in here?

kaspar030:
Otherwise, please address @aabadie's problem and rebase and squash. ;)

bergzand:
@kaspar030 @aabadie: Does it still make sense to name the driver `ili9341`? There are with @kaspar030's addition multiple different TFT display chips making use of this driver. Maybe some generic name is better in this case, but we can leave this rename for a follow up.

drivers/include/ili9341.h
```diff
@@ -0,0 +1,174 @@
...
+    gpio_t cs_pin;      /**< pin connected to the CHIP SELECT line */
+    gpio_t dcx_pin;     /**< pin connected to the DC line */
+    gpio_t rst_pin;     /**< pin connected to the reset line */
+} ili9341_params_t;
+
+/**
+ * @brief   Device descriptor for a ili9341
+ */
+typedef struct {
+    ili9341_params_t params;          /**< I2C device that sensor is connected to */
```
silkeh: :thinking:

tests/driver_ili9341/riot_logo.h
bergzand: It's an off-by-one here: https://github.com/RIOT-OS/RIOT/pull/9948/files#diff-601d0bbb085a269ae7e4f7d2ac5fe68bR257-R258

drivers/include/ili9341.h
bergzand: Renamed to pixmap and added some docs

drivers/ili9341/ili9341.c
bergzand: Reworked!

tests/driver_ili9341/riot_logo.h
bergzand: And fixed, I was able to reproduce your issue and it should be fixed in the latest iteration.

drivers/include/ili9341.h
bergzand: Fixed!

bergzand:
Rebased, fixed some issues with the driver, also cleaned a few things up

bergzand:
@aabadie @kaspar030: Still okay to squash?

bergzand:
> Maybe cherry-pick in here?

Cherry picked!

drivers/ili9341/Makefile
aabadie: still unaddressed

aabadie:
Please squash and rebase. Just a suggestion that could be nice to use with the test application: put the RIOT logo in a blob and include it to the build ? That would make it easy for testing with other images. What do you think ?

drivers/ili9341/Makefile
bergzand: Whoops, fixed

bergzand:
> Just a suggestion that could be nice to use with the test application: put the RIOT logo in a blob and include it to the build ?

I think this is not that easy, so I'd rather leave this for somebody else to follow up on. To use the blob Makefile construct, the supplied file needs to be the raw image in rgb 565 format without any metadata. I don't think there is any advantage to including that instead of the image in C header format. What I'd rather see is something similar to the blob construct, but for png images, add a number of PNG images to a Make variable and the build will convert them to C header format in raw rgb 565 format. Then we could even include a repo with RIOT logo images as external package.

bergzand:
@aabadie Rebased and squashed!

aabadie:
@bergzand, Murdock is reporting some remaining issues. Can you have a look ?

bergzand:
@aabadie I've excluded the image draw from the test when using AVR microcontrollers. On the AVR architecture, the image data ends up in the RAM. The approximately 16KB of image data doesn't fit on those devices. 

Storing them in flash would require something along the lines of the `PROGMEM` macro which is out of scope here :)

bergzand:
@aabadie I've added the `stm32f030f4-demo` to the insufficient memory list. This is the only board that causes build failures for now. Murdock says all green (except for the static test). 

Is it okay to squash again?

tests/driver_ili9341/Makefile
```diff
@@ -0,0 +1,15 @@
...
+USEMODULE += ili9341
+USEMODULE += xtimer
+
+CFLAGS += -DILI9341_LE_MODE
+
+include $(RIOTBASE)/Makefile.include
+
+# The AVR architecture stores the image in the RAM, this usually doesn't fit.
+# This flag excludes the image from the test
+ifneq (,$(filter arch_avr8,$(FEATURES_PROVIDED)))
```
aabadie:
```suggestion
ifneq (,$(filter arch_avr8,$(FEATURES_USED)))
```

Would be better I think.

bergzand: Fixed

aabadie:
Please squash!

bergzand:
Squashed!

fjmolinas:
Tested on `stm32429i-disc`, all good.

aabadie:
> Tested on stm32429i-disc, all good.

Thanks @fjmolinas !

aabadie:
All green, go!

bergzand:
:tada: 
Thanks @aabadie!

PR #9952, opened 2018-09-18
at driver: allow using same buffer for command and response in at_send_cmd_get_lines()

toonst:
### Contribution description
When reusing the same buffer for the at command and response, no command would be sent because the buffer was cleared. This is fixed by only clearing the buffer after the command has been sent.

### Testing procedure

Can be tested by using the same buffer in the `at_send_cmd_get_lines()` call for command and response:
```
char buf[16];
char *pos = buf;
int res;
pos += fmt_str(pos, "AT");
*pos = '\0';
res = at_send_cmd_get_lines(&at_dev, buf, buf, sizeof(buf), true, 10 * US_PER_SEC);
if (res < 0) {
    puts("Failed");
}
else {
    puts("Succeeded");
}
```

### DIFF
--- a/drivers/at/at.c
+++ b/drivers/at/at.c
@@ -121,14 +121,16 @@ ssize_t at_send_cmd_get_lines(at_dev_t *dev, const char *command,
     size_t bytes_left = len - 1;
     char *pos = resp_buf;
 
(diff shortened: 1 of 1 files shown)


vincent-d:
All green, let's go

PR #9958, opened 2018-09-19
periph/gpio: support for extension API (part 2)

ZetaR60:
### Contribution description
This is an implementation of GPIO redirection outlined by my proposal in [#9582](https://github.com/RIOT-OS/RIOT/issues/9582#issuecomment-410094702) (items 5-6) for the periph/gpio interface. In short, it allows other drivers (such as GPIO expanders) to be addressed using gpio_t so that all existent code can easily gain support through those drivers. For non-implementation detail and discussion, please see #9582

At @kYc0o 's suggestion, I am breaking the implementation of #9582 / #9690 into several pieces to make it easier to review. This PR only contains the API redirection code. For the intercept code (part 1) see #9860 

### Dependencies
This depends on #9860

### Testing procedure
A test case is provided that implements a soft-driver for the GPIO extension interface. The test case uses this driver to confirm that interception and redirection of the API call are working properly. This has been tested and is working properly on mega-xplained.

### Issues/PRs references
Partially replaces and closes #9190 
Implements items 5-6 (redirection) of my proposal in #9582 
For implementation of items 1-4 (interception) see #9860
Part of work tracked by #9690

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -160,7 +160,6 @@ ifneq (,$(filter ethos,$(USEMODULE)))
 endif
 
 ifneq (,$(filter extend_gpio,$(USEMODULE)))
-  FEATURES_REQUIRED += periph_gpio
   USEMODULE += extend
 endif
 
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -159,6 +159,11 @@ ifneq (,$(filter ethos,$(USEMODULE)))
   USEMODULE += tsrb
 endif
 
+ifneq (,$(filter extend_gpio,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_gpio
+  USEMODULE += extend
+endif
+
 ifneq (,$(filter feetech,$(USEMODULE)))
   USEMODULE += uart_half_duplex
 endif
@@ -166,11 +171,6 @@ endif
 ifneq (,$(filter fxos8700,$(USEMODULE)))
   USEMODULE += xtimer
   FEATURES_REQUIRED += periph_i2c
-endif
-
-ifneq (,$(filter extend_gpio,$(USEMODULE)))
-  FEATURES_REQUIRED += periph_gpio
-  USEMODULE += extend
 endif
 
 ifneq (,$(filter grove_ledbar,$(USEMODULE)))
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -170,7 +170,6 @@ endif
 
 ifneq (,$(filter extend_gpio,$(USEMODULE)))
   FEATURES_REQUIRED += periph_gpio
-  FEATURES_REQUIRED += periph_gpio_irq
   USEMODULE += extend
 endif
 
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -168,6 +168,12 @@ ifneq (,$(filter fxos8700,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter extend_gpio,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_gpio
+  FEATURES_REQUIRED += periph_gpio_irq
+  USEMODULE += extend
+endif
+
 ifneq (,$(filter grove_ledbar,$(USEMODULE)))
   USEMODULE += my9221
 endif
--- a/drivers/Makefile.include
+++ b/drivers/Makefile.include
@@ -86,6 +86,10 @@ ifneq (,$(filter fxos8700,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/drivers/fxos8700/include
 endif
 
+ifneq (,$(filter extend_gpio,$(USEMODULE)))
+  USEMODULE_INCLUDES += $(RIOTBASE)/drivers/extend/include
+endif
+
 ifneq (,$(filter grove_ledbar,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/drivers/grove_ledbar/include
 endif
--- a/drivers/Makefile.include
+++ b/drivers/Makefile.include
@@ -78,16 +78,16 @@ ifneq (,$(filter encx24j600,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/drivers/encx24j600/include
 endif
 
+ifneq (,$(filter extend_,$(USEMODULE)))
+  USEMODULE_INCLUDES += $(RIOTBASE)/drivers/extend/include
+endif
+
 ifneq (,$(filter feetech,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/drivers/feetech/include
 endif
 
 ifneq (,$(filter fxos8700,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/drivers/fxos8700/include
-endif
-
-ifneq (,$(filter extend_gpio,$(USEMODULE)))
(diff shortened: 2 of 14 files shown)


tests/extend_gpio/main.c
```diff
@@ -0,0 +1,191 @@
...
+{
+    (void)value;
+
+    printf("write on dev 0x%x with pin %u\n",
+        (uint16_t)(uintptr_t)dev, (unsigned int)pin);
+}
+
+int main(void)
+{
+    uint16_t num;
```
gschorcht: I started to adopt ESP8266 and ESP32 MCUs for your GPIO ext interface. This sould probably be 
```
 gpio_t num;
```
Otherwise you get the compile error below on platforms that use
```
typedef unsigned int gpio_t;
```
from `drivers/include/perihp/gpio.h` and where `unsigned int` is 32 bit, e.g., ESP8266 and ESP32.
```
/data/riotbuild/tests/extend_gpio/main.c:180:5: error: comparison is always true due to limited range of data type [-Werror=type-limits]
     for (num = 0; num <= GPIO_EXT_PIN_MASK; num++) {
```

drivers/Makefile.dep
```diff
@@ -152,6 +152,12 @@ ifneq (,$(filter fxos8700,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter extend_gpio,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_gpio
+  FEATURES_REQUIRED += periph_gpio_irq
```
gschorcht: Is it really necessary to always use the `periph_gpio_irq` module? IMHO, it should be the other way around. That is, the interrupt-related parts of the API should only be activated if the module `periph_gpio_irq` is used by someone.
```
#if MODULE_PERIPH_GPIO_IRQ
...
#endif
```

ZetaR60: You're probably right. Fixed.

drivers/extend/gpio_redir.c
```diff
@@ -0,0 +1,173 @@
...
+    if (gpio == GPIO_UNDEF) {
+        return NULL;
+    }
+
+    gpio_t devnum = gpio_ext_dev(gpio);
+
+    DEBUG("[gpio_ext_entry] list entry is %u\n", devnum);
+
+    /* device is greater than number of listed entries */
+    if (devnum > (sizeof(gpio_ext_list) / sizeof(gpio_ext_list[0]))) {
```
gschorcht: This has to be
```
if (devnum >= (sizeof(gpio_ext_list) / sizeof(gpio_ext_list[0]))) {
```
since the valid entry indices go from 0 ... number of entries - 1. For example, when you have only one entry and you try to access to `GPIO_EXT_PIN(1,0)`, `gpio_ext_dev` gives 1 which is equal to the number of entries and leads to a crash at the moment.

ZetaR60: Thanks for the catch. This is a holdover from when devnum was defined slightly differently. Fixed.

drivers/extend/gpio_redir.c
```diff
@@ -0,0 +1,173 @@
...
+ * @{
+ *
+ * @file
+ * @brief       GPIO extension redirection functions
+ *
+ * @author      Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * @}
+ */
+
```
gschorcht: Please add `#if  MODULE_EXTEND_GPIO ` here.

Otherwise you get compilation errors if module `extend` is enabled but not pseudomodule `extend_gpio`, e.g., because pseudomodule `extend_adc` is enabled.

drivers/Makefile.dep
```diff
@@ -152,6 +152,11 @@ ifneq (,$(filter fxos8700,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter extend_gpio,$(USEMODULE)))
```
gschorcht: I'm sorry if it seems to be a bit nit-picking, but could move this 
```
ifneq (,$(filter extend_gpio,$(USEMODULE)))
```
block before the 
```
ifneq (,$(filter feetech,$(USEMODULE)))
```
block to keep alphabetical order, please. Maybe you placed it there because you initially called it `gpio_extend`.

drivers/include/extend/gpio.h
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * @brief   Set pin mask in gpio_t
+ */
+#define GPIO_EXT_PIN_MASK    (gpio_t)(~(UINT_MAX << GPIO_EXT_DEV_LOC))
+#endif
+
+/**
+ * @brief   Callback typedef for gpio_init
+ *
+ * @see @ref drivers_periph_gpio
```
gschorcht: You might think about to replace
```suggestion
 * @see #gpio_init
```
to link to the function directly.

drivers/include/extend/gpio.h
```diff
@@ -0,0 +1,201 @@
...
+ */
+typedef int (*gpio_ext_init_t)(void *dev, gpio_t pin, gpio_mode_t mode);
+
+/**
+ * @brief   Callback typedef for gpio_init_int
+ *
+ * @see @ref drivers_periph_gpio
```
gschorcht: The same as above for all callback typedefs.

drivers/Makefile.include
```diff
@@ -82,6 +82,10 @@ ifneq (,$(filter fxos8700,$(USEMODULE)))
   USEMODULE_INCLUDES += $(RIOTBASE)/drivers/fxos8700/include
 endif
 
+ifneq (,$(filter extend_gpio,$(USEMODULE)))
```
gschorcht: Should probably be
```suggestion
ifneq (,$(filter extend,$(USEMODULE)))
```
to address any `extend_*` module.

drivers/extend/gpio_redir.c
ZetaR60: I feel like there should be a better way of accomplishing this, because this is not needed for e.g. periph, but I am not sure what it is. I have fixed it as you suggested for now.

drivers/extend/gpio_notsup.c
```diff
@@ -0,0 +1,124 @@
...
+ * @{
+ *
+ * @file
+ * @brief       GPIO extension not-supported functions
+ *
+ * @author      Matthew Blue <matthew.blue.neuro@gmail.com>
+ *
+ * @}
+ */
+
```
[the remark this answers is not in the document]
ZetaR60: Added anyway.

drivers/Makefile.dep
ZetaR60: Thanks for catching this. I find the almost-alphabetical order of the makefiles a bit annoying.

drivers/include/extend/gpio.h
ZetaR60: Fixed.

drivers/Makefile.include
ZetaR60: Fixed, and made a bit more alphabetical.

drivers/extend/gpio_redir.c
gschorcht: The reason is that `extend` module is compiled if any `extend_*` pseudomodule is enabled because of the dependency in `drivers/Makefile.dep`, e.g.,
```
ifneq (,$(filter extend_adc,$(USEMODULE)))
  USEMODULE += extend
endif
```
and all `*_extend` files reside in `drivers/extend` directory.

So, make changes to the `drivers/extend` directory if any 'extend_*` module is enabled and tries to compile all .c files found in this directory including `gpio_extend.c` independent on whether `extend_gpio` is enabled or not.

We could solve it to add a separate directory for each `extend_*` module but I would
prefer to let all `*_extend` files in one directory instead having a separate directory and to solve it with `#if MODULE_* ... #endif` wrappers.

PR #9967, opened 2018-09-19
tests/netdev_test: remove non required feature periph_timer

cladmi:
### Contribution description

Remove the non used periph_timer in test

Found out it is not actually linked.

### Testing procedure

If building works in murdock, everything should be ok.

### Issues/PRs references

Found while working on https://github.com/RIOT-OS/RIOT/pull/8711.
As the PR actually links `periph_timer` even if unused it resulted in an increased code size.


cladmi:
Thanks for reviewing.

PR #9972, opened 2018-09-20
Cleanup: drivers/makefile.dep

SemjonKerner:
This PR reorders modules in Makefile.dep alphabetically. 
Also it deletes code duplication of the srf02 module.


haukepetersen:
all green -> go

PR #9974, opened 2018-09-20
cpu/mips: remove `periph_gpio_irq` feature and stub

haukepetersen:
### Contribution description
The `gpio` driver for the `pic32` does not implement external interrupts (feature ´periph_gpio_irq`), it had simply a stub function for it. With #9035 the clean way to handle this is however to remove the stub and the feature declaration from the boards using that CPU, so applications that depend no this GPIO feature do not event build for those boards anymore...

### Testing procedure
build #9035 with and without this PR

### Issues/PRs references
Stumbled upon this while finalizing #9035
#9845 broke it :-)

### DIFF
--- a/boards/pic32-clicker/Makefile.features
+++ b/boards/pic32-clicker/Makefile.features
@@ -1,5 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
+FEATURES_PROVIDED += periph_gpio
 FEATURES_PROVIDED += periph_timer
 FEATURES_PROVIDED += periph_uart
 
--- a/boards/pic32-wifire/Makefile.features
+++ b/boards/pic32-wifire/Makefile.features
@@ -1,5 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
+FEATURES_PROVIDED += periph_gpio
 FEATURES_PROVIDED += periph_timer
 FEATURES_PROVIDED += periph_uart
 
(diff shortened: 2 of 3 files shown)


haukepetersen:
> these feature tags were added categorically using sed.

Even if so, I would expect at least the reviewers to notice these kind of things! Its just very annoying to have these simple fails all over the place all the time...

haukepetersen:
removed the patch for the `hifive1` board, as it uses a different cpu...

cladmi:
The `tests/periph_gpio` uses `gpio_init_int` so detects already if the function is there. So adding `gpio_irq` everywhere made sense except for the  boards in `BOARD_INSUFFICIENT_MEMORY`.

haukepetersen:
> The tests/periph_gpio uses gpio_init_int so detects already if the function is there. So adding gpio_irq everywhere made sense except for the boards in BOARD_INSUFFICIENT_MEMORY.

Sorry, but that is bullshit (and very unclean).

cladmi:
Murdock checks that `gpio_init_int` is declared for all boards that are compiled in `tests/periph_gpio` that is what I checked when reviewing the initial PR.

cladmi:
To keep a working git history, the commits should be in the inverted order with disabling the feature before removing the code.

haukepetersen:
Yes, but this is not about making Murdock pass somehow, but about using the feature defs as they are meant to: declare, if a certain feature is available on a certain platform!

I don't mean to be rude, its just that I think we need to be more careful when reviewing/refactoring/applying API changes, as we tend to be way to sloppy when doing these things... So next time lets look at these kind of changes more carefully!

haukepetersen:
> To keep a working git history

Last time I checked the convention was, that not every commit needs to be able to build... But I can adapt this.

haukepetersen:
changed order of commits.

gebart:
@haukepetersen it helps if every commit can build properly when doing bisections. It may not always be desirable due to the resulting size of a single commit in certain cases, but try to keep the build working for each commit if possible.

haukepetersen:
yepp, makes sense.

PR #9976, opened 2018-09-20
native: remove feature `periph_gpio_irq`

haukepetersen:
### Contribution description
See #9974. `native` does not implement any GPIO irq features, so it should not declare that feature in the board, and also the function stubs are not longer needed.

### Testing procedure
Buildtest should suffice.

### Issues/PRs references
#9974


cladmi:
To keep a working git history, the commits should be in the inverted order with disabling the feature before removing the code.

haukepetersen:
changed order of commits

haukepetersen:
All green -> go go go

PR #9983, opened 2018-09-20
examples, tests: Clean BOARD_* list in Makefiles

MrKevinWeiss:
### Contribution description

This PR rearranges the BOARD_INSUFFICIENT_MEMORY, BOARD_BLACKLIST and BOARD_WHITELIST to make add in new boards easier.

### Testing procedure

Passes murdock and review.

### Issues/PRs references

#9965


MrKevinWeiss:
I still need to run through the blacklist and whitelist (tomorrow morning).

Should we go for this until a better solution is out or should I just leave it in favor of #9081?

kaspar030:
@MrKevinWeiss thanks for tackling this! Honestly, I prefer the shorter representation (as it is now, as a short but wide block). How did you change the files? Maybe we can come up with some script?

MrKevinWeiss:
Ya I used a script but just quick and dirty python one (and looks pretty bad so I didn't want it merged).  If you say no that is fine but every time I or someone adds a board they should push the rest of the boards down.  Not the end of the world but seems unnecessary.

Your call @kaspar030, I just wanted to follow through with what I said I will do when ACKing the blackpill board.

kaspar030:
Any other opinions on line count vs. editability here?

@MrKevinWeiss how much work would it be to change your script to output blocks with line length < 80?

MrKevinWeiss:
@cladmi I would either like to (fix and) merge to close this PR.  Can you throw in your two cents so we can make a decision?

x3ro:
I'm just quoting my reason for preferring the "one board per line" option, hoping that it might get this merged :)

> @MrKevinWeiss @kaspar030 In regards to the "one board per line" vs. "80 characters per line" question, I favor one board per line because it results in much cleaner diffs. For example, in the 80 characters per line case, if I wanted to change the first board in a list, this would likely change all following lines of the list in the diff.

cladmi:
I re-arranged many applications that were not building for some boards and it was a pain.
Simplifying these task for dev would be better for me.

And one board per line does indeed help with this.

But as there is a tool that may do this, I would say it could be better to merge it along with enabling the tool check in CI.

MrKevinWeiss:
Alright closing in favor of #10111

PR #9988, opened 2018-09-21
sys/ecc: move hamming256 module to generic ecc

PeterKietzmann:
### Contribution description

This PR removes the specific `hamming256` module and moves it to generic `ecc`, similar to the structure in *sys/hashes*. This makes it clearer when adding further error correction codes.

### Testing procedure
From the RIOT root dorectory type`

`make -C tests/unittests tests-ecc all term`


kaspar030:
Doesn't this make a previously distinct module a part of "ecc"?

tests/unittests/tests-ecc/tests-ecc.c
```diff
@@ -11,15 +11,17 @@
  * @brief
  * @{
  *
- * @brief       Tests for Hamming Code implementation
+ * @brief       Tests for Error Correction Codes
+ *
  * @author      Lucas Jenß <lucas@x3ro.de>
+ * @author      Peter Kietzmann <peter.kietzmann@haw-hamburg.de>
```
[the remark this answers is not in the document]
PeterKietzmann: Whopps, originally this was part of #9989 I guess...

PeterKietzmann:
Yes, on purpose. IMO there's no need for a distinct module here and I followed the approach of *sys/hashes*. Do you revert your ACK?

kaspar030:
> the approach of _sys/hashes_. 

I think it is preferable to keep distinct modules. Not many applications will use *all* ecc algorithms (more than one screams for optimization :) ), thus there's a lot of wasted compilation.

> Do you revert your ACK?

Didn't ACK :)

> This makes it clearer when adding further error correction codes.

Can you elaborate on this? Is the "clearness" that now you can just drop a file in ecc/ and it will be compiled, vs. the mess if you want a module in a subfolder (with all the extra ```DIRS+=x```, ```USEMODULE_INCLUDES += bar```)?

PeterKietzmann:
> I think it is preferable to keep distinct modules.

Fine with me but then I propose we change *sys/hashes* as well, at least for consistency reasons.

> Didn't ACK :)

You did!

smlng

> Can you elaborate on this?

I think it was the mess if you create a module in a subfolder...

Before we change things back and forth I propose we wait for @jia200x's opinion. Afterwards I can address all changes.

PeterKietzmann:
@kaspar030, @smlng I'm sorry. Obviously I confused your comments. Your avatars just look too similar :-)

kaspar030:
> I think it was the mess if you create a module in a subfolder...

Yeah... How about using submodules? If ecc modules are contained within single files, this is much less of a hassle.

kaspar030:
> Obviously I confused your comments. Your avatars just look too similar :-)

I smiled when I saw the screenshot ;)

PeterKietzmann:
What is the benefit from submodules over regular modules (status quo in current *sys/ecc*)? Personally I still like the proposed solution better – as long as we're talking about a collection of such 'simple' functions. Anyway, let's wait for some other opinions.

kaspar030:
> What is the benefit from submodules over regular modules (status quo in current _sys/ecc_)?

When comparing one module with multiple C files to one module using submodules will only compile the C files that are depended on (e.g., selected through USEMODULE).

PeterKietzmann:
@kaspar030 that's for sure. I was just a bit confused about your proposal 

> Yeah... How about using submodules? If ecc modules are contained within single files, this is much less of a hassle.

... even if the current structure (w/o my PR) presents kind of submodule structure? However, I guess you're referring to an implementation like *sys/random*, right?

jia200x:
I personally like better this approach. For the user is quite easy to include an ECC module, a generic ECC header file and use the functions from there (same as `sys/hashes`)

> I think it is preferable to keep distinct modules. Not many applications will use all ecc algorithms (more than one screams for optimization :) ), thus there's a lot of wasted compilation.

There will be wasted compilation in some cases ([maybe not lots](https://en.wikipedia.org/wiki/Error_correction_code#List_of_error-correcting_codes) but some :) ). However, I'm not sure if this is an issue. The CI has to compile every file any way, and there's also some time spent in following Makefile targets. Also, these are usually small functions.

kaspar030:
> There will be wasted compilation in some cases ([maybe not lots](https://en.wikipedia.org/wiki/Error_correction_code#List_of_error-correcting_codes) but some :) ). 

There will be wasted compilation in *all* cases, as *none* of our applications will use *all* ecc algorithms.

> However, I'm not sure if this is an issue. The CI has to compile every file any way, and there's also some time spent in following Makefile targets. Also, these are usually small functions.

The point here is that the CI needs to be saved from compiling every file *thousands of times*, unnecessarily.

Sorry guys, either keep ecc algorithms as distinct modules or make them submodules. Unconditionally mixing different algorithms in one module's .a is counter-productive in terms of conditional compilation and RIOT modularity.

kaspar030:
> However, I guess you're referring to an implementation like _sys/random_, right?

yup!

jia200x:
> There will be wasted compilation in all cases, as none of our applications will use all ecc algorithms.

For the user, I don't see a fraction of second longer (if any) a big issue compared with to the fact of just importing a "library" for ECC. This is the usual approach for "crypto", "math", and some other libraries (see [here](https://github.com/weidai11/cryptopp) and [here](https://os.mbed.com/users/feb11/code/Crypto/)). I agree nobody will ever use 2 ECC algorithm in the same application, but I still think for the user is much easier to just import a library and use related functions.



> The point here is that the CI needs to be saved from compiling every file thousands of times, unnecessarily.

Couldn't be handled by cache?

kaspar030:
> For the user, I don't see a fraction of second longer (if any) a big issue compared with to the fact of just importing a "library" for ECC.

It must be really nice, if the problem you're solving right now can easily add a fraction of a second to compilation times because currently, compilation is *fast*. Let's keep it that way.

Seriously, fractions of a second add up. So do the compilation problems.

I you'd like to improve user experience, scan their application code for "#include "hashes/foo.h" and turn that into "USEMODULE += hashes_foo".

> This is the usual approach for "crypto", "math", and some other libraries (see [here](https://github.com/weidai11/cryptopp) and [here](https://os.mbed.com/users/feb11/code/Crypto/)).

From the second link:

> *Warning*

> If you are using SHA-384 or SHA-512, be aware that it produces large binary files and the compilation (using the online compiler) takes much longer to execute. It may happen that the compiler stops because it timed-out. In this case, just compile again and it should work.

Point here is that part of the code seems to make trouble with some compilers, thus it make sense to only compile what's needed.

> Couldn't be handled by cache?

Still not free, and CI is choking on jobs these days.

PeterKietzmann:
Ufff, that discussion escalated quickly. I'm trying to summarize what I understood and interpreted from all your comments. 

## User friendliness
IMO it doesn't matter if the user has to include *ecc*, *ecc/coder*, `USEMODULE+= ...` or similar. Also build times should be pretty uninteresting for single users. Here two examples:

### on master
```
Execution times (seconds)
 phase setup             :   0.00 ( 0%) usr   0.00 ( 0%) sys   0.00 ( 0%) wall    2432 kB (20%) ggc
 phase parsing           :   0.04 (67%) usr   0.03 (100%) sys  0.08 (80%) wall    8655 kB (69%) ggc
 phase opt and generate  :   0.00 ( 0%) usr   0.00 ( 0%) sys   0.01 (10%) wall      90 kB ( 1%) ggc
 phase last asm          :   0.02 (33%) usr   0.00 ( 0%) sys   0.01 (10%) wall    1270 kB (10%) ggc
 dump files              :   0.00 ( 0%) usr   0.00 ( 0%) sys   0.01 (10%) wall       0 kB ( 0%) ggc
 preprocessing           :   0.01 (17%) usr   0.00 ( 0%) sys   0.02 (20%) wall    3768 kB (30%) ggc
 lexical analysis        :   0.02 (33%) usr   0.00 ( 0%) sys   0.02 (20%) wall       0 kB ( 0%) ggc
 parser (global)         :   0.01 (17%) usr   0.01 (33%) sys   0.01 (10%) wall    1004 kB ( 8%) ggc
 parser struct body      :   0.00 ( 0%) usr   0.02 (67%) sys   0.00 ( 0%) wall    1100 kB ( 9%) ggc
 parser inl. func. body  :   0.00 ( 0%) usr   0.00 ( 0%) sys   0.01 (10%) wall     225 kB ( 2%) ggc
 symout                  :   0.02 (33%) usr   0.00 ( 0%) sys   0.03 (30%) wall    3813 kB (31%) ggc
 TOTAL                   :   0.06             0.03             0.10              12457 kB
"make" -C /media/DATA/RIOT/sys
"make" -C /media/DATA/RIOT/sys/ecc/hamming256
```

### with this PR (plus two additional codes in the ecc folder)
```
Execution times (seconds)
 phase setup             :   0.00 ( 0%) usr   0.00 ( 0%) sys   0.00 ( 0%) wall    2432 kB (20%) ggc
 phase parsing           :   0.07 (78%) usr   0.02 (100%) sys  0.09 (82%) wall    8655 kB (69%) ggc
 phase opt and generate  :   0.00 ( 0%) usr   0.00 ( 0%) sys   0.01 ( 9%) wall      90 kB ( 1%) ggc
 phase last asm          :   0.02 (22%) usr   0.00 ( 0%) sys   0.01 ( 9%) wall    1270 kB (10%) ggc
 preprocessing           :   0.04 (44%) usr   0.01 (50%) sys   0.08 (73%) wall    3768 kB (30%) ggc
 lexical analysis        :   0.01 (11%) usr   0.00 ( 0%) sys   0.00 ( 0%) wall       0 kB ( 0%) ggc
 parser struct body      :   0.01 (11%) usr   0.01 (50%) sys   0.01 ( 9%) wall    1100 kB ( 9%) ggc
 scheduling 2            :   0.00 ( 0%) usr   0.00 ( 0%) sys   0.01 ( 9%) wall       5 kB ( 0%) ggc
 symout                  :   0.03 (33%) usr   0.00 ( 0%) sys   0.01 ( 9%) wall    3813 kB (31%) ggc
 TOTAL                   :   0.09             0.02             0.11              12457 kB
"make" -C /media/DATA/RIOT/sys
"make" -C /media/DATA/RIOT/sys/ecc
```

-> 30 ms is nothing for one person.

## Build system integration
Actual modules seem somewhat overkill to me. If this would be the preferred way I'd ask if 'modularizing' `hashes` would be the way to go consequently. 
Submodules seem like a promising alternative **but** fiddling around with manual include dirs is (i) not developer friendly and (ii) might collide with ongoing efforts to improve resolving build dependencies and OS configurations (@jia200x ???)

## CI performance
The CI needs to build modules thousands of times and I understand that even milliseconds sum up. On the other hand I'm wondering if we should base this specific decision on CI properties. Furthermore I did not completely understand the issue with ccache. If the built module is cached, this overhead just adds on the first build, right?

kaspar030:
> `ecc_hamming256.a` doesn't exist.

Currently you'll have to add ```ecc_%``` to PSEUDOMODULES.

PeterKietzmann:
@kaspar030 would you ACK?

kaspar030:
Thanks @PeterKietzmann!

PeterKietzmann:
So I squashed my last commit. Murdock is happy and I see two approvals. Who will hit the button?

kaspar030:
I assume @smlng is alright with the current solution. &go.

Makefile.dep
```diff
@@ -811,3 +811,7 @@ USEPKG := $(sort $(USEPKG))
 ifneq ($(OLD_USEMODULE) $(OLD_USEPKG),$(USEMODULE) $(USEPKG))
   include $(RIOTBASE)/Makefile.dep
 endif
+
+ifneq (,$(filter ecc_%,$(USEMODULE)))
+  USEMODULE += ecc
+endif
```
maribu: @PeterKietzmann: Is it by intent that this was put out of the recursive dependency tracking and after the sorting/deduplication of `USEMODULE` entries? This looks like a bug to me.

PeterKietzmann: I don't quite understand. What would be the fix?

maribu: Move that code above this:
https://github.com/RIOT-OS/RIOT/blob/01fb5335a28551195c31b6c40a1a5185a46694a2/Makefile.dep#L859-L864

Personally, I would move it up even more, so that it ends up above all of theses lines:

https://github.com/RIOT-OS/RIOT/blob/01fb5335a28551195c31b6c40a1a5185a46694a2/Makefile.dep#L837-L864

maribu: To explain what the bug is: in
``` Makefile
ifneq ($(OLD_USEMODULE) $(OLD_USEPKG),$(USEMODULE) $(USEPKG)) 
endif 
```

The file `Makefile.dep` is included again until no changes in the used modules have taken place. E.g. lets assume at the top of the file some additional modules would have been added if module `ecc` was enabled, but that code did not trigger as `ecc` was not part of `USEMODULE` at that time. The recursive include makes sure that every dependency is caught. Moving the code below that will prevent this from happening.

Another issue is that this line

``` Makefile
USEMODULE := $(sort $(USEMODULE)) 
```

make sure that every module is only once in `USEMODULE` (it sorts and removes duplicates). If `USEMODULE` would contain duplicates, compilation would fail. E.g. if `ecc` was already part of `USEMODULE` the insertion of `ecc` in line 816 would result in such a duplicate.

Being honest, I'm not really an expert in the build system. Maybe @kaspar030 or @cladmi could confirm if I got the details right.

maribu: I was wrong,  duplicates in `USEMODULE` will not always result in build failures. But it will result at least in ugly warnings. E.g. when simply adding `USEMODULE += core_msg` at the bottom of `Makefile.dep` in the current master branch to have it listed twice (`core_msg` is a default module), I get the following output:

```
examples/hello-world $ make
Building application "hello-world" for "nucleo-f303re" with MCU "stm32f3".

"make" -C /home/maribu/Repos/software/RIOT/boards/nucleo-f303re
"make" -C /home/maribu/Repos/software/RIOT/boards/common/nucleo
"make" -C /home/maribu/Repos/software/RIOT/core
/home/maribu/Repos/software/RIOT/Makefile.base:82: target '/home/maribu/Repos/software/RIOT/examples/hello-world/bin/nucleo-f303re/core/msg.o' given more than once in the same rule
/home/maribu/Repos/software/RIOT/Makefile.base:82: target '/home/maribu/Repos/software/RIOT/examples/hello-world/bin/nucleo-f303re/core/msg.o' given more than once in the same rule
```

PeterKietzmann: @maribu thanks for your investigation but the problem is still not clear to my. Ergo, I won't open a PR. Maybe you better open it and have the discussion with @kaspar030 or @cladmi there?

maribu: I'm on it :-)

cladmi: The issue with the dependency being after `OLD_USEMODULE` lines, is that if `ecc` has a dependency, they will not be resolved. So being there is wrong. There is a reason why all the others are before :D 
There are cases with `default implementation handling` that may not work with the current scheme but it is not the case here.

The issue with `core_msg` is that `Makefile.base` does not handle having `USEMODULE` with duplicates when handling `SUBMODULES` and should just be handled there by doing `sort` before listing the `.c` files.

PR #9996, opened 2018-09-21
cpu/sam0_common/gpio: use gpio_irq feature

haukepetersen:
See #9992

### DIFF
--- a/cpu/sam0_common/periph/gpio.c
+++ b/cpu/sam0_common/periph/gpio.c
@@ -32,16 +32,18 @@
 #include "debug.h"
 
 /**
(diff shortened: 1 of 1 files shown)


dylad:
@haukepetersen `DISABLE_MODULE += periph_gpio_irq` leads to unsuccessful build because there is no guard in tests/periph_gpio. This issue is also true for others CPU implementations to me.
This should be fix before merging this one.

dylad:
Oh great it works once rebased, thanks @gebart :)

dylad:
All green -> Go.

PR #6512, opened 2017-01-28
drivers/ds18: Add Maxim Integrated ds18 driver

phectori:
This patch adds the driver for the Maxim Integrated [ds1822](https://www.maximintegrated.com/en/products/analog/sensors-and-sensor-interface/DS1822.html) and [ds18b20](https://www.maximintegrated.com/en/products/analog/sensors-and-sensor-interface/DS18B20.html) temperature sensors. These sensors use one-wire as communication protocol and can be powered using the data-line.

Currently I have only tested the ds1822, but according to the datasheet they should be software compatible.

The driver is SAUL compatible.

PR #10011, opened 2018-09-21
drivers/ds18: Add Maxim Integrated 1-Wire temperature sensor driver

leandrolanzieri:
### Contribution description
This PR is based on #6512.

What this PR adds:
- Fixes the temperature reading problem reported in #6512.
- Fixes some type errors on the API.
- Adds a timeout to the ```ds18_read_bit``` function.
- Test application for the driver.

### Testing procedure
1. Connect the sensor to a free GPIO on the board.
2. Define  ```DS18_PARAM_PIN``` and ```DS18_PARAM_PULL``` accordingly to your board.
3. Run the test application with something like: ```BOARD=lobaro-lorabox make flash term -C tests/driver_ds18```.
4. You should see an output similar to:

```
# DS18B20 test application
# 
# +------------Initializing------------+
# 
# +--------Starting Measurements--------+
# Temperature [ºC]: 25.87
# +-------------------------------------+
```

### Issues/PRs references
Based on #6512.


leandrolanzieri:
@aabadie Do you still have access to the hardware to test this?

aabadie:
> Do you still have access to the hardware to test this?

Normally yes but I have to find it in my mess :)

aabadie:
@leandrolanzieri, I found it but I still can't make the sensor to give a valid temperature (tested with different DS18_PARAM_PULL values PU/PD) but it always display the following (with debug enabled):
```sh
2018-09-25 11:08:08,323 - INFO # [DS18] Reset and read scratchpad
2018-09-25 11:08:08,325 - INFO # [DS18] Received byte: 0xff
2018-09-25 11:08:08,326 - INFO # [DS18] Received byte: 0xff
2018-09-25 11:08:08,326 - INFO # Temperature [ºC]: 163.77
```

I tested with an Arduino Zero and a Nucleo L073. I setup the pin correctly for both boards. Maybe my sensor is broken ? I only have this one so I cannot test with another one to check this.

leandrolanzieri:
@aabadie thanks a lot for testing!
On the lobaro_lorabox board I found that the problem was that ```DS18_SAMPLE_TIME``` was too big and the bits were always read as '1', so I reduced that and solved it. I don't know if this is caused by a not accurate time measurement.
I will try to get a sensor so I can perform further tests.

danpetry:
Partial ACK on the fundamentals.
PR reasoning and testing instructions are clearly laid out and I can't see any reason this breaks from current concepts. Rights of current PR authors are respected in commit history.

drivers/include/ds18.h
```diff
@@ -0,0 +1,87 @@
...
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_ds18 DS18 temperature sensor driver
+ * @ingroup     drivers_sensors
+ * @ingroup     drivers_saul
+ * @brief       Driver interface for the DS18 temperature sensors
+ *
+ * This driver provides @ref drivers_saul capabilities.
```
haukepetersen: Would you please also add some more documentation about the supported (or not supported) features and limitations of this driver?!

E.g.
- does not allow for addressing devices -> only supports a single device on the bus
- hardcodes the 1wire bus functionality into the driver
- does not allow for configuration of the sampling width

drivers/ds18/ds18.c
```diff
@@ -0,0 +1,196 @@
...
+    uint32_t start, measurement = 0;
+
+    /* Initiate read slot */
+    ds18_low(dev);
+    xtimer_usleep(1);
+    ds18_release(dev);
+
+    /* Measure time low of device pin, timeout after slot time*/
+    start = xtimer_now_usec();
+    while (!gpio_read(dev->pin) && measurement < DS18_DELAY_SLOT) {
```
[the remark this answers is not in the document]
haukepetersen: True, more-or-less precise timings are quite mandatory for using my proposed code block...

So until we get the timer imprecision sorted out, how about we go with a dual approach? We could include both code blocks in the code, `#if - #else`ing them based on a sub(pseudo)module, making my variant optional. This way people can use the more efficient code once they made sure their platform is handle the needed timings...

Of course we would need to document this in the module description :-)

leandrolanzieri:
I added a submodule `ds18_optimized` to change de read function. Also rebased.

drivers/ds18/ds18.c
```diff
@@ -0,0 +1,232 @@
...
+    *temperature = (int16_t)(measurement / 100);
+
+    return DS18_OK;
+}
+
+int ds18_get_temperature(ds18_t *dev, int16_t *temperature)
+{
+    int res;
+
+    res = ds18_reset(dev);
```
haukepetersen: please remove this here, it is done already in `ds18_trigger()`

drivers/ds18/ds18.c
```diff
@@ -0,0 +1,232 @@
...
+    dev->in_mode = (params->out_mode == GPIO_OD_PU) ? GPIO_IN_PU : GPIO_IN;
+
+    /* Initialize the device and the pin */
+    dev->pin = params->pin;
+    dev->out_mode = params->out_mode;
+    res = gpio_init(dev->pin, dev->in_mode) == 0 ? DS18_OK : DS18_ERROR;
+
+#if defined(MODULE_DS18_OPTIMIZED)
+    DEBUG("[DS18] Using optimized read function");
+#endif
```
haukepetersen: please remove this block, I think it is rather for internal debugging...

drivers/ds18/ds18_internal.h
```diff
@@ -0,0 +1,66 @@
...
+#define DS18_DELAY_RESET            (480U)
+#define DS18_DELAY_PRESENCE         (60U)
+#define DS18_DELAY_SLOT             (60U)
+#define DS18_SAMPLE_TIME            (10U)
+#define DS18_DELAY_CONVERT          (750U * US_PER_MS)
+#define DS18_DELAY_RW_PULSE         (1U)
+#define DS18_DELAY_R_RECOVER        (DS18_DELAY_SLOT - DS18_SAMPLE_TIME)
+/** @} */
+
+
```
haukepetersen: remove newline

drivers/include/ds18.h
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * @return                  0 on success
+ * @return                 -1 on error
+ */
+int ds18_read(ds18_t *dev, int16_t *temperature);
+
+/**
+ * @brief   convenience fuction for triggering a conversion and reading the
+ * value
+ *
```
haukepetersen: here should also be a note or description that this function will block for an amount of time (don't remember the value...).

sys/auto_init/auto_init.c
```diff
@@ -437,6 +437,42 @@ void auto_init(void)
     extern void auto_init_veml6070(void);
     auto_init_veml6070();
 #endif
+#ifdef MODULE_DS18
```
haukepetersen: seems like something went wrong when you rebased, most of these entries are unrelated to this PR, right?!

drivers/ds18/ds18.c
leandrolanzieri: Done.

drivers/ds18/ds18.c
leandrolanzieri: Removed

drivers/ds18/ds18_internal.h
leandrolanzieri: Done

sys/auto_init/auto_init.c
leandrolanzieri: Oops, I screwed up yep :(

drivers/include/ds18.h
leandrolanzieri: Added a note on that.

leandrolanzieri: I added a note on that.

danpetry:
Yes I'll do that tomorrow.

drivers/ds18/ds18.c
```diff
@@ -0,0 +1,223 @@
...
+    xtimer_usleep(DS18_DELAY_R_RECOVER);
+    return DS18_OK;
+#else
+    uint32_t start, measurement = 0;
+
+    /* Measure time low of device pin, timeout after slot time*/
+    start = xtimer_now_usec();
+    while (!gpio_read(dev->pin) && measurement < DS18_DELAY_SLOT) {
+        measurement = xtimer_now_usec() - start;
+    };
```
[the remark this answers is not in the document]
leandrolanzieri: Done

drivers/ds18/ds18_internal.h
```diff
@@ -0,0 +1,64 @@
+/*
```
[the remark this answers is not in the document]
leandrolanzieri: I haven't seen an `internal` folder in the common drivers structure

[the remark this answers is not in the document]
danpetry: Sorry, `/include`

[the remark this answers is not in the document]
leandrolanzieri: Here https://github.com/RIOT-OS/RIOT/pull/10011#discussion_r220998575 @haukepetersen asked to move it out of the folder so it is not in the global path.

leandrolanzieri:
After @danpetry reported failures in readings with `sensebox_samd21` board I performed some measurements, on both the `sensebox_samd21` and `samr21_xpro` boards. Here are the captures together with a summary of the measurements: [time_measurements.zip](https://github.com/RIOT-OS/RIOT/files/2506261/time_measurements.zip)

I found that the delayed introduced by the call to `xtimer_usleep` at the beginning of the read and write slots was causing some trouble, so I decided to remove it. The delay in the gpio toggle is enough to start a slot.

drivers/ds18/ds18_internal.h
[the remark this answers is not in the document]
danpetry: ok

PR #10017, opened 2018-09-23
cpu/esp8266: periph/i2c fix

gschorcht:
The PR contains a fix for the bit-banging protocol implementation of I2C for the esp8266 in case of error conditions, e.g., when the slave does not react to a byte transfer and a NAK is read by the master.

Without this fix, the protocol implementation simply returns with an error code and the last bit transfer condition is kept, i.e., the SCL signal is kept low. With the fix, a STOP condition is generated in case of error conditions to abort the failed transfer and to bring the I2C bus back into a defined state.

### DIFF
--- a/cpu/esp8266/periph/i2c.c
+++ b/cpu/esp8266/periph/i2c.c
@@ -53,8 +53,8 @@
 extern uint8_t system_get_cpu_freq(void);
 extern bool system_update_cpu_freq(uint8_t freq);
 
-/* max clock stretching counter */
-#define I2C_CLOCK_STRETCH 400
+/* max clock stretching counter (ca. 10 ms) */
+#define I2C_CLOCK_STRETCH 40000
 
 typedef struct
 {
@@ -128,9 +128,9 @@ static int _i2c_read_bit (_i2c_bus_t* bus, bool* bit);
 static int _i2c_write_byte (_i2c_bus_t* bus, uint8_t byte);
 static int _i2c_read_byte (_i2c_bus_t* bus, uint8_t* byte, bool ack);
 static int _i2c_arbitration_lost (_i2c_bus_t* bus, const char* func);
+static void _i2c_abort (_i2c_bus_t* bus, const char* func);
 
 /* implementation of i2c interface */
-
 void i2c_init(i2c_t dev)
 {
     if (I2C_NUMOF != sizeof(_i2c_bus)/sizeof(_i2c_bus_t)) {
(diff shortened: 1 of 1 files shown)


PeterKietzmann:
@MrKevinWeiss would you please have a look at this?!

MrKevinWeiss:
OK I should be able to get it reviewed at the Hack'n'Ack tonight (busy now)!

cpu/esp8266/periph/i2c.c
```diff
@@ -130,7 +130,7 @@ static int _i2c_read_byte (_i2c_bus_t* bus, uint8_t* byte, bool ack);
 static int _i2c_arbitration_lost (_i2c_bus_t* bus, const char* func);
 
 /* implementation of i2c interface */
-
+#include "xtimer.h"
```
MrKevinWeiss: Why add xtimer?

gschorcht: @MrKevinWeiss There is no reason. Probably, I forgot to to remove it when I tried to do some timings using xtimer. Should I remove it?

MrKevinWeiss: Yes please!

cpu/esp8266/periph/i2c.c
```diff
@@ -484,6 +496,8 @@ static /* IRAM */ int _i2c_stop_cond(_i2c_bus_t* bus)
     /* wait t_BUF - bus free time between a STOP and a START condition */
     /* min. in us: 4.7 (SM), 1.3 (FM), 0.5 (FPM), 0.16 (HSM); no max. */
     _i2c_delay (bus);
+    /* one additional delay */
+    _i2c_delay (bus);
```
MrKevinWeiss: Also why so many _i2c_delay?

gschorcht:
@MrKevinWeiss Please wait a bit more, during a test last night, the slave block SDA line again after I removed the additional delay in STOP condition. I have to investigate it again.

MrKevinWeiss:
OK, I will monitor.

gschorcht:
@MrKevinWeiss What I can tell, after adding this additional delay in STOP condition, the slave didn't block the SDA line for the last 5 hours with 100 bus accesses per seconds. It seems that the longer delay in STOP condition is critical. Since this additional delay of one quarter clock cycle is only in STOP condition, I don't expect any performance degradation.  I would prefer to leave it there for stability reasons.

gschorcht:
@MrKevinWeiss After a 24 h+ stress test with the slave that blocked the SDA line and the double delay in STOP condition passed.

Please don't wonder, I have replaced the execution of a complete STOP condition to abort the transfer in error conditions by a separate abort function which simply sets the bus signals back to the default state. Executing a complete STOP condition produces problems in some error conditions.

From my point of view it is finished and I can sqash and rebase if necessary.

MrKevinWeiss:
@gschorcht OK, sorry for not getting back sooner.  Please cleanup the //delay comment on line 512 and squash then it should be good.  Thanks for the extra work!

gschorcht:
@MrKevinWeiss Sorry, switching between too many branches today caused the mistake with the ```// _i2c_delay``` line. It was just to see whether I can reproduce the problem after all changes. No, it should be ok.

gschorcht:
Thanks

PR #10021, opened 2018-09-24
mips: Fix deps for MIPS newlib

gebart:
### Contribution description

Adjust dependencies and fix some duplicate symbol definitions in certain situations for MIPS

### Testing procedure

 - [ ] Check that all mips boards still build (CI)
 - [ ] Run some test application on a mips board (preferably mips-malta, but any pic32 works too)

### Issues/PRs references

Split from #8711

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -391,7 +391,14 @@ endif
 
 ifneq (,$(filter stdio_uart,$(USEMODULE)))
   USEMODULE += isrpipe
-  FEATURES_REQUIRED += periph_uart
+  ifneq (,$(filter newlib_syscalls_mips_uhi,$(USEMODULE)))
+    # The periph_uart driver is optional when using MIPS UHI syscalls (debug
+    # UART is used on PIC32 boards, but not on mips-malta)
+    FEATURES_OPTIONAL += periph_uart
+  else
+    # All other UART implementations require the periph_uart driver.
+    FEATURES_REQUIRED += periph_uart
+  endif
 endif
 
 ifneq (,$(filter isrpipe,$(USEMODULE)))
--- a/cpu/mips32r2_common/newlib_syscalls_mips_uhi/syscalls.c
+++ b/cpu/mips32r2_common/newlib_syscalls_mips_uhi/syscalls.c
@@ -46,6 +46,8 @@ extern char _sheap; /* start of the heap */
 extern char _eheap; /* end of the heap */
 char *heap_top = &_sheap + 4;
 
+/* MIPS newlib crt implements _init, _fini, and _exit, and manages the heap */
+#ifndef __mips__
 /**
  * @brief Free resources on NewLib de-initialization, not used for RIOT
  */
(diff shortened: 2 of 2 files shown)


Makefile.dep
```diff
@@ -392,7 +392,14 @@ endif
 
 ifneq (,$(filter stdio_uart,$(USEMODULE)))
   USEMODULE += isrpipe
-  FEATURES_REQUIRED += periph_uart
+  ifneq (,$(filter newlib_syscalls_mips_uhi,$(USEMODULE)))
+    # The periph_uart driver is optional when using MIPS UHI syscalls (debug
+    # UART is used on PIC32 boards, but not on mips-malta)
+    FEATURES_OPTIONAL += periph_uart
```
cladmi: Now that I look at it on its own, does this somehow means that `newlib_syscalls_mips_uhi` implements the `periph_uart` functions on mips malta and could be described the other way around ? Like adding this somewhere in the `mips3r2_generic` cpu for mips-malta:

    ifneq (,$(filter newlib_syscalls_mips_uhi,$(USEMODULE))
    endif

With of course a comment saying it is not really an uart but implements the interface.

The fact that it adds ` periph_uart` pseudomodule to USEMODULE should not be a problem in that case

cladmi: From doing `mips-mti-elf-objdump -t examples/hello-world/bin/mips-malta/hello-world.elf  | grep stdio ` I get no reference to the `stdio_uart` module. I could even compile `hello-world` when removing `stdio_uart`.

Maybe a solution would be more to not include `stdio_uart` for `mips_malta`.

It compiles with:

``` diff
index 25cdb71af..e12d35924 100644
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -367,7 +367,7 @@ ifneq (,$(filter newlib,$(USEMODULE)))
   ifeq (,$(filter newlib_syscalls_%,$(USEMODULE)))
   endif
-  ifeq (,$(filter stdio_rtt,$(USEMODULE)))
+  ifeq (,$(filter stdio_%,$(USEMODULE)))
     USEMODULE += stdio_uart
   endif
 endif
diff --git a/cpu/mips32r2_generic/Makefile.include b/cpu/mips32r2_generic/Makefile.include
index efb2a88b1..27e261919 100644
--- a/cpu/mips32r2_generic/Makefile.include
+++ b/cpu/mips32r2_generic/Makefile.include
@@ -2,6 +2,14 @@ export MEMORY_BASE=0x80000000
 export MEMORY_SIZE=1M
 export APP_START=0x80000000
 
+
+# newlib_syscalls_mips_uhi provides directly the 'stdio' functions
+# This prevents using 'stdio_uart' module
+ifeq ($(USE_UHI_SYSCALLS),1)
+  USEMODULE += stdio_mips_uhi
+endif
+PSEUDOMODULES += stdio_mips_uhi
+
 include $(RIOTMAKE)/arch/mips.inc.mk
 include $(RIOTCPU)/mips32r2_common/Makefile.include
 
```

I did not test it though.

cladmi: With this, it builds all compatible applications.

PR #10026, opened 2018-09-25
cpu/stm32_common: EEPROM speedup

olegart:
### Contribution description

Changing EEPROM access from byte to 32-bit word improves access speed by up to 8 times. Unaligned and less than 4 bytes access still possible using same functions.

P.S. _wait_for_pending_operations moved from flashpage to flash_common and used in eeprom.c to prevent flash controller busy errors.

### Testing procedure

'test' command was added to tests/periph_eeprom

Before (STM32L151CCU6 @ 32 MHz):
Writing 256 bytes to clean EEPROM: 1709029 usec
Writing 256 bytes to dirty EEPROM: 1728992 usec
Reading 256 bytes from EEPROM: 408 usec

After:
Writing 256 bytes to clean EEPROM: 216165 usec
Writing 256 bytes to dirty EEPROM: 432014 usec
Reading 256 bytes from EEPROM: 242 usec

### DIFF
--- a/cpu/stm32_common/periph/eeprom.c
+++ b/cpu/stm32_common/periph/eeprom.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2018 Inria
+ * Copyright (C) 2018 Unwired Devices LLC
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -15,11 +16,13 @@
  * @brief       Low-level eeprom driver implementation
  *
  * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
+ * @author      Oleg Artamonov <oleg@unwds.com>
  *
  * @}
  */
 
 #include <assert.h>
+#include <string.h>
 
 #include "cpu.h"
 
@@ -30,17 +33,47 @@
 
 extern void _lock(void);
 extern void _unlock(void);
+extern void _wait_for_pending_operations(void);
 
 #ifndef EEPROM_START_ADDR
 #error "periph/eeprom: EEPROM_START_ADDR is not defined"
 #endif
 
-uint8_t eeprom_read_byte(uint32_t pos)
+static union {
+    uint32_t data_word;
+    uint8_t  data_bytes[4];
+} eeprom_data;
+
+static void _eeprom_write_byte(uint32_t pos, uint8_t byte)
 {
     assert(pos < EEPROM_SIZE);
 
-    DEBUG("Reading data from EEPROM at pos %" PRIu32 "\n", pos);
-    return *(uint8_t *)(EEPROM_START_ADDR + pos);
+    _wait_for_pending_operations();
+    *(__IO uint8_t *)(EEPROM_START_ADDR + pos) = byte;
+}
+
+static void _eeprom_write_word(uint32_t pos, uint32_t word)
+{
+    assert(pos <= EEPROM_SIZE - sizeof(uint32_t));
+
+    _wait_for_pending_operations();
+    *(__IO uint32_t *)(EEPROM_START_ADDR + pos) = word;
+}
+
+static uint8_t _eeprom_read_byte(uint32_t pos)
+{
+    assert(pos < EEPROM_SIZE);
+
+    _wait_for_pending_operations();
+    return *(__IO uint8_t *)(EEPROM_START_ADDR + pos);
+}
+
+static uint32_t _eeprom_read_word(uint32_t pos)
+{
+    assert(pos <= EEPROM_SIZE - sizeof(uint32_t));
+
+    _wait_for_pending_operations();
+    return *(__IO uint32_t *)(EEPROM_START_ADDR + pos);
 }
 
 void eeprom_write_byte(uint32_t pos, uint8_t data)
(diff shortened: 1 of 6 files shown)


cpu/stm32_common/periph/eeprom.c
```diff
@@ -20,6 +20,9 @@
  */
 
 #include <assert.h>
+#include <string.h>
+
+#if defined(CPU_FAM_STM32L0) || defined(CPU_FAM_STM32L1)
```
aabadie: This is not needed as it is already handled by the build system

cpu/stm32_common/periph/eeprom.c
```diff
@@ -30,25 +33,155 @@
...
 #error "periph/eeprom: EEPROM_START_ADDR is not defined"
 #endif
 
-uint8_t eeprom_read_byte(uint32_t pos)
+static void _eeprom_write_byte(uint32_t pos, uint8_t byte)
 {
     assert(pos < EEPROM_SIZE);
+    
+    _wait_for_pending_operations();
+    *(__IO uint8_t *)(EEPROM_START_ADDR + pos) = byte;
```
aabadie: `__IO` is not needed I think. But if you have a good reason to keep it, then it should also be used in read functions.

cpu/stm32_common/periph/eeprom.c
```diff
@@ -30,25 +33,155 @@
...
+    _wait_for_pending_operations();
+    return *(uint32_t *)(EEPROM_START_ADDR + pos);
+}
+
 void eeprom_write_byte(uint32_t pos, uint8_t data)
 {
     assert(pos < EEPROM_SIZE);
 
-    DEBUG("Writing data '%c' to EEPROM at pos %" PRIu32 "\n", data, pos);
+    DEBUG("Writing data '%c' to EEPROM at pos %lu\n", data, pos);
```
aabadie: Unrelated change and won't build with CLang

cpu/stm32_common/periph/eeprom.c
```diff
@@ -30,25 +33,155 @@
...
+            if (data)
+                _eeprom_write_byte(pos + bytes, data[bytes]);
+            else
+                _eeprom_write_byte(pos + bytes, 0);
+            bytes++;
+        }
+    }
+    
+    /* write by words to speed up the process */
+    uint32_t words = (len - bytes)/4;
```
aabadie: In general, we put spaces around operators (although it's explicitly stated in [RIOT coding conventions](https://github.com/RIOT-OS/RIOT/wiki/Coding-conventions))

cpu/stm32_common/periph/eeprom.c
```diff
@@ -30,25 +33,155 @@
...
+    _lock();
+    
+    return bytes;
+}
+
+uint8_t eeprom_read_byte(uint32_t pos)
+{
+    assert(pos < EEPROM_SIZE);
+
+    DEBUG("Reading data from EEPROM at pos %lu\n", pos);
```
aabadie: Use PRIu32 insteand of %lu (see comment above)

cpu/stm32_common/periph/eeprom.c
```diff
@@ -30,25 +33,155 @@
...
+            data[bytes] = _eeprom_read_byte(pos + bytes);
+            bytes++;
+        }
+    }
+    _lock();
+    
+    return bytes;
+}
+
+#endif
```
aabadie: missing new line at end of file

cpu/stm32_common/periph/flash_common.c
```diff
@@ -40,6 +40,14 @@
 #define KEY_REG                (FLASH->KEYR)
 #endif
 
+typedef enum {
```
aabadie: This enum is not used

cpu/stm32_common/periph/flash_common.c
```diff
@@ -56,3 +64,23 @@ void _lock(void)
...
+            FLASH->SR |= (uint32_t)FLASH_SR_WRPERR;
+            while (FLASH->SR & FLASH_SR_BSY) {}
+        }
+    }
+
+    /* Clear 'end of operation' bit in status register */
+    if (FLASH->SR & FLASH_SR_EOP) {
+        FLASH->SR &= ~(FLASH_SR_EOP);
+    }
+}
```
aabadie: No newline at end of file

tests/periph_eeprom/main.c
```diff
@@ -98,17 +99,58 @@ static int cmd_write(int argc, char **argv)
     return 0;
 }
 
+static int cmd_test(int argc, char **argv)
+{
+    (void)argc;
+    (void)argv;
+    
+    uint8_t bytes[256];
+    eeprom_clear(0, 256);
```
aabadie: This function doesn't exist (so I don't think will ever build) but it's interesting to have it (something I have in mind already). Are you willing to propose this in another PR ?

tests/periph_eeprom/main.c
```diff
@@ -98,17 +99,58 @@ static int cmd_write(int argc, char **argv)
     return 0;
 }
 
+static int cmd_test(int argc, char **argv)
+{
+    (void)argc;
+    (void)argv;
+    
+    uint8_t bytes[256];
```
aabadie: You could put this 256 in a macro.

tests/periph_eeprom/main.c
```diff
@@ -98,17 +99,58 @@ static int cmd_write(int argc, char **argv)
...
+    { "test", "Test write and read speed", cmd_test},
     { NULL, NULL, NULL }
 };
 
 int main(void)
 {
     puts("EEPROM read write test\n");
     puts("Please refer to the README.md for more details\n");
+    
+    xtimer_init();
```
aabadie: I haven't checked but I'm wondering why you need this.

tests/periph_eeprom/main.c
olegart: Yep, we have it in our code and forget to remove it here. Actually it's just eeprom_write with all zeros.

tests/periph_eeprom/main.c
olegart: Can be removed actually - auto_init is expected to do it anyway.

olegart:
Ok, so

* comments noted. I left __IO modifier in place - usually it translates to 'volatile', and it's a bit safer to have volatile access to hardware storage. At least, no funny compilers' optimizations to expect.

* read/write functions further improved - there's no less-than-word access anymore. Accessing single bytes are really expensive and translates to multiple read-modify-clear-write operations, wearing eeprom out. All accesses with eeprom_read and eeprom_write are aligned 1-word operations on hardware level now.

* code tested on our production firmware (stores hundreds of bytes of data in EEPROM in chunks from 1 byte to few hundreds bytes in size), it works ok. Someone may want to check aligned / unaligned access with different data sizes and start addresses, but as I see it now, everything is ok.

* there will be no problems with flash_common - there's a single function implementation had been moved from flashpage to flash_common. FLASH->SR must be checked for BUSY and ERROR flags before any flash *or EEPROM* read/write operation, so _wait_for_pending_operations() should never be limited to flashpage.c in the first place.

* actually, eeprom_clear implementation is already in place - just call eeprom_write with NULL instead of data pointer.

aabadie:
Thanks for addressing my comments quickly @olegart. I may reconsider my request about contributing the clear functions in a separate PR.

As you can see in the PR status, some static check are still failing on Travis. They are about commit messages too long and trailing whitespaces.

Please fix them in fixup commits. We want the git history to remain as clean as possible, so as I see this PR now, you'll have to rework the git history. Let me know if this causes you problems, I can help.

olegart:
Ok, I'll fix it.

olegart:
@aabadie, I'm done fixing it.

aabadie:
Thanks @olegart, it looks super good now. I'll test the PR tomorrow on a couple of boards.

By the way, you removed the clear functions, do you plan to open another PR to add them ? It can be based on this one if you want.

olegart:
@aabadie, yep, clear functions needs to be added to atmega_common as well, so separate PR for it.

aabadie:
> clear functions needs to be added to atmega_common as well, so separate PR for it

I think there's no need to modify the CPU specific code for clear functions, see #10056. I also fixed there the bounds issues (last byte in EEPROM couldn't be read/written)

aabadie:
@olegart, after working on #10056, I had a second look at your changes and I see an important conceptual problem in your PR (sorry I missed it initially): you are reimplementing the eeprom_read/write in the STM32 common periph code. I'm wondering why it builds since this function is already defined in `drivers/periph_common/eeprom.c`.
You are also introducing internal `_eeprom_read_word` and `_eeprom_write_word` functions that are used to speed up global `eeprom_read/write` functions instead of the slow `eeprom_read/write_byte` one.
`word` is a name (weirdly) used by ST for 4 bytes values, on other architectures this is in general called dword (double word), and word is used for values on 2 bytes.

The actual API provides 4 functions:  eeprom_read, eeprom_write, eeprom_read_byte and eeprom_write_byte.

So I see 3 alternatives:
1. Keep the public API unchanged, implement a fast eeprom_read and eeprom_write and skip the ones in periph_common using defines only for STM32. AVR will provide their own slow (and EEPROM unsafe) implementations. This is basically what this PR does but it misses the defines to skip the implementation in periph_common. Even if it builds, it would make things cleaner.
The drawback of this method is that it duplicates more code (more functions defined in different places).
2. Add `eeprom_read_dword`, `eeprom_write_dword` function to the public API in drivers/periph_eeprom, and modify the public implementations of eeprom_read, eeprom_write like it's done in this PR. That would require the 2 new functions for AVR (but it can be a simple loop with a call to eeprom_read_byte, eeprom_write_byte. The drawback of this solution is that it will increase the code size of application built with periph_eeprom and on some really small AVR this can be a problem.
3. The third solution could be a mix of the 2 others: identify if an architecture provides eeprom_read/write_dword and use the optimized method in this case, otherwise use the slow/unsafe byte per byte read/write method. The implementation of eeprom_read/eeprom_write will be kept in `drivers/periph_eeprom` but with a conditional build using macro if dword fast functions are provided by an architecture. I would go for this solution but it requires a rework of your PR (and you won't have to touch AVR)

One last thing: I think the change about `_wait_for_pending_operations();` and `_IO` should be provided in another PR. This is quite orthogonal and could be merged fast.

olegart:
@aabadie I can't see why we need drivers/periph_common/eeprom.c at all.

MrKevinWeiss:
@olegart @aabadie is there any consensus?  I like solution 3 from @aabadie.

aabadie:
This PR is partially merged in master but the speedup is still missing. I'll take over this part.

PR #10032, opened 2018-09-25
pkg/openthread: use predefined env paths

smlng:
Adapt Makefiles to use already defined environment paths instead
of redefining them.

### Contribution description

This adapt Makefiles of the OpenThread package to use predefined path from RIOT build system, similar to all other packages.


jia200x:
works as expected

PR #10036, opened 2018-09-25
net/nanocoap: add module documentation

kb2ma:
### Contribution description

Adds high-level module documentation for nanocoap. This provides an outline for implementing a client or server, and pointers to the important API functions. It also clarifies how to use the minimal and struct-based APIs for options. The documentation is structured like the gcoap module documentation.

Also capitalized "Nanocoap" in the @defgroup doxygen tag, so it's in order in the Networking section of the documentation.
 
### Testing procedure

N/A, just documentation

### DIFF
--- a/sys/include/net/nanocoap.h
+++ b/sys/include/net/nanocoap.h
@@ -1,6 +1,7 @@
 /*
  * Copyright (C) 2016-17 Kaspar Schleiser <kaspar@schleiser.de>
  *               2018 Freie Universität Berlin
+ *               2018 Ken Bannister <kb2ma@runbox.com>
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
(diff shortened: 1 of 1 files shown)


sys/include/net/nanocoap.h
```diff
@@ -1,16 +1,99 @@
...
+ * which it responds. See the declarations of `coap_resources` and
+ * `coap_resources_numof`. The array contents must be ordered by the resource
+ * path, specifically the ASCII encoding of the path characters (digit and
+ * capital precede lower case). nanocoap provides the
+ * COAP_WELL_KNOWN_CORE_DEFAULT_HANDLER entry for `/.well-known/core`.
+ *
+ * ### Handler functions ###
+ *
+ * For each resource, you must implement a handler function. nanocoap provides
+ * functions to help implement the handler.
```
chrysn: A link to coap_handler_t might come in handy here.

sys/include/net/nanocoap.h
```diff
@@ -1,16 +1,99 @@
...
+ *
+ * If a response does not require specific CoAP options, use
+ * coap_reply_simple(). If there is a payload, it writes a Content-Format
+ * option with the provided value.
```
chrysn: I'd add
"Both coap_reply_simple and coap_build_reply build a response packet in the place of the received request package. The handler must extract all information relevant to it from the `pkt` handler argument into its stack before making `pkt` the response package using either of these functions."

sys/include/net/nanocoap.h
kb2ma: Addressed in 708c52c.

sys/include/net/nanocoap.h
kb2ma: Addressed in dcf346c. Used wording similar to gcoap. I would expect any server to reuse the request buffer, but I had to qualify that because it depends on how the server calls coap_handle_req().

kb2ma:
Thanks for the review! I am happy to hear this documentation clarifies how to use it. 

> Who is responsible for updating payload_len

The struct-based API option functions update these values as they are used. I added a sentence to that effect. Also, the coap_opt_add_xxx() function documentation notes this post condition.

sys/include/net/nanocoap.h
```diff
@@ -1,16 +1,102 @@
...
+ * - **minimal API** requires only a reference to the buffer for the message.
+ * However, the caller must provide the last option number written as well as
+ * the buffer position.
+ *
+ * - **struct-based API** uses a coap_pkt_t struct to conveniently track each
+ * option as it is written and prepare for any payload.
+ *
+ * For either API, the caller *must* write options in order by option number
+ * (see "CoAP option numbers", below).
+ *
```
haukepetersen: Should we add a note that it is normally not a good idea (or is it even possible) to mix both APIs when preparing a request/response packet?!

sys/include/net/nanocoap.h
```diff
@@ -1,16 +1,102 @@
...
+ * ### Minimal API ###
+ *
+ * Before starting, ensure the CoAP header has been initialized with
+ * coap_build_hdr(). For a response, coap_build_reply() includes a call to
+ * coap_build_hdr().
+ *
+ * Next, use the coap_opt_put_xxx() and coap_put_xxx() functions to write each
```
haukepetersen: I don't quite remember the current state of discussion, but did we already reach some kind of consensus to unify these functions to a common prefix like `coap_opt_put_xx`? Anyway, not really part of this PR...

sys/include/net/nanocoap.h
```diff
@@ -1,16 +1,102 @@
...
+ * struct.
+ *
+ * Next, use the coap_opt_add_xxx() functions to write each option, like
+ * coap_opt_add_uint(). When all options have been added, call
+ * coap_opt_finish().
+ *
+ * Finally, write any message payload at the coap_pkt_t `payload` pointer
+ * attribute. The `payload_len` attribute provides the available length in the
+ * buffer. The option functions keep these values current as they are used.
  *
```
haukepetersen: For completeness, I think adding another paragraph on how to read options from a packet might be useful to many users, right? But I would be completely fine by doing this in a follow up...

sys/include/net/nanocoap.h
```diff
@@ -1,16 +1,102 @@
...
  * General Public License v2.1. See the file LICENSE in the top level
  * directory for more details.
  */
 
 /**
- * @defgroup    net_nanocoap nanocoap small CoAP library
+ * @defgroup    net_nanocoap Nanocoap small CoAP library
  * @ingroup     net
- * @brief       Provides CoAP functionality optimized for minimal resource usage
+ * @brief       CoAP functionality optimized for minimal resource usage
```
haukepetersen: just a thought: I think `CoAP library [...]` might have a nicer sound to it.

haukepetersen:
Nice one, this should greatly improve the user experience when using nanocoap for many users that don't like to read code :-)

I have a couple of comments (see above), but I see none of them as blockers for this PR, hence I did not set my change request flag... So feel free to respond or just write 'no' to each of them :-)

sys/include/net/nanocoap.h
```diff
@@ -1,16 +1,102 @@
...
+ *               2018 Ken Bannister <kb2ma@runbox.com>
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
  * directory for more details.
  */
 
 /**
- * @defgroup    net_nanocoap nanocoap small CoAP library
+ * @defgroup    net_nanocoap Nanocoap small CoAP library
```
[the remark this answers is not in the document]
kb2ma: I made the 'N' upper case here only to support the case sensitive sort of the Doxygen tree view. Otherwise, nanocoap appears at the bottom of the list of Networking sub-modules. I'm not proposing to change case systematically. It just seemed worth it here to make the information easier to find. I don't see a way to change how Doxygen sorts the tree view.

Also, I'd prefer to leave the description as is for now. I might pick a different word than "minimal". :-)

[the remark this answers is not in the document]
kb2ma: I made the 'N' upper case here only to support the case sensitive sort of the Doxygen tree view. Otherwise, nanocoap appears at the bottom of the list of Networking sub-modules. I'm not proposing to change case systematically. It just seemed worth it here to make the information easier to find. I don't see a way to change how Doxygen sorts the tree view.

Also, I'd prefer to leave the description as is for now. I might pick a different word than "minimal". :-)

sys/include/net/nanocoap.h
kb2ma: I don't recall explicit discussion on use of a common prefix, but that is my preference for simplicity for the user. I expect we can deprecate our way there over time.

sys/include/net/nanocoap.h
kb2ma: I'll be direct. :-)

kb2ma:
Thanks for the review, @haukepetersen! Fixups coming soon.

kb2ma:
Just pushed two commits -- a fixup for @haukepetersen's comments, and a forehead-slapping fix to the minimal API description. I'm ready to squash if you are.

kb2ma:
Squashed. Also fixed the reference to "CoAP option numbers" which has been split out to a separate page.

> How about adding real code snippets ?

Thanks for reviewing, @aabadie. I referenced the nanocoap_server example for request handling and response generation. I agree that a simple example would be helpful for the client side. Considering that the docs have started from scratch, can we leave this work as an enhancement for now?

bergzand:
Anything still blocking this or can we merge this?

kb2ma:
I was hoping to hear back from @aabadie on deferring example code for now. He did not mark his comments as blocking though.

aabadie:
Sorry, I forgot to reply. I have no objections in improving the documentation in follow-up PRs.

Let's go then.

PR #9646, opened 2018-07-27
makefiles/docker.inc.mk: Use directories in RIOT when possible

cladmi:
### Contribution description

When building with docker, only mount the directories that are outside of RIOT.
This makes RIOT_FILE_RELATIVE be correctly relative to RIOTBASE when possible.

It also handles correctly `EXTERNAL_MODULE_DIRS`.

It fixes:

    -DRIOT_FILE_RELATIVE=\"/data/riotbuild/riotproject/tests/libfixmath_unittests/bin/pkg/stm32f0discovery/libfixmath/libfixmath/fix16_str.c\"

To

    -DRIOT_FILE_RELATIVE=\"tests/libfixmath_unittests/bin/pkg/stm32f0discovery/libfixmath/libfixmath/fix16_str.c\"

This fixes build size issues when `RIOT_FILE_RELATIVE` was used.

I also fixed `BUILDRELPATH` when building in `docker` for an example outside of RIOT to allow testing. (I could split this outside if wanted).

### Testing

Build `hello-world` with `docker` it should work and you can look at the list of `-v` and `-e` options given to docker.

Copy `hello-world` outside of `RIOT`, update `RIOTBASE` in the makefile to point to RIOT and build with docker. It should also work and you can notice that a new `-v` and `-e` options are given to docker.

I did not test with external cpu/board/make.

Now more advanced tests are described in these two comments:

* normal application, external applications and tests/external_modules: [#9646 (comment)](https://github.com/RIOT-OS/RIOT/pull/9646#issuecomment-433052095)
* External modules handling [#9646 (comment)](https://github.com/RIOT-OS/RIOT/pull/9646#issuecomment-436283027)

### Reviewing

I finished this a bit in a friday evening rush so would definitely need a second read.

### Issues/PRs references

Tracking issue https://github.com/RIOT-OS/RIOT/issues/9645
~https://github.com/RIOT-OS/RIOT/pull/9507~

PR #10038, opened 2018-09-25
make: introduce common BUILD_DIR

smlng:
### Contribution description

This introduces a new environment variable for a common directory
that holds all output of the build process, such as application or
package binaries. This would also allow to easily redirect output
to any other location, e.g. for out-of-source builds.

~~The 2nd commit adds a `build` directory with subdirs `build/bin` and `build/pkg`. In the (near) future we might add something like `build/tests` to store test results.~~

~~The 3rd commit adds the APPLICATION name as a subdirectory to the bin output folder, this is similar to the pkg subdirectory hierarchy and will also allow to build multiple application in one out-of-source build directory (in the future).~~

### Testing procedure

Everything should still build as usual. ~~However, as some directories changed (see above) it might break something~~

### Issues/PRs references

- needed by #10095 and #10147


aabadie:
> This would also allow to easily redirect output to any other location, e.g. for out-of-source builds.

I like. But let's wait for build system experts' opinion.

Makefile.include
```diff
@@ -15,9 +15,10 @@ RIOTTOOLS      ?= $(RIOTBASE)/dist/tools
 RIOTPROJECT    ?= $(shell git rev-parse --show-toplevel 2>/dev/null || pwd)
 GITCACHE       ?= $(RIOTTOOLS)/git/git-cache
 APPDIR         ?= $(CURDIR)
-BINDIRBASE     ?= $(APPDIR)/bin
-BINDIR         ?= $(BINDIRBASE)/$(BOARD)
-PKGDIRBASE     ?= $(BINDIRBASE)/pkg/$(BOARD)
+BUILDOUT_DIR   ?= $(APPDIR)/build
```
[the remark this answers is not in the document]
smlng: ah good point! Please, remember for all of us 😉

smlng:
> I like. But let's wait for build system experts' opinion.

that's the purpose of this PR, in parts I already talked about this with @cladmi - because we need a place to store test output+results, e.g. from HIL and so on, to they can be found and accessed more easily.

cladmi:
If we start change the name of the `bin` directory and touch this part of the BS API, I would also like to change one thing, remove the `override` to an absolute path of the directories inside that are board dependent. This is breaking things in the current build system with `buildtest` for example.

I would replace it by a check it is an absolute path if the user really wants to set it himself.

I would rather say, you can over

cladmi:
> > I was more thinking about a $(RIOTBASE)/build directory.
> > This would allow put things in common between builds, like tools, pythonlibs instead of currently cloning them directly in the source tree.
> 
> I think **waaaaaay back** `bin/` _was_ actually in `$(RIOTBASE)` for all applications... I think conflicts in configuration lead to the per application approach we have nowadays.

That is why I want to still keep a per application `bin` directory (or `build` directory). But introducing a common `build` directory could help migrate parts to a common one.

I would like to be able to build all applications without cleaning, but I would like to not have 120 times the 190M of the u8g2 package.

cladmi:
To explain a bit more my thought.

I see two parts to this PR.

Rename `bin` to `build` with a different hierarchy inside but still as a "per application" basis.
This changes the directory API but could be the occasion of doing a cleanup of the `override` directive of variables that are inside.
And I could go for this with the proper announcements.
I could also go for keeping the "ugly" name and put "bin/results" for the tests even if they are not `binaries` to keep the current API if it is preferred.


The second part you propose, is preparing this directory to be common between applications, by having the per `APPLICATION` subdirectory for example. And I would rather not do this but address it the other way around.
Because until 100% can be done in a common directory, the per application `build` directory cannot be moved to a common one and if at least one maintainers is not 100% conscious and careful that it can become common, new problems could be introduced.
Even now some packages need to be per application when they are based on external build systems. With `ccn-lite` for example, the `cmake` command depends on `INCLUDES` so on the application.

I would prefer introduce a`common` build directory, at `$(RIOTBASE)/build` or any other name with at the beginning nothing inside, and then, migrate one thing at a time to it. And maybe there will always be a need for per application things in the application bin directory. Or things could be migrated but we do not have enough working power to do the whole migration.

I think we would benefit more by separating it as early as possible to be able to put directly common things in it.
We should also define the rules on how to use it, like not allow having multiple build processes using the directory in parallel could prevent to use a lock for every transactions.

smlng:
I was also thinking to include the build commit hash (`git rev-parse --short HEAD`) into the hierarchy at some place, to allow for storage of multiple builds for the app but over time. Could be (at least) useful for CI and testing to store many binaries/results over time for comparison - or just to compare master with some PR.

smlng:
anyway thanks for your comments @cladmi, I guess you're right and I also like the idea to have `$(RIOTBASE)/build` - my idea was to simply introduce new *name* and hierarchy without changing the current structure too much.

Also for packages I wasn't aware (sure) that some might might be build in an app depend way, so maybe it makes sense to checkout/clone/download the package source to `build/pkg/<pkg_name>/` and then actually compile in `build/pkg/<pkg_name>/<app_name>` so each packages source is only needed once - is that what you mean by referring to u8g2?

Makefile.include
```diff
@@ -14,10 +14,11 @@ RIOTPKG        ?= $(RIOTBASE)/pkg
 RIOTTOOLS      ?= $(RIOTBASE)/dist/tools
 RIOTPROJECT    ?= $(shell git rev-parse --show-toplevel 2>/dev/null || pwd)
 GITCACHE       ?= $(RIOTTOOLS)/git/git-cache
+BUILDOUT_DIR   ?= $(RIOTBASE)/build
 APPDIR         ?= $(CURDIR)
-BINDIRBASE     ?= $(APPDIR)/bin
-BINDIR         ?= $(BINDIRBASE)/$(BOARD)
-PKGDIRBASE     ?= $(BINDIRBASE)/pkg/$(BOARD)
+BINDIRBASE     ?= $(BUILDOUT_DIR)/bin
```
cladmi: As explained in the main comments, I think the ones currently in `bin` should not been moved for the moment.
It is an API change I would not merge in the last weeks of a release.

Adding a new directory that needs to start being handled for new commands is ok.
The internal organization of this directory will also need some thinking.

Also, there is still some work to make this correctly handled by `docker` if it is set outside of RIOT. This could help https://github.com/RIOT-OS/RIOT/pull/9646
No idea what the `scan-build` does.

cladmi:
I would even add more hierarchy, why building in the 'source' part of the `build` directory.
It is still building in the source directory. And maybe `app_name` is a directory in your repository.

Choosing these directory hierarchy should also be a task on itself and I would like to discuss and propose some rough idea of a plan for review by the rest of the community.
I would use the `bitbake` `build` directory as a reference for comparison.

cladmi:
@smlng That should be handled by the CI. If I do `git commit --amend` to change the commit name it changes the build directory. Also a `dirty` repository is different from a clean one.
Also we should not include `git` commands to the directories as the release should be buildable from the archive.

A user can just, on its own, have a different 'build' directory per branch.

smlng:
yeah, I guess you're right @cladmi, thx for talking me out of it 😉 

Still, what about what's actually here, is it okay?

cladmi:
It's in my review todo list :)

makefiles/scan-build.inc.mk
```diff
@@ -7,6 +7,7 @@ SCANBUILD_ENV_VARS := \
   BINDIR \
   BINDIRBASE \
   BOARD \
+  BUILDOUT_DIR \
```
cladmi: I do not know how to test the `scan-build` feature.

makefiles/vars.inc.mk
```diff
@@ -28,6 +28,7 @@ export RIOTPROJECT           # Top level git root of the project being built, or
 export RIOTMAKE              # Location of all supplemental Makefiles (such as this file)
 export BINDIRBASE            # This is the folder where the application should be built in. For each BOARD a different subfolder is used.
 export BINDIR                # This is the folder where the application should be built in.
+export BUILDOUT_DIR          # The base folder for output of the RIOT build process
```
cladmi: Some subjective feedback.

I disagree with the current documentation and then the name of the variable.
What means 'output' here ? (which leads to buildOUT_dir).

If this folder is now supposed to store other things than test results it should have a more generic documentation and name.
I see it more as a folder used for "common build things", also with a subdirectory for "build artifacts(results/outputs)".
Maybe simply "BUILDDIR" could do, with a documentation saying it is a "common" one to differentiate to `BINDIR*` that are `application` specific. But I am bad at naming…

smlng: IIRC the naming *buildout* came up during our f2f discussion a while ago at the HAW. But I'm fine with changing the name and documentation.

makefiles/scan-build.inc.mk
smlng: I added it here bc BINDIR and BINBASE and others are listed, too. Better remove, for now?

cladmi:
I have a branch that should add my requests, I cannot test now as I do not have docker here.
I will test tomorrow but still want to put it here to not loose it

https://github.com/cladmi/RIOT/pull/new/pr/common/build

If it works I will do a PR on your branch.

cladmi:
This could go in for the release I think if you have time to review my changes.

smlng:
@cladmi integrated your PR, rebased and squashed

smlng:
to recap after all those changes back and forth, this PR **only** introduces a new directory it does **not** change any existing directories or where things are build and so on. This is only a first step to prepare such things, but also to have a common directory to put test results, i.e. produces by robot framework see related PRs

cladmi:
Changes look good to me.

cladmi:
Let me just re-test quickly.

Question, do you have any plans/doc for the internal organization of the directory?
Or do we assume the organization is currently on demand and can be changed whenever we want?

Because I see many different usages for this one, like `output/artifacts`, `cache/download/git`, `compiled_tools`, `tmp/bin/shared` and could be a good idea to sort usages by type somehow.

cladmi:
Just re-running murdock with the last master.

smlng:
@cladmi: as of now my immediate use case is to store HIL test results of the RobotFramework and therefore I currently was thinking of the following structure: `/robot/$(BOARD)/$(APPLICATION)` see #10774. But certainly the future goal is to allow for out-of-source builds and to make the RIOT-OS source dir read-only (or at least leave it unchanged), thus it makes sense to use this folder for caching and downloading of external stuff, too.

cladmi:
@smlng at least I would prefer `output/robot`. By having a common `output` or `artifacts` directory, it could be a directory that gets saved by murdock builders. And if something needs to add other outputs (documentation ?), we do not need to modify murdock code for every directory.

cladmi:
I will open an issue/PR for discussing an organization inside.

cladmi:
I noticed an issue, I forgot to grep for `BUILD_DIR` when reviewing:

https://github.com/RIOT-OS/RIOT/blob/7226daf33e9c6f0f6284492b9ecb7ff0c8315d36/pkg/jerryscript/Makefile.jerryscript#L1

I will do a PR to rename it.

PR #10046, opened 2018-09-26
net/nanocoap: use correct type param for _bulid_hdr()

haukepetersen:
### Contribution description
While cleaning up #9926 I noticed the unclean usage of the `COAP_REQ` define. As `coap_build_hdr()`'s second parameter is used as the actual message type value (e.g. CON/NON/ACK/RST), the define used for this parameter should be `COAP_TYPE_x`. As it happens, the value of `COAP_REQ` is equal to `COAP_TYPE_CON`, so the existing code worked just fine. But syntactically, I think this should be fixed :-)

### Testing procedure
Simple build test aka Murdock, as simply one define was exchanged with another one while both have the same value.

### Issues/PRs references
Noticed this while working on #9926

### DIFF
--- a/sys/include/net/nanocoap_sock.h
+++ b/sys/include/net/nanocoap_sock.h
@@ -44,7 +44,7 @@ extern "C" {
 int nanocoap_server(sock_udp_ep_t *local, uint8_t *buf, size_t bufsize);
 
 /**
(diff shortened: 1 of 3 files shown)


kb2ma:
Nice work! I agree with the evaluation and fix.

PR #10049, opened 2018-09-26
pkg/semtech-loramac: RU864 band added

olegart:
### Contribution description

Adds RU864-870 LoRaWAN band backported from LoRaMAC-Node 4.4.2, with duty cycle changed to 100 % (it is allowed in upper part of Russian ISM band).

RU864-870 is officially described in LoRaWAN Regional Parameters 1.0.3a and 1.1.

### Testing procedure

Build with LORA_REGION = RU864 and test with any RU864-compatible LoRaWAN setup.

Was tested with Kerlink iFemtoCell, Tektelic Kona Pico, Vega BS-1 base stations, Actility ThingPark, brocaar LoraServer, gotthard Lorawan-server network servers.

### DIFF
--- a/pkg/semtech-loramac/patches/0003-add-ru864-region.patch
+++ b/pkg/semtech-loramac/patches/0003-add-ru864-region.patch
@@ -1,0 +1,1997 @@
+From b8dc64684fb4e39d29cbfcb9a23b6e7807d251a0 Mon Sep 17 00:00:00 2001
+From: Oleg Artamonov <oleg@unwds.com>
+Date: Mon, 6 Aug 2018 21:43:54 +0300
+Subject: [PATCH 2/2] Add RU864 LoRaWAN region
+
+---
+ src/mac/LoRaMac.h            |    4 +
+ src/mac/region/Region.c      |   75 +++
+ src/mac/region/Region.h      |   28 +
+ src/mac/region/RegionRU864.c | 1018 ++++++++++++++++++++++++++++++++++
+ src/mac/region/RegionRU864.h |  459 +++++++++++++++
+ 5 files changed, 1584 insertions(+)
+ create mode 100644 src/mac/region/RegionRU864.c
+ create mode 100644 src/mac/region/RegionRU864.h
+
+diff --git a/src/mac/LoRaMac.h b/src/mac/LoRaMac.h
+index 8df8f9e..0067419 100644
+--- a/src/mac/LoRaMac.h
++++ b/src/mac/LoRaMac.h
+@@ -1739,6 +1739,10 @@ typedef enum eLoRaMacRegion_t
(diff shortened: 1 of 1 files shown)


aabadie:
By the way, now that #8864 is merged, I wanted to start working on porting the 4.4.2-rc2 version. But maybe you also want to have a look ?

olegart:
@aabadie, it's mostly not even _code_, it's just _settings_, a bunch of defines, actually, so I don't see why anyone should wait for some kind of official release. Let's say it was not backported but implemented according to LoRaWAN Regional Parameters 1.0.3a by a member of [LPWAN Russia working group](https://iotas.ru/committee/lora-lpwan/) that defined those parameters for Russia and proposed them to Alliance (that's me, right), what's the difference?

Regarding 4.4.2, I'm kind of scared of the scale of changes they did there (API, secure element, etc) so I think I'll wait until they release it. RC2 still has some problems, and anyway, our network providers are all on LoRaWAN Specifications 1.0.2 for now.

aabadie:
> it's mostly not even code, it's just settings, a bunch of defines, actually

RegionRU864.c, which is 100% added by the patch is about 1k lines, so nearly 50% of the change. I don't think "mostly" is the right word.

> Regarding 4.4.2, I'm kind of scared of the scale of changes they did there (API, secure element, etc) so I think I'll wait until they release it

This is one of the problem with their release workflow, patched versions contains lots of refactoring (see 4.4.0 => 4.4.1 which contains a switch to CMake build system...). But I understand your concern, now I'm not sure anymore what is best strategy. @jia200x or @miri64 what's you opinion, is it ok to merge in a package code a patch of 2k lines ?

jia200x:
@aabadie @olegart maybe not the cleanest and elegant solution, but I think it won't add big nightmares for us. Updating the pkg has been a hell anyway in all cases, so I don't think the bottleneck will be in a 2k lines patch :)

If it gets merged into the upstream, we remove the patch. If there's a conflict updating to the package, then we could see how to proceed.

olegart:
@jia200x there will be no need in this patch in the upcoming LoRaMAC 4.4.2, but there are so many internal changes in 4.4.2 I don't think RIOT adapt it anytime soon (and it's not released yet anyway).

Also, we were specifically asked by our largest network provider (ER-Telecom) NOT to update MAC level because they are sticking to LoRaWAN 1.0.2 (which means LoRaMAC 4.4.1) and do not want to have any potential troubles with newer protocol versions.

So all the reasons given, I believe there will be a lot of people prefer to stay with 4.4.1 for a while, at least until something like mid-2019.

aabadie:
I triggered Murdock and it's green. So given your arguments, now I'm fine with merging this PR. I let @jia200x giving his ACK and merge.

PR #10051, opened 2018-09-26
cpu/cortexm_common: add function to check address for validity

olegart:
### Contribution description

Adds extremely useful function to check memory addresses for validity on Cortex-M3/M4/M7 MCUs.

Can be used to determine memory sizes, peripherals availability, etc. in runtime to build firmware that runs effectively on a differenent MCUs without recompiling.

### Testing procedure

Build and run tests/cpu_cortexm_address_check for your board, then type 'check address' in the shell, where address is a memory address in hex notation (with '0x' prefix). Check your MCUs' datasheet for memory map.

E.g. on STM32L152RET6:
> check 0x1FF02000
Address 0x1ff02000 is NOT valid. Accessing it will result in BusFault

> check 0x1FF01999
Address 0x1ff01999 is valid. Feel free to access it

(0x1FF01999 is an upper address of the System Memory section)

### Issues/PRs references

~Cortex-M0 not supported, it needs completely different approach.~

P.S. Will provide real-world usage example in the next PR.

### DIFF
--- a/cpu/cortexm_common/cortexm_init.c
+++ b/cpu/cortexm_common/cortexm_init.c
@@ -72,3 +72,50 @@ void cortexm_init(void)
     cortexm_init_isr_priorities();
     cortexm_init_misc();
 }
+
+bool cpu_check_address(volatile const char *address)
+{
+#if defined(CPU_ARCH_CORTEX_M3) || defined(CPU_ARCH_CORTEX_M4) || \
+    defined(CPU_ARCH_CORTEX_M4F) || defined(CPU_ARCH_CORTEX_M7)
+    static const uint32_t BFARVALID_MASK = (0x80 << SCB_CFSR_BUSFAULTSR_Pos);
+
+    bool is_valid = true;
+
+    /* Clear BFARVALID flag */
+    SCB->CFSR |= BFARVALID_MASK;
+
+    /* Ignore BusFault by enabling BFHFNMIGN and disabling interrupts */
+    uint32_t mask = __get_FAULTMASK();
+    __disable_fault_irq();
+    SCB->CCR |= SCB_CCR_BFHFNMIGN_Msk;
+
+    *address;
+    /* Check BFARVALID flag */
+    if ((SCB->CFSR & BFARVALID_MASK) != 0)
+    {
+        /* Bus Fault occured reading the address */
+        is_valid = false;
+    }
+
+    /* Reenable BusFault by clearing  BFHFNMIGN */
+    SCB->CCR &= ~SCB_CCR_BFHFNMIGN_Msk;
+    __set_FAULTMASK(mask);
+
+    return is_valid;
+#else
+    /* Cortex-M0 doesn't have BusFault so we need to catch HardFault */
+    (void)address;
+
+    /* R5 will be set to 0 by HardFault handler */
+    /* to indicate HardFault has occured */
+    register uint32_t result __asm("r5") = 1;
+
+    __asm__ volatile (
+        "ldr  r1, =0xDEADF00D   \n" /* set magic number     */
+        "ldr  r2, =0xCAFEBABE   \n" /* 2nd magic to be sure */
+        "ldrb r3, [r0]          \n" /* probe address        */
+    );
+
+    return result;
+#endif
+}
--- a/tests/cpu_cortexm_address_check/main.c
+++ b/tests/cpu_cortexm_address_check/main.c
@@ -1,0 +1,65 @@
+/*
+ * Copyright (C) 2018 Unwired Devices LLC
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     tests
+ * @{
+ *
+ * @file
+ * @brief       Test application for Cortex-M address checks
+ *
+ * @author      Oleg Artamonov <oleg@unwds.com>
+ *
+ * @}
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+
+#include "cpu.h"
+#include "shell.h"
+
+static int cmd_check(int argc, char **argv)
+{
+    if (argc < 2) {
+        printf("usage: %s <hex address>\n", argv[0]);
+        printf("\t example: %s 0x08080000\n", argv[0]);
+        return 1;
+    }
+
+    char *address;
+    uint32_t address_value = strtol(argv[1], NULL, 0);
+
+    address = (char *)address_value;
+
+    if (cpu_check_address(address))
+        printf("Address 0x%08" PRIx32 " is valid. Feel free to access it\n", address_value);
+    else
+        printf("Address 0x%08" PRIx32 " is NOT valid. Accessing it will result in BusFault\n", address_value);
+
+    return 0;
+}
+
+static const shell_command_t shell_commands[] = {
+    { "check", "Check address", cmd_check},
+    { NULL, NULL, NULL }
+};
+
+int main(void)
+{
+    puts("Address check test\n");
+    puts("Please refer to the README.md for more details\n");
+    puts("usage: check <hex address>");
+    puts("\texample: check 0x08080000");
+
+    /* run the shell */
+    char line_buf[SHELL_DEFAULT_BUFSIZE];
+    shell_run(shell_commands, line_buf, SHELL_DEFAULT_BUFSIZE);
+    return 0;
+}
--- a/cpu/cortexm_common/include/cpu.h
+++ b/cpu/cortexm_common/include/cpu.h
@@ -189,6 +189,17 @@ static inline void cortexm_isr_end(void)
     }
 }
 
+/**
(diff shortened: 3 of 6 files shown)


olegart:
See #10052 for example

PeterKietzmann:
@olegart would you address the codays issue please?

cpu/cortexm_common/cortexm_init.c
```diff
@@ -72,3 +72,40 @@ void cortexm_init(void)
...
+
+    *address;
+    /* Check BFARVALID flag */
+    if ((SCB->CFSR & BFARVALID_MASK) != 0)
+    {
+        /* Bus Fault occured reading the address */
+        is_valid = false;
+    }
+
+    __asm volatile ("cpsie f;");
```
jcarrano: What if the mask was disabled when calling this function. They will be enabled. Maybe it's better to save/restore the flag.

cpu/cortexm_common/cortexm_init.c
```diff
@@ -72,3 +72,40 @@ void cortexm_init(void)
...
+#if defined(CPU_ARCH_CORTEX_M3) || defined(CPU_ARCH_CORTEX_M4) || \
+    defined(CPU_ARCH_CORTEX_M4F) || defined(CPU_ARCH_CORTEX_M7)
+    bool is_valid = true;
+
+    /* Clear BFARVALID flag */
+    SCB->CFSR |= BFARVALID_MASK;
+
+    /* Ignore BusFault by enabling BFHFNMIGN */
+    SCB->CCR |= SCB_CCR_BFHFNMIGN_Msk;
+    __asm volatile ("cpsid f;");
```
jcarrano: AFAICT this does not inhibit interrupts, meaning the bus fault could possibly be caused by something else.

olegart:
@PeterKietzmann Codacy is in disagreement with GCC here.

cpu/cortexm_common/cortexm_init.c
olegart: "cpsid f" disables interrupts and all fault handlers, according to 3.10.2 of the Cortex-M3 Devices Generic User Guide.

cpu/cortexm_common/cortexm_init.c
olegart: I'm pretty sure people should not play with FAULTMASK without a very solid reason, so it should be enabled at all times except when doind some very specific things.

Also, not sure if it's a good idea to write FAULTMASK register directly - there's MSR/MRS instructions to do so, but I can't find if there are any undesired consequences compared to using CPS instructions. I've tried and MSR/MRS (__get_FAULTMASK / __set_FAULTMASK from CMSIS) work, but still not sure if it's ok to use it so.

Regular interrupts (controlled by irq_enable / irq_disable) won't change anyway, it's PRIMASK, not FAULTMASK.

P.S. Changed assember to inline CMSIS instructions.

jcarrano:
> @PeterKietzmann Codacy is in disagreement with GCC here.

In any case, using `PRIx32` for the scanf should get rid of the problem.

cpu/cortexm_common/cortexm_init.c
jcarrano: > I'm pretty sure people should not play with FAULTMASK without a very solid reason, so it should be enabled at all times except when doind some very specific things.

Agreed, but I wanted to prevent any surprising weird interaction.

> Also, not sure if it's a good idea to write FAULTMASK register directly

If you look at https://github.com/RIOT-OS/RIOT/blob/master/cpu/cortexm_common/irq_arch.c it is doing that (it's even using cmsis builtins). I'm not sure why it is doing what it is doing in `irq_enable` (I just opened a bug report -  #10076).

It may work to have a fault_disable/fault_restore.

cpu/cortexm_common/cortexm_init.c
```diff
@@ -72,3 +72,40 @@ void cortexm_init(void)
...
+{
+#if defined(CPU_ARCH_CORTEX_M3) || defined(CPU_ARCH_CORTEX_M4) || \
+    defined(CPU_ARCH_CORTEX_M4F) || defined(CPU_ARCH_CORTEX_M7)
+    bool is_valid = true;
+
+    /* Clear BFARVALID flag */
+    SCB->CFSR |= BFARVALID_MASK;
+
+    /* Ignore BusFault by enabling BFHFNMIGN */
+    SCB->CCR |= SCB_CCR_BFHFNMIGN_Msk;
```
jcarrano: Sorry for not raising this before. Why is BFHFNMIGN set before disabling interrupts?

cpu/cortexm_common/cortexm_init.c
olegart: Ok, as no one sees no harm either, I've changed CPS to direct access with MRS with FAULTMASK state preservation. Moved BFHFNMIGN after interrupts disables too.

P.S. Don't think we need separate fault_disable function in public API.

olegart:
> In any case, using `PRIx32` for the scanf should get rid of the problem.

Doesn't work in newlib-nano. It builds, but produces some garbage.

cpu/cortexm_common/cortexm_init.c
```diff
@@ -72,3 +72,41 @@ void cortexm_init(void)
...
+    SCB->CCR &= ~SCB_CCR_BFHFNMIGN_Msk;
+    __set_FAULTMASK(mask);
+
+    return is_valid;
+#else
+    /* Cortex-M0 doesn't have BusFault */
+    (void) address;
+
+    printf("[warning] %s: Cortex-M0 doesn't have BusFault\n", __func__);
+    return true;
```
kaspar030: Is there a use-case for having this compiled in on Cortex-M0? If not, I suggest guarding the whole function.

(returning true seems wrong)

kaspar030: Or wouldn't it be possible to catch an "expected" hardfault and that way somehow replicate the functionality on M0?

olegart:
Added address probing for Cortex-M0 cores too.

M0 doesn't have BusFault, so we need to catch intended HardFaults by looking for some magic numbers in registers, than set some other register to indicated HardFault has occured and return from HardFault to the next instruction.

Here R1 and R2 are set with magic numbers (two different numbers just to be sure) and R5 is modified inside HardFault.

Tested with STM32F042K6T6:
> check 0x1FF02000
Address 0x1ff02000 is NOT valid. Accessing it will result in BusFault
> 
> check 0x40007999
Address 0x40007999 is valid. Feel free to access it

cpu/cortexm_common/cortexm_init.c
olegart: @kaspar030 Yes, just wrote a catcher for M0.

tests/cpu_cortexm_address_check/main.c
```diff
@@ -0,0 +1,67 @@
...
+    if (argc < 2) {
+        printf("usage: %s <hex address>\n", argv[0]);
+        printf("\t example: %s 0x08080000\n", argv[0]);
+        return 1;
+    }
+
+    char *address;
+    uint32_t address_value;
+
+    sscanf(argv[1] + 2, "%lx", &address_value); /* ignore 0x prefix */
```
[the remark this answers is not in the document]
jcarrano: And some code to check if the parse is successful.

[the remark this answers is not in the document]
jcarrano: It can be done, but just leave it, it's a test after all. The important thing is that it does not blow up and that is already fixed.

BTW, [here](https://linux.die.net/man/3/strtol) is how to do it:

> In particular, if *nptr is not '\0' but **endptr is '\0' on return, the entire string is valid.

jcarrano:
@olegart Squash, fix the whitespace issue and we're ready to go (I was waiting for @kaspar030 because I did not want to just dismiss his review.)

olegart:
Done.

> I wonder if there is some address that is invalid for most ARM chips

Addresses above 0xEFFFFFFF should be.

And 0xE0000000 should be valid on all Cortex-M as it's the beginning of a Cortex Private Peripheral Bus.

jcarrano:
Seems like clang does not like the assembly code.

Also, what happened to the test? Are you splitting it up?

olegart:
Yep, it seems GCC and CLANG have a bit different assembler understanding.

This should satisfy both.

jcarrano:
The test is failing because the CI is trying to build it for non-ARM boards. Unfortunately we don't have a way to blacklist based on architecture. Two options:

* Add all non-ARM boards to the blacklist.
* Add some M0 and some M3/M4 to the whitelist (adding the samr21xpro would be appreciated) . I think this is the option that makes sense.

fjmolinas:
@olegart @jcarrano Hi, I would like to use the PR to get low power for stm32l1, it seems this was ready to merge except for the white list part. @olegart do you think you can add the whitelist part and rebase it? if not, and if you don't mind I can use your branch and implement the changes needed to merge.

jcarrano:
@fjmolinas I'm not the author, but I think it should be fine. I already took over one of @olegart's PRs dealing with AES. He will get credited anyway. I'll review it.

PR #10052, opened 2018-09-26
cpu/stm32_common: minimize power consumption on STM32L1

olegart:
### Contribution description

Automatically configure all STM32L1 GPIOs as analog inputs.

GPIOs on STM32L1 MCUs are configured as digital inputs with Schmitt triggers, which means, if they are left floating, their power consumption is not 0 but few microamperes per GPIO due to random noise on high-impedance pin and constant random trigger switching.

To achieve minimum possible consumption, GPIOs must be reconfigured as analog inputs. It can be done automatically during MCU initialization.

Function needs cpu_check_address to check what GPIOs are available (it depends on MCU package and so can't be deduced from MCU model data available to RIOT).

Newer STM32 energy-efficient series (L0, L4) already have GPIO configured as AIN by default.

### Testing procedure

STM32L1 power consumption without this patch in STOP+RTC mode and GPIOs left floating may be as high as 100 uA. With the patch, it should be < 2 uA.

### Issues/PRs references

Depends on PR #10051

### DIFF
--- a/cpu/stm32_common/cpu_init.c
+++ b/cpu/stm32_common/cpu_init.c
@@ -56,10 +56,27 @@ void cpu_init(void)
     for (int i = 0; i < 12; i++) {
         GPIO_TypeDef *port;
         port = (GPIO_TypeDef *)(GPIOA_BASE + i*(GPIOB_BASE - GPIOA_BASE));
-        if (cpu_check_address((char *)port))
-            port->MODER = 0xffffffff;
-        else
+        if (cpu_check_address((char *)port)) {
+#if !defined (DISABLE_JTAG)
+            switch (i) {
+                /* preserve JTAG pins on PORTA and PORTB */
+                case 0:
+                    port->MODER = 0xABFFFFFF;
+                    break;
+                case 1:
+                    port->MODER = 0xFFFFFEBF;
+                    break;
+                default:
+                    port->MODER = 0xFFFFFFFF;
+                    break;
+            }
+#else
+            port->MODER = 0xFFFFFFFF;
+#endif
+        }
+        else {
             break;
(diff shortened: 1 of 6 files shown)


PeterKietzmann:
@MichelRottleuthner this is a good candidate for you to review, right :-)?

PeterKietzmann:
@olegart would you address the codacy issue please?

olegart:
> @olegart would you address the codacy issue please?

First one addressed, on the second issue (whether uint32_t is an unsigned int or a long unsigned int) Codacy is in disagreement with gcc-7.

cpu/stm32_common/cpu_init.c
```diff
@@ -49,6 +49,20 @@ void cpu_init(void)
     periph_clk_en(APB1, BIT_APB_PWREN);
     /* initialize the system clock as configured in the periph_conf.h */
     stmclk_init_sysclk();
+
+#if defined(CPU_FAM_STM32L1)
```
vincent-d: This approach is valid for the whole stm32 family AFAIK

olegart: It's valid, but not needed at least on L1 and L4 - they have GPIO configured as analog inputs by default (see default MODER value in the Reference Manual).

Not sure about F1 (not an energy-efficient series anyway), but probably newer F-series don't need it either.

olegart:
Added option (enabled by default) to preserve the state of JTAG pins - PA13, PA14, PA15, PB2 and PB3 configured as Alternate Function Mode.

Doesn't affect power consumption if those pins are left floated.

fjmolinas:
Closing since ##11832 was merged.

PR #10053, opened 2018-09-26
Feat/base64 const

pwillem:
### Contribution description

This merge request changes the api of the base64 functions:
- add `const` to input arguments
- use `void *` for the raw (byte) data (input for `base64_encode()`, output for `base64_decode()`)

### Testing procedure

`make BOARD=native -C tests/unittests tests-base64 test` still compiles and passes

### Issues/PRs references
None

### DIFF
--- a/sys/base64/base64.c
+++ b/sys/base64/base64.c
@@ -56,7 +56,7 @@ static char getsymbol(unsigned char code)
     return (char)BASE64_NOT_DEFINED;
 }
 
-int base64_encode(unsigned char *data_in, size_t data_in_size, \
+int base64_encode(const unsigned char *data_in, size_t data_in_size,
                   unsigned char *base64_out, size_t *base64_out_size)
 {
     size_t required_size = 4 * ((data_in_size + 2) / 3);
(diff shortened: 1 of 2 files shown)


sys/base64/base64.c
```diff
@@ -56,9 +56,10 @@ static char getsymbol(unsigned char code)
     return (char)BASE64_NOT_DEFINED;
 }
 
-int base64_encode(unsigned char *data_in, size_t data_in_size, \
+int base64_encode(const void *data_in, size_t data_in_size,
                   unsigned char *base64_out, size_t *base64_out_size)
 {
+    const unsigned char *in = data_in;
```
kaspar030: is this variable necesary?

kaspar030:
> @kaspar030 what's your assessment on this here?

+1!

sys/base64/base64.c
smlng: mhm, right I mean thats the point of using `void` right, to not hassle with casts.

pwillem: if I try to use the void. I get the following errors:
```
RIOT/sys/base64/base64.c: In function ‘base64_encode’:
RIOT/sys/base64/base64.c:89:28: error: pointer of type ‘void *’ used in arithmetic [-Werror=pointer-arith]
         tmpval = *(data_in + i);
                            ^
RIOT/sys/base64/base64.c:89:18: error: dereferencing ‘void’ pointer [-Werror]
         tmpval = *(data_in + i);
                  ^~~~~~~~~~~~~~
RIOT/sys/base64/base64.c:89:16: error: void value not ignored as it ought to be
         tmpval = *(data_in + i);
                ^
```

jcarrano: That's right, `void*` cannot be used in pointer arithmetic.

PR #10056, opened 2018-09-27
drivers/periph_eeprom: add clear and erase functions

aabadie:
### Contribution description

This PR adds 2 new functions, `clear` and `erase` functions to the periph_eeprom API. These 2 functions are here as helper since clearing a byte in EEPROM simply consists in writing a 0 at its position.

In the meantime, this PR provides a couple of commits that fix the maximum position asserts in read and write functions.

### Testing procedure

Build and flash `tests/periph_eeprom` on a STM32 L0,  L1 or AVR based board.

- For testing the EEPROM maximum bounds access (example with an EEPROM of 6144 bytes, nucleo-l073rz case), write and read the last byte:

```sh
> write 6143 Z  => one byte written
> read 6143 1   => one byte read, 'Z'
```

- For clear and erase functions:
```sh
> clear 6143 1 => one byte cleared
> read 6143 1 => byte read is empty
> write 1 A
> write 6143 Z
> erase  => this command can take a lot of time!
> read <any position> 1 => any byte read is empty
```

### Issues/PRs references

Removed from #10026 but could be useful.

### DIFF
--- a/drivers/include/periph/eeprom.h
+++ b/drivers/include/periph/eeprom.h
@@ -35,6 +35,13 @@ extern "C" {
 #endif
 
 /**
+ * @brief   Default value of the EEPROM clear byte
+ */
+#ifndef EEPROM_CLEAR_BYTE
+#define EEPROM_CLEAR_BYTE 0x00
+#endif
+
+/**
  * @brief   Read a byte at the given position in eeprom
  *
  * @param[in]  pos      position to read
@@ -78,6 +85,17 @@ void eeprom_write_byte(uint32_t pos, uint8_t data);
  * @return the number of bytes written
  */
 size_t eeprom_write(uint32_t pos, const uint8_t *data, size_t len);
+
+/**
+ * @brief   Set @p len bytes from the given position @p pos with value @p val
+ *
+ * @param[in] pos       start position in eeprom
+ * @param[in] val       the value to set
+ * @param[in] len       the number of bytes to set
+ *
+ * @return the number of bytes set
+ */
+size_t eeprom_set(uint32_t pos, uint8_t val, size_t len);
 
 /**
  * @brief   Clear @p len bytes from the given position @p pos
--- a/drivers/include/periph/eeprom.h
+++ b/drivers/include/periph/eeprom.h
@@ -79,6 +79,25 @@ void eeprom_write_byte(uint32_t pos, uint8_t data);
  */
 size_t eeprom_write(uint32_t pos, const uint8_t *data, size_t len);
 
+/**
+ * @brief   Clear @p len bytes from the given position @p pos
+ *
+ * Clearing a byte in EEPROM simply consists in setting it to 0
+ *
+ * @param[in] pos       start position in eeprom
+ * @param[in] len       the number of bytes to clear
+ *
+ * @return the number of bytes cleared
+ */
+size_t eeprom_clear(uint32_t pos, size_t len);
+
+/**
+ * @brief   Erase the whole EEPROM content
+ *
+ * @return the EEPROM_SIZE
+ */
+size_t eeprom_erase(void);
+
 #ifdef __cplusplus
 }
 #endif
--- a/cpu/atmega_common/include/periph_cpu_common.h
+++ b/cpu/atmega_common/include/periph_cpu_common.h
@@ -146,6 +146,11 @@ typedef struct {
 } pwm_conf_t;
 /** @} */
 
+/**
+ * @brief   EEPROM clear byte
+ */
+#define EEPROM_CLEAR_BYTE              (0xff)
+
 #ifdef __cplusplus
 }
 #endif
--- a/drivers/periph_common/eeprom.c
+++ b/drivers/periph_common/eeprom.c
@@ -40,15 +40,20 @@ void eeprom_write_byte(uint32_t pos, uint8_t byte)
     eeprom_write(pos, &byte, 1);
 }
 
-size_t eeprom_clear(uint32_t pos, size_t len)
+size_t eeprom_set(uint32_t pos, uint8_t val, size_t len)
 {
     assert(pos + len <= EEPROM_SIZE);
 
     for (size_t i = 0; i < len; i++) {
-        eeprom_write_byte(pos++, 0);
+        eeprom_write_byte(pos++, val);
     }
 
     return len;
+}
+
+size_t eeprom_clear(uint32_t pos, size_t len)
+{
+    return eeprom_set(pos, EEPROM_CLEAR_BYTE, len);
 }
 
 size_t eeprom_erase(void)
--- a/drivers/periph_common/eeprom.c
+++ b/drivers/periph_common/eeprom.c
@@ -40,4 +40,19 @@ void eeprom_write_byte(uint32_t pos, uint8_t byte)
     eeprom_write(pos, &byte, 1);
 }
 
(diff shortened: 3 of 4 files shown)


aabadie:
Added the API Change label, since it extends the existing API (but without modifying the existing functions)

jcarrano:
@aabadie This is like a `bzero` for eeprom. Could I suggest something more like `memset`?

aabadie:
> Could I suggest something more like memset?

I can add `eeprom_set(uint32_t pos, uint8_t val, size_t len)`.  `eeprom_clear` will then become a wrapper to it with `val = 0`. What do you think ?

aabadie:
@jcarrano, I added `eeprom_set`. Let me know what you think.

jcarrano:
@aabadie I added the bug tag because this is also fixing a bug.

jcarrano:
> I added eeprom_set. Let me know what you think.

I made the suggestion because I foresee that at some point someone will add a `set` function. I'm still not sure if having a separate clear function is that necessary (the C standard library does not have a memory clear, for example), esp. seeing how conscious is everybody here about binary size.

> write 6143 Z  => one byte written
> read 6143 1   => one byte read, 'Z'

What if the test did that automatically? (in other words, allow me to be laze, please).

> assert(pos + len < EEPROM_SIZE);

aabadie:
> seeing how conscious is everybody here about binary size.

I know, but it can be nice if one finds it there. And if it's not by the application, it won't be in the final archive (maybe I'm wrong?).

jcarrano:
> if it's not called by the application, it won't be in the final archive

True, but my point is that it leads to the kind of situation where one has to be aware that replacing `clear` by `set(0)` saves memory, and there is code doing one thing and code doing another.

> > There it died another bug.

The assert thing had a bug, which you fixed.

> >    What if the test did that automatically? (in other words, allow me to be laze, please).

> Yes, that's a good idea. And it's also part of #10026. We could even add later a Python test script to automatize all this.

for me it means that it can be tested (and approved) waaay faster.

drivers/include/periph/eeprom.h
```diff
@@ -73,6 +73,35 @@ void eeprom_write_byte(uint32_t pos, uint8_t data);
  */
 size_t eeprom_write(uint32_t pos, const uint8_t *data, size_t len);
 
+/**
+ * @brief   Set @p len bytes from the given position @p pos with value @p val
+ *
+ * @param[in] pos       start position in eeprom
+ * @param[in] len       the number of bytes to set
+ *
```
MrKevinWeiss: Should you add param[in] val to the doxygen param list as well?

jcarrano: Is `[in]` even necessary. Is it not the default?

aabadie: fixed

> Is [in] even necessary? Is it not the default?

Following [Python's philosophy](https://www.python.org/dev/peps/pep-0020/#id3), I prefer when it's explicit (rather than implicit)

olegart:
> the C standard library does not have a memory clear, for example

Because memory doesn't have to be cleared prior to writing. EEPROM does.

EEPROM writing time strongly depends on current EEPROM content, so if you need timings to be predictable, you have to clear EEPROM first.

Regarding eeprom_set() - I can't see how it would be useful except some very limited cases.

aabadie:
> Because memory doesn't have to be cleared prior to writing. EEPROM does.

To avoid confusion, this is done internally by the CPUs (STM32 specific): each time the program tries to write on a non null byte/word/dword, the memory interface performs a clear of the bytes And on STM32L1, there are fast write modes that assumes the bytes to write to are already cleared before writing (faster).

And regarding the "high-level" EEPROM api in RIOT, there's no need to call `eeprom_clear` before `eeprom_write` since it's the clear is handled automatically by the memory interface of the CPU, but it's slower.

olegart:
> And regarding the "high-level" EEPROM api in RIOT, there's no need to call `eeprom_clear` before `eeprom_write` since clearing the bytes is handled automatically by the memory interface of the CPU, but it's slower.

Yes, it's 2 times slower. So if you do care about timings (and remember, unless you have your code in RAM or a separate NVM bank, even interrupts won't be processed until EEPROM clear/write is finished), you have to clear EEPROM manually. And we are on milliseconds scale here (a single EEPROM operation takes ~7 ms)!

See [AN4808](https://www.st.com/content/ccc/resource/technical/document/application_note/group0/7c/6a/88/d1/27/04/46/2f/DM00260799/files/DM00260799.pdf/jcr:content/translations/en.DM00260799.pdf) for details.

So, eeprom_clear in public API is a must. eeprom_write_byte - I don't see why anyone needs it and what are benefits over eeprom_write(pos, data, 1).

aabadie:
> So, eeprom_clear in public API is a must. eeprom_write_byte - I don't see why anyone needs it and what are benefits over eeprom_write(pos, data, 1).

After some thought, the EEPROM implementation could be changed as follows (and this would help in #10026):
- instead of implementing eeprom_read/write_byte per architecture, implement `eeprom_read`, `eeprom_write`
- keep eeprom_clear, eeprom_read_byte and eeprom_write_byte in the common implementation. These function would be helper wrapper around `eeprom_read/write` functions.

This way it's possible to implement optimized versions for STM32s while keeping the API generic and without changing it. Would you be happy with that @olegart ?

olegart:
@aabadie Absolutely!

aabadie:
@jcarrano, @olegart, see #10093. I tested it on nucleo-l073rz and arduino-mega2560 and it works.

By the way, I realized that this PR (#10056) might be broken on Atmega: EEPROM bytes are cleared when 0xFF is written and here I was assuming 0x00 like on STM32s.

olegart:
@aabadie, I see no problem here - let's clear ATMega EEPROMs by writing 0xFF. It may be confusing for someone who decides to check manually if EEPROM is empty, but it's not a problem really.

MrKevinWeiss:
@aabadie Any updates to this?

MrKevinWeiss:
Sounds good!

MrKevinWeiss:
Sounds good, I will look/test later today!

MrKevinWeiss:
As is tradition since I merged the xtimer before @cladmi wanted it merged I will wait a respectable amount of time before merging to let any nay-sayers say.

MrKevinWeiss:
BTW @aabadie I will eventually want to move that test up to a python layer but still figuring out the specifics.

MrKevinWeiss:
almost forgot @aabadie please squash!

aabadie:
> I will eventually want to move that test up to a python layer but still figuring out the specifics.

I don't understand the purpose of this comment. The actual python test is consistent with others. Why do you want to change that ?

> please squash!

Well, I already kept the git history 'clean' (no fixup! commits), so I don't get either why you want me to squash...

MrKevinWeiss:
> I don't understand the purpose of this comment. The actual python test is consistent with others. Why do you want to change that ?
Don't worry about it now.  I was thinking to add test logic into the higher layer instead of test, this would reduce binary size (and maybe even combined with other tests to reduce flashing).  Either way it good for now.

> Well, I already kept the git history 'clean' (no fixup! commits), so I don't get either why you want me to squash...
True, I thought it was very verbose but each commit does make sense.

aabadie:
Thanks for merging @MrKevinWeiss !

PR #10060, opened 2018-09-27
static_tests: add build system checks

cladmi:
### Contribution description

Add a script to execute sanity checks on build system files.
It should prevent bad patterns to re-appear after being cleaned.

Currently adds a check for using the content of `FEATURES` instead of
`USEMODULE`.

~There is one case currently ignored as fixing it requires handing
CPU dependencies and the script could go in without it.~ Fixed by https://github.com/RIOT-OS/RIOT/pull/10153

#### Fixes

This PR also include fixes for the test that can be split in separate PRs.

I think mainly the `hifive1/fe310` changes should have a separate review as they are a bit specific.

### Testing procedure

The script can be run directly with:

    ./dist/tools/buildsystem_sanity_check/check.sh

To test on master, it requires to cherry-picking just the commit adding the script and running it.

### Issues/PRs references

It is part of issues found while working on https://github.com/RIOT-OS/RIOT/issues/9913
The `fatfs` issue was also found in https://github.com/RIOT-OS/RIOT/pull/9307

Depending on commits regarding `hifive1` ~https://github.com/RIOT-OS/RIOT/pull/10062~ and the script split in ~https://github.com/RIOT-OS/RIOT/pull/10179~

### DIFF
--- a/dist/tools/ci/build_and_test.sh
+++ b/dist/tools/ci/build_and_test.sh
@@ -93,6 +93,7 @@ then
         run ./dist/tools/coccinelle/check.sh
         run ./dist/tools/flake8/check.sh
         run ./dist/tools/headerguards/check.sh
+        run ./dist/tools/buildsystem_sanity_check/check.sh
         exit $RESULT
     fi
 
--- a/pkg/fatfs/Makefile.dep
+++ b/pkg/fatfs/Makefile.dep
@@ -2,11 +2,5 @@ USEMODULE += fatfs_diskio_mtd
 USEMODULE += auto_init_storage
 USEMODULE += mtd
 
-include $(RIOTBASE)/boards/$(BOARD)/Makefile.features
-
-#if periph_rtc is available use it. Otherwise use static timestamps
-ifneq (, $(filter periph_rtc, $(FEATURES_PROVIDED)))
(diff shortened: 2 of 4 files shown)


jcarrano:
Why is murdock complaining that `There is nothing to exclude from by :(exclude) patterns.`?

cladmi:
Hmm I should check on `ubuntu` as I developed it

cladmi:
In git `2.7.4` from ubuntu 16.04, the `pathspec` exclude functionnality needs to have a inclusive pattern before https://stackoverflow.com/questions/10423143/how-to-exclude-certain-directories-files-from-git-grep-search/30084612#30084612

I updated the script accordingly.

cladmi:
Murdock does not complain about it anymore. Just the waiting for the other PR dependencies to be squashed.

I can the PR and re-order the commits to have parts integrated earlier if wanted.

Like adding the script alone and only enabling it here when everything is fixed.
And also split the other modules fixes in different PRs.

What do you think ?

jcarrano:
I think it is a good idea to add the tool and enable it on separate commits.

cladmi:
Rebased since https://github.com/RIOT-OS/RIOT/pull/10179 was merged.

I can split the fixes out if you want.

PR #10064, opened 2018-09-27
board/mips-malta: include the cpu features instead of common cpu

cladmi:
### Contribution description

board/mips-malta: include the cpu features instead of common cpu

This change was missed when the cpu 'mips32r2_generic' was separated
from 'mips32r2_common'. The target file already includes 'mips32r2_common'.

### Testing procedure

The `FEATURES_PROVIDED` do not change between master and this PR for `mips-malta`:

    make --no-print-directory -C examples/hello-world/ info-debug-variable-FEATURES_PROVIDED BOARD=mips-malta
    periph_timer cpp periph_pm

### Issues/PRs references

Inconsistency found while working on https://github.com/RIOT-OS/RIOT/issues/9913

PR #10078, opened 2018-09-28
boards: always include cpu features

cladmi:
### Contribution description

This adds a sanity check that CPU/Makefile.features is always used and replace the optional '-include' from boards as CPU Makefile.features always exist (done with `sed`).

### Testing procedure

Verify that murdock correctly builds 128 times (== all boards) for `examples/hello-world`.

### Issues/PRs references

~Waiting for: https://github.com/RIOT-OS/RIOT/pull/10064 and https://github.com/RIOT-OS/RIOT/pull/10063~

Part of working on https://github.com/RIOT-OS/RIOT/issues/9913

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -246,6 +246,13 @@ include $(RIOTBOARD)/$(BOARD)/Makefile.include
 INCLUDES += -I$(RIOTCPU)/$(CPU)/include
 include $(RIOTCPU)/$(CPU)/Makefile.include
 
+# Sanity check
+# The check is only done after 'include $(RIOTBOARD)/$(BOARD)/Makefile.include'
+# because we need to have the 'CPU' variable defined
+ifeq (,$(filter $(RIOTCPU)/$(CPU)/Makefile.features,$(MAKEFILE_LIST)))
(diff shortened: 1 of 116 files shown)


cladmi:
@smlng Glad that you asked :) and you are right it should be but it is also an upcoming PR.

The problem is that currently `CPU` is defined by `$(BOARD)/Makefile.include` and so is defined after parsing `Makefile.features`.
It is in my plans to move `CPU` to `Makefile.features`, and other variables, to achieve this. And all the small PRs I am doing right now are here to be ready for this, and prevent new boards to break the migration.

Why not including it after `Makefile.include`, or anything else?
That is what I describe in "Dependencies processing order issues #9913" and ask for feedback on.
So I would rather ask you to read it than trying to copy-paste things here :)

cladmi:
But it's true I forgot to reference the main issue in this PR description. I fixed it.

smlng:
please squash

cladmi:
Thank you for reviewing.

I rebased and squashed. I also re-ran the `sed` command to check there were no new boards that were introduced.

PR #10082, opened 2018-09-30
drivers: add driver for L3GD20H 3-axis gyroscope

gschorcht:
### Contribution description

[UPDATE] The driver has been completely reworked.

This PR provides a driver for the ST L3Gxxxx 3-axis gyroscope sensor familiy. The driver supports the following sensors which have to be specified by using the respective pseudomodule.

Sensor Variant   | Pseudomodule
:---------------------|:-------------------------------------------------------
L3GD20H          | `l3gd20h`
L3GD20            | `l3gd20`
L3G4200D        | `l3g4200_ng`
A3G4250D        | `a3g4250d`
I3G4250D         | `i3g4250d`

By default, the driver has a minimum size and supports only the most basic functionalities:
- a fixed configuration defined by the default parameter set in `l3gxxxx_params.h`
- the polling of raw data or angular rates in millidegrees per second (mdps)
- the power-down and power-up of the sensor
- the use of the I2C or the SPI interface

This allows efficient initialization by simply writing a burst of control registers to the sensor. After initialization the sensor is operational. All other functionalities have to be explicitly enabled by pseudomodules:

Pseudomodule        | Functionality
:-------------------|:-------------------------------------------------------
`l3gxxxx_low_odr`   | Low output data rates (L3GD20H only)
`l3gxxxx_fifo`           | 32 level FIFO
`l3gxxxx_irq_data`   | Data ready and FIFO status interrupt (`INT2/DRDY`) handling
`l3gxxxx_irq_event` | Axes movement and Wake-up interrupt (`INT1`) handling
`l3gxxxx_sleep`       | Sleep and wakeup functions
`l3gxxxx_config`      | Functions for changing configurations at runtime

~This PR adds a driver for the ST L3GD20H 3-axis gyroscope. Although it is written for L3GD20H, it should also work with L3GD20 and L3G4200D.~

~The driver supports most of the features of the L3GD20H:~
~- all measurement modes and full scales~
~- 16 bit raw data and 32 bit converted/normalized data output with mdps resolution~
~- FIFO configuration and handling~
~- data-ready interrupts and FIFO interrupts ~
~- event detection configuration and interrupts~
~- filter handling~
~- I2C and SPI interface~
~- power-down and sleep mode handling~

~The driver follows the approach to split the functionality in different variants of the driver defined as pseudomodules:~
~- **l3gd20h** is the small base module which offers only very basic functions such as initializing the sensor and reading data through polling and the SAUL interface.~
~- **l3gd20h_full** module provides additional features like filtering, interrupts, event detection, and FIFO handling.~
~- **l3gd20h_i2c** module realizes the I2C interface of the L3GD20H device.~
~- **l3gd20h_spi** module realizes the SPI interface of the L3GD20H device.~

### Test

[UPDATE] The test application included in this PR should work. The variable `DRIVER` is used to define the sensor variant to be used. It is set to `l3gd20h` by default. It must also be set if the board definition already includes the sensor variant as a module.
```
DRIVER=l3gd20h BOARD=... make -C tests/driver_l3gxxxx flash term
```
The default configuration uses the I2C interface (module `l3gxxxx_i2c`) of the sensor and polls the data periodically every 200 ms. If the SPI interface of the sensor should be used, module `l3gxxxx_spi` has to be used:
```
DRIVER=l3gd20h USEMODULE=l3gxxxx_spi BOARD=... make -C tests/driver_l3gxxxx flash term
```
Override the SPI CS pin (`L3GXXXX_SPI_CS`) in case the default configuration does not match your hardware configuration.

The test application uses different functionalities of the driver when different pseudomodules are used.

- Module `l3gxxxx_irq_data`: Enable the data ready interrupt or FIFO status interrupts if module `l3gxxxx_fifo`
  is used at the same time.
    ```
    DRIVER=l3gd20h USEMODULE='l3gxxxx_irq_drdy l3qxxxx_fifo' \
    CFLAGS='-DL3GXXXX_INT2_PIN=GPIO_PIN\(0,5\) BOARD=... make -C tests/driver_l3gxxxx flash term
    ```
- Module `l3gxxxx_irq_event`: Enable event interrupt generation as configured in the `l3gxxxx_params.h`
    ```
    DRIVER=l3gd20h USEMODULE='l3gxxxx_irq_event' \
    CFLAGS='-DL3GXXXX_INT1_PIN=GPIO_PIN\(0,6\) BOARD=... make -C tests/driver_l3gxxxx flash term
    ```
- Module `l3gxxxx_config`: The configuration of the event interrupt generation is changed at runtime.

If not all of them are tested, at least the compilation should work if all possible pseudomodules are enabled.
```
DRIVER=l3gd20h \
USEMODULE='l3gxxxx_i2c l3gxxxx_spi l3gxxxx_fifo l3gxxxx_irq_data l3gxxxxx_irq_event l3gxxxx_config l3gxxxx_low_odr l3gxxxx_sleep' \
BOARD=... make -C tests/driver_l3gxxxx
```

Kconfig could be tested with
```
TEST_KCONFIG=1 DRIVER=l3gd20h \
BOARD=nucleo-f103rb make -C tests/driver_l3gxxxx menuconfig
```

~```tests/driver_l3gd20h``` provides a configurable application which allows to test most features.~

~The default configuration uses the I2C interface (module ```l3gd20h_i2c```) of the sensor and the base module ```l3gd20h``` to initialize the sensor and to fetch data periodically. With module ```l3gd20h_spi```, the SPI interface of the sensor can be used. Using module ```l3gd20h_full```, the LPF and HPF filters are configured and used.~

~With following defines different functions of the driver can be tested:~

~- USE_L3GD20H_INT_DATA_READY - Instead of polling the sensor, data-ready interrupt is used to read the data.~
~- USE_L3GD20H_INT_EVENT - Event detection is configured which triggers interrupts when angular rate of any axis exceeds a configured threshold.~
~- USE_L3GD20H_FIFO - Instead of using single data samples, the FIFO of the sensor is activated and used in conjunction with FIFO interrupts to read out all data sample with one burst access.~


gschorcht:
PR is waiting for review.

benpicco:
This chip can be found on the venerable [`stm32f429i-disc1`](https://www.st.com/en/evaluation-tools/32f429idiscovery.html) board.

Linux supports it too https://lkml.org/lkml/2020/5/12/147

drivers/l3gd20h/l3gd20h.c
```diff
@@ -0,0 +1,991 @@
...
+        EXEC_RET(l3gd20h_reg_read(dev, L3GD20H_REG_FIFO_SRC, &reg, 1), res);
+        return _get_reg_bit(reg, L3GD20H_FIFO_FFS);
+    }
+}
+
+/*
+ * scale factors for conversion of raw sensor data to degree for possible
+ * sensitivities according to mechanical characteristics in datasheet
+ */
+static const float L3GD20H_SCALES[3] = { (0.00875F), (0.0175F), (0.070F) };
```
benpicco, line 253:
```suggestion
static const uint16_t L3GD20H_SCALES[3] = { 35, 70, 280 };
```

multiplied everything by 4000

drivers/l3gd20h/l3gd20h.c
```diff
@@ -0,0 +1,991 @@
...
+
+    int res = L3GD20H_OK;
+
+    l3gd20h_raw_data_t raw;
+
+    EXEC_RET(l3gd20h_read_raw (dev, &raw), res);
+
+    data->x = ((double)raw.x * L3GD20H_SCALES[dev->params.scale]) * 1000;
+    data->y = ((double)raw.y * L3GD20H_SCALES[dev->params.scale]) * 1000;
+    data->z = ((double)raw.z * L3GD20H_SCALES[dev->params.scale]) * 1000;
```
benpicco, line 270:
```suggestion
    data->x = (raw.x * L3GD20H_SCALES[dev->params.scale]) >> 2;
    data->y = (raw.y * L3GD20H_SCALES[dev->params.scale]) >> 2;
    data->z = (raw.z * L3GD20H_SCALES[dev->params.scale]) >> 2;
```

drivers/l3gd20h/l3gd20h.c
```diff
@@ -0,0 +1,991 @@
...
+
+    if ((num = l3gd20h_read_raw_fifo(dev, raw)) < 0) {
+        DEBUG_DEV("reading raw data samples from FIFO failed", dev);
+        return num;
+    }
+
+    for (int i = 0; i < num; i++) {
+        data[i].x = ((double)raw[i].x * L3GD20H_SCALES[dev->params.scale]) * 1000;
+        data[i].y = ((double)raw[i].y * L3GD20H_SCALES[dev->params.scale]) * 1000;
+        data[i].z = ((double)raw[i].z * L3GD20H_SCALES[dev->params.scale]) * 1000;
```
benpicco, line 366:
```suggestion
        data[i].x = (raw[i].x * L3GD20H_SCALES[dev->params.scale]) >> 2;
        data[i].y = (raw[i].y * L3GD20H_SCALES[dev->params.scale]) >> 2;
        data[i].z = (raw[i].z * L3GD20H_SCALES[dev->params.scale]) >> 2;
```

aabadie:
This needs rebase.

It will take time to review this PR in-depth (diff is +3K lines). At first sight, I'm impressed by the level of documentation you put into this @gschorcht.

> it should also work with L3GD20 and L3G4200D.

There's already a driver available in RIOT for the l3gd4200d sensor but your version clearly overtakes it, since all possible modes are implemented (interrupt, fifo, i2c, spi, etc).
Maybe the implementation should be renamed to something more generic and pseudo-modules be used to include a driver for a specific module.
At least, it's possible to test the driver on iotlab-m3 because it provides a L3G4200D.

benpicco:
I have an `stm32f429i-disco` which has a I3G4250D - does this belong to the same family?

gschorcht:
> I have an `stm32f429i-disco` which has a I3G4250D - does this belong to the same family?

According to the datasheet it seems to be very similar. The registers are the same. After a first look, the only difference I found is that the L3GD20H supports low ODRs which are not supported by I3G4250D. However, the driver checks if the device is an L3GD20H before it tries to use the low ODRs. There are some other very small differences, but these are not handled by the driver.

So I would say, that it should be possible to use this driver for the I3G4250D by defining the I3G4250D identifier and the I3G4250D enumeration.

gschorcht:
> Maybe the implementation should be renamed to something more generic and pseudo-modules be used to include a driver for a specific module.

@aabadie There are only 3 ST gyro sensors still active. I3G4250 for industrial applications, A3G4250 for automotive and L3GD20H for consumer applications. All of them have same register structure with only a few differences and should work with the driver. A3G4250 supports only a full scale pf +/-245 dps while I3G4250 and L3GD20H support different full scales up to +/-2000 dps.

@benpicco I3G4250 and A3G4250 have same ID and I2C address as L3G4200D and should work out of the box with the driver as L3G4200D.

Finally, the question is what generic name we should use for the driver, x3gxxxx isn't an option. We could call it `stx3gxxxx` and the pseudomodules `l3gd20h`, `l3g4200d` (which collides with the existing driver module name), `i3g4250d` and `a3g4250d`.

gschorcht:
@aabadie @benpicco
I have completely reworked the driver. I have completely reworked the driver. Now the driver has a minimal size by default and supports only the most basic functionalities.
- a fixed configuration of the sensor by the default parameter set of `l3gd20h_params.h`
- the polling of raw data or angular rates in millidegrees per second (mdps)
- the power-down and power-up of the sensor
- the use of the I2C or SPI interface

This allows efficient initialization by simply writing a burst of control registers to the sensor. All other additional functionalities have to be explicitly enabled by pseudomodules:

Pseudomodule        | Functionality
:-------------------|:-------------------------------------------------------
`l3gd20h_low_odr`   | Low output data rates (L3GD20H only)
`l3gd20h_fifo`      | 32 level FIFO
`l3gd20h_irq_drdy`  | Data ready and FIFO interrupt (`INT2/DRDY`) handling
`l3gd20h_irq_event` | Axes movement and Wakeup interrupt (`INT1`) handling
`l3gd20h_sleep`     | Sleep and wakup functions
`l3gd20h_config`    | Functions for changing configurations at runtime

Would it be OK to force push the branch because almost everything has been changed which doesn't result in a diff that could be followed?

The PR is more than 3 years old and I would be really happy if we could make some progress before it becomes obsolete. It works also with the newest ST gyro sensors.

benpicco:
That's great news!
I'll try to review the driver sooner this time around. 

> Would it be OK to force push the branch because almost everything has been changed which doesn't result in a diff that could be followed?

Sure, we are not dogmatic about that :wink:

drivers/saul/init_devs/auto_init_l3gxxxx.c
```diff
@@ -0,0 +1,71 @@
...
+ * @brief   Define the number of saul info
+ */
+#define L3GD20H_INFO_NUM    ARRAY_SIZE(l3gd20h_saul_info)
+
+/**
+ * @brief   Reference the driver struct
+ */
+extern saul_driver_t l3gd20h_saul_driver;
+
+void auto_init_l3gd20h(void)
```
benpicco, line 50:
```suggestion
void auto_init_l3gxxxx(void)
```

drivers/saul/init_devs/auto_init_l3gxxxx.c
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+/*
+ * @ingroup     sys_auto_init_saul
+ * @brief       Auto initialization of L3GD20H 3-axis gyroscope sensor
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ */
+
+#ifdef MODULE_L3GD20H
```
benpicco, line 16:
```suggestion
#ifdef MODULE_L3GXXXX
```
same for the rest of the file

drivers/l3gxxxx/l3gxxxx.c
```diff
@@ -0,0 +1,965 @@
...
+    return _GET_REG_BIT(reg, L3GXXXX_ANY_DATA_READY);
+#endif
+}
+
+/*
+ * scale factors for conversion of raw sensor data to degree for possible
+ * sensitivities according to mechanical characteristics in datasheet
+ * multiplied by 4,000
+ */
+static const int L3GXXXX_SCALES[3] = { 35, 70, 280 };
```
benpicco, line 212:
```suggestion
static const uint16_t L3GXXXX_SCALES[3] = { 35, 70, 280 };
```

saves a few bytes

drivers/include/l3gxxxx.h
```diff
@@ -0,0 +1,1928 @@
...
+ * This function is available only if module `l3gxxxx_fifo` is used.
+ *
+ * @param[in]   dev     device descriptor of L3Gxxxx sensor
+ * @param[out]  raw     array of 32 raw data structures
+ *
+ * @return  number of data samples read on success, or negative error code,
+ *          see #l3gxxxx_error_codes_t
+ */
+int l3gxxxx_read_raw_fifo(const l3gxxxx_t *dev,
+                          l3gxxxx_raw_data_fifo_t raw);
```
benpicco, line 1637: Does it make sense to provide this as a public function?

tests/driver_l3gxxxx/main.c
```diff
@@ -0,0 +1,280 @@
...
+#if IS_USED(MODULE_L3GXXXX_IRQ_DRDY)
+    /* init INT2/DRDY signal pin and enable the interrupt */
+    gpio_init_int(l3gxxxx_params[0].int2_pin, GPIO_IN, GPIO_RISING,
+                  l3gxxxx_int2_isr, 0);
+#endif /* IS_USED(L3GXXXX_IRQ_DRDY) */
+
+#if IS_USED(MODULE_L3GXXXX_IRQ_EVENT)
+    /* init INT1 signal pin and enable the interrupt */
+    gpio_init_int(l3gxxxx_params[0].int1_pin, GPIO_IN, GPIO_RISING,
+                  l3gxxxx_int1_isr, 0);
```
benpicco, line 230: In [`lis2dh12`](https://github.com/RIOT-OS/RIOT/blob/master/drivers/lis2dh12/lis2dh12.c#L473) I provided a wrapper function for this that blocks until an interrupt is received.

This was to prevent users from doing SPI directly inside the ISR to get the data.

drivers/saul/init_devs/auto_init_l3gxxxx.c
gschorcht: Ups, just forgotten when renaming. Unfortunately, it is not enough to recompile it after renaming.

drivers/include/l3gxxxx.h
gschorcht: Hm, when I wrote the driver for the CCS811 gas sensor for another project some years ago, someone insisted on having a function to read the raw data from sensor, although calculating the physical data from raw sensor data is quite difficult for such sensors. Since that time I also expose the function to read the raw data from the sensor.

I can hide it if you think it is better.

tests/driver_l3gxxxx/main.c
gschorcht: This is a different approach. In `lis2dh12` implementation the running thread calls `lis2dh12_wait_event` and blocks until the event happens and the driver takes care that the application doesn't access I2C or SPI in interrupt context.

At least I tried to document this [here](https://github.com/RIOT-OS/RIOT/blob/25dc4c4370d45a777cb4a4a7334337046b58954f/drivers/include/l3gxxxx.h#L780). In the L3Gxxxx driver, the application must use a mutex or messages in the ISR to inform the application thread that an interrupt has occurred. I believe that most of our drivers do it this way.

I can change this if you think it is more safe if the application does not access the sensor directly in the interrupt context.

gschorcht:
@leandrolanzieri May I ask you to have a look on Kconfig related parts of this PR and if it is Ok how I realized it, especially the part where board definitions already set some configurations, e.g. [boards/stm32f429i-disc1/Kconfig](https://github.com/RIOT-OS/RIOT/blob/461f9b103e630d5a69162158980f2edd683eb350/boards/stm32f429i-disc1/Kconfig#L31)?

boards/stm32f429i-disc1/Makefile.dep
```diff
@@ -9,3 +10,8 @@ endif
 ifneq (,$(filter touch_dev,$(USEMODULE)))
   USEMODULE += stmpe811
 endif
+
+ifneq (,$(filter l3gxxxx,$(USEMODULE)))
+  USEMODULE += i3g4250d
```
gschorcht, line 15: @benpicco Here we have a problem with different revisions of the board: L3GD20 (Rev. C01, D01), I3G4250D (Rev. E02)

boards/stm32f3discovery/Makefile.dep
```diff
@@ -1,4 +1,10 @@
 ifneq (,$(filter saul_default,$(USEMODULE)))
   USEMODULE += saul_gpio
   USEMODULE += lsm303dlhc
+  USEMODULE += l3gxxxx
+endif
+
+ifneq (,$(filter l3gxxxx,$(USEMODULE)))
+  USEMODULE += i3g4250d
```
gschorcht, line 8: @benpicco Here we have a problem with different revisions of the board: L3GD20 (Rev. C01, D01), I3G4250D (Rev. E02)

boards/stm32f429i-disc1/Makefile.dep
benpicco: Can the sensor type be detected at run-time?

gschorcht: In the former implementation it was detected during initialization. At least L3GD20* and [IAL]3G4* have different chip IDs. When I optimized the code for size, I removed this chip detection and defined respective pseudomodules, since the capabilities differ slightly. With the pseudo modules, only the code for the capabilities that the sensor has is compiled in.

For the I2C interface we have additionally the problem that the I2C addresses differ.

gschorcht: BTW, the L3GD20 variant isn't active any longer. That's why newer revisions of the board only have I3G4250.

benpicco: Could we select multiple pseudo-modules and only do the detection for the pseudo-modules that are compiled in? If there is only a single one, that should not increase the code size.

gschorcht: [Here](https://github.com/RIOT-OS/RIOT/blob/461f9b103e630d5a69162158980f2edd683eb350/drivers/l3gxxxx/l3gxxxx.c#L137) we would have to add an additional check for L3GD20H if pseudomodule `i3g4250d` is together with pseudomodules `l3gd20h` and `l3gxxxx_low_odr` since the I3G4250D doesn't have the register `LOW_ORD` and doesn't support low data rates. The detection itself would add additional code.

drivers/l3gxxxx/Kconfig
```diff
@@ -0,0 +1,432 @@
...
+if MODULE_L3GXXXX_FIFO
+
+choice
+    bool "Select FIFO mode"
+    default L3GXXXX_FIFO_MODE_FIFO
+    help
+        The 32 level FIFO is used to collect a number of sensor
+        data samples before the data have to be fetched by the MCU.
+
+config L3GXXXX_FIFO_MODE_BYPASS
```
leandrolanzieri, line 258: Why is there an option to have the FIFO but not use it?

drivers/l3gxxxx/Kconfig
```diff
@@ -0,0 +1,432 @@
...
+    help
+        Threshold values are defined for raw data values. Their resolution
+        therefore depends on the selected full range:
+        254 dps -> LSB value ~7.5 mdps,
+        500 dps -> LSB value ~15.3 mdps,
+        2000 dps -> LSB value ~61 mdps
+
+config L3GXXXX_INT1_X_GT_THRESH
+    bool "Interrupt generation enabled for |X| > X threshold (X high event)"
+    default true
```
leandrolanzieri, line 358: here and bellow: Kconfig uses `y` and `n` for logic true and false

drivers/l3gxxxx/Kconfig
```diff
@@ -0,0 +1,432 @@
...
+    bool
+
+config HAS_L3G4200D_NG
+    bool
+
+config HAS_A3G4250D
+    bool
+
+config HAS_I3G4250D
+    bool
```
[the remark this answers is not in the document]
leandrolanzieri: Also: during migration the `HAS_` prefix is reserved to symbols that match the features provided also in `Makefile.features`, for now move please change these to `HAVE_` (see https://doc.riot-os.org/kconfig-in-riot.html#autotoc_md1742)

tests/driver_l3gxxxx/app.config.test
```diff
@@ -0,0 +1,2 @@
+# application configuration. This is only needed during migration.
+CONFIG_MODULE_L3GXXXX=y
```
leandrolanzieri, line 2:
```suggestion
CONFIG_MODULE_L3GXXXX=y
CONFIG_MODULE_L3GD20H=y
```

drivers/l3gxxxx/Kconfig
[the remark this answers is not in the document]
leandrolanzieri: Finally: maybe we can do something like done here? https://github.com/RIOT-OS/RIOT/blob/d52077207ebe5f68660416dc0ae728864b0b1e1a/drivers/lis2dh12/Kconfig#L8-L12

Then `MODULE_SAUL_DEFAULT` would activate the driver if the board has the device.

leandrolanzieri (commented):
@gschorcht it looks pretty good (and detailed) to me already! Here are some comments

tests/driver_l3gxxxx/main.c
gschorcht: To be honest, I really like the idea that the driver ensures that an application doesn't access I2C or SPI in interrupt context. However, this has the disadvantage that the thread using the sensor cannot be used for anything else, which in turn increases the number of threads required for an application.

The advantage of the approach where the ISR is defined by the application and not by the driver is that the same thread can be used for other tasks if the ISR simply sets a flag used by the thread to determine when an event has occurred. This allows the thread to do other things between checking the flag.

drivers/l3gxxxx/Kconfig
```diff
@@ -0,0 +1,432 @@
...
+
+choice
+    bool "Select sensor variant"
```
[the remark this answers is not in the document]
gschorcht: What do you recommend, just to remove the leading "Select"? I used select only for choices where the user has to select one.

drivers/l3gxxxx/Kconfig
```diff
@@ -0,0 +1,432 @@
...
+
+endchoice
+
+if MODULE_L3GD20H
+
+config MODULE_L3GXXXX_LOW_ODR
+    bool "Support of low output data rates"
+    depends on MODULE_L3GD20H
+    help
+        Low data rates are only supported on L3GXXXX.
```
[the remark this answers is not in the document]
gschorcht: Yes, this was changed by an auto-replacement when I renamed the driver from `l3gd20h` to the more common name `l3gxxxx`. I saw this too when I was removing the if ... endif for this configuration.

drivers/l3gxxxx/Kconfig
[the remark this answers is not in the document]
leandrolanzieri: > What do you recommend, just to remove the leading "Select"? I used select only for choices where the user has to select one.

Yes, I would drop the "Select", as it is (IMO) understood from how `choice`s are displayed.

drivers/l3gxxxx/Kconfig
gschorcht: Good question. On the one hand it's for completeness to be compatible with the enumeration as defined in the datasheet. On the other hand, an application could change the FIFO mode during runtime, for example configure `BYPASS` mode first by default and switch to `STREAM` mode only after a configured event occured. The L3GD20H can do this automatically (`BYPASS_TO_STREAM`), for other variants of the sensor this has to be done by the application.

tests/driver_l3gxxxx/app.config.test
gschorcht: Wouldn't this configure the L3GD20H permanently as a sensor variant? What if a board has another sensor variant like [stm32f429i-disc1](https://github.com/RIOT-OS/RIOT/blob/461f9b103e630d5a69162158980f2edd683eb350/boards/stm32f429i-disc1/Kconfig#L33). The test application enables module `l3gd20h` by default if no sensor variant is defined.

gschorcht:
@leandrolanzieri Thanks for reviewing `Kconfig` related parts, it helps a lot.

gschorcht:
static tests fail with
```
boards/stm32f429i-disc1/include/board.h:52: warning: Member BTN0_PIN (macro definition) of file board.h is not documented.
boards/stm32f429i-disc1/include/board.h:60: warning: Member L3GXXXX_SPI_DEV (macro definition) of file board.h is not documented.
boards/stm32f429i-disc1/include/board.h:61: warning: Member L3GXXXX_SPI_CS (macro definition) of file board.h is not documented.
boards/stm32f429i-disc1/include/board.h:62: warning: Member L3GXXXX_INT1_PIN (macro definition) of file board.h is not documented.
boards/stm32f429i-disc1/include/board.h:63: warning: Member L3GXXXX_INT2_PIN (macro definition) of file board.h is not documented.
```
Why the hell do such simple and self-explanatory definitions need to be documented now?

boards/stm32f429i-disc1/Makefile.dep
gschorcht: While trying to change it to detect the sensor at runtime, I realized the following problems:
-  I2C addresses differ for the [variants](https://github.com/RIOT-OS/RIOT/blob/bd270e8e064b793d90a68d07ea288a738ba7048e/drivers/include/l3gxxxx.h#L991), so it is not clear which one should be used as default configuration if pseudomodules `l3gd20*` and `x3g42*` are used simultaneously.
- Full scaling is limited to 245 dps for the A3G4250D. This cannot be reliably checked since its chip ID is the same as for the I3G4250D and L3G4200D.
- `Kconfig` requires the selection of a sensor variant, which in turn defines some dependencies on other configurations like [low ODR availability](https://github.com/RIOT-OS/RIOT/blob/bd270e8e064b793d90a68d07ea288a738ba7048e/drivers/l3gxxxx/Kconfig#L102) or [possible full scales](https://github.com/RIOT-OS/RIOT/blob/bd270e8e064b793d90a68d07ea288a738ba7048e/drivers/l3gxxxx/Kconfig#L172).

Maybe, you have an idea how to deal with such differences. I don't know how A3G4250D behaves if the application tries to set a full scale of 2000 dps.

gschorcht: @benpicco I have pushed a commit that includes these changes for testing. If this approach works, I would split it up into separate fixup commits of the driver. What the approach can't solve is that `Kconfig` uses the variant and the corresponding dependencies that is found first.

tests/driver_l3gxxxx/app.config.test
leandrolanzieri: > Wouldn't this configure the L3GD20H permanently as a sensor variant?
Yes, it would.

> What if a board has another sensor variant like [stm32f429i-disc1](https://github.com/RIOT-OS/RIOT/blob/461f9b103e630d5a69162158980f2edd683eb350/boards/stm32f429i-disc1/Kconfig#L33). The test application enables module `l3gd20h` by default if no sensor variant is defined.

Ok, I missed this check in the `Makefile`. Would this match the 'automatic' variant selection in the driver Kconfig (based on the features)? In that case this is fine.

drivers/l3gxxxx/Kconfig
```diff
@@ -0,0 +1,456 @@
...
+        The cutoff frequency is specified as a value from 0 to 9, which defines
+        the cutoff frequency depending on the selected sensor mode and the
+        output data rate. Details can be found in the data sheet for the CTRL2
+        register field HPCF.
+
+endif # L3GXXXX_HPF_ONLY || L3GXXXX_HPF_AND_LPF2 || L3GXXXX_INT1_HPF_ONLY || L3GXXXX_INT1_HPF_AND_LPF2
+
+config MODULE_L3GXXXX_FIFO
+    bool "Support for the 32 level FIFO"
+    depends on MODULE_L3GXXXX
```
leandrolanzieri, line 259: As all this part is already inside `if MODULE_L3GXXXX` the `depends on` is redundant, right?

drivers/l3gxxxx/Kconfig
leandrolanzieri: Hmm I see. The on/off switch for the FIFO feature in the driver (`MODULE_L3GXXXX_FIFO`) makes totally sense, as it is a configuration of the driver itself. The choice of the initial mode is a bit more on the edge of being a hardware boot option IMO: it applies to a specific device instance, not to the driver itself. Still, we have been including these default values in Kconfig in many cases, so I'm fine with keeping this if it can be changed in runtime be the application.

tests/driver_l3gxxxx/main.c
gschorcht: Which approach is preferred in RIOT to handle interrupts and the GPIO pins for these interrupts in drivers? After looking at a number of drivers, I am really unsure and the following two questions arose:

1. Who should be responsible to initialize the GPIO pin that is used for interrupts, e.g. when new sensor data are available? The driver or the application?
2. Should the ISR be part of the driver that uses a mutex and a blocking function call by the application? Or should the ISR be part of the application that is registered with the driver and called back when the interrupt occurs?

I have seen both approaches for both questions. We are talking about sensor drivers where interrupts are not used for internal functions but to indicate the occurrence of an event to the application.

tests/driver_l3gxxxx/app.config.test
gschorcht: The [default selection of the variant](https://github.com/RIOT-OS/RIOT/blob/8e7ea65fbf3310d17142153a49e7d00e6ca25f51/drivers/l3gxxxx/Kconfig#L53) in `Kconfig` depends on on the variant a board definition has. I guess the first config entry of the choice is used if the board doesn't define a variant which would be L3GD20H, right?

leandrolanzieri: Yes, if none of the `default`s could be applied the first valid option is the default instead.

drivers/l3gxxxx/Kconfig
gschorcht: right.

gschorcht:
@benpicco After you pointed out the implementation for LIS2DH12 which is very similar to L3GXXXX sensors, I took a closer look into PR #15871. The following questions arose:

- In `lis2dh12` driver, functions to configure all sensor parameters are defined and used in `init` function. Furthermore, functions for FIFO configuration and interrupt configuration are implemented by default and. The test application uses all these functions to configure the sensor.
   The initial implementation of `l3gd20h` was structured in same way. However, to reduce ROM/RAM usage, I changed this structure so that a basic version of the driver just uses a fixed configuration defined in the `params` file that is directly written into sensor registers. According to the [Writing a Device Driver in RIOT guide](https://doc.riot-os.org/driver-guide.html), a sensor should be operational after initialization and should minimize the ROM/RAM usage. Therefore, all additional features like FIFO, interrupt handling and sensor configuration are now compiled in only when additional pseudomodules are used.

   The question is now what the preferred approach is.

- In `lis2dh12` driver, a lot of functions are used like `_acquire`, `_read`, `_write`, `_release` which all can return with error code that isn't evaluated by the driver. It's clear using `if` blocks arround each called function increases the code size a lot. However the [Writing a Device Driver in RIOT guide](https://doc.riot-os.org/driver-guide.html) "Other functions should check the dev pointer is not void, and should also handle error return values from the lower layer peripheral driver implementations".

  Since I am currently reworking all older PRs, the question now is what the best approach is and what the best compromise between code size and fault tolerance is. One of my approaches was to use something like this
   ```
   #if IS_USED(MODULE_L3GXXXX_NO_ERRORS)
   #define EXEC(f) f
   #else
   #define EXEC(f) \
       do { \
           int _r; \
           if ((_r = f) != L3GXXXX_OK) { \
               DEBUG("[l3gxxxx] %s: error code %d\n", __func__, _r); \
               return _r; \
           } \
       } while (0)
   #endif
  ```
  where functions are called through a macro which either realizes the `if` block around the called function or just calls the function without the evaluation of the return code.

Recommendations are welcome.

boards/stm32f429i-disc1/Makefile.dep
gschorcht: @benpicco have you had time to take a look into this approach?

benpicco:
tbh with the lis2dh12 function extension I was happy when we eventually had the PR in a mergable state, I did a follow-up #16328 to clean up the API (and prevent user error by calling SPI functions in the ISR), but I didn't attempt to get this into a perfect state.

If your driver is more modular, that's all for the better, there just wasn't the time to also do that for lis2dh12.

As for the error checking, I don't think you need to add an extra case without it, it shouldn't add much overhead. 
But I guess there is a case for only checking errors if it makes sense, often an `assert()` will work too if it's a configuration error that can never work  (#15902)

gschorcht:
> I did a follow-up #16328 to clean up the API

I see.

With this PR you have also changed the enumerations from real register values to values that only keep the order of the register values, but then have to be shifted to the correct position and have to be masked.

As I also saw in this driver, the structure for each register is defined as a union. I like this approach very much, it is used extensively in the ESP32/ESP8266 platform definitions. Since I haven't seen it in any other driver, I haven't used this approach for sensor register definitions yet. What do you think is more efficient, masking and logical shifts or access via the union?

benpicco:
> What do you think is more efficient, masking and logical shifts or access via the union?

I would expect the compiler to create the same code either way, but I did not verify this.

gschorcht:
> I would expect the compiler to create the same code either way, but I did not verify this.

If the enumeration comprises complete register values, the fastest approach is probably to define enumeration values that correspond to the register values and to write them directly to the register without masking and shifting.

gschorcht:
> As for the error checking, I don't think you need to add an extra case without it, it shouldn't add much overhead.

@benpicco Another question about error handling. When a number of bytes need to be transferred, an `acquire` call is usually followed by a number of `read`/`write` calls before `release` is called. However, `read`/`write` over a communication channel can always fail. If we were to consistently check for errors, we would have to check for errors after each `read`/write` and would also have the code for calling `acquire` in each `if` block. 

This is probably the reason why most drivers don't use the return values of these functions and rely on that the communication was successful like in `lis2dh12`.

I often use a wrapper function in my drivers that calls `aquire`, `read`/`write` and `release` for each register read/write operation. I know that releasing the communication peripherals between single register read/write operations isn't optimal, but using such a function reduces the code size.

So I have these two options
- don't use return values
- use `release` after each single `read`/`write` in a wrapper function

What would you suggest?

benpicco:
- use `release` after each single read/write in a wrapper function

This can be pretty bad since the peripheral will be powered down on `release`. E.g. on the [sam0 platform](https://github.com/RIOT-OS/RIOT/blob/master/cpu/sam0_common/periph/spi.c#L378) a acquire - release cycle can be 1ms on slow family members.
I made that 'mistake' with the `at86rf215` driver and still hope to correct it when porting that to the new radio HAL.

Do you have to do error checks after each transfer? You can `|=` together return values and check in the end if any of them was not 0. 

But I wouldn't try to gracefully handle errors that are the result of a wrong configuration - just blow an assertion on those.
As for random bus errors (EMI? cosmic rays?) I'm not sure if there is a good way to deal with that anyway. If we can return an error that's great, but it's not important to know which instruction caused the error.

gschorcht:
> * use `release` after each single read/write in a wrapper function
> 
> This can be pretty bad since the peripheral will be powered down on `release`. E.g. on the [sam0 platform](https://github.com/RIOT-OS/RIOT/blob/master/cpu/sam0_common/periph/spi.c#L378) a acquire - release cycle can be 1ms on slow family members.

I see.

> Do you have to do error checks after each transfer? You can `|=` together return values and check in the end if any of them was not 0.

:+1: good idea.

> As for random bus errors (EMI? cosmic rays?) I'm not sure if there is a good way to deal with that anyway. If we can return an error that's great, but it's not important to know which instruction caused the error.

Yes, that is probably the point. I've always wondered what benefit it has for the higher layer to know which individual communication failed. It is only important that a communication has failed, especially since in general most communication errors are usually combined into only one error code.

Thanks, I will try to change my approach.

fjmolinas:
@gschorcht @benpicco should we target this one for next release? If so please add the milestone!

gschorcht:
> @gschorcht @benpicco should we target this one for next release? If so please add the milestone!

Let me try to improve the `_aquire`, `_read`/`_write`, `_release` sequence before review.

gschorcht:
> > @gschorcht @benpicco should we target this one for next release? If so please add the milestone!
> 
> Let me try to improve the `_aquire`, `_read`/`_write`, `_release` sequence before reviewing it.

I2C and SPI access changed.

tests/driver_l3gxxxx/main.c
gschorcht: @benpicco Finally, I decided to change the interrupt handling to the approach as used in `lis2dh12` driver. So I label the PR as WIP until I have finished it.

gschorcht:
@benpicco I have finished my changes for interrupt handling and sensor variant detection.

benpicco:
Murdock still has some issues

gschorcht:
> Murdock still has some issues

Ups, I didn't see this compilation problem with xtensa-gcc.

drivers/l3gxxxx/l3gxxxx.c
```diff
@@ -0,0 +1,1163 @@
...
+    if (_IS_DEV_I2C) {
+        DEBUG_DEV("I2C device initialized: bus %d, addr=%02x",
+                  dev, _I2C_DEV, _I2C_ADDR);
+    }
+#endif /* IS_USED(MODULE_L3GXXXX_I2C) */
+
+#if IS_USED(MODULE_L3GXXXX_SPI)
+    if (_IS_DEV_SPI) {
+        DEBUG_DEV("SPI device initialized: bus %d, cs=%u",
+                  dev, _SPI_DEV, _SPI_CS);
```
benpicco, line 193:
```suggestion
                  dev, (unsigned)_SPI_DEV, (unsigned)_SPI_CS);
```

drivers/l3gxxxx/l3gxxxx.c
```diff
@@ -0,0 +1,1163 @@
...
+#define ASSERT_PARAM(cond) assert(cond);
+#define DEBUG_DEV(m, d, ...)
+
+#endif /* ENABLE_DEBUG */
+
+#define _SET_REG_BIT(r, m, v) r = ((v) & ~m) ? r | m : r & ~m
+#define _GET_REG_BIT(r, m) (r & m) ? 1 : 0
+
+#define _SET_REG_VALUE(r, m, v) r = (r & ~m) | (((v) << m ## _S) & m)
+#define _GET_REG_VALUE(r, m) ((r & ~m) >> m ## _S)
```
benpicco, line 52: Ok now this is a bit too magic to be left without a comment :wink:

drivers/l3gxxxx/l3gxxxx.c
```diff
@@ -0,0 +1,1163 @@
...
+
+        uint8_t data[6];
+
+        /* read raw data sample */
+        res |= _read(dev, L3GXXXX_REG_OUT_X_L, data, 6);
+
+        /* L3GXXXX_REG_CTRL4.BLE = 0, Data LSB @ lower address */
+        raw[i].x = (data[1] << 8) | data[0];
+        raw[i].y = (data[3] << 8) | data[2];
+        raw[i].z = (data[5] << 8) | data[4];
```
benpicco, line 369: You could also do
```suggestion
        raw[i].x = byteorder_lebuftohs(&data[0]);
        raw[i].y = byteorder_lebuftohs(&data[2]);
        raw[i].z = byteorder_lebuftohs(&data[4]);
```
in case one of these mythical big endian machines shows up one day. But then again they are pretty much extinct for all practical purpose.

drivers/l3gxxxx/l3gxxxx.c
```diff
@@ -0,0 +1,1163 @@
...
+static int _is_available(l3gxxxx_t *dev)
+{
+    DEBUG_DEV("", dev);
+
+    int res;
+
+    uint8_t chip_id;
+
+    /* try to read the chip id from L3GXXXX_REG_WHO_AM_I */
+    res = _read(dev, L3GXXXX_REG_WHO_AM_I, &chip_id, 1);
```
benpicco, line 881: Is there some power-up delay? On `stm32f429i-disco` this would always read `0xFF` (or 0) when `ENABLE_DEBUG` is *not* set, but work when set, so looks like the initial debug print is delay enough.

What works is

```C
    uint8_t tries = 255;

    /* try to read the chip id from L3GXXXX_REG_WHO_AM_I */
    do {
        res = _read(dev, L3GXXXX_REG_WHO_AM_I, &chip_id, 1);
    } while ((chip_id == 0xFF || chip_id == 0x0) && --tries);
```

drivers/l3gxxxx/l3gxxxx.c
```diff
@@ -0,0 +1,1163 @@
...
+    if (dev->params.if_params.type == L3GXXXX_SPI) {
+        ASSERT_PARAM(len < L3GXXXX_SPI_BUF_SIZE)
+
+        reg &= 0x3f;
+        reg |= L3GXXXX_SPI_READ_FLAG;
+        reg |= L3GXXXX_SPI_AUTO_INC_FLAG;
+
+        /* prepare input and output by stream */
+        static uint8_t mosi[L3GXXXX_SPI_BUF_SIZE];
+        static uint8_t miso[L3GXXXX_SPI_BUF_SIZE];
```
benpicco, line 1027: Why are those static buffers needed here?

drivers/l3gxxxx/l3gxxxx.c
```diff
@@ -0,0 +1,1163 @@
...
+
+        reg &= 0x7f;
+
+        /* first byte in output is the register address */
+        mosi[0] = addr;
+
+        /* shift data one byte right, first byte in output is the register address */
+        for (uint8_t i = 0; i < len; i++) {
+            mosi[i + 1] = data[i];
+        }
```
benpicco, line 1110: Why can't you just do

```C
    spi_transfer_bytes(_SPI_DEV, _SPI_CS, true, &addr, 0, 1);
    spi_transfer_bytes(_SPI_DEV, _SPI_CS, false, data, 0, len);
```

benpicco (commented):
This looks pretty good!
But let's get rid of those `static` buffers.

With the ID-read retry this works on my `stm32f429i-disco`

```
2021-11-28 22:49:35,460 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:35,660 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:35,860 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:36,061 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:36,261 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:36,461 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:36,661 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:36,861 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:37,061 # gyro [dps] x:    +0, y:   +30, z:   -14
2021-11-28 22:49:37,262 # gyro [dps] x:   -20, y:   -81, z:   -16
2021-11-28 22:49:37,462 # gyro [dps] x:    -7, y:    +1, z:    +6
2021-11-28 22:49:37,662 # gyro [dps] x:    +0, y:   -62, z:    -7
2021-11-28 22:49:37,862 # gyro [dps] x:    -8, y:    +4, z:    -2
2021-11-28 22:49:38,062 # gyro [dps] x:    -2, y:    +1, z:    +8
2021-11-28 22:49:38,262 # gyro [dps] x:   +31, y:   +15, z:   +24
2021-11-28 22:49:38,463 # gyro [dps] x:    +7, y:   +14, z:    -1
2021-11-28 22:49:38,663 # gyro [dps] x:    +7, y:   +30, z:    +3
2021-11-28 22:49:38,863 # gyro [dps] x:    +0, y:    +5, z:    +0
2021-11-28 22:49:39,063 # gyro [dps] x:    +0, y:    -1, z:    +1
2021-11-28 22:49:39,263 # gyro [dps] x:    +1, y:    +5, z:    +0
2021-11-28 22:49:39,463 # gyro [dps] x:    +0, y:    +3, z:    +0
2021-11-28 22:49:39,664 # gyro [dps] x:    +0, y:    +2, z:    -2
2021-11-28 22:49:39,864 # gyro [dps] x:   +21, y:    +8, z:   +16
2021-11-28 22:49:40,064 # gyro [dps] x:    +7, y:   -76, z:   -11
2021-11-28 22:49:40,264 # gyro [dps] x:    -1, y:    -4, z:    -6
2021-11-28 22:49:40,464 # gyro [dps] x:   +11, y:   -18, z:    -5
2021-11-28 22:49:40,664 # gyro [dps] x:    +3, y:   +15, z:    +2
2021-11-28 22:49:40,865 # gyro [dps] x:    -3, y:   -14, z:    +0
2021-11-28 22:49:41,065 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:41,265 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:41,465 # gyro [dps] x:    +0, y:    +0, z:    +0
2021-11-28 22:49:41,665 # gyro [dps] x:    +0, y:    +0, z:    +0
```

benpicco:
let's not let this get stale again :wink:

This is almost good to go!

gschorcht:
> let's not let this get stale again wink
> 
> This is almost good to go!

Shit, I missed your comments for some reason. Now the `2022.01 branch` is created and for this PR probably too late for the next version :worried:

benpicco:
It will be in 2022.04 then ¯\\_(ツ)_/¯

I guess most people will just pull the current `master` anyway

drivers/l3gxxxx/l3gxxxx.c
gschorcht: Removed by splitting the transfer into
```
spi_transfer_byte(_SPI_DEV, _SPI_CS, true, reg);
spi_transfer_bytes(_SPI_DEV, _SPI_CS, false, 0, data, len);
```

drivers/l3gxxxx/l3gxxxx.c
gschorcht: The same here.

drivers/l3gxxxx/l3gxxxx.c
gschorcht: Hm, there is some power-up delay, but only for the gyros that are switched off by default after power-up because the sensor is in power-down mode. But the communication interfaces are of course always active to be able to communicate with the sensor also in power-down mode. There is no power-up delay is specified for the digital block in the datasheet.

In the initial version of the driver I had a delay of 100 ms at initialization, but this caused the dependency on `xtimer`. Now that we have `ztimer_msec`, we could add such a delay again, or we could try to read a valid ID several times before aborting.

gschorcht: > or we could try to read a valid ID several times before aborting.

Ah ok, I see. I have overlooked your suggestion with the retries. So I guess that you prefer this approach. But the `ztimer_msec` could also be usedul in `l3gxxxx_power_up` to wait for the gyros. At the moment, switching from power-down to normal mode results in some invalid measurements until the gyros are stable.

drivers/saul/init_devs/auto_init_l3gxxxx.c
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+/*
+ * @ingroup     sys_auto_init_saul
+ * @brief       Auto initialization of ST L3Gxxxx 3-axis gyroscope sensor
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ */
+
+#ifdef MODULE_L3GXXXX
```
aabadie, line 16: This is not needed anymore

drivers/l3gxxxx/l3gxxxx.c
benpicco: It's not wrong to use a timer, I just thought

```patch
diff --git a/drivers/l3gxxxx/l3gxxxx.c b/drivers/l3gxxxx/l3gxxxx.c
index bc5240d2d3..6caed8d47c 100644
--- a/drivers/l3gxxxx/l3gxxxx.c
+++ b/drivers/l3gxxxx/l3gxxxx.c
@@ -880,10 +880,12 @@ static int _is_available(l3gxxxx_t *dev)
 
     int res;
 
-    uint8_t chip_id;
+    uint8_t chip_id, tries = 255;
 
     /* try to read the chip id from L3GXXXX_REG_WHO_AM_I */
-    res = _read(dev, L3GXXXX_REG_WHO_AM_I, &chip_id, 1);
+    do {
+        res = _read(dev, L3GXXXX_REG_WHO_AM_I, &chip_id, 1);
+    } while ((chip_id == 0xFF || chip_id == 0x0) && --tries);
```

would be the most easy and robust solution.

boards/stm32f3discovery/Makefile.dep
benpicco: Then better leave a comment here, so the user knows what to do. (Or better yet, also add a note about this to `board.txt`)

boards/stm32f3discovery/doc.txt
```diff
@@ -107,6 +108,7 @@ An 3-axis gyroscope is soldered on the board.
 | INT1       | PE0 (IN)      |
 | INT2/DRDY  | PE1 (IN)      |
 
+[1] L3GD20 (up to rev. D01), IG34250D (from rev. E02)
```
benpicco, line 111: should have a note here that `Makefile.dep` needs to be adapted for this - or does the driver support run-time detection now?

benpicco:
Just needed a [simple patch](https://github.com/RIOT-OS/RIOT/pull/10082#discussion_r785495287) but then it's working fine on my `stm32f429i-disco`!

boards/stm32f429i-disc1/Makefile.dep
leandrolanzieri: What about adding a board configuration to determine which revision to use? Based on that the correct sensor variant is used. The configuration can be a pseudomodule, having the latest revision as the default one.

boards/stm32f3discovery/doc.txt
gschorcht: The driver detects the right sensor type for all sensor types that are enabled by pseudomodules. In case of the STM32F3Discovery board, `l3gd20` and `i3g4250d` are enabled so that the driver can detect them.

gschorcht: I added a short statement to the footnote.

gschorcht:
> Just needed a [simple patch](https://github.com/RIOT-OS/RIOT/pull/10082#discussion_r785495287) but then it's working fine on my `stm32f429i-disco`!

Patched as suggested.

benpicco:
Please squash!

benpicco (approved):

drivers/l3gxxxx/Kconfig
```diff
@@ -50,33 +50,39 @@ menuconfig MODULE_L3GXXXX
...
-    help
-        The driver supports L3GD20H, L3GD20, L3G4200D, I3G4250D and A3G4250D.
-        Select one.
+menu "Sensor variants used"
 
     config MODULE_L3GD20H
         bool "L3GD20H"
+        default y if HAVE_L3GD20H
+        help
+            Enables the support for L3GD20H.
```
benpicco, line 59:
```suggestion
            Enables the support for L3GD20H.
```

drivers/l3gxxxx/Kconfig
```diff
@@ -50,33 +50,39 @@ menuconfig MODULE_L3GXXXX
...
         bool "L3GD20H"
+        default y if HAVE_L3GD20H
+        help
+            Enables the support for L3GD20H. 
 
     config MODULE_L3GD20
         bool "L3GD20"
+        default y if HAVE_L3GD20
+        help
+            Enables the support for L3GD20.
```
benpicco, line 65:
```suggestion
            Enables the support for L3GD20.
```

drivers/l3gxxxx/Kconfig
```diff
@@ -50,33 +50,39 @@ menuconfig MODULE_L3GXXXX
...
         bool "L3GD20"
+        default y if HAVE_L3GD20
+        help
+            Enables the support for L3GD20. 
 
     config MODULE_L3G4200D_NG
         bool "L3G4200D"
+        default y if HAVE_L3G4200D_NG
+        help
+            Enables the support for L3G4200D.
```
benpicco, line 71:
```suggestion
            Enables the support for L3G4200D.
```

drivers/l3gxxxx/Kconfig
```diff
@@ -50,33 +50,39 @@ menuconfig MODULE_L3GXXXX
...
         bool "L3G4200D"
+        default y if HAVE_L3G4200D_NG
+        help
+            Enables the support for L3G4200D. 
 
     config MODULE_A3G4250D
         bool "A3G4250D"
+        default y if HAVE_A3G4250D
+        help
+            Enables the support for A3G4250D.
```
benpicco, line 77:
```suggestion
            Enables the support for A3G4250D.
```

drivers/l3gxxxx/Kconfig
```diff
@@ -50,33 +50,39 @@ menuconfig MODULE_L3GXXXX
...
         bool "A3G4250D"
+        default y if HAVE_A3G4250D
+        help
+            Enables the support for A3G4250D. 
 
     config MODULE_I3G4250D
         bool "I3G4250D"
+        default y if HAVE_I3G4250D
+        help
+            Enables the support for I3G4250D.
```
benpicco, line 83:
```suggestion
            Enables the support for I3G4250D.
```

benpicco (commented):
Please squash

benpicco:
CI does not like the latest Kconfig changes :weary:

gschorcht:
Hm, with the former version of `Kconfig` it wasn't happy for `stm32f429i-disc1` and `stm32f3discovery`. The reason was that `Makefile.dep` pulled in two sensor variants by pseudomodules while the `Kconfig` was a choice of a single one.

To solve it, I changed the choice to selections which are set by default if the board has the according sensor variants. But if a board doesn't define to have a variant, no variant is selected by `Kconfig` so that this error pops up.

I could define `CONFIG_MODULE_L3GD20H=y` in `app.config.test`, but this can't be overriden by the board definition or in `menuconfig`.

benpicco:
Maybe @leandrolanzieri has an idea

drivers/l3gxxxx/Kconfig
```diff
@@ -0,0 +1,457 @@
...
+
+if MODULE_L3GXXXX
+
+menu "Sensor variants used"
+
+    config MODULE_L3GD20H
+        bool "L3GD20H"
+        default y if HAVE_L3GD20H
+        help
+            Enables the support for L3GD20H.
```
leandrolanzieri, line 59: One option I see is to mimic what makefile is doing, defaulting to this one when no other is selected:
```suggestion
    config MODULE_L3GD20H
        bool "L3GD20H"
        default y if HAVE_L3GD20H
        default y if !MODULE_L3GD20 && !MODULE_L3G4200D_NG && !MODULE_A3G4250D && !MODULE_I3G4250D

        help
            Enables the support for L3GD20H.
```

leandrolanzieri:
Another thing I'm thinking: are we saving a lot of code by defining the model in a module at compile-time? Clearly we can detect the model at startup (like done for the `stm32f3discovery`), so would it make sense to do something like in https://github.com/RIOT-OS/RIOT/pull/17540 ?

gschorcht:
> Another thing I'm thinking: are we saving a lot of code by defining the model in a module at compile-time? Clearly we can detect the model at startup (like done for the `stm32f3discovery`), so would it make sense to do something like in #17540 ?

Saving code was not the primary goal. Rather, defining the type by a pseudomodule limits the features supported. Sure, we could compile the code for all features and check at runtime each time which feature is allowed with the actual variant present.

leandrolanzieri:
> Sure, we could compile the code for all features and check at runtime each time which feature is allowed with the actual variant present.

If we can determine the variant only from the chip ID it would be even easier than having the user specifying it in the params.

gschorcht:
> If we can determine the variant only from the chip ID

Unfortunately, this is not possible because several variants use the same ID.

The check for different variants at runtime requires that only variants with different IDs are enabled at the same time. And this should be controlled by enabling the pseudomodules or by Kconfig in future.

In the end, the recognition of the variant at runtime based on the ID was just a workaround for boards such as the `stm32f3discovery`, which can have one of two variants of the sensor depending on the revision, the L3GD20 and the I3G4250D. Since these variants have different IDs, the check for the variant at runtime works.

To be honest, I don't like the idea of changing the definition of variants by pseudomodules to user defined parameters. The PR is now more than 2 years old, during which I had to make extensive changes over and over again after a series of rebases. After I have completely reworked the driver again last October and the documentation with 1000 lines refers the pseudomodules for the variants extensively, I do not want to adapt the PR and completely revise the documentation again.

Therefore, I would like to continue with your first suggestion to select a default variant if no other one is selected.

OlegHahm:
@gschorcht, ping!

gschorcht:
Rebased due to conflicts.

aabadie:
[200802d](https://github.com/RIOT-OS/RIOT/pull/10082/commits/200802d32eb6825b774708231e3eff305d001094) needs squashing. Sorry

aabadie:
The Update commit was still there. Any reason for not squashing it ?

gschorcht:
> The Update commit was still there. Any reason for not squashing it ?

@aabadie I'm sorry, I definitely did a rebase  to squash it before I force-pushed it the last time. I still have the command in the history. I don't know what went wrong. I didn't realize that the rebase went wrong in any way. I thought it was fine and removed the "Needs Squashing" label. I was a little stressed this morning. My fault. I'm really sorry.

PR #10083, opened 2018-09-30
drivers: add driver for Honeywell HMC5883L magnetometer

gschorcht:
### Contribution description

This PR adds a driver for the Honeywell HMC5883L 3-axis digital magnometer as it is used for example on GY-85 9-DOF (Accelerometer ADXL345, Gyroscope ITG3200, Magnetometer HMC5883) module. Honeywell HMC5883L is a very easy to use, ultra-low-power magnetometer.

The driver implements
- basic polling mechanism
- power-down/up functions
- SAUL interface

### Testing procedure

This PR includes `tests/driver_hmc5883l` which can be used for testing. This test application demonstrates the usage of the driver and shows the different approaches to fetch data, by using either
- the `hmc5883l_read function` at a lower rate than the the DOR, or
- the data-ready interrupt (**DRDY**).

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -292,6 +292,14 @@ ifneq (,$(filter hih6130,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter hmc5883l,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_i2c
+endif
+
+ifneq (,$(filter hmc5883l_int,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_gpio_irq
+endif
+
 ifneq (,$(filter hts221,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
--- a/drivers/hmc5883l/Makefile
+++ b/drivers/hmc5883l/Makefile
@@ -1,0 +1,1 @@
+include $(RIOTBASE)/Makefile.base
--- a/drivers/hmc5883l/hmc5883l.c
+++ b/drivers/hmc5883l/hmc5883l.c
@@ -1,0 +1,301 @@
+/*
+ * Copyright (C) 2018 Gunar Schorcht
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_hmc5883l
+ * @brief       Device driver for the Honeywell HMC5883L 3-axis digital compass
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ * @{
+ */
+
+#include <string.h>
+#include <stdlib.h>
+
+#include "hmc5883l_regs.h"
+#include "hmc5883l.h"
+
+#include "log.h"
+#include "xtimer.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+#if ENABLE_DEBUG
+
+#define DEBUG_DEV(f, d, ...) \
+        DEBUG("[hmc5883l] %s i2c dev=%d addr=%02x: " f "\n", \
+              __func__, d->dev, HMC5883L_I2C_ADDRESS, ## __VA_ARGS__);
+
+#else /* ENABLE_DEBUG */
+
+#define DEBUG_DEV(f, d, ...)
+
+#endif /* ENABLE_DEBUG */
+
+#define ERROR_DEV(f, d, ...) \
+        LOG_ERROR("[hmc5883l] %s i2c dev=%d addr=%02x: " f "\n", \
+                  __func__, d->dev, HMC5883L_I2C_ADDRESS, ## __VA_ARGS__);
+
+#define EXEC_RET(f, r) \
+    if ((r = f) != HMC5883L_OK) { \
+        DEBUG("[hmc5883l] %s: error code %d\n", __func__, res); \
+        return res; \
+    }
+
+#define EXEC_RET_CODE(f, r, c) \
+    if ((r = f) != HMC5883L_OK) { \
+        DEBUG("[hmc5883l] %s: error code %d\n", __func__, res); \
+        return c; \
+    }
+
+/** Forward declaration of functions for internal use */
+
+static int _is_available(const hmc5883l_t *dev);
+
+static int _reg_read(const hmc5883l_t *dev, uint8_t reg, uint8_t *data, uint16_t len);
+static int _reg_write(const hmc5883l_t *dev, uint8_t reg, uint8_t data);
+
+int hmc5883l_init(hmc5883l_t *dev, const hmc5883l_params_t *params)
+{
+    int res = HMC5883L_OK;
+
+    assert(dev != NULL);
+    assert(params != NULL);
+    DEBUG_DEV("params=%p", dev, params);
+
+    /* init sensor data structure */
+    dev->dev = params->dev;
+#if MODULE_HMC5883L_INT
+    dev->int_pin = params->int_pin;
+#endif
+    dev->op_mode = params->op_mode;
+    dev->gain = params->gain;
+
+    /* check availability of the sensor */
+    EXEC_RET(_is_available(dev), res)
+
+    /* set configuration register A and B */
+    EXEC_RET(_reg_write(dev, HMC5883L_REG_CFG_A, params->meas_avg |
+                                                 params->meas_mode |
+                                                 params->dor), res);
+    EXEC_RET(_reg_write(dev, HMC5883L_REG_CFG_B, params->gain), res);
+
+    /* set operation mode */
+    EXEC_RET(_reg_write(dev, HMC5883L_REG_MODE, params->op_mode), res);
+
+    /* to set the LOCK flag, read the first data sample that is not valid */
+    uint8_t data[6];
+    EXEC_RET(_reg_read(dev, HMC5883L_REG_OUT_X_MSB, data, 6), res);
+
+    return res;
+}
+
+#ifdef MODULE_HMC5883L_INT
+
+int hmc5883l_init_int(hmc5883l_t *dev, hmc5883l_drdy_int_cb_t cb, void *arg)
+{
+    assert(dev != NULL);
+    assert(dev->int_pin != GPIO_UNDEF);
+    DEBUG_DEV("", dev);
+
+    if (gpio_init_int(dev->int_pin, GPIO_IN, GPIO_FALLING, cb, arg)) {
+        return HMC5883L_ERROR_COMMON;
+    }
+
+    return HMC5883L_OK;
+}
+
+#endif /* MODULE_HMC5883L_INT */
+
+int hmc5883l_data_ready(const hmc5883l_t *dev)
+{
+    assert(dev != NULL);
+    DEBUG_DEV("", dev);
+
+    int res = HMC5883L_OK;
+
+    uint8_t reg;
+
+    EXEC_RET(_reg_read(dev, HMC5883L_REG_STATUS, &reg, 1), res);
+    return (reg == HMC5883L_REG_STATUS_RDY) ? HMC5883L_OK : HMC5883L_ERROR_NO_DATA;
+}
+
+/*
+ * Scale factors for conversion of raw sensor data to uGs for possible
+ * sensitivities according to the datasheet.
+ */
+static const uint16_t HMC5883L_RES[] = {
+     730,      /* uG/LSb for HMC5883L_GAIN_1370 with range +-0.88 Gs */
+     917,      /* uG/LSb for HMC5883L_GAIN_1090 with range +-1.3 Gs */
+    1220,      /* uG/LSb for HMC5883L_GAIN_820  with range +-1.9 Gs */
+    1515,      /* uG/LSb for HMC5883L_GAIN_660  with range +-2.5 Gs */
+    2273,      /* uG/LSb for HMC5883L_GAIN_440  with range +-4.0 Gs */
+    2564,      /* uG/LSb for HMC5883L_GAIN_390  with range +-4.7 Gs */
+    3030,      /* uG/LSb for HMC5883L_GAIN_330  with range +-5.6 Gs */
+    4348,      /* uG/LSb for HMC5883L_GAIN_230  with range +-8.1 Gs */
+};
+
+int hmc5883l_read(const hmc5883l_t *dev, hmc5883l_data_t *data)
+
+{
+    assert(dev != NULL);
+    assert(data != NULL);
+    DEBUG_DEV("data=%p", dev, data);
+
+    int res = HMC5883L_OK;
+
+    hmc5883l_raw_data_t raw;
+
+    EXEC_RET(hmc5883l_read_raw (dev, &raw), res);
+
+    /*
+     * The range of raw data is -2048 ... -2047. That is, raw data multiplied
+     * by scale fit into 32 bit integer.
+     */
+    data->x = ((int32_t)raw.x * HMC5883L_RES[dev->gain >> HMC5883L_REG_CFG_B_GN_S]) / 1000;
+    data->y = ((int32_t)raw.y * HMC5883L_RES[dev->gain >> HMC5883L_REG_CFG_B_GN_S]) / 1000;
+    data->z = ((int32_t)raw.z * HMC5883L_RES[dev->gain >> HMC5883L_REG_CFG_B_GN_S]) / 1000;
+
+    return res;
+}
+
+int hmc5883l_read_raw(const hmc5883l_t *dev, hmc5883l_raw_data_t *raw)
+{
+    assert(dev != NULL);
+    assert(raw != NULL);
+    DEBUG_DEV("raw=%p", dev, raw);
+
+    int res = HMC5883L_OK;
+
+    uint8_t data[6];
+
+    /* read raw data sample */
+    EXEC_RET_CODE(_reg_read(dev, HMC5883L_REG_OUT_X_MSB, data, 6),
+                  res, HMC5883L_ERROR_RAW_DATA);
+
+    /* data MSB @ lower address */
+    raw->x = (data[0] << 8) | data[1];
+    raw->y = (data[4] << 8) | data[5];
+    raw->z = (data[2] << 8) | data[3];
+
+    return res;
+}
+
+int hmc5883l_power_down(hmc5883l_t *dev)
+{
+    assert(dev != NULL);
+    DEBUG_DEV("", dev);
+
+    /* set operation mode to Idle mode with only 5 uA current */
+    return _reg_write(dev, HMC5883L_REG_MODE, HMC5883L_OP_MODE_IDLE);
+}
+
+int hmc5883l_power_up(hmc5883l_t *dev)
+{
+    assert(dev != NULL);
+    DEBUG_DEV("", dev);
+
+    /* set operation mode to last operation mode */
+    return _reg_write(dev, HMC5883L_REG_MODE, dev->op_mode);
+}
+
+/** Functions for internal use only */
+
+/**
+ * @brief   Check the chip ID to test whether sensor is available
+ */
+static int _is_available(const hmc5883l_t *dev)
+{
+    DEBUG_DEV("", dev);
+
+    int res = HMC5883L_OK;
+
+    uint8_t id_c[] = HMC5883L_ID;
+    uint8_t id_r[HMC5883L_ID_LEN];
+
+    /* read the chip id from HMC5883L_REG_ID_X */
+    EXEC_RET(_reg_read(dev, HMC5883L_REG_ID_A, id_r, HMC5883L_ID_LEN), res);
+
+    if (memcmp(id_r, id_c, HMC5883L_ID_LEN)) {
+        DEBUG_DEV("sensor is not available, wrong id %02x%02x%02x, "
+                  "should be %02x%02x%02x",
+                  dev, id_r[0], id_r[1], id_r[2], id_c[0], id_c[1], id_c[2]);
+        return HMC5883L_ERROR_WRONG_ID;
+    }
+
+    return res;
+}
+
+static int _reg_read(const hmc5883l_t *dev, uint8_t reg, uint8_t *data, uint16_t len)
+{
+    assert(dev != NULL);
+    assert(data != NULL);
+    assert(len != 0);
+
+    DEBUG_DEV("read %d byte from sensor registers starting at addr 0x%02x",
+              dev, len, reg);
+
+    if (i2c_acquire(dev->dev)) {
+        DEBUG_DEV("could not acquire I2C bus", dev);
+        return HMC5883L_ERROR_I2C;
+    }
+
+    int res = i2c_read_regs(dev->dev, HMC5883L_I2C_ADDRESS, reg, data, len, 0);
+    i2c_release(dev->dev);
+
+    if (res == 0) {
+        if (ENABLE_DEBUG) {
+            printf("[hmc5883l] %s i2c dev=%d addr=%02x: read following bytes: ",
+                   __func__, dev->dev, HMC5883L_I2C_ADDRESS);
+            for (unsigned i = 0; i < len; i++) {
+                printf("%02x ", data[i]);
+            }
+            printf("\n");
+        }
+    }
+    else {
+        DEBUG_DEV("could not read %d bytes from sensor registers "
+                  "starting at addr %02x, reason %d (%s)",
+                  dev, len, reg, res, strerror(res * -1));
+        return HMC5883L_ERROR_I2C;
+    }
+
+    return res;
+}
+
+static int _reg_write(const hmc5883l_t *dev, uint8_t reg, uint8_t data)
+{
+    assert(dev != NULL);
+
+    DEBUG_DEV("write register 0x%02x", dev, reg);
+
+    if (ENABLE_DEBUG) {
+        printf("[hmc5883l] %s i2c dev=%d addr=%02x: write following bytes: ",
+               __func__, dev->dev, HMC5883L_I2C_ADDRESS);
+        printf("%02x ", data);
+        printf("\n");
+    }
+
+    if (i2c_acquire(dev->dev)) {
+        DEBUG_DEV("could not acquire I2C bus", dev);
+        return HMC5883L_ERROR_I2C;
+    }
+
+    int res = i2c_write_regs(dev->dev, HMC5883L_I2C_ADDRESS, reg, &data, 1, 0);
+    i2c_release(dev->dev);
+
+    if (res != 0) {
+        DEBUG_DEV("could not write to sensor registers "
+                  "starting at addr 0x%02x, reason %d (%s)",
+                  dev, reg, res, strerror(res * -1));
+        return HMC5883L_ERROR_I2C;
+    }
+
+    return res;
+}
--- a/drivers/hmc5883l/hmc5883l_saul.c
+++ b/drivers/hmc5883l/hmc5883l_saul.c
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2018 Gunar Schorcht
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_hmc5883l
+ * @brief       HMC5883L adaption to the RIOT actuator/sensor interface
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ */
+
+#include <string.h>
+
+#include "saul.h"
+#include "hmc5883l.h"
+
+static int read(const void *dev, phydat_t *res)
+{
+    hmc5883l_data_t data;
+    int ret = hmc5883l_read((const hmc5883l_t *)dev, &data);
+    if (ret < 0) {
+        return -ECANCELED;
+    }
+    res->val[0] = data.x;
+    res->val[1] = data.y;
+    res->val[2] = data.z;
+    res->unit = UNIT_GS;
+    res->scale = -3;
+    return 3;
+}
+
+const saul_driver_t hmc5883l_saul_driver = {
+    .read = read,
+    .write = saul_notsup,
+    .type = SAUL_SENSE_MAG,
+};
--- a/drivers/hmc5883l/include/hmc5883l_params.h
+++ b/drivers/hmc5883l/include/hmc5883l_params.h
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2018 Gunar Schorcht
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_hmc5883l
+ * @brief       Default configuration for the Honeywell HMC5883L 3-axis digital compass
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ * @{
+ */
+
+#ifndef HMC5883L_PARAMS_H
+#define HMC5883L_PARAMS_H
+
+#include "board.h"
+#include "hmc5883l.h"
+#include "saul_reg.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @name    Set default configuration parameters
+ * @{
+ */
+#ifndef HMC5883L_PARAM_DEV
+#define HMC5883L_PARAM_DEV          I2C_DEV(0)
+#endif
+#ifndef HMC5883L_PARAM_DOR
+#define HMC5883L_PARAM_DOR          (HMC5883L_DOR_15)
+#endif
+#ifndef HMC5883L_PARAM_MEAS_MODE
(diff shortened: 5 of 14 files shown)


haukepetersen:
I just ordered some HMC5883L devices to read the gas meter in my flat. Will give this PR a review and test as soon as they arrive (hopefully during the next 2 weeks).

gschorcht:
@haukepetersen Sounds good and it would be really great, but are we talking about the same thing? HMC5883L is a three axis compass. The use case for the use of a compass for reading gas meters is not clear to me.

haukepetersen:
Yes, we are talking about the same thing :-) Gas meters in Germany (at least a subset of them) have their least significant digit, this analog spin wheel, encoded with a magnet at one position. So when this digit wheel spins around, you can use the HMC5883L to detect the change in magenetic flux and with that detect a revolution of the digit wheel... So at least in theory, I will let you know if it works in practice for me :-)

-> see https://www.kompf.de/tech/gascountmag.html for more info

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,338 @@
...
+
+    uint8_t cfg_a = 0;
+    uint8_t cfg_b = 0;
+    uint8_t mode = 0;
+
+    /* set configuration register A and B */
+    _set_reg_bit(&cfg_a, HMC5883L_REG_CFG_A_MA, params->meas_avg);
+    _set_reg_bit(&cfg_a, HMC5883L_REG_CFG_A_MS, params->meas_mode);
+    _set_reg_bit(&cfg_a, HMC5883L_REG_CFG_A_DO, params->dor);
+    _set_reg_bit(&cfg_b, HMC5883L_REG_CFG_B_GN, params->gain);
```
haukepetersen: This seems much too complicated to me. I suggest to do something like
```c
uint8_t tmp = (params->meas->avg | params->meas_mode | params->dor);
_reg_write(dev->dev, dev->addr, HMC5883L_REG_CFG_A, tmp, 0); // of course test return value for error...
tmp = params->gain;
_reg_write(..., HMC5883L_REG_CFG_B, tmp, 0);
[...]
```
Now the only thing that is needed to be changed is to assing more specifig values to the config param enums, e.g.:
```c
typedef enum {
    HMC5883L_DOR_0_75 = 0x00,  /**< 0.75 Hz           */
    HMC5883L_DOR_1_5 = 0x04,       /**<  1.5 Hz           */
    HMC5883L_DOR_3 = 0x08,         /**<    3 Hz           */
    HMC5883L_DOR_7_5 = 0xc,       /**<  7.5 Hz           */
    HMC5883L_DOR_15 = 0x10,        /**<   15 Hz (default) */
    HMC5883L_DOR_30 = 0x14,        /**<   30 Hz           */
    HMC5883L_DOR_75 = 0x18,        /**<   75 Hz           */
} hmc5883l_dor_t;
```
This should be much easier to read and I would think it should also be more efficient?!

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,338 @@
...
+
+    /* wait 6 ms accoring to data sheet */
+    xtimer_usleep(6 * US_PER_MS);
+
+    return res;
+}
+
+int hmc5883l_data_ready(const hmc5883l_t *dev)
+{
+    ASSERT_PARAM(dev != NULL);
```
haukepetersen: personally I prefer a simple `assert(dev != NULL)`. IMO the added debug output message is only taking ROM for limited additional benefit, as the failed assertion on its own is already pretty obvious what has happened...

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,338 @@
...
+    int res = HMC5883L_OK;
+
+    uint8_t data[6];
+
+    /* read raw data sample */
+    EXEC_RET_CODE(_reg_read(dev, HMC5883L_REG_OUT_X_MSB, data, 6),
+                  res, -HMC5883L_ERROR_RAW_DATA);
+
+    /* data MSB @ lower address */
+    raw->x = (data[0] << 8) | data[1];
```
haukepetersen: We don't do it any better way in most of the drivers in RIOT, but maybe its a good point to start: using proper byteorder functions here might make this code also runnable on big-endian systems...

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,338 @@
...
+    ASSERT_PARAM(dev != NULL);
+    DEBUG_DEV("", dev);
+
+    uint8_t mode = 0;
+
+    /* set operation mode to Idle mode with only 5 uA current */
+    _set_reg_bit(&mode, HMC5883L_REG_MODE_HS, 0);
+    _set_reg_bit(&mode, HMC5883L_REG_MODE_MD, HMC5883L_OP_MODE_IDLE);
+
+    return _reg_write(dev, HMC5883L_REG_MODE, &mode, 1);
```
haukepetersen: This funtions could boil down to a one-liner (+ assertion):
```c
return _reg_write(dev->dev, dev->addr, HMC5883L_REG_MODE, HMC5883L_REG_MODE_MD);
```

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,338 @@
...
+
+    if (!data || !len) {
+        res = i2c_write_byte(dev->params.dev, dev->params.addr, reg, 0);
+    }
+    else {
+        res = i2c_write_regs(dev->params.dev, dev->params.addr, reg, data, len, 0);
+    }
+    i2c_release(dev->params.dev);
+
+    if (res != HMC5883L_OK) {
```
haukepetersen: should be `(res != 0)`, as `0` is the official `OK` value for the I2C driver...

gschorcht:
@haukepetersen Thanks for your first review.

gschorcht:
> * the `_set_reg_bit()` and `_get_reg_bit()` functions seem very strange to me. I don't think they are needed at all. See inline comments for a proposal to remove them.

Principially, I agree that is easy to OR individual bits using corresponding enum values. The same is true for reading. However, for multiple bit values, you would need to define a mask and a shift value for each register member, for example
```
tmp = (val << HMC5883L_REG_MODE_MD_SHIFT) & HMC5883L_REG_MODE_MD_MASK;
...
val = (tmp & HMC5883L_REG_MODE_MD_MASK) >> HMC5883L_REG_MODE_MD_SHIFT;
```
From my point of view, using these functions allows more readable source code, especially when many multi-bit register values ​​need to be set.
```
_set_bit(&tmp, HMC5883L_REG_MODE_MD, val);
...
_get_bit(val, HMC5883L_REG_MODE_MD, val);
```
Ok, these functions should be called `_set_bit` and `_get_bit` and they could be defined inlined or even as macros. BTW, PR #12238 will introduce such functions for common use.

> * I think I would move the handling for the `DRDY` pin inside of the driver: initialize that GPIO pin if the configuration sets it to something different than `GPIO_UNDEF`. Also we could build a convenience function for `read_blocking()` or similar, that blocks until a sample is ready. Maybe even put all of this into a pseudo-sub-module, so its only compiled in if needed/wanted.

Good idea, I will try it in that way. There are other drivers where I did it exactly in that way.

> * using these preprocessor functions/shortcuts would not be my personal style (and its not done too often in the RIOT code base), but this is merely subjective I guess :-)

Defining `DEBUG_DEV` and `ERROR_DEV` macros based on existing output macros ensures consistent output without writing down again and again exactly the same output prolog.

The same is true for `EXEC_RET`. But in addition, it allows to replace the four lines `if (...) { DEBUG(...); return error; }` statements by a single line. It's not a problem when only one register has to be written. But if dozends of registers have to be written, using the four lines `if ...` statements, the code would become unreadable. A good example is another driver I wrote
https://github.com/RIOT-OS/RIOT/blob/39c3406deaafcd16087662ad4c09a4fbae8bc730/drivers/vl53l1x/vl53l1x.c#L610-L655

I know that macros should be used wisely according to the [Guidelines for RIOT programming](https://github.com/RIOT-OS/RIOT/wiki/Best-Practice-for-RIOT-Programming). However, in this particular case, the code size is not affected. Even more, it helps to keep the source code readable.

drivers/hmc5883l/hmc5883l.c
gschorcht: OK, my heart does not hang on the additional error message. The developer can take a look into the source code when the assertion fails.

drivers/hmc5883l/hmc5883l.c
gschorcht: I'm not very clear, the endianess would be relevant only if `data` would be mapped directly the `int16_t` member. 

`data` contains the bytes in the order as they come from the sensor. This order is always little-endian. `(data[0] << 8) | data[1])` gives always a 16 bit integer which is assigned to the 16 bit integer `raw->x`. The order in which this integer is then stored in memory doesn't matter here. The code should work on little as well as on bigendian. Finally, the code is doing the same as 
`byteorder_bebuftohs` but for integer.

BTW, the byteorder functions in `core/include/bytorder.h` are only defined for unsigned integers which could lead to compiler errors for nitpicking compilers when used with integers.

drivers/hmc5883l/hmc5883l.c
gschorcht: Good point.

drivers/hmc5883l/hmc5883l.c
gschorcht: The question is whether we should use `assert` on an I2C error which can always happen from time to time. From my point of view, `assert` should only be used in case of configuration issues, but not for runtime errors.

drivers/hmc5883l/hmc5883l.c
haukepetersen: Exactly - and the developer has to do that in any case when debugging :-)

drivers/hmc5883l/hmc5883l.c
haukepetersen: yes, makes sense. I did not look into the semantics of the `raw->x/y/z` fields close enough, so never mind.

drivers/hmc5883l/hmc5883l.c
haukepetersen: No, I don't meant to assert the i2c errors, that would be bad practice, indeed! I was rather referring to assert `dev != NULL` and then calling the `_reg_write()` func afterwards. So all I2C error are evaluated inside the `_reg_write()` function during runtime, right?!

haukepetersen:
> Principially, I agree that is easy to OR individual bits using corresponding enum values. The same is true for reading. However, for multiple bit values, you would need to define a mask and a shift value for each register member, for example [...]

I don't see the need for shifting here. All bit positions are static, so it all boils down to OR and NOT AND operations for setting and clearing. For multi-bit values there is further no need to set the bits inidividually, they can be applied 'at-once' using their static values.

drivers/hmc5883l/hmc5883l.c
gschorcht: ORing fix values is definitely more efficient than to shift them bit-wise in a loop. Since they are only set during setup or when powering off/on, performance shouldn't be a problem.

But from my point of view, using fixed values instead of masks and shifts is more error-prone in programming. I had just to learn it when I tried to change these values it in this driver

gschorcht:
> * I think I would move the handling for the `DRDY` pin inside of the driver: initialize that GPIO pin if the configuration sets it to something different than `GPIO_UNDEF`. Also we could build a convenience function for `read_blocking()` or similar, that blocks until a sample is ready. Maybe even put all of this into a pseudo-sub-module, so its only compiled in if needed/wanted.

After thinking about the options, I would prefer an approach as it is realized in [ADS101X driver](https://doc.riot-os.org/group__drivers__ads101x.html). The driver defines a function by which the interrupt is enabled. The function is called by the application and has a callback function as parameters. Defining a blocking function would require to use any synchronization mechanism like message passing or thread flags which seems to be too complex for a low level driver.

The problem is that all of our sensor drivers only realize the polling approach so that we have no examples for best practice. But using the callback mechanism for sensor interrupts should be ok.

gschorcht:
@haukepetersen While thinking about the complexity the driver should have for interrupt handling, I am becoming unsure what the best approach is. This is probably a problem we will have with all drivers that require I2C access while handling interrupts.

The simplest approach from the driver's point of view would be for the driver to call back the application code directly from the ISR. However, this would be done in the interrupt context. That is, mutexes for synchronizing access to the I2C bus will not work and the application would have to take care of it.

Another approach would be to immediately return from the ISR and to call back the application later in the thread context. For this purpose, we introduced the "Single Interrupt Thread", which serializes interrupt events from various drivers that need to access I2C, for example. However, this approach costs an additional thread.

What do you think?

gschorcht:
@haukepetersen Fixed most of the issues. Please take a look at how the interrupt handling is now realized. I tested the driver again and it still seems to work. The remaining open point is the question of what to do with the macros. While you do not like macros, I like them a lot because they help avoid writing down dozens of copies of the same code over and over again.

haukepetersen:
The optional interrupt handling via callback function looks nice, just what I had mind :-)

I hope my devices arrive soon, can't wait to try it out!

haukepetersen:
I got my devices, but after some trial and error it turned out, that I did not receive actual `hmc5883l` devices, but I got `qmc5883l` ones. Unfortunately these have a different register map and use a different style of configuration, so I had to write a new driver from scratch (see #12464).

As I now won't be able to actually test this PR, I will try to a least mark some problems I see now after working with that other device for a night...

drivers/Makefile.dep
```diff
@@ -202,6 +202,12 @@ ifneq (,$(filter hih6130,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter hmc5883l,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_i2c
+  FEATURES_REQUIRED += periph_gpio_irq
```
haukepetersen: this feature is only required if the actual interrupt submodule is used, right?

drivers/hmc5883l/Makefile
```diff
@@ -0,0 +1,3 @@
+MODULE = hmc5883l
```
haukepetersen: no need to specify this line, it is deducted automatically by the make system

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,292 @@
...
+
+    /* check availability of the sensor */
+    EXEC_RET(_is_available(dev), res)
+
+    uint8_t cfg_a = 0;
+
+    /* set configuration register A and B */
+    cfg_a = params->meas_avg | params->meas_mode | params->dor;
+    EXEC_RET(_reg_write(dev, HMC5883L_REG_CFG_A, cfg_a), res);
+    EXEC_RET(_reg_write(dev, HMC5883L_REG_CFG_B, params->gain), res);
```
haukepetersen: There is quite some unneeded overhead introduced here by aquiring/releasing the i2c bus for every singe byte transferred. It would be more elegant to acquire once, do the needed transfers and release after...

drivers/hmc5883l/hmc5883l_saul.c
```diff
@@ -0,0 +1,40 @@
...
+    hmc5883l_data_t data;
+    int ret = hmc5883l_read((const hmc5883l_t *)dev, &data);
+    if (ret < 0) {
+        return -ECANCELED;
+    }
+    res->val[0] = data.x / 1000;
+    res->val[1] = data.y / 1000;
+    res->val[2] = data.z / 1000;
+    res->unit = UNIT_GS;
+    res->scale = -3;
```
haukepetersen: The doc below states that `hmc5883l_read()` returns the values in micro Gauss (`uG`). So this should be `-6`, right?!

drivers/hmc5883l/include/hmc5883l_params.h
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * @name    Set default configuration parameters
+ * @{
+ */
+#ifndef HMC5883L_PARAM_DEV
+#define HMC5883L_PARAM_DEV          I2C_DEV(0)
+#endif
+#ifndef HMC5883L_PARAM_ADDR
+#define HMC5883L_PARAM_ADDR         (HMC5883L_I2C_ADDRESS)
```
haukepetersen: The device address is static and can not be changed, so no need of including the addresss in the configuration data set.

drivers/include/hmc5883l.h
```diff
@@ -0,0 +1,315 @@
...
+#include <stdint.h>
+
+#include "periph/gpio.h"
+#include "periph/i2c.h"
+
+/** HMC5883L I2C address*/
+#define HMC5883L_I2C_ADDRESS    (0x1e)
+
+/** HMC5883L chip id defined in Identification Registers A..C */
+#define HMC5883L_ID             { 0x48, 0x34, 0x33 }
```
haukepetersen: IMO there is no need in having this as part of the public driver API. It could safely be moved into the `hmc5883l_regs.h` file...

drivers/include/hmc5883l.h
```diff
@@ -0,0 +1,315 @@
...
+    HMC5883L_ERROR_COMMON,         /**< common error */
+} hmc5883l_error_codes_t;
+
+/**
+ * @brief   Data output rates (DOR)
+ *
+ * Values correspond to bits <4:2> of HMC5883L_REG_CFG_A register
+ */
+typedef enum {
+
```
haukepetersen: I am not a fan of these newlines here and below, and they are not common in the RIOT code base. In any case, the formatting should at least be consistent inside this file :-)

drivers/include/hmc5883l.h
```diff
@@ -0,0 +1,315 @@
...
+/** HMC5883L chip id defined in Identification Registers A..C */
+#define HMC5883L_ID             { 0x48, 0x34, 0x33 }
+
+/** HMC5883L chip id length */
+#define HMC5883L_ID_LEN         (3)
+
+/** Definition of error codes */
+typedef enum {
+    HMC5883L_OK,                   /**< success */
+    HMC5883L_ERROR_I2C,            /**< any I2C communication error */
```
haukepetersen: to make the error checking less error prone, I think it is a good idea to assign them to negative values in the enum definition here (e.g. `HMC5883L_ERROR_I2C = -1` ...). Right now, the code like the following would yield unexpected behavior:
```c
int res = hmc5883l_init(...);
if (res == HMC5883L_ERROR_I2C) {   // fail: minus sign is missing
    // do some specific error handling
```
Assigning these enum values would further make this code consistent to other device drivers in RIOT.

drivers/include/hmc5883l.h
```diff
@@ -0,0 +1,315 @@
...
+    hmc5883l_gain_t      gain;      /**< Gain (default #HMC5883L_GAIN_1090) */
+
+} hmc5883l_params_t;
+
+/**
+ * @brief   HMC5883L sensor device data structure type
+ */
+typedef struct {
+
+    hmc5883l_params_t params;    /**< device initialization parameters */
```
haukepetersen: If I see it correctly, there is actually no need in copying the complete params struct into RAM. Some of the static configuration values are never used again after init() was called, and others can be compacted...

drivers/Makefile.dep
gschorcht: Yes, I just missed this dependence here when I introduced `hmc5883l_int`. Thanks.

drivers/hmc5883l/hmc5883l_saul.c
gschorcht: Don't remember, I have to check again the datasheet.

drivers/include/hmc5883l.h
gschorcht: OK

drivers/include/hmc5883l.h
gschorcht: Yeah, that's a problem. When I started to port some drivers I already used in other projects to RIOT, I was trying to figure out, what the right way in RIOT is to define error codes. I had to realize that all approaches are used:

1. negative codes that are returned directly (for example `drivers/bmx280`)
2. positive codes that are returned inverted (for example `drivers/bmp180`)
3. positive codes that are returned directly.

Even for same kind of drivers, they differ. Most of the drivers use 1. and 2.

Therefore, I decided to define and use them like standard error codes, i.e., positive codes returned inverted.

But, I can change it if it is really necessary. The documentation refers negative values.

drivers/include/hmc5883l.h
gschorcht: I will check.

drivers/hmc5883l/Makefile
gschorcht: A number of existing drivers are defining it in same way. So I thought that it is necessary.

drivers/hmc5883l/hmc5883l_saul.c
gschorcht: That should be correct. `hmc5883l_read` returns 32-bit values ​​in uGs. This divided by 1000 gives mGs. I decided to use 32-bit values ​​in `hmc5883l_read` not to lose accuracy when the raw data is multiplied by the gain.

drivers/hmc5883l/include/hmc5883l_params.h
gschorcht: OK

drivers/include/hmc5883l.h
gschorcht: OK

drivers/include/hmc5883l.h
gschorcht: It's probably a matter of taste. According to style guide this is not forbidden. Personally, I find it easier to read. But, I can change it.

drivers/include/hmc5883l.h
gschorcht: Yes, It is just the way as used in dozens of other drivers where the params are simply copied instead of picking single elements from them.

drivers/hmc5883l/hmc5883l.c
gschorcht: Hm, expanding `_reg_write` to `i2c_*` primtives would let explode the source code. Could you live with something like that
```
    uint8_t data[3] = { params->meas_avg | params->meas_mode | params->dor,
                        params->gain, 
                        params->op_mode };
    EXEC_RET(_reg_write(dev, HMC5883L_REG_CFG_A, data, 3), res);
```
that uses the fact that the registers are in subsequent order?

Regarding the code size, the former implementation required
```
   text	   data	    bss	    dec	    hex
  13148	    424	    953	  14525	   38bd
```
This implementation would require less memory.
```
   text	   data	    bss	    dec	    hex
  13130	    424	    953	  14507	   
```

gschorcht: The implementation of `_reg_write` and `_reg_read` with `acquire` and `release` as in #12464 requires much more memory:
```
  13202	    424	    953	  14579	   38f3
```

gschorcht: So, I would tend to leave my implementation as it is. The only place where the release in between subsequent register accesses happens is in `hmc5883_init`.

drivers/Makefile.dep
gschorcht: Fixed

drivers/hmc5883l/Makefile
gschorcht: Fixed

drivers/hmc5883l/include/hmc5883l_params.h
gschorcht: Fixed

drivers/include/hmc5883l.h
gschorcht: Fixed

drivers/include/hmc5883l.h
gschorcht: Fixed, only params that are required later are copied to the device structure.

drivers/include/hmc5883l.h
gschorcht: I changed it as proposed.

drivers/include/hmc5883l.h
gschorcht: Done

drivers/hmc5883l/hmc5883l_saul.c
gschorcht: I decided to change everything to milli-Gauss (mGs) since the precission is far away from 1 mGs.

gschorcht: I decided to change it back to uGs.

gschorcht:
@haukepetersen I added the computation of the heading angle in the test.

gschorcht:
@haukepetersen How do we continue with this PR?

benpicco:
@haukepetersen what's left here to do?

fjmolinas:
ping @haukepetersen

benpicco:
Btw.: Is there anything that needs to be done for this chip to show up?
I also have a GY-85 9-DOF module, only connected the I2C and power lines, yet `i2c_scan` only reports

```
2020-01-14 00:03:48,417 # addr not ack'ed = "-", addr ack'ed = "X", addr reserved = "R", error = "E"
2020-01-14 00:03:48,418 #      0 1 2 3 4 5 6 7 8 9 a b c d e f
2020-01-14 00:03:48,418 # 0x00 R R R R R R R R R R R R R R - -
2020-01-14 00:03:48,419 # 0x10 - - - - - - - - - - - - - - - -
2020-01-14 00:03:48,420 # 0x20 - - - - - - - - - - - - - - - -
2020-01-14 00:03:48,420 # 0x30 - - - - - - - - - - - - - - - -
2020-01-14 00:03:48,421 # 0x40 - - - - - - - - - - - - - - - -
2020-01-14 00:03:48,422 # 0x50 - - - X - - - - - - - - - - - -
2020-01-14 00:03:48,422 # 0x60 - - - - - - - - X - - - - - - -
2020-01-14 00:03:48,423 # 0x70 - - - - - - - - R R R R R R R R
```

gschorcht:
It might be a problem of your module. I get the following output where `0x1e` is the address of UMC5883L:
```
> i2c_scan 0
Scanning I2C device 0...
addr not ack'ed = "-", addr ack'ed = "X", addr reserved = "R", error = "E"
     0 1 2 3 4 5 6 7 8 9 a b c d e f
0x00 R R R R R R R R R R R R R R - -
0x10 - - - - - - - - - - - - - - X -
0x20 - - - - - - - - - - - - - - - -
0x30 - - - - - - - - - - - - - - - -
0x40 - - - - - - - - - - - - - - - -
0x50 - - - X - - - - - - - - - - - -
0x60 - - - - - - - - X - - - - - - -
0x70 - - - - - - - - R R R R R R R R

```

gschorcht:
I have just rebased to resolve conflicts. Furthermore, I have removed the dependency of this driver from `xtimer'.

gschorcht:
I have just rebased to resolve conflicts. Furthermore, I have removed the dependency of this driver from `xtimer'.

leandrolanzieri:
@haukepetersen were your comments addressed?

benpicco:
Looks like you've addresses all comments - please squash.

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,301 @@
...
+
+    EXEC_RET(_reg_read(dev, HMC5883L_REG_STATUS, &reg, 1), res);
+    return (reg == HMC5883L_REG_STATUS_RDY) ? HMC5883L_OK : HMC5883L_ERROR_NO_DATA;
+}
+
+/*
+ * Scale factors for conversion of raw sensor data to uGs for possible
+ * sensitivities according to the datasheet.
+ */
+static const int32_t HMC5883L_RES[] = {
```
benpicco:
```suggestion
static const int16_t HMC5883L_RES[] = {
```

should be enough here.

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,301 @@
...
+
+    /* data MSB @ lower address */
+    raw->x = (data[0] << 8) | data[1];
+    raw->y = (data[4] << 8) | data[5];
+    raw->z = (data[2] << 8) | data[3];
+
+    return res;
+}
+
+int hmc5883l_power_down (hmc5883l_t *dev)
```
benpicco:
```suggestion
int hmc5883l_power_down(hmc5883l_t *dev)
```

drivers/hmc5883l/hmc5883l.c
```diff
@@ -0,0 +1,301 @@
...
+{
+    assert(dev != NULL);
+    DEBUG_DEV("", dev);
+
+    /* set operation mode to Idle mode with only 5 uA current */
+    return _reg_write(dev, HMC5883L_REG_MODE, HMC5883L_OP_MODE_IDLE);
+}
+
+int hmc5883l_power_up (hmc5883l_t *dev)
```
benpicco:
```suggestion
int hmc5883l_power_up(hmc5883l_t *dev)
```

drivers/include/hmc5883l.h
```diff
@@ -0,0 +1,296 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "periph/gpio.h"
+#include "periph/i2c.h"
+
+/** Definition of error codes */
+typedef enum {
+    HMC5883L_OK             = 0,   /**< success */
```
benpicco:
```suggestion
    HMC5883L_OK             =  0,  /**< success */
```

drivers/include/hmc5883l.h
```diff
@@ -0,0 +1,296 @@
...
+ * Changes the sensor operation mode to #HMC5883L_OP_MODE_IDLE in which
+ * almost all internal blocks are switched off. I2C interface is
+ * still active. The content of the configuration registers is preserved.
+ *
+ * @param[in]   dev     Device descriptor of HMC5883L device to read from
+ *
+ * @retval  HMC5883L_OK      on success
+ * @retval  HMC5883L_ERROR_* negative error code, see #hmc5883l_error_codes_t
+ */
+int hmc5883l_power_down (hmc5883l_t *dev);
```
benpicco:
```suggestion
int hmc5883l_power_down(hmc5883l_t *dev);
```

drivers/include/hmc5883l.h
```diff
@@ -0,0 +1,296 @@
...
+ * @brief   Power up the sensor
+ *
+ * Swichtes the sensor back into the last active operation mode.
+ *
+ * @param[in]   dev     Device descriptor of HMC5883L device to read from
+ *
+ * @retval  HMC5883L_OK      on success
+ * @retval  HMC5883L_ERROR_* negative error code, see #hmc5883l_error_codes_t
+ */
+int hmc5883l_power_up (hmc5883l_t *dev);
```
benpicco:
```suggestion
int hmc5883l_power_up(hmc5883l_t *dev);
```

drivers/saul/init_devs/auto_init_hmc5883l.c
```diff
@@ -0,0 +1,72 @@
...
+#include "log.h"
+#include "saul_reg.h"
+
+#include "hmc5883l.h"
+#include "hmc5883l_params.h"
+
+/**
+ * @brief   Define the number of configured sensors
+ */
+#define HMC5883L_NUM    (sizeof(hmc5883l_params) / sizeof(hmc5883l_params[0]))
```
benpicco:
```suggestion
#define HMC5883L_NUM    ARRAY_SIZE(hmc5883l_params)
```

drivers/saul/init_devs/auto_init_hmc5883l.c
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * @brief   Memory for the SAUL registry entries
+ */
+static saul_reg_t saul_entries[HMC5883L_NUM];
+
+/**
+ * @brief   Define the number of saul info
+ */
+#define HMC5883L_INFO_NUM    (sizeof(hmc5883l_saul_info) / sizeof(hmc5883l_saul_info[0]))
```
benpicco:
```suggestion
#define HMC5883L_INFO_NUM    ARRAY_SIZE(hmc5883l_saul_info)
```

tests/driver_hmc5883l/README.md
```diff
@@ -0,0 +1,32 @@
...
+- Normal mode, no biasing
+- Gain 1090 LSB/Gs
+- No averaging of data samples
+
+The application can use different approaches to get new data:
+
+- using the #hmc5883l_read function at a lower rate than the the DOR
+- using the data-ready interrupt (**DRDY**), see #hmc5883l_init_int.
+
+To use the data-ready interrupt (**DRDY), the application has to enable
```
benpicco:
```suggestion
To use the data-ready interrupt (**DRDY**), the application has to enable
```

drivers/saul/init_devs/auto_init_hmc5883l.c
gschorcht: Yeah, the PR is older than the `ARRAY_SIZE` macro.

gschorcht:
Squashed. How do we continue, it is still blocked by @haukepetersen.

benpicco:
Please squash!

benpicco:
Can you provide the output of `tests/driver_hmc5883l`, also squash.

gschorcht:
> Can you provide the output of `tests/driver_hmc5883l`, also squash.

Squashed. The output is:
```
main(): This is RIOT! (Version: 2020.04-devel-1403-ge9c0f-drivers_hmc5883l)
HMC5883L magnetometer driver test application

Initializing HMC5883L sensor
[OK]

magn [mGs] x = 152, y = 203, z = 331
head [deg] 56
magn [mGs] x = 148, y = 200, z = 332
head [deg] 56
```

PR #10087, opened 2018-10-01
net/nanocoap: fix response type

kb2ma:
### Contribution description
coap_build_reply() uses response type of ACK for non-confirmable requests as well as confirmable. This PR fixes that error.

This error was due to nanocoap's focus on confirmable messaging as well as its use of internal macro constants for CoAP types. #10046 began the process of de-emphasizing and removing these confusing and unnecessary constants. This PR continues that work.

In fact, there is only one use remaining for the internal type constants. So, if anyone wants to write that PR, I will approve it!

### Testing procedure
Send a request of non-confirmable type to the nanocoap server example. We expect a response of non-confirmable type, but instead nanocoap responds with an ACK type.

I also added unit tests to cover this scenario, so be sure to run 'tests-nanocoap'.

### DIFF
--- a/sys/net/application_layer/nanocoap/nanocoap.c
+++ b/sys/net/application_layer/nanocoap/nanocoap.c
@@ -370,8 +370,16 @@ ssize_t coap_build_reply(coap_pkt_t *pkt, unsigned code,
         return -ENOSPC;
     }
 
-    /* if code is COAP_CODE_EMPTY (zero), use RST as type, else RESP */
-    unsigned type = code ? COAP_RESP : COAP_RST;
+    /* if code is COAP_CODE_EMPTY (zero), assume Reset (RST) type */
+    unsigned type = COAP_TYPE_RST;
+    if (code) {
+        if (coap_get_type(pkt) == COAP_TYPE_CON) {
+            type = COAP_TYPE_ACK;
+        }
+        else {
+            type = COAP_TYPE_NON;
+        }
+    }
 
     coap_build_hdr((coap_hdr_t *)rbuf, type, pkt->token, tkl, code,
                    ntohs(pkt->hdr->id));
--- a/tests/unittests/tests-nanocoap/tests-nanocoap.c
+++ b/tests/unittests/tests-nanocoap/tests-nanocoap.c
@@ -14,6 +14,7 @@
 #include <errno.h>
 #include <stdint.h>
 #include <stdbool.h>
+#include <string.h>
 
 #include "embUnit.h"
 
@@ -22,12 +23,15 @@
 #include "unittests-constants.h"
 #include "tests-nanocoap.h"
 
+
+#define _BUF_SIZE (128U)
+
 /*
  * Validates encoded message ID byte order and put/get URI option.
  */
 static void test_nanocoap__hdr(void)
 {
-    uint8_t buf[128];
+    uint8_t buf[_BUF_SIZE];
     uint16_t msgid = 0xABCD;
     char path[] = "/test/abcd/efgh";
     char loc_path[] = "/foo/bar";
@@ -59,7 +63,7 @@ static void test_nanocoap__hdr(void)
  */
 static void test_nanocoap__get_req(void)
 {
-    uint8_t buf[128];
+    uint8_t buf[_BUF_SIZE];
     coap_pkt_t pkt;
     uint16_t msgid = 0xABCD;
     uint8_t token[2] = {0xDA, 0xEC};
@@ -95,7 +99,7 @@ static void test_nanocoap__get_req(void)
  */
 static void test_nanocoap__put_req(void)
 {
-    uint8_t buf[128];
+    uint8_t buf[_BUF_SIZE];
     coap_pkt_t pkt;
     uint16_t msgid = 0xABCD;
     uint8_t token[2] = {0xDA, 0xEC};
@@ -123,7 +127,7 @@ static void test_nanocoap__put_req(void)
     len = coap_opt_finish(&pkt, COAP_OPT_FINISH_PAYLOAD);
     TEST_ASSERT_EQUAL_INT(total_hdr_len + uri_opt_len + fmt_opt_len + 1, len);
     TEST_ASSERT_EQUAL_INT(0xFF, *(pkt.payload - 1));
-    TEST_ASSERT_EQUAL_INT(&buf[0] + 128 - pkt.payload, pkt.payload_len);
+    TEST_ASSERT_EQUAL_INT(&buf[0] + _BUF_SIZE - pkt.payload, pkt.payload_len);
 }
 
 /*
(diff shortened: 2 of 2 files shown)


sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -370,8 +370,19 @@ ssize_t coap_build_reply(coap_pkt_t *pkt, unsigned code,
         return -ENOSPC;
     }
 
-    /* if code is COAP_CODE_EMPTY (zero), use RST as type, else RESP */
-    unsigned type = code ? COAP_RESP : COAP_RST;
+    /* if code is COAP_CODE_EMPTY (zero), assume Reset (RST) type */
+    unsigned type;
```
smlng: minor suggestion: `unsigned type = COAP_TYPE_RST;` and omit the `else` below for readability. Otherwise looks good

tests/unittests/tests-nanocoap/tests-nanocoap.c
```diff
@@ -250,6 +254,108 @@ static void test_nanocoap__get_path_too_long(void)
...
+    TEST_ASSERT_EQUAL_INT(COAP_METHOD_GET, coap_get_code(&pkt));
+    TEST_ASSERT_EQUAL_INT(COAP_TYPE_CON, coap_get_type(&pkt));
+}
+
+/* Builds on test_nanocoap__server_reply_simple to test confirmable request. */
+static void test_nanocoap__server_reply_simple_con(void)
+{
+    uint8_t buf[_BUF_SIZE];
+    coap_pkt_t pkt;
+    char payload[2] = "0";
```
smlng: I guess `char payload[2] = { 0 };` is (more) correct here?

tests/unittests/tests-nanocoap/tests-nanocoap.c
```diff
@@ -250,6 +254,108 @@ static void test_nanocoap__get_path_too_long(void)
...
+    coap_get_uri_path(&pkt, (uint8_t *)&uri[0]);
+    TEST_ASSERT_EQUAL_STRING((char *)path, (char *)uri);
+}
+
+/* Response for server GET request using coap_reply_simple(). */
+static void test_nanocoap__server_reply_simple(void)
+{
+    uint8_t buf[_BUF_SIZE];
+    coap_pkt_t pkt;
+    char payload[2] = "0";
```
smlng: same here

tests/unittests/tests-nanocoap/tests-nanocoap.c
```diff
@@ -250,6 +254,108 @@ static void test_nanocoap__get_path_too_long(void)
...
 }
 
+/*
+ * Helper for server_get tests below.
+ * Request for nanocoap server example /riot/value resource.
+ * Includes 2-byte token; non-confirmable.
+ */
+static int _read_riot_value_req(coap_pkt_t *pkt, uint8_t *buf)
+{
+    uint8_t pkt_data[] = {
```
smlng: where do these come from, I mean what's the content,  how is it generated?

tests/unittests/tests-nanocoap/tests-nanocoap.c
```diff
@@ -250,6 +254,108 @@ static void test_nanocoap__get_path_too_long(void)
...
+}
+
+/*
+ * Helper for server_get tests below.
+ * Request for nanocoap server example /riot/value resource.
+ * Includes 2-byte token; confirmable.
+ */
+static int _read_riot_value_req_con(coap_pkt_t *pkt, uint8_t *buf)
+{
+    uint8_t pkt_data[] = {
```
smlng: same here, looks different from the other one, still unclear to me what and why

tests/unittests/tests-nanocoap/tests-nanocoap.c
smlng: mhm, okay the comment says something, but still how is it "generated" -> captured?

tests/unittests/tests-nanocoap/tests-nanocoap.c
```diff
@@ -22,12 +23,15 @@
 #include "unittests-constants.h"
 #include "tests-nanocoap.h"
 
+
+#define _BUF_SIZE (128)
```
smlng: might be good to define as `(128U)`

sys/net/application_layer/nanocoap/nanocoap.c
kb2ma: OK, no problem.

tests/unittests/tests-nanocoap/tests-nanocoap.c
kb2ma: I'd like for this to read like string creation to make it a realistic example. Would it be clearer if I used:

`char *payload = "0"`

tests/unittests/tests-nanocoap/tests-nanocoap.c
kb2ma: I probably used libcoap to generate it, but I'm sure I reviewed it in Wireshark. I'll add a note to that effect.

tests/unittests/tests-nanocoap/tests-nanocoap.c
kb2ma: Byte 1 is different since confirmable, and this is noted in the comment. Bytes 3 and 4 encode a different message ID, but I didn't think that was worth commenting on.

tests/unittests/tests-nanocoap/tests-nanocoap.c
kb2ma: Thanks for pointing this out because I'm unclear about it. When is it necessary/correct to specify unsigned for macros?

kb2ma:
Thanks for the review, @smlng. Responses inline.

tests/unittests/tests-nanocoap/tests-nanocoap.c
smlng: I think it's always good to be as specific as possible. Also I think it might avoid problems when compiling on different platforms, i.e. 8, 16 or 32 bit.

kb2ma:
Addressed comments.

smlng:
tested ACK now, works as described, i.e. answering with NON when request was NON, too.

Please squash

kb2ma:
Squashed and rebased. Hopefully Murdock will smile down upon it.

kb2ma:
Hard to imagine how this PR would cause tests/cpp11_mutex to fail to find include files. :-/

smlng:
mhm, maybe somethings wrong with that Murdock-worker (its `ifive` @kaspar030 ?), so bad luck, I'll trigger murdock again, and hopefully the tasks are dealt differently.

kb2ma:
Looks like the problem was an instance of #10166. @smlng, are we ready to merge?

smlng:
yes we are, ACK and GO

PR #10093, opened 2018-10-01
drivers/periph_eeprom: refactor implementation and test application

aabadie:
### Contribution description

This PR follows discussions in #10026 and #10056 and refactor the architecture specific implementation:
Before, each CPU architecture with an internal had to implement eeprom_read_byte/eeprom_write_byte functions and eeprom_read/eeprom_write were just wrappers on top of the former (basically looping on each byte).
This PR changes this to the opposite: each CPU now provides a specific implementation of eeprom_read and eeprom_write and eeprom_read_byte/eeprom_write_byte are special case of them. The API is kept unchanged.

This will allow architecture specific implementations to provide optimization (like what's proposed in #10026) without having to change the API.

In the meantime, I refactor slightly the test application:
- added 2 new commands for manipulating eeprom_read_byte/eeprom_write_byte functions
- added a new test command that performs different calls to the different API functions and that checks everything is correctly written/read
- added an automatic python script that calls this test function

Thus, this PR should be merged before the 2 other related PRs #10026 and #10056 (since it provides an automatic testing procedure and other fixes).

### Testing procedure

- Grab a board with a CPU providing internal EEPROM: arduino-mega2560, nucleo-l073rz, nucleo-l152re
- Run the following command:
```
$ make BOARD=<your board> -C tests/periph_eeprom flash test
```

It should run without error.

### Issues/PRs references

Following the discussion in #10056 and should be helpful for #10026.

### DIFF
--- a/cpu/atmega_common/periph/eeprom.c
+++ b/cpu/atmega_common/periph/eeprom.c
@@ -24,37 +24,52 @@
 #include "cpu.h"
 #include "periph/eeprom.h"
 
-uint8_t eeprom_read_byte(uint32_t pos)
+#define ENABLE_DEBUG (0)
+#include "debug.h"
+
+size_t eeprom_read(uint32_t pos, uint8_t *data, size_t len)
 {
-    assert(pos < EEPROM_SIZE);
+    assert(pos + len <= EEPROM_SIZE);
 
-    /* Wait for completion of previous operation */
-    while (EECR & (1 << EEPE)) {}
+    uint8_t *p = data;
 
-    /* Set up address register */
-    EEAR = pos;
+    DEBUG("Reading data from EEPROM at pos %" PRIu32 ": ", pos);
+    for (size_t i = 0; i < len; i++) {
+        while (EECR & (1 << EEPE)) {}
 
-    /* Start eeprom read by writing EERE */
-    EECR |= (1 << EERE);
+        /* Set up address register */
+        EEAR = pos++;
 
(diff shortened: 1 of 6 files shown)


aabadie:
@MrKevinWeiss, would you like to review and give this one a try ?

MrKevinWeiss:
Why is len returned on eeprom_write and eeprom_read if nothing is actionable?

MrKevinWeiss:
Also not a big deal but why not write the test logic in the test.py not in the c file.  This should allow it to be more extendable and have a smaller binary.

MrKevinWeiss:
I did run it though the tests and it passed, also did some manual playing around and everything looks correct.  This was done on the nucleo-l152re and arduino-mega2560.

aabadie:
> Why is len returned on eeprom_write and eeprom_read if nothing is actionable?

If nothing is actionable, it won't change anything because the program will either crash (due to an assertion failed) or hang (because the EEPROM cannot be read/write). This was the previous behaviour by the way.

MrKevinWeiss:
Ok, part of keeping the API then fair enough.

aabadie:
> why not write the test logic in the test.py not in the c file

That's a matter of taste. I prefer it this way ;)

MrKevinWeiss:
Heh, alright then, maybe it's the Canadian in me but I won't push back.

aabadie:
@MrKevinWeiss, CI is all green there !

PR #10099, opened 2018-10-02
sys/auto_init/saul: fix of #10098

### LINKED ISSUE #10098: sys/autoinit/saul:compilation problem in auto_init_*.c
#### Description

Compilation of ```tests/saul``` fails for most modules when the architecture doesn't  include ```thread.h``` in ```$(RIOTCPU)/$(CPU)/include/cpu.h```.

All ```sys/auto_init/saul/auto_init_*.c``` files use ```assert``` as following
```
 assert(<driver>_INFO_NUM == <driver>_NUM);
```
to ensure that the number of ```<driver>_params``` entries matches with the number of ```<driver>_saul_info entries```.

Unfortunatly, ```assert.h``` is not included. Nevertheless, it works for most platforms / boards because ```assert.h``` is included via the following path:
```
<driver>_params.h -> board.h -> cpu.h -> thread.h -> cib.h -> assert.h
```
Even though it would be easy to include ```thread.h``` in every ```$(RIOTCPU)/$(CPU)/include/cpu.h```, from my point of view, the file which is using a certain function is responsible to include everything it needs. Therefore, I would suggest to add
```
#include "assert.h"
```
in all ```sys/auto_init/saul/auto_init_*.c``` files. I could provide an according PR if desired.

```
USEMODULE="si7021" make flash BOARD=esp8266-esp-12x -C tests/saul QEMU=0
```

#### Versions

Observed with ESP8266.

gschorcht:
### Contribution description

The PR fixes the compilation problem of undeclared ```assert``` which happens only if architecture ```cpu.h``` does not include ```thread.h``` and thus ```assert.h```.

The PR includes:

- missing ```#include "assert.h"``` added
- moved ```/sys/auto_init/shtx1.c``` to ```/sys/auto_init/saul/```
- sorted the order of calls of saul ```auto_inits_<driver>``` modules in ```/sys/auto_init/auto_init.c```
- added the mission call of ```auto_inits_sht1x```

### Testing procedure

Compilation with command:
```
USEMODULE="adcxx1c adc081c ads101x adxl345 bmp180 bmx055 bmx280 dht fxos8700 grove_ledbar hdc1000 hts221 io1_xplained isl29020 jc42 l3g4200d lis2dh12_spi lis3dh lis3mdl lps331ap lsm303dlhc lsm6dsl mag3110 mma7660 mma8x5x mpl3115a2 mpu9150 pir pulse_counter sht1x si114x si7021 tcs37727 tmp006 tsl2561 veml6070" make BOARD=esp8266-esp-12x -C tests/saul
```

### Issues/PRs references

fixes #10098


sys/auto_init/auto_init.c
```diff
@@ -284,42 +284,106 @@ void auto_init(void)
...
-    auto_init_gpio();
-#endif
 #ifdef MODULE_SAUL_ADC
     extern void auto_init_adc(void);
     auto_init_adc();
 #endif
-#ifdef MODULE_LSM303DLHC
-    extern void auto_init_lsm303dlhc(void);
-    auto_init_lsm303dlhc();
+    #ifdef MODULE_SAUL_GPIO
```
aabadie: There's an extra indent here.

aabadie:
Murdock failure seems unrelated btw.

sys/auto_init/auto_init.c
gschorcht: Oops :worried:

gschorcht: Done. May I reset the branch to commit sys/auto_init/auto_init.c separately?

aabadie:
@gschorcht, I see 3 kinds of changes in this PR:
1. alphabetical ordering of call to `auto_init_<xxx>` functions
2. include assert.h (original fix)
3. a rename of a macro `XXXX_NUM` to `XXXX_NUMOF`

All changes are perfectly fine but please put each of them in separate commits. Don't use the number of the issue it fixes in the comment, this is too much GitHub dependent. Thanks.

gschorcht:
@aabadie Changed, hopefully in the right way :smile:

maribu:
@aabadie, @gschorcht: Regarding commit dc8c363c45722d3dd483bc582d7deff01cd67035, I wanted to point out that `auto_init_sht1x.c` is only partly related to SAUL. There are dedicated shell commands for the SHT1X that require the auto-initializing of the driver provided by it, but are not depending on SAUL. The file does perform the SAUL registration as well, but only if SAUL is actually used. I'm not really convinced that moving it into the saul folder was a good idea, because one easily assumes now it depends on SAUL, which it does not.

gschorcht:
@maribu I apologize a lot, it was not really clear to me that the SHT1x initialization is handled differently than with all other sensors.

The question for me now is why the SHT1x initialization is handled differently. Should not all sensors follow the same interface? SHT1x is the only sensor initialized by ```auto_init``` and not by the application.

I see the following options:

- It is possible that the initialization of sensors and actuators can be performed by ```auto_init```. But, in that case sensors and actuators should get a separate subdirectory, like ```netif``` or ```storage```. Or ```auto_init_<driver>.c``` is placed in the directory of the driver. Having a single ```auto_init_<driver>.c``` in ```sys/auto_init``` directory is probably not desirable.
- It should not be possible to perform the initialization of sensors and actuators automatically via ```auto_init```.

maribu:
No need to apologize at all ;-)

I agree that the original place don't fit 100% either and your proposed to options make totally sense to me.

Regarding the second option: Having the driver auto-initialized makes developing the application much easier. I believe there is no real use case when someone decides to add the driver to USEMODULES, but do not want to use it. (One could image that the driver is only used every 10th boot and the other 9 boots the overhead of the initialization could be skipped, but that seems purely hypothetical to me.) So outsourcing the initialization to auto_init makes life easier and helps preventing bugs. (The auto_init code will be more widely tested and reviewed, thus it will be less bug-prone.)

A short grep over the files in sys/auto_init/saul shows that the SHT1X driver is indeed the only one only partly relevant to SAUL. There would be no confusion if all auto-init code their would have the same initialization scheme, that is initialize any enabled actuator/sensor driver, but only register it to SAUL if SAUL is also enabled. All required to achieve this would be adding two pairs of `#ifdef MODULE_AUTO_INIT_SAUL` and `#endif` to each file. I could create an PR to just do that. But some code won't make much sense without SAUL, e.g. `auto_init_adc.c` or `auto_init_gpio.c` seem to make little sense without SAUL to me.

@haukepetersen might have an opinion about this as author of the SAUL-API.

gschorcht:
> Having the driver auto-initialized makes developing the application much easier.

Fully agree.

> I believe there is no real use case when someone decides to add the driver to USEMODULES, but do not want to use it.

The advantage of doing initialization in application's ```main.c``` is that it becomes possible to override some default parameters without having a complete ```<driver>_params.h``` in application's directory, e.g.,

```
#define LIS3DH_PARAM_INT1           (GPIO_PIN(0, 5))

#include "lis3dh.h"
#include "lis3dh_params.h"
```
Having the define ```LIS3DH_PARAM_INT1``` in front of the includes, overrides the default definition.

maribu:
> The advantage of doing initialization in application's main.c is that it becomes possible to override some default parameters without having a complete `<driver>_params.h` in application's directory [...]

This is indeed a good point. It would be possible using the Makefile by adding `CFLAGS += '-DLIS3DH_PRAM_INIT=(GPIO_PIN(0, 5))'` as well, but this might be less readable.

But this problem affects e.g. network device drivers in any case, as manual initialization is not an option there. A really nice feature would be if the RIOT build system would look for a specifically named header file, e.g. `global_defines.h` in the application folder, and include that file via `CFLAGS += "-include $(APPDIR)/global_define.h"` - but only if the file is present.

gschorcht:
> A really nice feature would be if the RIOT build system would look for a specifically named header file, e.g. global_defines.h in the application folder, and include that file via CFLAGS += "-include $(APPDIR)/global_defines.h" - but only if the file is present.

Yes, that would also be my preferred approach. That's why I started a discussion (# 9825) on this topic a few weeks ago.

PR #10108, opened 2018-10-03
coap: Replace `COAP_CT_` with `COAP_FORMAT_`

bergzand:
### Contribution description

According to the IANA defines, there is only CoAP content-formats. and no CoAP content types. This PR merges the list of content types into the list of content formats and replaces all `COAP_CT_.*`occurences in our code with `COAP_FORMAT_.*`. I merged into the `COAP_FORMAT_` options because those were more often used than the `COAP_CT_` defines.

Not sure whether to call this an API_CHANGE, but all example code uses `COAP_FORMAT_`, so the old defines "should" not have been used outside our implementations, but if somebody wants to call this an api change I'm fine with it.

### Testing procedure

A small test to verify that the `.well-known/core` still has content number 40 should be enough here.

### Issues/PRs references

None

### DIFF
--- a/sys/include/net/coap.h
+++ b/sys/include/net/coap.h
@@ -123,6 +123,8 @@ extern "C" {
 
 /**
  * @name    Content types
+ * @deprecated  Deprecated in favour of [COAP_FORMAT_](@ref net_coap_format)
+ *              style defines
  * @{
  */
 #define COAP_CT_LINK_FORMAT     (40)
@@ -134,11 +136,14 @@ extern "C" {
 
 /**
  * @name    Content-Format option codes
+ * @anchor  net_coap_format
  * @{
  */
 #define COAP_FORMAT_TEXT         (0)
 #define COAP_FORMAT_LINK        (40)
+#define COAP_FORMAT_XML         (41)
 #define COAP_FORMAT_OCTET       (42)
+#define COAP_FORMAT_EXI         (47)
 #define COAP_FORMAT_JSON        (50)
 #define COAP_FORMAT_CBOR        (60)
 /** @} */
--- a/sys/net/application_layer/gcoap/gcoap.c
+++ b/sys/net/application_layer/gcoap/gcoap.c
@@ -981,7 +981,7 @@ uint8_t gcoap_op_state(void)
 int gcoap_get_resource_list(void *buf, size_t maxlen, uint8_t cf)
 {
     (void)cf; /* only used in the assert below. */
-    assert(cf == COAP_CT_LINK_FORMAT);
+    assert(cf == COAP_FORMAT_LINK);
 
     /* skip the first listener, gcoap itself (we skip /.well-known/core) */
     gcoap_listener_t *listener = _coap_state.listeners->next;
--- a/sys/net/application_layer/nanocoap/nanocoap.c
+++ b/sys/net/application_layer/nanocoap/nanocoap.c
@@ -842,7 +842,7 @@ ssize_t coap_well_known_core_default_handler(coap_pkt_t *pkt, uint8_t *buf, \
     coap_block2_init(pkt, &slicer);
     uint8_t *payload = buf + coap_get_total_hdr_len(pkt);
     uint8_t *bufpos = payload;
-    bufpos += coap_put_option_ct(bufpos, 0, COAP_CT_LINK_FORMAT);
+    bufpos += coap_put_option_ct(bufpos, 0, COAP_FORMAT_LINK);
     bufpos += coap_opt_put_block2(bufpos, COAP_OPT_CONTENT_FORMAT, &slicer, 1);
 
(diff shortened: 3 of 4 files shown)


smlng:
one thing: isn't this somehow an API CHANGE? Some RIOT external APP could use the old defines, right?

bergzand:
> one thing: isn't this somehow an API CHANGE? Some RIOT external APP could use the old defines, right?

Yeah, that's why I mentioned it in the opening post. If everybody nicely based their applications on our examples it should not be a problem, but that's a huge if there. Anyway, I'm fine with this being considered an API change.

smlng:
I guess it's good to have the label for documentation and release changelog creation.

smlng:
though I'd like to have second ACK on this one then, too

bergzand:
> I guess it's good to have the label for documentation and release changelog creation.

Fair enough

> though I'd like to have second ACK on this one then, too

Agreed

Thanks for the review!

jia200x:
> So, does it make sense to mark the COAP_CT_xxx macros as deprecated for a couple of releases? I would like to see us get into this mechanism for nanocoap/gcoap as the API will be evolving for function names. For example, coap_put_option_ct() (with the same problem as these macros) should change to coap_opt_put_format() or something like that. So, it makes sense to me to deprecate the current function name for a while.

I'm +1 for adding a `@deprecated` tag to these macros

bergzand:
Restored the removed defines and added a `@deprecated` tag with a message to them.

kb2ma:
Thanks, @bergzand. The deprecated tag really stands out in Doxygen. This all looks good and tests fine for me.

There are a few uses of the "deprecated" tag in the code base. I don't know about historical use though. So, I would like to merge this PR, and then I plan to send create an RFC Issue to confirm there is no concern with this approach systematically for evolving the gcoap/nanocoap APIs forward. It helps to have an example. :-)

@smlng, any concerns? If not, I'm ready to squash and go.

deprecated

sys/include/net/coap.h
```diff
@@ -138,7 +139,9 @@ extern "C" {
  */
```
miri64: You could add an anchor to this group and reference it above:

```C
 /**
  * @name    Content-Format option codes
  * @anchor  net_coap_format
  * @{
  */
```

```
@deprecated  Deprecated in favour of [`COAP_FORMAT_`](@ref net_coap_format)
             style defines
```

bergzand: > You could add an anchor to this group and reference it above:

Awesome, I was looking for something like that!

bergzand: Done

kb2ma:
Link is an improvement (thanks @miri64), and bold looks even better. Unfortunately the use of black color for text means it doesn't stand out as a link.

deprecated2

bergzand:
> Unfortunately the use of black color for text means it doesn't stand out as a link.

Yeah, also noticed that, not sure if we can fix that without modifying the css file.

smlng:
Looks good to me, please squash

bergzand:
Squashed and rebased. I've reworded the second commit message to better reflect the current changes.

kb2ma:
I saw the build failure. What went wrong?

Also, and this is drifting off topic, but when is it appropriate to tag with 'run tests'? I looked around yesterday for more background, but could not find an answer.

bergzand:
> @bergzand you broke the core wink I'll rerun Murdock

But how...?

smlng:
@bergzand I was joking, I think the test is flawed or at least not deterministic in results on native

kb2ma:
Thanks everyone for talking this through and improvements.

bergzand:
Now we only have to remember to actually remove the defines in a release or two :)

miri64:
> Now we only have to remember to actually remove the defines in a release or two :)

At some point we need to go through [this list](http://doc.riot-os.org/deprecated.html) anyway and start removing stuff ;-)

jia200x:
Great! Thanks so much to everyone involved!

> At some point we need to go through this list anyway and start removing stuff ;-)

+1. There could be some kind of automation between releases :)

miri64:
(also the interval deprecated should be removed is something that should very much be decided on a case-by-case basis)

kb2ma:
Thanks all for the comments! I plan to post a deprecation RFC in the next few days, and will look for consensus on a trial approach to this workflow.

BTW @miri64, I know I have seen that "Deprecated List" in Doxygen hundreds of times as I click into something else, but it didn't register in my mind that it existed when I looked for existing deprecated code. Thanks for calling that out.

jia200x:
> I think letting some bot remove stuff from our code is something that would lead to desaster.

I mean something to warn that a feature should be deprecated. We could tag the time somehow.
Never intended to automatically remove tags but more into the direction of inform maintainers :)

PR #10110, opened 2018-10-04
drivers: htu21d relative humidity and temperature sensor driver

gschorcht:
### Contribution description

This PR adds a driver for the TE Connectivity HTU21D humidity and temperature sensor. The driver could also be used with Sensirion SHT21 and with SI7021.

The sensor only supports single measurements and polling. This driver provides ```drivers/saul``` capabilities.

**Please note:**
Even though the HTU21D driver also works with SI7021 sensor for which ```drivers/si70xx``` already exists, there the following differences:

- ```drivers/si70xx``` checks some registers during intialization that do no exist in HTU21D sensor (heater control register and revision registers).
- ```drivers/si70xx``` uses clock stretching which blocks the bus and the master during  the measurement which can take up to 85 ms. ```drivers/htu21d``` does not use clock stretching to avoid this blocking.
- ```drivers/si70xx``` does not implement the CRC check while ```drivers/htu21d``` implement and use it.

As an alternative, I could provide a PR that
- defines a pseudomodule for HTU21D,
- modifies ```drivers/si70xx``` so that it would also work also with HTU21D,
- includes wrapper files ```drivers/include/htu21d.h``` and ```drivers/htu21d/include/htu21d_params.h which just redefine symbols with prefix ```htu21d_```, and
- includes a ```tests/driver_htu21d``` as test application.

I have it already realized it also in that way and could provide it. The same could be done for SHT21.
But please keep in mind that a pseudo/wrapper module would use  ```drivers/si70xx``` which blocks the bus and the master while waiting for measurement results.

### Testing procedure

USEMODULE=htu21d make flash -C tests/saul BOARD=...

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -175,6 +175,10 @@ ifneq (,$(filter hts221,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter htu21d,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_i2c
+endif
+
 ifneq (,$(filter ina220,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
--- a/drivers/htu21d/Makefile
+++ b/drivers/htu21d/Makefile
@@ -1,0 +1,3 @@
+MODULE = htu21d
+
+include $(RIOTBASE)/Makefile.base
--- a/drivers/htu21d/htu21d.c
+++ b/drivers/htu21d/htu21d.c
@@ -1,0 +1,381 @@
+/*
+ * Copyright (C) 2018 Gunar Schorcht
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_htu21d
+ * @brief       Device driver for the HTU21D humidity and temperature sensor
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ * @{
+ */
+
+#include <errno.h>
+#include <string.h>
+#include <stdlib.h>
+
+#include "htu21d_regs.h"
+#include "htu21d.h"
+
+#include "log.h"
+#include "xtimer.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+#if ENABLE_DEBUG
+
+#define ASSERT_PARAM(cond) \
+    if (!(cond)) { \
+        DEBUG("[htu21d] %s: %s\n", \
+              __func__, "parameter condition (" # cond ") not fulfilled"); \
+        assert(cond); \
+    }
+
+#define DEBUG_DEV(f, d, ...) \
+        DEBUG("[htu21d] %s i2c dev=%d addr=%02x: " f "\n", \
+              __func__, d->params.dev, d->params.addr, ## __VA_ARGS__);
+
+#else /* ENABLE_DEBUG */
+
+#define ASSERT_PARAM(cond) assert(cond);
+#define DEBUG_DEV(f, d, ...)
+
+#endif /* ENABLE_DEBUG */
+
+#define ERROR_DEV(f, d, ...) \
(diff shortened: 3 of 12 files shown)


maribu:
@gschorcht: I could review this PR, but I don't have the hardware to test it. Should I still review?
@basilfx: Maybe you could get involved as author of the `si70xx`.

To me the advantage of not using clock stretching to allow using multiple I2C device on the same bus is obvious. Is there any technical argument to not use this approach in the `si70xx`? If not, I personally would favor a unified driver for all devices that would work in a non-blocking fashion, unless one of the following would result:

 - The resulting unified driver would be bigger than two separate drivers
 - The resulting driver would perform significantly poorer (e.g. regarding performance, features, ROM size) than each of the two separate driver
 - The resulting driver would be harder to maintain than two separate drivers (e.g. code gets way more complex)
 - Simultaneous use of `si70xx`, `htu21d`, and `sht21` devices would not be possible by one unified driver, but it would be by separate drivers.

Note: I did not look yet into either this code or the `si70xx` drivers. So maybe the above questions are obvious.

gschorcht:
@maribu Thanks for your offer to review it. We should wait some more time for an answer from @basilfx.

basilfx:
I'm in the middle of a moving, so I won't be able to test this soon.

gschorcht:
PR is waiting for review.

aabadie:
Needs rebase

drivers/htu21d/Makefile
```diff
@@ -0,0 +1,3 @@
+MODULE = htu21d
+
```
aabadie, line 2:
```suggestion
```

drivers/htu21d/htu21d.c
```diff
@@ -0,0 +1,381 @@
...
+
+#define EXEC_RET_CODE(f, c) { \
+    int _r; \
+    if ((_r = f) != HTU21D_OK) { \
+        DEBUG("[htu21d] %s: error code %d\n", __func__, _r); \
+        return c; \
+    } \
+}
+
+/** Forward declaration of functions for internal use */
```
aabadie, line 70:
```suggestion
/* Forward declaration of functions for internal use */
```

drivers/Makefile.dep
```diff
@@ -175,6 +175,10 @@ ifneq (,$(filter hts221,$(USEMODULE)))
   FEATURES_REQUIRED += periph_i2c
 endif
 
+ifneq (,$(filter htu21d,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_i2c
```
aabadie, line 179:
```suggestion
  FEATURES_REQUIRED += periph_i2c
  USEMODULE += xtimer
```

Because there's a strong dependency on xtimer in the driver implementation.

drivers/htu21d/htu21d.c
```diff
@@ -0,0 +1,381 @@
...
+                user_reg.resolution_1 = 0;
+                user_reg.resolution_0 = 1;
+                break;
+    }
+    EXEC_RET(_user_reg_write(dev, user_reg.val));
+
+    return HTU21D_OK;
+}
+
+static const uint8_t _times_tmp[] = { 85, 22, 43, 11 };
```
aabadie, line 121: Could we have some defines for this constants ?

drivers/htu21d/htu21d.c
```diff
@@ -0,0 +1,381 @@
...
+            if (res != HTU21D_OK && res != -HTU21D_ERROR_MEASURING) {
+                return res;
+            }
+        } while (res != HTU21D_OK);
+    }
+
+    return HTU21D_OK;
+}
+
+/** Functions for internal use only */
```
aabadie, line 249:
```suggestion
/* Functions for internal use only */
```

maribu:
> This PR adds a driver for the TE Connectivity HTU21D humidity and temperature sensor. The driver could also be used with Sensirion SHT21 and with SI7021.

Could this replace the `sth2x` and the `si70xx` driver? If so, I really would love to have less code to maintain to provide the same features.

This would also be greatly beneficial in case someone (for whatever reason) connects a bunch of temperature+humidity sensors of the various flavors, as with the single driver only ROM space for one implementation would be needed.

aabadie:
Please rebase. There are also some pending review comments.

mguetschow:

PR #7976, opened 2017-11-09
make: allow alias for board names

haukepetersen:
Some boards have unique, but quite ugly names (e.g. `pba-d-01-kw2x`) and we use different names in the day-to-day operation (e.g. `phynode` for the former). Event though this 'nicer' name ist not a 100% correct (as there are also phynodes based on other CPUs out there, just not in the RIOT universe...), it is sometimes much easier to remember and use. Also, there are some boards that have a specific, cryptic name (e.g. `SLTB001A`), but also a nice user friendly one (`thunderboard-sense` in that case). So I think it would be nice, to be able to use both (or even more?) aliases for a certain board.

This PR contains a quick draft, on how this could be implemented. It works, but is it nice enough?
```
BOARD=phynode make ...   -> will build `BOARD=pba-d-01-kw2x`
BOARD=linux make ... -> will build 'BOARD=native`, just for demo purposes...
```

Maybe we can even think of something that combines this approach with my idea about board revisions ( #7975). What do you think?

PR #10116, opened 2018-10-05
Make support for the stm32f429i-disco more obvious

x3ro:
### Contribution description

There are two different iterations of the stm32f429i discovery board,
the `-DISCO` and the `-DISC1`. Both of these are virtually the same,
except for their use of a different version of ST-Link.

When trying to use the `-DISCO` variation of the board, the fact that I
have to use the `-DISC1` board was not obvious to me. I tried to go with
the `stm32f4-discovery`, but that one uses a different MCU altogether.

This PR intends to make clearer that the board configuration can be used
for both boards by changing the `-disc1` suffix to `-disc` and improving
the board specific documentation.

### Testing procedure

Follow the provided board documentation to flash onto the `stm32f429i-disco` and/or `stm32f429i-disc1`.

### Issues/PRs references

- N/A


x3ro:
Hey @aabadie, since you're the author of the original board, maybe you could take a look? :)

boards/stm32f429i-disc/Makefile.include
```diff
@@ -14,6 +14,7 @@ include $(RIOTMAKE)/tools/serial.inc.mk
 
 # this board has an on-board ST-link adapter
 export DEBUG_ADAPTER ?= stlink
+export STLINK_VERSION ?= 2-1
```
aabadie: useless, 2-1 is already the default. See [here](https://github.com/RIOT-OS/RIOT/blob/master/makefiles/tools/openocd-adapters/stlink.inc.mk#L3)

boards/stm32f429i-disc/doc.txt
```diff
@@ -0,0 +1,36 @@
+/**
+ * @defgroup    boards_stm32f429i-disc STM32F429I-DISCO and -DISC1
+ * @ingroup     boards
+ * @brief       Support for the `STM32F429-DISCO` and `STM32F429-DISC1` boards.
+
```
aabadie: You don't need all those blank lines between sections. One would be enough

boards/stm32f429i-disc/Makefile.include
x3ro: I know its the default. This was mainly for reference, since it's intended to be changed per board. In other words, by looking at the makefile of the board I see the relevant variables. I can remove it if you feel strongly about this, though.

boards/stm32f429i-disc/doc.txt
x3ro: Yeah, this is just my style of writing markdown I guess. Is there a markdown styleguide for RIOT? I don't recall.

boards/stm32f429i-disc/Makefile.include
aabadie: If you look at other boards that uses st-link v2-1, this variable is not used. You could add a comment if you prefer.

boards/stm32f429i-disc/doc.txt
aabadie: > Is there a markdown styleguide for RIOT?

I don't think so. And looking at other files there's no 4 blank lines after the doxygen header in general. The number of blank lines before section is between 1 and 3 in general.
i prefer using the minimum blank lines possible in general but this is personal taste.

aabadie:
After some more thought, I finally don't like this PR in the current state. The title says "make support for the stm32f429i-disco more obvious" but in fact, it's worse for both `stm32f429i-disc1` and `stm32f429i-disco` because the board name (from the build system point of view) is neither one or the other.
If one wants to build for one of the 2 boards, he has to know it's now called stm32f429i-disc (without 1 or o).
Also, according to [ST website](https://www.st.com/en/evaluation-tools/32f429idiscovery.html), stm32f429i-disco is replaced by stm32f429i-disco, so it is perfectly fine to keep the disc1 as default.

To improve things, I propose to just update the documentation of the board with a mention to the old disco version (about how to flash and how to get the stdio).
In a second step, revamping #7976 will also help: this PR was proposing an aliasing mechanism for boards and I think it will perfectly fit here. With this mechanism it might also be possible to override automatically the ST-Link version when building for `stm32f429i-disco`, which will make things even more simple.

x3ro:
I admit that is somewhat disappointing, and I don't particularly agree with your reasoning. I do agree that an alias would be preferable, but that mechanism does not exist and it doesn't seem like anyone is working on it.

That being said, this is too low on my list of priorities to warrant further discussion, so I'll simply go ahead and close it. I'll open a separate PR to update the documentation, since that seems to far removed from the original intent of this PR to simply modify this one.

PR #10118, opened 2018-10-05
tests/sx127x: set b-l072z-lrwan1 as default board

aabadie:
### Contribution description

While trying to build the `tests/driver_sx127x` (LoRa PHY) application for my preferred LoRa board (b-l072z-lrwan1), I noticed 2 things:
- by default (when setting no BOARD from the command line), the application is built for samr21-xpro, whereas the default should be (from the Makefile) nucleo-l152re
- the default is nucleo-l152re whereas this board doesn't provide a LoRa radio par default (you need an external shield)

This PR fixes the 2 points. Now the default is set to be b-l072z-lrwan1 and when no BOARD is set, it's this one that is used.

### Testing procedure

make -C tests/driver_sx127x

The board should be b-l072z-lrwan1

### Issues/PRs references

None

### DIFF
--- a/tests/driver_sx127x/Makefile
+++ b/tests/driver_sx127x/Makefile
@@ -1,8 +1,8 @@
+BOARD ?= b-l072z-lrwan1
+
 include ../Makefile.tests_common
 
 BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-uno nucleo-f031k6
-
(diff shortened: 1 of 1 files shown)


MrKevinWeiss:
Rerunning murdock just in case.

aabadie:
And you know what I would like to tell you @MrKevinWeiss !

PR #10119, opened 2018-10-06
tests: fix including Makefile.tests_common

olegart:
### Contribution description

Fixes `include ../Makefile.tests_common` in Makefiles — it should be placed at the end of the file, before the final `include $(RIOTBASE)/Makefile.include`, otherwise it can interfere with other variables.

E.g. `BOARD ?= something` does nothing as BOARD already defined in Makefile.tests_common

### Issues/PRs references

#10118

### DIFF
--- a/tests/bench_timers/Makefile
+++ b/tests/bench_timers/Makefile
@@ -1,5 +1,3 @@
-include ../Makefile.tests_common
-
 BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-uno
 
 # These boards only have a single timer in their periph_conf.h, needs special
@@ -53,4 +51,5 @@ test-kinetis-lptmr: all
 # Reset the default goal.
 .DEFAULT_GOAL :=
 
+include ../Makefile.tests_common
 include $(RIOTBASE)/Makefile.include
--- a/tests/emb6/Makefile
+++ b/tests/emb6/Makefile
@@ -1,6 +1,5 @@
 # overwrite board, do not set native as default
 BOARD ?= samr21-xpro
-include ../Makefile.tests_common
 
 # MSP-430 doesn't support C11's atomic functionality yet
 BOARD_BLACKLIST := msb-430 msb-430h pic32-clicker pic32-wifire \
@@ -41,4 +40,5 @@ USEMODULE += $(DRIVER)
 CFLAGS += -Wno-unused-parameter -Wno-unused-function -Wno-type-limits
 CFLAGS += -Wno-sign-compare -Wno-missing-field-initializers
 
+include ../Makefile.tests_common
 include $(RIOTBASE)/Makefile.include
--- a/tests/float/Makefile
+++ b/tests/float/Makefile
@@ -1,5 +1,3 @@
-include ../Makefile.tests_common
-
 DISABLE_MODULE += auto_init
 
 # for native we can go do a couple of more operations in reasonable time...
@@ -9,4 +7,5 @@ endif
 
 TEST_ON_CI_WHITELIST += all
 
+include ../Makefile.tests_common
 include $(RIOTBASE)/Makefile.include
--- a/tests/gnrc_sock_dns/Makefile
+++ b/tests/gnrc_sock_dns/Makefile
@@ -1,5 +1,3 @@
-include ../Makefile.tests_common
-
 RIOTBASE ?= $(CURDIR)/../..
 
 BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-mega2560 arduino-uno \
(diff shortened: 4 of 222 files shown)


tests/bench_timers/Makefile
```diff
@@ -53,4 +51,5 @@ test-kinetis-lptmr: all
 # Reset the default goal.
 .DEFAULT_GOAL :=
 
+include ../Makefile.tests_common
```
cladmi: This one should also be before the 

```
ifeq (,$(findstring TIM_TEST_DEV,$(CFLAGS)))
```

This handling is already a bit dirty so would not touch it with this PR.

tests/emb6/Makefile
```diff
@@ -41,4 +40,5 @@ USEMODULE += $(DRIVER)
 CFLAGS += -Wno-unused-parameter -Wno-unused-function -Wno-type-limits
 CFLAGS += -Wno-sign-compare -Wno-missing-field-initializers
 
+include ../Makefile.tests_common
```
cladmi: Same here as `tests_at86rf2xx`, we are using `BOARD` for configuring so would still prefer keeping the include before. I know a default value is defined, but I would prefer to keep it coherent.

tests/float/Makefile
```diff
@@ -9,4 +7,5 @@ endif
 
 TEST_ON_CI_WHITELIST += all
 
+include ../Makefile.tests_common
```
cladmi: Here also we use `BOARD`, a conditional definition of `CFLAGS` could be done, or just changing the C file would even be better…

tests/gnrc_sock_dns/Makefile
```diff
@@ -26,4 +24,5 @@ ifeq ($(BOARD),$(filter $(BOARD),$(LOW_MEMORY_BOARDS)))
               -DNRC_IPV6_NIB_OFFL_NUMOF=1
 endif
 
+include ../Makefile.tests_common
```
cladmi: I would also put it before the `ifeq ($(BOARD)` to not use before defining.

tests/libfixmath_unittests/Makefile
```diff
@@ -19,4 +17,5 @@ USEMODULE += printf_float
 
 TEST_ON_CI_WHITELIST += native
 
+include ../Makefile.tests_common
```
cladmi: Also using `BOARDS` before the include here.

The LINKFLAGS could be defined by a `LINKFLAGS += $(if $(filter native,$(BOARD)),-lm)`.

tests/pkg_fatfs/Makefile
```diff
@@ -50,4 +48,5 @@ image:
 compressed-image:
 	@./create_fat_image_file.sh $(FATFS_IMAGE_FILE_SIZE_MIB)
 
+include ../Makefile.tests_common
```
cladmi: I would also put this one before the `ifeq ($(BOARD),native)` even if the default is already set to `native`.

tests/pkg_fatfs_vfs/Makefile
```diff
@@ -49,6 +48,7 @@ BOARD_WHITELIST := airfy-beacon arduino-due arduino-duemilanove arduino-mega2560
 
 TEST_DEPS += image
 
+include ../Makefile.tests_common
```
cladmi: Also this one before `ifeq ($(BOARD),native)` and even more as no default value is set here.

tests/pkg_microcoap/Makefile
```diff
@@ -36,4 +34,5 @@ ifneq (,$(filter $(BOARD),$(LOW_MEMORY_BOARDS)))
   USEMODULE += prng_minstd
 endif
 
+include ../Makefile.tests_common
```
cladmi: This one is also using `BOARD`, so move it before.

tests/thread_cooperation/Makefile
```diff
@@ -22,4 +20,5 @@ CFLAGS += -DPROBLEM=$(PROBLEM)
 
 TEST_ON_CI_WHITELIST += all
 
+include ../Makefile.tests_common
```
cladmi: This one should also go before using `$(BOARD)` to be consistent.

tests/unittests/Makefile
```diff
@@ -253,6 +252,7 @@ INCLUDES += -I$(RIOTBASE)/tests/unittests/common
 # list of boards to run CI tests on
 TEST_ON_CI_WHITELIST += all
 
+include ../Makefile.tests_common
```
cladmi: I would keep the `unittest` Makefile untouched, it is a special one where we do not re-define `BOARD` anyway in the Makefile.
Also lot of things are done using `BOARD` before here.

PR #10124, opened 2018-10-07
pkg/semtech-loramac: switch to RIOT AES

olegart:
### Contribution description

Patch to adapt LoRaMAC 4.4.1 to RIOT AES implementation.

1) Code size: RIOT AES significantly bigger without #10037, ~1KB bigger with #10037, about the same size as LoRaMAC AES with #10037 and aes_decrypt disabled (LoRaWAN doesn't need it)

2) Performance: generally, RIOT AES is significantly (over 2 times) faster, especially on Cortex-M3 and above. On Cortex-M0, decryption is about the same speed, encryption is almost 2 times faster with RIOT AES.

Such difference is to be expected, as RIOT uses faster T-tables implementation while LoRaMAC uses S-boxes.

3) Compatibility: LoRaMAC AES and RIOT AES can't be used simultaneously because of the same function names aes_encrypt() and aes_decrypt(). So, **porting LoRaMAC to RIOT AES is a must if one needs AES for any other function, not just LoRaWAN**.

### Testing procedure

Build examples/lorawan, flash, connect.

Tested on custom STM32L0/SX1276 hardware and Er-Telecom/Actility network. Works fine. Test was done on a slightly different semtech-loramac code and patches, not exactly from the RIOT repo, but this should be of no significance.

### DIFF
--- a/pkg/semtech-loramac/Makefile.include
+++ b/pkg/semtech-loramac/Makefile.include
@@ -1,3 +1,5 @@
+USEMODULE += crypto
+
 INCLUDES += -I$(RIOTBASE)/pkg/semtech-loramac/include
 
 DIRS += $(RIOTBASE)/pkg/semtech-loramac/contrib
--- a/pkg/semtech-loramac/patches/0004-switch-to-riot-aes.patch
+++ b/pkg/semtech-loramac/patches/0004-switch-to-riot-aes.patch
@@ -1,0 +1,206 @@
+From fa275bbb8259fd3ddc9960f85e94bd444e1034d3 Mon Sep 17 00:00:00 2001
+From: Oleg Artamonov <oleg@unwds.com>
+Date: Sun, 7 Oct 2018 21:37:45 +0300
+Subject: [PATCH] Switch to RIOT AES implementation
+
+---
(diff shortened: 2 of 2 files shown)


jia200x:
@olegart the changes makes sense to me. I will test them, do a deeper review and report back.

miri64:
The dependency to `crypto` has to be made optional though (by using `#ifdef MODULE_CRYPTO` around your changes) or added explicitly to `pkg/semtech-loramac/Makefile.dep`. Otherwise the application always has to have `USEMODULE += crypto` added manually.

olegart:
Yep. Added USEMODULE += crypto to pkg's Makefile, also changed patch a bit to fix it for RIOT's LoRaMAC version.

jia200x:
IIRC we also need [`-DCRYPTO_AES`](http://doc.riot-os.org/group__sys__crypto.html) to the CFLAGS.

PeterKietzmann:
@jia200x according to the labels you set it's only testing missing. Can you provide that?

tcschmidt:
@jia200x can you test/review integration?

kYc0o:
I've tested this PR with both TTN and ChirpStack, join procedure fails. I didn't dig that much yet on the reason, but current master succeeds on both.

PR #10125, opened 2018-10-08
cpu/esp8266: cpu frequency settings feature added

gschorcht:
### Contribution description

With this PR it becomes possible to set the default CPU frequency for the non-SDK and the SDK-Version as well. For that purpose,

- the define ```ESP8266_CPU_FREQUENCY``` is introduced, and
- the function ```system_update_cpu_freq``` is implemented in non-SDK version.

Using ```-DESP8266_CPU_FREQENCY=<freq>``` in CFLAGS where ```<freq>``` can be 80 or 160 is used to define the CPU frequency to be 80 MHz or 160 MHz. If ```ESP8266_CPU_FREQENCY``` is not defined, ```ESP8266_CPU_FREQUENCY_DEFAULT``` is used which is defined to be 80 MHz.

### Testing procedure

Using
```
make flash BOARD=esp8266-esp-12x -C tests/bench_thread_yield_pingpong
```
compiles and flashes the benchmark test with the default CPU frequency of 80 MHz. Using command
```
CFLAGS="-DESP8266_CPU_FREQUENCY=160" make flash BOARD=esp8266-esp-12x -C tests/bench_thread_yield_pingpong
```
compiles and flashes the benchmark test with a CPU frequency of 160 MHz. The result of the benchmark test should be the double.

### DIFF
--- a/cpu/esp8266/include/user_config.h
+++ b/cpu/esp8266/include/user_config.h
@@ -5,6 +5,13 @@
  * General Public License v2.1. See the file LICENSE in the top level
  * directory for more details.
  */
+/**
+ * @ingroup     cpu_esp8266
+ * @brief       Default configurations required by the SDK
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ * @{
+ */
 
 #ifndef USER_CONFIG_H
 #define USER_CONFIG_H
@@ -13,10 +20,17 @@
 extern "C" {
 #endif
 
-/* This file is just to satisfy SDK */
+/**
+ * @brief   Default CPU frequency in MHz.
+ * Possible values are 80 and 160.
+ */
+#ifndef ESP8266_CPU_FREQUENCY
+#define ESP8266_CPU_FREQUENCY   80
+#endif
 
 #ifdef __cplusplus
 }
 #endif
 
 #endif /* USER_CONFIG_H */
+/** @} */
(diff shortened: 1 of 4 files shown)


cpu/esp8266/include/user_config.h
```diff
@@ -13,10 +20,15 @@
 extern "C" {
 #endif
 
-/* This file is just to satisfy SDK */
+/**
+ * @brief   Default CPU frequency in MHz.
+ * Possible values are 80 and 160.
+ */
+#define ESP8266_CPU_FREQUENCY_DEFAULT   80
```
leandrolanzieri: Why do you define a default and then check if `ESP8266_CPU_FREQUENCY` is defined instead of defining it like..?
```C
#ifndef ESP8266_CPU_FREQUENCY
#define ESP8266_CPU_FREQUENCY   80
#endif
```
That way you wouldn't need the `#ifdef...` on the `user_init` function. Is there something I'm missing?

MrKevinWeiss:
ping @gschorcht , Leandro does make a decent point (unless it has something to do with API's).  I can review and test if you either give a reason or make the change.

MrKevinWeiss:
I do think @leandrolanzieri solution is better but if you feel strongly that the clarity is worth the extra lines then I can accept it.

leandrolanzieri:
I just think that it's a common and known pattern in RIOT to define defaults as `#ifndef`s.

gschorcht:
Changed it.

MrKevinWeiss:
Thanks!

gschorcht:
squashed.

PR #10126, opened 2018-10-08
cpu/esp8266: doc extended

gschorcht:
### Contribution description

Documentation in ```cpu/esp8266/doc.txt``` was extended by section ```Erasing the Device``` which describes
- how to erase the flash of the device, and
- more important what to do after erasing the flash to get the device working again.

### Testing procedure

It's just documentation.


gschorcht:
@jia200x The disadvantage of handling board documentation as part of the code and as normal PRs is that it takes a lot of time before the changes become active. This PR is opened for almost 3 months.

Another problem I'm running into is that it becomes more and more difficult to merge the changes of this PR with changes of the documentation I had to make as part of other PRs :worried:

cpu/esp8266/README.md
```diff
@@ -347,11 +347,23 @@ Optional features of ESP8266 can be enabled by ```USEMODULE``` definitions in th
...
-esp_spiffs | Enables the SPIFFS file system, see section [SPIFFS Device](#esp8266_spiffs_device)
-esp_sw_timer | Enables software timer implementation, implies the setting ```USE_SDK=1```, see section [Timers](#esp8266_timers)
+[esp_gdb](#esp8266_qemu_mode_and_gdb) | Enable the compilation with debug information, which is equivalent to using ```ENABLE_GDB=1```
+[esp_sdk](#esp8266_sdk_task_handling) | Enable the SDK version, which is equivalent to using ```USE_SDK=1```
+[esp_spiffs](#esp8266_spiffs_device) | Enable the SPIFFS drive in on-board flash memory
+[esp_sw_timer](#esp8266_timers) | Enable software timer implementation, implies the setting ```USE_SDK=1``` (module ```esp_sdk```)
 
 </center><br>
 
+For example, to activate the a SPIFFS drive in on-board flash memory, the makefile of application has simply to add the ```esp_spiffs``` module to ```USEMODULE``` make variable:
```
aabadie: There's an extra `a`. I think it should be `to activate the SPIFFS drive`

cpu/esp8266/README.md
```diff
@@ -347,11 +347,23 @@ Optional features of ESP8266 can be enabled by ```USEMODULE``` definitions in th
...
+```
+USEMODULE += esp_spiffs
+```
+
+Modules can be also be activated temporarily at the command line when calling the make command:
```
aabadie: Extra `be`: `Modules can also be activated...`

cpu/esp8266/README.md
```diff
@@ -613,22 +636,19 @@ SDCARD_SPI_PARAM_CS    | SPI0_CS0_GPIO | can be overridden
...
 # <a name="esp8266_application_specific_configurations"> Application-Specific Configurations </a> &nbsp;[[TOC](#esp8266_toc)]
 
-Configuration used for peripheral devices and for device driver modules, such as GPIO pins, bus interfaces or bus speeds are typically defined in the board-specific configurations ```board.h``` and ```periph_conf.h``` or in the driver parameter configuration ```< driver>_params.h```. Most of these definitions are enclosed by
-```
-#ifndef ANY_PARAMETER
-...
-#endif
-```
-constructs, so it is possible to override them by defining them in front of these constructs.
+The board-specific configuration files ```board.h``` and ```periph_conf.h``` as well well as the driver parameter configuration files ```<driver>_params.h``` define the default configurations for peripherals and device driver modules. These are, for example, the GPIOs used, bus interfaces used or available bus speeds. Because there are many possible configurations and many different application requirements, these default configurations are usually only a compromise between different requirements.
```
aabadie: Extra `well`

aabadie: And I think this line could be split, but this comment would apply to large parts of this file.

cpu/esp8266/doc.txt
```diff
@@ -339,24 +340,35 @@ Option | Values | Default | Description
...
-esp_spiffs | Enables the SPIFFS file system, see section [SPIFFS Device](#esp8266_spiffs_device)
-esp_sw_timer | Enables software timer implementation, implies the setting ```USE_SDK=1```, see section [Timers](#esp8266_timers)
+[esp_gdb](#esp8266_qemu_mode_and_gdb) | Enable the compilation with debug information, which is equivalent to using ```ENABLE_GDB=1```
+[esp_sdk](#esp8266_sdk_task_handling) | Enable the SDK version, which is equivalent to using ```USE_SDK=1```
+[esp_spiffs](#esp8266_spiffs_device) | Enable the SPIFFS drive in on-board flash memory
+[esp_sw_timer](#esp8266_timers) | Enable software timer implementation, implies the setting ```USE_SDK=1``` (module ```esp_sdk```)
 
 </center><br>
 
+For example, to activate the a SPIFFS drive in on-board flash memory, the makefile of application has simply to add the ```esp_spiffs``` module to ```USEMODULE``` make variable:
```
aabadie: Extra `a` copy pasted from the README.md. Just a thought: can we just keep one of these files ? (doc.txt)

gschorcht: `README.md` and `doc.txt` are identical. I use `README.md` to write the documentation to have immediate feedback in markdown viewer. It is quite tedious to write complex documentation for `doxygen` and execute `make doc` to see the results.

Using `doc.txt` is the new approach for board documentation. So, if only one file should be kept, it should be `doc.txt` :worried:

cpu/esp8266/README.md
gschorcht: I remember to out discussion regarding 80 characters per line also for documentation. These files were written before. I you insist on it, I can split the lines.

gschorcht:
@aabadie Many thanks for your review.

cpu/esp8266/doc.txt
gschorcht: I kept `README.md` for the moment so that you can track requested changes. I will remove it before squashing.

gschorcht:
@aabadie What do you think, should I remove README.md first? Yes, it is difficult to keep them consistent.

aabadie:
> should I remove README.md first? Yes, it is difficult to keep them consistent.

That would be better indeed.

gschorcht:
@aabadie  I would say that, the Codacy problem is a false positive. Can we merge inspite of it?

aabadie:
Just in case @gschorcht, your last commit (71ff08216f2ded8628f71aa45c8c8aec503cf31f) is removing the wrong README.md (in esp32 instead of esp8266) :)

The Codacy failure should be tackled if you fix this ;)

gschorcht:
@aabadie Ups :sunglasses: Fixed.

aabadie:
All green, let's merge!

gschorcht:
@aabadie Thanks for your review.

PR #10128, opened 2018-10-08
sys/ubjson: Unconditionally include <sys/types.h>

gebart:
### Contribution description

msp430_types.h is included implicitly if needed by sys/types.h
native Linux also needs sys/types.h for ssize_t, no idea why it was excluded originally..
This PR is necessary in order to build on native without `-std=gnu99`

### Testing procedure

Run unit tests (tests-ubjson)


kaspar030:
(If it still builds, it should be OK)

PR #10130, opened 2018-10-08
[WIP] Apply automatic board list formatting from #10111 to Makefiles

x3ro:
### Contribution description

Applies the formatting mechanism from #10111 to all relevant Makefiles. Still marked as WIP since there are some corner cases to be fixed.

### DIFF
--- a/tests/mpu_stack_guard/Makefile
+++ b/tests/mpu_stack_guard/Makefile
@@ -1,48 +1,196 @@
 BOARD ?= samr21-xpro
 include ../Makefile.tests_common
 
-BOARD_WHITELIST += arduino-due      # cortex-m3
-BOARD_WHITELIST += arduino-zero     # cortex-m0plus
-BOARD_WHITELIST += cc2538dk         # cortex-m3
-BOARD_WHITELIST += cc2650stk        # cortex-m3
-BOARD_WHITELIST += ek-lm4f120xl     # cortex-m4f
-BOARD_WHITELIST += f4vi1            # cortex-m4f
-BOARD_WHITELIST += fox              # cortex-m3
-BOARD_WHITELIST += frdm-k64f        # cortex-m4
-BOARD_WHITELIST += ikea-tradfri     # cortex-m4f
-BOARD_WHITELIST += iotlab-a8-m3     # cortex-m3
-BOARD_WHITELIST += iotlab-m3        # cortex-m3
-BOARD_WHITELIST += limifrog-v1      # cortex-m3
-BOARD_WHITELIST += maple-mini       # cortex-m3
-BOARD_WHITELIST += mbed_lpc1768     # cortex-m3
-BOARD_WHITELIST += msbiot           # cortex-m4f
-BOARD_WHITELIST += mulle            # cortex-m4
-BOARD_WHITELIST += nrf52dk          # cortex-m4f
-BOARD_WHITELIST += nucleo-f103rb    # cortex-m3
-BOARD_WHITELIST += nucleo-f207zg    # cortex-m3
-BOARD_WHITELIST += nucleo-f303re    # cortex-m4f
-BOARD_WHITELIST += nucleo-f334r8    # cortex-m4f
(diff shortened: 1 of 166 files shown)


kaspar030:
am I the only one who's bothered by having to scroll down first thing after opening a test's Makefile?

tests/socket_zep/Makefile
```diff
@@ -1,7 +1,14 @@
...
 
-BOARD_WHITELIST = native    # socket_zep is only available on native
+BOARD_WHITELIST = # \
+                  available \
+                  is \
+                  native \
+                  native \
+                  on \
+                  only \
+                  socket_zep
```
gebart: Oops!

gebart:
@kaspar030 
> am I the only one who's bothered by having to scroll down first thing after opening a test's Makefile?

Not sure what you mean? Do you mean that the board blacklist/whitelists are too long and you have to scroll down to see the USEMODULE etc lines in the Makefile?

kaspar030:
> Do you mean that the board blacklist/whitelists are too long and you have to scroll down to see the USEMODULE etc lines in the Makefile?

Yes...

tests/socket_zep/Makefile
x3ro: See #10111

x3ro:
> am I the only one who's bothered by having to scroll down first thing after opening a test's Makefile?

So far yes

waehlisch:
You can just search: ```/USEMODULE```

kaspar030:
> You can just search: `/USEMODULE`

Yup. Everytime I open the file, which might be hundred times more often than I modify the lists. Multiplied by the number of people that read at our code, that is quite often.

To me it seems like the board lists get much more screen real estate as they are relevant.
See [this](https://github.com/x3ro/RIOT/blob/pr/consistently-format-board-lists/examples/dtls-echo/Makefile) vs [this](https://github.com/RIOT-OS/RIOT/blob/master/examples/dtls-echo/Makefile).

If there's a script that does the formatting anyways, why not have that output lines that fill 80 characters?

MrKevinWeiss:
Maybe we just move the BOARD_* below the USEMODULE?

kaspar030:
> Maybe we just move the BOARD_* below the USEMODULE?

Yeah, why not? as far down as possible.

x3ro:
> If there's a script that does the formatting anyways, why not have that output lines that fill 80 characters?

To make diffs clean, relevant and reviewable.

> Maybe we just move the BOARD_* below the USEMODULE?

Seems doable

A-Paul:
What about including from (a) separate files(s)?

x3ro:
> What about including from (a) separate files(s)?

Could you elaborate how this would improve the situation over putting the lists at the end of the file?

tests/mpu_stack_guard/Makefile
```diff
@@ -1,48 +1,196 @@
...
 
 USEMODULE += mpu_stack_guard
 
-include $(RIOTBASE)/Makefile.include
-
 ifeq (llvm,$(TOOLCHAIN))
   CFLAGS += -Wno-infinite-recursion
 endif
+
+BOARD_WHITELIST := \
```
MrKevinWeiss: Maybe manually alter this so it doesn't look so silly?  I don't it is worth the effort adding this corner case into the script.

tests/socket_zep/Makefile
```diff
@@ -12,4 +10,14 @@ CFLAGS += -DDEVELHELP
 
 TERMFLAGS ?= -z [::1]:17754
 
+BOARD_WHITELIST := \
```
MrKevinWeiss: Also manually fix this guy.

MrKevinWeiss:
Darn...

MrKevinWeiss:
Maybe see if the updated script works on this.  If there are only a few corner cases that go against convention I think manual fixes are fine.

x3ro:
Closing this. I'd prefer to open a new PR with a clean discussion.

Issue #10122, opened 2018-10-07
Indeterministic hard fault in _mutex_lock(), with nRF52 SoftDevice


pekkanikander:
#### Description

With https://github.com/RIOT-OS/RIOT/pull/9473, with the updated support for nRF52 SoftDevice, the way interrupts are re-enabled in `_mutex_unlock` seems to cause indeterministic hard faults.

#### Steps to reproduce the issue

The problem is hard to reproduce.  We saw it a few weeks ago when working on our own nRF52 board with relatively heavy I/O and BlueTooth load.  Applying the change proposed below increased the stability a lot, but not completely.  (There seems to be a thread deadlock issue triggered in `_xtimer_tsleep`, also related to `_mutex_lock`, but that is not clear enough to me even to describe.)

I will try to reproduce this problem with nRF52 DK so that others can debug this too, but I think it is better to document this now and start some discussion already before that.  Hence this issue.

#### Expected results

RIOT should not cause hard faults.

#### Actual results

Indeterministic crashes, with hard fault triggered on line 62 in `mutex.c`:
https://github.com/RIOT-OS/RIOT/blob/master/core/mutex.c#L62

#### Versions

arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2017-q4-major) 7.2.1 20170904 (release) [ARM/embedded-7-branch revision 255204]

#### Suggested fix

Change the order in which interrupts are enabled again and higher priority threads are allowed to run:

```
   --- a/core/mutex.c
   +++ b/core/mutex.c
   @@ -58,8 +58,8 @@ int _mutex_lock(mutex_t *mutex, int blocking)
            else {
                thread_add_to_list(&mutex->queue, me);
            }
   -        irq_restore(irqstate);
            thread_yield_higher();
   +        irq_restore(irqstate);
            /* We were woken up by scheduler. Waker removed us from queue.
             * We have the mutex now. */
            return 1;
```

#### Discussion

For Cortex-M series, `thread_yield_higher()` simply triggers the `PendSV` interrupt:
https://github.com/RIOT-OS/RIOT/blob/master/cpu/cortexm_common/thread_arch.c#L291

In RIOT, if a thread called `_mutex_lock`, the PendSV interrupt will take place immediately, since by default all interrupts are on the same priority level, see https://github.com/RIOT-OS/RIOT/pull/3511

[...]

PR #10137, opened 2018-10-09
cpu/cortexm_common: Add support for Cortex-M interrupt sub-priorities

pekkanikander:
### Contribution description

This commit enables Cortex-M CPU interrupt sub-priorities and allows the PendSV interrupt to have a priority different from the default one.  Together these two new preprocessor defines can be used to have PendSV always run as the last interrupt, before returning from the interrupt stack back to the user space.

Running PendSV as the last interrupt before returning to the user space is recommended by ARM, as it increases efficiency.  Furthermore, that change enhances stability a lot with the new nRF52 SoftDevice support, currently being worked in PR #9473.

This commit merely enables sub-priorities and a separate PendSV priority to be used without changing the default RIOT behaviour.

### Testing procedure

As such, this should not change anything.  All the previous tests should run as before, and no new ones are introduced.

### Issues/PRs references

See also #9473 and https://github.com/RIOT-OS/RIOT/issues/10122


gebart:
@pekkanikander Thanks for working on this!

I did not look into the code yet, but does this work the same way on all Cortex-M CPUs, even M0/M0+? Does it depend on any optional features in the NVIC or are all vendors the same here?

Background info: There are some features in other parts of the CPU which are mandatory on the CM3 and up, but optional on the CM0.

pekkanikander:
@gebart Unfortunately I am not a definite expert here, but AFAIK and as far as I was able to check quickly, the patch is ok for all Cortex-M CPUs.  

Apparently Cortex-M0 doesn't have this feature, while Cortex-M3 and M4 do.  However, Cortex-M0 CMSIS does not implement `NVIC_SetPriorityGrouping`, so setting this on Cortex-M0 should lead to a compile-time error (though I haven't tested that.)

http://infocenter.arm.com/help/topic/com.arm.doc.dui0553a/Cihehdge.html
http://infocenter.arm.com/help/topic/com.arm.doc.dui0552a/Cihehdge.html
http://infocenter.arm.com/help/topic/com.arm.doc.dui0497a/Cihehdge.html

https://embeddedgurus.com/state-space/2014/02/cutting-through-the-confusion-with-arm-cortex-m-interrupt-priorities/

gebart:
This looks ready. I will test and review tonight or tomorrow

gebart:
Looks good I added one comment on the wording. Accidentally commented on the commit instead of on the PR, but it is visible above this post

pekkanikander:
Changed the wording as suggested and squashed.

pekkanikander:
What should be done with this?  No interest from RIOT side and so close?  Or someone should rebase?  (I don't have any recent RIOT tree on my laptop, since I haven't been working with RIOT for a while.)

benpicco:
@pekkanikander this has already been approved, but there is a merge conflict that needs to be resolved before it can be merged.

benpicco:
No change in default behavior, so no reason why we shouldn't have that option.

pekkanikander:
@benpicco Thanks for taking care of this!  (I don't have RIOT build environment, so I couldn't have been able to rebase or test.)

PR #10139, opened 2018-10-10
net/rdcli_simple: do not send CF opt in POST msg

haukepetersen:
### Contribution description
Through the resource directory interop today @chrysn discovered a bug in the simple registration endpoint implementation: For the initial `POST` message, the content format option should not be set. This PR fixes that.

### Testing procedure
Run the `examples/rdcli_simple` example and observe the traffic in wireshark. Without this PR, the content format option in the initial `POST` message is set. With this PR, this option is not set anymore.

### Issues/PRs references
none


haukepetersen:
@bergzand would you mind to give this a quick review/test today? I would be nice to have this merged as there will be another CoRE RD interop test tomorrow afternoon...

If you don't have time, I can find someone else :-)

PR #10147, opened 2018-10-11
testing: add periph I2C robot tests for HIL

smlng:
### Contribution description

This adds a new test suite for the I2C peripheral drivers based on the RobotFramework and PHiLIP to allow for hardware-in-the-loop (HIL) testing.

### Testing procedure

To run this you need a board (device under test) hooked up to a PHILIP test board, see https://github.com/riot-appstore/PHiLIP . Then run `BOARD=<fill in> make -C tests/periph_i2c flash robot-test`.

### Issues/PRs references

depends on #10095


MrKevinWeiss:
I ran the test on the nrf52dk (that supports reset on p21) with PHiLIP-N and all tests passed but `make robot-open` seems to have some problems with it.
When I tried everything together I received:
`open: Unable to open /dev/tty2: Permission denied`

When I tried with sudo and just `robot-open` I got:
`objcopy not found. Hex file will not be created.`

smlng:
I think I remove the `robot-open` target, it was intended as a short-hand to open the generated html output and on macOS it works nicely, but Linux seems to have a different meaning of `open`.

tests/periph_i2c/tests/periph_i2c.keywords.txt
```diff
@@ -0,0 +1,18 @@
...
+Library             I2Cdevice  port=%{PORT}  baudrate=%{BAUD}  timeout=${10}
+
+Resource            riot_base.keywords.txt
+Resource            riot_util.keywords.txt
+Resource            api_shell.keywords.txt
+Resource            philip.keywords.txt
+
+*** Keywords ***
+DUT Must Have Periph I2C Application
+    API Call Should Succeed     I2C Get ID
```
MrKevinWeiss: To prevent initial error from connecting and resetting I think it would be good to send the I2C Get ID command once without any checks first, then send and check the id.  Either include it here or do it before calling this keyword.

tests/periph_i2c/tests/02__periph_i2c_write_register.robot
```diff
@@ -0,0 +1,17 @@
+*** Settings ***
+Documentation       Data driven tests to verify writing to a valid register.
+
+Suite Setup         Run Keywords    Reset DUT and PHILIP
+...                                 DUT Must Have Periph I2C Application
+Test Setup          Reset DUT and PHILIP
```
MrKevinWeiss: I think each test should include a DUT Must Have Periph I2C Application or at least send the I2C Get ID without checks.  Maybe make a clear buffers command that does that.

MrKevinWeiss:
also add the riot_pal version to the metadata

MrKevinWeiss:
... Also grep for verfiy, I think there are a few places that this occurs.

tests/periph_i2c/tests/periph_i2c.keywords.txt
smlng: okay, will add a simple `I2C Get ID` without checking the result before this line

tests/periph_i2c/tests/02__periph_i2c_write_register.robot
smlng: so basically Test Setup should be the same as Suite Setup

smlng:
> Also I think it would be good for all data driven tests to have a acquire and release around it (maybe this can be done with the setup and teardown)

yeah Test Setup and Test Teardown would be the best way, so we do not clutter the actual test

smlng:
> ... Also grep for verfiy, I think there are a few places that this occurs.

what do you mean?

smlng:
> hmmm, my suggestions still don't solve the remote-revb problems. Maybe a retry command n times would be better. Sometimes it takes 2 timeouts after the reset to recover.

you mean `I2C Get ID` should be called multiple times, before running the test? I guess then it would be better to have something like: *retry as long as it times out, but max n times*

MrKevinWeiss:
Very nice.  I like the changes and it seems to improve the non i2c related issues.

kaspar030:
Would you mind creating a version that does not depend on robot? No need to stall HIL testing for a framwork discussion.

tcschmidt:
> 
> 
> Would you mind creating a version that does not depend on robot? No need to stall HIL testing for a framework discussion.

No issue here: HIL testing is deployed and working. No need to intermingle with the asynchronous framework discussion.

kaspar030:
> HIL testing is deployed and working. No need to intermingle with the asynchronous framework discussion.

Then we can close this?

MrKevinWeiss:
@kaspar030 please don't, at the moment we are using this for the HiL results in the nighties (and I have been using it for i2c improvements that can use some review :wink:)

MrKevinWeiss:
> Would you mind creating a version that does not depend on robot?

Checkout the tests/periph_i2c/tests/test.py (it doesn't use test runner and was meant as an example of what should be tested but it is an independent test).

kaspar030:
> please don't, at the moment we are using this for the HiL results in the nighties

I know, and I didn't intent to. I tried to clarify what @tcschmidt means by "this is deployed and working" vs. there's an open PR.

kaspar030:
> please don't, at the moment we are using this for the HiL results in the nighties

I know, and I didn't intent to. I tried to clarify what @tcschmidt means by "this is deployed and working" vs. there's an open PR.

tcschmidt:
@kaspar030 the point is simple: Framework discussion is asynchronous, nothing more.

smlng:
closing in favour of #10774

PR #10148, opened 2018-10-11
drivers: Add support for tsl4531x visible light intensity sensor

danpetry:
### Contribution description

Driver for TSL4531x range of light intensity sensors.

This device is used on the SenseBox Light Sensor.

The driver includes two power modes: low and high. In low power mode, the sensor switches off after a reading, and a read will wake the sensor up to perform a single integration cycle, after which the sensor will go back to sleep. The integration cycle is of the order of several hundred milliseconds, and so this read can be performed in a syncronous or asynchronous manner. More details are in the API documentation in tsl4531x.h.

A general description, including protocol format, can be viewed [here](https://ams.com/documents/20143/36005/TSL4531_DS000182_3-00.pdf), and a datasheet [here](https://media.digikey.com/pdf/Data%20Sheets/Austriamicrosystems%20PDFs/TSL4531.pdf).

The implementation also includes SAUL support.

### Testing procedure

To test the driver, please compile tests/driver_tsl4531x for a board which has this sensor. 

To test SAUL integration, please compile examples/saul for a board which has this sensor. If you are testing using the SenseBox, `USEMODULE += tsl4531x` will need to be added in the Makefile.

Switching between low and high power modes has also been tested qualitatively by measuring the current draw, using an oscilloscope probe connected across a 10K resistor. The voltage across the resistor was observed to switch between high and low levels as the power modes changed.

This testing has only been performed on the tsl45315, in absence of the other devices in the range. However, they vary only in the I2C address and I2C Vbus, as indicated on the datasheet.

### Acknowledgements

This work is based on @aabadie 's tsl2561 driver and @jcarrano 's further work in adapting this to the tsl4531x range.


drivers/include/tsl4531x.h
```diff
@@ -0,0 +1,178 @@
...
+ * @brief Integration times
+ */
+typedef enum {
+    TSL4531X_INTEGRATE_400MS = 0,       /* 0b00 */
+    TSL4531X_INTEGRATE_200MS = 1,       /* 0b01 */
+    TSL4531X_INTEGRATE_100MS = 2,       /* 0b10 */
+}tsl4531x_intgn_time_t;
+
+/**
+ * @brief Part numbers for different members of the TSL4531x series
```
leandrolanzieri: Maybe adding a `@name` to group these and the following?

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,180 @@
...
+    if (dev->low_power_mode) {
+        i2c_write_reg(dev->i2c_dev,
+                      dev->i2c_addr,
+                      TSL4531X_COMMAND(TSL4531X_CONTROL_REG),
+                      TSL4531X_CONTROL(TSL4531X_MODE_SINGLE_ADC_CYCLE),
+                      0);
+
+    }
+
+    /* In low power mode, or just after changing to high power mode, this
```
leandrolanzieri: Is it needed to way just after changing to high power mode?

drivers/include/tsl4531x.h
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * @return     Zero on success
+ * @return     -ENODEV if I2C bus can't be acquired
+ * @return     -ENXIO if device can't be read or configured
+ * @return     -ENOTSUP if ID, once read, is wrong
+ */
+int tsl4531x_init(tsl4531x_t *dev, const tsl4531x_params_t *params);
+
+/**
+ * Set the power mode of the driver.
```
leandrolanzieri: As you pass an `ON / OFF` value maybe say _Set low power mode on or off_.

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,180 @@
...
+    dev->low_power_mode = params->low_power_mode;
+    dev->sample_start_time = 0; /* Device assumed to start up at same time as
+                                   microcontroller - i.e. when it hits this line */
+
+    return 0;
+}
+
+int tsl4531x_set_low_power_mode(tsl4531x_t *dev, bool low_power_mode)
+{
+    assert(dev);
```
jcarrano: You are missing [i2c_acqure()](http://doc.riot-os.org/group__drivers__periph__i2c.html#ga01056db645f8e561ac8bdab1f531a6db)/i2c_release() in all these functions

drivers/tsl4531x/tsl4531x.c
danpetry: Yes, because an integration cycle needs to be performed before reading a value if the driver was previously off

drivers/tsl4531x/include/tsl4531x_params.h
```diff
@@ -0,0 +1,84 @@
...
+#endif
+#ifndef TSL4531X_LOW_POWER_MODE
+#define TSL4531X_LOW_POWER_MODE        (false)
+#endif
+#ifndef TSL4531X_PARAMS
+#define TSL4531X_PARAMS                { .i2c_dev     = TSL4531X_PARAM_I2C_DEV, \
+                                         .i2c_addr    = TSL4531X_PARAM_I2C_ADDR, \
+                                         .integration_time = TSL4531X_PARAM_INTEGRATION, \
+                                         .low_power_mode = TSL4531X_LOW_POWER_MODE, \
+                                         .part_number = TSL45315_PARTNO }
```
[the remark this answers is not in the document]
danpetry: yes, well caught

drivers/include/tsl4531x.h
```diff
@@ -0,0 +1,178 @@
...
+ * @param      dev           Initialized device descriptor
+ * @param      low_power_on  Bool indicating whether low power mode is on or off
+ *
+ * @return     Zero
+ */
+int tsl4531x_set_low_power_mode(tsl4531x_t *dev, bool low_power_on);
+
+/**
+ * Start collecting sample in low power mode. This provides asynchronous operation
+ * along with tsl4531x_is_sample_ready and tsl4531x_get_sample. It should not be
```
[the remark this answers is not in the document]
danpetry: Sounds good to me. As discussed offline, this would allow people to write the following code that would work in both modes:

```
tsl4531x_start_sample(&dev);
sleep(tsl4531x_is_sample_ready(&dev));
tsl4531x_get_sample(&dev);
```
And can allow the power mode to be changed elsewhere.

drivers/include/tsl4531x.h
```diff
@@ -0,0 +1,178 @@
...
+ * with that function and tsl4531x_get_sample, provides asynchronous operation.
+ * In high power mode, this indicates whether adequate time has passed to collect
+ * a full sample since the driver was switched into high power mode or started up.
+ *
+ * The waiting time depends on the integration time set in the device initialisation
+ * parameters, plus 5% margin to encompass the max times indicated in the datasheet.
+ *
+ * @param      dev           Initialized device descriptor
+ *
+ * @return     Bool indicating whether sample is ready to be read
```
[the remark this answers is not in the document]
danpetry: Yes good idea, will do

drivers/include/tsl4531x.h
```diff
@@ -0,0 +1,178 @@
...
+bool tsl4531x_is_sample_ready(const tsl4531x_t *dev);
+
+/**
+ * Reads the sample from the device immediately. In high power mode, this does
+ * the same as tsl4531x_simple_read. In low power mode, this provides asynchronous
+ * operation along with tsl4531x_start_sample and tsl4531x_is_sample_ready.
+ *
+ * @param      dev           Initialized device descriptor
+ *
+ * @return     The sample value in Lux
```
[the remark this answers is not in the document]
danpetry: Yes will do

drivers/tsl4531x/include/tsl4531x_internals.h
```diff
@@ -0,0 +1,104 @@
...
+
+/* This allows either full power mode or power down mode. */
+#define TSL4531X_POWER_MODE(mode) ((!mode) * (0x03))
+
+/* The full integration time is a sum of:
+ * 1. A worst-case time for each of the integration times, which according to
+ *    the datasheet is 5% more than the nominal time
+ * 2. 60, 30 or 15 ms, according to the integration time, if PowerSave is on;
+ *    which is the power down period in between integration cycles in this mode.
+ *    Note that in the current implementation, PowerSave is hard-coded to be off,
```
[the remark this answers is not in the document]
danpetry: Will clarify this

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,180 @@
...
+
+    dev->low_power_mode = low_power_mode;
+
```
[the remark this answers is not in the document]
danpetry: Will include

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,180 @@
...
+    i2c_write_reg(dev->i2c_dev,
```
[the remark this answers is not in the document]
danpetry: @haukepetersen it would be good to get your feedback on this as the author of the driver guidelines

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    return 0;
+}
+
+bool tsl4531x_is_sample_ready(const tsl4531x_t *dev)
+{
+    assert(dev);
+
+    return ((xtimer_now_usec() - dev->sample_start_time) >
```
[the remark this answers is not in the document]
danpetry: As discussed offline, this will change anyway to the following functionality:
- Return time till sample ready
- Clamp sample ready time at zero if sample's ready

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,180 @@
...
+    return MULTIPLY_DATA((((uint16_t)als_data[_DATAHIGH]) << 8)
+                         + als_data[_DATALOW], dev->integration_time);
+}
+
+int tsl4531x_simple_read(const tsl4531x_t *dev)
+{
+    assert(dev);
+
+    if (dev->low_power_mode) {
+        i2c_write_reg(dev->i2c_dev,
```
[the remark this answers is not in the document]
danpetry: This wasn't possible due to SAUL's requirement for `*dev` to be `const`, however we've discussed an alternative which removes the code duplication by removing the hard requirement to use the `is_sample_ready` function and the `dev->sample_start_time` field

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,180 @@
...
+                      dev->i2c_addr,
+                      TSL4531X_COMMAND(TSL4531X_CONTROL_REG),
+                      TSL4531X_CONTROL(TSL4531X_MODE_SINGLE_ADC_CYCLE),
+                      0);
+
+    }
+
+    /* In low power mode, or just after changing to high power mode, this
+       function blocks for several hundred milliseconds. */
+    while (!tsl4531x_is_sample_ready(dev)) {}
```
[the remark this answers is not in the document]
danpetry: Also as discussed this function needs fixing because I'm not resetting `dev->is_sample_ready`. In any case I will put a sleep in here of 10 ms rather than busy loop

tests/driver_tsl4531x/main.c
```diff
@@ -0,0 +1,122 @@
...
+        lux = tsl4531x_simple_read(&dev);
+        printf("Illuminance       | Low power mode  | Synchronous read                 |");
+        printf(" [lx] | %u\n", lux);
+
+        /* Test asynchronous read - low power mode */
+        if (tsl4531x_start_sample(&dev) < 0) {
+            puts("[Error] Not in low power mode");
+        }
+
+        /* Test how long the sample took to be ready. */
```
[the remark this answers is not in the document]
danpetry: As discussed offline, this tests the API and demonstrates that it works as documented

drivers/include/tsl4531x.h
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * @param      dev           Initialized device descriptor
+ *
+ * @return     Zero
+ */
+int tsl4531x_start_sample(tsl4531x_t *dev);
+
+/**
+ * Deliver time in microseconds until sample is ready, or zero if it is ready. In low power mode, this counts down from the point at which tsl4531x_start_sample is called, and along
+ * with that function and tsl4531x_get_sample, provides asynchronous operation.
```
jcarrano: Control long lines.

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,197 @@
...
+
+    uint32_t t;
+
+    if (dev->low_power_mode) {
+        /* Call the const *dev version, so this basic read also takes a const *dev
+           and therefore is compatible with the SAUL API. */
+        t = _start_sample(dev);
+    }
+    else {
+        /* Wait until the device has collected at least one sample. Note that
```
jcarrano: This is not nice. There should be a way to fix this.

drivers/tsl4531x/tsl4531x.c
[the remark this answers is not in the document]
haukepetersen: I am with @jcarrano, you should check and return an error if you encounter one. The section in the device driver howto guide is actually outdated. So as general rule: if the underlying module is able to detect and return errors, these should be checked and propagated (in a sensible way regarding aggregation of error codes...).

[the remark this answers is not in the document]
danpetry: I will update the driver guidelines and the code

drivers/tsl4531x/tsl4531x.c
danpetry: As discussed offline, I added a flag to indicate whether the device has started up. this is set if
a) the sample ready countdown timer has hit zero, and
b) the sample has been checked (necessary to detect the device having been started up)

The read is now non-const, and in the saul read function I've handed a pointer to a dev pointer which gets around saul's requirement for a const pointer. (i.e. the pointer pointer is const, but the dev pointer isn't)

drivers/tsl4531x/tsl4531x.c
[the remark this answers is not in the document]
jcarrano: Also, as a matter of style and code simplicity, I'm *for* **goto**s in this kind of error handling code. specially if one has to release resources like with the i2c bus here, it makes it easier to do it in only one place.

drivers/tsl4531x/tsl4531x.c
```diff
@@ -136,62 +136,93 @@ int tsl4531x_start_sample(tsl4531x_t *dev)
...
 
-    uint32_t elapsed = (xtimer_now_usec() - dev->sample_start_time);
+    uint32_t t = TSL4531X_GET_INTEGRATION_TIME_USEC(dev->integration_time, TSL4531X_PSAVESKIP_ON) -
+                 (xtimer_now_usec() - dev->sample_start_time);
 
-    return (TSL4531X_GET_INTEGRATION_TIME_USEC(dev->integration_time, TSL4531X_PSAVESKIP_ON) > elapsed ?
-            TSL4531X_GET_INTEGRATION_TIME_USEC(dev->integration_time, TSL4531X_PSAVESKIP_ON) - elapsed :
-            0);
+    /* Clamp t at zero */
+    t = (t <= TSL4531X_GET_INTEGRATION_TIME_USEC(dev->integration_time, TSL4531X_PSAVESKIP_ON) ?
```
jcarrano: if 
* `t = TSL4531X_GET_INTEGRATION_TIME_USEC(...) - (xtimer_now_usec() - dev->sample_start_time);`
* `(xtimer_now_usec() - dev->sample_start_time) > 0` (which is always true because "now" is always later than the time we started the sample)
then
* `t <= TSL4531X_GET_INTEGRATION_TIME_USEC(...)` is always true.

drivers/tsl4531x/tsl4531x_saul.c
```diff
@@ -27,7 +27,7 @@
 static int _read(const void *dev, phydat_t *data)
 {
     memset(data, 0, sizeof(phydat_t));
-    data->val[0] = tsl4531x_simple_read((const tsl4531x_t *)dev);
+    data->val[0] = tsl4531x_simple_read(*(tsl4531x_t **)dev);
```
jcarrano: Replace by `(tsl4531x_t **)` by `(tsl4531x_t * const*)`, otherwise you are casting to an incompatible pointer type (you can check it by compiling with -Wcast-qual).
The end result is the same, of course.

drivers/include/tsl4531x.h
```diff
@@ -96,7 +96,8 @@ typedef struct {
     i2c_t i2c_dev;                              /**< I2C device which is used */
     i2c_t i2c_addr;                             /**< I2C address of sensor */
     tsl4531x_intgn_time_t integration_time;     /**< integration time */
```
jcarrano: If you are going to use bitfields, why not make `tsl4531x_intgn_time_t` a bitfield of size 2 and save 4 bytes for free?

drivers/tsl4531x/tsl4531x.c
danpetry: However, if 
`(xtimer_now_usec() - dev->sample_start_time)`
is greater than 
`TSL4531X_GET_INTEGRATION_TIME_USEC(...)`
then t becomes negative, which wraps around to a large positive number because t is unsigned.
so,
`t > TSL4531X_GET_INTEGRATION_TIME_USEC(...)`. at that point.

drivers/include/tsl4531x.h
danpetry: I implemented this and looked at the build sizes - it actually doesn't save any space in RAM but adds 32 bytes to flash because of the instructions required to handle the bitfields

drivers/tsl4531x/tsl4531x_saul.c
danpetry: will do, thanks

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,228 @@
...
+
+int tsl4531x_set_low_power_mode(tsl4531x_t *dev, bool low_power_mode)
+{
+    assert(dev);
+
+    int r;
+
+    dev->low_power_mode = low_power_mode;
+
+    if ((r = i2c_acquire(dev->i2c_dev) < 0)) {
```
jcarrano: Thanks @codacy-bot for pointing this out.

danpetry: done

drivers/tsl4531x/tsl4531x.c
```diff
@@ -0,0 +1,228 @@
...
+}
+
+int tsl4531x_get_sample(const tsl4531x_t *dev)
+{
+    assert(dev);
+
+    int r;
+    uint8_t als_data[2]; /* = {[DATALOW], [DATAHIGH]} */
+
+    if ((r = i2c_acquire(dev->i2c_dev) < 0)) {
```
jcarrano: Still needs fixing.
```suggestion
    if ((r = i2c_acquire(dev->i2c_dev)) < 0) {
```

PR #10150, opened 2018-10-11
sys/arduino: add arduino time functions micros() and delayMicroseconds()

fedepell:
### Contribution description
This PR adds two standard time related functions in the Arduino module, making it more complete and therefore making it easier to use Arduino unmodified sketches on top of RIOT.
The added functions are:

- micros(): https://www.arduino.cc/reference/en/language/functions/time/micros/
- delayMicroseconds(): https://www.arduino.cc/reference/en/language/functions/time/delaymicroseconds/

### Testing procedure
There is no automatic tests for the Arduino module. To test manually you can call the two new functions in your Arduino sketch.

### DIFF
--- a/sys/arduino/base.cpp
+++ b/sys/arduino/base.cpp
@@ -62,6 +62,16 @@ void delay(unsigned long msec)
     xtimer_usleep(1000 * msec);
 }
 
+void delayMicroseconds(unsigned long usec)
+{
+    xtimer_usleep(usec);
+}
+
+unsigned int micros()
+{
+    return xtimer_now_usec();
+}
+
 /*
  * Bitfield for the state of the ADC-channels.
  * 0: Not initialized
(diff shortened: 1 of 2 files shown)


sys/arduino/base.cpp
```diff
@@ -62,6 +62,16 @@ void delay(unsigned long msec)
     xtimer_usleep(1000 * msec);
 }
 
+void delayMicroseconds(unsigned long usec)
+{
+    xtimer_usleep(usec);
+}
+
+unsigned int micros()
```
leandrolanzieri: Here the return type should be `unsigned long`.

fedepell: Many thanks for tests and spotting this! Corrected now, sorry was ok in the header but wrong here!

Thanks!
F.

leandrolanzieri:
Great, then this is an ACK.

PR #10151, opened 2018-10-11
cpu: change adc_sample return to int32_t

olegart:
### Contribution description

adc_sample return value type changed from `int` to `int32_t`

1) There are 16-bit CPUs on the market with 18-bit or even 24-bit ADCs (e.g. some MSP430s)
2) MISRA-C:2004 rule 6.3 recommends to avoid using types without explicitly specified size and signedness

### Testing procedure

Build tests/periph_adc for the board of your choice.

### DIFF
--- a/cpu/atmega_common/periph/adc.c
+++ b/cpu/atmega_common/periph/adc.c
@@ -99,9 +99,9 @@ int adc_init(adc_t line)
     return 0;
 }
 
-int adc_sample(adc_t line, adc_res_t res)
+int32_t adc_sample(adc_t line, adc_res_t res)
 {
-    int sample = 0;
+    int32_t sample = 0;
 
     /* check if resolution is applicable */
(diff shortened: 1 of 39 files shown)


gebart:
I agree this is a good improvement.

gebart:
Could you grep through the source and update the adc_sample users too?

olegart:
1) ADC resolutions 18/20/22/24 bits added to adc.h and CPU configuration files
2) SAUL rejects ADC resolutions over 16 bits
3) adc_sample users fixed (Arduino's AnalogRead return value still declared as `int` — it is fixed at ADC_RES_10BIT anyway)

olegart:
Regarding SAUL — it is deliberately declared as int16 in phydat.h

olegart:
Will address build issues/typos on weekend.

PR #10154, opened 2018-10-11
sys/arduino: add possibility to customize Arduino serial port at compile time

fedepell:
### Contribution description
This PR gives the possibility to easily customize the serial port used by the Arduino module that was previously hardcoded to be UART0 in the Arduino module source code. With this PR the UART used can be customized in the usual board specific arduino_board.h header file. The default has been kept as UART0 for backwards compatibility. 

### Testing procedure
Manual procedure: use a board setup with the arduino module with more than one serial. Change the ARDUINO_UART_DEV, rebuild and see the different serial being used.

### DIFF
--- a/sys/arduino/doc.txt
+++ b/sys/arduino/doc.txt
@@ -112,6 +112,12 @@
  * @endcode
  *   This links to the third entry in the `arduino_pinmap` array.
  *
+ * - a define `ARDUINO_UART_DEV` that defines the UART to use as the Arduino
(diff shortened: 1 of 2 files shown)


MrKevinWeiss:
Nice feature, everything looks good but I can't test at the moment.

leandrolanzieri:
I tested this on `arduino-mega2560` and `arduino-due` with different UARTS and works fine. So ACK

Issue #10166, opened 2018-10-15
LLVM: C++ includes are not found


jia200x:
#### Description
This issue was detected with the [output of Murdock](https://ci.riot-os.org/RIOT-OS/RIOT/9426/dc6b193e2a332230db8751a31f002108cc460c0c/output.html) after the latest workers update.

The location of the C++ includes changed in the provided ARM toolchain in RIOT-OS/riotdocker#53 and the build system is not able to find them.

#### Steps to reproduce the issue
E.g
`BUILD_IN_DOCKER=1 TOOLCHAIN=llvm make BOARD=samr21-xpro`

#### Expected results
The application should build without any issues.

#### Actual results

```
Building application "tests_cpp11_mutex" for "samr21-xpro" with MCU "samd21".

/data/riotbuild/riotproject/tests/cpp11_mutex/main.cpp:20:10: fatal error: 'string' file not found
#include <string>
```
#### Versions
riot/riotdocker:latest

cladmi:
Any idea what is the header location to use in that case ?

jia200x:
These seem to be the locations:

jia200x:
This seem to be the location:
```
/opt/gcc-arm-none-eabi-7-2018-q2-uopt/gcc-arm-none-eabi-7-2018-q2-update/arm-none-eabi/include/c++/7.3.1/

cladmi:
Thanks.

cladmi:
I am thinking about replacing everything with what was done in `llvm.inc.mk` to ask gcc about its headers location.

cladmi:
From my tests in `arch linux`, ubuntu 16.04 `gcc`, the previous `docker` image and the current docker image, the `GCC_MULTI_DIR` should be useless.

Replacing the directory with the same directory with `GCC_MULTI_DIR` should not change anything. The content of the base directory is the same, with only more directories in it, so I will remove from a proposed solution.

jia200x:
great! Thank you. I will test it on my side

PR #10167, opened 2018-10-15
llvm.inc.mk: completely replace detection of GCC includes

cladmi:
### Contribution description

This completely removes the hardcoded/os specific path setting for includes.
It directly queries gcc include directories.

It takes what was already done by `makefiles/libc/llvm.inc.mk`.

~I also removed the 'GCC_MULTI_DIR' variable as the pointed directory content was already contained in the base directory.
I checked this on arm with `arch-linux`, `ubunt16.04`, the `gcc-arm` ppa, and the arm website toolchain.~
I replaced the `GCC_MULTI_DIR` handling by givining `CFLAGS_CPU` when
searching for include directories.
`CFLAGS` cannot be used as it will crash when `-target $(CPU_ARCH)` is
added.
It currently requires using deferred variables as `CFLAGS_CPU` can be
overwritten later in the build process.

I completely replaced the original code because I found many issues and could not really find a fix:

* TARGET_TRIPLE is undefined for a long time but was still used
* the base code for `CXX` was always putting things in the variable so the backup lines https://github.com/RIOT-OS/RIOT/blob/55b026b3498f3287fb6c6a02df203be9fa9f3419/makefiles/toolchain/llvm.inc.mk#L77-L78
 could never be used
* The backup code before is not adding all sub directories and is not working for c++
* Hardcoded path do not take into account the current gcc version used as one could try using an old version
* For `arm` boards `newlib` is also used so will also query `gcc` for its path.

### Testing procedure

Test compiling the following examples for `BOARD=iotlab-m3` and `TOOLCHAIN=llvm`
with different environments:

`examples/riot_and_cpp/  tests/cpp11_condition_variable  tests/cpp11_mutex  tests/cpp11_thread`

```
make -C examples/riot_and_cpp BOARD=iotlab-m3 TOOLCHAIN=llvm clean all
```

It should compile without issues.

It would be great if you could also give the output of the `GCC_CXX_INCLUDE_DIRS` variable with the different environments:

```
make --no-print-directory -C examples/riot_and_cpp/  TOOLCHAIN=llvm info-debug-variable-GCC_CXX_INCLUDE_DIRS BOARD=iotlab-m3
```
* arch linux
    * [x] iotlab-m3
* ubuntu 16.04
    * [x] iotlab-m3
    * [x] samr21-xpro
* docker arm-gcc ppa
    * [x] iotlab-m3
    * [x] samr21-xpro
* docker new image
    * [x] iotlab-m3
    * [x] samr21-xpro

I could compile and run the tests when there was one correctly on the given list.

The patch also fixes that even if I was using /opt/gcc-arm-none-eabi-6-2017-q2-update the include directories were still search into /etc/alternatives/gcc-arm-none-eabi-include/c++/4.9.3.

Now the CXXINCLUDES also contains the C includes path which was not the case before. They will just be present two times in the command line as we use $(CXXINCLUDES) $(INCLUDES) when compiling.

* [x] I will test tomorrow but still wants to run `murdock` and maybe get feedback.
* [x] Test compiling `tests/cpp11*` on several different systems and testing the output with `TOOLCHAIN=llvm`
* [x] verify what I said about `GCC_MULTI_DIR`. I will explain how to do it or add a static check for testing.
* [ ] Propagate to use the `GCC_C_INCLUDES` variable for `makefiles/libc/newlib.mk`
    * -> Should come in a follow up PR

### Issues/PRs references

Fixes https://github.com/RIOT-OS/RIOT/issues/10166

### DIFF
--- a/makefiles/toolchain/llvm.inc.mk
+++ b/makefiles/toolchain/llvm.inc.mk
@@ -32,72 +32,41 @@ export SIZE        = $(LLVMPREFIX)size
 # We use GDB for debugging for now, maybe LLDB will be supported in the future.
 include $(RIOTMAKE)/tools/gdb.inc.mk
 
-ifneq (,$(TARGET_ARCH))
-  # Clang on Linux uses GCC's C++ headers and libstdc++ (installed with GCC)
-  # Ubuntu and Debian use /etc/alternatives/gcc-$(TARGET_ARCH)-include/c++/$(GCC_VERSION)
-  # Arch uses /usr/$(TARGET_ARCH)/include/c++/$(GCC_VERSION)
-  # Gentoo uses /usr/lib/gcc/$(TARGET_ARCH)/$(GCC_VERSION)/include/g++-v5
-  GCC_CXX_INCLUDE_PATTERNS ?= \
-  /etc/alternatives/gcc-$(TARGET_ARCH)-include/c++/*/ \
-  /usr/$(TARGET_ARCH)/include/c++/*/ \
-  /usr/lib/gcc/$(TARGET_ARCH)/*/include/g++-v8 \
-  /usr/lib/gcc/$(TARGET_ARCH)/*/include/g++-v7 \
-  /usr/lib/gcc/$(TARGET_ARCH)/*/include/g++-v6 \
-  /usr/lib/gcc/$(TARGET_ARCH)/*/include/g++-v5 \
-  #
+# Include directories from gcc
+#   $1: language <c|cpp>
+#
+#   `realpath` is used instead of `abspath` to support Mingw32 which has issues
+#   with windows formatted gcc directories
+#
+# CFLAGS_CPU is used to get the correct multilib include header.
+gcc_include_dirs = $(realpath \
+    $(shell $(PREFIX)gcc $(CFLAGS_CPU) -v -x $1 -E /dev/null 2>&1 | \
+        sed \
+        -e '1,/\#include <...> search starts here:/d' \
+        -e '/End of search list./,$$d' \
+        -e 's/^ *//')\
+)
 
-  # Try to find the proper multilib directory using GCC, this may fail if a cross-
-  # GCC is not installed.
-  ifeq ($(GCC_MULTI_DIR),)
-    GCC_MULTI_DIR := $(shell $(PREFIX)gcc $(CFLAGS) -print-multi-directory 2>/dev/null)
+ifneq (,$(TARGET_ARCH))
+  ifeq (,$(CFLAGS_CPU))
+    $(error CFLAGS_CPU must have been defined to use `llvm`.)
   endif
 
   # Tell clang to cross compile
-  export CFLAGS     += -target $(TARGET_ARCH)
-  export CXXFLAGS   += -target $(TARGET_ARCH)
+  CFLAGS     += -target $(TARGET_ARCH)
+  CXXFLAGS   += -target $(TARGET_ARCH)
   # We currently don't use LLVM for linking (see comment above).
-  #export LINKFLAGS  += -target $(TARGET_ARCH)
+  # LINKFLAGS  += -target $(TARGET_ARCH)
 
-  # Use the wildcard Makefile function to search for existing directories matching
-  # the patterns above. We use the -isystem gcc/clang argument to add the include
-  # directories as system include directories, which means they will not be
-  # searched until after all the project specific include directories (-I/path)
-  # We sort the list of found directories and take the last one, it will likely be
-  # the most recent GCC version. This avoids using old headers left over from
-  # previous tool chain installations.
-  GCC_CXX_INCLUDES ?= \
-      $(addprefix \
-          -isystem $(firstword \
-              $(foreach pat, $(GCC_CXX_INCLUDE_PATTERNS), $(lastword $(sort $(wildcard $(pat)))))), \
-          /. /$(TARGET_ARCH)/$(GCC_MULTI_DIR) /backward \
-      )
+  # Clang on Linux uses GCC's C and C++ headers and libstdc++ (installed with GCC)
 
-  # If nothing was found we will try to fall back to searching for a cross-gcc in
-  # the current PATH and use a relative path for the includes
-  ifeq (,$(GCC_CXX_INCLUDES))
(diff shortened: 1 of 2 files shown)


miri64:
One (?) of the workers can't find their Sam. Restarting Murdock.

miri64:
#scnr

cladmi:
@gebart Thanks for the feedback.

We were only using it to put it into `CXXINCLUDES` so is not provided to the linker.
But indeed we are currently linking with `gcc`.

I had a version were I replace directories from `GCC_CXX_INCLUDE_DIRS` with `d/$(GCC_MULTI_DIR)` if it exists but did not commit it as did not find a good reason to do it yet, so prefer to not add it and ask.

Also it looks like I can get the correct `gcc_include_dirs` with the multilib directory when providing `CFLAGS`. I will update accordingly.
I there a variable I could use instead of the whole `CFLAGS` ?

cladmi:
I used deferred variables for the moment so it does not help. I could replace them with immediate ones.
It looks like I can use `CFLAGS_CPU`. It is defined in all the `makefiles/arch/*` files.

I can also put an assert that it is defined to detect issues if a new cpu tries to use llvm.

cladmi:
Even with an immediate variable, `CFLAGS_FPU` can currently be modified later in build parsing https://github.com/RIOT-OS/RIOT/blob/f376b4669119ec88dae244dfde1b07dc7c416979/pkg/nordic_softdevice_ble/Makefile.include#L27

`nordic_softdevice_ble` is currently blacklisted for `llvm` but if enabled the value would change without being evaluated correctly.

gebart:
@cladmi I think this is beginning to get out of hand. This may be a time to re-evaluate build system tools (laze, cmake, etc.)

cladmi:
The problem will still be there. If a package is allowed to overwrite `FPU` the same behavior will be required. And if it is with a language without deferred then the definition order will be even more important than now.

If it should not be allowed and requires another mechanism to do it, we should also be doing it here.

jia200x:
> @cladmi I think this is beginning to get out of hand. This may be a time to re-evaluate build system tools (laze, cmake, etc.)

+1

miri64:
I've turned off the failing worker for now.

gebart:
> The problem will still be there. If a package is allowed to overwrite `FPU` the same behavior will be required. And if it is with a language without deferred then the definition order will be even more important than now.
> 
> If it should not be allowed and requires another mechanism to do it, we should also be doing it here.

What I mean is that when running a Makefile generator tool it will be necessary to configure the CPU CFLAGS once per build. This should also implicitly improve the speed of some of the informational makefile targets, such as `make info-buildsize-diff` which is slow as hell right now for no particular reason other than that the build system will do all these pre-build checks on the module dependencies and configure the compiler before running the `size` tool on the binary.
I only suggested that we compare our options, if the benefits are less than the cost of refactoring, then there is no reason to use an extra tool.

gebart:
@jia200x 
Looks like an incompatibility of the vendor headers. In C++ apparently, unnamed bitfields are not allowed to have qualifiers (volatile, const). I guess the image has an older Clang version without the check implemented.

jia200x:
> @jia200x
Looks like an incompatibility of the vendor headers. In C++ apparently, unnamed bitfields are not allowed to have qualifiers (volatile, const). I guess the image has an older Clang version without the check implemented.

Yes, it seems so :/

jia200x:
the proposed solution seems ok to me. I verified the paths given by ` arm-none-eabi-gcc -mcpu=cortex-m0plus -mlittle-endian -mthumb -mfloat-abi=soft -v -x c -E /dev/null 2>&1` and they are correct.

I'm +1 with @miri64. I think it's OK to merge this one and tackle the (possible) CFLAGS_FPU evaluation issue later. `nordic_softdevice_ble` is blacklisted anyway, so shouldn't affect us

jia200x:
@cladmi could you please squash?

cladmi:
I have the same errors for `samr21-xpro` on `arch linux`. But the issue is also present on `master` so unrelated to this PR.

cladmi:
@jia200x there is no `CFLAGS_FPU` evaluation issue now with `nordic_softdevice_ble` as I am using a normal deferred variable. So it will use the updated value when set.

However there was one before.

miri64:
Can we *please* merge this PR first, if only to fix the symptom and focus on the detail work later? This is stalling the whole integration process at the moment.

jia200x:
> Can we please merge this PR first, if only to fix the symptom and focus on the detail work later? This is stalling the whole integration process at the moment.

Yes, please let's continue the work in follow-ups.

cladmi:
Tests:

* [ ] arch linux
    * [ ] iotlab-m3
* [ ] ubuntu 16.04
    * [ ] iotlab-m3
    * [ ] samr21-xpro
* [ ] docker arm-gcc ppa
    * [ ] iotlab-m3
    * [ ] samr21-xpro
* [ ] docker new image
    * [ ] iotlab-m3
    * [ ] samr21-xpro

cladmi:
I could compile and run the tests when there was one correctly on the given list.

The patch also fixes that even if I was using `/opt/gcc-arm-none-eabi-6-2017-q2-update` the include directories were still search into `/etc/alternatives/gcc-arm-none-eabi-include/c++/4.9.3`.

Now the `CXXINCLUDES` also contains the C includes path which was not the case before. They will just be present two times in the command line as we use `$(CXXINCLUDES) $(INCLUDES)` when compiling.

cladmi:
I rebased and updated the main description with the last updates.

kYc0o:
Tested on OSX and it solves some problems, taken into account that it never worked good on OSX.

I have this error on current master, which this PR solves.

```console
snake:RIOT facosta$ BOARD=samr21-xpro make -C tests/cpp11_mutex/ TOOLCHAIN=llvm clean all flash test
make: Entering directory '/Users/facosta/git/RIOT-OS/RIOT/tests/cpp11_mutex'
Building application "tests_cpp11_mutex" for "samr21-xpro" with MCU "samd21".

/Users/facosta/git/RIOT-OS/RIOT/tests/cpp11_mutex/main.cpp:20:10: fatal error: 'string' file not found
#include <string>
         ^
1 error generated.
```

jia200x:
ok, Murdock green. Let's GO then

jia200x:
thanks for the fix @cladmi!

PR #9596, opened 2018-07-18
tests/xtimer_drift: added test

Josar:
Added a test to be able to investigate problems like stated in https://github.com/RIOT-OS/RIOT/issues/5103.

Problems arising with this PR can be solved with
https://github.com/RIOT-OS/RIOT/pull/9211
https://github.com/RIOT-OS/RIOT/pull/9595

This PRs have to be merged before this problem is solved.

Additional https://github.com/RIOT-OS/RIOT/pull/8990 can be applied.

This test compares the timing between received messages with the host clock and prints when an error occured.

```
2018-07-18 13:41:16,388 - INFO # main(): This is RIOT! (Version: 2018.10)
2018-07-18 13:41:16,389 - INFO # [START]
2018-07-18 13:41:16,389 - INFO # 
2018-07-18 13:41:17,394 - INFO # now=1.046264 (0x0001fedf ticks), drift=552 us, jitter=552 us
2018-07-18 13:41:19,967 - INFO # now=2.045584 (0x0003e6d2 ticks), drift=-128 us, jitter=-680 us
2.045584: Invalid timebetween messages, expected 0.9 < 1.9108633995056152 < 1.1
2018-07-18 13:41:23,067 - INFO # now=3.045520 (0x0005cf12 ticks), drift=-192 us, jitter=-64 us
3.04552: Invalid timebetween messages, expected 0.9 < 3.1008362770080566 < 1.1
2018-07-18 13:41:25,642 - INFO # now=4.045456 (0x0007b752 ticks), drift=-256 us, jitter=-64 us
4.045456: Invalid timebetween messages, expected 0.9 < 2.5776400566101074 < 1.1
2018-07-18 13:41:27,164 - INFO # now=5.045392 (0x00099f92 ticks), drift=-320 us, jitter=-64 us
5.045392: Invalid timebetween messages, expected 0.9 < 1.517474889755249 < 1.1
2018-07-18 13:41:29,741 - INFO # now=6.045328 (0x000b87d2 ticks), drift=-384 us, jitter=-64 us
6.045328: Invalid timebetween messages, expected 0.9 < 2.6143975257873535 < 1.1
2018-07-18 13:41:31,791 - INFO # now=7.045264 (0x000d7012 ticks), drift=-448 us, jitter=-64 us
```

PR #10173, opened 2018-10-16
test/xtimer_drift: DEBUG_PINS

Josar:
This PR adds debug pins to probe this test with an oscilloscope. 
This helps debugging https://github.com/RIOT-OS/RIOT/pull/9595

Before the merge of https://github.com/RIOT-OS/RIOT/pull/9211#issuecomment-430401266 we can observe the hanging behaviour of the system.

image

After https://github.com/RIOT-OS/RIOT/pull/9211#issuecomment-430401266 is merged a pariodic  a race condition appears which can be seen in the prints of the test https://github.com/RIOT-OS/RIOT/pull/9596. 

When https://github.com/RIOT-OS/RIOT/pull/9595 is merged this will result in an stable System.


PeterKietzmann:
@leandrolanzieri is this helpful? Mind to have a look?

Josar:
Set generic pin definition as in https://github.com/RIOT-OS/RIOT/pull/10174/files

PeterKietzmann:
@MichelRottleuthner, @leandrolanzieri you took care of #10174, how about this PR?

tests/xtimer_drift/Makefile
```diff
@@ -5,4 +5,19 @@ BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-uno \
 
 USEMODULE += xtimer
 
+# Port and pin configuration for probing with oscilloscope
+# Define Test pin for hardware timer interrupt, hardware dependent
+# ATmega
+#CFLAGS += -DDEBUG_TIMER_PORT=PORTF
+#CFLAGS += -DDEBUG_TIMER_DDR=DDRF
+#CFLAGS += -DDEBUG_TIMER_PIN=PORTF4
+
```
A-Paul: Lines 8-14: This description is already in *<cpu/atmega_common/periph/timer.c>* since #9025.
Just place a reference, please.

Josar: @A-Paul A-Paul i am willing to remove the comments if this is the showstopper here.

Josar:
@A-Paul can we continue here?

Josar:
@A-Paul  @PeterKietzmann  @kYc0o is it possible to get this merged as this is a part of the xtimer improvement i started a long time ago.

I somehow get the feeling that there is no interrest from the maintainers to get xtimer working proprerly.

Now old and known issues are coming up again https://github.com/RIOT-OS/RIOT/issues/11149.

kYc0o:
So far the only thing missing is the testing part. Unfortunately I don't have the tools anymore to proceed with a good testing here, however I also think that this PR has waited too much and would be beneficial to merge it.

@A-Paul would you have time for the testing?

tests/xtimer_drift/Makefile
PeterKietzmann: This test should be independent of the atmega implementation. Ideally, we would have a generic debug pin for this purpose configured for every platform. Now that this is out of scope for your PR, I'd say we remove it.

tests/xtimer_drift/Makefile
```diff
@@ -5,4 +5,19 @@ BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-uno \
...
+# Define test probing pins GPIO API based.
+# Port number should be found in port enum e.g in cpu/include/periph_cpu.h
+#FEATURES_REQUIRED += periph_gpio
```
PeterKietzmann: Is there a single platform that has no GPIO driver but xtimer support? In my opinion it should be fine to always include that feature (otherwise we would see what Murdoch says later)

tests/xtimer_drift/Makefile
```diff
@@ -5,4 +5,19 @@ BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-uno \
...
+# Jiminy probing Pins
```
PeterKietzmann: If you really wanna have the jiminy config here, please define the CFLAGS in 
```
ifneq (,$(filter jiminy,$(BOARD)))

endif
```

tests/xtimer_drift/main.c
```diff
@@ -67,6 +72,11 @@ void *slacker_thread(void *arg)
     (void) arg;
     timex_t now;
 
+#if defined(SLACKER_THREAD_PIN)
+    gpio_t slacker_pin = SLACKER_THREAD_PIN;
+    gpio_init(slacker_pin, GPIO_OUT);
```
PeterKietzmann: Why not directly `gpio_init(SLACKER_THREAD_PIN, GPIO_OUT);` here and following?

tests/xtimer_drift/main.c
```diff
@@ -79,7 +89,10 @@ void *slacker_thread(void *arg)
         struct timer_msg *tmsg = m.content.ptr;
         xtimer_now_timex(&now);
         xtimer_usleep(TEST_MSG_RX_USLEEP);
-
+#if defined(SLACKER_THREAD_PIN)
+        gpio_set(slacker_pin);
+        gpio_clear(slacker_pin);
```
PeterKietzmann: Probably a matter of taste, but wouldn't a single toggle be simpler and easier to read on the scope?

Josar:
@PeterKietzmann i think its a 32 bit timer which is used for xtimer, waiting for that to overflow takes some time. Maybe you could configure it to use a 16bit timer.

PeterKietzmann:
> waiting for that to overflow takes some time

I wasn't waiting for an overflow :-). Are you saying this because I mentioned that my results look slightly different to yours? I'm fine with the concept of this PR and IMO it only needs some small polish in order to be merged

PR #10174, opened 2018-10-16
test/xtimer_hang: DEBUG_PINS

Josar:
This PR adds debug pins to probe this test with an oscilloscope.
This helps debugging https://github.com/RIOT-OS/RIOT/pull/9211

When runnig this on main branch it can be observed that the xtimer are not fired when they are expected. The whole system is interrupted for a whole timer circle untill the set target is reached again.

image

This is solved by https://github.com/RIOT-OS/RIOT/pull/9211 

What we wold expect is that the main pin is on for ca. 100ms and that in a regular interval
Also we would expect the worker to be after 1ms and 1.1ms. (As this falls in the XTIMER_ISR_BACKOFF the first timer is delayed untill the second is ready. Thus the time interval is longer.)

The second figure also shows cicled in red when there is a hardware interrupt before the second worker thread timer is set. This leads to a separation of the timer interrupts, As we only count 7 until they are in the same hardware interrupt again we lost one worker time 2, which is expected as it has a lower priority then worker timer 1. And thus in the moment when the hardware interrupt for the 2 worker is executet it has to wait for the 1 worker timer because of the XTIMER_ISR_BACKOFF and ten the first timer is executed first as it has the higher priority.

image

### DIFF
--- a/tests/xtimer_hang/README.md
+++ b/tests/xtimer_hang/README.md
@@ -9,3 +9,14 @@ the test has failed.
 
 Please note (again), this is a runtime test to check if xtimer runs into a
 deadlock, it is not about clock stability nor accuracy of timing intervals.
+
+When debug pins are used and observed the expected output is as follows:
+The MAIN_THREAD_PIN is on for ca. 100ms and that in a regular interval. If this interval is not regular or has gaps this
+is an error. The WORKER_THREAD_PIN should toggle after 1ms and 1.1ms. (As this might fall in the XTIMER_ISR_BACKOFF the
+first pin toggle is delayed untill the second is ready. Thus the time interval can be longer.)
+If the Timer fall in the same interrupt there might be a interrupt before the second worker thread timer is set.
+This leads to a separation of the timer interrupts until they again fall in the same interrupt.
+It might happen that from the seperation of the interrupts till the merge there is only a uneven count of
+WORKER_THREAD_PIN toggles, which means a loss of one worker time 2, which is expected as it has a lower priority then
+worker timer 1. And thus in the moment when the hardware interrupt for the 2 worker is executet it has to wait for the
+1 worker timer because of the XTIMER_ISR_BACKOFF and so the first timer is executed first as it has the higher priority.
--- a/tests/xtimer_hang/main.c
+++ b/tests/xtimer_hang/main.c
@@ -29,6 +29,11 @@
 #include "thread.h"
 #include "log.h"
 
+#if defined(MAIN_THREAD_PIN) || defined(WORKER_THREAD_PIN)
+#include "board.h"
+#include "periph/gpio.h"
+#endif
+
 #define TEST_TIME_S             (10LU)
 #define TEST_INTERVAL_MS        (100LU)
 #define TEST_TIMER_STACKSIZE    (THREAD_STACKSIZE_DEFAULT)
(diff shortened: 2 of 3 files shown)


PeterKietzmann:
@leandrolanzieri is this helpful? Mind to have a look?

tests/xtimer_hang/main.c
```diff
@@ -38,14 +43,31 @@ char stack_timer2[TEST_TIMER_STACKSIZE];
 
 void* timer_func(void* arg)
 {
+#ifdef WORKER_THREAD_PIN
+    printf("Debug worker thread port 0x%02x pin %d\n",  WORKER_THREAD_PORT, WORKER_THREAD_PIN);
+    gpio_t worker_pin = GPIO_PIN(WORKER_THREAD_PORT, WORKER_THREAD_PIN);
```
basilfx: Instead of

```c
gpio_t worker_pin = GPIO_PIN(WORKER_THREAD_PORT, WORKER_THREAD_PIN);
```

why not 

```
gpio_t worker_pin = WORKER_THREAD_PIN;
```

and define 

```
-DWORKER_THREAD_PIN=GPIO_PIN\(PORTF, 6\)
```

in the Makefile? We do that at [several](https://github.com/RIOT-OS/RIOT/blob/master/tests/pkg_u8g2/Makefile#L19) [places](https://github.com/RIOT-OS/RIOT/blob/master/tests/driver_nrf24l01p_lowlevel/Makefile#L14).

(@kYc0o what do you think?)

leandrolanzieri: +1 on defining `WORKER_THREAD_PIN` and also `MAIN_THREAD_PIN`

leandrolanzieri:
I believe this is an interesting feature to have. @Josar maybe a note could be added on what to expect on the pins, like you did here https://github.com/RIOT-OS/RIOT/pull/9211#issuecomment-430401266 ?

Josar:
@basilfx @leandrolanzieri done.

kYc0o:
@leandrolanzieri would you mind to ACK if you agree with this PR? Have you tested?

kYc0o:
Please squash and @leandrolanzieri please set the label for CI when it's squashed.

Josar:
Squashed and rebased.

tests/xtimer_hang/README.md
```diff
@@ -9,3 +9,14 @@ the test has failed.
...
+
+When debug pins are used and observed the expected output is as follows:
+The MAIN_THREAD_PIN is on for ca. 100ms and that in a regular interval. If this interval is not regular or has gaps this
+is an error. The WORKER_THREAD_PIN should toggle after 1ms and 1.1ms. (As this might fall in the XTIMER_ISR_BACKOFF the
+first pin toggle is delayed untill the second is ready. Thus the time interval can be longer.)
+If the Timer fall in the same interrupt there might be a interrupt before the second worker thread timer is set.
+This leads to a separation of the timer interrupts until they again fall in the same interrupt.
+It might happen that from the seperation of the interrupts till the merge there is only a uneven count of
+WORKER_THREAD_PIN toggles, which means a loss of one worker time 2, which is expected as it has a lower priority then
+worker timer 1. And thus in the moment when the hardware interrupt for the 2 worker is executet it has to wait for the
```
leandrolanzieri: Trailing whitespace here. Please remove.

Josar: Done.

PR #9890, opened 2018-09-06
ethos: Only accept frame if previous frame is read

bergzand:
### Contribution description

This PR changes ethos to only accept a new frame if the previous frame is read. It only changes whether the bytes are actually stored to prevent disrupting the ethos state machine.

I've also included the assert from #9889 to prevent a similar breakage in the future.

### Testing procedure

same as #9889 (flood a gnrc_border_router with ping), but now it should keep working. 

### Issues/PRs references

Fixes #9889

PR #10178, opened 2018-10-17
drivers/ethos: remove unnecessary casts.

### LINKED ISSUE #10177: Warn when casts do no preserve qualifiers (-Wcast-qual)
### Description

With the compiler options we are using now, explicit casts that discard a qualifier (like `volatile` or `const`) do not trigger a warning.

I propose to enable these warnings by adding the compiler flag `-Wcast-qual`.

I tried compiling a couple of simple examples to see how feasible it is and I found I found several violations. Some of them are in the kernel. **That does not mean the kernel is buggy, read below**.

### Kernel

Before somebody says "we could disable this warning for stuff in /core". Yes, but we could also fix it, and in any case, the core is the most critical part, so we better have lots of checks there.

The kernel defines some stuff, like `sched_active_thread` as volatile, because the active thread can change at any moment.

Of course, when the kernel is running the thread is not changing because it is the kernel who changes threads, so in that situation it is "ok" to discard the qualifier.

I'd argue, however, that it is not that "it's ok to discard the qualifier", but  rather that "in that situation the variable is not const". The hard part is how we explain that to the compiler.

#### `containerof`

When using GCC `containerof` is defined as:

```c
#   define container_of(PTR, TYPE, MEMBER) \
        (__extension__ ({ \
            __extension__ const __typeof__ (((TYPE *) 0)->MEMBER) *__m____ = (PTR); \
            ((TYPE *) ((char *) __m____ - offsetof(TYPE, MEMBER))); \
        }))
```

Why is that `const` there? Is it necessary?

### Other code

Do a `WERROR=0 CFLAGS=-Wcast-qual make` and you will see a number of violations, most of them discarding `const`.

jcarrano:
### Contribution description

The code contained superfluous casts which did not preserve the const qualifiers. Functionality is unchanged, but there are no more warnings reported by -Wcast-qual for this module.

### Testing procedure

Same as `ethos`.

### Issues/PRs references

Trying to fix_ #10177 or at least improve the situation.

### DIFF
--- a/drivers/ethos/ethos.c
+++ b/drivers/ethos/ethos.c
@@ -200,16 +200,16 @@ static size_t iolist_count_total(const iolist_t *iolist)
 
 static void _write_escaped(uart_t uart, uint8_t c)
 {
-    uint8_t *out;
+    const uint8_t *out;
     int n;
 
     switch(c) {
         case ETHOS_FRAME_DELIMITER:
-            out = (uint8_t*)_esc_delim;
+            out = _esc_delim;
             n = 2;
             break;
         case ETHOS_ESC_CHAR:
-            out = (uint8_t*)_esc_esc;
+            out = _esc_esc;
             n = 2;
             break;
         default:
@@ -244,7 +244,7 @@ void ethos_send_frame(ethos_t *dev, const uint8_t *data, size_t len, unsigned fr
 
     /* send frame content */
     while(len--) {
-        _write_escaped(dev->uart, *(uint8_t*)data++);
+        _write_escaped(dev->uart, *data++);
     }
 
     /* end of frame */


bergzand:
@smlng I opened #9890 because it also gave me some troubles.

jcarrano:
> Question: does ethos work right now, anyway? 

I don't know. I just know that it was throwing casting things back and forth for no reason so I fixed it.

drivers/ethos/ethos.c
```diff
@@ -198,18 +198,18 @@ static size_t iolist_count_total(const iolist_t *iolist)
     return result;
 }
 
-static void _write_escaped(uart_t uart, uint8_t c)
+static void _write_escaped(uart_t uart, const uint8_t c)
```
bergzand: Adding `const` here doesn't really make sense and doesn't add anything, as `c` is not a pointer.

jcarrano: I guess I should practice what I preach. I though it would complain below when I assigned it's address to a `const uint8_t *`.

aabadie:
@bergzand, I think your change request has been addressed here. Do you ACK ?

bergzand:
@aabadie No longer blocking this. I would like to test ethos quickly but don't have the hardware here at the moment.

aabadie:
> I would like to test ethos quickly but don't have the hardware here at the moment.

@bergzand, you have it on IoT-LAB with A8 nodes and their connected M3. See this [tutorial](https://www.iot-lab.info/tutorials/riot-public-ipv6-a8-m3/).

bergzand:
> @bergzand, you have it on IoT-LAB with A8 nodes and their connected M3. See this [tutorial](https://www.iot-lab.info/tutorials/riot-public-ipv6-a8-m3/).

I get the hint :)

Tested it today, everything works fine for me. ACK

PR #10181, opened 2018-10-17
pkg/nimble: use STACKTEST flag for controller thread

haukepetersen:
### Contribution description
Just debugged some hard-fault in nimble and suspected a stack overflow, until I noticed that `ps` did not show the actual values due to the missing `STACKTEST` flag... So this PR enables the THREAD_CREATE_STACKTEST flag when creating the controller thread in case DEVELHELP is enabled.

### Testing procedure
build with and without DEVELHELP should suffice.

### Issues/PRs references
none

### DIFF
--- a/pkg/nimble/contrib/controller_init.c
+++ b/pkg/nimble/contrib/controller_init.c
@@ -39,7 +39,8 @@ void nimble_riot_controller_init(void)
      * Create task where NimBLE LL will run. This one is required as LL has its
      * own event queue and should have highest priority.
      */
-    thread_create(stack, sizeof(stack), NIMBLE_CONTROLLER_PRIO, 0,
-                  (thread_task_func_t)nimble_port_ll_task_func,
-                  NULL, "nimble_ctrl");
+    thread_create(stack, sizeof(stack), NIMBLE_CONTROLLER_PRIO,
+                  THREAD_CREATE_STACKTEST,
+                  (thread_task_func_t)nimble_port_ll_task_func, NULL,
+                  "nimble_ctrl");
 }


bergzand:
I assume that I should compile the `examples/nimble_gatt` for testing this? Compilation with `DEVELHELP=1` works, compilation with `DEVELHELP=0` fails on:
```
/home/hydrazine/dev/RIOT/examples/nimble_gatt/bin/pkg/nrf52dk/nimble/nimble/host/src/ble_att_clt.c: In function 'ble_att_clt_tx_write_cmd':
/home/hydrazine/dev/RIOT/examples/nimble_gatt/bin/pkg/nrf52dk/nimble/nimble/host/src/ble_att_clt.c:754:9: error: variable 'rc' set but not used [-Werror=unused-but-set-variable]
     int rc;
         ^
/home/hydrazine/dev/RIOT/examples/nimble_gatt/bin/pkg/nrf52dk/nimble/nimble/host/src/ble_hs.c: In function 'ble_hs_event_start':
/home/hydrazine/dev/RIOT/examples/nimble_gatt/bin/pkg/nrf52dk/nimble/nimble/host/src/ble_hs_conn.c: In function 'ble_hs_conn_addrs':
/home/hydrazine/dev/RIOT/examples/nimble_gatt/bin/pkg/nrf52dk/nimble/nimble/host/src/ble_hs_conn.c:363:9: error: variable 'rc' set but not used [-Werror=unused-but-set-variable]
     int rc;
         ^
/home/hydrazine/dev/RIOT/examples/nimble_gatt/bin/pkg/nrf52dk/nimble/nimble/host/src/ble_hs.c:496:9: error: variable 'rc' set but not used [-Werror=unused-but-set-variable]
     int rc;
         ^
/home/hydrazine/dev/RIOT/examples/nimble_gatt/bin/pkg/nrf52dk/nimble/nimble/host/src/ble_hs.c: In function 'ble_hs_init':
/home/hydrazine/dev/RIOT/examples/nimble_gatt/bin/pkg/nrf52dk/nimble/nimble/host/src/ble_hs.c:619:9: error: variable 'rc' set but not used [-Werror=unused-but-set-variable]
     int rc;
```

Obviously unrelated to this PR (also fails on master). @haukepetersen I'm willing to ACK this, changes make sense and it builds without the changes here which should cover `FLAGS=0` case too. Furthermore, building with `DEVELHELP=0 make -j BOARD=nrf52dk -i`, it shows that `controller_init.c` which is changed here compiles without any issues.

pkg/nimble/contrib/controller_init.c
```diff
@@ -27,6 +27,12 @@
 #include "nrf_clock.h"
 #endif
 
+#ifdef DEVELHELP
+#define FLAGS           (THREAD_CREATE_STACKTEST)
+#else
+#define FLAGS           (0)
+#endif
```
bergzand: I don't think this distinction between compiling with and without `DEVELHELP` is a good idea. If I read `core/thread.c` correctly, `THREAD_CREATE_STACKTEST` is only used when either `DEVELHELP` or `SCHED_TEST_STACK` is enabled and is ignored otherwise. Changing the handling here to `DEVELHELP` only could cause confusion with people when stacktest is not enable while they might have enabled `SCHED_TEST_STACK`

haukepetersen: i was already thinking that we covered this thing globally somewhere, but I could not remember so hence this PR. Will adapt to get rid of the duplication...

haukepetersen:
this should be straight forward now. Will look into the develhelp issue in a separate PR.

bergzand:
Murdock says that compilation works, Go!

Issue #10187, opened 2018-10-17
tests/unittests: Move large unittests to individual tests


bergzand:
I propose breaking out a number of large tests from the unittests to their own test. This to decrease the flash size of the full unittest binary a bit.

## Tests:

Flash sizes as mentioned below are based on the size of the required flash of the package itself, or when it is not a package, the flash required for the test itself. All flash sizes are measured for the nrf52840dk board.

- [ ] tweetnacl (9KB) #10184 
- [ ] qDSA (13KB) #10185 
- [ ] libcose #10183
- [ ] hacl (10KB)
- [ ] spiffs (16KB)
- [ ] gnrc_ipv6_nib (38KB)

### Optionally:

- [ ] hashes (19KB)
- [ ] Bloom (15KB)

Packages that would still remain if all these tests are moved outside are cn-cbor at almost 2KB and heatshrink at 1.5KB

smlng:
good idea, but I would rephrase to: *move out all unittests that require external packages*.

smlng:
move package tests from unittests, which are

```
cd tests/unittests && git grep USEPKG
tests-cn_cbor/Makefile.include:USEPKG += cn-cbor
tests-hacl/Makefile.include:USEPKG += hacl
tests-heatshrink/Makefile.include:USEPKG += heatshrink
tests-libcose/Makefile.include:USEPKG += libcose
tests-qDSA/Makefile.include:USEPKG += qDSA
tests-relic/Makefile.include:USEPKG += relic
tests-tweetnacl/Makefile.include:USEPKG += tweetnacl
```

- [ ] cn_cbor
- [ ] hacl
- [ ] heatshrink
- [ ] libcose #10183 
- [ ] qDSA #10185
- [ ] relic
- [ ] tweetnacl #10184

bergzand:
My goal here is shrink the size of the unittest binary a bit. Removing all packages from the unittests helps a lot and I agree that they should not be in the unittests. However there are a few tests in there that are also large enough that they in my opinion qualify for having a separate test.

bergzand:
at #10183 we reached the conclusion that the crypto tests only get to run on native as not to increase the load on the hardware CI units too much. Those tests take around 1-2mins per crypto tests on a samr21-xpro and could significantly impact the runtime of the CI.

smlng:
[...]

PR #10185, opened 2018-10-17
tests/qdsa: Move from unittests to regular tests

bergzand:
And another!

### Contribution description

similar to #10183, this is another large package that only increases the size of the unittests.

### Testing procedure

Run the `tests/pkg_qdsa` test.

### Issues/PRs references

Part of https://github.com/RIOT-OS/RIOT/issues/10187

PR #10184, opened 2018-10-17
tests/tweetnacl: Move from unittests to regular test

bergzand:
### Contribution description

Similar to #10183. 

- I think only qDSA and HACL* require a lot of stack size, removing those from the unittests probably removes the large stack requirement

### Testing procedure

I suspect that the test works on AVR and msp430, but I'm not sure. Testing this is preferred in case those need to be blacklisted

Insufficient memory list is manually generated based on the `tests/pkg_libcose` blacklist and memory list.

### Issues/PRs references

PR #10189, opened 2018-10-18
tests/relic: move from unittests to regular test

smlng:
### Contribution description

This moves tests for the relic package from unittests to a regular
test, which should help to decrease binary size of unittests.

### Testing procedure

run the tests, i.e. `BOARD=<your-favourite> make -C tests/pkg_relic flash test`

### Issues/PRs references

- issue #10187
- similar PRs: #10183, #10184, #10185, #10188

### DIFF
--- a/pkg/relic/Makefile.include
+++ b/pkg/relic/Makefile.include
@@ -1,1 +1,8 @@
 INCLUDES += -I$(PKGDIRBASE)/relic/include
+
+ifneq (,$(filter cortex-m%,$(CPU_ARCH)))
+  # relic package package is not using system includes right now, so
+  # many newlib headers (not even stdio.h) are not found.
+  # Fixed in #9821 for jerryscript, should be applicable here too.
+  TOOLCHAINS_BLACKLIST += llvm
+endif
--- a/tests/pkg_relic/Makefile
+++ b/tests/pkg_relic/Makefile
@@ -1,0 +1,45 @@
+include ../Makefile.tests_common
+
+# The following boards are known to fail or have not been tested.
+BOARD_BLACKLIST :=  arduino-duemilanove \
+                    arduino-mega2560 \
+                    arduino-uno \
+                    chronos \
+                    f4vi1 \
+                    jiminy-mega256rfr2 \
+                    mega-xplained \
+                    msb-430 \
+                    msb-430h \
+                    msbiot \
+                    redbee-econotag \
+                    stm32f0discovery \
+                    stm32f3discovery \
+                    telosb \
+                    waspmote-pro \
+                    wsn430-v1_3b \
+                    wsn430-v1_4 \
+                    z1 \
+                    #
+
+BOARD_INSUFFICIENT_MEMORY :=    nucleo-f030r8 \
+                                nucleo-f031k6 \
+                                nucleo-f042k6 \
+                                nucleo-l031k6 \
+                                nucleo-l053r8 \
+                                #
+
+CFLAGS += -DTHREAD_STACKSIZE_MAIN=\(5*THREAD_STACKSIZE_DEFAULT\)
+
+USEPKG += relic
+USEMODULE += embunit
+
+# -DWORD=32 : 	Specifies the word width of the target system. This is
+#  				currently not automatically detected so adjusted to your target
+#  				platform.
+
+# The rest of the parameters are configuration parameters for RELIC described in its documentation.
+export RELIC_CONFIG_FLAGS=-DARCH=NONE -DOPSYS=NONE -DQUIET=off -DWORD=32 -DFP_PRIME=255 -DWITH="BN;MD;DV;FP;EP;CP;BC;EC" -DSEED=ZERO
+
+TEST_ON_CI_WHITELIST += all
+
+include $(RIOTBASE)/Makefile.include
--- a/tests/pkg_relic/tests/01-run.py
+++ b/tests/pkg_relic/tests/01-run.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+
+# Copyright (C) 2017 Freie Universität Berlin
+#
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import sys
+from testrunner import run
+
+
+def testfunc(child):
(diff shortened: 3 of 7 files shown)


cladmi:
Question, the `DISABLE_TEST_FOR_ARM_CORTEX_M := tests-relic` and others has not been propagated here. What is the reason for this ?

tests/pkg_relic/Makefile
```diff
@@ -0,0 +1,46 @@
...
+                    #
+
+BOARD_INSUFFICIENT_MEMORY :=    nucleo-f030r8 \
+                                nucleo-f031k6 \
+                                nucleo-f042k6 \
+                                nucleo-l031k6 \
+                                nucleo-l053r8 \
+                                #
+
+TEST_BOARDS_LLVM_COMPILE := native \
```
cladmi: This should be done in the `relic` package to blacklist llvm if it is not compatible using `TOOLCHAIN_BLACKLIST += llvm` with the correct conditionals.

tests/pkg_relic/Makefile
```diff
@@ -0,0 +1,46 @@
...
+                            #
+
+CFLAGS += -DTHREAD_STACKSIZE_MAIN=\(4*THREAD_STACKSIZE_DEFAULT\)
```
cladmi: This was not set for this test in `unittests` why now ?

tests/pkg_relic/Makefile
```diff
@@ -0,0 +1,46 @@
+include ../Makefile.tests_common
+
```
cladmi: This one should also be enable for CI

```suggestion

TEST_ON_CI_WHITELIST += all
```

But as `arm` was not supported before maybe it should just be `+= native`.

cladmi:
The test currently fails on `iotlab-m3` and `samr21-xpro`.

With a stack size of `-DTHREAD_STACKSIZE_MAIN=\(5*THREAD_STACKSIZE_DEFAULT\)` test successfully executes on both boards.

tests/pkg_relic/Makefile
```diff
@@ -28,7 +28,7 @@ BOARD_INSUFFICIENT_MEMORY :=    nucleo-f030r8 \
                                 nucleo-l053r8 \
                                 #
 
-export TEST_BOARDS_LLVM_COMPILE := native
+TOOLCHAIN_BLACKLIST += llvm
```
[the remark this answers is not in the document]
cladmi: Indeed it is the same issue, so you can reuse the same comment as `jerryscript`.

It compiles with `llvm` with this fix, I can do a PR for the fix:

``` diff
diff --git a/pkg/relic/Makefile b/pkg/relic/Makefile
index 7e2466c9d..d73aecbee 100644
--- a/pkg/relic/Makefile
+++ b/pkg/relic/Makefile
@@ -18,7 +18,7 @@ all: $(PKG_BUILDDIR)/Makefile

 $(PKG_BUILDDIR)/Makefile: $(TOOLCHAIN_FILE)
        cd $(PKG_BUILDDIR) && \
-       COMP="$(filter-out -Werror -Werror=old-style-definition -Werror=strict-prototypes -std=gnu99, $(CFLAGS) ) " \
+       COMP="$(INCLUDES) $(filter-out -Werror -Werror=old-style-definition -Werror=strict-prototypes -std=gnu99, $(CFLAGS) ) " \
        cmake -DCMAKE_TOOLCHAIN_FILE=$(TOOLCHAIN_FILE) \
                  -DCHECK=off -DTESTS=0 -DBENCH=0 -DSHLIB=off -Wno-dev $(RELIC_CONFIG_FLAGS) .

```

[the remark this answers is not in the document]
smlng: so just to clarify: its either blacklisting llvm for the relic package **or** applying your fix for jerryscript to make it work with llvm? Or is both somehow needed?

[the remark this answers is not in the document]
cladmi: I would say it is now blacklisting because it does not work for `llvm` and `cortex-m`, it is a fact, and it is an immediate solution.

Making it work for `llvm` is an improvement that needs work, so a new feature somehow with specific review. So can come in a second step.

I do not delay this PR because of something that was broken for years.
I was just adding details so the `blacklisting` comment would be accurate.

pkg/relic/Makefile.include
```diff
@@ -1 +1,8 @@
 INCLUDES += -I$(PKGDIRBASE)/relic/include
+
+ifneq (,$(filter cortex-m%,$(CPU_ARCH)))
+  # jerryscript package package is not using system includes right now, so
```
cladmi: Same but with `relic` instead of `jerryscript`. :p

smlng: yeah fixed the wording

cladmi:
I added the `run tests` label as it uses `CI_WHITELIST` and re-triggered murdock.

cladmi:
I will finish reviewing after murdock.

tests/pkg_relic/tests/01-run.py
```diff
@@ -0,0 +1,18 @@
...
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import sys
+from testrunner import run
+
+
+def testfunc(child):
+    child.expect('OK \(\d+ tests\)', timeout=120)
```
cladmi: `flake8` is correctly complaining about this one, this should be a `regex` string 
```suggestion
    child.expect(r'OK \(\d+ tests\)', timeout=120)
```

When rebasing to `riot/master` it will not be detected anymore as it will be using `riotdocker` image that uses the repository `flake8` version but the error is still valid.

cladmi:
@smlng Please tell when you force push, github does not send any notification about it

cladmi:
Just waiting for murdock and it will be good to merge.

smlng:
@cladmi sorry about the force-push. On another note: the regex is wrong in some other tests in master too.

cladmi:
I know, it was already mentioned in another test migration out of tests/unittests, but `static-tests` are mainly only done on changed files.

cladmi:
Thank you for doing the migration, I find it magic all the issues there can be with simple tasks like this.

PR #10193, opened 2018-10-18
coap: Update list of defines for "PATCH and FETCH Methods" and SenML

bergzand:
### Contribution description

Updated our list of defines with the ones from rfc8132 (Fetch, patch and ipatch) and accompanying response codes. I decided to rewrite the existing hexadecimal response codes to decimal to better match the formatting of the [IANA response code list](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#response-codes)

Also added content types from rfc8428 (senml).

Decided to bundle it in a single PR because it is defines only.

### Testing procedure

Test if nanocoap and/or gcoap still compiles.

### Issues/PRs references

None

### DIFF
--- a/sys/include/net/coap.h
+++ b/sys/include/net/coap.h
@@ -65,6 +65,9 @@ extern "C" {
 #define COAP_METHOD_POST        (2)
 #define COAP_METHOD_PUT         (3)
 #define COAP_METHOD_DELETE      (4)
+#define COAP_METHOD_FETCH       (5)
+#define COAP_METHOD_PATCH       (6)
+#define COAP_METHOD_IPATCH      (7)
 /** @} */
 
 /**
@@ -104,9 +107,11 @@ extern "C" {
 #define COAP_CODE_METHOD_NOT_ALLOWED         ((4 << 5) | 5)
 #define COAP_CODE_NOT_ACCEPTABLE             ((4 << 5) | 6)
 #define COAP_CODE_REQUEST_ENTITY_INCOMPLETE  ((4 << 5) | 8)
+#define COAP_CODE_CONFLICT                   ((4 << 5) | 9)
 #define COAP_CODE_PRECONDITION_FAILED        ((4 << 5) | 12)
 #define COAP_CODE_REQUEST_ENTITY_TOO_LARGE   ((4 << 5) | 13)
 #define COAP_CODE_UNSUPPORTED_CONTENT_FORMAT ((4 << 5) | 15)
+#define COAP_CODE_UNPROCESSABLE_ENTITY       ((4 << 5) | 22)
 /** @} */
 
 /**
--- a/sys/include/net/coap.h
+++ b/sys/include/net/coap.h
@@ -145,13 +145,23 @@ extern "C" {
  * @anchor  net_coap_format
  * @{
  */
-#define COAP_FORMAT_TEXT         (0)
-#define COAP_FORMAT_LINK        (40)
-#define COAP_FORMAT_XML         (41)
-#define COAP_FORMAT_OCTET       (42)
-#define COAP_FORMAT_EXI         (47)
-#define COAP_FORMAT_JSON        (50)
-#define COAP_FORMAT_CBOR        (60)
+#define COAP_FORMAT_TEXT                      (0)
+#define COAP_FORMAT_LINK                     (40)
+#define COAP_FORMAT_XML                      (41)
+#define COAP_FORMAT_OCTET                    (42)
+#define COAP_FORMAT_EXI                      (47)
+#define COAP_FORMAT_JSON                     (50)
+#define COAP_FORMAT_JSON_PATCH_JSON          (51)
+#define COAP_FORMAT_MERGE_PATCH_JSON         (52)
+#define COAP_FORMAT_CBOR                     (60)
+#define COAP_FORMAT_SENML_JSON              (110)
+#define COAP_FORMAT_SENSML_JSON             (111)
+#define COAP_FORMAT_SENML_CBOR              (112)
+#define COAP_FORMAT_SENSML_CBOR             (113)
+#define COAP_FORMAT_SENML_EXI               (114)
+#define COAP_FORMAT_SENSML_EXI              (115)
+#define COAP_FORMAT_SENML_XML               (310)
+#define COAP_FORMAT_SENSML_XML              (311)
 /** @} */
 
 /**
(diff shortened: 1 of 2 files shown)


bergzand:
Now with a sensible subject :laughing:

kb2ma:
What to do about the nanocoap method flag macros for coap_resource_t, defined in nanocoap.h? All things being equal, I would add them in this PR like:
```
COAP_FETCH  (0x10)
```
However, this need might be a good opportunity also to rename the prefix for these macros. IMO the current naming scheme is confusing relative to the COAP_METHOD_xxx macros. (I'm thinking @haukepetersen will agree here :-)). So, the new prefix could be NANOCOAP_FLAG_xxx,  for example
```
NANOCOAP_FLAG_FETCH  (0x10)
```
For existing values we could deprecate the old names and add new names, and also add the new ones for fetch and patch at the same time. 

We could do this work in a separate commit in this PR, or in a separate PR for increased visibility since they are used extensively. What do you think?

sys/include/net/coap.h
```diff
@@ -139,13 +144,21 @@ extern "C" {
...
-#define COAP_FORMAT_JSON        (50)
-#define COAP_FORMAT_CBOR        (60)
+#define COAP_FORMAT_TEXT                      (0)
+#define COAP_FORMAT_LINK                     (40)
+#define COAP_FORMAT_XML                      (41)
+#define COAP_FORMAT_OCTET                    (42)
+#define COAP_FORMAT_EXI                      (47)
+#define COAP_FORMAT_JSON                     (50)
+#define COAP_FORMAT_CBOR                     (60)
+#define COAP_FORMAT_APPLICATION_SENML_JSON  (110)
```
kb2ma: I see in RFC 7252 that the name of the format for JSON, for example, is "application/json", yet we don't use the 'application' part in the macro. Do we need to add 'application' for the new macros? It makes the names significantly longer, and it seems like 'COAP_FORMAT_xxx' already eliminates any ambiguity. I'm not really schooled on format names, so maybe I'm missing something though.

bergzand:
> When I first saw this PR, I thought "new macros, this should be easy". Oh well, it is what it is. :-) Definitely worth a little discussion.

Yeah, I was hoping to easily get this merged, but your points on the name and `nanocoap.h` are valid.

sys/include/net/coap.h
bergzand: I don't mind dropping the `_APPLICATION` part. We could also shorten it to `_APP`, but I see no real harm in dropping it completely.

bergzand:
> IMO the current naming scheme is confusing relative to the COAP_METHOD_xxx macros. (I'm thinking @haukepetersen will agree here :-)). So, the new prefix could be NANOCOAP_FLAG_xxx,

I think I agree with you here.

> We could do this work in a separate commit in this PR, or in a separate PR for increased visibility since they are used extensively. What do you think?

Yes please! I really don't want to piggy-back an API change here :)

kb2ma:
Sounds good -- agreed on separate PR to convert COAP_xxx macros to NANOCOAP_FLAG_xxx.

sys/include/net/nanocoap.h
```diff
@@ -166,10 +166,13 @@ extern "C" {
...
  */
-#define COAP_GET                (0x1)
-#define COAP_POST               (0x2)
-#define COAP_PUT                (0x4)
-#define COAP_DELETE             (0x8)
+#define COAP_GET                (0x01)
+#define COAP_POST               (0x02)
+#define COAP_PUT                (0x04)
+#define COAP_DELETE             (0x08)
+#define COAP_PATCH              (0x10)
```
kb2ma: COAP_PATCH -> COAP_FETCH

Typo. I expect we will fix this one to stay consistent with the ordering of method codes.

kb2ma:
Don't we also need to add COAP_FORMAT_xxx codes for json-patch+json and merge-patch+json defined in 8132?

bergzand:
> Don't we also need to add COAP_FORMAT_xxx codes for json-patch+json and merge-patch+json defined in 8132? I have not looked closely, just noticed them in the IANA Considerations section.

Yeah, nothing against adding them. I didn't bother with it in the first place because we don't use really use JSON a lot with RIOT

bergzand:
> > Don't we also need to add COAP_FORMAT_xxx codes for json-patch+json and merge-patch+json defined in 8132? I have not looked closely, just noticed them in the IANA Considerations section.
> 
> Yeah, nothing against adding them. I didn't bother with it in the first place because we don't use really use JSON a lot with RIOT

After a second look, those are from a different RFC, I'll add them to the list here, but that's probably the original reason why I skipped them

sys/include/net/coap.h
```diff
@@ -139,13 +144,24 @@ extern "C" {
...
-#define COAP_FORMAT_CBOR        (60)
+#define COAP_FORMAT_TEXT                      (0)
+#define COAP_FORMAT_LINK                     (40)
+#define COAP_FORMAT_XML                      (41)
+#define COAP_FORMAT_OCTET                    (42)
+#define COAP_FORMAT_EXI                      (47)
+#define COAP_FORMAT_JSON                     (50)
+#define COAP_FORMAT_JSON_PATCH_JSON          (51)
+#define COAP_FORMAT_MERGE_PATCH_JSON         (52)
+#define COAP_FORMAT_JSON                     (50)
```
kb2ma: Only need to define once ;-)

kb2ma:
Except for the duplicate COAP_FORMAT_JSON, looks fine to me. Please squash with the fix, and I think we'll be good to go.

I just did some automated testing on native -- unit tests and release specs.

bergzand:
@kb2ma squashed and rebased!

kb2ma:
@bergzand ping

As soon as you fix the duplicate COAP_FORMAT_JSON, and rebase and resolve the conflicts (easy), I am happy to ACK.

bergzand:
> As soon as you fix the duplicate COAP_FORMAT_JSON, and rebase and resolve the conflicts (easy), I am happy to ACK.

Removed the duplicate define and fixed the alignment. Took a bit too long to fix it. I'll rework the commits to a sensible set of changes when squashing.

kb2ma:
Looks good to me. Please squash.

I added a little code to the gcoap example to test making a FETCH request, and that worked. LwM2M is on my personal roadmap, and v1.1 uses FETCH and iPATCH. Looking forward to addition of support for these methods.

kb2ma:
I was too quick to ACK due to the merge conflict. Apologies for the confusion.

bergzand:
Rebased

bergzand:
Squashed

bergzand:
@kb2ma I think the set of should be good as they are now. Can you please check that I didn't mix  different sets of defines in the single commit.

bergzand:
@kb2ma Thanks!

PR #10195, opened 2018-10-18
Makefile.base: use thin static archives.

jcarrano:
## Contribution description

Normal, or thick archives contain a copy of the object code. Thin archives, on the contrary, are just an index to the .o files.

This patch does two things:

1. Change ARFLAGS to enable the "T" options.
2. Call AR with all relative paths.

The second step is necessary because the build system handles all absolute paths. If the index in the thin archive contains absolute paths, archives created in docker are no usable outside, and moving the objects breaks the archive.

If all arguments to AR are relative, the resulting archive contains filenames *relative to the .a file* and nothing should break as long as the relative location of the .a and .o remains unchanged.

Compilation time is unchanged, but **disc usage** is **reduced** by approximately **2x**:

## Disk space savings

The following numbers were extracted from these branches:

- https://github.com/cladmi/RIOT/tree/wip/du/buildtest (vanilla)
- https://github.com/jcarrano/RIOT/tree/wip/du/buildtest-with-thinarc (the previous branch with this PR on top)

After compiling everything, there will be an output directory containing the sizes of the build artefacts. I used these command lines:

```sh
$ find . -name "*.bin.bindirsize" -type f -exec tail  -n1 '{}' \; | cut -f 1 | awk '{s+=$1} END {printf "%.0f", s}'
$ find . -name "*.pkg.bindirsize" -type f -exec tail  -n1 '{}' \; | cut -f 1 | awk '{s+=$1} END {printf "%.0f", s}'
$ find .  -type f -exec tail  -n1 '{}' \; | cut -f 1 | awk '{s+=$1} END {printf "%.0f", s}'
```

Sizes are broken down into package and non-package files because package data is mostly unaffected by this change. Units is million KiB (sorry for the mixture of 1000 and 1024 base)

| Thin Archive      |  no               |  yes                   | Savings (%) |
| --------------------- | ---------------: | -------------------: | ---------------- |
| pkg (10e6 KiB)   | 1 790          | 905                   | 49%             |
| Non pkg             |       71         |  71                    | 1%               |
| Total                   | 1 812          | 976                   | 46 %            |

### Thin archives

One of the issues with thin archives is how paths to object files are stored in .a files. If done incorrectly, there will be absolute paths, which means that docker-generated file won't be usable outside the container and, more generally, that it is not possible to move the bin directory.

I added a new target that checks if the archives have absolute paths. Run:

```
BOARD=<your-board> make archive-check
```

Inside and outside of docker (the command is docker-enabled).

### ~Make functions~

~I added tests for the new make functions, in `tests/build_system_utils`. Just run `make`. I refactored the code and improved the output to make it clearer which tests are being run.~

## Other stuff

- ~The commits following the first one may need some tidying up/reworking.~
- ~I would like to have feedback regarding `utils.inc.mk` (is that the correct place to put the code?).~
- ~I have no idea why disc space savings are so high. I was expecting 50% percent reduction or so, not 90%.~

## Dependencies

~Depends on #11218~
Depends on #12155.

### DIFF
--- a/Makefile.base
+++ b/Makefile.base
@@ -67,10 +67,12 @@ $(BINDIR)/$(MODULE)/:
 
 $(BINDIR)/$(MODULE).a $(OBJ): | $(BINDIR)/$(MODULE)/
 
+relpath = $(shell realpath --relative-to=$(abspath .) $(1))
+
 $(BINDIR)/$(MODULE).a: $(OBJ) | $(DIRS:%=ALL--%)
 	@# Recreate archive to cleanup deleted/non selected source files objects
 	$(Q)$(RM) $@
-	$(Q)$(AR) $(ARFLAGS) $@ $^
+	$(Q)$(AR) $(ARFLAGS) $(foreach f,$@ $^,$(call relpath,$f))
 
 CXXFLAGS = $(filter-out $(CXXUWFLAGS), $(CFLAGS)) $(CXXEXFLAGS)
 CCASFLAGS = $(filter-out $(CCASUWFLAGS), $(CFLAGS)) $(CCASEXFLAGS)
--- a/Makefile.include
+++ b/Makefile.include
@@ -498,6 +498,49 @@ print-size: $(ELFFILE)
 
 endif # BUILD_IN_DOCKER
 
+# Rules to check the correctness of thin archives.
+
+relpath = $(shell realpath --relative-to=$(abspath .) $(1))
+
+# Each ARCHECK file contains all the absolute paths found inside the archive.
+BASELIB_ARCHECKS = $(patsubst %.a,%.a-check,$(filter %.a,$(BASELIBS)))
+
+# For each a file, print the absolute paths found inside it
+# If "ar t" is called with an absolute path it will print an abs path regardless
+# of how the archive is internally
+# In case of a malformed archive, ar prints to stderr (and sets an error code).
+# Doing `2>&1` is hacky, the correct thing would be to get the exit code.
+# The `| %.a` is necessary to be able to check file produced in docker.
+%.a-check: %.a
+	$(Q)$(AR) t $(call relpath,$<) 2>&1 | grep '^/' | '$(LAZYSPONGE)' $(LAZYSPONGE_FLAGS) '$@'
+
+# There's no point on keeping files whose content is later copied to another file
+.INTERMEDIATE: $(BASELIB_ARCHECKS)
+
+ARCHIVE_CHECK = $(BINDIR)/$(APPLICATION).archive-check
+
+$(ARCHIVE_CHECK): $(BASELIB_ARCHECKS)
+	$(Q)cat $^ | '$(LAZYSPONGE)' $(LAZYSPONGE_FLAGS) '$@'
+
+# Rule to check if thin archives are correctly produced, that is, with a correct
+# relative path.
+ifeq ($(BUILD_IN_DOCKER),1)
+archive-check: ..in-docker-container
+else
+archive-check: $(ARCHIVE_CHECK) FORCE
+	@if [ -s '$<' ] ; then \
+		$(COLOR_ECHO) '$(COLOR_RED)Found the following absolute paths in archives' ;\
+		cat '$<';\
+		$(COLOR_ECHO) '$(COLOR_RESET)' ;\
+		exit 1;\
+	elif [ -f '$<' ] ; then \
+		$(COLOR_ECHO) '$(COLOR_GREEN)Archives correctly formed$(COLOR_RESET)' ;\
+	else \
+		$(COLOR_ECHO) '$(COLOR_RED)Unexpected error (file not found)$(COLOR_RESET)' ;\
+		exit 1;\
+	fi
+endif # BUILD_IN_DOCKER
+
 # Check given command is available in the path
 #   check_cmd 'command' 'description'
 define check_cmd
--- a/makefiles/cflags.inc.mk
+++ b/makefiles/cflags.inc.mk
@@ -65,5 +65,5 @@ CFLAGS += $(filter-out $(OPTIONAL_CFLAGS_BLACKLIST),$(OPTIONAL_CFLAGS))
 # Default ARFLAGS for platforms which do not specify it.
 # Note: make by default provides ARFLAGS=rv which we want to override
 ifeq ($(origin ARFLAGS),default)
-  ARFLAGS = rcs
+  ARFLAGS = rcTs
 endif
--- a/makefiles/docker.inc.mk
+++ b/makefiles/docker.inc.mk
@@ -9,6 +9,7 @@ export DOCKER_MAKECMDGOALS_POSSIBLE = \
   scan-build \
   scan-build-analyze \
   tests-% \
+  archive-check \
   #
 export DOCKER_MAKECMDGOALS = $(filter $(DOCKER_MAKECMDGOALS_POSSIBLE),$(MAKECMDGOALS))
 
--- a/sys/arduino/Makefile.include
+++ b/sys/arduino/Makefile.include
@@ -2,7 +2,7 @@
 
 # Define application sketches module, it will be generated into $(BINDIR)
 SKETCH_MODULE     ?= arduino_sketches
-SKETCH_MODULE_DIR ?= $(BINDIR)/$(SKETCH_MODULE)
+SKETCH_MODULE_DIR ?= $(BINDIR)/$(SKETCH_MODULE)_src
 SKETCHES           = $(wildcard $(APPDIR)/*.sketch)
 include $(RIOTBASE)/sys/arduino/sketches.inc.mk
 
--- a/sys/arduino/sketches.inc.mk
+++ b/sys/arduino/sketches.inc.mk
@@ -19,7 +19,8 @@ SKETCH_GENERATED_FILES = $(SKETCH_MODULE_DIR)/Makefile $(SKETCH_MODULE_DIR)/$(SK
 # Building the module files
 #   Do not use $^ in receipes as Makefile is also a prerequisite
 $(SKETCH_MODULE_DIR)/Makefile: $(SKETCH_MODULE_DIR)/$(SKETCH_CPP)
-	$(Q)echo 'SRCXX = $(SKETCH_CPP)'               > $@
+	$(Q)echo 'MODULE = $(SKETCH_MODULE)'           > $@
+	$(Q)echo 'SRCXX = $(SKETCH_CPP)'              >> $@
 	$(Q)echo 'include $$(RIOTBASE)/Makefile.base' >> $@
 $(SKETCH_MODULE_DIR)/$(SKETCH_CPP): $(SKETCHES_ALL)
 	@mkdir -p $(@D)


Makefile.base
```diff
@@ -67,10 +67,12 @@ $(BINDIR)/$(MODULE)/:
 
 $(BINDIR)/$(MODULE).a $(OBJ): | $(BINDIR)/$(MODULE)/
 
+relpath = $(shell realpath --relative-to=$(abspath .) $(1))
+
 $(BINDIR)/$(MODULE).a: $(OBJ) | $(DIRS:%=ALL--%)
 	@# Recreate archive to cleanup deleted/non selected source files objects
 	$(Q)$(RM) $@
-	$(Q)$(AR) $(ARFLAGS) $@ $^
+	$(Q)$(AR) $(ARFLAGS) $(foreach f,$(abspath $@ $^),$(call relpath,$f))
```
gebart: instead of forking a new subshell we could patsubst away the BINDIR prefix from the objects, since the output (`$@`) is always located in the top of the BINDIR tree where all the objects are
```suggestion
	$(Q)$(AR) $(ARFLAGS) $@ $(patsubst $(dir $@)%,%,$^)
```

kaspar030:
pls try on OS X

jcarrano:
> pls try on OS X

Can a OSX user do this for me, please?

Makefile.base
jcarrano: The problem is that make is running from a different directory than BINDIR, so those paths will be wrong. Also "$@" also needs to be relative.

jcarrano: I tried stripping RIOTBASE, but that will break external apps and external modules.

gebart: Why does `$@` have to be relative?
I thought the relative paths were supposed to be relative to the archive file, I guess it needs a different approach.

gebart: I played around a bit with the proposed makefile change.
Using this line I get fully working thin archives both in Docker and outside. `ar -t bin/frdm-kw41z/core.a` works on the host after building the archive inside Docker with `make BUILD_IN_DOCKER=1`, and the file names are shown as relative to the current directory instead of absolute.

```suggestion
	$(Q)cd $(dir $@) && $(AR) $(ARFLAGS) $@ $(patsubst $(dir $@)%,%,$^)
```

jcarrano:
For the cool kids using Mac, how about they [Think Different™](https://www.youtube.com/watch?v=cFEarBzelBs)

```make
# remaining words in a list
_rest = $(wordlist 2,$(words $(1)),$(1))
# return an empty string if the arguments are different
_streq = $(filter $(1),$(2))
_putdots = $(patsubst %,..,$(1))
_relpath = $(if $(call _streq,$(firstword $(1)),$(firstword $(2))),\
$(call _relpath,$(call _rest,$(1)),$(call _rest,$(2))),\
$(call _putdots,$(1)) $(2))

_joinpath = $(if $(1),$(addsuffix $(call _joinpath,$(call _rest,$(1))),$(firstword $(1))/))
_splitpath = $(subst /, ,$(1))

relpath = $(call _joinpath,$(call _relpath,$(call _splitpath,$(1)),$(call _splitpath,$(2))))

all:
	echo $(call relpath,/1/2/3/5/6/7,/1/2/3/4/5/6/7)
	echo $(call relpath,/1/2/3/5/6/7,/)
	echo $(call relpath,/,/1/2/3/5/6/)
	# these fail but it should not matter because $1 should always be a dir
	# and $2 a file
	echo $(call relpath,/,/)
	echo $(call relpath,/1/2/3,/1/2/3)
```

Now we don't need `realpath`/`coreutils` anymore.

Makefile.base
jcarrano: We did some experiments with @cladmi on `ar` with relative and absolute paths. In all `ar` invocations, the paths given to the tool must make sense when interpreted from the current working directory.

- `ar rcTs ABSOLUTE.a ABSOLUTE.o` always works, but the path of the o file is stored as absolute in the a file.
- `ar rcTs  RELATIVE.a RELATIVE.o` always works, and `ar` stores the path of the o file relative to the a file (it converts it internally).
- `ar rcTs  ABSOLUTE.a RELATIVE.o` causes ar to store the path to the .o file **unchanged**, as it was given in the command line. This gives an invalid .a, unless the command is ran from the same directory where the a file is located.

jcarrano:
@smlng @kYc0o Would you mind giving this a try on Mac?

smlng:
output on macOS:

```
 BOARD=samr21-xpro make -C tests/minimal/ archive-check
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/boards/samr21-xpro
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/core
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/samd21
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/cortexm_common
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/cortexm_common/periph
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/sam0_common
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/sam0_common/periph
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/samd21/periph
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/drivers
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/drivers/periph_common
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/isrpipe
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/newlib_syscalls_default
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/pm_layered
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/stdio_uart
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/tsrb
Archives correctly formed
```

cladmi:
I plan to review this after the release as it should not go in after soft feature freeze.
But I agree with the direction and was actually already discussing about it with @jcarrano IRL.

Makefile.include
```diff
@@ -457,6 +457,41 @@ print-size: $(ELFFILE)
...
+else
+archive-check: $(ARCHIVE_CHECK) FORCE
+	@if [ -s '$<' ] ; then \
+		$(COLOR_ECHO) '$(COLOR_RED)Found the following absolute paths in archives' ;\
+		cat '$<';\
+	elif [ -f '$<' ] ; then \
+		$(COLOR_ECHO) '$(COLOR_GREEN)Archives correctly formed' ;\
+	else \
+		$(COLOR_ECHO) '$(COLOR_RED)Unexpected error (file not found)' ;\
+	fi
```
smlng: the color is not reset to the default, i.e. before changing to red or green. So everything after this check is run is printed red or green respectively.

jcarrano:
Can we Murdock this thing, see if it works for all tests/examples/boards?

Makefile.include
jcarrano: Done.

jcarrano:
The initial space savings measurements were way off. I updated the description with improved statistics. The savings are now close to the expected 50%.

cladmi:
I think you can apply this diff to make it run in murdock for all build cases:

``` diff
diff --git a/.murdock b/.murdock
index 5d03a2a09..37daef4ec 100755
--- a/.murdock
+++ b/.murdock
@@ -155,7 +155,7 @@ compile() {

     # compile
     CCACHE_BASEDIR="$(pwd)" BOARD=$board TOOLCHAIN=$toolchain RIOT_CI_BUILD=1 \
-        make -C${appdir} clean all -j${JOBS:-4}
+        make -C${appdir} clean all archive-check -j${JOBS:-4}
     RES=$?

     # run tests
```

I could not directly push to your branch.

Makefile.base
```diff
@@ -70,7 +70,7 @@ $(BINDIR)/$(MODULE).a $(OBJ): | $(BINDIR)/$(MODULE)/
 $(BINDIR)/$(MODULE).a: $(OBJ) | $(DIRS:%=ALL--%)
 	@# Recreate archive to cleanup deleted/non selected source files objects
 	$(Q)$(RM) $@
-	$(Q)$(AR) $(ARFLAGS) $@ $^
+	$(Q)$(AR) $(ARFLAGS) $(foreach f,$(abspath $@ $^),$(call relpath,$(abspath .),$f))
```
cladmi: The `relpath` implementation should take care of the `foreach` like most makefile functions do.
Also I would prefer the python arguments ordering https://docs.python.org/3.7/library/os.path.html#os.path.relpath

```
relpath, directories, optional_ref_directory
```
Where optional_ref_directory is replaced by `$(abspath .)` if it is empty. It can be done with `$(if $2,$2,$(abspath .))` in the function call.

Makefile.include
```diff
@@ -457,6 +457,47 @@ print-size: $(ELFFILE)
...
+.INTERMEDIATE: $(BASELIB_ARCHECKS)
+
+ARCHIVE_CHECK = $(BINDIR)/$(APPLICATION).archive-check
+
+$(ARCHIVE_CHECK): $(BASELIB_ARCHECKS)
+	$(Q)cat $^ | '$(LAZYSPONGE)' $(LAZYSPONGE_FLAGS) '$@'
+
+# Rule to check if thin archives are correctly produced, that is, with a correct
+# relative path.
+ifeq ($(BUILD_IN_DOCKER),1)
```
cladmi: The `archive-check` does not need to be done in docker I think.

Using relative path for archives was done to ensure that a file generated in docker are still valid in the host system.
So checking on the host system when build in docker should be enough.

makefiles/utils.inc.mk
```diff
@@ -0,0 +1,87 @@
+# Assorted make utilities
+# TODO: move tests out of here
+
+test-all-utils: test-streq test-assert-positive test-assert_eq-positive
+
+# Return an empty string if the arguments are different (works only for single
+# words
+streq = $(filter $(1),$(2))
```
cladmi: This should be more `findstring`, maybe with a `strip` after, to not use patter matching.

makefiles/utils.inc.mk
```diff
@@ -0,0 +1,87 @@
...
+	@echo empty,empty: $(call streq,,)
+	@echo empty,x: $(call streq,,x)
+	@echo x, empty: $(call streq,x,)
+	@echo x, y: $(call streq,abc,abcw)
+	@echo y, x: $(call streq,abcw,abc)
+	@echo x,x: $(call streq,abcw,abcw)
+
+# Produce an error if the argument is empty. The second argument defines the
+# message.
+assert = $(if $(1),,$(error $(2)))
```
cladmi: This is now the similar to `ensure_value` from https://github.com/RIOT-OS/RIOT/pull/10342 so definitely putting ithem in common would help :)

makefiles/utils.inc.mk
```diff
@@ -0,0 +1,87 @@
...
+
+# Split a path into its components
+splitpath = $(subst /, ,$(1))
+
+# Given a directory and a path (both absolute), calculate the path of the second
+# argument relative to the first one.
+relpath = $(call joinpath,$(call _relpath,$(call splitpath,$(1)),$(call splitpath,$(2))))
+
+# Use coreutils to calculate a relative path
+sys_relpath = $(shell realpath -m --relative-to=$(1) $(2))
```
cladmi: This cannot be tested with mac, so replacing it with a `python` version would be better.

makefiles/utils.inc.mk
```diff
@@ -0,0 +1,87 @@
...
+	@echo OK? $(call assert_eq,$(call relpath,/,/1/2/3/5/6/),$(call sys_relpath,/,/1/2/3/5/6/),)
+
+test-relpath-negtive:
+	# these fail but it should not matter because $1 should always be a dir
+	# and $2 a file
+	echo $(call relpath,/,/)
+	echo $(call relpath,/1/2/3,/1/2/3)
+
+# Return a path relative to the current directory
+relpath_here = $(call relpath,$(abspath .),$(abspath $(1)))
```
cladmi: As said in the first comment, this should be handled by using a default value.

makefiles/utils.inc.mk
```diff
@@ -0,0 +1,87 @@
...
+
+# Intercalate: concatenate words in a list with a separator between each one.
+intercal = $(firstword $(2))$(call concat,$(addprefix $(1),$(call rest,$(2))))
+
+test-intercal:
+	@echo OK? $(call assert_eq,$(call intercal,-,ab xy z),ab-xy-z,)
+	@echo OK? $(call assert_eq,$(call intercal,-,ab),ab,"This should not fail")
+
+_putdots = $(patsubst %,..,$(1))
+_relpath = $(if $(call streq,$(firstword $(1)),$(firstword $(2))),\
```
cladmi: A documentation explaining this one would be great as it is the main implementation.

makefiles/utils.inc.mk
```diff
@@ -0,0 +1,87 @@
...
+assert_eq = $(if $(call streq,$(1),$(2)),,$(error $(if $(3),$(3),Expected $(2) and got $(1))))
+
+test-assert_eq-positive:
+	@echo OK? $(call assert_eq,x,x,"This should not fail")
+
+test-assert_eq-negative:
+	@echo OK? $(call assert_eq,z,x,"This should fail")
+
+# Get the remaining words in a list
+rest = $(wordlist 2,$(words $(1)),$(1))
```
cladmi: Some functions like this one would maybe better be private namespaced functions as the name is not really explicit and could collide with another name.

jcarrano:
> Question, maybe also for my future me testing, does the generated archive contains timestamp of the objects ? or is there a way to generate it with a creation time of the newest object.

If the ".o" is a prerequisite of the ".a" then the latter will be re-generated whenever the former changes, so this is not an issue (as long as dependencies are correctly declared).

makefiles/utils.inc.mk
jcarrano: I think assert can be made an alias for "ensure_value"

cladmi:
@jcarrano I realized after that I was mixing file content and file timestamp in my head. So whenever we re-create it it will be ok.

makefiles/utils.inc.mk
jcarrano: One requirement for this (I should document it) is that two empty strings are considered different.

Makefile.include
jcarrano: If I disable docker support for this one it means that either:

1. The user must have the toolchain available locally (to use $(AR))
2. The native `ar` could be used, but then another variable should be created to hold it's name (someting like $(NATIVE_AR))

I don't think this is a big deal since it is just a diagnostic tool.

makefiles/utils.inc.mk
jcarrano: "Overloading" `relpath` to do something special when the first (or second) argument is empty exposes us to bugs: one of the arguments could come from a variable that is not defined (think a typo) and then, instead of getting an- obviously wrong- empty result one gets a path relative to the current directory.

jcarrano:
I rebased, integrated the make stuff with the `build_system_tests` and fixupsquashed the commits.

jcarrano:
I rolled back all changes related to the wacky text- and path-processing functions in make. It was holding this change back.

Mac users: get coreutils. Really, it is not that hard.

jcarrano:
This is failing because of #12155.

benpicco:
#12155 is merged, try rebasing.

jcarrano:
@benpicco There's no need to rebase, I just toggled the "ready for ci" label.

jcarrano:
On @cladmi's suggestion I rebased so that the pr branch can be directly tested.

benpicco:
Arduino still has issues:
> `arduino_sketches.a: error adding symbols: Malformed archive`

jcarrano:
mmm, that's weird. I only seems to happen with llvm.

jcarrano:
@benpicco 

I found the issue:

```
rm -f /home/jcarrano/source/vanillaRIOT/examples/arduino_hello-world/bin/stm32f4discovery/arduino_sketches.a
llvm-ar rcTs ../arduino_sketches.a arduino_sketches.o
```

See, the archive is being created from (i.e. the PWD is) the `bin/stm32f4discovery/arduino_sketches/`, where it should be `bin/stm32f4discovery`. Indeeed, creating the archive manually by doing:

```
cd bin/stm32f4discovery
llvm-ar rcTs arduino_sketches.a arduino_sketches/arduino_sketches.o
```

And then (manually) running the linker step works.

benpicco:
This breaks `esp_[wifi|now]` :cry: #12258

PR #10201, opened 2018-10-19
tests/libfixmath: Add delay at start of test

### LINKED ISSUE #9766: Initial UART data corruption with libfixmath test after flash
#### Description
It appears that sending lots of data after flashing has corrupted data when using `make flash term BOARD=...`.    I tried this on nucleo boards and samr21 with the same result.  I tried adding extra timeout after resetting.  If a reset button is pressed after the it seems to be able to handle the data.  If a sleep or delay is added before spitting out data in the main function if fixes it.  I tried with the nucelo connection (stlink), a separate uart to usb converter and the samr debug port.  Maybe the tests should always have something that starts running instead of just dumping data like a [START] command or something?

`make BOARD=samr21-xpro flash term -C tests/libfixmath`

#### Expected results
`Unary` in the first line after the connection is complete

#### Actual results
Varying and corrupt data, usually missing the first 20 or so characters.

MrKevinWeiss:
### Contribution description

Add a delay to the start of the test since many boards start outputting before a connection is formed and the input buffer gets flooded.

### Testing procedure

`BOARD=nucleo-f103rb make flash term -C tests/libfixmath`
check the first output, it should be `Unary.`
compare against master.

_I have tested on nucleo but it should be tested on other boards especially with different default clock speeds since it uses a dumb delay_ 
### Issues/PRs references

fixes #9766

### DIFF
--- a/tests/libfixmath/Makefile
+++ b/tests/libfixmath/Makefile
@@ -2,7 +2,8 @@ include ../Makefile.tests_common
 
 USEPKG += libfixmath
 USEMODULE += libfixmath
+USEMODULE += xtimer
 
-TEST_ON_CI_WHITELIST += native
+TEST_ON_CI_WHITELIST += all
 
 include $(RIOTBASE)/Makefile.include
--- a/tests/libfixmath/main.c
+++ b/tests/libfixmath/main.c
@@ -30,6 +30,7 @@
 
 #include <stdio.h>
 
+#include "xtimer.h"
 #include "fix16.h"
 
 #ifndef M_PI
@@ -184,6 +185,8 @@ static void unary_ops(void)
 
 int main(void)
 {
+    /* Delay output to prevent flooding of buffer */
+    xtimer_sleep(1);
     puts("Unary.");
     unary_ops();
 


tests/libfixmath/main.c
```diff
@@ -184,7 +191,10 @@ static void unary_ops(void)
 
 int main(void)
 {
+    /* Delay output to prevent flooding of buffer */
+    dumb_delay(1000000);
```
jia200x: Please use a macro here (e.g `#define UART_DELAY 1000000`)

jia200x:
Btw I'm OK with the proposed fix. Although, I think we should cover these kind of race conditions issues with other mechanism (echos, etc).

I will run some tests here

aabadie:
I've updated the initial comment of this PR using the automatic related issue closing syntax.

tests/libfixmath/main.c
MrKevinWeiss: Something like that.

tests/libfixmath/main.c
```diff
@@ -184,6 +195,8 @@ static void unary_ops(void)
 
 int main(void)
 {
+    /* Delay output to prevent flooding of buffer */
+    dumb_delay(PREVENT_UART_FLOOD_DELAY);
```
cladmi: I would prefer to use `xtimer_sleep_s` here.

tests/libfixmath/main.c
```diff
@@ -36,6 +36,10 @@
 #   define M_PI 3.14159265359
 #endif
 
+#ifndef M_PI
```
cladmi: Wrong macro name.

cladmi:
The test works with an `iotlab-m3` on my computer with an added delay.

jia200x:
please squash! :)

tests/libfixmath/Makefile
```diff
@@ -2,6 +2,7 @@ include ../Makefile.tests_common
 
 USEPKG += libfixmath
 USEMODULE += libfixmath
+USEMODULE += xtimer
 
 TEST_ON_CI_WHITELIST += native
```
cladmi: If it is working it can now be enabled on CI also.

You can inline in the squash

```suggestion
TEST_ON_CI_WHITELIST += all
```

cladmi:
I successfully tested on IoT-LAB with

    "arduino-zero"
    "iotlab-m3"
    "nrf52dk"
    "samr21-xpro"
    "samr30-xpro"

And on my computer with `iotlab-m3`,  `samr21-xpro`.


For information, it fails on `arduino-mega2560` even with a 60s timeout, but also failed in master so I did not investigate why for the moment.
`Timeout in expect script at "child.expect('{}\(-?\d+.\d+\) = -?\d+.\d+'.format(op_name))" (tests/libfixmath/tests/01-run.py:14)`



The commit description is currently not reflecting the problem.
What this does is something like (with my not so nice English writing)

    The delay allows time to pexpect to empty `make term` output buffer from the previous output before the test starts and tries to fiil it again.

The problem is in the output buffer of make term and not the input one.

MrKevinWeiss:
I do know the arduinomega-2560 takes several seconds to kick out of the bootloader after flashing...  Should I just blacklist?

cladmi:
@MrKevinWeiss I just tested it because it was on my desk. It was already broken in master so no need to address it with this PR. I may try to look at it during the release.

MrKevinWeiss:
Great, if that is the case @jia200x and @cladmi have all your comments been addressed.  (Thanks for testing btw @cladmi)

cladmi:
And thanks for addressing the issue, I was still thinking about fixing the problem globally instead of first fixing the test.

MrKevinWeiss:
It isn't just pexpect, it happens with just `make flash term`.  Anyways I updated the commit message.

cladmi:
@MrKevinWeiss it is normal that it does not work with `make flash term`, you basically start a music, with your sound being muted, and you turn on the sound 30 seconds later and expect to get the first second of the music. I would not like to have 30 seconds blank at the start of each music just in case.

In the tests we do `make term`, and then `make reset` to get the first output.

@jia200x I am good with the changes, I let you finish your review and merge when you want.

jia200x:
I'm also good with the changes

jia200x:
Let's go, it's needed for the Release tests

Issue #10206, opened 2018-10-19
Lua_basic example doesn't flash in b-l072z-lrwan1


jia200x:
#### Description
The `lua_basic` example fails to flash on `b-l072z-lrwan1` in current RIOT master. It seems there's a tiny overflow that's not being detected by the build system.

#### Steps to reproduce the issue
Flash `lua_basic` example in `b-l072z-lrwan1`.

#### Expected results

The application is flashed without errors
#### Actual results

```
make RIOT_CI_BUILD=1 CC_NOCOLOR=1 --no-print-directory -C /home/jialamos/Development/RIOT/examples/lua_basic flash-only test
make[1]: Nothing to be done for 'Makefile.include'.
make[1]: Nothing to be done for 'Makefile.include'.
/home/jialamos/Development/RIOT/dist/tools/openocd/openocd.sh flash
### Flashing Target ###
Open On-Chip Debugger 0.10.0+dev-00226-g1c2e3d41d (2018-01-02-18:28)
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
WARNING: interface/stlink-v2-1.cfg is deprecated, please switch to interface/stlink.cfg
hla_swd
adapter speed: 300 kHz
adapter_nsrst_delay: 100
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
none separate
srst_only separate srst_nogate srst_open_drain connect_assert_srst
Info : Unable to match requested speed 300 kHz, using 240 kHz
Info : Unable to match requested speed 300 kHz, using 240 kHz
Info : clock speed 240 kHz
Info : STLINK v2 JTAG v28 API v2 SWIM v18 VID 0x0483 PID 0x374B
Info : using stlink api v2
Info : Target voltage: 3.252174
Info : stm32l0.cpu: hardware has 4 breakpoints, 2 watchpoints
    TargetName         Type       Endian TapName            State       
--  ------------------ ---------- ------ ------------------ ------------
 0* stm32l0.cpu        hla_target little stm32l0.cpu        reset
Info : Unable to match requested speed 300 kHz, using 240 kHz
Info : Unable to match requested speed 300 kHz, using 240 kHz
adapter speed: 240 kHz
target halted due to debug-request, current mode: Thread 
xPSR: 0xf1000000 pc: 0x08000414 msp: 0x20000200
auto erase enabled
Info : Device: STM32L0xx (Cat.5)
[...]

PR #10213, opened 2018-10-19
boards/common: change openocd config file for stm32l1x

kYc0o:
### Contribution description
The nucleo-l152re is the only stm32l1x supported board, and it hast two flash banks according to openocd config. Thus, adding the right config file makes the whole flash available.

### Testing procedure

Try to flash a very large firmware (more than 256KB) and it fails on current master, since openocd is not able to write on the second bank (256KB). Normally tests/unittests was enough but with all the recent splitting maybe not anymore.

With this PR it should succeed.

### Issues/PRs references
None.

### DIFF
--- a/boards/nucleo-l152re/dist/openocd.cfg
+++ b/boards/nucleo-l152re/dist/openocd.cfg
@@ -1,0 +1,3 @@
+source [find target/stm32l1x_dual_bank.cfg]
+reset_config srst_only
+$_TARGETNAME configure -rtos auto


aabadie:
Nice one! I checked and dual bank support for stm32l1x is supported by openocd 0.10.

Note that a similar fix could be proposed for #10206, but requires a more recent version of openocd. The dual bank support for stm32l0 was added in this [commit](https://github.com/ntfreak/openocd/commit/b375052c58a4ef9b855ce9aa119cc1abf432d0c8) which is 3 days older than the 0.10 release...

> The nucleo-l152re is the only stm32l1x supported board

No, there are also limifrog-v1, lobaro-lorabox and nz32-sc151. Only limifrog-v1 also uses openocd so it would also benefit from the dual bank fix proposed by this PR.

aabadie:
I've looked more closely at the stm32l1x cpus and in fact only stm32l152re has a dual bank flash and only nucleo-l152re has one of these. So there's no need for adapting other l1 based boards like limifrog-v1.

kYc0o:
Yeah I was looking too and found those ones, then we need a way to distinguish them regarding their ROM banks.

aabadie:
I've just tested this PR on a nucleo-l152re. It works but OpenOCD is reporting error while flashing:

```
[...]
hla_swd
adapter speed: 300 kHz
adapter_nsrst_delay: 100
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
none separate
srst_only separate srst_nogate srst_open_drain connect_deassert_srst
Info : Unable to match requested speed 300 kHz, using 240 kHz
Info : Unable to match requested speed 300 kHz, using 240 kHz
Info : clock speed 240 kHz
Info : STLINK v2 JTAG v25 API v2 SWIM v14 VID 0x0483 PID 0x374B
Info : using stlink api v2
Info : Target voltage: 3.264951
Error: jtag status contains invalid mode value - communication failure
Polling target stm32l1.cpu failed, trying to reexamine
Examination failed, GDB will be halted. Polling again in 100ms
Info : Previous state query failed, trying to reconnect
Error: jtag status contains invalid mode value - communication failure
Polling target stm32l1.cpu failed, trying to reexamine
Examination failed, GDB will be halted. Polling again in 300ms
Info : Listening on port 43481 for gdb connections
    TargetName         Type       Endian TapName            State       
--  ------------------ ---------- ------ ------------------ ------------
 0* stm32l1.cpu        hla_target little stm32l1.cpu        unknown
Info : Unable to match requested speed 300 kHz, using 240 kHz
Info : Unable to match requested speed 300 kHz, using 240 kHz
adapter speed: 240 kHz
Error: mem2array: Read @ 0xe0042004, w=4, cnt=1, failed
/usr/local/bin/../share/openocd/scripts/mem_helper.tcl:6: Error: 
in procedure 'reset' 
in procedure 'ocd_bouncer' 
in procedure 'ocd_process_reset' 
in procedure 'ocd_process_reset_inner' called at file "embedded:startup.tcl", line 248
in procedure 'stm32l1.cpu' called at file "embedded:startup.tcl", line 299
in procedure 'ocd_bouncer' 
in procedure 'mmw' 
in procedure 'mrw' called at file "/usr/local/bin/../share/openocd/scripts/mem_helper.tcl", line 25
at file "/usr/local/bin/../share/openocd/scripts/mem_helper.tcl", line 6

Info : Previous state query failed, trying to reconnect
target halted due to debug-request, current mode: Thread 
xPSR: 0x01000000 pc: 0x08000454 msp: 0x20000200
Polling target stm32l1.cpu failed, trying to reexamine
Info : stm32l1.cpu: hardware has 6 breakpoints, 4 watchpoints
auto erase enabled
Info : Device: STM32L1xx (Cat.5/Cat.6)
Info : STM32L flash has dual banks. Bank (0) size is 256kb, base address is 0x8000000
target halted due to breakpoint, current mode: Thread 
xPSR: 0x61000000 pc: 0x2000000e msp: 0x20000200
target halted due to breakpoint, current mode: Thread 
xPSR: 0x61000000 pc: 0x2000000e msp: 0x20000200
[...]
```

These messages are not there in master.

kYc0o:
I modified the PR to only cover the nucleo-l152re board.

kYc0o:
>  It works but OpenOCD is reporting error while flashing

That error looks like your board was in a weird state before being flashed, something was disturbing openocd from getting information from the CPU (e.g. `State Unknown`).

PR #10218, opened 2018-10-21
crypto/helper: Change equals arguments to const

bergzand:
### Contribution description

Small cleanup changing the arguments to the `crypto_equals` function to const.

### Testing procedure

Should all be fine if murdock agrees.

### Issues/PRs references

None


bergzand:
Thanks!

PR #10219, opened 2018-10-21
crypto/helper: Add secure wipe function

bergzand:
### Contribution description

Adds a cryptographically secure wipe function to wipe structs with
sensitive data. Works by first casting the pointer to a `volatile`
pointer to ensure that the compiler doesn't optimize the "memset" away.

### Testing procedure

For now it is mostly if this compiles and if the source makes sense

### Issues/PRs references

None

### DIFF
--- a/sys/include/crypto/helper.h
+++ b/sys/include/crypto/helper.h
@@ -49,6 +49,21 @@ void crypto_block_inc_ctr(uint8_t block[16], int L);
  */
 int crypto_equals(uint8_t *a, uint8_t *b, size_t len);
 
+/**
+ * @brief   Secure wipe function.
+ *
+ * This wipe function zeros the supplied buffer in a way that the compiler is
+ * not allowed to optimize. This can be used to erase secrets from memory.
+ *
+ * Note that this function on its own could be insufficient against (data
+ * remanence) attacks. It is outside the scope of this function to thoroughly
+ * shred the memory area.
+ *
+ * @param[in]   buf     buffer to wipe
+ * @param[in]   len     size of the buffer in bytes
+ */
+void crypto_secure_wipe(void *buf, size_t len);
+
 #ifdef __cplusplus
 }
 #endif
--- a/sys/crypto/helper.c
+++ b/sys/crypto/helper.c
@@ -33,3 +33,12 @@ int crypto_equals(uint8_t *a, uint8_t *b, size_t len)
 
     return diff;
 }
+
+/* Compiler should not be allowed to optimize this */
+void crypto_secure_wipe(void *buf, size_t len)
+{
+    volatile uint8_t *vbuf = (uint8_t*)buf;
+    for (size_t i = 0; i < len; i++) {
+        vbuf[i] = 0;
+    }
+}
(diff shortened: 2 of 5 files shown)


basilfx:
What about data remanence attacks? We should either list that this isn't taken care of, or implement a more secure wipe (multiple times, random data, memory type dependent).

I guess the first is the easiest option :-)

basilfx:
What about data remanence attacks? We should either list that this isn't taken care of, or implement a more secure wipe (multiple times, random data, memory type dependent).

I guess the first is the easiest option :-)

bergzand:
Yeah, let's go with the first option. 

Most crypto libraries I know of handle a "secure" wipe this way and only protect against normal memory reads and not against other memory black magic :)

bergzand:
I'm thinking of renaming it to `crypto_secure_wipe` to prevent name collisions with at least Monocypher. Other suggestions are welcome

Marking as WIP to prevent merging before the rename

bergzand:
@basilfx I've added a small note to the function. I've also renamed the function to hopefully avoid naming collision.

sys/include/crypto/helper.h
```diff
@@ -50,12 +50,18 @@ void crypto_block_inc_ctr(uint8_t block[16], int L);
 int crypto_equals(uint8_t *a, uint8_t *b, size_t len);
 
 /**
- * @brief   Secure wipe function
+ * @brief   Secure wipe function.
+ *
+ * This wipe function zeros the supplied buffer in a way that the compiler is
+ * not allowed to optimize. This can be used to erase secrets from memory.
+ *
+ * Note that this function on its owm could be insufficient. It is outside the
```
basilfx: Typo ;-)

basilfx: maybe add:

`.. could be insufficient against (data remanence) attacks.`

This clarifies why it is insufficient.

bergzand: Fixed, thanks for the suggestion

bergzand: Fixed!

basilfx:
I'm OK with the changes, looks good. Haven't tested it though.

bergzand:
I can add a quick unittest for this, only issue is that verifying the result is a read operation which also  prevents the compiler from optimizing the wipe away. But it would test that the memory region is zero'd out.

basilfx:
Would make sense, since a unit test verifies the code, not the compiler specifics ;-)

bergzand:
Added a simple test to the crypto unittest.

bergzand:
> It is not being optimized away (at least not now).

@jcarrano Thanks for checking. Is this with or without the test assert reading the values after the call?

jcarrano:
Oops! It was with the read. I commented it and dissassembled it again. The code is still there, so it is still secure.

It is declared as volatile, so it *should* get written eventually. The user just needs to be aware of what reorderings are possible (there is a C standard for that).

bergzand:
@basilfx @jcarrano Ready to squash here?

basilfx:
I would say yes. I fully trust @jcarrano checks!

bergzand:
squashed!

bergzand:
@jcarrano Mind pressing the big green button here? :)

bergzand:
@jcarrano Thanks!

PR #10222, opened 2018-10-22
net: add CoRE RD lookup client implementation

pokgak:
### Contribution description

This PR add CoRE Resource Directory lookup client functionality to RIOT. The code basically done and should work already but I cannot test this yet because problem on my laptop.

This PR adds RD lookup client functionality to RIOT.

### Testing

For testing we will need 3 components:

1. A resource directory (RD) server. We will use `aiocoap-rd` example from the development version of aicoap. Run following to install, taken from the [aiocoap install page]:
```
pip3 install --upgrade "git+https://github.com/chrysn/aiocoap@0.4b3#egg=aiocoap[all]"
```
2. A resource directory endpoint to register resources to the RD server. We will use `examples/cord_ep`.
3. A resource directory lookup client. This is provided by this PR at `examples/cord_lc`.

Steps:

1. Start the RD server: `aiocoap-rd`
2. Start a RIOT native instance as a RD endpoint and register to the RD server, replacing `RD_SERVER_ADDRESS` with link-local address of `tapbr0` interface: 

```
$ cd $RIOTROOT/examples/cord_ep
$ PORT=tap0 make all term

> cord_ep register [RD_SERVER_ADDRESS]
Registering with RD now, this may take a short while...
RD endpoint event: now registered with a RD
registration successful

CoAP RD connection status:
RD address: coap://[fe80::2cb3:23ff:fee4:fd57]:5683
   ep name: RIOT-6B06232323232323
  lifetime: 60s
    reg if: /resourcedirectory
  location: /reg/1/
```

3. Start a RIOT native instance as a RD lookup client and query the RD for registered endpoints/resources:

```
$ cd $RIOTROOT/examples/cord_lc
$ PORT=tap1 make all term

> cord_lc [RD_SERVER_ADDRESS] resource
cord_lc [fe80::2cb3:23ff:fee4:fd57] resource
Performing lookup now, this may take a short while...
Found resource/endpoint
Target: coap://[fe80::388d:f7ff:fecf:dfcb%tapbr0]/node/info
'anchor': 'coap://[fe80::388d:f7ff:fecf:dfcb%tapbr0]'

# Parsed resouce lookup
> cord_lc [fe80::14ea:3dff:fed9:d818] resource
cord_lc [fe80::14ea:3dff:fed9:d818] resource
Performing lookup now, this may take a short while...
Found resource/endpoint
Target: coap://[fe80::e466:f1ff:fe8e:36f7%tapbr0]/node/info
'anchor': 'coap://[fe80::e466:f1ff:fe8e:36f7%tapbr0]'

# Raw resource lookup
> cord_lc [fe80::14ea:3dff:fed9:d818] -r resource
cord_lc [fe80::14ea:3dff:fed9:d818] -r resource
Performing lookup now, this may take a short while...
finished lookup
Lookup result:
<coap://[fe80::e466:f1ff:fe8e:36f7%tapbr0]/node/info>;anchor="coap://[fe80::e466:f1ff:fe8e:36f7%tapbr0]",<coap://[fe80::e466:f1ff:fe8e:36f7%tapbr0]/sense/hum>;anchor="coap://[fe80::e466:f1ff:fe8e:36f7%tapbr0]",<coap://[fe80::e466:f1ff:fe8e:36f7%tapbr0]/sense/temp>;anchor="coap://[fe80::e466:f1ff:fe8e:36f7%tapbr0]"

```


Makefile.dep
```diff
@@ -773,6 +773,15 @@ ifneq (,$(filter cord_ep,$(USEMODULE)))
   endif
 endif
 
+ifneq (,$(filter cord_lc,$(USEMODULE)))
+  USEMODULE += cord_common
+  USEMODULE += core_thread_flags
+  USEMODULE += gcoap
```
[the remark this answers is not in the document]
pokgak: You mean new `cord` module? Does it make sense to put this into `cord_common` instead and `cord_ep` and `cord_lc` can use `cord_common` as dependency?

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,264 @@
...
+
+static uint8_t buf[BUFSIZE];
+
+static void _lock(void)
+{
+    mutex_lock(&_mutex);
+    _waiter = sched_active_thread;
+}
+
+static int _sync(void)
```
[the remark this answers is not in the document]
pokgak: How about moving both `_lock` and `_sync` to `cord_common` and rename to `cord_common_lock` and `cord_common_sync`?

sys/include/net/cord/config.h
```diff
@@ -90,6 +90,20 @@ extern "C" {
...
+#ifndef CORD_LC_MAX_SCRITS
+#define CORD_LC_MAX_SCRITS      (1)
+#endif
+
+/**
+ * @brief  Default length of lookup result buffer
+ */
+#ifndef CORD_LC_RES_BUF_LEN
+#define CORD_LC_RES_BUF_LEN     (128)
+#endif
```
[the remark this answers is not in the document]
pokgak: Fixed that.

Makefile.dep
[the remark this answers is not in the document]
haukepetersen: So we could add this below (and remove the corresponding USEMODULE statements accordingly):
```
ifneq (,$(filter cord_lc cord_ep,$(USEMODULE)))
  ifneq (,$(filter shell_commands,$(USEMODULE)))
    USEMODULE += sock_util
  endif
endif
```
Also we might want to move the `gcoap` 'include' into `cord_common`, as you suggested.

haukepetersen:
> Both cord_ep and cord_lc need to first discover details about the RD. 

True, but I would like to keep both submodules more independent of each other. Also, for the current version of `cord_ep`, only interacting with a single RD at the time is possible (mostly due to (partly previous) restrictions in `gcoap`). So here it does not really make sense to use that proposed state struct for now.

As of now, I tend to keep the `rd_init()` part tied to the `cord_lc_x` submodule. But we can always refactor this at a later point in time. I have however been thinking of generalizing the `cord_ep` interface towards being able to interact with multiple RDs (and opened PRs for `gcaop` on this). Once this is done, we should talk again :-)

> I will work on this next week and hopefully be ready for next round of review. =)

Very nice, looking forward to it!

Makefile.dep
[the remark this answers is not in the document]
pokgak: I pushed a commit (37bd548da) addressing this comment but I'm not sure I understand your last comment correctly. Please correct me if there is anything wrong with it.

examples/cord_lc/main.c
```diff
@@ -0,0 +1,41 @@
...
+#include "msg.h"
+
+#define MAIN_QUEUE_SIZE     (8)
+static msg_t _main_msg_queue[MAIN_QUEUE_SIZE];
+
+int main(void)
+{
+    /* we need a message queue for the thread running the shell in order to
+     * receive potentially fast incoming networking packets */
+    msg_init_queue(_main_msg_queue, MAIN_QUEUE_SIZE);
```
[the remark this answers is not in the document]
pokgak: I removed that section and tested it by pinging to itself and got a FAILED_ASSERTION. So maybe it is still needed. Note: I'm using master branch from Oct 21.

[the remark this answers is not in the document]
miri64: > If I remember correctly, @miri64 told me that this is not needed anymore, right @miri64?!

If you are not using the message queue of the thread for netapi, that is correct, yes. Since `sock`s use their own message queues (mbox) they don't need it.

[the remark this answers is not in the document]
miri64: (`sock` uses `mbox` since its inception, so your version *should* work without if *if* you are not using vanilla `gnrc_netapi`).

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -191,70 +209,241 @@ static void _on_lookup(unsigned req_state, coap_pkt_t *pdu,
...
+        strncpy(ep->base, val, sizeof(ep->base));
+    }
+    else if (strcmp(attr, "ep") == 0) {
+        strncpy(ep->ep, val, sizeof(ep->ep));
+    }
+
+    /* ignore unrecognized attributes */
+}
+
+static size_t _parse_res(const char *source, size_t len, cord_lc_res_t *results,
```
pokgak: _parse_res and _parse_ep basically do the same thing - parses the result buffer from _lookup_raw and saves the endpoints/resources to respective structs. I tried to combine the two functions together so that there is less redundant code but cannot find a good solution. Suggestions are welcomed.

pokgak:
@haukepetersen I pushed the new code, now based on your suggestion.
Excuse the force-pushes, I missed some extra whitespace and some old code that are not needed anymore during the first push.

sys/include/net/cord/lc.h
```diff
@@ -0,0 +1,181 @@
...
+ * @param[in]  content_format   data format wanted from RD server
+ * @param[out] result           buffer holding the raw response
+ * @param[in]  maxlen           max length of @p result
+ *
+ * @return length of result on success
+ * @return CORD_LC_NORSC if there is no resource lookup interface at @p rd
+ * @return CORD_LC_TIMEOUT if lookup timed out
+ * @return CORD_LC_ERR on any other internal error
+ */
+ssize_t cord_lc_res_raw(const cord_lc_rd_t *rd, uint8_t content_format,
```
kb2ma: An array of `clif_param_t` parameter structs from #11189 could be used as the input filter.

sys/include/net/cord/lc.h
```diff
@@ -0,0 +1,181 @@
...
+ * @param[in]  rd               the target RD server
+ * @param[out] resources        list of resources at @p rd
+ * @param[in]  limit            max number of resources prepared to receive
+ *
+ * @return number of found resources on success
+ * @return CORD_LC_NORSC if there is no resource lookup interface at @p rd
+ * @return CORD_LC_TIMEOUT if lookup timed out
+ * @return CORD_LC_ERR on any other internal error
+ */
+ssize_t cord_lc_res(const cord_lc_rd_t *rd, cord_lc_res_t *resources, size_t limit);
```
kb2ma: This function is resource intensive. I would consider an iterative function on the collected string from coap_lc_res_raw().

```
cord_lc_read_res(const char *source, clif_t *link)
```
where link is the output parameter. However, this function is similar to `coap_decode_links()` from #11189, so not that useful. 

On the other hand, I would not be surprised to see a block based response from the RD. How to incorporate the possibility into the API? Rather than separately call `cord_lc_res_raw()` first, maybe something like this:

```
cord_lc_read_res(const cord_lc_rd_t *rd, char *input, unsigned input_len, clif_t *link)
```
where the cord_lc_rd_t param could be extended to include the block number. That means it could transparently retrieve the next block.

pokgak: > This function is resource intensive. I would consider an iterative function on the collected string from coap_lc_res_raw().

Sorry, I don't quite understand here. What part of the function, do you mean, is resource intensive? Internally, it currently uses _lookup_raw, which are also used by cord_lc_res_raw, and then parses the result and put it in cord_lc_res_t.

I do however find that using iterative function on the collected string is a good idea. As clif_decode_link only decodes one link per call, cord_lc_read_res can be used as a wrapper to it to decode all the links at once and returns it to the user.

> On the other hand, I would not be surprised to see a block based response from the RD.

Block based response would be nice actually. The default GCOAP_PDU_BUF_SIZE is too small even for an RD ep with only 3 resources. Currently, I just set the buffer size bigger but using block option would solve this, I think.

I found #11056 but haven't take a deeper look at it and doesn't really know how to use the API and integrate it here yet, but as you understands it better than me, I'll keep your suggestion in mind ;)

sys/include/net/cord/lc.h
pokgak: Yup, lots of the code in this PR can be replaced by clif e.g. clif_t instead of cord_lc_res_t and cord_lc_ep_t as both of them is just links actually. I think there's a lot more than can replaced.

sys/include/net/cord/lc.h
kb2ma: > As clif_decode_link only decodes one link per call, cord_lc_read_res can be used as a wrapper to it to decode all the links at once and returns it to the user.

That's the resource intensive part. ;-) By iterative I meant that it would be less intensive for the user to repeatedly call `cord_lc_read_res()` with the output from the query, and receive one clif_t link per call. Maybe the cord_lc_rd_t parameter could remember the last character read from the query result, as well as remember the last block number it retrieved.

Let's wait for clif, and I'd be happy to discuss more.

pokgak: Aah okay. That's what you mean.

> Maybe the cord_lc_rd_t parameter could remember the last character read from the query result, as well as remember the last block number it retrieved.

There are actually [page and count parameters][1] that can be used for the lookup. So this should be easier to implement. Just as a note for now.

[1]: https://tools.ietf.org/html/draft-ietf-core-resource-directory-20#section-6.2

kb2ma: Thanks for the link. I actually saw that earlier but thought, "why do we need this when we already have block"? Looking at it more closely, I see that the interface is super simple for a client and eliminates the need for the client to stitch 'sliced' links together from block's fixed size response. On the other hand, if packet space is really limited, block provides certainty in the quantity of data received.

It might actually be worthwhile to ask the core mailing list about the reasoning. If we're concerned about data size, use of CBOR would be valuable here, but I guess that use is orthogonal to how the data is sliced. I also agree with you that we don't need to resolve this issue today.

examples/cord_lc/main.c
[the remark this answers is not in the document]
miri64: (it is however still needed for the `ping6` command, if you include that)

tcschmidt:
@pokgak @kb2ma @haukepetersen any reason why this is in rest? I think it is worth completing.

pokgak:
Sorry forgot to update the current status. I'm waiting for #11189  to be merged and so that it can be used for parsing link format. My current implementation is a bit hacky IMHO.

pokgak:
I updated this PR to use the CLIF module (#11189 ) to parse link format response from RD server.  This is currently ready for review.

I'm trying to write some script/Dockerfile so that this can be tested directly without having to setup the test dependencies first. I'll update this later when this is ready.

tcschmidt:
@haukepetersen @kb2ma are you available to move this forward?

kb2ma:
It's going to be a little while before I can look at this. The Wakaama package and @pokgak's integration of DTLS with gcoap are higher priorities right now.

If only I could spend more time on RIOT, things would happen faster. ;-)

tcschmidt:
> It's going to be a little while before I can look at this. The Wakaama package and @pokgak's integration of DTLS with gcoap are higher priorities right now.
> 
> If only I could spend more time on RIOT, things would happen faster. ;-)

That's the problem for most of us ;) - maybe @smlng can kick off?

pokgak:
Updated the OP with steps for testing.

examples/cord_lc/Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+# The default size of GCOAP_PDU_BUF_SIZE can be too small for the response
+# from the RD server. Use a larger value if the client drops response packet
+# from RD server.
+CFLAGS += -DGCOAP_PDU_BUF_SIZE=2048
+
+# Comment this out to disable code in RIOT that does safety checking
+# which is not needed in a production environment but helps in the
+# development process:
+CFLAGS += -DDEVELHELP
```
smlng: this was changed to `DEVELHELP ?= 1`, a while ago

examples/cord_lc/cord_lc_cli.c
```diff
@@ -0,0 +1,194 @@
...
+        }
+        if (retval < 0) {
+            printf("Error during lookup %d\n", retval);
+            return -1;
+        }
+        printf("Lookup result:\n%.*s\n", retval, bufpool);
+    }
+    else if (!strcmp(argv[2], "resource")) {
+        cord_lc_res_t resource;
+	clif_attr_t attrs[5];
```
smlng: indention is wrong here

examples/cord_lc/cord_lc_cli.c
```diff
@@ -0,0 +1,194 @@
...
+                             sizeof(bufpool));
+        if (retval < 0) {
+            printf("Error during lookup %d\n", retval);
+            return -1;
+        }
+        _print_lookup_result(&resource);
+    }
+    else if (!strcmp(argv[2], "endpoint")) {
+        cord_lc_ep_t endpoint;
+	clif_attr_t attrs[5];
```
smlng: same here, copy-pasta?

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,421 @@
...
+
+    /* add filters */
+    cord_lc_filter_t *f = filters;
+    while (f) {
+        for (unsigned i = 0; i < f->len; i++) {
+	    char key[32];
+	    char val[32];
+	    snprintf(key, f->array[i].key_len + 1, "%s", f->array[i].key);
+	    snprintf(val, f->array[i].value_len + 1, "%s", f->array[i].value);
+            gcoap_add_qstring(&pkt, key, val);
```
smlng: indention

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,421 @@
...
+        break;
+    }
+
+    res = gcoap_req_init(&pkt, buf, sizeof(buf), COAP_METHOD_GET, lookif);
+    if (res < 0) {
+        retval = CORD_LC_ERR;
+        DEBUG("cord_lc: failed gcoap_req_init()\n");
+        goto end;
+    }
+
```
smlng: remove dup empty line

examples/cord_lc/Makefile
pokgak: Fixed.

examples/cord_lc/cord_lc_cli.c
pokgak: Fixed.

examples/cord_lc/cord_lc_cli.c
pokgak: Fixed.

sys/net/application_layer/cord/lc/cord_lc.c
pokgak: fixed

sys/net/application_layer/cord/lc/cord_lc.c
pokgak: fixed.

pokgak:
> There is lots of code duplicates for cord_lc_ep and cord_lc_res, I think this could simplified with a common internal function.

I addressed this in b073f8635 but am not sure if that is easy to understand or not because `cord_lc_res()` will call `_lookup_result()` which then calls another internal function `_lookup_raw()`.

pokgak:
Fixed the error from Travis CI and rebased to master.

smlng:
is this still WIP?

pokgak:
Nope. It is now ready for review :)

tcschmidt:
@haukepetersen It looks like your comments are addressed, so can your block be removed?
Otherwise: What about testing? I suppose, this is a candidate for the next release ...

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,432 @@
...
+    thread_flags_set((thread_t *)_waiter, flag);
+}
+
+static ssize_t _add_filters_to_lookup(coap_pkt_t *pkt, cord_lc_filter_t *filters)
+{
+    cord_lc_filter_t *f = filters;
+    while (f) {
+        for (unsigned i = 0; i < f->len; i++) {
+            char key[NANOCOAP_QS_MAX];
+            char val[NANOCOAP_QS_MAX];
```
haukepetersen: Allocating 128bytes per default (and possibly more if someone decides to redefine the value of `NANOCOAP_QS_MAX`) is not the best idea. Using a statically allocated buffer is IMO the better option here (see e.g. https://github.com/RIOT-OS/RIOT/pull/12774 for the problems this could cause)

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,432 @@
...
+
+            clif_attr_t *kv = &(f->array[i]);
+            if ((kv->key_len + kv->value_len + 1) >= NANOCOAP_QS_MAX ) {
+                DEBUG("cord_lc: filter key value pair overflow\n");
+                return CORD_LC_OVERFLOW;
+            }
+            snprintf(key, kv->key_len + 1, "%s", kv->key);
+            snprintf(val, kv->value_len + 1, "%s", kv->value);
```
haukepetersen: Looking at his, I am confused: what is the purpose of copying `kv->key` and `kv->value` to the intermediate buffers? Is is just to get a `\0` termination for the strings? If so, this method seems awefully expensive to me... 

I might be wrong, but aren't those values already `\0` terminated by the `clif` module? But if I am wrong, then the better way to solve this is to add an additional, lower-level function to `gcoap` that accepts non-`\0`-terminated strings, e.g. `gcoap_add_qstring(coap_pkt_t *pdu, const char *key, size_t key_len, const char *val, size_t val_len)`. The existing `gcoap_add_qstring()` can then simply call this function after figuring out the buffer sizes using `strlen`.

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,432 @@
...
+    return CORD_LC_OK;
+}
+
+static ssize_t _lookup_raw(const cord_lc_rd_t *rd, unsigned content_format,
+                           unsigned lookup_type, cord_lc_filter_t *filters,
+                           void *result, size_t maxlen)
+{
+    assert(rd->remote);
+
+    uint8_t buf[GCOAP_PDU_BUF_SIZE];
```
haukepetersen: same here, not a good practice to allocate this on the stack.

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,432 @@
...
+    int res;
+    int retval;
+    coap_pkt_t pkt;
+    ssize_t pkt_len;
+
+    char lookif[NANOCOAP_QS_MAX];
+    memset(lookif, 0, sizeof(lookif));
+    switch (lookup_type) {
+    case CORD_LC_RES:
+        memcpy(lookif, rd->res_lookif, rd->res_lookif_len);
```
haukepetersen: why copy this value? A simple pointer should do the trick. Like this its just a waste of memory.

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,432 @@
...
+    case CORD_LC_EP:
+        memcpy(lookif, rd->ep_lookif, rd->ep_lookif_len);
+        break;
+    }
+
+    res = gcoap_req_init(&pkt, buf, sizeof(buf), COAP_METHOD_GET, lookif);
+    if (res < 0) {
+        retval = CORD_LC_ERR;
+        DEBUG("cord_lc: failed gcoap_req_init()\n");
+        goto end;
```
haukepetersen: No need for using `goto` here -> simple put `return CORD_LC_ERR` here. `goto` would only be viable if there is some repetitive cleanup that needs to be done before returning...

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,432 @@
...
+        DEBUG("cord_lc: unsupported content format\n");
+        return CORD_LC_ERR;
+    }
+    coap_hdr_set_type(pkt.hdr, COAP_TYPE_CON);
+    coap_opt_add_uint(&pkt, COAP_OPT_ACCEPT, content_format);
+
+    pkt_len = gcoap_finish(&pkt, 0, COAP_FORMAT_NONE);
+    if (pkt_len < 0) {
+        retval = CORD_LC_ERR;
+        goto end;
```
haukepetersen: see above, no need for `goto`

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,432 @@
...
+    }
+    res = gcoap_req_send(buf, pkt_len, rd->remote, _on_lookup, NULL);
+    if (res < 0) {
+        retval = CORD_LC_ERR;
+        goto end;
```
haukepetersen: `goto`...

sys/include/net/cord/lc.h
```diff
@@ -0,0 +1,193 @@
...
+ * @ingroup     net_cord
+ * @brief       Library for using RIOT as CoRE Resource Directory lookup client
+ *
+ * This module implements a CoRE Resource Directory lookup client library, that
+ * allows RIOT nodes to lookup resources, endpoints and groups with resource
+ * directories.
+ * It implements the standard lookup functionality as defined in
+ * draft-ietf-core-resource-directory-20.
+ * @see https://tools.ietf.org/html/draft-ietf-core-resource-directory-20
+ *
```
haukepetersen: I think it would be good to add a paragraph documenting the limitations of this implementation, e.g. that the user is not able to make more than a single request concurrently and that lookup function calls are fully synchronously.

By the way: there is actually no reason for this anymore. Back when I implemented `core_ep`, there were issues with `gcoap` that prevented me from using it asynchronously. But those are no longer existing, so we could actually rewrite the code to lift these limitations. But that is probably best done in some future work :-)

sys/net/application_layer/cord/lc/cord_lc.c
```diff
@@ -0,0 +1,432 @@
...
+        DEBUG("cord_lc: no endpoint link found\n");
+        retval = CORD_LC_ERR;
+    }
+    return retval;
+}
+
+ssize_t cord_lc_res(cord_lc_rd_t *rd, cord_lc_res_t *resource,
+                    cord_lc_filter_t *filters, void *buf, size_t maxlen)
+{
+    _lock();
```
haukepetersen: You can simplify this code a bit: move the `_lock()` and `mutex_unlock()` calls to `_lookup_result()`, and move `cord_lc_res()` and `cord_lc_ep()` into the header file and make them `static inline` as they are simple one-liners then.

sys/net/application_layer/cord/lc/cord_lc.c
pokgak: fixed in f0857f7e0

sys/net/application_layer/cord/lc/cord_lc.c
pokgak: fixed in f0857f7e0

sys/net/application_layer/cord/lc/cord_lc.c
pokgak: fixed in f0857f7e0

Makefile.dep
```diff
@@ -882,16 +881,21 @@ endif
...
+ifneq (,$(filter cord_lc,$(USEMODULE)))
+  USEMODULE += cord_common
+  USEMODULE += clif
 endif
 
 ifneq (,$(filter cord_common,$(USEMODULE)))
   USEMODULE += fmt
   USEMODULE += luid
+  USEMODULE += gcoap
+  USEMODULE += core_thread_flags
```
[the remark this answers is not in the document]
pokgak: I cannot come out with a way to factor these dependencies only fo cord_ep and cord_lc. I moved this out from cord_common for now.

examples/cord_lc/README.md
```diff
@@ -0,0 +1,13 @@
+CoRE Resource Directory: Lookup Example
+=========================================
+
+This example application demonstrates the usage of RIOT's Resource Directory
+(RD) lookup module, called `cord_lc`. This module supports the lookup of
+resources and endpoints as defined in
+[draft-ietf-core-resource-directory-20](https://tools.ietf.org/html/draft-ietf-core-resource-directory-20).
```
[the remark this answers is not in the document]
pokgak: Changed the link to point to version 23. See 4e14b195b

sys/net/application_layer/cord/lc/cord_lc.c
pokgak: > Looking at his, I am confused: what is the purpose of copying `kv->key` and `kv->value` to the intermediate buffers? Is is just to get a `\0` termination for the strings?

Yes. AFAIK `clif` does not assume that the string is `\0` terminated and always defines the length of the string with the pointer. Correct me if I'm wrong here @leandrolanzieri.

Adding a lower-level function to gcoap seems like the acceptable solution here. I'll try something out and make a PR for that.

sys/include/net/cord/lc.h
pokgak: Added doc documenting the limitation and also the different lookup modes available in 6f7d2e17a6ab3c38c3cfd2ffa15e5bb1e5dfb138.

> By the way: there is actually no reason for this anymore. Back when I implemented `core_ep`, there were issues with `gcoap` that prevented me from using it asynchronously. But those are no longer existing, so we could actually rewrite the code to lift these limitations. But that is probably best done in some future work :-)

:+1:

sys/net/application_layer/cord/lc/cord_lc.c
pokgak: If I understand you correctly, doing that will require `_lookup_result()` to be moved to the header file too, which is not good I guess because it should only be a helper function and not supposed to be exposed outside of the file.

sys/include/net/cord/lc.h
kb2ma: > Back when I implemented `core_ep`, there were issues with `gcoap` that prevented me from using it asynchronously.

Perhaps an async approach would help with #12884.

haukepetersen: It seems to me that #12884 is unrelated to this and can be fixed on its own -> will comment there

sys/net/application_layer/cord/lc/cord_lc.c
haukepetersen: Yes. I don't see any problem with exposing `_lookup_result()` in the public API. There is nothing in this function signature that would prevent users to use it directly, in the contrary, there might even be possible use cases where it could come in handy to programmatically define the type parameter...

But even if you don't like this and leave it like it is, you should move the mutex related lines into `_lookup_result()` to reduce some code duplication...

sys/net/application_layer/cord/lc/cord_lc.c
pokgak: Opened #13022 to add a `gcoap_add_qstring2()` function to gcoap.

tcschmidt:
@pokgak @haukepetersen are issues addressed? Is this ready for progress?

examples/cord_lc/cord_lc_cli.c
```diff
@@ -0,0 +1,177 @@
...
+#include <stdlib.h>
+
+#include "net/cord/config.h"
+#include "net/cord/lc.h"
+#include "net/gcoap.h"
+#include "net/sock/util.h"
+
+#define CORD_SERVER_PORT    (5683)
+#define FILTER_BUF_SIZE     (64)
+#define RESULT_BUF_SIZE     (1024)
```
smlng: not used

examples/cord_lc/cord_lc_cli.c
```diff
@@ -0,0 +1,177 @@
...
+}
+
+static void _print_usage(void) {
+    puts("usage: cord_lc <server_addr> "
+         "{ resource | endpoint | raw { resource | endpoint } [key=value]"
+         "example: cord_lc [2001:db8:3::dead:beef]:5683 raw resource count=1 page=2");
+}
+
+int cord_lc_cli_cmd(int argc, char **argv) {
+    char bufpool[1024];
```
smlng: maybe you want to use `RESULT_BUF_SIZE` here instead

sys/include/net/cord/lc.h
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * @brief  Information about RD server and its lookup interface resources
+ */
+typedef struct {
+    const sock_udp_ep_t *remote;        /**< Remote endpoint of RD server */
+    char *res_lookif;                   /**< Resource lookup interface */
+    size_t res_lookif_len;              /**< Length of @p res_lookif */
+    char *ep_lookif;                    /**< Endpoint lookup interface */
+    size_t ep_lookif_len;               /**< Length of @p ep_lookif */
```
smlng: looking below these should be `\0` terminated strings to be used as path in `gcoap_req_init`, hence you can omit the length attribute and the memcpy below

sys/net/application_layer/cord/lc/cord_lc.c
smlng: I agree, though it requires `res_lookif` and `ep_lookif` to be null terminated strings, but I guess its worth 1 byte to save the length param and the `lookif` buffer here plus the `memcpy`, too.

pokgak:
> @pokgak @haukepetersen are issues addressed? Is this ready for progress?

I addressed some of the issues (opened #13022 for https://github.com/RIOT-OS/RIOT/pull/10222#discussion_r358257934) but there are still other issues.

I cannot focus more on this until next month. Sadly this PR been open for too long. I'll be focusing to get this merge in February.

pokgak:
@smlng @haukepetersen 

I pushed commits addressing the comments. Rebased it on the latest master and squashed it directly because otherwise, it will show too many unrelated commits. Hope you don't mind.

Summary of changes that I did since last time:

- removed `res_lookif_len` and `ep_lookif_len` from `cord_lc_rd_t`. This also removes `memcpy` from other parts of the code as pointed in the last review.
- removed unused defines in `cord_lc_cli.c`
- used `coap_opt_add_uquery2` from #13022 
- moved allocation of buffers out of the stack

tcschmidt:
@pokgak we need someone for testing this. Do you want to poke around?

examples/cord_lc/cord_lc_cli.c
```diff
@@ -0,0 +1,168 @@
...
+static sock_udp_ep_t remote;
+static char rdbuf[2 * NANOCOAP_URI_MAX] = {0};
+static unsigned rd_initialized = 0;
+
+static int _make_sock_ep(sock_udp_ep_t *ep, const char *addr) {
+    ep->port = 0;
+    if (sock_udp_str2ep(ep, addr) < 0) {
+        return -1;
+    }
+    ep->family = AF_INET6;
```
smlng: is already set by `sock_udp_str2ep` according to given `addr` string, and should not be manually be overriden, bc in theory this could also work with IPv4.

examples/cord_lc/cord_lc_cli.c
```diff
@@ -0,0 +1,168 @@
...
+#include "net/gcoap.h"
+#include "net/sock/util.h"
+
+static cord_lc_rd_t rd;
+static sock_udp_ep_t remote;
+static char rdbuf[2 * NANOCOAP_URI_MAX] = {0};
+static unsigned rd_initialized = 0;
+
+static int _make_sock_ep(sock_udp_ep_t *ep, const char *addr) {
+    ep->port = 0;
```
smlng: no needed, ep is initialised with zeros in `sock_udp_str2ep`

examples/cord_lc/cord_lc_cli.c
```diff
@@ -0,0 +1,168 @@
...
+        if (!key_end) {
+            f->key_len = strlen(f->key);
+            f->value = NULL;
+            f->value_len = 0;
+            continue;
+        }
+
+        f->key_len = key_end - f->key;
+        f->value = key_end + 1;
+        f->value_len = strlen(f->value);
```
smlng: use `else` instead

```suggestion
        }
        else {
            f->key_len = key_end - f->key;
            f->value = key_end + 1;
            f->value_len = strlen(f->value);
        }
```

tcschmidt:
> 
> 
> needs rebase and #13022 merged first

With #13022 being merged and rebase being straightforward, we are back with looking for testers.

examples/cord_lc/cord_lc_cli.c
pokgak: fixed.

examples/cord_lc/cord_lc_cli.c
pokgak: fixed

examples/cord_lc/cord_lc_cli.c
pokgak: fixed

examples/cord_lc/cord_lc_cli.c
```diff
@@ -0,0 +1,177 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "net/cord/config.h"
+#include "net/cord/lc.h"
+#include "net/gcoap.h"
+#include "net/sock/util.h"
+
+#define CORD_SERVER_PORT    (5683)
+#define FILTER_BUF_SIZE     (64)
```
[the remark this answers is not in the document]
pokgak: already removed in the latest commit.

pokgak:
I updated the testing instruction and tested it on Ubuntu 19.10.

pokgak:
Fixed the static-test error. Rebased and squashed directly.

tcschmidt:
@haukepetersen are your comments addressed?
@kb2ma do you have capacity for testing?

haukepetersen:
Just tried to give this a quick test: seems there is still something not quite in order with the test application:
```
> cord_lc [fe80::6c05:adff:fe85:121c] raw
cord_lc [fe80::6c05:adff:fe85:121c] raw
Performing lookup now, this may take a short while...
make: *** [/home/hauke/dev/riot/RIOT/examples/cord_lc/../../Makefile.include:685: term] Segmentation fault (core dumped)
```
I guess it should not segfault?! :-)

Did run this under `native` with all prepration steps done as quoted in the PR description...

sys/net/application_layer/cord/lc/cord_lc.c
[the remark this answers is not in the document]
haukepetersen: It would still be nice not to duplicate this code, but then again I don't see this as a show stopper for now. So lets postpone this step until after this PR is merged...

pokgak:
> Just tried to give this a quick test: seems there is still something not quite in order with the test application:

`raw` is supposed to be followed by either `resource` or `endpoint`. Passing in `raw` without either of those two now prints usage info. I guess the command can be confusing. I changed `raw` to `-r` options. gcoap cli example also uses this style (`-c` for CON messages). I also added both example for raw and normal lookup command to the usage info.

haukepetersen:
Could you please rebase (and retest)? After rebasing to master, the PR does not build anymore: Error `NANOCOAP_URI_MAX` undeclared...

haukepetersen:
And when using your branch as is, the example application is still segfaulting:
```
hauke@schelm:~/dev/riot/RIOT/examples/cord_lc$ make term
/home/hauke/dev/riot/RIOT/examples/cord_lc/bin/native/cord_lc.elf tap0 
LED_RED_OFF
LED_GREEN_ON

main(): This is RIOT! (Version: 2020.04-devel-1077-g28f6c-HEAD)
CoRE RD lookup client example!

> help
help
cord_lc              Cord LC example
reboot               Reboot the node
ps                   Prints information about running threads.
ping6                Ping via ICMPv6
random_init          initializes the PRNG
random_get           returns 32 bit of pseudo randomness
nib                  Configure neighbor information base
ifconfig             Configure network interfaces
> cord_lc
cord_lc
usage: cord_lc <server_addr> [-r] { resource | endpoint } [key=value]
Options:
      -r get raw result
example: cord_lc [2001:db8:3::dead:beef]:5683 -r resource count=1 page=2
example: cord_lc [2001:db8:3::dead:beef]:5683 endpoint
> cord_lc [fe80::a0ef:f8ff:fefb:20db]:5683 endpoint   
cord_lc [fe80::a0ef:f8ff:fefb:20db]:5683 endpoint
Performing lookup now, this may take a short while...
make: *** [/home/hauke/dev/riot/RIOT/examples/cord_lc/../../Makefile.include:685: term] Segmentation fault (core dumped
```

pokgak:
I cannot reproduce the segfault on the current version:

```
user@riot-vm:~/RIOT$ PORT=tap0 make -C examples/cord_lc term
make: Entering directory '/home/user/RIOT/examples/cord_lc'
/home/user/RIOT/examples/cord_lc/bin/native/cord_lc.elf tap0
LED_RED_OFF
LED_GREEN_ON

main(): This is RIOT! (Version: 2020.04-devel-1077-g28f6c-pr/add-cord-lc)
CoRE RD lookup client example!

> cord_lc
cord_lc
usage: cord_lc <server_addr> [-r] { resource | endpoint } [key=value]
Options:
      -r get raw result
example: cord_lc [2001:db8:3::dead:beef]:5683 -r resource count=1 page=2
example: cord_lc [2001:db8:3::dead:beef]:5683 endpoint
> cord_lc [fe80::d054:feff:fe4a:124b] resource
cord_lc [fe80::d054:feff:fe4a:124b] resource
Performing lookup now, this may take a short while...
Found resource/endpoint
Target: coap://[fe80::d054:feff:fe4a:124c%tapbr0]/node/info
'anchor': 'coap://[fe80::d054:feff:fe4a:124c%tapbr0]'
> cord_lc [fe80::d054:feff:fe4a:124b] resource
cord_lc [fe80::d054:feff:fe4a:124b] resource
Found resource/endpoint
Target: coap://[fe80::d054:feff:fe4a:124c%tapbr0]/sense/hum
'anchor': 'coap://[fe80::d054:feff:fe4a:124c%tapbr0]'
> cord_lc [fe80::d054:feff:fe4a:124b] -r resource
cord_lc [fe80::d054:feff:fe4a:124b] -r resource
Lookup result:
<coap://[fe80::d054:feff:fe4a:124c%tapbr0]/node/info>;anchor="coap://[fe80::d054:feff:fe4a:124c%tapbr0]",<coap://[fe80::d054:feff:fe4a:124c%tapbr0]/sense/hum>;anchor="coap://[fe80::d054:feff:fe4a:124c%tapbr0]",<coap://[fe80::d054:feff:fe4a:124c%tapbr0]/sense/temp>;anchor="coap://[fe80::d054:feff:fe4a:124c%tapbr0]"
> cord_lc [fe80::d054:feff:fe4a:124b] endpoint
cord_lc [fe80::d054:feff:fe4a:124b] endpoint
Found resource/endpoint
Target: /reg/1/
'ep': 'RIOT-4821232348212323'
'base': 'coap://[fe80::d054:feff:fe4a:124c%tapbr0]'
'rt': 'core.rd-ep'
> cord_lc [fe80::d054:feff:fe4a:124b] -r endpoint
cord_lc [fe80::d054:feff:fe4a:124b] -r endpoint
Lookup result:
</reg/1/>;ep="RIOT-4821232348212323";base="coap://[fe80::d054:feff:fe4a:124c%tapbr0]";rt="core.rd-ep"
```

pokgak:
And after rebasing I got cannot connect to the RD, this also happens on master (e8389a58470edf53800c3a9ad202c25d7c663112). I'll open an issue for this.

pokgak:
> And after rebasing I got cannot connect to the RD, this also happens on master ([e8389a5](https://github.com/RIOT-OS/RIOT/commit/e8389a58470edf53800c3a9ad202c25d7c663112)). I'll open an issue for this.

See #14399

pokgak:
Rebased.

> And after rebasing I got cannot connect to the RD, this also happens on master ([e8389a5](https://github.com/RIOT-OS/RIOT/commit/e8389a58470edf53800c3a9ad202c25d7c663112)). I'll open an issue for this.

Using a global address (e.g `fd00::1`), this PR still works after rebase.

pokgak:
Rebased to include fix for #14399. Retested again. Log:

```
main(): This is RIOT! (Version: 2020.07-devel-1689-g84dea-pr/add-cord-lc)
CoRE RD lookup client example!

> cord_lc [fe80::101a:7dff:fe03:18f6] endpoint
cord_lc [fe80::101a:7dff:fe03:18f6] endpoint
Performing lookup now, this may take a short while...
Found resource/endpoint
Target: /reg/1/
'ep': 'RIOT-9836232398362323'
'base': 'coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]'
'rt': 'core.rd-ep'
> cord_lc [fe80::101a:7dff:fe03:18f6] -r endpoint
cord_lc [fe80::101a:7dff:fe03:18f6] -r endpoint
Lookup result:
</reg/1/>;ep="RIOT-9836232398362323";base="coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]";rt="core.rd-ep"
> cord_lc [fe80::101a:7dff:fe03:18f6] resource page=2
cord_lc [fe80::101a:7dff:fe03:18f6] resource page=2
Found resource/endpoint
Target: coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]/sense/temp
'anchor': 'coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]'
> cord_lc [fe80::101a:7dff:fe03:18f6] -r resource page=1
cord_lc [fe80::101a:7dff:fe03:18f6] -r resource page=1
Error during lookup -2
> cord_lc [fe80::101a:7dff:fe03:18f6] -r resource
cord_lc [fe80::101a:7dff:fe03:18f6] -r resource
Lookup result:
<coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]/node/info>;anchor="coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]",<coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]/sense/hum>;anchor="coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]",<coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]/sense/temp>;anchor="coap://[fe80::1c2a:e0ff:fec4:fe93%tapbr0]"

```

pokgak:
@haukepetersen Can you share with me how you setup your testing environment? I'll try to see if I can reproduce the segfault then.
Mine is using the RIOT vagrant VM as the base with aiocoap version 0.4b3 installed for the RD.

haukepetersen:
Of course. I just re-tested, the segfault still happens as before. I did exactly the following:

- fresh checkout of this branch
- rebased this branch on master
- start aiocoap-rd (version 1a7f8bbbe079714ed6ba4a979a70339688148bc6)

I just did run my test again, with the same result of RIOT segfaulting. But then I updated my `aiocoap` (before: 1a7f8bb, now: bd9faf7) and voila, no segfault anymore. Without knowing any particular, it seems to me that something in the behavior of `aiocoap` changed I guess. Although everything with this PR seems to be fine now, it leaves a little bitter mark, as there is apparently a combination of input data, that makes the implementation crash...

Anyway, as the tests run fine now, I'd say we should go ahead with this PR as is.

miri64:
@pokgak please squash

pokgak:
Squashed.

> Although everything with this PR seems to be fine now, it leaves a little bitter mark, as there is apparently a combination of input data, that makes the implementation crash...

Agree. That might be something worth looking into in the future. For now we'll leave it as a note here.

miri64:
`im880b` needs to be added to the insufficient memory list. You can squash that directly.

pokgak:
Thanks everyone!

PR #10228, opened 2018-10-23
Fletcher16: extend with multi-part functions

bergzand:
### Contribution description

This commit extends the fletcher16 function with a set of init, update
and finish functions to allow for multi-part checksums. This is useful
when the data that has to be checksummed is not available in a single
continuous buffer.

### Testing procedure

There is a unittest for verification of the algorithm

### Issues/PRs references

None

### DIFF
--- a/sys/checksum/fletcher16.c
+++ b/sys/checksum/fletcher16.c
@@ -20,21 +20,41 @@
 
 #include "checksum/fletcher16.h"
 
+static inline void _reduce(fletcher16_ctx_t *ctx)
+{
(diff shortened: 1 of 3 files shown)


bergzand:
Of course!

bergzand:
Added a test.

PeterKietzmann:
I don't know why travis indicates that the build is in progress. The detailed view shows success. ACK and go

bergzand:
@PeterKietzmann Thanks!

PR #10237, opened 2018-10-24
travis.yml: Use lesscpy from pip.

jcarrano:
### Contribution description

The version of lesscpy in Ubuntu 14 (in travis) is super outdated and causes `make docs` to fail.

lesscpy is used to generate riot.css from riot.less (the rule is only enabled if the tool is found on the system)

### Testing procedure

Use #9819 (it has a test script rebased on top of this PR)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 dist: trusty
 
 before_install:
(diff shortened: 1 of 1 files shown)


cladmi:
`doxygen` version in `travis` looks too old to like our documentation.

PR #10243, opened 2018-10-24
boards/common: add common place for nrf51 based boards

aabadie:
### Contribution description

 @haukepetersen's review in #9373 made me realized that there were a lot of code duplication between nrf51 based boards.

This PR is introducing a common location for what can be factorized and nrf51 based boards are adapted to take advantage of it:
- airfy-beacon
- microbit
- nrf51dongle
- nrf6310
- yunjia-nrf51822

### Testing procedure

If Murdock build pass, it shoud be ok. But testing some application on the modified platform would also make sense.

### Issues/PRs references

Found thanks to @haukepetersen's review in #9373


haukepetersen:
very nice! This kind of cleanup puts a smile on my face :-)

Can't look in detail right now, but will review this tomorrow. But below are some first quick findings

boards/common/nrf51/Makefile.features
```diff
@@ -0,0 +1,15 @@
+# Put defined MCU peripherals here (in alphabetical order)
+FEATURES_PROVIDED += periph_adc
+FEATURES_PROVIDED += periph_gpio periph_gpio_irq
+FEATURES_PROVIDED += periph_i2c
+FEATURES_PROVIDED += periph_rtt
+FEATURES_PROVIDED += periph_spi
+FEATURES_PROVIDED += periph_timer
+FEATURES_PROVIDED += periph_uart
```
haukepetersen: pulling all of these into the common folder seems to invalid to me, as there might be nrf51-based boards that do not provide all of these.

My suggestion: 
- all features that do not depend on board-specific (e.g. pin) configurations (gpio, gpio_irq) should be put into `cpu/nrf51/Makefile.features`. 
- features that are configured in the shared `boards/common/nrf51/include/periph_conf_common.h` should stay in this file
- all others should go into the corresponding board folders

boards/common/nrf51/Makefile.features
```diff
@@ -0,0 +1,15 @@
...
+
+# Various other features (if any)
+
+# The board MPU family (used for grouping by the CI system)
+FEATURES_MCU_GROUP = cortex_m0_2
```
haukepetersen: should be moved into the CPUs makefile.features, right?!

boards/common/nrf51/Makefile.include
```diff
@@ -0,0 +1,22 @@
...
+
+# define the default port depending on the host OS
+PORT_LINUX ?= /dev/ttyUSB0
+PORT_DARWIN ?= $(firstword $(sort $(wildcard /dev/tty.SLAB_USBtoUART*)))
+
+# setup serial terminal
+include $(RIOTMAKE)/tools/serial.inc.mk
+
+# most of nrf51 boards use an ST-Link v2 debug adapter
+export DEBUG_ADAPTER ?= stlink
```
haukepetersen: I would prefer to define the default programmer explicitly for each board.

boards/common/nrf51/Makefile.include
```diff
@@ -0,0 +1,22 @@
...
+export STLINK_VERSION ?= 2
+
+# include this module into the build
+INCLUDES += -I$(RIOTBOARD)/common/nrf51/include
+USEMODULE += boards_common_nrf51
+
+# use common openocd configuration for nrf51
```
haukepetersen: only needed in case openocd is actually used as a programmer. So does it make sense to ifdef this based on the used programmer?

boards/common/nrf51/board.c
```diff
@@ -7,11 +7,11 @@
  */
 
 /**
- * @ingroup     boards_yunjia-nrf51822
+ * @ingroup     boards_yunjia-boards_common_nrf51
```
haukepetersen: shouldn't this be `boards_common_nrf51`?

boards/nrf51dongle/Makefile.include
```diff
@@ -2,6 +2,9 @@
 export CPU = nrf51
```
haukepetersen: can be removed, right?!

boards/common/nrf51/Makefile.features
aabadie: I think this can be completely removed as I don't think it's still used by Murdock

aabadie:
I wrote this PR in a hurry earlier and didn't self review it before opening it. There's still a lot to improve. I'll try to fix them asap.

aabadie:
@haukepetersen, I addressed your first batch of comments: 
- I tried to provided a common way to manage multiple programmers supported by nrf51 boards
- I couldn't resist in adding openocd in the list of programmers for the microbit (but kept fscopy as the default one)
- nrf6310 now also uses common jlink definitions and scripts. Not sure if this is valid though
- periph_gpio and periph_gpio_irq provided features have been moved to `cpu/nrf51/Makefile.features`

The CI is now passing, which is already an improvement.

haukepetersen:
nice, will review now. Just one question before hand: does this mean that you integrated #9373 with this PR? Would be completely fine with me, just so I know what to look for...

boards/common/nrf51/include/periph_conf_common.h
```diff
@@ -0,0 +1,56 @@
+/*
```
haukepetersen: As a further improvement I think it would make sense to split this file to keep timer and RTT config in separate files. This way, boards using the shared nrf51 code can override either one independently. This would be similar to the clock config header files that we use for the shared stm32 boards. What do you think?

boards/microbit/include/board.h
```diff
@@ -21,20 +21,13 @@
 #define BOARD_H
 
 #include "cpu.h"
+#include "board_common.h"
 #include "periph_conf.h"
```
haukepetersen: we can drop `cpu.h` (included from `board_common.h`) and `periph_conf.h` (not used) from the includes. Probably this is true for all the other board.h files, too?!

boards/nrf6310/Makefile.include
```diff
@@ -19,3 +18,9 @@ export RESET_FLAGS = $(BINDIR)
 
 # setup serial terminal
 include $(RIOTMAKE)/tools/serial.inc.mk
+
+# use jlink to program this board
+PROGRAMMER = jlink
```
haukepetersen: does it make sense to use `?=` here? I think at least in theory it should be also possible to use openocd to program this device, right?!

haukepetersen:
One more change request: could you please also use the shared board code for the calliope-mini? Thanks.

haukepetersen:
Confirmed all changes working on the `nrf51dongle` and the `microbit`

boards/nrf6310/Makefile.include
aabadie: > I think at least in theory it should be also possible to use openocd to program this device, right?!

Indeed, I'll make the change. I also think in adding the possibility to use a variable containing the list of supported programmers, like `SUPPORTED_PROGRAMMERS`. The common makefile.include would check if the supplied programmer is there and return an error if it's not the case. What do you think ?

aabadie:
@haukepetersen, comments addressed.

> This would be similar to the clock config header files that we use for the shared stm32 boards. What do you think?

Maybe this can be done in a follow-up PR ?

haukepetersen:
> Maybe this can be done in a follow-up PR ?

certainly.

haukepetersen:
one note: did not test every board, only the ones mentioned above...

aabadie:
> did not test every board, only the ones mentioned above

I'm pretty confident that everything will be fine with airfy-beacon, calliope-mini and yunjia but less with nrf6310, especially the flashing change (make use of factorized jlink scripts).
Do you know someone that can test on this platform ?

aabadie:
I'll open a PR with the periph_conf optimizations.

haukepetersen:
> Do you know someone that can test on this platform ?

Unfortunately not... @PeterKietzmann do you have this board in HH?

aabadie:
@haukepetersen, see #10277 for factorizing the clock configurations.

aabadie:
From what I could find, the nrf6310 (nrfGO starter kit) is flashed using JLink so I don't see any reason for changes provided by this PR to not work. Of course, it would be better if we could test...
@haukepetersen, do you think we could merge this PR anyway  (CIs are green and you gave an ACK) ?

aabadie:
ping @haukepetersen, how do we proceed here ?

aabadie:
ping @haukepetersen

haukepetersen:
just went through this PR once more to make sure - my ACK holds.

So lets go!

PR #10245, opened 2018-10-24
boards/nucleo-l476rg: add I2C configuration

MichelRottleuthner:
### Contribution description
The title says it all.
I'm just adding this so we have another platform to test the changes that will come with #9804 .

### Testing procedure
Did a quick test with both provided configs and an SRF08 sensor.

### Issues/PRs references
Clocks are probably not correct as of now. #9804 will fix this.

### DIFF
--- a/boards/nucleo-l476rg/include/periph_conf.h
+++ b/boards/nucleo-l476rg/include/periph_conf.h
@@ -244,6 +244,29 @@ static const spi_conf_t spi_config[] = {
 /** @} */
 
 /**
+ * @name I2C configuration
+ * @{
+ */
+static const i2c_conf_t i2c_config[] = {
+    {
+        .dev            = I2C1,
+        .speed          = I2C_SPEED_NORMAL,
+        .scl_pin        = GPIO_PIN(PORT_B, 8),
+        .sda_pin        = GPIO_PIN(PORT_B, 9),
+        .scl_af         = GPIO_AF4,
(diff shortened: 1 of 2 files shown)


boards/nucleo-l476rg/include/periph_conf.h
```diff
@@ -243,6 +243,41 @@ static const spi_conf_t spi_config[] = {
...
+        .scl_af         = GPIO_AF4,
+        .sda_af         = GPIO_AF4,
+        .bus            = APB1,
+        .rcc_mask       = RCC_APB1ENR1_I2C1EN,
+        .irqn           = I2C1_ER_IRQn,
+    },
+    {
+        .dev            = I2C2,
+        .speed          = I2C_SPEED_NORMAL,
+        .scl_pin        = GPIO_PIN(PORT_B, 10),
```
aabadie: The second I2C can potentially conflicts with some PWM pins, see above in the file. I'm not sure if this is allowed in the default configuration of a board (I remember reviewers asking for this to be changed).

MichelRottleuthner:
@aabadie thanks for having a look. I don't think that this is not a real problem because the pwm is not initialized by default. If the application initializes the pwm it should be clear that nothing else (i2c) is connected to that pin.
I removed it for now anyway so we can move on.
For the future a solution to split configuration and (auto)initialization would be nice.

PR #10252, opened 2018-10-25
tools/ci: correcly report flake8 version. 

jcarrano:
~_This PR is rebased on top if #10251 for easy testing, but it is not required._~

### Contribution description

This PR contains two commits. I can split it if necessary.

**Change the way version detection works**

Detect command line tool versions without using "command". "command" may be a builting in some shells, leading to unportability.

The new version uses the status code to correctly detect a non-existent command. This allows it to differentiate between error in the tool and not-found errors.

It also works with compound commands, for example `python -m callable_module".

**Fix flake8 detection in some systems (for example, riotdocker)**

If the flake8 executable is not found, the static test script reports the tool as missing. It may happen that the flake8 module is installed, but the console entry point is not.

In the flake8 shell script, flake is invoked via `python -m`. The result is a confusing error message where static-test reports the tools as missing, yet the flake8 tests are run.

This patch makes the toolchain version script use the same command as the flake8 script.

### Testing procedure

The easiest way to see the issue is to look at Murdock output from any PR ([example](https://ci.riot-os.org/RIOT-OS/RIOT/10243/cbc72d44a20ca2dbae5d7c10f2bfea070f58425f/output.html)) and see that flake8 is reported as missing, but the test that uses flake8 passes.

~Travis uses a different image that does not show the problem, that's why I rebased on top of #10251. That PR's Travis output has the issue, while this one should not.~

With this PR, the static-tests can detect flake8 without problems.

### Why is this important?

Because if we have tests it is because we want to run them and if the output from the static test is contradictory, how do we know if the tests are running or not?

### DIFF
--- a/dist/tools/ci/print_toolchain_versions.sh
+++ b/dist/tools/ci/print_toolchain_versions.sh
@@ -5,21 +5,17 @@ get_cmd_version() {
         return
     fi
 
-    local cmd="$1"
-    if command -v "$cmd" 2>&1 >/dev/null; then
-        ver=$("$cmd" --version 2> /dev/null | head -n 1)
-        # some tools (eg. openocd) print version info to stderr
-        if [ -z "$ver" ]; then
-            ver=$("$cmd" --version 2>&1 | head -n 1)
-        fi
-        if [ -z "$ver" ]; then
-            ver="error"
-        fi
-    else
-        ver="missing"
+    VERSION_RAW=$( ($@ --version) 2>&1)
+    ERR=$?
+    VERSION=$(echo "$VERSION_RAW" | head -n 1)
+
+    if [ $ERR -eq 127 ] ; then # 127 means command not found
+        VERSION="missing"
+    elif [ $ERR -ne 0 ] ; then
+        VERSION="error: ${VERSION}"
     fi
 
-    printf "%s" "$ver"
+    printf "%s" "$VERSION"
 }
 
 get_define() {
@@ -123,7 +119,7 @@ for c in \
          python2 \
          python3 \
          ; do
-    printf "%23s: %s\n" "$c" "$(get_cmd_version $c)"
+    printf "%23s: %s\n" "$c" "$(get_cmd_version "${c}")"
 done
 printf "%23s: %s\n" "coccinelle" "$(get_cmd_version spatch)"
 
--- a/dist/tools/ci/print_toolchain_versions.sh
+++ b/dist/tools/ci/print_toolchain_versions.sh
@@ -111,7 +111,6 @@ for c in \
          cmake \
          cppcheck \
          doxygen \
-         flake8 \
          git \
          make \
          openocd \
@@ -121,6 +120,7 @@ for c in \
          ; do
     printf "%23s: %s\n" "$c" "$(get_cmd_version "${c}")"
 done
+printf "%23s: %s\n" "flake8" "$(get_cmd_version "python3 -Wignore -m flake8")"
 printf "%23s: %s\n" "coccinelle" "$(get_cmd_version spatch)"
 
 exit 0


dist/tools/ci/print_toolchain_versions.sh
```diff
@@ -6,11 +6,11 @@ get_cmd_version() {
     fi
 
     local cmd="$1"
-    if command -v "$cmd" 2>&1 >/dev/null; then
-        ver=$("$cmd" --version 2> /dev/null | head -n 1)
+    if command -v $cmd 2>&1 >/dev/null; then
```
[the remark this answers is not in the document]
gebart: How about adding optional extra arguments to this function instead, which can be used to pass arguments?
```bash
get_cmd_version() {
    if [ -z "$1" ]; then
        return
    fi

    local cmd="$1"
    shift 1
    local args=--version
    if [ $# -gt 0 ]; then
        args="$@"
    fi
    if command -v $cmd 2>&1 >/dev/null; then
        ver=$($cmd $args 2> /dev/null | head -n 1)
        # some tools (eg. openocd) print version info to stderr
        if [ -z "$ver" ]; then
            ver=$($cmd $args 2>&1 | head -n 1)
        fi
        if [ -z "$ver" ]; then
            ver="error"
        fi
    else
        ver="missing"
    fi

    printf "%s" "$ver"
}
```
and
```sh
printf "%23s: %s\n" "flake8" "$(get_cmd_version python3 -m flake8 --version)"
```

Though, on failure we get this error message instead (because of the captured stderr output):
```
flake8: /usr/lib/python-exec/python3.6/python3: No module named fake8
```

dist/tools/ci/print_toolchain_versions.sh
```diff
@@ -5,21 +5,17 @@ get_cmd_version() {
...
-        # some tools (eg. openocd) print version info to stderr
-        if [ -z "$ver" ]; then
-            ver=$("$cmd" --version 2>&1 | head -n 1)
-        fi
-        if [ -z "$ver" ]; then
-            ver="error"
-        fi
-    else
-        ver="missing"
+    VERSION_RAW=$( ($@ --version) 2>&1)
```
smlng: changed behaviour, was `2> /dev/null` which is actually needed because with `2>&1` you get the following on macOS:

```
python3 -m flake8 --version 2>&1
/usr/local/lib/python3.7/site-packages/pycodestyle.py:113: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
3.5.0 (mccabe: 0.6.1, pycodestyle: 2.3.1, pyflakes: 1.6.0) CPython 3.7.0 on Darwin
```

while as before:

```
python3 -m flake8 --version 2>/dev/null
3.5.0 (mccabe: 0.6.1, pycodestyle: 2.3.1, pyflakes: 1.6.0) CPython 3.7.0 on Darwin
```

jcarrano: That is because your version of `pycodestyle` is old. I can reproduce the issue myself if I force the installation of those versions via PIP.

In any case, if I change the behaviour I'd fix it, but I'd rather handle that case by case, see my comment below.

dist/tools/ci/print_toolchain_versions.sh
```diff
@@ -98,9 +94,10 @@ printf "%23s: %s\n" "avr-libc" "$(avr_libc_version avr-gcc)"
 printf "\n"
 printf "%s\n" "Installed development tools"
 printf "%s\n" "---------------------------"
-for c in cmake cppcheck doxygen flake8 git make openocd python python2 python3; do
-    printf "%23s: %s\n" "$c" "$(get_cmd_version $c)"
+for c in cmake cppcheck doxygen git make openocd python python2 python3; do
+    printf "%23s: %s\n" "$c" "$(get_cmd_version "${c}")"
 done
+printf "%23s: %s\n" "flake8" "$(get_cmd_version "python3 -m flake8")"
```
jcarrano: @smlng This should make the warning go away:

```suggestion
printf "%23s: %s\n" "flake8" "$(get_cmd_version "python3 -Wignore -m flake8")"
```

jcarrano:
I rebased-squashed-fixed conflicts.

@gebart Are you OK with this PR?

dist/tools/ci/print_toolchain_versions.sh
```diff
@@ -5,21 +5,17 @@ get_cmd_version() {
...
-            ver=$("$cmd" --version 2>&1 | head -n 1)
-        fi
-        if [ -z "$ver" ]; then
-            ver="error"
-        fi
-    else
-        ver="missing"
+    VERSION_RAW=$( ($@ --version) 2>&1)
+    ERR=$?
+    VERSION=$(echo "$VERSION_RAW" | head -n 1)
```
gebart: it is more reliable to use `printf %s "$VERSION_RAW"` instead of echo to avoid getting strange outputs in certain situations, especially since `$VERSION_RAW` may contain anything.

dist/tools/ci/print_toolchain_versions.sh
```diff
@@ -115,16 +111,16 @@ for c in \
...
          make \
          openocd \
          python \
          python2 \
          python3 \
          ; do
-    printf "%23s: %s\n" "$c" "$(get_cmd_version $c)"
+    printf "%23s: %s\n" "$c" "$(get_cmd_version "${c}")"
 done
+printf "%23s: %s\n" "flake8" "$(get_cmd_version "python3 -Wignore -m flake8")"
```
gebart:
```suggestion
printf "%23s: %s\n" "flake8" "$(get_cmd_version python3 -Wignore -m flake8)"
```

smlng:
@jcarrano can you have another look at @gebart comments, maybe we can get this in quick?!

dist/tools/ci/print_toolchain_versions.sh
jcarrano: This has the `head` to make sure we get the fist line in commands that have more than one.

dist/tools/ci/print_toolchain_versions.sh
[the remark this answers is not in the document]
jcarrano: "$@" will treat "a b" as the name of a command instead of command a with argument b.

dist/tools/ci/print_toolchain_versions.sh
jcarrano: See comment above for $@.

jcarrano:
@smlng I responded to @gebart's comment. He does not seem active at the moment, so if you think I answered satisfactorily dismiss his review and we merge.

smlng:
@cladmi can you have a quick look on this one, thx!

cladmi:
Oops, sorry I missed your message, it got lost in the stale bot flood.

PR #10065, opened 2018-09-27
bootloader: add riotboot minimal bootloader

kYc0o:
##### Overview
 
`riotboot` is the name of a minimal bootloader application and infrastructure.
 It consists of
  - the application "riotboot" in the bootloaders folder which serves as
    minimal bootloader
  - the module "riot_hdr" used to recognise RIOT firmware images
    which riotboot can boot.
  - the module "slot_util" used to manage the images (slots) with a
    RIOT header attached to them
  - a tool in dist/tools/riot_hdr for header generation.
  - several make targets to glue everything together

##### Concept
`riotboot` expects the flash to be separated in slots: slot 0 contains
 the bootloader and slot 1 a firmware image.

A RIOT image with a single slot looks like:
```
|------------------------------- FLASH -------------------------------------|
|- RIOTBOOT_SLOT0_SIZE -|-------- RIOTBOOT_FW_SLOT_SIZE (slot 1) -----------|
       (slot 0)         |---- RIOT_HDR_SIZE ------|
 ---------------------------------------------------------------------------
|       bootloader      | riot_hdr_t + filler (0) |     RIOT firmware       |
 ---------------------------------------------------------------------------
```
Please note that `RIOT_HDR_SIZE` depends on the architecture of the
MCU, since it needs to be aligned to either 256B or 512B. This is
fixed regardless of `sizeof(riot_hdr_t)`

The bootloader will, on reset, verify the checksum of the first slot, then
boot it. If the slot doesn't have a valid checksum, no image will be
booted and the bootloader will enter "while(1);".

The header contains

  - "RIOT" as magic number
  - an application version
  - a checksum

The bootloader "riotboot" only cares about checksum.

### Testing procedure

There's a test provided in tests/riotboot. This has only tested for now in:
  - [x] samr21-xpro
  - [x] iotlab-m3 (desk)

Please add above the boards that you have tested and worked.

### Issues/PRs references

#9342 
#9969

PR #10253, opened 2018-10-25
sys: add riot_hdr module

kYc0o:
### Contribution description

This PR adds a module called riot_hdr which is mainly used to identify "partitions" on flash. Sections on flash can be identified by parsing a header.

The header contains
```c
/**
 * @brief Structure to store image header - All members are little endian
 * @{
 */
typedef struct {
    uint32_t magic_number;     /**< header magic_number (always "RIOT")              */
    uint32_t version;          /**< Integer representing the partition version       */
    uint32_t start_addr;       /**< Address after the allocated space for the header */
    uint32_t chksum;           /**< checksum of riot_hdr                             */
} riot_hdr_t;
/** @} */
```

This is a split of #10065, since that PR is too complex to be reviewed as is.

### Testing procedure

`BOARD=samr21-xpro make -C tests/riot_hdr/ flash test`

### Issues/PRs references

#10065

### DIFF
--- a/sys/Makefile
+++ b/sys/Makefile
@@ -140,7 +140,6 @@ ifneq (,$(filter cord_ep,$(USEMODULE)))
     DIRS += net/application_layer/cord/ep
 endif
 
-
 DIRS += $(dir $(wildcard $(addsuffix /Makefile, $(USEMODULE))))
 
 include $(RIOTBASE)/Makefile.base
--- a/sys/riotboot/hdr.c
+++ b/sys/riotboot/hdr.c
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2017 Kaspar Schleiser <kaspar@schleiser.de>
+ *                    Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     sys_riotboot_hdr
+ * @{
+ *
+ * @file
+ * @brief       RIOT header helpers and tools
+ *
+ * @author      Kaspar Schleiser <kaspar@schleiser.de>
+ * @author      Francisco Acosta <francisco.acosta@inria.fr>
+ *
+ * @}
+ */
+
+#include <string.h>
+#include <stddef.h>
+#include <stdio.h>
+
+#ifdef RIOT_VERSION
+#include "log.h"
+#else
+#define LOG_INFO(...) printf(__VA_ARGS__)
+#endif
+
+#include "riotboot/hdr.h"
+#include "checksum/fletcher32.h"
+#include "byteorder.h"
+
+#if __BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__
+#   error "This code is implementented in a way that it will only work for little-endian systems!"
+#endif
+
+void riotboot_hdr_print(const riotboot_hdr_t *riotboot_hdr)
+{
+    printf("Image magic_number: 0x%08x\n", (unsigned)riotboot_hdr->magic_number);
+    printf("Image Version: 0x%08x\n", (unsigned)riotboot_hdr->version);
+    printf("Image start address: 0x%08x\n", (unsigned)riotboot_hdr->start_addr);
+    printf("Header chksum: 0x%08x\n", (unsigned)riotboot_hdr->chksum);
+    printf("\n");
+}
+
+int riotboot_hdr_validate(const riotboot_hdr_t *riotboot_hdr)
+{
+    if (riotboot_hdr->magic_number != RIOTBOOT_MAGIC) {
+        LOG_INFO("%s: riotboot_hdr magic number invalid\n", __func__);
+        return -1;
+    }
+
+    int res = riotboot_hdr_checksum(riotboot_hdr) == riotboot_hdr->chksum ? 0 : -1;
+    if (res) {
+        LOG_INFO("%s: riotboot_hdr checksum invalid\n", __func__);
+    }
+
+    return res;
+}
+
+uint32_t riotboot_hdr_checksum(const riotboot_hdr_t *riotboot_hdr)
+{
+    return fletcher32((uint16_t *)riotboot_hdr, offsetof(riotboot_hdr_t, chksum) / sizeof(uint16_t));
+}
--- a/tests/riotboot_hdr/Makefile
+++ b/tests/riotboot_hdr/Makefile
@@ -1,0 +1,10 @@
+include ../Makefile.tests_common
+
+USEMODULE += riotboot_hdr
+USEMODULE += embunit
+
+HDR_LOG_LEVEL ?= LOG_NONE
+
+CFLAGS += -DLOG_LEVEL=$(HDR_LOG_LEVEL)
+
+include $(RIOTBASE)/Makefile.include
--- a/tests/riotboot_hdr/tests/01-run.py
+++ b/tests/riotboot_hdr/tests/01-run.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+
+# Copyright (C) 2018 Francisco Acosta <francisco.acosta@inria.fr>
+#
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import sys
+from testrunner import run
+
+
+def testfunc(child):
+    child.expect(r"OK \(\d+ tests\)")
+
+
+if __name__ == "__main__":
+    sys.exit(run(testfunc))
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -791,6 +791,11 @@ ifneq (,$(filter uuid,$(USEMODULE)))
   USEMODULE += random
 endif
 
+ifneq (,$(filter riotboot_hdr, $(USEMODULE)))
+  USEMODULE += checksum
+  USEMODULE += riotboot
+endif
+
 # Enable periph_gpio when periph_gpio_irq is enabled
 ifneq (,$(filter periph_gpio_irq,$(USEMODULE)))
   FEATURES_REQUIRED += periph_gpio
--- a/makefiles/pseudomodules.inc.mk
+++ b/makefiles/pseudomodules.inc.mk
@@ -54,6 +54,7 @@ PSEUDOMODULES += pktqueue
 PSEUDOMODULES += printf_float
 PSEUDOMODULES += prng
 PSEUDOMODULES += prng_%
+PSEUDOMODULES += riotboot_%
 PSEUDOMODULES += saul_adc
 PSEUDOMODULES += saul_default
 PSEUDOMODULES += saul_gpio
--- a/sys/include/riotboot/hdr.h
+++ b/sys/include/riotboot/hdr.h
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2018 Kaspar Schleiser <kaspar@schleiser.de>
+ *                    Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    sys_riotboot_hdr RIOT header helpers and tools
+ * @ingroup     sys
+ * @{
+ *
+ * The header contains
+ *
+ * - "RIOT" as magic number
+ * - the application version
+ * - the address where the RIOT firmware is found
+ * - the checksum of the three previous fields
+ *
+ * @file
+ * @brief       RIOT "partition" header and tools
+ *
+ * @author      Kaspar Schleiser <kaspar@schleiser.de>
+ * @author      Francisco Acosta <francisco.acosta@inria.fr>
+ *
+ * @}
+ */
+
+#ifndef RIOTBOOT_HDR_H
+#define RIOTBOOT_HDR_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+
+/**
+ * @brief  Magic number for riotboot_hdr
+ *
+ */
+#define RIOTBOOT_MAGIC         0x544f4952 /* "RIOT" */
+
+/**
+ * @brief Structure to store image header - All members are little endian
+ * @{
+ */
+typedef struct {
+    uint32_t magic_number;      /**< Header magic number (always "RIOT")              */
+    uint32_t version;           /**< Integer representing the partition version       */
+    uint32_t start_addr;        /**< Address after the allocated space for the header */
+    uint32_t chksum;            /**< Checksum of riotboot_hdr                         */
+} riotboot_hdr_t;
+/** @} */
+
+/**
+ * @brief  Print formatted riotboot_hdr_t to STDIO
+ *
+ * @param[in] riotboot_hdr  ptr to image header
+ *
+ */
+void riotboot_hdr_print(const riotboot_hdr_t *riotboot_hdr);
+
+/**
+ * @brief  Validate image header
+ *
+ * @param[in] riotboot_hdr  ptr to image header
+ *
+ * @returns 0 if OK
+ * @returns -1 if not OK
+ */
+int riotboot_hdr_validate(const riotboot_hdr_t *riotboot_hdr);
+
+/**
+ * @brief  Calculate header checksum
+ *
+ * @param[in] riotboot_hdr  ptr to image header
+ *
+ * @returns the checksum of the given riotboot_hdr
+ */
+uint32_t riotboot_hdr_checksum(const riotboot_hdr_t *riotboot_hdr);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* RIOTBOOT_HDR_H */
--- a/sys/riotboot/Makefile
+++ b/sys/riotboot/Makefile
@@ -1,0 +1,3 @@
+SUBMODULES := 1
+
+include $(RIOTBASE)/Makefile.base
--- a/tests/riotboot_hdr/main.c
+++ b/tests/riotboot_hdr/main.c
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2018 Inria
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup    tests
+ * @{
+ *
+ * @file
+ * @brief      Tests for module riotboot_hdr
+ *
+ * @author     Francisco Acosta <francisco.acosta@inria.fr>
+ */
+
+#include <stdio.h>
+
+#include "riotboot/hdr.h"
+#include "embUnit.h"
+
+const riotboot_hdr_t riotboot_hdr_good = {
+    .magic_number = RIOTBOOT_MAGIC,
+    .version = 0x5bd19bff,
+    .start_addr = 0x00001100,
+    .chksum = 0x02eda672
+};
+
+const riotboot_hdr_t riotboot_hdr_bad_magic = {
+    .magic_number = 0x12345678,
+    .version = 0x5bd19bff,
+    .start_addr = 0x00001100,
+    .chksum = 0x02eda672
+};
+
+const riotboot_hdr_t riotboot_hdr_bad_chksum = {
+    .magic_number = RIOTBOOT_MAGIC,
+    .version = 0x5bd19bff,
+    .start_addr = 0x00001100,
+    .chksum = 0x02000000
+};
+
+static void test_riotboot_hdr_01(void)
+{
+    int ret = riotboot_hdr_validate(&riotboot_hdr_good);
+
+    TEST_ASSERT_EQUAL_INT(0, ret);
+}
+
+static void test_riotboot_hdr_02(void)
+{
+    int ret = riotboot_hdr_validate(&riotboot_hdr_bad_magic);
+
+    TEST_ASSERT_EQUAL_INT(-1, ret);
+}
+
+static void test_riotboot_hdr_03(void)
+{
+    int ret = riotboot_hdr_validate(&riotboot_hdr_bad_chksum);
+
+    TEST_ASSERT_EQUAL_INT(-1, ret);
+}
+
+static void test_riotboot_hdr_04(void)
+{
+    uint32_t chksum = riotboot_hdr_checksum(&riotboot_hdr_good);
+
+    TEST_ASSERT_EQUAL_INT(0x02eda672, chksum);
+}
+
+Test *tests_riotboot_hdr(void)
+{
+    EMB_UNIT_TESTFIXTURES(fixtures) {
+        new_TestFixture(test_riotboot_hdr_01),
+        new_TestFixture(test_riotboot_hdr_02),
+        new_TestFixture(test_riotboot_hdr_03),
+        new_TestFixture(test_riotboot_hdr_04),
+    };
+
+    EMB_UNIT_TESTCALLER(riotboot_hdr_tests, NULL, NULL, fixtures);
+
+    return (Test *)&riotboot_hdr_tests;
+}
+
+int main(void)
+{
+    TESTS_START();
+    TESTS_RUN(tests_riotboot_hdr());
+    TESTS_END();
+    return 0;
+}


bergzand:
Is this code also supposed to work on big endian systems? If not could you place a compile time warning/error in the file? Something similar to the one in `sys/crypto/chacha.c`.

kYc0o:
> Is this code also supposed to work on big endian systems? 

No, only little endian.

> If not could you place a compile time warning/error in the file? Something similar to the one in sys/crypto/chacha.c.

Thanks for the hint! Will update.

kaspar030:
Maybe move to ```sys/ota/riot_hdr```?

Why can't this be a part of riotboot?

bergzand:
> Maybe move to `sys/ota/riot_hdr`?

Or `sys/firmware`, It could be used for things outside of the OTA part.

kYc0o:
> Maybe move to sys/ota/riot_hdr?

I think this is not meant to be exclusively for OTA, it can be used in other contexts and I think it can live alone.

> Why can't this be a part of riotboot?

riotboot grew a lot and it cannot be reviewed as is, I'm splitting to ease the review task.

kYc0o:
> Or sys/firmware, It could be used for things outside of the OTA part.

I agree it can be used somewhere else, but I'm not sure `firmware` is meaningful. You might want to put the header to identify other stuff which is not a firmware (though I don't have use cases at hand).

kaspar030:
> riotboot grew a lot

Maybe, but riot_hdr_t is a struct with 4 members and a function call. IMO that doesn't deserve it's own module in sys/. This is also pretty much a part of riotboot. Sure it'll be exported, but still...

kYc0o:
Side note: I'm not sure if I need to activate the board tests on Murdock, I didn't succeed to catch the output on my OSX, maybe in Linux it's a bit slower and the expect can catch it.

kaspar030:
> I'm not sure `firmware` is meaningful.

`upgrade`?

bergzand:
> `upgrade`?

`slot`?

kYc0o:
> `slot`?

+1

bergzand:
> No, only little endian.

To really start nitpicking, we also have to be careful when generating this header on a big endian development station such as the PowerPC architecture :)

kYc0o:
> To really start nitpicking, we also have to be careful when generating this header on a big endian development station such as the PowerPC architecture :)

I understand your concern, at least this PR doesn't contain any header generation means (contrary to #10065 ). Should the provided functions in this PR take care of it?

kYc0o:
> If it is home-grown tool that generates the header it just needs to be made aware of the endianness of the target it is generating a header for. The same thing applies for any tools that wants to read raw headers.

This is more or less what's proposed in #10065. However, the home-grown tool there IMHO it's just a proposition which might be eventually used to generate the header, but I think any serious usage of riotboot should generate it's own header with its own tools.

bergzand:
> If it is home-grown tool that generates the header it just needs to be made aware of the endianness of the target it is generating a header for. The same thing applies for any tools that wants to read raw headers.

Exactly. People need to be aware that this is something they have to keep in mind when building the tooling.

kYc0o:
> Exactly. People need to be aware that this is something they have to keep in mind when building the tooling.

The big warning should be enough?

bergzand:
> The big warning should be enough?

I don't think there is much more we can do here

jcarrano:
> The big warning should be enough?

The big warning is misleading. It makes you think that this code is the one that cannot be used in big endian. __This code is alright wrt. endianness__, the real issue is the code that _generates_ the headers. This code _expects_ the header to be in the endianness of the target, and if that is the spec it is fine. External tools should adapt to the spec (and that is easy on tools that run with PCs).

kYc0o:
> The big warning is misleading. It makes you think that this code is the one that cannot be used in big endian. This code is alright wrt. endianness, the real issue is the code that generates the headers. This code expects the header to be in the endianness of the target, and if that is the spec it is fine. External tools should adapt to the spec (and that is easy on tools that run with PCs).

The code that generates the header in #10065 makes use of this module. If someone tries (weird...) to generate it in a big endian machine the big warning will be thrown.

bergzand:
> The big warning is misleading. It makes you think that this code is the one that cannot be used in big endian. **This code is alright wrt. endianness**,

No, I don't think it is going to work on a big endian MCU in the current form. The documentation clearly states little endianess variables in the struct. If one would follow that specification then the code would not work on a big endian system. The start address for example would have to be converted from little endian to big endian to resolve to the start address. That is not something the current code takes into account.

jcarrano:
> If one would follow that specification then the code would not work on a big endian system.

Is that spec definitive? Is it possible that the spec to mandate target endianness instead? I think it is smarter to make slightly more complex tools and make life easier for the poor MCU.

bergzand:
> Is that spec definitive? Is it possible that the spec to mandate target endianness instead? I think it is smarter to make slightly more complex tools and make life easier for the poor MCU.

That's what I proposed in #10065.

bergzand:
> That's what I proposed in #10065.

[In this thread] to be exact.

[In this thread]: https://github.com/RIOT-OS/RIOT/pull/10065#discussion_r223385745

kYc0o:
> That's what I proposed in #10065.

AFAIK there we aim to add a bootloader for ARM Cortex-M0+/3/4/7 which all of them are little endian. It will never exist any of these for big endian.

So what I suggest, is that whenever someone brings the support of this specific bootloader into a big endian machine, it needs also to perform the needed conversions. Is all RIOT code able to speak big endian if ever that comes to happen?

jcarrano:
> whenever someone brings the support of this specific bootloader into a big endian machine, it needs also to perform the needed conversions.

Why put the burden of byte-order conversion on the poor microcontroller, if data can be delivered in the correct endianness very easily? If you say "the byte order is the same as the target's" then you need to change nothing in your code, whoever wants to run it in a big endian has to change nothing too. 

The only (apparent) consequence may be a slightly more complex header generation tool. I say apparent because if you do it right you have to ensure the endianness is right regardless of the platform the tool runs on.

sys/include/riot_hdr.h
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * @brief Structure to store image header - All members are little endian
+ * @{
+ */
+typedef struct {
+    uint32_t magic_number;     /**< header magic_number (always "RIOT")              */
+    uint32_t version;          /**< Integer representing the partition version       */
+    uint32_t start_addr;       /**< Address after the allocated space for the header */
+    uint32_t chksum;           /**< checksum of riot_hdr                             */
```
jcarrano: Why it does not indicate the length of the image?

bergzand: @jcarrano The header struct here is for a minimal bootloader application. It is however possible to extend this struct for different types of bootloaders. So if the current bootloader does not require it, I'd rather keep it out of the struct.

sys/riot_hdr/riot_hdr.c
```diff
@@ -0,0 +1,72 @@
...
+#include "byteorder.h"
+
+#if __BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__
+#   error "This code is implementented in a way that it will only work for little-endian systems!"
+#endif
+
+/**
+ *  @brief Number of header bytes that get checksummed
+ */
+#define RIOT_HDR_CHECKSUM_LEN      (12)
```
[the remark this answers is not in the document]
kYc0o: Addressed. Can you mark your comment as change requests while you review? As comments I don't know if you want me to do it or if is just a suggestion.

sys/include/riot_hdr.h
kYc0o: I'd say in the worse case you always have the slot size which can be used for that purpose. Otherwise, while receiving the data allocated to that slot you always get the size, so if you need to perform other operations you can pass that information if needed.

Anyways, the bootloader proposed on #10065 is not ready to perform any swapping, and a big change would be needed for that. When those changes come, modifications to this struct might be done.

kYc0o:
Someone in the discussion (@bergzand @jcarrano) did a review or tested it? It sounds that the points were addressed, let me know if you have more comments/suggestions.

kaspar030:
> > `upgrade`?
> 
> `slot`?

Without context, what do you think when you read "slot"? I'm not sure anyone would think about the slots of a firmware upgrade.

kYc0o:
> Without context, what do you think when you read "slot"? I'm not sure anyone would think about the slots of a firmware upgrade.

True, maybe something more verbose would be necessary. As far as I can tell, the reason to keep this module out of sys/ is that is too small? I think it can apply to several contexts which are not all related to upgrade or OTA or the slots. Why don't we leave it there?

danpetry:
Hi @kYc0o , would we be able to collect the header generation tool into this PR as well? I was confused by why we specified little-endian platforms only - the tool needs to be looked at to understand this, and it's not in this PR. Could we also involve this tool in the tests, and separate out stuff to do with headers from riotboot.mk (and anywhere else)? 

I.e. everything header related should be included in this PR and encapsulated from other bootloader code, for extensibility later, even if it's the only place the header is used at the moment.

danpetry:
re `slot` vs `firmware` etc: how about `img_hdr`? Or even just `header` if we want to keep it completely generic? If at some point someone wants to use it to describe a data partition or something?

danpetry:
For me, sys/riot_hdr is fine. A header for a RIOT image, nothing else assumed.

danpetry:
I'd also like to see all header-related things from https://github.com/RIOT-OS/RIOT/pull/10065 included here. Including the home-grown tool and a test that encompasses it. It'd make it easier to review and encourage clean abstraction of header stuff from the rest of the bootloader.

tests/riot_hdr/main.c
```diff
@@ -0,0 +1,96 @@
...
+{
+    uint32_t chksum = riot_hdr_checksum(&riot_hdr_good);
+    TEST_ASSERT_EQUAL_INT(0x02eda672, chksum);
+}
+
+static void test_riot_hdr_05(void)
+{
+    uint32_t chksum = riot_hdr_checksum(&riot_hdr_good);
+    TEST_ASSERT_EQUAL_INT(0x02eda672, chksum);
+}
```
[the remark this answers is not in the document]
kYc0o: I think it was intended to check a bad checksum and was a non-modified copy paste.

Good catch! will change it asap.

kYc0o:
Comments addressed.

danpetry:
Thanks Paco. Any thoughts on whether we should include the tool and a related test in this PR?

sys/riot_hdr/riot_hdr.c
```diff
@@ -0,0 +1,68 @@
...
+#include "checksum/fletcher32.h"
+#include "byteorder.h"
+
+#if __BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__
+#   error "This code is implementented in a way that it will only work for little-endian systems!"
+#endif
+
+void riot_hdr_print(const riot_hdr_t *riot_hdr)
+{
+    printf("Image magic_number: 0x%08x\n", (unsigned)riot_hdr->magic_number);
```
miri64: [`printf()` requires `<stdio.h>`](https://ci.riot-os.org/RIOT-OS/RIOT/10253/e03b1855225192251e1a6ff16dd086f05829f660/output/compile/tests/riot_hdr/esp32-olimex-evb:gnu.txt) ;-)

sys/include/riot_hdr.h
```diff
@@ -0,0 +1,88 @@
...
+#define RIOT_HDR_MAGIC         0x544f4952 /* "RIOT" */
+
+/**
+ * @brief Structure to store image header - All members are little endian
+ * @{
+ */
+typedef struct {
+    uint32_t magic_number;     /**< header magic_number (always "RIOT")              */
+    uint32_t version;          /**< Integer representing the partition version       */
+    uint32_t start_addr;       /**< Address after the allocated space for the header */
```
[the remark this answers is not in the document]
kYc0o: The reason is that manipulating uint32_t variables is far easier than pointers (and it's arithmetic, since we do several operations with it), it avoids lots of casting and thus points of failures when something is not cast/managed correctly. The vendor headers also represent this kind of addresses using integers, which simplifies their usage.

tests/riot_hdr/main.c
```diff
@@ -0,0 +1,89 @@
...
+    int ret = riot_hdr_validate(&riot_hdr_bad_chksum);
+    TEST_ASSERT_EQUAL_INT(-1, ret);
+}
+
+static void test_riot_hdr_04(void)
+{
+    uint32_t chksum = riot_hdr_checksum(&riot_hdr_good);
+    TEST_ASSERT_EQUAL_INT(0x02eda672, chksum);
+}
+
```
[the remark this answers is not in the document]
kYc0o: The problem with the tool is that is hard to manipulate the binaries generated by it, then transform them to code to finally be integrated in the tests. The tools makes usage of header code so some tests are already integrated here. Do you have a specific suggestion of which kind of test would you like to have for the tool? That might also be a separated PR.

sys/include/riot_hdr.h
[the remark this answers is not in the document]
kYc0o: Well, at the end we don't know which start address is (image, parameters, other info...).

sys/include/riot_hdr.h
```diff
@@ -0,0 +1,88 @@
...
+extern "C" {
+#endif
+
+#include <stdint.h>
+
+/**
+ * @brief  Magic number for riot_hdr
+ *
+ */
+#define RIOT_HDR_MAGIC         0x544f4952 /* "RIOT" */
```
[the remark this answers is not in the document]
kYc0o: I strongly think it belongs here. What's the problem/advantages by setting it here or from the tool?

sys/include/riot_hdr.h
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+/**
+ * @defgroup    sys_riot_hdr RIOT header helpers and tools
+ * @ingroup     sys
+ * @{
+ *
+ * The header contains
+ *
+ * - "RIOT" as magic number
```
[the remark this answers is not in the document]
kYc0o: The principle of having a magic number is that it's recognisable by this or other tools. Making it variable removes the static nature of the header, and thus makes easier losing track of headers/magic numbers.

I propose to leave it as is, and if there are use cases which need it variable, then modifying it accordingly.

kaspar030:
> and I'm personally happy with the location and name of the module.

I'll bikeshed some more. ;)

This is part of riotboot, right? The PR description says so ("was split out because PR too complex"). There's *no* other user. It's name is "riot_hdr", but there's plenty of other riot, and there's plenty of other hdr in the tree, so calling anything "riot_hdr" cannot be speaking for itself.
This is a simple header with <5 fields, and the header belongs to riotboot.

Creating "sys/riot_hdr" is like creating "sys/net_hdr" but only referring to gnrc's ethernet header.

Anyhow, I'll push some commits that move "sys/riot_hdr" to "sys/riotboot" as hdr.c and "include/riotboot/hdr.h" so we see how it looks like. do you have convincing arguments on moving it back up one level, considering there's more of riotboot to come?

kaspar030:
> This is a split of #10065, since that PR is too complex to be reviewed as is.

Splitting a PR doesn't imply splitting the module in question, which was done here...

sys/include/riot_hdr.h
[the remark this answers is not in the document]
kaspar030: +1

@danpetry What would be the use case for having this configurable?

kYc0o:
> Anyhow, I'll push some commits that move "sys/riot_hdr" to "sys/riotboot" as hdr.c and "include/riotboot/hdr.h" so we see how it looks like. do you have convincing arguments on moving it back up one level, considering there's more of riotboot to come?

I think this might be a good idea, I didn't want to tie the header for unique riotboot usage, that's why I thought a module by itself would have been useful. Also, this adds a bit of decoupling for a simple review of a module that can live by itself.

danpetry:
I also think that this shouldn't be tied to riotboot.  
I guess the underlying question is about the granularity of RIOT modules. When should they be separate etc. I think they should be separate if there are decent conceivable uses for them. (As @kaspar030 stated above.)  
Here, the header _could_ be used independently of the bootloader, but are there any use cases? I can see that people might want to use it to assign metadata to any area of memory. Program code partitions (as it's being used in this case), and data partitions as well. Are there any headers in RIOT that are doing this at the moment?  
Agree that the name `riot_hdr` is too generic because it's simply global. But I think it should be an independent re-usable module in this case.

tests/riot_hdr/main.c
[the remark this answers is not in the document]
danpetry: This comment wasn't really meant to refer to integrating the generation tool into tests, more whether the `riot_hdr_print` function should be covered in the tests.

Re covering the tool in tests, I'm not sure what you mean in the comment - I'm not sure why they'd need to be transformed into code. I was thinking something along the lines of:

1. Compile tool
2. Generate header
3. Generate main program binary with some test function that uses header
4. Run tests above

Is this feasible?

sys/include/riot_hdr.h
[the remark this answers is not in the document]
danpetry: Ok thanks for explaining this. Agreed
Maybe we could consider changing it after we've decided the purpose of the module (i.e. whether it's tied to the bootloader or a re-usable module). If `riot_hdr` is too generic, maybe the magic number `RIOT` is also too generic and should be changed, in line with whatever we decide the purpose of the module is?

danpetry:
Also, please note the previous review comments - they're marked "outdated" because of pushing name changes of files etc. but still need to be addressed. @kYc0o hope your new job is going ok, if you're reduced in time I'm happy to help out with coding changes etc, if you wish.

emmanuelsearch:
concerning module or not: how about we keep things simple for now?
If other uses emerge, we can refine.
(Context: we urgently need to move on with a basic bootloader support)

danpetry:
Ok, I'm happy to make this aspect non-blocking, to move things on.

sys/riot_hdr/riot_hdr.c
```diff
@@ -0,0 +1,68 @@
...
+#include <stdio.h>
+#define LOG_INFO(...) printf(__VA_ARGS__)
+#endif
+
+#include "riot_hdr.h"
+#include "checksum/fletcher32.h"
+#include "byteorder.h"
+
+#if __BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__
+#   error "This code is implementented in a way that it will only work for little-endian systems!"
```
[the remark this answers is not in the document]
danpetry: After talking with @cladmi and @emmanuelsearch offline, I'm resolving this because basically there are well explained reasons for it being here:
- It's common to define a serialization format that the platform must support (e.g. also unicode/utf-8/etc for text files)
- most of our platforms are little-endian anyway

tests/riotboot_hdr/tests/01-run.py
```diff
@@ -0,0 +1,18 @@
...
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import sys
+from testrunner import run
+
+
+def testfunc(child):
+    child.expect(r"OK \(5 tests\)")
```
miri64:
```suggestion
    child.expect(r"OK \(\d+ tests\)")
```

Otherwise this line needs to be always rewritten when the number of test cases changes.

emmanuelsearch:
@danpetry thanks for the speed + contructiveness !
Is that a provisional ACK (assuming yoru nits are corrected)?

kYc0o:
Thanks @danpetry for the fixups!, I'm actually quite busy now but the changes are OK, they should be merged here now.

danpetry:
yes, ACK from me, please squash/rebase and address Murdock if it complains.

kYc0o:
> I also think that this shouldn't be tied to riotboot.

@danpetry you have acked also the changes made by @kaspar030 ?

danpetry:
> Ok, I'm happy to make this aspect non-blocking, to move things on.
was my slightly later comment.
In other words, yes I've acked also the changes made by @kaspar030 .

kYc0o:
Hmmm... I kind of find strange to merge something introducing the module riotboot (before riotboot actually exists) when the PR doesn't contain anything about it, besides the description about the splitting.

If everyone is ok with this I don't care to merge, but a little discussion around it doesn't hurt I think.

emmanuelsearch:
@kYc0o I'd say let's now focus on #10065 and get this ACKed too.
Then: merge the cluster  #10065, #10253 in that order?

danpetry:
Yes sure. Me and @emmanuelsearch discussed this offline this afternoon, sorry for not summarizing the discussion here, an oversight on my part. I pointed out the same point - i.e. that if something is split off, it should operate on its own, i.e. be its own module. @emmanuelsearch disagreed and said it's ok, and that in this case there's not an immediate, practical use case for a generic image header. We agreed at the end that, along with the wider issue about what should be its own module, and the degree of encapsulation that modules should have, this was something more global than this PR and consequently better addressed in an issue discussion.
  
In this situation, I'm happy with going forward with this, because
- the tests pass and give adequate coverage IMO
- it's not /core
- I'm as eager as @emmanuelsearch to get the bootloader OTA implementation moving along.

sys/Makefile
```diff
@@ -139,6 +139,9 @@ endif
 ifneq (,$(filter cord_ep,$(USEMODULE)))
     DIRS += net/application_layer/cord/ep
 endif
+ifneq (,$(filter riotboot_%,$(USEMODULE)))
```
kYc0o: This needs to point to `riotboot` only instead. I'll add another fixup for this. @kaspar030 do you agree?

sys/Makefile
```diff
@@ -139,6 +139,9 @@ endif
 ifneq (,$(filter cord_ep,$(USEMODULE)))
     DIRS += net/application_layer/cord/ep
 endif
+ifneq (,$(filter riotboot,$(USEMODULE)))
```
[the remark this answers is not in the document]
kYc0o: I'd like to leave it for the sake of completeness. Is it ok?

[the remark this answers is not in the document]
cladmi: It's the role of the line after to handle the directories with the same name as the module, so here it is duplicating something already handled after for no reason.

sys/riotboot/hdr.c
```diff
@@ -0,0 +1,68 @@
...
+    printf("Image Version: 0x%08x\n", (unsigned)riotboot_hdr->version);
+    printf("Image start address: 0x%08x\n", (unsigned)riotboot_hdr->start_addr);
+    printf("Header chksum: 0x%08x\n", (unsigned)riotboot_hdr->chksum);
+    printf("\n");
+}
+
+int riotboot_hdr_validate(const riotboot_hdr_t *riotboot_hdr)
+{
+    if (riotboot_hdr->magic_number != RIOTBOOT_MAGIC) {
+        LOG_INFO("%s: riotboot_hdr magic number invalid\n", __func__);
```
aabadie: But since it's an error case, it should be `LOG_ERROR`. The same goes line 59 in this file.

sys/riotboot/hdr.c
```diff
@@ -0,0 +1,68 @@
...
+ * @author      Francisco Acosta <francisco.acosta@inria.fr>
+ *
+ * @}
+ */
+
+#include <string.h>
+#include <stddef.h>
+#include <stdio.h>
+
+#ifdef RIOT_VERSION
```
aabadie: I don't understand why `log.h` is included with RIOT_VERSION is defined. I see no good reason for just having `#include "log.h"`. Maybe there is one ?

tests/riotboot_hdr/Makefile
```diff
@@ -0,0 +1,8 @@
+include ../Makefile.tests_common
+
+USEMODULE += riotboot_hdr
+USEMODULE += embunit
+
+CFLAGS += -DLOG_LEVEL=LOG_NONE
```
aabadie: This seems useless in a test application where one would expect a good logging level (DEBUG) with everything displayed.

sys/riotboot/hdr.c
kYc0o: The reason is that when this file is compiled without RIOT the `LOG_*` macros won't be available, unless log.h is included, but it only exists on the RIOT context. Thus, whenever it doesn't compile for RIOT, the `LOG_DEBUG` macro is defined as a simple `prinft()`.

sys/riotboot/hdr.c
kYc0o: In this and the other case it would imply to redefine all the `LOG_*` macros according to the debug level (in other words, re-implement the whole logging module after the `#ifdef`), which is not the goal in this use case. The goal here is to just inform if something went wrong even if there are failures out of the RIOT context (e.g. a external tool which might want to validate a header).

Another option would be to put `#ifdef`s before each all to match the context on which this file is compiled.

tests/riotboot_hdr/Makefile
miri64: Not if the test application also should showcase the size benefits ;-).

kYc0o: It's a leftover from the main use case (riotboot).

I'll remove it. Just note that without this, the booloader size would increase dramatically due to the need of a `printf()` function.

miri64: But maybe make it overridable, so people can modify it when it is required?

kYc0o: > Not if the test application also should showcase the size benefits ;-)

I think here anyways the size doesn't matter that much.

>But maybe make it overridable, so people can modify it when it is required?

However I agree with this, I'll make it this way.

sys/Makefile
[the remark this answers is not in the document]
kYc0o: Ok, I removed it.

kYc0o:
Finally green! Merge at will.

tests/riotboot_hdr/Makefile
kaspar030: I don't really see the use of this. Why would anyone ever not have the output in the test?
As @kYc0o noted, the code is partially shared between the bootloader, the RIOT library code and the header generation code (which runs on host).

The bootloader needs to be as minimal as possible and has otherwise *no* output in order to drop any dependency to the (giant) newlib stdio code. That's why it is compiled with LOG_LEVEL=LOG_NONE. This test application doesn't really benefit from it.

Anyhow, I'll open a removal PR once this is in.

miri64: If you don't want to make it configurable, then remove the log messages completely. Otherwise, they make no sense.

kaspar030: The log messages are as configurable as anywhere else, just compile the corresponding module with "-DLOG_LEVEL=loglevel".

The question here is whether the test case needs two extra makefile lines where developers can set the log level, as if that has any extra meaning for this particular test.

The log messages are interesting if the code is used as a RIOT library or for the header tools. It cannot be compiled in when used in the bootloader for size reasons. That's all pretty much irrelevant for the test.

sys/Makefile
```diff
@@ -140,7 +140,6 @@ ifneq (,$(filter cord_ep,$(USEMODULE)))
     DIRS += net/application_layer/cord/ep
 endif
```
kaspar030: this is unrelated but I think we can live with it. ;)

kYc0o: Sorry, small mistake...

miri64:
:tada:

PR #10264, opened 2018-10-26
core: Rename typedef schedstat -> schedstat_t

gebart:
### Contribution description

To follow naming conventions, the `schedstat` type should be named `schedstat_t`

### Testing procedure

Compile test should prove that there are no references left to the old name.


smlng:
@kaspar030 do you see any problems with that, it's a change in `core` (though only minimal) so I'd like to have a second ACK (and preferably yours).

kaspar030:
No problem! ACK & go.

PR #10267, opened 2018-10-26
sys/auto_init: Fixed initialization of sht1x

maribu:
### Contribution description

Commit cef81a102c61455dd418225a009ff78558b5b9c2 introduced two bugs:

 - the driver for sht1x is initialized two times
 - the second initialization is done only when SAUL is used, but sht1x needs
   to be initialized in any case. (SAUL registration is also done there, but
   only when SAUL is actually being used.)

This commit fixes both.

### Testing procedure

E.g. use `examples/default` on a board that features an SHT1x sensor, e.g. the MSB-A2. Prior to this PR RIOT will not boot due to the double initialization, with this PR applied RIOT will boot again.

### DIFF
--- a/sys/auto_init/auto_init.c
+++ b/sys/auto_init/auto_init.c
@@ -428,10 +428,6 @@ void auto_init(void)
     extern void auto_init_pulse_counter(void);
     auto_init_pulse_counter();
 #endif
-#ifdef MODULE_SHT1X
-    extern void auto_init_sht1x(void);
-    auto_init_sht1x();
-#endif
 #ifdef MODULE_SI114X
     extern void auto_init_si114x(void);
     auto_init_si114x();
--- a/sys/auto_init/auto_init.c
+++ b/sys/auto_init/auto_init.c
@@ -296,6 +296,10 @@ void auto_init(void)
 
 /* initialize sensors and actuators */
 #ifdef MODULE_SHT1X
+    /* The sht1x module needs to be initialized regardless of SAUL being used,
+     * as the shell commands rely on auto-initialization. auto_init_sht1x also
+     * performs SAUL registration, but only if module auto_init_saul is used.
+     */
     DEBUG("Auto init SHT1X module (SHT10/SHT11/SHT15 sensor driver).\n");
     extern void auto_init_sht1x(void);
     auto_init_sht1x();
--- a/sys/auto_init/saul/auto_init_sht1x.c
+++ b/sys/auto_init/saul/auto_init_sht1x.c
@@ -16,6 +16,11 @@
  *
  * @author      Marian Buschsieweke <marian.buschsieweke@ovgu.de>
  *
+ * Please note that the sht1x differs from other sensor modules, as
+ * initialization is performed regardless of the use of SAUL. However, when
+ * the module `auto_init_saul` is *not* used, the SAUL registration is skipped
+ * and the initialization does *not* depend on SAUL in this case.
+ *
  * @}
  */
 


maribu:
@PeterKietzmann, @MichelRottleuthner: As you have an MSB-A2 board, does one of you mind to test this PR?

maribu:
After thinking about it I realized that the bug was introduced because adequate documentation of differences between the initialization of the sht1x and other sensors is missing. I will update this PR tomorrow to provide this documentation.

PeterKietzmann:
@jia200x do you want this to be backported to the release branch?

jia200x:
@PeterKietzmann yes, let's backport it ;). I will add the label

jia200x:
@maribu could you provide a backport PR to `2018.10-branch`?

PeterKietzmann:
@maribu thanks for the documentation. To be honest, I don't understand what makes this driver so different from others. SAUL is one feature, but each driver should be auto-initialized if included via Makefile, no matter about SAUL.
@jia200x I didn't set the *Reviewed: 5-documentation* label. Can you please take over from here as I'm unavailable the next days?!

jia200x:
> @jia200x I didn't set the Reviewed: 5-documentation label. Can you please take over from here as I'm unavailable the next days?!

Sure, I take it-

> @jia200x: Sure. Both the added documentation and the bugfix, or only the bugfix? (I see little potential of breaking things by adding documentation, so I personally would go for both.)

I think in this case it's not necessary. If the commits are the same and doesn't break master, then there are really low chances that breaks the Release. If something gets broken, it should be fixed in the `master` branch as well as the `2018.10-branch`.

So, I would just backport this whole PR :)

maribu:
> To be honest, I don't understand what makes this driver so different from others. SAUL is one feature, but each driver should be auto-initialized if included via Makefile, no matter about SAUL.

I do agree with that. Providing known-to-work auto initialization will make life easier for application developers and using the `foobar_params_t` mechanism should still provide all the flexibility needed. However, all other sensor/actuator drivers will only perform auto initialization if SAUL is used as well, as they are wrapped into `#ifdef MODULE_AUTO_INIT_SAUL` ... `#endif`.

There is one drawback with the approach in the sht1x: The API is a bit clumsy. It asks for a pointer to an `sht1x_dev_t`, which is located in an global array in `sht1x_auto_init.c`.

This code snipped shows the clumsy-ness quite well:
https://github.com/RIOT-OS/RIOT/blob/289e635e5a6eed77c689ec973d945e1e96ea3d6c/sys/shell/commands/sc_sht1x.c#L32-L56

The shell command expects the user to give the number of the SHT1X device (e.g. 0 when only one is present) and has to get to the correct device "by hand". I was thinking about suggesting:

1) Use the same auto_init approach for every sensor/actuator as sht1x (that is: initialize the driver for each configured device regardless of SAUL, but only perform SAUL registration when SAUL is used.)

2. a) Change the API of all sensors and actuators so that the device descriptor is not passed as argument, but the array index of the device. (E.g. if two devices are configured via foobar_params_t devices 0 and 1 will be present.)

or

2. b) Extend the API with `sht1x_dev_t sht1x_by_index(unsigned idx)`, so that the device descriptor can be obtained easily. (This is clearly less intrusive and more consistent to the rest of RIOT's APIs, but not as straight-forward as 2.a))


Once the cc110x rewrite is done I will create an PR that converts the sht1x API to suggestion b) to receive feedback on it.

maribu:
@jia200x: The backport is done :-)

jia200x:
ok, the documentation makes sense.
I would still try to make to make it compliant to other drivers in follow up PRs. We should maybe open an issue ;)

PR #10268, opened 2018-10-26
cpu/nrf52/radio: initial support for nrf52's ieee802.15.4 radio

SemjonKerner:
This PR provides an initial driver implementation for nrf52's ieee802.15.4 radio device.
Therefor it adds:
- Function to filter mac header in the common ieee802.15.4 layer
- Autoinit functionalities
- Board support for nrf52840dk
- Support for nrf52 boards in examples/default

The driver does not include acknowledgements or retransmission on the link layer.
Also it does not utilize the carrier sense features of the nrf52. This is documented and should be tackled in future PRs.

I tested this driver with nrf52840dk and samr21-xpro with examples/gnrc_networking both sending and receiving packages continuous.

### DIFF
--- a/cpu/nrf52/include/nrf802154.h
+++ b/cpu/nrf52/include/nrf802154.h
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2019 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_nrf52_802154 IEEE802.15.4 Driver for nRF52840 SoCs
+ * @ingroup     drivers_netdev
+ * @brief       Driver for using the nRF52's radio in IEEE802.15.4 mode
+ *
+ * ## Implementation state ##
+ * Netdev events supported:
+ *
+ * - NETDEV_EVENT_RX_COMPLETE
+ * - NETDEV_EVENT_TX_COMPLETE
+ *
+ * Transmission options not yet impemented:
+ * - Send acknowledgement for packages
+ * - Request acknowledgement
+ * - Retransmit unacked packages
+ * - Carrier Sense Multiple Access (CSMA) and Implementation of Clear Channel
+ *   Assessment Control (CCACTRL)
+ *
+ * @{
+ *
+ * @file
+ * @brief       Driver interface for using the nRF52 in IEEE802.15.4 mode
+ *
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Semjon Kerner <semjon.kerner@fu-berlin.de>
+ */
+
+#ifndef NRF802154_H
+#define NRF802154_H
+
+#include "net/netdev/ieee802154.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Export the netdev device descriptor
+ */
+extern netdev_ieee802154_t nrf802154_dev;
+
+/**
+ * @brief   IEEE 802.15.4 radio timer configuration
+ *
+ *          this radio relies on a dedicated hardware timer to maintain IFS
+ *          the default timer may be overwritten in the board configuration
+ */
+#ifndef NRF802154_TIMER
+#define NRF802154_TIMER TIMER_DEV(1)
+#endif
+
+#endif /* NRF802154_H */
+/** @} */
--- a/cpu/nrf52/radio/nrf802154/nrf802154.c
+++ b/cpu/nrf52/radio/nrf802154/nrf802154.c
@@ -1,0 +1,446 @@
+/*
+ * Copyright (C) 2019 Freie Universität Berlin
+ *               2019 HAW Hamburg
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_nrf52_802154
+ * @{
+ *
+ * @file
+ * @brief       Implementation of the radio driver for nRF52 radios
+ *
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ * @author      Dimitri Nahm <dimitri.nahm@haw-hamburg.de>
+ * @author      Semjon Kerner <semjon.kerner@fu-berlin.de>
+ * @}
+ */
+
+#include <string.h>
+#include <errno.h>
+
+#include "cpu.h"
+#include "luid.h"
+#include "mutex.h"
+
+#include "net/ieee802154.h"
+#include "periph/timer.h"
+#include "net/netdev/ieee802154.h"
+#include "nrf802154.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+static const netdev_driver_t nrf802154_netdev_driver;
+
+netdev_ieee802154_t nrf802154_dev = {
+    {
+        .driver = &nrf802154_netdev_driver,
+        .event_callback = NULL,
+        .context = NULL,
+    },
+#ifdef MODULE_GNRC
+#ifdef MODULE_GNRC_SIXLOWPAN
+    .proto = GNRC_NETTYPE_SIXLOWPAN,
+#else
+    .proto = GNRC_NETTYPE_UNDEF,
+#endif
+#endif
+    .pan = IEEE802154_DEFAULT_PANID,
+    .short_addr = { 0, 0 },
+    .long_addr = { 0, 0, 0, 0, 0, 0, 0, 0 },
+    .chan = IEEE802154_DEFAULT_CHANNEL,
+    .flags = 0
+};
+
+static uint8_t rxbuf[IEEE802154_FRAME_LEN_MAX + 3]; /* len PHR + PSDU + LQI */
+static uint8_t txbuf[IEEE802154_FRAME_LEN_MAX + 3]; /* len PHR + PSDU + LQI */
+
+#define RX_COMPLETE         (0x1)
+#define TX_COMPLETE         (0x2)
+#define LIFS                (40U)
+#define SIFS                (12U)
+#define SIFS_MAXPKTSIZE     (18U)
+#define TIMER_FREQ          (250000UL)
+static volatile uint8_t _state;
+static mutex_t _txlock;
+
+/**
+ * @brief   Set radio into DISABLED state
+ */
+static void _disable(void)
+{
+    /* set device into DISABLED state */
+    if (NRF_RADIO->STATE != RADIO_STATE_STATE_Disabled) {
+        NRF_RADIO->EVENTS_DISABLED = 0;
+        NRF_RADIO->TASKS_DISABLE = 1;
+        while (!(NRF_RADIO->EVENTS_DISABLED)) {};
+        DEBUG("[nrf802154] Device state: DISABLED\n");
+    }
+}
+
+/**
+ * @brief   Set radio into RXIDLE state
+ */
+static void _enable_rx(void)
+{
+    DEBUG("[nrf802154] Set device state to RXIDLE\n");
+    /* set device into RXIDLE state */
+    if (NRF_RADIO->STATE != RADIO_STATE_STATE_RxIdle) {
+        _disable();
+    }
+    NRF_RADIO->PACKETPTR = (uint32_t)rxbuf;
+    NRF_RADIO->EVENTS_RXREADY = 0;
+    NRF_RADIO->TASKS_RXEN = 1;
+    while (!(NRF_RADIO->EVENTS_RXREADY)) {};
+    DEBUG("[nrf802154] Device state: RXIDLE\n");
+}
+
+/**
+ * @brief   Set radio into TXIDLE state
+ */
+static void _enable_tx(void)
+{
+    DEBUG("[nrf802154] Set device state to TXIDLE\n");
+    /* set device into TXIDLE state */
+    if (NRF_RADIO->STATE != RADIO_STATE_STATE_TxIdle) {
+        _disable();
+    }
+    NRF_RADIO->PACKETPTR = (uint32_t)txbuf;
+    NRF_RADIO->EVENTS_TXREADY = 0;
+    NRF_RADIO->TASKS_TXEN = 1;
+    while (!(NRF_RADIO->EVENTS_TXREADY)) {};
+    DEBUG("[nrf802154] Device state: TXIDLE\n");
+}
+
+/**
+ * @brief   Reset the RXIDLE state
+ */
+ static void _reset_rx(void)
+ {
+    if (NRF_RADIO->STATE != RADIO_STATE_STATE_RxIdle) {
+        return;
+    }
+
+    /* reset RX state and listen for new packets */
+    _state &= ~RX_COMPLETE;
+    NRF_RADIO->TASKS_START = 1;
+ }
+
+static void _set_chan(uint16_t chan)
+{
+    assert((chan >= IEEE802154_CHANNEL_MIN) && (chan <= IEEE802154_CHANNEL_MAX));
+    /* Channel map between 2400 MHZ ... 2500 MHz
+     * -> Frequency = 2400 + FREQUENCY (MHz) */
+    NRF_RADIO->FREQUENCY = (chan - 10) * 5;
+    nrf802154_dev.chan = chan;
+}
+
+static int16_t _get_txpower(void)
+{
+    int8_t txpower = (int8_t)NRF_RADIO->TXPOWER;
+    if (txpower < 0) {
+        return (int16_t)(0xff00 | txpower);
+    }
+    return (int16_t)txpower;
+}
+
+static void _set_txpower(int16_t txpower)
+{
+    if (txpower > 8) {
+        NRF_RADIO->TXPOWER = RADIO_TXPOWER_TXPOWER_Pos8dBm;
+    }
+    if (txpower > 1) {
+        NRF_RADIO->TXPOWER = (uint32_t)txpower;
+    }
+    else if (txpower > -1) {
+        NRF_RADIO->TXPOWER = RADIO_TXPOWER_TXPOWER_0dBm;
+    }
+    else if (txpower > -5) {
+        NRF_RADIO->TXPOWER = RADIO_TXPOWER_TXPOWER_Neg4dBm;
+    }
+    else if (txpower > -9) {
+        NRF_RADIO->TXPOWER = RADIO_TXPOWER_TXPOWER_Neg8dBm;
+    }
+    else if (txpower > -13) {
+        NRF_RADIO->TXPOWER = RADIO_TXPOWER_TXPOWER_Neg12dBm;
+    }
+    else if (txpower > -17) {
+        NRF_RADIO->TXPOWER = RADIO_TXPOWER_TXPOWER_Neg16dBm;
+    }
+    else if (txpower > -21) {
+        NRF_RADIO->TXPOWER = RADIO_TXPOWER_TXPOWER_Neg20dBm;
+    }
+    else {
+        NRF_RADIO->TXPOWER = RADIO_TXPOWER_TXPOWER_Neg40dBm;
+    }
+}
+
+static void _timer_cb(void *arg, int chan)
+{
+    (void)arg;
+    (void)chan;
+    mutex_unlock(&_txlock);
+    timer_stop(NRF802154_TIMER);
+    timer_clear(NRF802154_TIMER, 0);
+}
+
+static int _init(netdev_t *dev)
+{
+    (void)dev;
+
+    int result = timer_init(NRF802154_TIMER, TIMER_FREQ, _timer_cb, NULL);
+    assert(result >= 0);
+    (void)result;
+
+    /* initialize local variables */
+    mutex_init(&_txlock);
+
+    /* reset buffer */
+    rxbuf[0] = 0;
+    txbuf[0] = 0;
+    _state = 0;
+
+    /* power on peripheral */
+    NRF_RADIO->POWER = 1;
+
+    /* make sure the radio is disabled/stopped */
+    _disable();
+
+    /* we configure it to run in IEEE802.15.4 mode */
+    NRF_RADIO->MODE = RADIO_MODE_MODE_Ieee802154_250Kbit;
+    /* and set some fitting configuration */
+    NRF_RADIO->PCNF0 = ((8 << RADIO_PCNF0_LFLEN_Pos) |
+                        (RADIO_PCNF0_PLEN_32bitZero << RADIO_PCNF0_PLEN_Pos) |
+                        (RADIO_PCNF0_CRCINC_Include << RADIO_PCNF0_CRCINC_Pos));
+    NRF_RADIO->PCNF1 = IEEE802154_FRAME_LEN_MAX;
+    /* set start frame delimiter */
+    NRF_RADIO->SFD = IEEE802154_SFD;
+    /* set MHR filters */
+    NRF_RADIO->MHRMATCHCONF = 0;              /* Search Pattern Configuration */
+    NRF_RADIO->MHRMATCHMAS = 0xff0007ff;      /* Pattern mask */
+    /* configure CRC conform to IEEE802154 */
+    NRF_RADIO->CRCCNF = ((RADIO_CRCCNF_LEN_Two << RADIO_CRCCNF_LEN_Pos) |
+                         (RADIO_CRCCNF_SKIPADDR_Ieee802154 << RADIO_CRCCNF_SKIPADDR_Pos));
+    NRF_RADIO->CRCPOLY = 0x011021;
+    NRF_RADIO->CRCINIT = 0;
+
+    /* assign default addresses */
+    luid_get(nrf802154_dev.short_addr, IEEE802154_SHORT_ADDRESS_LEN);
+    luid_get(nrf802154_dev.long_addr, IEEE802154_LONG_ADDRESS_LEN);
+
+    /* set default channel */
+    _set_chan(nrf802154_dev.chan);
+
+    /* configure some shortcuts */
+    NRF_RADIO->SHORTS = RADIO_SHORTS_RXREADY_START_Msk | RADIO_SHORTS_TXREADY_START_Msk;
+
+    /* enable interrupts */
+    NVIC_EnableIRQ(RADIO_IRQn);
+    NRF_RADIO->INTENSET = RADIO_INTENSET_END_Msk;
+
+    /* switch to RX mode */
+    _enable_rx();
+
+    return 0;
+}
+
+static int _send(netdev_t *dev,  const iolist_t *iolist)
+{
+    (void)dev;
+
+    DEBUG("[nrf802154] Send a packet\n");
+
+    assert(iolist);
+
+    mutex_lock(&_txlock);
+
+    /* copy packet data into the transmit buffer */
+    unsigned int len = 0;
+    for (; iolist; iolist = iolist->iol_next) {
+        if ((IEEE802154_FCS_LEN + len + iolist->iol_len) > (IEEE802154_FRAME_LEN_MAX)) {
+            DEBUG("[nrf802154] send: unable to do so, packet is too large!\n");
+            mutex_unlock(&_txlock);
+            return -EOVERFLOW;
+        }
+        memcpy(&txbuf[len + 1], iolist->iol_base, iolist->iol_len);
+        len += iolist->iol_len;
+    }
+
+    /* specify the length of the package. */
+    txbuf[0] = len + IEEE802154_FCS_LEN;
+
+    /* trigger the actual transmission */
+    _enable_tx();
+    DEBUG("[nrf802154] send: putting %i byte into the ether\n", len);
+
+    /* set interframe spacing based on packet size */
+    unsigned int ifs = (len > SIFS_MAXPKTSIZE) ? LIFS : SIFS;
+    timer_set_absolute(NRF802154_TIMER, 0, ifs);
+
+    return len;
+}
+
+static int _recv(netdev_t *dev, void *buf, size_t len, void *info)
+{
+    (void)dev;
+    (void)info;
+
+    size_t pktlen = (size_t)rxbuf[0] - IEEE802154_FCS_LEN;
+
+    /* check if packet data is readable */
+    if (!(_state & RX_COMPLETE)) {
+        DEBUG("[nrf802154] recv: no packet data available\n");
+        return 0;
+    }
+
+    if (buf == NULL) {
+        if (len > 0) {
+            /* drop packet */
+            DEBUG("[nrf802154] recv: dropping packet of length %i\n", pktlen);
+        }
+        else {
+          /* return packet length */
+          DEBUG("[nrf802154] recv: return packet length: %i\n", pktlen);
+          return pktlen;
+        }
+    }
+    else if (len < pktlen) {
+        DEBUG("[nrf802154] recv: buffer is to small\n");
+        return -ENOBUFS;
+    }
+    else {
+        DEBUG("[nrf802154] recv: reading packet of length %i\n", pktlen);
+        memcpy(buf, &rxbuf[1], pktlen);
+    }
+
+    _reset_rx();
+
+    return (int)pktlen;
+}
+
+static void _isr(netdev_t *dev)
+{
+    if (!nrf802154_dev.netdev.event_callback) {
+        return;
+    }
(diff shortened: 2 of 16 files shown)


aabadie:
Very nice, good job !
FYI, today I added an nrf52840dk board in IoT-LAB (Saclay site): there you can test interop with samr21-xpro, iotlab-m3 and arduino-zero/xbee.

I gave a quick try and had some issues:
- with gnrc_networking, calling ifconfig crash the node
- I can send L2 bcast packets between nodes => there it quite works
- If I send L2 bcast packets from iotlab-m3/samr21-xpro, it works well => all nodes receive it
- If I send L2 bcast packets from nrf52840dk, it doesn't work well => all nodes receive it only once. After that none of them work anymore => no bcast paquet can be received afterward. I have to reboot/reset the nodes.

gebart:
@aabadie 
>If I send L2 bcast packets from nrf52840dk, it doesn't work well => all nodes receive it only once. After that none of them work anymore => no bcast paquet can be received afterward. I have to reboot/reset the nodes.

That seems serious. Did the nrf52840dk node cause a DoS condition in the iotlab-m3/samr21-xpro nodes?

SemjonKerner:
@aabadie thanks for testing already!

Indeed this seems to be a problem created by the nrf52, but maybe this is even a problem on all sides, if the hardfaults happen on all devices? My first guess is to look at some asserts in the common ieee15.4 layer. Can you provide some logs?

Also I used ifconfig alot and had not have any problems can you provide more info there, too?

miri64:
@aabadie can you pleaes post exactly what you did (i.e. dump of the commands typed / copy of the C code you used)? Because I tried what you described right now, with `ping6 ff02::1`, with `udp send ff02::1 ...` and even (with `gnrc_txtsnd` compiled in) with `txtsnd 7 bcast "abcd"` and I wasn't able to reproduce what you described regarding broadcast. Likewise, as @SemjonKerner I wasn't able to reproduce the `ifconfig` issue.

aabadie:
I used the gnrc_networking and default examples of RIOT on IoT-LAB nodes:
1. Check that I'm on the right branch:
```
$ git log -1 --oneline
2509ae5fb (HEAD -> add_nrf154, SemjonKerner/add_nrf154) examples/default: add ieee802.15.4 radio support for all nrf52840 boards
```
2. Start an experiment with different nodes:
```
$ iotlab-experiment -n test_pr_10268 -d 200 -l saclay,m3,1 -l saclay,samr21,1 -l nrf52840dk,1
```
3. Build/flash/term gnrc_networking on nrf52840dk:
```
$ make IOTLAB_NODE=nrf52840dk-1.saclay.iot-lab.info BOARD=nrf52840dk -C examples/gnrc_networking flash term
> help
help
udp                  send data over UDP and listen on UDP ports
reboot               Reboot the node
ps                   Prints information about running threads.
ping6                Ping via ICMPv6
random_init          initializes the PRNG
random_get           returns 32 bit of pseudo randomness
nib                  Configure neighbor information base
ifconfig             Configure network interfaces
rpl                  rpl configuration tool ('rpl help' for more information)
> reboot
reboot
main(): This is RIOT! (Version: 2018.10-RC1-24-g2509ae-trek-saclay-add_nrf154)
RIOT network stack example application
All up, running the shell now
> ifconfig
ifconfig  => hangs here
```
4. Build/flash/term default application on all nodes (in different shells):
```
$  make IOTLAB_NODE=nrf52840dk-1.saclay.iot-lab.info BOARD=nrf52840dk -C examples/default flash term
$ make IOTLAB_NODE=m3-1.saclay.iot-lab.info BOARD=iotlab-m3 -C examples/default flash term
$ make IOTLAB_NODE=samr21-1.saclay.iot-lab.info BOARD=samr21-xpro -C examples/default flash term
```
5. Send a bcast from the samr21 node:
```
> txtsnd 4 bcast qlkjksdjfksdjflkdsjf
```
All other nodes (m3 and nrf52840dk) prints a packet dump:
```
PKTDUMP: data received:
~~ SNIP  0 - size:  20 byte, type: NETTYPE_UNDEF (0)
00000000  71  6C  6B  6A  6B  73  64  6A  66  6B  73  64  6A  66  6C  6B
00000010  64  73  6A  66
~~ SNIP  1 - size:  12 byte, type: NETTYPE_NETIF (-1)
if_pid: 4  rssi: 4008  lqi: 0
flags: BROADCAST 
src_l2addr: 30:1A
dst_l2addr: FF:FF
~~ PKT    -  2 snips, total size:  32 byte
```
Send again and all other nodes still dumps a received packet. This can be reproduced any time from the m3 and samr21 node.
6. Send a bcast from the **nrf52840dk** node:
```
> txtsnd 4 bcast qlkjksdjfksdjflkdsjf
```
Other nodes receive it only once. The next sends trigger no dump. Even worse, any raw packet sent from samr21 or m3 is not received by other nodes.

miri64:
> 3. Build/flash/term gnrc_networking on nrf52840dk:

I can reproduce that on IoT-LAB as well :-/. But not on the local `nrf52840dk` we have here. Maybe we have a different revision?

nrf52840dk front

nrf52840dk back

5. and 6. I also can reproduce *in* IoT-Lab but not locally :-/

aabadie:
I'm not sure (and won't be before next week ;) ) but I think the IoT-LAB board is a DK version and the one you are using is the PDK (Preview-DK). I don't know how this could make a difference since I haven't check the datasheets in detail.

aabadie:
Another thing on the IoT-LAB node: there's an X-nucleo extension (with sensors) plugged on it. It uses the Arduino-like pinout. I'll try to remove it and test again but I'm not sure it will change something.

haukepetersen:
The boards we have around here are `nrf52pdk` (p for preview). But from what I have seen, there should not be any significant difference between it and the `nrf52dk` as used in the IoT-LAB. The pin layout should also not matter, as the radio does not need any pins. I also scanned the board datasheets once more side-by-side, and I can't find anything that would lead to different behavior.

So next thing would be to get the actual MCU stepping of the boards we use. The pdk I have here uses a `nrf52840 QIAAAA` -> hardware revision 3rd`A`, engineering device 4th 'A'. For that line of CPU there are some issues with the radio described in the [errata sheet](http://infocenter.nordicsemi.com/pdf/nRF52840_Engineering_A_Errata_v1.6.pdf), but they don't seem to lead to any 'DOS' like behavior.

@aabadie would you mind to check the MCU version present on the boards in the IoT-LAB? Thx

SemjonKerner:
@aabadie Okay, i can reproduce the ifconfig failure in the iot-lab. There seems to be a blocking mutex lock while changing states. The cause is still unclear to me.
I am looking into the statemachines, so can you still provide the revision number of the saclay nrf52840dk cpu? This will probably help a lot with debugging.

aabadie:
> would you mind to check the MCU version present on the boards in the IoT-LAB?

I just checked the board in IoT-LAB and it's a nrf52840DK (not Preview) with nrf52840 rev 1. The CPU model is QIAA-Cx0.

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,412 @@
...
+    DEBUG("Send a packet\n");
+
+    assert(iolist);
+
+    mutex_lock(&txlock);
+    /* copy packet data into the transmit buffer */
+    int len = 0;
+    for (; iolist; iolist = iolist->iol_next) {
+        if ((IEEE802154_FCS_LEN + len + iolist->iol_len) > (IEEE802154_FRAME_LEN_MAX)) {
+            DEBUG("[nrf802154] send: unable to do so, packet is too large!\n");
```
bergzand: I think you at least want to unlock the mutex here.

bergzand:
After some debugging this is what I think is happening:
 1. Frame submitted for transfer
 2. Radio enters TX state
 3. Frame is transmitted
 4. `EVENTS_END` is set, IRQ is called. At this point the radio is still in the Tx state and not yet in the TxIdle state.
 5. The `isr_radio` doesn't handle anything since the radio periph is not in the RxIdle state or in the TxIdle state.

At this point the mutex is still locked and the netif thread hangs at a `mutex_lock` statement trying to send a second frame. This also causes the `netif` shell command to hang due to netopt requests not being answered.

cpu/nrf52/radio/nrf802154/nrf802154.c
SemjonKerner: Definitely, this is a huge mistake. Thank you for spotting this.

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,412 @@
...
+    .long_addr = { 0, 0, 0, 0, 0, 0, 0, 0 },
+    .chan = IEEE802154_DEFAULT_CHANNEL,
+    .flags = 0
+};
+
+static uint8_t rxbuf[IEEE802154_FRAME_LEN_MAX + 3]; /* len PHR + PSDU + LQI */
+static uint8_t txbuf[IEEE802154_FRAME_LEN_MAX + 3]; /* len PHR + PSDU + LQI */
+
+#define RX_AVAIL (1)
+static volatile uint8_t _state;
```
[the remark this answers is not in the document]
SemjonKerner: Right now the state variable just holds information about the buffer states and thus should only be written with very specific values. Anyway your suggestion is very reasonable and I'll probably implement that.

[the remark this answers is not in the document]
bergzand: What I was trying to say (early in the morning) is that `1 << SOME_NETDEV_EVENT` can result in something larger than `UINT8_MAX`. Not sure if that was clear in my original wording.

bergzand:
@SemjonKerner I've opened semjonkerner#1 to have my changes to the ISR handler at least somewhere public.

bergzand:
> If my description of what happens is indeed what is happening on the IoT-Lab testbed, then the DOS issue could be caused by the nrf52840dk never leaving the Tx Idle state. If, in that state it is transmitting a whole lot of nothing, then the other radios in the vicinity are unable to transmit due to a failing CSMA-CA check.

This was tested with a rev 1 chip by the way.

SemjonKerner:
@bergzand What you describe is exactly what I found on the IoT-lab boards. Strangely after triggering the interrupt the nrf52840 in the IoT-lab takes something around 500 cpu-cycles to go into the TxIdle state, while my preview-board changes the state somewhat simultaneously with the interrupt.
I did not find anything about that in the errata, so right now I can not evaluate if this is a bug or a feature.
I assume there is another (unrelated) interrupt after the payload but before the FCS and thus the mutex would run into a deadlock in the send function. (Also ifconfig uses the get function which then is blocked, too.)

But your refactoring seems to work well on both boards. Thank you very much for your effort!

SemjonKerner:
@bergzand I tested your code and talked with @haukepetersen about it. There may be a bug in the statemachine and I will ask in the nordic forum for directions. For me your refactored isr looks like a workaround, but we agreed that it is a forgiving approach. So, thank you again.

@aabadie Thank you for adding the nrf52840dk and for your verbose test description. It helped a lot  to reproduce the deadlock, especially because i am new to IoT-lab. From my point of view the new isr function should fix all mentioned problems. So feel free to have another go :)

bergzand:
> @bergzand I tested your code and talked with @haukepetersen about it. There may be a bug in the statemachine and I will ask in the nordic forum for directions. For me your refactored isr looks like a workaround, but we agreed that it is a forgiving approach. So, thank you again.

I agree that it could be a bug in their state machine somewhere. It's pretty weird that the interrupt already triggers while the state machine itself is not yet in the state activating the interrupt. Let's wait what the nordic guys say on the forum, I'm pretty curious what is causing the trouble here.

miri64:
I tested the current state on the testbed as well and am able to confirm, that the issues pointed out by @aabadie don't come up anymor.

miri64:
I can't however ping the `nrf` on the testbed from the other nodes using `gnrc_networking`. I'll try the same in my local set-up.

aabadie:
I'm happy to read the great news here. I'll give another try tomorrow.
Thanks for the good work @SemjonKerner @bergzand !

miri64:
> I can't however ping the nrf on the testbed from the other nodes using gnrc_networking. I'll try the same in my local set-up.

Ah, it was because I did not deactivate `ACK_REQ` on the `samr21-xpro` and `iotlab-m3` ^^"

aabadie:
Just in case, I ran some test on IoT-LAB with the new version of this PR and it works well. I also had to disable ACK_REQ flag (`ifconfig 7 -ack_req`) to be able to ping from iotlab-m3, samr21 nodes.
I could send and receive UDP packets between all nodes. Great!

miri64:
I'll try to test this tomorrow

miri64:
I've tested it now. Basic operations work like `txtsnd` with different combinations of long, short or broadcast address as source or destination (broadcast of course not as source) and different channels and PAN IDs (on the local setup, did not test in IoT-LAB yet). I was also able to configure TX power without breaking things. However, when trying to 6LoWPAN functionalities I ran into the problem that fragmented packets can't be received by other nodes. I'm suspecting that the value returned by `NETOPT_MAX_PACKET_SIZE` is wrong and thus the fragments can't be reassembled properly. If you are at the office tomorrow we can have a look at the PCAP dump together to figure out what's happening.

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,432 @@
...
+    DEBUG("[nrf802154] Device state: RXIDLE\n");
+}
+
+/**
+ * @brief   Set radio into TXIDLE state
+ */
+static void _enable_tx(void)
+{
+    DEBUG("[nrf802154] Set device state to TXIDLE\n");
+    NRF_RADIO->PACKETPTR = (uint32_t)txbuf;
```
bergzand: This lightly smells like a race condition here. Is it possible that right after the address is changed here to point to the TX buffer, but while the radio is still in RX mode, a frame is received and accidentally written to the TX buffer instead of the RX buffer?

To be on the safe side, wouldn't it be better to move this register write below the `_disable()` call?
https://github.com/RIOT-OS/RIOT/blob/427111035fcbed0947efe283872d6ca26f2c103d/cpu/nrf52/radio/nrf802154/nrf802154.c#L109-L112

miri64:
> I'm suspecting that the value returned by NETOPT_MAX_PACKET_SIZE is wrong and thus the fragments can't be reassembled properly. If you are at the office tomorrow we can have a look at the PCAP dump together to figure out what's happening.

It wasn't that after all. The NRF just didn't send some of the subsequent fragments. As a solution to that I proposed #10403. However, there is still a problem with reception of packets on the *`samr21-xpro`* (so it might be totally unrelated to this PR): for some reason for the second packet the reception interrupt is not fired. The NRF can handle the second packet just fine.

bergzand:
> It wasn't that after all. The NRF just didn't send some of the subsequent fragments. As a solution to that I proposed #10403. However, there is still a problem with reception of packets on the _`samr21-xpro`_ (so it might be totally unrelated to this PR): for some reason for the second fragment the reception interrupt is not fired (when sending a packet fragmented in two). The NRF can handle the second packet just fine.

Could it be that the nrf52840 is transmitting too fast (outside the spec of 802.15.4)? There is a `TIFS` register on the nrf52840 to configure the inter frame spacing. According to the datasheet this should be automatically configured when the peripheral is in "assisted mode" though.

miri64:
> Could it be that the nrf52840 is transmitting too fast (outside the spec of 802.15.4)? There is a TIFS register on the nrf52840 to configure the inter frame spacing. According to the datasheet this should be automatically configured when the peripheral is in "assisted mode" though.

Maybe if this is the problem #10403 is not necessary?

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,432 @@
...
+        nrf802154_dev.netdev.event_callback(dev, NETDEV_EVENT_TX_COMPLETE);
+        _state &= ~NRF_TX_COMPLETE;
+    }
+}
+
+static int _get(netdev_t *dev, netopt_t opt, void *value, size_t max_len)
+{
+    assert(dev);
+
+    DEBUG("[nrf802154] get: %s\n", netopt2str(opt));
```
miri64: Can you apply the following, so we can also compile without `USEMODULE=netopt` and `ENABLE_DEBUG==1`
```suggestion
#ifdef MODULE_NETOPT
    DEBUG("[nrf802154] get: %s\n", netopt2str(opt));
#else
    DEBUG("[nrf802154] get: %d\n", opt);
#endif
```

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,432 @@
...
+                                          opt, value, max_len);
+    }
+}
+
+static int _set(netdev_t *dev, netopt_t opt,
+                const void *value, size_t value_len)
+{
+    assert(dev);
+
+    DEBUG("[nrf802154] set: %s\n", netopt2str(opt));
```
miri64: Can you apply the following, so we can also compile without `USEMODULE=netopt` and `ENABLE_DEBUG==1`
```suggestion
#ifdef MODULE_NETOPT
    DEBUG("[nrf802154] set: %s\n", netopt2str(opt));
#else
    DEBUG("[nrf802154] set: %d\n", opt);
#endif
```

SemjonKerner:
> Could it be that the nrf52840 is transmitting too fast (outside the spec of 802.15.4)? There is a `TIFS` register on the nrf52840 to configure the inter frame spacing. According to the datasheet this should be automatically configured when the peripheral is in "assisted mode" though.

So TIFS is probably not enforced, because it needs the End->Disable and Disabled->TxEnable Shortcuts to be triggered. Right now I am not sure if there is an easy method to utilize those shortcuts.

hubertmis:
Shorts with TIFS are tricky. We were using TIFS with shorts for some time in the https://github.com/NordicSemiconductor/nRF-IEEE-802.15.4-radio-driver project, but it resulted in unwanted transmission in rare conditions. We switched to TIMER peripheral instead to overcome problems.

aabadie:
@SemjonKerner, any updates here ? It would be to have it in the release, since it seems to be nearly ready.

SemjonKerner:
@aabadie sadly the last 3 weeks i was very busy with other things and did not find the source of the fragmentation bug.
It seems like the driver is loosing the very first fragment (reproduceable) of a fragmented packet.
I tried a workaround, such that when a second fragment shall be send, we wait a moment (printf magic) and then everything works and i can send packets of at least 1000 bytes. Sadly this is not a way to go and didn't give me further information on the bug.
I assume, the first fragment is beeing overwritten (with the second one) by 6lowpan even before it is locked with the mutex, but as I said, i had no time to find the exact source of the bug.
Once a fragment made it through the mutex guard everything is fine and all other fragments are send and received, but this is of no use without the first fragment.

Also I could not utilize timings. I tried to set a timestamp at the end of the interrupt and waitloop the maximum intervall before sending the next packet, but than again the first fragment didn't even raise an interrupt since it was overwritten before sending.

SemjonKerner:
@aabadie I will not probably work on the driver until the 22.Feb (due to exams).
Please check out the nordic driver, I am sure it is a very good implementation, yet it extends a lot of features that are not required for riot. Also the style is not intuitive to me, so i'd waste days to accomplish that.
I would prefer if some experienced users like you or @bergzand could have a look into the bug and possibly present a pull request, this i could review and merge quickly.

SemjonKerner:
So @aabadie @bergzand @miri64 , the fragmentation should work now. I addressed all requests, removed L2 Netstat and also added a timer to achieve correct interframe spacing. I tested the fix on both board revisions. Please feel free to test it.

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,448 @@
...
+    timer_clear(TIMER_DEV(1), 0);
+}
+
+static int _init(netdev_t *dev)
+{
+    assert(dev);
+    (void)dev;
+
+    if (timer_init(TIMER_DEV(1), 250000UL, _timer_cb, NULL) < 0) {
+        DEBUG("Timer Init failed\n");
```
haukepetersen: This would be a good place to use an `assert` :-)

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,448 @@
...
+
+    /* specify the length of the package. */
+    txbuf[0] = len + IEEE802154_FCS_LEN;
+
+    /* trigger the actual transmission */
+    _enable_tx();
+    DEBUG("[nrf802154] send: putting %i byte into the ether\n", len);
+
+    /* set interframe spacing based on packet size */
+    if ((unsigned int)len > NRF_SIFS_MAXPKTSIZE)
```
haukepetersen: to prevent you adding the missing parenthesis, how about 
```c
unsigned ifs = (len > NRF_SIFS_MAXPKTSIZE) ? NRF_LIFS : NRF_SIFS;
timer_set_absolute(TIMER_DEV(1), 0, ifs);
```
:-)

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,448 @@
...
+    (void)dev;
+
+    DEBUG("[nrf802154] Send a packet\n");
+
+    assert(iolist);
+
+    mutex_lock(&txlock);
+
+    /* copy packet data into the transmit buffer */
+    int len = 0;
```
haukepetersen: and any reason we don't simply make `len` `unsigned`? Would save some casting below and the value is always >0 anyway, right?

drivers/Makefile.dep
```diff
@@ -342,6 +342,13 @@ ifneq (,$(filter my9221,$(USEMODULE)))
   USEMODULE += xtimer
 endif
 
+ifneq (,$(filter nrf802154,$(USEMODULE)))
```
haukepetersen: From a structural point of view, I would move this dependency into the CPUs Makefile.dep. On paper this is of course a driver, but it is completely CPU related and none of its code resides in the `/driver` code path, so I'd say its dependency defs should also not reside here.

drivers/include/net/netdev/ieee802154.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016 Freie Universität Berlin
+ * Copyright (C) 2018 Freie Universität Berlin
```
haukepetersen: rather `2016,2018`?

sys/include/net/ieee802154.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-16 Freie Universität Berlin
+ * Copyright (C) 2018 Freie Universität Berlin
```
haukepetersen: see comment above

drivers/Makefile.dep
SemjonKerner: I get the reasoning, but which Makefile.dep do you mean exactly?

benpicco:
Nice!

fyi, when I do `ping -I lowpan0 fe80::2c24:f6ca:24ea:740e -s 1000` on Linux
I get `gnrc_netif: possibly lost interrupt.` and 100% packet loss.

Those messages start to appear with payload size > 850, but are not accompanied by packet loss at this point yet.

Smaller payloads (<850) are working fine.

SemjonKerner:
> fyi, when I do `ping -I lowpan0 fe80::2c24:f6ca:24ea:740e -s 1000` on Linux
> I get `gnrc_netif: possibly lost interrupt.` and 100% packet loss.
> 
> Those messages start to appear with payload size > 850, but are not accompanied by packet loss at this point yet.

Thanks for testing @benpicco ! I found a "boundary" too, mine was a bit higher (1000ish) and always acompanied with errors by 6lowpan. I am not to deep into the standard, but my first guess is, that this boundary is fine.

However, first thing you should do is testing without ack_req and autoack on the other board, since this driver does not implement acks yet. Then at least your duplicates should vanish, and probably your boundary goes up, too.

benpicco:
With `-autoack` and `-ack_req` the two boards can talk to each other.

SemjonKerner:
> I can confirm that with `-autoack` and `-ack_req` the two boards can talk to each other.

And do you achieve a higher maximum package size?

benpicco:
No, the maximum ping payload size is not affected by this. The result is the same no matter if I'm pinging from RIOT or from Linux.

emmanuelsearch:
@SemjonKerner I tested on IoT-lab, between two nrf52840-dk, and between an nrf52840-dk and an samr21-xpro.
- nrf52840-dk => nrf52840-dk seems to work (I noticed some packet loss with -c 10 -s 150, but could be due to other aspects)
- nrf52840-dk => samr21-xpro seems to work too (same remark as above)
- samr21-xpro => does not work at all, even for normal size ping

Can you reproduce the problem? (I checked that samr21-xpro => samr21-xpro pinging works too).

emmanuelsearch:
and I confirm that with modified IF settings on the samr21, things work fine!

miri64:
> fyi, when I do `ping -I lowpan0 fe80::2c24:f6ca:24ea:740e -s 1000` on Linux
> I get `gnrc_netif: possibly lost interrupt.` and 100% packet loss.
> 
> Those messages start to appear with payload size > 850, but are not accompanied by packet loss at this point yet.
> 
> Smaller payloads (<850) are working fine.

#11067 should fix that.

SemjonKerner:
So @miri64 and I discussed this problem and found, that the netif message queue might be to small for certain radios, thus dropping interrupts for fragments once the queue is full. @miri64 proposed a PR #11067 that raises the queue size and exposes it, so other drivers may change it.

We also found that this problem occured on the kw2x radio and is fixed with the PR. Hence we decided this is the way to go.

cpu/nrf52/include/nrf802154.h
```diff
@@ -0,0 +1,59 @@
...
+#endif
+
+/**
+ * @brief   Export the netdev device descriptor
+ */
+extern netdev_ieee802154_t nrf802154_dev;
+
+/**
+ * @brief   IEEE 802.15.4 radio timer configuration
+ * @note    this radio relies on a dedicated hardware timer to maintain IFS
```
miri64: Nit-pick: rather looks like a detailed description to me, so just drop the `@note`

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,446 @@
...
+#include "net/netdev/ieee802154.h"
+#include "nrf802154.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+static const netdev_driver_t nrf802154_netdev_driver;
+
+netdev_ieee802154_t nrf802154_dev = {
+    {
```
miri64: Very elegant way to not require a set-up routine. ;-) Maybe we should adapt this for other drivers if applicable ;-).

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,446 @@
...
+    },
+#ifdef MODULE_GNRC
+#ifdef MODULE_GNRC_SIXLOWPAN
+    .proto = GNRC_NETTYPE_SIXLOWPAN,
+#else
+    .proto = GNRC_NETTYPE_UNDEF,
+#endif
+#endif
+    .pan = IEEE802154_DEFAULT_PANID,
+    .short_addr = { 0, 0 },
```
miri64: Everything not mentioned here will be set to 0, so I suggest to leave the address parts out.

sys/include/net/ieee802154.h
```diff
@@ -32,14 +32,17 @@ extern "C" {
 #endif
 
 /**
- * @brief IEEE 802.15.4 address lengths
```
miri64: Any reason for these unrelated doc removals?

sys/include/net/ieee802154.h
```diff
@@ -112,43 +115,33 @@ extern const uint8_t ieee802154_addr_bcast[IEEE802154_ADDR_BCAST_LEN];
...
 #ifndef IEEE802154_DEFAULT_PANID
 #define IEEE802154_DEFAULT_PANID            (0x0023U)
 #endif
 
-/**
- * @brief IEEE802.15.4 default TX power (in dBm)
- */
 #ifndef IEEE802154_DEFAULT_TXPOWER
-#define IEEE802154_DEFAULT_TXPOWER          (0)
+#define IEEE802154_DEFAULT_TXPOWER          (0) /* in dBm */
```
miri64: This is already documented in the `@brief` you removed a few lines up ;-).

examples/default/Makefile
```diff
@@ -37,7 +37,7 @@ USEMODULE += ps
 USEMODULE += saul_default
 
 BOARD_PROVIDES_NETIF := acd52832 airfy-beacon b-l072z-lrwan1 cc2538dk fox iotlab-m3 iotlab-a8-m3 mulle \
-        microbit native nrf51dk nrf51dongle nrf52dk nrf6310 openmote-cc2538 pba-d-01-kw2x \
```
miri64: Merge error to remove `nrf51dk`?

cpu/nrf52/radio/nrf802154/nrf802154.c
SemjonKerner: This is work of either Hauke or Dimitri, but in their name thanks for the compliment.

sys/include/net/ieee802154.h
SemjonKerner: There is nothing removed here, I just inserted `IEEE802154_SFD` above. This is just githubs highlighting :)

cpu/nrf52/radio/nrf802154/nrf802154.c
SemjonKerner: I dont really understand the reasoning.
I have no strong feelings on this code, but it makes sense to me, to set everything in one place. Especially if the memory is used anyway.
Could you explain this a bit more?

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,446 @@
...
+static uint8_t txbuf[IEEE802154_FRAME_LEN_MAX + 3]; /* len PHR + PSDU + LQI */
+
+#define RX_COMPLETE         (0x1)
+#define TX_COMPLETE         (0x2)
+#define LIFS                (40U)
+#define SIFS                (12U)
+#define SIFS_MAXPKTSIZE     (18U)
+#define TIMER_FREQ          (250000UL)
+static volatile uint8_t _state;
+static mutex_t txlock;
```
haukepetersen: nitpicky: shouldn't this variable also start with an underscore?

Citrullin:
I guess this one does not with the nRF52832, right?

haukepetersen:
@Citrullin yes, correct. Only the newer nRF52840 does have support for IEEE802.15.4.

haukepetersen:
@bergzand would you mind to re-check this PR once more and see if all your change requests have been addressed? Thanks!

cpu/nrf52/radio/nrf802154/nrf802154.c
```diff
@@ -0,0 +1,446 @@
...
+    NRF_RADIO->MHRMATCHMAS = 0xff0007ff;      /* Pattern mask */
+    /* configure CRC conform to IEEE802154 */
+    NRF_RADIO->CRCCNF = ((RADIO_CRCCNF_LEN_Two << RADIO_CRCCNF_LEN_Pos) |
+                         (RADIO_CRCCNF_SKIPADDR_Ieee802154 << RADIO_CRCCNF_SKIPADDR_Pos));
+    NRF_RADIO->CRCPOLY = 0x011021;
+    NRF_RADIO->CRCINIT = 0;
+
+    /* assign default addresses */
+    luid_get(nrf802154_dev.short_addr, IEEE802154_SHORT_ADDRESS_LEN);
+    luid_get(nrf802154_dev.long_addr, IEEE802154_LONG_ADDRESS_LEN);
```
bergzand: Minor nit: Other drivers request a single `IEEE802154_LONG_ADDRESS_LEN` luid and reuse part of it to set the `IEEE802154_SHORT_ADDRESS`, see [here](https://github.com/RIOT-OS/RIOT/blob/master/drivers/at86rf2xx/at86rf2xx.c#L74) and [here](https://github.com/RIOT-OS/RIOT/blob/master/drivers/mrf24j40/mrf24j40.c#L57). Not blocking on this, only want to point it out that this diverges from other drivers.

bergzand:
No further comments besides the nit above, my issues have been addressed. Thanks for the awesome work @SemjonKerner

haukepetersen:
Thanks. I will simply merge this PR for now and prepare a small fix in a separate PR for the short address 'issue'.

Thanks again @SemjonKerner for the good work!

haukepetersen:
See https://github.com/RIOT-OS/RIOT/pull/11100 for a fix on re-using the parts of the long address as short address.

PR #10271, opened 2018-10-26
Fix return value of aes_init for keys of incorrect length

### LINKED ISSUE #10175: No error returned from aes_init when a key with a bad size is used
#### Description

The `aes_init` function in the crypto module takes a key as a parameter which has to be either 16, 24 or 32 bytes long.
But the function does not check the key's length, even though the documentation of `cipher_init` states that it should. An error won't be thrown until the `aes_encrypt`or `aes_decrypt` functions are called.

Call `aes_init` with a *keySize` value other then 16, 24 or 32.

#### Expected results

CIPHER_ERR_INVALID_KEY_SIZE

#### Actual results

CIPHER_INIT_SUCCESS

mtausig:
### Contribution description

A proper error code is returned if a key with unsupported (either by the implementation or the AES algorithm) length is passed to `aes_init`.

### Testing procedure

The unit tests in the `tests-crypto` folder have been updated.

### Issues/PRs references

This fixes #10175

### DIFF
--- a/sys/crypto/aes.c
+++ b/sys/crypto/aes.c
@@ -124,7 +124,8 @@ static const u32 Te0[256] = {
     #define Te1(n)  ((Te0[n] >>  8) | (Te0[n] << 24))
     #define Te2(n)  ((Te0[n] >> 16) | (Te0[n] << 16))
     #define Te3(n)  ((Te0[n] >> 24) | (Te0[n] <<  8))
-    #define Te4(n)  (((Te0[n] & 0x00FFFF00) >> 8) | ((Te0[n] & 0x00FFFF00) << 8))
+    #define Te4(n)  (((Te0[n] & 0x00FFFF00) >> 8) | \
+                     ((Te0[n] & 0x00FFFF00) << 8))
 #else /* MODULE_CRYPTO_AES_PRECALCULATED */
     #define Te0(n)  (Te0[n])
     #define Te1(n)  (Te1[n])
@@ -472,11 +473,12 @@ static const u32 Td0[256] = {
     #define Td2(n)  ((Td0[n] >> 16) | (Td0[n] << 16))
     #define Td3(n)  ((Td0[n] >> 24) | (Td0[n] <<  8))
 
-    /* helper to prevent the u8 to be promoted to signed int, which would turn
-     * left shift by 24 into undefined behaviour */
+/* helper to prevent the u8 to be promoted to signed int, which would turn
+ * left shift by 24 into undefined behaviour */
     #define Td4u(n) ((u32)Td4[n])
 
-    #define Td4(n)  (Td4u(n) | (Td4u(n) << 8) | (Td4u(n) << 16) | (Td4u(n)  << 24))
+    #define Td4(n)  (Td4u(n) | (Td4u(n) << 8) | (Td4u(n) << 16) | \
+                     (Td4u(n) << 24))
 
 static const u8 Td4[256] = {
     0x52U, 0x09U, 0x6aU, 0xd5U, 0x30U, 0x36U, 0xa5U, 0x38U,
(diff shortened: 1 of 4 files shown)


jcarrano:
Please make the first commit message shorter. Maybe something like: 
`sys/crypto: aes_init(): fail when called with unsupported key length.`

(also, while you are at it you could paste your PR description into the commit message too)

bergzand:
> The conditional that is added to `aes_init` to check if `keySize != AES_KEY_SIZE` seems to conflict with the construct to extend the key when it is not long enough.

Having an error message when the supplied key is too long is something I can get behind though.

mtausig:
The implementation of AES that is currently in RIOT does only support a single key size (16 bytes) which is defined in AES_KEY_SIZE. The functions `aes_set_decrypt_key` and `aes_set_encrypt_key` which are called by `aes_decrypt`/`aes_encrypt` uses the this fixed key size. If I read the code correctly, trying to pass a 17 byte key to `aes_init` should lead to the same result as passing the 16 byte subkey.

The automatic lengthening of a key shorter than 16 byte by copying the first bytes is a bit weird IMO (and also undocumented) and should be avoided. The AES algorithm is specified for exactly three keylengths (16, 24, and 32 bytes), passing anything else but those should cause an error.

bergzand:
> If I read the code correctly, trying to pass a 17 byte key to `aes_init` should lead to the same result as passing the 16 byte subkey.

That's also how I'm reading the code and I agree that that case should indeed result in an error as it would give a developer a false sense of (heightened) security.

> The automatic lengthening of a key shorter than 16 byte by copying the first bytes is a bit weird IMO (and also undocumented) and should be avoided. The AES algorithm is specified for exactly three keylengths (16, 24, and 32 bytes), passing anything else but those should cause an error.

Not that I don't agree with you, but do you have a citation/documentation for this claim?
The other thing is that only allowing 16 byte keys is essentially an API change here. Again, not necessarily a problem, but more an active decision on whether we want to change the current behaviour.

mtausig:
> > The automatic lengthening of a key shorter than 16 byte by copying the first bytes is a bit weird IMO (and also undocumented) and should be avoided. The AES algorithm is specified for exactly three keylengths (16, 24, and 32 bytes), passing anything else but those should cause an error.
> 
> Not that I don't agree with you, but do you have a citation/documentation for this claim?

Quote from the book *The Design of Rijndael, AES — The Advanced Encryption Standard* by *Joan Daemen* and *Vincent Rijmen* (http://jda.noekeon.org/JDA_VRI_Rijndael_2002.pdf) (3.1, p46):

    The AES fixes the block length to 128 bits, and supports key lengths of 128, 192 or 256 bits only.

> The other thing is that only allowing 16 byte keys is essentially an API change here. Again, not necessarily a problem, but more an active decision on whether we want to change the current behaviour.

I agree. But the previous behaviour was errornous, anyone trying to use AES with a 8 or 19 byte key should not be made believing that anything good and secure and predictable will come out of it.

mtausig:
I updated the documentation to make developers aware of the implementations limitations (8466341).

bergzand:
> Quote from the book _The Design of Rijndael, AES — The Advanced Encryption Standard_ by _Joan Daemen_ and _Vincent Rijmen_ (http://jda.noekeon.org/JDA_VRI_Rijndael_2002.pdf) (3.1, p46):

Thanks!

> > The other thing is that only allowing 16 byte keys is essentially an API change here. Again, not necessarily a problem, but more an active decision on whether we want to change the current behaviour.
> 
> That is correct. But the previous behaviour was errornous, anyone trying to use AES with a 8 or 19 byte key should not be made believing that anything good and secure and predictable will come out of it.

I think we agree here that the current behaviour is not something that we want. The only problem is that people might be relying on it.

> I updated the documentation to make developers aware of the implementations limitations ([8466341](https://github.com/RIOT-OS/RIOT/commit/8466341a211d9cb534718b14b5472f49bc3bd75f)).

Thanks, that helps.

I'll try to properly review the whole change set today/this week.

jcarrano:
@bergzand , what's the veredict on this. I think it is a desirable behaviour that invalid parameters fail as early as possible, during initialization. On the other hand, it is true that there may be code that is not checking the error. The documentation, however, made it clear that this could happen:

https://github.com/RIOT-OS/RIOT/blob/66e0d4803773f7a60d8d17f42d089eadab5ff2e0/sys/include/crypto/ciphers.h#L127-L133

In conclusion, I'm _for_ this change.

mtausig:
Ping

mtausig:
@jcarrano Done.

bergzand:
> @bergzand Do you have any concern?

Nope, concerns have been addressed. Thanks for picking up the reviewing here!

Issue #10259, opened 2018-10-25
tests/leds doesn't work, on ESP32 Thing


snej:
#### Description
I'm trying to get RIOT-OS to do anything on my ESP32 Thing board. I can't get any terminal I/O from the tutorial apps, so I dropped down to a lower level LED test.

When I try to run `tests/leds`, the build and flashing appear to succeed, but the [blue LED](https://cdn.sparkfun.com/assets/learn_tutorials/5/0/7/top-annotated.jpg) (attached to pin 5) just stays on without blinking.

#### Steps to reproduce the issue
1. Check out latest master branch of RIOT
2. Install ESP32 toolchain and ESP-IDF and set environment vars accordingly
3. Connect ESP32 Thing to USB port
4. `cd tests/leds`
5. `BOARD=esp32-wroom-32 PORT=/dev/tty.usbserial-DN041OIM make all flash term`

#### Expected results
Blue LED blinks

#### Actual results
Blue LED stays steady

Here's the build output, following the compile/link stages:
```
...
[...]

PR #10273, opened 2018-10-26
cpu/esp32: allow explicit ESP32 crystal freq configuration

### LINKED ISSUE #10272: Terminal I/O broken on ESP32 boards with 26MHz crystal
#### Description
On some ESP32 dev boards [like mine], USB terminal I/O fails — all you see is garbage mixed with some of the ROM bootloader logging, but nothing from RIOT or the app itself.

This happens on boards with a 26MHz (not 40MHz) main clock crystal. The ESP docs say that "some" systems have this, and the SparkFun ESP32 Thing appears to be one. If this isn't compensated for at boot time, the UART timing will be wrong and terminal I/O will fail, resulting in nothing but garbage. IDF has a [config setting](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/kconfig.html#config-esp32-xtal-freq-sel) to specify 26MHz, which has to be set manually. RIOT apparently needs something similar.

(See [this note](https://docs.espressif.com/projects/esp-idf/en/latest/get-started-cmake/index.html#monitor) in the IDF getting-started docs, describing the symptoms and workaround.)

I have a fix and will be submitting a PR shortly...

1. Connect an ESP32 board with 26MHz crystal
2. Build & run any RIOT test/example app, such as test/leds

#### Expected results
RIOT logs appear in the terminal

#### Actual results
Terminal output is mostly garbage with some output from the ROM bootloader but nothing from RIOT.

#### Versions
* Building on: macOS 10.13.6
* Toolchain: xtensa-esp32-elf-gcc-5.2.0 (crosstool-NG crosstool-ng-1.22.0-80-g6c4433a) 5.2.0
* RIOT: Latest master commit 25-Oct-18 (99ae495)
* Hardware: SparkFun ESP32 Thing

snej:
### Contribution description

Some ESP32 boards (like my SparkFun ESP32 Thing) have a main clock
crystal that runs at 26MHz, not 40MHz. RIOT appears to assume 40MHz.
The mismatch causes the UART to not sync properly, resulting in
garbage written to the terminal instead of log output.

I’ve added:

* A new configuration constant CONFIG_ESP32_XTAL_FREQ that defaults
  to 0, but can be overridden by a board def or at build time to
  force a specific value (i.e. 26).
* Some code spliced into system_clk_init() to check this constant and
  call rtc_clk_init() to set the correct frequency. This code is
  mostly copied verbatim out of ESP-IDF; it’s a hack that works for
  me, but it **should definitely be cleaned up before merging**,
  by someone who knows what they’re doing!

### Testing procedure

See the "steps to reproduce" in #10272. I ran the `test/leds` app on my ESP32 Thing after making this change, prepending `CFLAGS="-DCONFIG_ESP32_XTAL_FREQ=26"` to the `make` command; all the expected RIOT log messages appeared, and the LED still blinks.

I do not have access to any other ESP32 boards to test with.
The changes are exclusively in ESP32 code, so no other CPUs are affected.

### Issues/PRs references

Fixes #10272

### DIFF
--- a/cpu/esp32/include/sdk_conf.h
+++ b/cpu/esp32/include/sdk_conf.h
@@ -66,6 +66,16 @@ extern "C" {
 #define CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ       80
 #endif
 
+/**
+ * ESP32 specific configuration
+ *
+ * Main clock crystal frequency (MHz). Zero means to auto-configure.
+ * This is configured at the board level, defaulting to 40.
+ */
+#ifndef CONFIG_ESP32_XTAL_FREQ
+#define CONFIG_ESP32_XTAL_FREQ                  ESP32_XTAL_FREQ
+#endif
+
 #define CONFIG_ESP32_RTC_XTAL_BOOTSTRAP_CYCLES  100
 #define CONFIG_ESP32_RTC_CLK_CAL_CYCLES         1024
 
--- a/cpu/esp32/startup.c
+++ b/cpu/esp32/startup.c
@@ -14,6 +14,7 @@
  * @brief       Implementation of the CPU initialization
  *
  * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @author      Jens Alfke <jens@mooseyard.com>
  * @}
  */
 
@@ -88,6 +89,7 @@ extern void esp_panic_wdt_stop (void);
 extern void spi_ram_init(void);
 extern void spi_ram_heap_init(void);
 extern uint32_t hwrand (void);
+extern void bootloader_clock_configure(void);
 
 /* forward declarations */
 static void system_init(void);
@@ -193,6 +195,12 @@ static void IRAM system_clk_init (void)
     /* first initialize RTC with default configuration */
     rtc_config_t rtc_cfg = RTC_CONFIG_DEFAULT();
     rtc_init_module(rtc_cfg);
+
+    /* configure main crystal frequency if necessary */
+    if (CONFIG_ESP32_XTAL_FREQ != RTC_XTAL_FREQ_AUTO
+            && CONFIG_ESP32_XTAL_FREQ != rtc_clk_xtal_freq_get()) {
+        bootloader_clock_configure();
+    }
 
     /* set FAST_CLK to internal low power clock of 8 MHz */
     rtc_clk_fast_freq_set(RTC_FAST_FREQ_8M);
--- a/cpu/esp32/vendor/esp-idf/esp_funcs.c
+++ b/cpu/esp32/vendor/esp-idf/esp_funcs.c
@@ -545,3 +545,48 @@ void sc_ack_send_stop(void)
     NOT_SUPPORTED();
 }
 #endif
+
+/*
+ * source: /path/to/esp-idf/components/bootloader_support/src/bootloader_clock.c
+ */
+void bootloader_clock_configure(void)
+{
+    // ROM bootloader may have put a lot of text into UART0 FIFO.
+    // Wait for it to be printed.
+    // This is not needed on power on reset, when ROM bootloader is running at
+    // 40 MHz. But in case of TG WDT reset, CPU may still be running at >80 MHZ,
+    // and will be done with the bootloader much earlier than UART FIFO is empty.
+    uart_tx_wait_idle(0);
+
+    /* Set CPU to 80MHz. Keep other clocks unmodified. */
+    rtc_cpu_freq_t cpu_freq = RTC_CPU_FREQ_80M;
+#ifndef RIOT_VERSION
+    /* On ESP32 rev 0, switching to 80MHz if clock was previously set to
+     * 240 MHz may cause the chip to lock up (see section 3.5 of the errata
+     * document). For rev. 0, switch to 240 instead if it was chosen in
+     * menuconfig.
+     */
+    uint32_t chip_ver_reg = REG_READ(EFUSE_BLK0_RDATA3_REG);
+    if ((chip_ver_reg & EFUSE_RD_CHIP_VER_REV1_M) == 0 &&
+            CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ == 240) {
+        cpu_freq = RTC_CPU_FREQ_240M;
+    }
+#endif
+
+    rtc_clk_config_t clk_cfg = RTC_CLK_CONFIG_DEFAULT();
+    clk_cfg.xtal_freq = CONFIG_ESP32_XTAL_FREQ;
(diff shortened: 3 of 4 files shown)


aabadie:
IIUC, this PR also fixes #10259

gschorcht:
@aabadie 

> IIUC, this PR also fixes #10259

Not really. The LED problem described in #10259 was not a bug but caused by missing LED pin configuration. The garbage output was just an observation during the investigation of the LED problem.

cpu/esp32/startup.c
```diff
@@ -188,12 +188,67 @@ NORETURN void IRAM call_start_cpu0 (void)
 extern uint32_t esp_clk_slowclk_cal_get(void);
 extern void IRAM_ATTR rtc_select_slow_clk(rtc_slow_freq_t slow_clk);
 
+
+// <<<< BEGIN PROVISIONAL CODE
+// Note: Copied verbatim from IDF's bootloader_clock.c except where noted.
+// I have no idea whether setting the CPU frequeny to 80MHz is needed, but it
+// looks like rtc_clk_init() needs _some_ value to set it to. --snej
+static void bootloader_clock_configure(void)
```
gschorcht: Please move this function either to file ```cpu/esp32/vendor/esp-idf/esp_funcs.c``` which is a collection of single functions that were extracted from different ESP-IDF files or create a new subdirectory ```cpu/esp32/vendor/esp-idf/bootloader_support``` and place the original file with your modifications there for the following reasons:

- The function is extracted from ESP-IDF as it is.
- since all files in ```cpu/esp32/vendor/``` are vendor files, it is not necessary to fulfill RIOT's coding style where ```//``` comments are not allowed. Furthermore, it becomes clear that the code was extracted from ESP-IDF.
- I prefere to keep ```startup.c``` small and not to overload it with supporting function code.

cpu/esp32/startup.c
```diff
@@ -188,12 +188,67 @@ NORETURN void IRAM call_start_cpu0 (void)
...
+    // ROM bootloader may have put a lot of text into UART0 FIFO.
+    // Wait for it to be printed.
+    // This is not needed on power on reset, when ROM bootloader is running at
+    // 40 MHz. But in case of TG WDT reset, CPU may still be running at >80 MHZ,
+    // and will be done with the bootloader much earlier than UART FIFO is empty.
+    uart_tx_wait_idle(0);
+
+    /* Set CPU to 80MHz. Keep other clocks unmodified. */
+    rtc_cpu_freq_t cpu_freq = RTC_CPU_FREQ_80M;
+#if 0 // `EFUSE_*` constants are not available in current RIOT sources --snej
```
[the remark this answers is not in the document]
gschorcht: It might be better to use
```
#ifndef RIOT_VERSION
...
#endif 
```
This makes the changes made for RIOT more clear and would make the comment unnecessary.

cpu/esp32/startup.c
snej: Will do.

cpu/esp32/include/sdk_conf.h
```diff
@@ -66,6 +66,18 @@ extern "C" {
...
+/**
+ * ESP32 specific configuration
+ *
+ * Main clock crystal frequency (MHz). Zero means to auto-configure.
+ * (Most board are 40, but some, like the SparkFun ESP32 Thing, are 26.)
+ * An obvious side effect of a mismatch is that the UART won't sync and the
+ * development terminal will show garbage instead of log output.
+ */
+#ifndef CONFIG_ESP32_XTAL_FREQ
+#define CONFIG_ESP32_XTAL_FREQ                  0
```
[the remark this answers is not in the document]
snej: Never mind, found it -- boards/common/esp32/include/board_common.h (right?)

snej:
I've pushed an updated commit with the suggested fixes, and verified that it works for me.

jia200x:
@gschorcht @snej @aabadie this bugfix is important IMO for the Release. Any news here? ;)

gschorcht:
sorry, my fault.

kYc0o:
So, @snej can you adapt your commit message?

cpu/esp32/vendor/esp-idf/esp_funcs.c
```diff
@@ -545,3 +545,48 @@ void sc_ack_send_stop(void)
...
+/*
+ * source: /path/to/esp-idf/components/bootloader_support/src/bootloader_clock.c
+ */
+void bootloader_clock_configure(void)
+{
+    // ROM bootloader may have put a lot of text into UART0 FIFO.
+    // Wait for it to be printed.
+    // This is not needed on power on reset, when ROM bootloader is running at
+    // 40 MHz. But in case of TG WDT reset, CPU may still be running at >80 MHZ,
+    // and will be done with the bootloader much earlier than UART FIFO is empty.
```
kYc0o: Just a side comment here, for further reference: `//` comments are not permitted according to our [coding conventions](https://github.com/RIOT-OS/RIOT/wiki/Coding-conventions).

It is not necessary to change it here because it's part of the vendor sources, but for further contributions on the RIOT code you should take it into account.

gschorcht: @kYc0o 
> Just a side comment here, for further reference: // comments are not permitted according to our coding conventions.

I saw it too when I reviewed the code. But since it is original code from vendor and placed in the vendor subdirectory, I told @snef that it is not necessary to adopt the code to fulfill RIOT's coding style. It makes no sense to change vendor code.

snej:
I've updated the commit title to match what @gschorcht changed the issue title to; is that enough? (Sorry, I don't know what your guidelines for commit messages are.)

jia200x:
> I've updated the commit title to match what @gschorcht changed the issue title to; is that enough? (Sorry, I don't know what your guidelines for commit messages are.)

It looks fine to me. @kYc0o are you OK with this PR? IMO it's relevant to have this PR into the Release.

kYc0o:
>  IMO it's relevant to have this PR into the Release.

Is this a bugfix?

gschorcht:
No, supported ESP32 boards are using 40 MHz crystal and work without this PR. You need the PR only if you want to use RIOT with an unsupported board that uses 26 MHz like the Sparkfun ESP32 Thing. Without having done a deep search, I would guess there is only a handful of such boards. So I think it is an enhancement.

kYc0o:
Great! Then let's go :-)

jia200x:
> No, supported ESP32 boards are using 40 MHz crystal and work without this PR. You need the PR only if you want to use RIOT with an unsupported board that uses 26 MHz like the Sparkfun ESP32 Thing. Without having done a deep search, I would guess there is only a handful of such boards. So I think it is an enhancement.

Ok, thanks for the info ;)

PR #10277, opened 2018-10-28
boards/common/nrf51: add common place for clock configuration

aabadie:
### Contribution description

This PR follows #10243 by adding a common place for clock configurations.

### Testing procedure

Building any application for each concerned board (airfy-beacon, calliope-mini, microbit, nrf51dongle, nrf5310, yunjia-nrf51822) should work.

### Issues/PRs references

depends on #10243

### DIFF
--- a/boards/yunjia-nrf51822/include/periph_conf.h
+++ b/boards/yunjia-nrf51822/include/periph_conf.h
@@ -21,25 +21,11 @@
 
 #include "periph_cpu.h"
 #include "periph_conf_common.h"
+#include "cfg_clock_16_0.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
-
-/**
- * @name    Clock configuration
- *
- * @note    The radio will not work with the internal RC oscillator!
- *
- * @{
- */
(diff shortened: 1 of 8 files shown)


aabadie:
@haukepetersen, are you interested by this one ?

haukepetersen:
@aabadie yes, I am :-)

would you mind to rebase (since I just merged #10243)? Then this one should be quite easy to review and ack...

aabadie:
Thanks @haukepetersen. I just rebased this one.

boards/yunjia-nrf51822/include/periph_conf.h
```diff
@@ -21,26 +21,12 @@
 
 #include "periph_cpu.h"
 #include "periph_conf_common.h"
+#include "cfg_clock_16_1.h"
```
haukepetersen: The original config did not enable the LFCLK. I am actually not sure if there is a lf crystal on that board. By looking at it, I can't find one. So I'd say keep it save and better disable it?!

aabadie:
@haukepetersen, all green there, do you ACK ?

haukepetersen:
and go.

PR #10280, opened 2018-10-29
boards/common/silabs: add Arduino API support for SiLabs boards

fedepell:
### Contribution description
Adds the Arduino API support for the Silicon Labs evaluation boards (ie. EFM32 STK3700 (Giant Gecko Starter Kit).
This enables Arduino-like programming supported by RIOT also on this boards.

### Testing procedure
Arduino automatic test sys_arduino can be used for basic functionalities tests, for example on the STK3700:
```
cd tests/sys_arduino/
BOARD=stk3700 make all flash test
```
Additionally one can manually use and play with the _examples/arduino_hello-world/_ for other tests (ie. playing with I/O and leds)

### DIFF
--- a/boards/common/silabs/include/arduino_board.h
+++ b/boards/common/silabs/include/arduino_board.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C)  2018 Federico Pellegrin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     boards_common_silabs
+ * @brief       SiLabs Boards configuration for the Arduino API
+ * @file
+ * @author      Federico Pellegrin <fede@evolware.org>
+ * @{
+ */
+
+#ifndef ARDUINO_BOARD_H
+#define ARDUINO_BOARD_H
+
+#include "arduino_pinmap.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   Arduino's digital pins mappings
+ */
+static const gpio_t arduino_pinmap[] = {
+    ARDUINO_PIN_0,
+    ARDUINO_PIN_1,
+    ARDUINO_PIN_2,
+    ARDUINO_PIN_3
+};
+
+/**
+ * @brief   Arduino's analog pins mappings
+ */
+static const adc_t arduino_analog_map[] = {
+    ARDUINO_A0
+};
+
+/**
+ * @brief   On-board LED mapping
+ */
+#define ARDUINO_LED         (0)
+
+/**
+ * @brief   On-board serial port mapping
+ */
+#define ARDUINO_UART_DEV         UART_DEV(0)
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* ARDUINO_BOARD_H */
+/** @} */
--- a/boards/common/silabs/Makefile.features
+++ b/boards/common/silabs/Makefile.features
@@ -1,0 +1,2 @@
+# Various other features (if any)
+FEATURES_PROVIDED += arduino
--- a/boards/common/silabs/include/arduino_pinmap.h
+++ b/boards/common/silabs/include/arduino_pinmap.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C)  2018 Federico Pellegrin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     boards_common_silabs
+ * @brief       SiLabs Boards configuration for the Arduino API
+ * @file
+ * @author      Federico Pellegrin <fede@evolware.org>
+ * @{
+ */
+
+#ifndef ARDUINO_PINMAP_H
+#define ARDUINO_PINMAP_H
+
+#include "board.h"
+#include "periph_cpu.h"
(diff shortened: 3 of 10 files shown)


basilfx:
This would be valid for multiple boards, so would this be something that can be generalized into `boards/common/silabs/include`?

All boards have LED0, LED1, PB0 and PB1.

fedepell:
@basilfx : many thanks for the heads up! I moved as suggested so it is more generic, squashed changed and updated descriptions.

boards/common/silabs/include/arduino_board.h
```diff
@@ -0,0 +1,60 @@
...
+#define ARDUINO_BOARD_H
+
+#include "periph_cpu.h"
+#include "periph/adc.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
```
basilfx: This must then be changed to:

```c
/**
 * @brief   Arduino's digital pins mappings
 */
static const gpio_t arduino_pinmap[] = {
    LED0_PIN,
    LED1_PIN,
    PB0_PIN,
    PB1_PIN
};
```

boards/common/silabs/include/arduino_board.h
```diff
@@ -0,0 +1,60 @@
...
+ * @brief   On-board LED mapping
+ */
+#define ARDUINO_LED         (0)
+
+/**
+ * @brief   On-board serial port mapping
+ */
+#define ARDUINO_UART_DEV         UART_DEV(0)
+
```
basilfx: One too many newline :-)

boards/common/silabs/include/arduino_board.h
fedepell: Agreed, much better! Thanks! Done!

boards/common/silabs/include/arduino_board.h
fedepell: Done!

fedepell:
@basilfx : should I squash the two in one commit or can be fine like this?

basilfx:
I will test this on several boards soon. So far, it looks good!

fedepell:
@basilfx : I rewrote a bit the structure of the header files since I wanted to make them (after comments on #10314 ) more similar to other Arduino boards definitions. So i extracted the purely pin definitions to a separate arduino_pinmap.h and left the remainder in arduino_board.h.
Hope it is better / more streamlined now!

fedepell:
@leandrolanzieri : Thanks for noting this out! I added the common Makefile.feature and included from all the files. 
I believe also some other features could be centralized, but I think it's better to do a separate PR as that would be a clean-up unrelated to the Arduino feature itself, do you agree?
Hope the changes are fine!

fedepell:
I blacklisted some tests for Arduino since they don't have the required hardware / pins to run it.

fedepell:
And squashed now :)

leandrolanzieri:
> I blacklisted some tests for Arduino since they don't have the required hardware / pins to run it.
 
Would it be possible to define more pins? That way the silabs board would need to be blacklisted from that driver test.

fedepell:
> > I blacklisted some tests for Arduino since they don't have the required hardware / pins to run it.
> 
> Would it be possible to define more pins? That way the silabs board would need to be blacklisted from that driver test.

Hmm not sure I understand, I mean: what would we match (#define) that pins on as the hardware (the display) is not actually there?

leandrolanzieri:
> > > I blacklisted some tests for Arduino since they don't have the required hardware / pins to run it.
> > 
> > 
> > Would it be possible to define more pins? That way the silabs board would need to be blacklisted from that driver test.
> 
> Hmm not sure I understand, I mean: what would we match (#define) that pins on as the hardware (the display) is not actually there?

Sorry for the delay. I meant that if there is a possibility to define more GPIOs in the `gpio_t arduino_pinmap[]`. The blacklist in the LCD example is there just to avoid trying to compile it in boards that don't have enough GPIOs (as a matter of fact, the stm32f4discovery no longer belongs there as it now has 8 arduino pins defined).

fedepell:
> > > > I blacklisted some tests for Arduino since they don't have the required hardware / pins to run it.
> > > 
> > > 
> > > Would it be possible to define more pins? That way the silabs board would need to be blacklisted from that driver test.
> > 
> > 
> > Hmm not sure I understand, I mean: what would we match (#define) that pins on as the hardware (the display) is not actually there?
> 
> Sorry for the delay. I meant that if there is a possibility to define more GPIOs in the `gpio_t arduino_pinmap[]`. The blacklist in the LCD example is there just to avoid trying to compile it in boards that don't have enough GPIOs (as a matter of fact, the stm32f4discovery no longer belongs there as it now has 8 arduino pins defined).

My only doubt is that these extra pin definitions would not be common to all boards. While the currently defined pins are common for all of the boards (PB0/PB1/LED0/LED1) I cannot find another set of 4 pins that have a common mapping/usage between all the boards. For example for the STK3x00 I could use PC8-PC11 but that's not goot for SLSTK ones. I struggle to find 4 that would be equivalent (and not already mapped to some better use)
That's why I have doubts on adding pins in this "common" if they are not really common for all of them. 

I believe we could eventually better add on the specific boards more pins. (so ie. for the STK I mention the PC8-10, for others find other available pins)

leandrolanzieri:
> I struggle to find 4 that would be equivalent (and not already mapped to some better use)
> That's why I have doubts on adding pins in this "common" if they are not really common for all of them.

That's fine, just wanted to know if there was a chance to define that for all of them. So from my side I tested on the slstk3402a board. @basilfx did you have a chance to test it on other boards?

leandrolanzieri:
@basilfx any news? Do you think we can go?

PR #10291, opened 2018-10-29
at86rf2xx: framebuffer idle_state

Josar:
The `idle_state` is save when starting to transmitt. But this is not obvious and not intuitiv to understand.
Setting the idle state when changing the state is more intuitiv and should result in the same behaviour.

https://github.com/RIOT-OS/RIOT/pull/9509#issuecomment-433866754


gebart:
Did you experience any issues without this PR?
What happens when recv is called multiple times (without dropping the frame buffer)?
From just looking at the code it seems like PLL_ON would be written to dev->idle_state on the second invocation, causing the transceiver to remain in PLL_ON after the frame has been read out instead of going back to RX_AACK_ON. I did not have time to test this PR yet.

Josar:
@gebart you are right. I think this PR is wrong. I was confused when reviewing https://github.com/RIOT-OS/RIOT/pull/9509 . As it is hard to understand when the idle state is saved. This should be better documented, imho.

PR #10292, opened 2018-10-29
pkg/baselibc: [WIP] a light-weight C library

jcarrano:
### Contribution description

I stole the C library from MyNewt, see https://mynewt.apache.org/latest/os/modules/baselibc.html . This is still a WIP, I'm publishing early to see if there is interest.

__33% reduction in code size for hello world__

Baselibc is derived from klibc (part of the linux kernel initialization system) with the GPL stuff removed. __It does not replace newlib__. Rather, it replaces several of it's functions by simpler, smaller and probably less efficient versions.

`printf` is based on tinyprintf, so many things won't be supported. The version in mynewt is slightly different from the one used here, so I might try pulling in those changes (their version has optional float support).

The interesting thing is that `stdio-uart` would need only small modifications to work both with and without this library (compare the stdio_read2, stdio_write2 functions I placed in main.c with the ones in stdio-uart).

### Testing procedure

I modified hello-world so that it uses this package:

1. Compile hello world (`BOARD=samr21-xpro WERROR=0 make`
2. Get the binary (`arm-none-eabi-objcopy -O binary bin/samr21-xpro/hello-world.elf new.bin`)
3. In the makefile, comment the line that says `USEPKG+=baselibc`.
4. Repeat (1) and (2)- use `old.bin` or another filename.
5. Compare sizes:

```sh
$ ls -l *.bin
-rwxr-xr-x 1 jcarrano jcarrano 5724 oct 29 17:19 new.bin
-rwxr-xr-x 1 jcarrano jcarrano 8572 oct 29 17:18 old.bin
```

### Related PRs

To integrate this properly, #9258 would be really helpful.

### DIFF
--- a/examples/hello-world/Makefile
+++ b/examples/hello-world/Makefile
@@ -12,6 +12,9 @@ RIOTBASE ?= $(CURDIR)/../..
 # development process:
 DEVELHELP ?= 1
 
+
+USEPKG+=baselibc
+
 # Change this to 0 show compiler invocation lines by default:
 QUIET ?= 1
 
--- a/examples/hello-world/main.c
+++ b/examples/hello-world/main.c
@@ -21,6 +21,39 @@
 
 #include <stdio.h>
 
+#if MODULE_BASELIBC
+
+#include "periph/uart.h"
+#include "stdio_uart.h"
+#include "isrpipe.h"
+
+extern isrpipe_t stdio_uart_isrpipe;
+
+size_t stdio_read2(FILE* dummy, char* buffer, size_t count)
+{
+    (void) dummy;
+
+    return (ssize_t)isrpipe_read(&stdio_uart_isrpipe, buffer, count);
+}
+
+size_t stdio_write2(FILE* dummy, const char* buffer, size_t len)
+{
+    (void) dummy;
+    uart_write(STDIO_UART_DEV, (const uint8_t *)buffer, len);
+
+    return len;
+}
+
+const struct File_methods _uart_methods = {
+    stdio_write2, stdio_read2
+};
+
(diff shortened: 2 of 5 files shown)


kaspar030:
Yup, the libc has quite some optimization potential. Even better, it is much easier to use the same libc on all platforms, if the C library is a little more approachable.

I experimented with actually [replacing](https://github.com/kaspar030/RIOT/tree/add_libc) newlib, as most RIOT applications use only a super small subset of libc.

jcarrano:
The nice thing about this one is that you don't need to replace anything (in fact, it _needs_ the toolchain-supplied library to provide stdint.h, stddef.h, etc). Filesystem-related stuff is missing, for example, though most apps won't even use that. I thing it gives a good "bang for the buck" in terms of benefits vs complexity.

I would't expect everything to work 100% the same with this libc (especially code that uses string formatting) but enabling and disabling it is just a matter of enabling or disabling a package.

I don't like the all the extra indirection in putchar, where it could consist of just sending a char to the UART:

```c
__extern_inline size_t fwrite(const void *buf, size_t size, size_t nmemb, FILE *stream)
{
    if (stream->vmt->write == NULL) return 0;
    return stream->vmt->write(stream, (char*)buf, size*nmemb) / size;
}

__extern_inline int fputc(int c, FILE *f)
{
	unsigned char ch = c;
	return fwrite(&ch, 1, 1, f) == 1 ? ch : EOF;
}

#define putchar(c) fputc((c),stdout)
```

The next step (after properly implementing the UART callbacks) is copying [MyNewt's version](https://github.com/apache/mynewt-core/tree/master/libc/baselibc) of this lib and diffing it to see what they changed.

PR #10293, opened 2018-10-29
example/nanocoap: fix block2 payload

kb2ma:
### Contribution description
The recent addition of Block2 support in nanocoap included a new resource for the nanocoap_server example, /riot/ver. This resource provides a description of the version of RIOT as well as the board and MCU. However, the construction of the payload includes the null terminator for the component strings. This PR reworks the construction to avoid addition of the \0 bytes.

### Testing procedure
To reproduce, send a GET request to the /riot/ver resource. Your implementation may require you to set a block size for the request. You also can use the attached Python script [task04_py.txt](https://github.com/RIOT-OS/RIOT/files/2526497/task04_py.txt) (renamed to satisfy GitHub), which requires a recent copy of [aiocoap](https://github.com/chrysn/aiocoap).

With aiocoap, the faulty response looks like:
```
Result: 2.05 Content
b'This is RIOT (Version: \x002018.04-devel-3293-g289e6\x00) running on a \x00native\x00 board with a \x00native\x00 MCU.'
```

The fixed response looks like:
```
Result: 2.05 Content
b'This is RIOT (Version: 2018.04-devel-3293-g289e6) running on a native board with a native MCU.'
```
### Issues/PRs references
N/A

### DIFF
--- a/examples/nanocoap_server/coap_handler.c
+++ b/examples/nanocoap_server/coap_handler.c
@@ -42,13 +42,13 @@ static ssize_t _riot_block2_handler(coap_pkt_t *pkt, uint8_t *buf, size_t len, v
     *bufpos++ = 0xff;
 
     /* Add actual content */
-    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, block2_intro, sizeof(block2_intro));
-    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, (uint8_t*)RIOT_VERSION, sizeof(RIOT_VERSION));
+    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, block2_intro, sizeof(block2_intro)-1);
+    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, (uint8_t*)RIOT_VERSION, strlen(RIOT_VERSION));
     bufpos += coap_blockwise_put_char(&slicer, bufpos, ')');
-    bufpos += coap_blockwise_put_bytes(&slicer, bufpos, block2_board, sizeof(block2_board));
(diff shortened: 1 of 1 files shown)


kb2ma:
@jia200x, do you agree that this fix is worth a backport to 2018.10?

jia200x:
@kb2ma I agree. Added the label.

kb2ma:
Thanks, @jia200x! Backport in #10298.

kaspar030:
Can the compile optimize the strlen call to a constant? If not, every time the block2 handler is called, strlen would actually count the langth. Not entirely bad, but also unnecessary.

If this doesn't get optimized, would ```sizeof(RIOT_FOO)-1``` have the same effect?

kb2ma:
> If this doesn't get optimized, would sizeof(RIOT_FOO)-1 have the same effect?

Yeah, that's the way I wrote it originally. But I noticed that _riot_board_handler() uses strlen(). I didn't think the optimization was worth it for an example. In terms of flash size, we probably wouldn't spell out 'MCU' in the real world either. ;-)

jia200x:
> Yeah, that's the way I wrote it originally. But I noticed that _riot_board_handler() uses strlen(). I didn't think the optimization was worth it for an example. In terms of flash size, we probably wouldn't spell out 'MCU' in the real world either. ;-)

Then IMO let's keep the `strlen` variant ;)

jia200x:
there seems to be an unrelated bug with `bench_runtime_coreapis` in Murdock...

kaspar030:
> In terms of flash size, we probably wouldn't spell out 'MCU' in the real world either. ;-)

:smile:

kaspar030:
> do you have any additional comments?

nope, let's go.

jia200x:
thank you!

PR #9740, opened 2018-08-08
drivers/at: fix invalid function pointer cast.

jcarrano:
### Contribution description

Fix an invalid cast of a function pointer. ARM GCC 8.2.0 (on Arch linux) was (rightly) complaining about this.

As a note, I'd regard any explicit cast of function pointers suspicious. That's something that simply should not be done.

PR #10299, opened 2018-10-30
drivers/at: Fix UART callback function signature

leandrolanzieri:
### Contribution description
This PR fixes the UART RX callback function signature to match the required by the `uart_init` function, by defining an intermediate function for callback. This problem appears when using GCC 8 or above apparently.

### Testing procedure
Try to compile `tests/driver_at` application, it should success.

### Issues/PRs references
None


leandrolanzieri:
Sorry, already addressed in #9740

Issue #10276, opened 2018-10-27
periph_init() never called on the MSB-A2, leaving the SPI bus uninitialized


maribu:
#### Description
On boot up of the MSB-A2 `periph_init()` is never called. This leaves e.g. the SPI interface uninitalized, as `spi_init()` is never called.

#### Steps to reproduce the issue
Boot the MSB-A2 e.g. with the cc110x driver enabled.

#### Expected results
`periph_init()` gets called on boot up, which in turn calls `spi_init()`

#### Actual results
`periph_init()` is not called and so is - therefore - `spi_init()`. However, the cc110x driver calls `spi_init_pins()` before every SPI transfer to the transceiver - which is apparently enough to get the SPI interface working. But e.g. adding an `assert(0)` to `periph_init()` or  to `spi_init()` is an easy way to confirm that neither function is called.

I stumbled upon this issue while rewriting the `cc110x` driver. I'm using a different approach to prevent the SPI interface from being used while the CC110x transceiver is not yet ready, so constantly configuring the MISO-Pin as GPIO pin and as SPI pin is not needed.

#### How to fix
Actually, a quick and dirty would be this:

``` patch
diff --git a/boards/msba2/board_init.c b/boards/msba2/board_init.c
index a386a9984c..bcdfbb5120 100644
--- a/boards/msba2/board_init.c
+++ b/boards/msba2/board_init.c
@@ -26,6 +26,7 @@
 
 #include "board.h"
 #include "cpu.h"
+#include "periph/init.h"
 
 void bl_init_ports(void)
 {
@@ -41,6 +42,7 @@ void bl_init_ports(void)
 
     LED0_OFF;
     LED0_OFF;
+    periph_init();
 }
 
 void init_clks1(void)
```

But I'm not sure if this is the right way to do it. I'm not familiar with how RIOT initializes periphs, but it seems the LPC2387 is handled different. It seems other CPUs provide `cpu_init()`, which in turn calls `periph_init()`. But the LPC2387 does not provide `cpu_init()`. Maybe the start up process in RIOT has changed and the LPC2387 has not yet been updated to the new way?

PeterKietzmann:
> Maybe the start up process in RIOT has changed and the LPC2387 has not yet been updated to the new way

[...]

PR #10301, opened 2018-10-30
boards/msba2: Added call to periph_init

maribu:
### Contribution description
Upon start up `periph_init()` should be called and initializes periphery such as SPI or I2C interface. `periph_init()` is called by other devices at the end of `cpu_init()` - however the start up process of the LPC2387 in RIOT is different and does use `cpu_init()`. Instead, the MSB-A2 is initialized in `board_init.c`. This PR adds the missing call to `periph_init()` to board initialization routine. (Obviously, it would be better to update the start up code of the LPC2387 to follow the same approach other RIOT platforms use, but with the MSB-A2 being an legacy platform that is barely used, the effort would be to high.)

#### Verifying that there is indeed a problem and this is addressed
E.g. by adding `assert(0);` into `periph_init()`. The expected result would be that the device does not boot because of the failed assertion. However, before this PR `periph_init()` is not called, thus the device will boot. With the PR applied the assertion will fail.

#### Testing for regressions
E.g. by checking if the MSB-A2 still boots reliable with this PR applied.

### Issues/PRs references
Fixes https://github.com/RIOT-OS/RIOT/issues/10276

### DIFF
--- a/boards/msba2/board_init.c
+++ b/boards/msba2/board_init.c
@@ -26,6 +26,7 @@
 
 #include "board.h"
 #include "cpu.h"
+#include "periph/init.h"
 
 void bl_init_ports(void)
 {
@@ -41,6 +42,7 @@ void bl_init_ports(void)
 
     LED0_OFF;
     LED0_OFF;
+    periph_init();
 }
 
 void init_clks1(void)


PeterKietzmann:
@maribu you are working a lot with that platform. Wouldn't a unified CPI initialization procedure be super nice :-)? I don't really much like the board specific fix in this PR. However, I woudn't NACK if you don't wanna clean-up the CPU port. Will test again

maribu:
> @maribu you are working a lot with that platform. Wouldn't a unified CPI initialization procedure be super nice :-)?

Sure it would be super nice :-) I'm not too sure how much we will still use... Maybe just doing it to learn more about the details of RIOT would be a good reason. This also might also be a chance to replace the startup assembly code by C code with sparsely used inline-assembly were absolutely needed. I believe the LLVM toolchain does work with inline assembly (even though I'm really not an expert at all), so if I'm correct this would fix the build with the LLVM toolchain as well. I'll think about it.

> For some strange reason everything seems to work fine on current master but with your PR the startup gets stuck. Would you mind to have an other look?

Sure. I'll check it out.

maribu:
I created this PR before the auto_init fix of the sht1x module was merged. This is why it crashes on startup in `example/default`. A rebase against current master fixed the issue.

PR #10305, opened 2018-10-30
tests/driver_my9221: fix for loop indexes

jia200x:
### Contribution description
In the `my9221` test there are some for loops used for testing each led in the driver. However, the indexes should go always from 0 to 9 and the actual test generate these indexes from 1 to 10, so there's a failed assertion [here](https://github.com/RIOT-OS/RIOT/blob/68dc5b0d6e0422df9c3653c2cb8021fc35974aae/drivers/my9221/my9221.c#L133).

This PR fixes this issue.

### Testing procedure
Configure `drivers_my9221` test in its Makefile, and run `make flash test`. Verify there's no `FAILED ASSERTION` => `[SUCCESS]` message in the end.

### DIFF
--- a/tests/driver_my9221/main.c
+++ b/tests/driver_my9221/main.c
@@ -51,7 +51,8 @@ int main(void)
         my9221_set_led(&dev, i, MY9221_LED_OFF);
     }
     xtimer_usleep(TEST_WAIT);
-    for (unsigned i=dev.params.leds; i > 0 ; --i) {
+    for (unsigned j=dev.params.leds; j > 0 ; --j) {
+        unsigned i = j-1;
         my9221_set_led(&dev, i, MY9221_LED_ON);
         xtimer_usleep(TEST_WAIT);
         my9221_set_led(&dev, i, MY9221_LED_OFF);


tests/driver_my9221/main.c
```diff
@@ -45,16 +45,16 @@ int main(void)
     }
     /* run some tests */
     LOG_INFO("- light up all LEDs one by one.\n");
-    for (unsigned i=0; i < dev.params.leds; ++i) {
+    for (unsigned i=0; i < dev.params.leds; i++) {
```
cgundogan: this change is unrelated

tests/driver_my9221/main.c
```diff
@@ -45,16 +45,16 @@ int main(void)
...
         my9221_set_led(&dev, i, MY9221_LED_ON);
         xtimer_usleep(TEST_WAIT);
         my9221_set_led(&dev, i, MY9221_LED_OFF);
     }
     xtimer_usleep(TEST_WAIT);
-    for (unsigned i=dev.params.leds; i > 0 ; --i) {
-        my9221_set_led(&dev, i, MY9221_LED_ON);
+    for (unsigned i=dev.params.leds; i > 0 ; i--) {
```
cgundogan: also unrelated

cladmi: I would rather change the boundaries, because in the same function `i` was first a led number, then a led +1.
But as it is an unsigned so of `>= 0` cannot be used.

What is our normal way of handling these ?
Iterate on `j` and define `i = j - 1` as an intermediate variable ?

tests/driver_my9221/main.c
jia200x: you are right, late programming effects :)

tests/driver_my9221/main.c
jia200x: that's why I prefer the `while(i--)` pattern for reverse counts :)

But yeah, it's cleaner with an intermediate variable. I will push it.

jia200x:
Done. I squashed since the change is super small.

PR #9012, opened 2018-04-24
frdm-kl43z: NXP Kinetis FRDM-KL43Z development board

gebart:
### Contribution description

Add initial CPU support for NXP Kinetis KL43Z and an initial configuration for the FRDM-KL43Z development board.

This PR lays the foundation for the KL series support and will enable easily adding support for other Kinetis L based boards.

#### What works?

 - Clock configuration (MCG_Lite), 2 MHz, 8 MHz, 48 MHz, external clock
 - UART/LPUART
 - GPIO
 - I2C
 - Timers (RTC/RTT, PIT, LPTMR)
 - MMA8451 accelerometer
 - MAG3110 magnetometer

#### What is missing?

 - SPI (different hardware module than used in the other supported Kinetis CPUs)
 - PWM
 - Segment LCD driver

### Issues/PRs references

#7154

PR #10306, opened 2018-10-31
drivers/mma8x5x: remove useless type from initialization parameters

aabadie:
### Contribution description

This PR slightly rework the way variants of the mma8x5x (mma8451/2/3 and mma8652/3) can be used in RIOT: they are explicitly passed via USEMODULE as pseudomodules and preprocessor macros are used to determine the right device ID.

In the meantime, the microbit is activating its mma8653 when saul is loaded.

### Testing procedure

- Try the test driver on a board that provides it (microbit provides mma8653 variant):
```
make BOARD=microbit DRIVER=mma8653 -C tests/driver_mma8x5x
```
- Try default with the microbit
```
make BOARD=microbit -C examples/default
```

### Issues/PRs references

None

### DIFF
--- a/drivers/mma8x5x/mma8x5x.c
+++ b/drivers/mma8x5x/mma8x5x.c
@@ -51,11 +51,18 @@ int mma8x5x_init(mma8x5x_t *dev, const mma8x5x_params_t *params)
 
     /* test if the target device responds */
     i2c_read_reg(BUS, ADDR, MMA8X5X_WHO_AM_I, &reg, 0);
-    if (reg != dev->params.type) {
-        i2c_release(BUS);
-        DEBUG("[mma8x5x] init - error: invalid WHO_AM_I value [0x%02x]\n",
-               (int)reg);
-        return MMA8X5X_NODEV;
+    switch (reg) {
+        case MMA8X5X_TYPE_MMA8451:
+        case MMA8X5X_TYPE_MMA8452:
+        case MMA8X5X_TYPE_MMA8453:
+        case MMA8X5X_TYPE_MMA8652:
+        case MMA8X5X_TYPE_MMA8653:
+            break;
+        default: /* invalid device type */
+            i2c_release(BUS);
+            DEBUG("[mma8x5x] init - error: invalid WHO_AM_I value [0x%02x]\n",
+                  (int)reg);
+            return MMA8X5X_NODEV;
     }
 
     /* reset the device */
--- a/boards/microbit/Makefile.dep
+++ b/boards/microbit/Makefile.dep
@@ -5,6 +5,7 @@ endif
 
 ifneq (,$(filter saul_default,$(USEMODULE)))
   USEMODULE += saul_gpio
+  USEMODULE += mma8x5x
 endif
 
 ifneq (,$(filter gnrc_netdev_default netdev_default,$(USEMODULE)))
--- a/boards/microbit/include/board.h
+++ b/boards/microbit/include/board.h
@@ -69,7 +69,6 @@ extern "C" {
  */
 #define MMA8X5X_PARAM_I2C           I2C_DEV(0)
 #define MMA8X5X_PARAM_ADDR          0x1d
-#define MMA8X5X_PARAM_TYPE          0x5a
 /** @} */
 
 /**
--- a/drivers/include/mma8x5x.h
+++ b/drivers/include/mma8x5x.h
@@ -95,7 +95,6 @@ enum {
 typedef struct {
     i2c_t i2c;                  /**< I2C bus the device is connected to */
     uint8_t addr;               /**< I2C bus address of the device */
-    uint8_t type;               /**< device type */
     uint8_t rate;               /**< sampling rate to use */
     uint8_t range;              /**< scale range to use */
     uint8_t offset[3];          /**< data offset in X, Y, and Z direction */
--- a/drivers/mma8x5x/include/mma8x5x_params.h
+++ b/drivers/mma8x5x/include/mma8x5x_params.h
@@ -37,9 +37,6 @@ extern "C" {
 #ifndef MMA8X5X_PARAM_ADDR
 #define MMA8X5X_PARAM_ADDR      (MMA8X5X_I2C_ADDRESS)
 #endif
-#ifndef MMA8X5X_PARAM_TYPE
-#define MMA8X5X_PARAM_TYPE      (MMA8X5X_TYPE_MMA8652)
-#endif
 #ifndef MMA8X5X_PARAM_RATE
 #define MMA8X5X_PARAM_RATE      (MMA8X5X_RATE_200HZ)
 #endif
@@ -53,13 +50,12 @@ extern "C" {
 #ifndef MMA8X5X_PARAMS
 #define MMA8X5X_PARAMS          { .i2c    = MMA8X5X_PARAM_I2C,   \
                                   .addr   = MMA8X5X_PARAM_ADDR,  \
-                                  .type   = MMA8X5X_PARAM_TYPE,  \
                                   .rate   = MMA8X5X_PARAM_RATE,  \
                                   .range  = MMA8X5X_PARAM_RANGE, \
                                   .offset = MMA8X5X_PARAM_OFFSET }
 #endif
 #ifndef MMA8X5X_SAUL_INFO
-#define MMA8X5X_SAUL_INFO       { .name = "mma8652" }
+#define MMA8X5X_SAUL_INFO       { .name = "mma8x5x" }
 #endif
 /**@}*/
 
--- a/tests/driver_mma8x5x/Makefile
+++ b/tests/driver_mma8x5x/Makefile
@@ -1,6 +1,6 @@
 include ../Makefile.tests_common
 
-USEMODULE += mma8x5x
+USEMODULE += mma8x5x  # will work with all variants
 USEMODULE += xtimer
 
 include $(RIOTBASE)/Makefile.include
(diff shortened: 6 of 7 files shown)


drivers/mma8x5x/mma8x5x.c
```diff
@@ -51,7 +63,7 @@ int mma8x5x_init(mma8x5x_t *dev, const mma8x5x_params_t *params)
 
     /* test if the target device responds */
     i2c_read_reg(BUS, ADDR, MMA8X5X_WHO_AM_I, &reg, 0);
-    if (reg != dev->params.type) {
+    if (reg != MMA8X5X_TYPE) {
```
jcarrano: Instead if doint this comparison, and having pseudomodules to change the value of MMA8X5X_TYPE, would it be simple to check if `reg` is equal to any of `MMA8X5X_TYPE_MMA8???`. Especially since, in the rest of the module, there is no other code dependent on the device subtype.

jcarrano: What's the point of hard-coding the variant in the module's code? How does it improve anything? It is not possible any longer to have two variants in the same firmware.

Also, there's no code to check that conflicting pseudomodules are not enabled.

I'd argue that checking that the type matches a specific sub-type is close to useless. Even in the case where one has different subtypes connected and wants to differentiate them, it can be done by adding a method that returns the subtype.

aabadie: I think you are commenting on the outdated version. The new version allows both:
- if the firmware should be used with one (or several) sensors of the same type, then the application makefile can just pass the explicit variant using the corresponding pseudo-module. The driver will use the check for that specific version
- if the firmware should be used with multiple and different variant of the sensor, then no pseudo-module is used (just import the generic driver version) and a loose comparison is used (again 0x0a which is common to all variants).

aabadie: I pushed a new version that keep the type parameter but its default value is now defined based on the pseudomodule passed to the build system. Let me know what you think of it.

aabadie:
ping @jcarrano @kaspar030. I'm not sure the current version is what you were thinking of. Let me know

jcarrano:
You reverted to the old behaviour, where initialization will fail if the `type` register is not exactly equal to the type set at build time. To me, relaxing that check was the most valuable contribution of this PR, so much that I wondered why would anybody want to fix the type at compile time.

aabadie:
> where initialization will fail if the type register is not exactly equal to the type set at build time

it's more subtle than that. Load the explicit module will set the right type at build time and ensures that the device is correctly matching.
With @kaspar030 's use case, it's also possible to have multiple devices of different type and overload the MMA8X5X_PARAMS macro with the list of devices auto-initialized at init with the right type for each.

kaspar030:
> With @kaspar030 's use case, it's also possible to have multiple devices of different type and overload the MMA8X5X_PARAMS macro with the list of devices auto-initialized at init with the right type for each. In this case, loading mma8x5x in the Makefile will be enough.

Not sure I understand. Before, I had to recompile for different variants of the chip. Now, I have to recompile.

Why not just accept all MMA variants in all cases? They're all using only the same (limited) feature set. As soon as more is supported, that suport can be ifdef'ed on "type". At that point it makes sense to maybe use ```#ifdef MMA_VARIANT_foo``` to not compile in dead code, but not now.

aabadie:
I think we will understand things better with code ;)
1. First case, a single device used by the application. Just load the right pseudo-module and don't care about type parameter:
- `Makefile`:
```mk
USEMODULE += mma8563
```
- `main.c` (or whatever code in charge of initializing the device):
```c
#include "mma8x5x.h"
#include "mma8x5x_params.h"

static mma8x5x_t dev;

int main(void)
{
    ...
    mma8x5x_init(&dev, &mma8x5x_params[0])
    ...
    return;
}
```
2. Multiple devices, possibly different variants. As required by @kaspar030:
- `Makefile`:
```mk
USEMODULE += mma8x5x
```
- in main.c (or whatever code in charge of setting the intialization params of the device**s**):
```c
#include "mma8x5x.h"

#define MMA8X5X_PARAMS          { \
    { \
      .i2c    = I2C_DEV(0),   \
      .addr   = MMA8X5X_I2C_ADDRESS,  \
      .type   = MMA8X5X_TYPE_MMA8652,  \
      .rate   = MMA8X5X_RATE_200HZ,  \
      .range  = MMA8X5X_RANGE_2G, \
      .offset = { 0, 0, 0 } \
    }, \
    { \
      .i2c    = I2C_DEV(1),   \
      .addr   = MMA8X5X_I2C_ADDRESS,  \
      .type   = MMA8X5X_TYPE_MMA8653,  \
      .rate   = MMA8X5X_RATE_200HZ,  \
      .range  = MMA8X5X_RANGE_2G, \
      .offset = { 0, 0, 0 } \
    }, \
    { \
      .i2c    = I2C_DEV(0),   \
      .addr   = 0x42,  \
      .type   = MMA8X5X_TYPE_MMA8451,  \
      .rate   = MMA8X5X_RATE_200HZ,  \
      .range  = MMA8X5X_RANGE_2G, \
      .offset = { 0, 0, 0 } \
    }, \
} \

#include "mma8x5x_params.h"

static mma8x5x_t dev0; /* mma8652 */
static mma8x5x_t dev1; /* mma8653 */
static mma8x5x_t dev2; /* mma8451 */

int main(void)
{
    ...
    mma8x5x_init(&dev0, &mma8x5x_params[0])
    mma8x5x_init(&dev1, &mma8x5x_params[1])
    mma8x5x_init(&dev2, &mma8x5x_params[2])
    ...
    return;
}
```

kaspar030:
>     1. Multiple devices, possibly different variants. As required by @kaspar030:

I have one device, unknown variant.

Just accept all MMA variants, dammit! ;)

gebart:
Why do you even need the type parameter? There is functionally no difference between them in this implementation. Just check that the who am I register contains any of the known compatible variants to avoid problems when another device uses the same address.

aabadie:
> Why do you even need the type parameter?

That was my initial motivation. But [this comment](https://github.com/RIOT-OS/RIOT/pull/10306#discussion_r230025254) confused me a bit.
Anyway, I updated again this PR: the type parameter is gone again, the _whoami_ check accepts all types, pseudo-modules are still there. I hope this time I can squash.

aabadie:
> I hope this time I can squash.

Sorry, I squashed by accident...

drivers/mma8x5x/mma8x5x.c
```diff
@@ -51,7 +51,9 @@ int mma8x5x_init(mma8x5x_t *dev, const mma8x5x_params_t *params)
 
     /* test if the target device responds */
     i2c_read_reg(BUS, ADDR, MMA8X5X_WHO_AM_I, &reg, 0);
-    if (reg != dev->params.type) {
+    if ((reg != MMA8X5X_TYPE_MMA8451) && (reg != MMA8X5X_TYPE_MMA8452) &&
```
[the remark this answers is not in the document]
aabadie: Ok, I changed it to use the switch. Thanks

kaspar030:
pls squash! I'll give it a quick test run as soon as I find my mma breakout.

gebart:
@kaspar030 I can test this on the frdm KL43Z later today if you don't find your board.

aabadie:
squashed. This can also be tested an any micro:bit (There's one on IoT-LAB, but fixed)

gebart:
tests/driver_mma8x5x works on frdm-kl43z (#9012), MMA8451 sensor.
Though I think the DRIVER and PSEUDOMODULES are just an unnecessary complexity. IMO this change would be better with only the type `switch` block in init and dropping the `.type` member of the params.

aabadie:
> the DRIVER and PSEUDOMODULES are just an unnecessary complexity

+1 for the DRIVER, this is a left-over from my previous tries.
-1 for the PSEUDOMODULES: I think it gives a more explicit hint about what variant is provided by a board or required by an application.

kaspar030:
@aabadie I think this PR is going into the classical "too many changes, cannot make everyone happy" state.

If you cut out https://github.com/RIOT-OS/RIOT/pull/10306/commits/7e625bdf5f7e6fecad0a1ed53b912f63cd188d06 into it's own PR, I'd ACK that, and @gebart indicated that he'd prefer that for simplicty reasons, too.

gebart:
@aabadie sounds good to me

aabadie:
@kaspar030 @gebart @jcarrano, this PR is updated again and commits are squashed. If you are all happy with it, I think it can go.

aabadie:
if one wants to test this PR, just flash and run examples/default on `microbit-1.saclay.iot-lab.info`.

@jcarrano, are you ok with the lastest changes ?

jcarrano:
@aabadie I'm fine. Only one detail, the comment in the first commit is inaccurate when it refers to  the MODULE_MMA6X8X macro).

Unfortunately I cannot approve because I do not have the HW.

aabadie:
> Only one detail, the comment in the first commit is inaccurate when it refers to the MODULE_MMA6X8X macro).

Ok, I can update this.

> Unfortunately I cannot approve because I do not have the HW.

If you want to try, there is a microbit available on IoT-LAB (Saclay site). It's static so the accelerometer will always return the same values, but it should be enough to test the initialization of the driver.

aabadie:
@jcarrano, the message of commit a63d76b3465db35f34fa3f0bc50c58a223ef1aed have been updated.

Anyone to test this and ACK ? @kaspar030 ?

aabadie:
> I checked around and I see that the MMA is on the pba-kw pba-d-01-kw2x board

You won't believe me but I just found one in the old @kYc0o's office :)
Could be a good candidate to put in IoT-LAB, no ?

aabadie:
Ping. It's all green but I prefer that someone else than me merges this PR.

kaspar030:
Thanks @aabadie!

PR #10310, opened 2018-11-01
kinetis: Perform auto-trim of IRC when possible

jnohlgard:
### Contribution description

This will perform a auto-trim (or self calibration) of the internal reference oscillator during boot, if the selected clocking mode allows it.
The benefit is that this may improve the correctness of the frequency of the internal MCGIRCLK clock source which can be used to clock peripherals in low power modes, such as the LPUART.
Requires that the clocking mode is configured to one of the external
modes, otherwise there is no clock available to use as reference.

Tests with an oscilloscope shows that one of my frdm-kw41z boards has a factory trimmed MCGIRCLK which runs at 3.84 MHz, after the auto-trim has completed the clock runs at 3.97 MHz instead.

### Testing procedure

Use the frdm-kw41z board, which is default configured to use LPUART clocked from MCGIRCLK, change clocking mode in periph_conf.h to KINETIS_MCG_MODE_FBE, change CLOCK_CORECLOCK to CLOCK_RADIOXTAL (see diff below)

Test that normal shell I/O still works over UART. This means that the MCGIRCLK is running close to 4 MHz.

A more elaborate test involves checking with an oscilloscope the CLKOUT pin configured to output MCGIRCLK, however this is overkill for this small change.

```diff
diff --git a/boards/frdm-kw41z/include/periph_conf.h b/boards/frdm-kw41z/include/periph_conf.h
index 2330c1a386..253f85972c 100644
--- a/boards/frdm-kw41z/include/periph_conf.h
+++ b/boards/frdm-kw41z/include/periph_conf.h
@@ -54,7 +54,7 @@ static const clock_config_t clock_config = {
         0,
     /* Using FEI mode by default, the external crystal settings below are only
      * used if mode is changed to an external mode (PEE, FBE, or FEE) */
-    .default_mode       = KINETIS_MCG_MODE_FEI,
+    .default_mode       = KINETIS_MCG_MODE_FBE,
     /* The crystal connected to RSIM OSC is 32 MHz */
     .erc_range          = KINETIS_MCG_ERC_RANGE_VERY_HIGH,
     .osc_clc            = 0, /* no load cap configuration */
@@ -67,7 +67,7 @@ static const clock_config_t clock_config = {
 /* Radio xtal frequency, either 32 MHz or 26 MHz */
 #define CLOCK_RADIOXTAL              (32000000ul)
 /* CPU core clock, the MCG clock output frequency */
-#define CLOCK_CORECLOCK              (48000000ul)
+#define CLOCK_CORECLOCK              (CLOCK_RADIOXTAL)
 #define CLOCK_BUSCLOCK               (CLOCK_CORECLOCK / 2)
 #define CLOCK_MCGIRCLK               (4000000ul)
 /** @} */
```

The implementation follows the recipe in the CPU reference manual for computing the target constants.

### Issues/PRs references

### DIFF
--- a/cpu/kinetis/periph/mcg.c
+++ b/cpu/kinetis/periph/mcg.c
@@ -177,6 +177,66 @@ static void kinetis_mcg_init_mcgirclk(void)
 }
 
 /**
+ * @brief   Internal helper for performing the actual auto-trim of the reference oscillators
+ *
+ * This is called internally by @ref kinetis_mcg_auto_trim_irc
+ */
+static int kinetis_mcg_auto_trim_exec(unsigned fast_irc)
+{
+    uint32_t expected_count = (CLOCK_BUSCLOCK * 21u);
+    if (fast_irc) {
+        bit_set8(&MCG->SC, MCG_SC_ATMS_SHIFT);
+        /* Fast IRC is expected to run at 4 MHz */
+        /* The 64 bit multiplication will be optimized away and replaced by a constant */
+        expected_count = ((uint64_t)expected_count * 128) / 4000000ul;
+    }
+    else {
+        bit_clear8(&MCG->SC, MCG_SC_ATMS_SHIFT);
+        expected_count /= 32768;
+    }
+    if (expected_count > 0xffff) {
+        /* Something is very wrong with the configuration */
+        return -2;
+    }
+    /* Load expected counts */
+    MCG->ATCVL = expected_count & 0xff;
+    MCG->ATCVH = (expected_count >> 8) & 0xff;
+    /* initiate auto trim procedure */
+    bit_set8(&MCG->SC, MCG_SC_ATME_SHIFT);
+    /* ATME bit will de-assert automatically when the procedure completes */
+    while (MCG->SC & MCG_SC_ATME_MASK) {}
+    if (MCG->SC & MCG_SC_ATMF_MASK) {
+        /* auto trim failure */
+        return -1;
+    }
+    return 0;
+}
+
+/**
+ * @brief   Perform auto-trim of the internal reference oscillators
+ *
+ * This requires that an external crystal or clock is being used to drive the
+ * bus clock, i.e. the current clock mode must be one of FBE, FEE, BLPE, PBE, PEE
+ *
+ * The accuracy of the internal reference oscillators may be better after
+ * performing this calibration. This may be extra useful when clocking
+ * peripherals from the MCGIRCLK, e.g. LPUART.
+ */
+static void kinetis_mcg_auto_trim_irc(void)
+{
+    /* Try to auto-trim the internal oscillators */
+    /* Trim the 32768 Hz slow IRC */
+    int res = kinetis_mcg_auto_trim_exec(0);
+    if (res < 0) {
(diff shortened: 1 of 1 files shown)


cpu/kinetis/periph/mcg.c
```diff
@@ -176,6 +176,66 @@ static void kinetis_mcg_init_mcgirclk(void)
     }
 }
 
+/**
+ * @brief   Internal helper for performing the actual auto-trim of the reference oscillators
+ *
+ * This is called internally by @ref kinetis_mcg_auto_trim_irc
+ */
+static int kinetis_mcg_auto_trim_exec(unsigned fast_irc)
```
MrKevinWeiss: Are you missing a data type here?

cpu/kinetis/periph/mcg.c
```diff
@@ -176,6 +176,66 @@ static void kinetis_mcg_init_mcgirclk(void)
...
+{
+    uint32_t expected_count = (CLOCK_BUSCLOCK * 21u);
```
MrKevinWeiss: Does it make sense to change magic numbers to defines?  I am not too picky if you don't want to though.

cpu/kinetis/periph/mcg.c
```diff
@@ -176,6 +176,66 @@ static void kinetis_mcg_init_mcgirclk(void)
...
+        /* Something is very wrong with the configuration */
+        return -2;
+    }
+    /* Load expected counts */
+    MCG->ATCVL = expected_count & 0xff;
+    MCG->ATCVH = (expected_count >> 8) & 0xff;
+    /* initiate auto trim procedure */
+    bit_set8(&MCG->SC, MCG_SC_ATME_SHIFT);
+    /* ATME bit will de-assert automatically when the procedure completes */
+    while (MCG->SC & MCG_SC_ATME_MASK) {}
```
MrKevinWeiss: Is there a need for a timeout or will this always end or is it acceptable to potentially wait forever?

MrKevinWeiss:
@aabadie if you confirm with UART I would be satisfied with the testing!

cpu/kinetis/periph/mcg.c
gebart: I usually omit `int` in `unsigned int`, for brevity. Same goes for `long` vs. `long int` or `unsigned long` vs `unsigned long int`

cpu/kinetis/periph/mcg.c
gebart: I don't know if a macro is necessary, maybe adding a comment would explain things enough? The formula is copied straight from the reference manual, which is given without any further explanation what the 21 comes from:

image

aabadie:
I tested this PR and unfortunately, after changing `boards/frdm-kw41z/include/periph_conf.h` as described in the testing procedure, I only get garbage on the UART.

gebart:
Let's put this on hold for a while, I will not have time to debug and work with this PR in the near future.

MrKevinWeiss:
Alrighty, I should be able to get a board to test with a scope and maybe help out when you have time

benpicco:
@benemorius can you give this a try?

PR #10311, opened 2018-11-01
frdm-kw41z: Use 1280 multiplier for FEI

gebart:
### Contribution description

The reference manual recommends not using the DMX32 setting (1464
multiplier) when using the internal reference clock as the FLL clock
source (FEI, FBI modes).

> **25.5.2 Using a 32.768 kHz reference**
> ... snip ...
>In FBI and FEI modes, setting C4[DMX32] bit is not recommended. If the internal
reference is trimmed to a frequency above 32.768 kHz, the greater FLL multiplication
factor could potentially push the microcontroller system clock out of specification,
potentially resulting in unpredictable behavior.

This risk is likely very small because the slow IRC should be factory trimmed to 32768 Hz @ 25 C, but it may be up to +/- 3 % wrong in certain environmental conditions, according to the CPU data sheet.

### Testing procedure

Build and run some of the test applications, see that it runs and that the UART shell output works correctly.

### DIFF
--- a/boards/frdm-kw41z/include/periph_conf.h
+++ b/boards/frdm-kw41z/include/periph_conf.h
@@ -34,9 +34,9 @@ static const clock_config_t clock_config = {
     /*
      * This configuration results in the system running with the internal clock
      * with the following clock frequencies:
-     * Core:  48 MHz
-     * Bus:   24 MHz
-     * Flash: 24 MHz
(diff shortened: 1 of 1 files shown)


benemorius:
I remind myself to look at this soon. At a glance, I expect this is a correct change.

PR #10312, opened 2018-11-01
sys/stdio_uart: do not cast function pointer.

jcarrano:
### Contribution description

There was a function pointer cast in stdio_uart. The warning was being explicitly supressed in the module's makefile (see #9121).

This commit removes the suppression and fixes the warning by providing a wrapper function.

The wrapper was already defined in the AT driver (see #9740), as part of a fix for a similar bug there, so the definition was moved to the isrpipe module, the logic being that using isrpipe in conjunction with a periph_uart is a common enough use-case.

### Testing procedure

`tests/driver_at` should still compile, same as `examples/hello_world`.

### Issues/PRs references

Undoes #9121.

### DIFF
--- a/drivers/at/at.c
+++ b/drivers/at/at.c
@@ -22,17 +22,12 @@
 #define AT_PRINT_INCOMING (0)
 #endif
 
-static void _isrpipe_write_one_wrapper(void *_isrpipe, uint8_t data)
-{
-    isrpipe_write_one(_isrpipe, (char)data);
-}
-
 int at_dev_init(at_dev_t *dev, uart_t uart, uint32_t baudrate, char *buf, size_t bufsize)
 {
     dev->uart = uart;
     isrpipe_init(&dev->isrpipe, buf, bufsize);
 
-    return uart_init(uart, baudrate, _isrpipe_write_one_wrapper,
+    return uart_init(uart, baudrate, isrpipe_write_uartcb,
                      &dev->isrpipe);
 }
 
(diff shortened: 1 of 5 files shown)


bergzand:
@jcarrano Is there a reason for providing a wrapper instead of changing the `isrpipe_write_one` from `char c` to `uint8_t c`? From what I can find, with this PR, the only user of `isrpipe_write_one` is `ethos.c` and a cast from `char` to `uint8_t` can be easily added there.

jcarrano:
@bergzand 

1. I wanted to avoid an API change.
2. What makes those function pointers incompatible is not the arguments, but the return type (void vs int).

Regarding (2), it's usually innocuous in RISC architectures where the first results are returned in registers, which explains why the code is not blowing up right now.

jcarrano:
I rebased on top of master and fixed the conflicts.

jcarrano:
There are conflicts, I'm not rebasing this again. I give up - fixing bugs is clearly not a priority.

aabadie:
Is rebasing so difficult ? I don't think there are that much conflicts. I'm surprised this PR didn't get much review: it's not a difficult change, 3 reviewers were assigned.

jcarrano:
No, it is not difficult, but it is a waste of time to rebase every N months (which shows that there _is_ work being done on that module, just no interest in bug fixes). If anyone is interested on this fix, they can feel free to take over it.

aabadie:
>  there is work being done on that module, just no interest in bug fixes

I think the conflict was because #11598 was merged recently. So yes there's work on that module, and it was for fixing a bug. So argument doesn't stand.

PR #10320, opened 2018-11-02
kinetis: LPUART: Compute oversampling rate at init

gebart:
### Contribution description

By not forcing a fixed oversampling rate we can achieve better baud rate
accuracy than otherwise possible.
For example, when requesting 115200 with a module clock of 4 MHz,
picking an oversampling rate of 17 (instead of hardware
default 16) yields 117647 baud instead of 125000 baud as the best
matching rate. Better matching baud rate between receiver and
transmitter results in a lower probability of transmission errors.

### Testing procedure

Flash frdm-kw41z with a shell application
Test that the shell works correctly when using the default BAUD settings (115200)
The expected baud rate is around 117000 with this PR, 125000 without this PR. The USB to UART adapter in the dev board does handle the rate mismatch even for 125000, but there is a smaller margin for error.

### Issues/PRs references

split from #7897

### DIFF
--- a/cpu/kinetis/periph/uart.c
+++ b/cpu/kinetis/periph/uart.c
@@ -58,9 +58,14 @@
 #endif
 #endif
 
-#ifndef LPUART_OVERSAMPLING_RATE
-/* Use 16x oversampling by default (hardware defaults) */
-#define LPUART_OVERSAMPLING_RATE (16)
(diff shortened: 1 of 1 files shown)


maribu:
The code looks good to me, sadly, I don't have the hardware to test it :-(

PR #10323, opened 2018-11-05
boards: Enable STM32F103C8 ROM hack via env var

maribu:
The STM32F103C8 secretly comes with 128KiB flash instead of 64KiB. Still, only
64KiB of it are tested and guaranteed to work. However, most of the times the
whole 128KiB flash works just fine. In the BluePill documentation this fact is
already documented and by using

    $ make BOARD=bluepill CPU_MODEL=stm32f103cb

the whole 128 KiB can be used by RIOT. When using this hack routinely, it easier
to use environment variables instead. But allowing to overwrite CPU_MODEL via
environment variables seems to be a bad thing, as it is easy to forget to clear
that environment variable when changing the BOARD variable.

This commit introduces the new STM32F103C8_ROM_HACK variable, which unlocks the
128KiB ROM when set to "y". The BluePill documentation has been updated
accordingly.

### DIFF
--- a/boards/bluepill/doc.txt
+++ b/boards/bluepill/doc.txt
@@ -61,12 +61,12 @@ you can flash the device with:
 
 To make use of the entire 128 KB flash, compile your application with:
 
-    $ make BOARD=bluepill CPU_MODEL=stm32f103cb
+    $ make STM32F103C8_FLASH_HACK=1 BOARD=bluepill
 
 This sets the `CPU_MODEL` make variable to `stm32f103cb`, the default
 value is `stm32f103c8`. These two CPU models basically only have one
 major difference, the latter has 128 KB flash while the former has 64
-KB.
+KB. You may also set `STM32F103C8_FLASH_HACK` as environment variable.
 
 If you want to flash a binary compiled this way you either need to
 figure out how to adjust the OpenOCD configuration to make it use 128 KB
--- a/boards/common/stm32f103c8/Makefile.include
+++ b/boards/common/stm32f103c8/Makefile.include
@@ -1,6 +1,11 @@
 ## the cpu to build for
 export CPU = stm32f1
-export CPU_MODEL = stm32f103c8
+STM32F103C8_FLASH_HACK ?= 0
+ifneq ($(STM32F103C8_FLASH_HACK),0)
+  export CPU_MODEL = stm32f103cb
+else
+  export CPU_MODEL = stm32f103c8
+endif
 
 INCLUDES += -I$(RIOTBOARD)/common/stm32f103c8/include
 


maribu:
Good point. I renamed it as suggested and squashed the commits.

kaspar030:
If you feel like, change use of "n/y" to "0/1".

maribu:
> If you feel like, change use of "n/y" to "0/1".

Done and rebased :-)

maribu:
@kaspar030: Do you mind to let murdock run?

MrKevinWeiss:
> The STM32F103C8 secretly comes with 128KiB flash instead of 64KiB

I guess you mean the the STM32F103CB not STM32F103C8 has the additional memory.

I only have the STM32F103C8 but can confirm different behavior when enabling the STM32F103C8_ROM_HACK.
STM32F103C8_ROM_HACK=0 or default -> `region rom overflowed by 12196 bytes`
STM32F103C8_ROM_HACK=1 ->
```
diff 0 address 0x08010000. Was 0xff instead of 0x13
diff 1 address 0x08010001. Was 0xff instead of 0x48
diff 2 address 0x08010002. Was 0xff instead of 0x01
diff 3 address 0x08010003. Was 0xff instead of 0xf0
...
```
I don't know if this qualifies as tested but I too give it my ACK.

kaspar030:
Tested OK using ```STM32F103C8_FLASH_HACK=1 BOARD=bluepill make -C tests/unittests tests-gnrc_ipv6 tests-gnrc_ipv6_hdr tests-qDSA tests-spiffs tests-sixlowpan_ctx tests-sixlowpan tests-ubjson tests-uuid flash -j8```. This is a random collection of unittests that happen to cross the 64k flash boundary. Without this PR (or the FLASH hack) linking fails, with hack enabled it succeeds.

Tests also complete OK. ;)

kaspar030:
(I'm puzzled because my openocd always thinks the bluepill has 128k RAM).

gebart:
@kaspar030 doesn't OpenOCD look at the actual CPU register values for the flash specs on STM32 though?
On Kinetis the flash banks are automatically configured to the actual size and layout for the connected CPU by reading some flash info registers.
This can obviously not be done for the linking stage, so we need to manually tell the linker about the memory layout.

maribu:
As minor addition to @gebart: The documentation in the bluepill points that out and suggests to either fiddle with the openocd config (to ignore the flash size the MCU reports in this particular case) or to use a patched version st-flash. I also so far had no problems with flashing more than 64KiB using GDB and the blackmagic probe. (I'm not sure if these tools are aware that the STM32F103C8 has in fact 128 KiB ROM, or if they just ignore whatever the MCU reports at all. I guess the latter.)

PR #10289, opened 2018-10-29
Autoinit fix backportAutoinit fix backport (backport)

maribu:
Backport of https://github.com/RIOT-OS/RIOT/pull/10267

PR #10326, opened 2018-11-05
sys/auto_init: Fixed initialization of sht1x [backport 2018.10]

jia200x:
# Backport of #10267

### Contribution description
(Same as #10289 but pointing to the release branch)

Commit cef81a102c61455dd418225a009ff78558b5b9c2 introduced two bugs:

 - the driver for sht1x is initialized two times
 - the second initialization is done only when SAUL is used, but sht1x needs
   to be initialized in any case. (SAUL registration is also done there, but
   only when SAUL is actually being used.)

This commit fixes both.

### Testing procedure

E.g. use `examples/default` on a board that features an SHT1x sensor, e.g. the MSB-A2. Prior to this PR RIOT will not boot due to the double initialization, with this PR applied RIOT will boot again.

### DIFF
--- a/sys/auto_init/auto_init.c
+++ b/sys/auto_init/auto_init.c
@@ -428,10 +428,6 @@ void auto_init(void)
     extern void auto_init_pulse_counter(void);
     auto_init_pulse_counter();
 #endif
-#ifdef MODULE_SHT1X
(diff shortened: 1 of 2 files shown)


jia200x:
rebased

Issue #10134, opened 2018-10-08
`struct netdev_radio_rx_info` RSSI value is "wasteful"


maribu:
# Background
`netcev_radio_rx_info_t` is defined as:

``` C
struct netdev_radio_rx_info {
    int16_t rssi;       /**< RSSI of a received packet in dBm */
    uint8_t lqi;        /**< LQI of a received packet */
}; 
```

According to [Wikipedia's article on dBm](https://en.wikipedia.org/wiki/DBm) a value of 60 dBm equals an microwave oven and -192.5 dBm is about the received signal strength of on 1Hz thermal noise at 4 Kelvin in outer space. I expect that no IoT device will ever transmit up to 80 dBm and that being able to detect signals at -192.5 dBm is out of reach for the near future for any low power receiver.

# The Problem
This would fit perfectly in one byte of data (using an offset). On the other side, many transceivers will be able to output an RSSI value with a precision of mare than 1 dBm. (E.g. the CC1101/CC1100 has a resolution of 0.5 dBm.)

# The Proposal
By defining `struct netdev_radio_rx_info::rssi` to not be in dBm, but in E-02 dBm a resolution of 0.01 dBm would be possible while still having a range of -327.68 dBm to 327.67 dBm. This should be future proof for quite some time. (The range should be even sufficient for forever.)

maribu:
@miri64 Do you wan't to have a look?

bergzand:
Something like this was discussed before in #6887. One of the lessons from there is that we should also take the range of LoRa radios into account. This is not an argument against or in favor of your idea, but more a precaution against repeating a previous discussion

maribu:
@bergzard: -192.5 dBm as thermal noise of outer space should be a really good lower bound - everything below would be indistinguishable from that background noise. And a microwave oven (60 dBm) a pretty good upper bound. But up to 327 dBm could be stored. I lt seems to require more than 1E29 Watt of electricity for 327dBm. This is way above any IoT scenario I can image for the next few millennia. I feal confident that the range is big enough :-)

olegart:
* LoRa fits here, on hardware level it uses 1 byte with an offset anyway
* I'm not sure saving a byte is enough to justify offset introduction

maribu:
@olegart: What do think about keeping it two bytes long, but having it in E-02 dBm instead of dBm?

olegart:
[...]

PR #10334, opened 2018-11-06
{drivers/saul,sys/phydat}: Added RSSI sensor and dBm unit

maribu:
### Contribution description

The contribution is as minor as the title suggest:
 - An RSSI sensor type was added to SAUL
 - The unit dBm was added to sys/phydat (used by RSSI sensors)

### Motivation

Some network cards allow constant reading of the current RSSI value (e.g. the CC110x transceivers). This could be used for channel monitoring.

### Testing procedure

I think with only 5 lines added real testing is not required. I think if murdock still builds this should be fine.

However: In my [cc110x branch](https://github.com/maribu/RIOT/tree/cc110x) the rewrite of the cc110x driver I'm working on is providing a constant RSSI output via SAUL. This could be tested e.g. on the MSB-A2 or the MSB-IoT. (I did this and it works fine, even though the CC110x transceiver's RSSI is not too accurate.)

### Issues/PRs references

This might also be a solution to issue https://github.com/RIOT-OS/RIOT/issues/10134.

### DIFF
--- a/drivers/include/saul.h
+++ b/drivers/include/saul.h
@@ -98,6 +98,7 @@ enum {
     SAUL_SENSE_TVOC     = 0x90,     /**< sensor: TVOC Gas */
     SAUL_SENSE_OCCUP    = 0x91,     /**< sensor: occupancy */
     SAUL_SENSE_PROXIMITY= 0x92,     /**< sensor: proximity */
+    SAUL_SENSE_RSSI     = 0x93,     /**< sensor: RSSI */
     SAUL_CLASS_ANY      = 0xff      /**< any device - wildcard */
     /* extend this list as needed... */
 };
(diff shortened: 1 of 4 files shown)


gebart:
would it be possible to somehow hook netdev get NETOPT_LAST_ED_LEVEL into this?

maribu:
@gebart:

> would it be possible to somehow hook netdev get NETOPT_LAST_ED_LEVEL into this?

Sounds like a nice idea too me. But I would like to keep that out of this PR to get this one merged soon, as one of the (not yet opened) PRs for my cc110x rewrite will depend on this.

gebart:
Yes, of course. Hooking netdev into this should be a separate PR and discussion.

PR #10343, opened 2018-11-07
boards/stm32l0: Fix openocd to prevent flash locking

### LINKED ISSUE #10341: driver_my9221 prevents further flashing of nucleo-l073rz
#### Description
When running the compile_and_test_for_board script the driver_my9221 test gets flashed on the nucleo-l073rz and as a result all further attempts to flash and reset are prevented.  This can be fixed by manually holding down the reset pin just before flashing.  Once flashed with working firmware after the manual reset button press it appears to be able to flash like normal again.

`BOARD=nucleo-l073rz make flash -C tests/driver_my9221`
`BOARD=nucleo-l073rz make flash -C examples/hello-world`

#### Expected results
hello-world can flash

#### Actual results
```Error: init mode failed (unable to connect to the target)
in procedure 'init' 
in procedure 'ocd_bouncer'

~/RIOT/tests/driver_my9221/../../Makefile.include:517: recipe for target 'flash' failed
make: *** [flash] Error 1
```

#### Versions
```
Operating System Environment
-----------------------------
       Operating System: "Ubuntu" "16.04.5 LTS (Xenial Xerus)"
                 Kernel: Linux 4.15.0-38-generic x86_64 x86_64

Installed compiler toolchains
-----------------------------
             native gcc: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609
                avr-gcc: avr-gcc (GCC) 4.9.2
       mips-mti-elf-gcc: missing
             msp430-gcc: missing
   riscv-none-embed-gcc: missing
                  clang: clang version 5.0.0-3~16.04.1 (tags/RELEASE_500/final)

Installed compiler libs
-----------------------
   arm-none-eabi-newlib: "2.5.0"
    mips-mti-elf-newlib: missing

Installed development tools
---------------------------
                  cmake: cmake version 3.12.3
               cppcheck: missing
                doxygen: missing
                 flake8: 3.5.0 (mccabe: 0.6.1, pycodestyle: 2.3.1, pyflakes: 1.6.0) CPython 3.5.2 on Linux
                    git: git version 2.7.4
                openocd: Open On-Chip Debugger 0.10.0+dev-00399-g09076d1 (2018-04-12-17:12)
                 python: Python 2.7.12
                python2: Python 2.7.12
                python3: Python 3.5.2
             coccinelle: missing
```

MrKevinWeiss:
### Contribution description

This commit add the connect_assert_srst which requires the SRST to be asserted before trying to connect.
Certain firmwares will cause it to get to a state where flashing fails and the reset button must be pushed at the correct time to recover.

### Testing procedure
on 7645c66aeb584a643d7356ba41f1ecf80b06dfe8 commit
`BOARD=nucleo-l073rz make flash -C tests/driver_my9221`
then try flashing anything after.
_It seems in master the driver_my9221 doesn't cause lockup anymore but if it is already locked up it doesn't fix it (this pr does)._

### Issues/PRs references
Fixes #10341

### DIFF
--- a/boards/b-l072z-lrwan1/dist/openocd.cfg
+++ b/boards/b-l072z-lrwan1/dist/openocd.cfg
@@ -1,5 +1,0 @@
-source [find target/stm32l0.cfg]
-
-reset_config srst_only connect_assert_srst
-
-$_TARGETNAME configure -rtos auto
--- a/boards/common/stm32/dist/stm32l0.cfg
+++ b/boards/common/stm32/dist/stm32l0.cfg
@@ -1,3 +1,3 @@
 source [find target/stm32l0.cfg]
-reset_config srst_only
+reset_config srst_only connect_assert_srst
 $_TARGETNAME configure -rtos auto


aabadie:
> Certain firmwares will cause it to get to a state where flashing fails and the reset button must be pushed at the correct time to recover.

This happens mainly when the firmware uses sleep pm modes (off or standby). I agree that this PR is useful for such cases.

MrKevinWeiss:
I think we should may even expand this, the reason why the assert was left out seems to be so one could debug without resetting it catching a failure state.  I am not sure if this should be the default action though, maybe it would be better to just set a flag to stop the reset assert from happening?  If that is the case than many should be changed.

smlng:
so, how to move this forward?

MrKevinWeiss:
It would be good to see if the problem occurs on the other nucleo boards and if this fix is valid.

aabadie:
> It would be good to see if the problem occurs on the other nucleo boards and if this fix is valid.

I think it is. Some months ago, in a previous PR (maybe #8475), I already tried to pass this change. IT was working on all L0 nucleo boards (for sure). But this is not the question.
This is the question:
> the reason why the assert was left out seems to be so one could debug without resetting it catching a failure state. I am not sure if this should be the default action though, maybe it would be better to just set a flag to stop the reset assert from happening?

It is also related to #10479 and #8976 I think.

MrKevinWeiss:
> Since you are reusing the same openocd configuration than the one use by the b-l072z-lrwan1 board, maybe you could use the shared openocd configuration with this board as well ?

Done, and it makes everything slightly cleaner!

smlng:
so @aabadie is this mergeable? seems to work with the L0s a we have...

MrKevinWeiss:
I think there was a PR that changed the reset assert to default and allows you to select if you want to flash without reset (so you can get into debug mode)...  I don't know if that takes care of this.

MrKevinWeiss:
Ya, I think the issue is that one cannot enter into debug mode without it resetting.  I will ask around, I thought there was a pr that would give the option but I couldn't find it.

I would be nice to have this in for everything, it seems to make the flasher more robust.

ping @kaspar030  @gebart @smlng

aabadie:
I'm fine with this PR, but once it's merged we'll also have to merge #10479 rather soonish, otherwise `make debug` won't work on stm32l0 (and possibly other stm32).

aabadie:
I retriggered Murdock just in case and it's still green. Let's go!

MrKevinWeiss:
Thanks @aabadie!  I forgot about this one.

PR #10346, opened 2018-11-07
sys/Makefile.include: remove some module specific code.

jcarrano:
### Contribution description

This is a small step towards making modules more self-contained and allow code to be added with less modifications to shared files.

Ideally each *module* should take care of itself. That is the point of *modul*arity. A user adding a new module should not need to touch any global file.

Currently sys/Makefile.include has an `if` statement for each module that needs it's own `Makefile.include` included. I replaced it with a generic patterned include.

I also used this opportunity to take out some module specific logic and put it in a separate file.

### Testing procedure

Modules possibly affected are: 

- embunit (affects all unit tests)
- arduino (examples/arduino_hello-world)
- ssp (tests/ssp)
- newlib_syscalls_default (too many things)
- cpp11-compat (tests/cpp11_mutex)

All of those have tests / examples, so the CI should catch any failure to compile. To test it locally just run the listed applications.


sys/Makefile.include
```diff
@@ -35,26 +35,7 @@ ifneq (,$(filter vfs,$(USEMODULE)))
...
-    CFLAGS += -DOUTPUT=OUTPUT_TEXT
-  else ifeq ($(OUTPUT),COMPILER)
-    CFLAGS += -DOUTPUT=OUTPUT_COMPILER
-  else ifeq ($(OUTPUT),COLORTEXT)
-    CFLAGS += -DOUTPUT=OUTPUT_COLORTEXT
-  else ifeq ($(OUTPUT),COLOR)
-    CFLAGS += -DOUTPUT=OUTPUT_COLOR
-  endif
-endif
+-include $(patsubst %,$(RIOTBASE)/sys/%/Makefile.include,$(USEMODULE))
```
cladmi: As discussed IRL, I am not sure I want to go in this direction for the moment.

It goes in a direction were module names must be a directory name at the root of `sys`, and we know it does not work or apply for everything.

I would find it weird that for example the posix module gets put in a flat hierarchy:

```
ls sys/posix/
doc.txt  include  inet.c  Makefile  pthread  semaphore  sockets  time
```
as 
```
sys/posix
sys/posix_pthread
sys/posix_semaphore
...
```

And would not want to rename the directory inside to `posix/posix_pthread`.

It pushes to a new hierarchy, without giving the flexibility currently required for what we want to do.

----

However this IRL long intense discussions ended up going more for `posix/pthread` module names to have a deterministic way of finding directly where a module is implemented if you know all the module trees.

PR #10349, opened 2018-11-08
cpu/nrf5x: update vendor headers

ghost:
This PR updates vendor headers for nordic nrf51, nrf52 and nrf52840 CPUs with actual version from [nordics nrfx repository](https://github.com/NordicSemiconductor/nrfx).

I tested building default/example with calliope-mini and also flashing with nrf52dk and nrf52840dk.


haukepetersen:
did just start to test, but it makes sense to get rid of the compiler warnings first... :-)

PR #10355, opened 2018-11-09
radios: Remove default radio flags from init

bergzand:
### Contribution description

This removes the default flag configuration from the initialization function of the supported radios. Instead, the upper layer network function configures the required/enabled flags on the radio.

The idea is to have a more centralized configuration of these flags instead of every radio enabling a different set of flags on init.

### Testing procedure

Test if every radio correctly configures flags based on the configuration in `sys/include/net/ieee802154.h`

 - [ ] cc2538 (Supports only autoack as far as I know)
 - [ ] at86rf2xx
 - [ ] cc2420
 - [ ] kw2xrf
 - [ ] mrf24j40

### Issues/PRs references

requires #9592


drivers/kw2xrf/kw2xrf.c
```diff
@@ -104,8 +104,6 @@ void kw2xrf_reset_phy(kw2xrf_t *dev)
 
     kw2xrf_set_rx_watermark(dev, 1);
 
-    kw2xrf_set_option(dev, KW2XRF_OPT_AUTOACK, true);
-    kw2xrf_set_option(dev, KW2XRF_OPT_ACK_REQ, true);
     kw2xrf_set_option(dev, KW2XRF_OPT_AUTOCCA, true);
```
bergzand: I kept the `AUTOCCA` flag here as it has a slightly different meaning than the `CSMA` netopt. It also has its own netopt option.

gebart: An alternative could be to try setting csma first, and if that fails, try to set autocca. What do you think of this?

gebart: We can copy the behavior from kw41zrf. I did exactly what you describe in order to provide csma for that radio.

gebart: Someone else will have to provide the PR. I don't have the hardware and have no interest in the kw2xdrf

bergzand: > Someone else will have to provide the PR. I don't have the hardware and have no interest in the kw2xdrf

Same for me to be honest

PeterKietzmann: Well, we're all short in time but we (HAW Hamburg) have certain interest in that radio so it might be us to do the follow-up. @gebart, @bergzand could you please open a feature request issue?

gebart: @PeterKietzmann I opened #10364 as a reminder

PeterKietzmann:
@jia200x I assume you know this PR. IMO we should try to address this one in conjunction with #11473

PR #10363, opened 2018-11-10
drivers/vl53l1x: VL53L1X ToF ranging sensor added as package

gschorcht:
### Contribution description

This PR was originally intended only as an example for the discussion on how to add a device driver as a package. In the meantime I finished the driver. It is fully functional and therefore ready for review.

#### Overview

This PR adds a device driver for ST VL53L1X Time-of-Flight laser-ranging sensor. The VL53L1X allows accurate and fast ranging up to 4 m and with a frequency of up to 50 Hz.

There are three variants of the driver which differ in functionality and size:

Module         | Driver                | Short Description
:-------------------|:----------------------|:----------------------------
`vl53l1x`        | Standard driver       | Common functionality and medium size
`vl53l1x_st_api` | ST VL53L1X API driver | Complete functionality and large size
`vl53l1x_basic`  | Basic driver          | Minimum functionality and small size

This `vl53l1x_st_api` variant of the driver uses the ST STSW-IMG007 VL53L1X API as a package (`pkg/driver_vl53l1x_st_api`). In this case, the driver is simply a wrapper for the API which provides a driver interface that is compatible with the other driver variants. Since the ST STSW-IMG007 VL53L1X API package is quite complex and very large, this driver variant can only be used when memory requirements are not an issue. It has to be used for best accuracy, for calibration and if threshold interrupts shall be used.

The `vl53l1x_std` driver variant (default) is a compromise of size and functionality. It provides the application with most functionality of the sensor. The driver can be used when memory requirements are important and most of the functionality should be used.

The `vl53l1x_basic` driver variant is the smallest driver variant which only provides some basic functions such as the distance measurement and the data-ready interrupt.

All driver variants provide SAUL capabilities and data-ready interrupt functionality. However, only the `vl53l1x_st_api` driver variant allows to use threshold interrupts.

Function / Property                    | `vl53l1x_basic` | `vl53l1x_std` (default) | `vl53l1x_st_api`
:--------------------------------------|:------:|:----:|:--------------:
Distance results in mm                 | X      | X    | X
Signal rate results in MCPS            |        | X    | X
Measurement status information          |        | X    | X
SAUL capability                        | X      | X    | X
Distance mode configuration            |        | X    | X
Timing budget configuration            |        | X    | X
Inter-measurement period configuration |        | X    | X
Region of Interest (ROI) configuration |        | X    | X
Data-ready interrupts                  | X      | X    | X
Threshold interrupts                   |        |      | X
Calibration functions                  |        |      | X [1]
Limit check configuration              |        |      | X [1]
Accuracy                               | medium | high | high
Size on reference platform in kByte [2]   | 1.0    | 2.9  | 19.8

[1] These functions are available by using the ST VL53L1X API directly.
[2] Reference platform: STM32F411RE

### Testing procedure

The driver variants can be tested by using the test application with different modules where the standard variant `vl53l1x_std` is used by default:
```
make flash -C tests/driver_vl53l1x BOARD=...
```
To use the other driver variants, module `vl53l1x_st_api` or module `vl53l1x_basic`
have to be specified at make command line
```
USEMODULE=vl53l1x_st_api make flash -C tests/driver_vl53l1x BOARD=...
```
or
```
USEMODULE=vl53l1x_basic make flash -C tests/driver_vl53l1x BOARD=... 
```
If the configuration parameter `VL53L1X_PARAM_PIN_INT ` is defined, e.g.,
```
CFLAGS='-DVL53L1X_PARAM_PIN_INT=\(GPIO\(0,\1\)' make flash -C tests/driver_vl53l1x BOARD=...
``` 
interrupts are used to get data instead of polling for new data. In the case of driver variant ```vl53l1x_st_api```, threshold interrupts are configured.

In all cases, the sensor is configured with following default parameters:
- timing budget of 50 ms
- intermeasurement period of 100 ms

### Issues/PRs references

Some weeks ago @dylad was providing a sensor driver for the Bosch BME680 gas sensor, see PR #9909. Since the calculations are quite complex, Bosch provides a ready-to-use API. Therefore, he and @aabadie decided that it would be better to add Bosch's BME680 API driver as package.

Since the ST VL53L1X is even more complex, I decided to try it in the same way. This PR is a try to structure a driver as a package.

See also PR #9909


gschorcht:
It's waiting for review.

javierfileiv:
Hi! This is a great PR! Thanks @gschorcht ! I was working on a native VL53L0X, but it was turning into a burden with all this i2c settings that come out of nowhere. I'ḿ not sure if your porting is compatible with my breakbord. Otherwise I can port it in a similar way as you did. Thanks!

maribu:
I do have some VL52L*0*X sensors lying around. I assume those are incompatible, is this right?

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+     * apply correction gain
+     * gain factor of 2011 is tuning parm default
+     * (VL53L1_TUNINGPARM_LITE_RANGING_GAIN_FACTOR_DEFAULT)
+     * Basically, this appears to scale the result by 2011/2048, or about 98%
+     * (with the 1024 added for proper rounding).
+     */
+    data->distance = ((uint32_t)range * 2011 + 0x0400) / 0x0800;
+    /* Copyrighted POLOLU / ST API code end */
+
+    #if !MODULE_VL53L1X_BASIC
```
maribu, line 345:
```suggestion
#if !IS_USED(MODULE_VL53L1X_BASIC)
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    /* ambient rate in in 9.7 fixed point format */
+    data->ambient_rate  = (uint32_t)raw.ambient_count_rate_mcps_sd0_hi << 8;
+    data->ambient_rate |= raw.ambient_count_rate_mcps_sd0_lo;
+
+    /* convert values from 9.7 to 16.16 */
+    data->signal_rate  <<= 9;
+    data->ambient_rate <<= 9;
+
+    #endif /* !MODULE_VL53L1X_BASIC */
```
maribu, line 358:
```suggestion
#endif /* !MODULE_VL53L1X_BASIC */
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    /* restart measurement */
+    EXEC_RET(_start(dev));
+
+    #endif /* MODULE_VL53L1X_ST_API */
```
maribu, line 363:
```suggestion
#endif /* MODULE_VL53L1X_ST_API */
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    if (!dev->int_init) {
+        dev->int_init = true;
+        gpio_init_int(dev->params.pin_int, GPIO_IN_PU, GPIO_FALLING,
+                      isr, isr_arg);
+    }
+
+    #if MODULE_VL53L1X_ST_API
+    ASSERT_PARAM(cfg != NULL);
+
+    EXEC_RET(VL53L1_StopMeasurement((VL53L1_Dev_t*)dev));
```
maribu, line 399:
```suggestion
    EXEC_RET(VL53L1_StopMeasurement(&dev->dev));
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    dcfg.Distance.CrossMode = cfg->dist_mode;
+    dcfg.Distance.High = cfg->dist_high;
+    dcfg.Distance.Low = cfg->dist_low;
+    dcfg.Rate.CrossMode = cfg->rate_mode;
+    dcfg.Rate.High = cfg->rate_high;
+    dcfg.Rate.Low = cfg->rate_low;
+    dcfg.IntrNoTarget = 0;
+
+    EXEC_RET(VL53L1_SetThresholdConfig((VL53L1_Dev_t*)dev, &dcfg));
+    EXEC_RET(VL53L1_StartMeasurement((VL53L1_Dev_t*)dev));
```
maribu, line 412:
```suggestion
    EXEC_RET(VL53L1_SetThresholdConfig(&dev->dev, &dcfg));
    EXEC_RET(VL53L1_StartMeasurement(&dev->dev));
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    #else
```
maribu, line 414:
```suggestion
#else
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    DEBUG_DEV("configuration is not used", dev);
+    #endif
```
maribu, line 416:
```suggestion
#endif
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    return VL53L1X_OK;
+}
+
+#if !MODULE_VL53L1X_BASIC
```
maribu, line 421:
```suggestion
#if !IS_USED(MODULE_VL53L1X_BASIC)
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+}
+
+int vl53l1x_get_timing_budget(const vl53l1x_t *dev, uint32_t *budget)
+{
+    ASSERT_PARAM(dev != NULL);
+    ASSERT_PARAM(budget != NULL);
+    DEBUG_DEV("budget=%p", dev, budget);
+
+    #if MODULE_VL53L1X_ST_API
+    return VL53L1_GetMeasurementTimingBudgetMicroSeconds((VL53L1_Dev_t*)dev,
```
maribu, line 462:
```suggestion
    return VL53L1_GetMeasurementTimingBudgetMicroSeconds(&dev->dev,
```

Btw: Are the IT programmers payed by character count? :rofl:

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+                                                         budget);
+    #else
```
maribu, line 464:
```suggestion
#else
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    EXEC_RET(_read_word(dev, VL53L1X_RANGE_CONFIG__TIMEOUT_MACROP_A, &word));
+    timeout_us = _timeout_clk_to_us(_dec_timeout(word), period_us);
+
+    *budget = 2 * timeout_us + _timing_guard;
+    /* Copyrighted POLOLU / ST API code end */
+
+    return VL53L1X_OK;
+
+    #endif
```
maribu, line 486:
```suggestion
#endif
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+}
+
+int vl53l1x_set_timing_budget(vl53l1x_t *dev, uint32_t budget_us)
+{
+    ASSERT_PARAM(dev != NULL);
+    DEBUG_DEV("budget_us=%u", dev, budget_us);
+
+    #if MODULE_VL53L1X_ST_API
```
maribu, line 494:
```suggestion
#if IS_USED(MODULE_VL53L1X_ST_API)
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    EXEC_RET(VL53L1_SetMeasurementTimingBudgetMicroSeconds((VL53L1_Dev_t*)dev,
```
maribu, line 496:
```suggestion
    EXEC_RET(VL53L1_SetMeasurementTimingBudgetMicroSeconds(&dev->dev,
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+                                                           budget_us));
+    #else
```
maribu, line 498:
```suggestion
#else
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    /* Update MM Timing B timeout, see comment about MM Timing A timeout. */
+    clks = _enc_timeout(_timeout_us_to_clk(1, period_us));
+    EXEC_RET(_write_word(dev, VL53L1X_MM_CONFIG__TIMEOUT_MACROP_B, clks));
+
+    /* Update Range Timing B timeout */
+    clks = _enc_timeout(_timeout_us_to_clk(timeout_us, period_us));
+    EXEC_RET(_write_word(dev, VL53L1X_RANGE_CONFIG__TIMEOUT_MACROP_B, clks));
+
+    /* Copyrighted POLOLU / ST API code end */
+    #endif
```
maribu, line 560:
```suggestion
#endif
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    return VL53L1X_OK;
+}
+
+int vl53l1x_set_measurement_period(vl53l1x_t *dev, uint32_t period_ms)
+{
+    ASSERT_PARAM(dev != NULL);
+    ASSERT_PARAM((period_ms + 4) * 1000 > dev->params.budget);
+    DEBUG_DEV("period_ms=%u", dev, period_ms);
+
+    #if MODULE_VL53L1X_ST_API
```
maribu, line 574:
```suggestion
#if IS_USED(MODULE_VL53L1X_ST_API)
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    EXEC_RET(VL53L1_SetInterMeasurementPeriodMilliSeconds((VL53L1_Dev_t*)dev,
+                                                          period_ms));
+    #else
```
maribu, line 578:
```suggestion
#else
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    /* Copyrighted POLOLU / ST API code begin */
+
+    /* from VL53L1_set_inter_measurement_period_ms() */
+    uint32_t period = period_ms * dev->osc_calibrate_val;
+    EXEC_RET(_write_dword(dev, VL53L1X_SYSTEM__INTERMEASUREMENT_PERIOD, period));
+
+    /* Copyrighted POLOLU / ST API code end */
+    #endif
```
maribu, line 586:
```suggestion
#endif
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    return VL53L1X_OK;
+}
+
+int vl53l1x_set_distance_mode(vl53l1x_t *dev, vl53l1x_dist_mode_t mode)
+{
+    ASSERT_PARAM(dev != NULL);
+    DEBUG_DEV("mode=%u", dev, mode);
+
+    #if MODULE_VL53L1X_ST_API
```
maribu, line 599:
```suggestion
#if IS_USED(MODULE_VL53L1X_ST_API)
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    EXEC_RET(VL53L1_SetDistanceMode((VL53L1_Dev_t*)dev, mode));
+
+    #else
```
maribu, line 603:
```suggestion
#else
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+            EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__WOI_SD0, 0x0f));
+            EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__WOI_SD1, 0x0d));
+            EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__INITIAL_PHASE_SD0, 14));
+            EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__INITIAL_PHASE_SD1, 14));
+            break;
+
+        default:
+            /* unrecognized mode - do nothing */
+            return VL53L1X_ERROR_INV_ARG;
+    }
```
maribu, line 655:
```suggestion
    switch (mode) {
    case VL53L1X_DIST_SHORT:
        /* from VL53L1_preset_mode_standard_ranging_short_range() */
        /* timing config */
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VCSEL_PERIOD_A, 0x07));
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VCSEL_PERIOD_B, 0x05));
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VALID_PHASE_HIGH, 0x38));

        /* dynamic config */
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__WOI_SD0, 0x07));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__WOI_SD1, 0x05));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__INITIAL_PHASE_SD0, 6));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__INITIAL_PHASE_SD1, 6));
        break;

    case VL53L1X_DIST_MEDIUM:
        /* from VL53L1_preset_mode_standard_ranging() */
        /* timing config */
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VCSEL_PERIOD_A, 0x0b));
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VCSEL_PERIOD_B, 0x09));
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VALID_PHASE_HIGH, 0x78));

        /* dynamic config */
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__WOI_SD0, 0x0b));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__WOI_SD1, 0x09));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__INITIAL_PHASE_SD0, 10));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__INITIAL_PHASE_SD1, 10));
        break;

    case VL53L1X_DIST_LONG:
        /* from VL53L1_preset_mode_standard_ranging_long_range() */
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VCSEL_PERIOD_A, 0x0f));
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VCSEL_PERIOD_B, 0x0d));
        EXEC_RET(_write_byte(dev, VL53L1X_RANGE_CONFIG__VALID_PHASE_HIGH, 0xb8));

        /* dynamic config */
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__WOI_SD0, 0x0f));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__WOI_SD1, 0x0d));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__INITIAL_PHASE_SD0, 14));
        EXEC_RET(_write_byte(dev, VL53L1X_SD_CONFIG__INITIAL_PHASE_SD1, 14));
        break;

    default:
        /* unrecognized mode - do nothing */
        return VL53L1X_ERROR_INV_ARG;
    }
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    /* reapply timing budget */
+    EXEC_RET(vl53l1x_set_timing_budget(dev, budget_us));
+
+    /* Copyrighted POLOLU / ST API code end */
+    #endif
```
maribu, line 661:
```suggestion
#endif
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    ASSERT_PARAM(roi->x_tl < roi->x_br);
+    ASSERT_PARAM(roi->y_tl > roi->y_br);
+    ASSERT_PARAM(roi->x_br - roi->x_tl >= 3);
+    ASSERT_PARAM(roi->y_tl - roi->y_br >= 3);
+
+    DEBUG_DEV("roi=%p", dev, roi);
+
+    #if MODULE_VL53L1X_ST_API
+    return VL53L1_SetUserROI((VL53L1_Dev_t*)dev, (VL53L1_UserRoi_t*)roi);
+    #else /* MODULE_VL53L1X_ST_API */
```
maribu, line 685:
```suggestion
#if IS_USED(MODULE_VL53L1X_ST_API)
    return VL53L1_SetUserROI(&dev->dev, (VL53L1_UserRoi_t*)roi);
#else /* MODULE_VL53L1X_ST_API */
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    uint8_t w   = (roi->x_br - roi->x_tl);
+    uint8_t h   = (roi->y_tl - roi->y_br);
+    uint8_t dim = (h << 4) + w;
+
+    /* Copyrighted ST API code end */
+
+    EXEC_RET(_write_byte(dev, VL53L1X_ROI_CONFIG__USER_ROI_CENTRE_SPAD, spad));
+    EXEC_RET(_write_byte(dev, VL53L1X_ROI_CONFIG__USER_ROI_REQUESTED_GLOBAL_XY_SIZE, dim));
+
+    #endif /* MODULE_VL53L1X_ST_API */
```
maribu, line 704:
```suggestion
#endif /* MODULE_VL53L1X_ST_API */
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+int vl53l1x_get_roi(const vl53l1x_t *dev, vl53l1x_roi_t *roi)
+{
+    ASSERT_PARAM(dev != NULL);
+    ASSERT_PARAM(roi != NULL);
+
+    DEBUG_DEV("roi=%p", dev, roi);
+
+    #if MODULE_VL53L1X_ST_API
+    return VL53L1_GetUserROI((VL53L1_Dev_t*)dev, (VL53L1_UserRoi_t*)roi);
+    #else /* MODULE_VL53L1X_ST_API */
```
maribu, line 718:
```suggestion
#if IS_USED(MODULE_VL53L1X_ST_API)
    return VL53L1_GetUserROI(&dev->, (VL53L1_UserRoi_t*)roi);
#else /* MODULE_VL53L1X_ST_API */
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    uint8_t w = dim & 0x0f;
+
+    /* from VL53L1_GetUserROI() */
+    roi->x_tl = (2 * col - w) >> 1;
+    roi->y_tl = (2 * row - h) >> 1;
+    roi->x_br = (2 * col + w) >> 1;
+    roi->y_br = (2 * row + h) >> 1;
+
+    /* Copyrighted ST API code end */
+    #endif /* MODULE_VL53L1X_ST_API */
```
maribu, line 752:
```suggestion
#endif /* MODULE_VL53L1X_ST_API */
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+}
+
+#define VL53L1X_BOOT_TIMEOUT    10      /* 10 ms */
+
+static int _reset(const vl53l1x_t *dev)
+{
+    #if MODULE_VL53L1X_ST_API
+
+    /* write reset impuls for 150 us */
+    VL53L1_software_reset((VL53L1_Dev_t*)dev);
```
maribu, line 930:
```suggestion
    VL53L1_software_reset(&dev->dev);
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    #endif /* MODULE_VL53L1X_BASIC */
+
+    /* start measurement */
+    EXEC_RET(_start(dev));
+
+    return VL53L1X_OK;
+}
+
+#define VL53L1X_BOOT_TIMEOUT    10      /* 10 ms */
```
maribu, line 923: Should this define be in an internal header, so that Doxygen can see it?

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    #if MODULE_VL53L1X_ST_API
+
+    /* write reset impuls for 150 us */
+    VL53L1_software_reset((VL53L1_Dev_t*)dev);
+
+    /* give the sensor some time to boot */
+    xtimer_usleep(4 * US_PER_MS);
+
+    /* wait until the sensor has been booted */
+    EXEC_RET(VL53L1_WaitDeviceBooted((VL53L1_Dev_t*)dev));
```
maribu, line 936:
```suggestion
    EXEC_RET(VL53L1_WaitDeviceBooted(&dev->dev));
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    /* set VDDIO */
+    EXEC_RET(VL53L1_UpdateByte((VL53L1_Dev_t*)dev,
```
maribu, line 939:
```suggestion
    EXEC_RET(VL53L1_UpdateByte(&dev->dev,
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+                               VL53L1X_PAD_I2C_HV__EXTSUP_CONFIG,
+                               0xfe, dev->params.vddio_2v8 ? 1 : 0));
+
+    #else /* MODULE_VL53L1X_ST_API */
```
maribu, line 943:
```suggestion
#else /* MODULE_VL53L1X_ST_API */
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+        if ((_read_byte(dev, VL53L1X_FIRMWARE__SYSTEM_STATUS, &res) == VL53L1X_OK) &&
+            (res & 0x01) == 1) {
+            break;
+        }
+        xtimer_usleep(1 * US_PER_MS);
+    }
+    if (timeout == 0) {
+        DEBUG_DEV("could not boot the sensor", dev);
+        return VL53L1X_ERROR_TIMEOUT;
+    }
```
maribu, line 963: Are you sure that the postdecrement in the `while()` doesn't trigger after `timeout` has reached zero? I have the gutfeeling that it does.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    /* set VDDIO */
+    EXEC_RET(_update(dev, VL53L1X_PAD_I2C_HV__EXTSUP_CONFIG,
+                          0xfe, dev->params.vddio_2v8 ? 1 : 0));
+
+    #endif /* MODULE_VL53L1X_ST_API */
```
maribu, line 969:
```suggestion
#endif /* MODULE_VL53L1X_ST_API */
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    return VL53L1X_OK;
+}
+
+#if !MODULE_VL53L1X_ST_API
```
maribu, line 974:
```suggestion
#if !IS_USED(MODULE_VL53L1X_ST_API)
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    EXEC_RET(_write_byte(dev, VL53L1X_SYSTEM__INTERRUPT_CLEAR, 0x01));
+    EXEC_RET(_write_byte(dev, VL53L1X_SYSTEM__MODE_START, 0x40));
+
+    /* Copyrighted POLOLU / ST API code end */
+    return VL53L1X_OK;
+}
+
+/* Copyrighted POLOLU / ST API code begin */
+
+#if !MODULE_VL53L1X_BASIC
```
maribu, line 990:
```suggestion
#if !IS_USED(MODULE_VL53L1X_BASIC)
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+}
+
+/* Copyrighted POLOLU / ST API code end */
+
+#define VL53L1X_BUFSIZ   32
+static uint8_t _buffer[VL53L1X_BUFSIZ] = {};
+
+static int _read_byte(const vl53l1x_t *dev, uint16_t index, uint8_t *data)
+{
+    return _read (dev, index, data, 1);
```
maribu, line 1168:
```suggestion
    return _read(dev, index, data, 1);
```

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+}
+
+static int _read_word(const vl53l1x_t *dev, uint16_t index, uint16_t *data)
+{
+    uint8_t bytes[2];
+    EXEC_RET(_read (dev, index, bytes, 2));
+    *data = (bytes[0] << 8) + (bytes[1]);
```
maribu, line 1175:
```suggestion
    *data = ((uint16_t)bytes[0] << 8) + (bytes[1]);
```

I *think* this is correct without the cast suggested, as all types smaller than `int` get promoted up to `int` prior to shifting. And `int` is if I recall correctly guaranteed to be at least 16 bits wide. But I would have to consult the C standard to be sure, and at this point I think it would be easier to just add the cast.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+static int _write_byte(const vl53l1x_t *dev, uint16_t index, uint8_t data)
+{
+    return _write (dev, index, &data, 1);
+}
+
+static int _write_word(const vl53l1x_t *dev, uint16_t index, uint16_t data)
+{
+    uint8_t bytes[2] = { (data >> 8) & 0xff, data & 0xff };
+    return _write (dev, index, (uint8_t*)&bytes, 2);
```
maribu, line 1230:
```suggestion
    return _write(dev, index, bytes, 2);
```

`&bytes` is a of type `uint8_t **`, so pointer to a pointer to `uint8_t`. So previously this would have written two bytes of the address of `bytes`.

maribu (commented):
I haven't seen all files yet, but some comments inline. I can resume review when I have the hardware for testing.

gschorcht:
@maribu Ups, I have to appologize, my plan was to improve the code a bit before because the code is quite old and contains a lot of code style violations :worried: Anyway, thanks for starting the review.

gschorcht:
@maribu I would have to touch the code anyway because of the conflicts and the changes in autoinitialization.

maribu:
My hardware arrived. Feel free to ping me when I should proceed with the review.

drivers/vl53l1x/include/vl53l1x_register_map.h
```diff
@@ -0,0 +1,11881 @@
+/*
```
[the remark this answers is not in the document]
gschorcht: This file is part of the ST VL53L1X driver package. Maybe we could pull the package even if it is not used.

drivers/include/vl53l1x.h
```diff
@@ -0,0 +1,586 @@
...
+typedef enum {
+    #if MODULE_VL53L1X_ST_API
+    VL53L1X_DIST_SHORT  = VL53L1_DISTANCEMODE_SHORT,  /**< up to 1.3 m */
+    VL53L1X_DIST_MEDIUM = VL53L1_DISTANCEMODE_MEDIUM, /**< up to 2 m */
+    VL53L1X_DIST_LONG   = VL53L1_DISTANCEMODE_LONG,   /**< up to 4 m */
+    #else
+    VL53L1X_DIST_SHORT,     /**< up to 1.3 m */
+    VL53L1X_DIST_MEDIUM,    /**< up to 2 m */
+    VL53L1X_DIST_LONG,      /**< up to 4 m */
+    #endif
```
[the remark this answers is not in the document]
gschorcht: These symbols are defined in `pkg/driver_vl53l1x_st_api/vl53l1x-st-api/vl53l1_def.h`. They are available only if the ST VL53L1X driver is used by enabling the `vl53l1x_st_api` module.

drivers/include/vl53l1x.h
```diff
@@ -0,0 +1,586 @@
...
+ *
+ * @param[in]   dev     device descriptor of VL53L1X sensor
+ * @param[out]  mm      ranging data in mm
+ *
+ * @retval  VL53L1X_OK         on success
+ * @retval  VL53L1X_ERROR_*    negative error code
+ */
+int vl53l1x_read_mm(vl53l1x_t *dev, int16_t *mm);
+
+#if !MODULE_VL53L1X_BASIC
```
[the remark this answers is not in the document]
gschorcht: It shouldn't be necessary because `MODULE_VL53L1X_BASIC` isn't defined when doxygen is executed so that the condition is true.

drivers/include/vl53l1x.h
```diff
@@ -0,0 +1,586 @@
...
+ */
+#define VL53L1X_OK              (0)                      /**< success */
+#define VL53L1X_ERROR_I2C       (VL53L1X_ERROR_BASE - 1) /**< I2C communication error */
+#define VL53L1X_ERROR_WRONG_ID  (VL53L1X_ERROR_BASE - 2) /**< wrong id read */
+#define VL53L1X_ERROR_NO_DATA   (VL53L1X_ERROR_BASE - 3) /**< no data available */
+#define VL53L1X_ERROR_RAW_DATA  (VL53L1X_ERROR_BASE - 4) /**< read raw data failed */
+#define VL53L1X_ERROR_NO_PIN    (VL53L1X_ERROR_BASE - 5) /**< pin not defined */
+#define VL53L1X_ERROR_TIMEOUT   (VL53L1X_ERROR_BASE - 6) /**< timeout happened */
+#define VL53L1X_ERROR_INV_ARG   (VL53L1X_ERROR_BASE - 7) /**< invalid argument */
+#define VL53L1X_ERROR_INV_PARAM (VL53L1X_ERROR_BASE - 8) /**< invalid params */
```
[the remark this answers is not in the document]
gschorcht: > Otherwise, it is now preferred to return 0 on success and negative errno codes on failure. 

I know, I was involved in this discussion some years ago in issue #10686. The use of POSIX error codes is not trivial for drivers, since it is often not clear which error should be assigned to which POSIX error code.

In the case of this driver, it is even more difficult because the original error codes returned by functions of the ST VL53L1X package are returned directly by the driver functions. It would be difficult to convert them, especially if a number of functions of the ST VL53L1X package are called in a single driver function:

drivers/vl53l1x/vl53l1x.c
gschorcht: Is this change really necessary? I really don't like this style because it reduces the readability.

I know, according to the Linux coding style it is the preferred way, but does is it mean that it is forbidden to use double Indentation? In Linux style 8 chars are used for indentation and the preferred switch/case indentation is a way to reduce the indentation levels to avoid that the code moves too far to the right on a 80 char terminal. 

It might be considered as a violation of their own "idea behind indentation to clearly define where a block of control starts and ends". The switch statement starts a block and all statements inside a block would have to be indented including the case statements.

I know, we could discuss about it for hours. Finally, when I'm crawling through RIOT code, both indentation schemes are used for switch/case statements, also in mature code. Therefore, I would prefer to leave it as it is. Would that be OK?

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+                                VL53L1X_RANGESTATUS_RANGE_VALID_NO_WRAP_CHECK_FAIL :
+                                VL53L1X_RANGESTATUS_RANGE_VALID;
+                 break;
+        case 12: data->status = VL53L1X_RANGESTATUS_XTALK_SIGNAL_FAIL; break;
+        case 13: data->status = VL53L1X_RANGESTATUS_MIN_RANGE_FAIL; break;
+        case 17: data->status = VL53L1X_RANGESTATUS_HARDWARE_FAIL; break;
+        case 18: data->status = VL53L1X_RANGESTATUS_SYNCRONISATION_INT; break;
+        case 19: data->status = VL53L1X_RANGESTATUS_RANGE_VALID_NO_WRAP_CHECK_FAIL; break;
+        default: data->status = VL53L1X_RANGESTATUS_NONE;
+    }
```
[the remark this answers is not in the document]
gschorcht: > It was recently clarified that we don't want the double indent.

Sorry, I haven't seen this when I left my comment for the other switch/case change request. I have read all coding convention very carefully. The question for me was whether "preferred way" means "must" and not only "should". As said in my comment there, from my point of view, the switch/case indentation style of Linux is a violation of their own idea of indentation and the situation with 8 chars indentation is different. We have only 4 characters indentation and 80 chars is a should. And we have a lot of mature code in RIOT that uses double indentation. Having something inside braces that is not indented is a basic conflict for me, I hate it :worried:

Anyway, was it a discussion that can be read somewhere, for example, an issue?

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    ASSERT_PARAM(params->period >= params->budget/1000 + 4);
+    #endif /* !MODULE_VL53L1X_BASIC */
+    DEBUG_DEV("params=%p", dev, params);
+
+    /* init sensor data structure */
+    dev->params = *params;
+    dev->int_init = false;
+
+    #if MODULE_VL53L1X_ST_API
+    dev->dev.I2cDevAddr = VL53L1X_I2C_ADDRESS;
```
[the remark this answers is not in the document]
gschorcht: The VL53L1X has a fixed default I2C address. Although it can be changed with an ST-API function, I don't know how this would work if there are multiple devices in the system that all have the same default slave address to begin with.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    }
+    #endif
+
+    /* init the sensor and start measurement */
+    _init(dev);
+
+    return VL53L1X_OK;
+}
+
+int vl53l1x_data_ready (const vl53l1x_t *dev)
```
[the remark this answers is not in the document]
gschorcht: Yeah that is an old style problem of mine. Because the syntax highlighting of my editor was too bad, I had to leave a space there.

drivers/vl53l1x/vl53l1x.c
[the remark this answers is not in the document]
gschorcht: Ok, I have changed it but I'm not lucky with. It looks really strange.

What if blocks are used in cases, for example? This looks crazy.
```
switch {
case 1: {
    ...
}
case 1: {
    ...
}
default: {
    ...
}
}
```
Again the question, does "preferred" mean "must"?

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+#include "vl53l1x.h"
+
+#include "irq.h"
+#include "log.h"
+#include "xtimer.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+#if ENABLE_DEBUG
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/ce64675d203931b45fa0777ba7edf52637e9160b

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+#include <stdlib.h>
+
+#include "vl53l1x_regs.h"
+#include "vl53l1x.h"
+
+#include "irq.h"
+#include "log.h"
+#include "xtimer.h"
+
+#define ENABLE_DEBUG    (0)
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+/* Copyrighted POLOLU / ST API code end */
+
+#endif /* MODULE_VL53L1X_ST_API */
+
+int vl53l1x_init(vl53l1x_t *dev, const vl53l1x_params_t *params)
+{
+    /* some parameter sanity checks */
+    ASSERT_PARAM(dev != NULL);
+    ASSERT_PARAM(params != NULL);
+    #if !MODULE_VL53L1X_BASIC
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    ASSERT_PARAM(params->budget <= 1000000);
+    ASSERT_PARAM(params->period >= params->budget/1000 + 4);
+    #endif /* !MODULE_VL53L1X_BASIC */
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    DEBUG_DEV("params=%p", dev, params);
+
+    /* init sensor data structure */
+    dev->params = *params;
+    dev->int_init = false;
+
+    #if MODULE_VL53L1X_ST_API
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    dev->dev.I2cDevAddr = VL53L1X_I2C_ADDRESS;
+    #elif !MODULE_VL53L1X_BASIC
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    dev->calibrated = false;
+    #endif
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    #if !MODULE_VL53L1X_BASIC
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    /* if shutdown pin is defined, it is initialized first and set */
+    if (params->pin_shutdown != GPIO_UNDEF) {
+        gpio_init(params->pin_shutdown, GPIO_OUT);
+        gpio_write(params->pin_shutdown, 1);
+    }
+    #endif
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+    EXEC_RET(VL53L1_GetMeasurementDataReady((VL53L1_Dev_t*)dev, &ready));
+    return (ready) ? VL53L1X_OK : -VL53L1X_ERROR_NO_DATA;
+    #else
+    uint8_t byte;
+    EXEC_RET(_read_byte(dev, VL53L1X_GPIO__TIO_HV_STATUS, &byte));
+    return ((byte & 0x01) == 0) ? VL53L1X_OK : -VL53L1X_ERROR_NO_DATA;
+    #endif
+}
+
+#if MODULE_VL53L1X_ST_API
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+int vl53l1x_read_details (const vl53l1x_t *dev,
+                          VL53L1_RangingMeasurementData_t *details)
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    EXEC_RET(vl53l1x_data_ready(dev));
+    EXEC_RET(VL53L1_GetRangingMeasurementData((VL53L1_Dev_t*)dev, details));
+    EXEC_RET(VL53L1_ClearInterruptAndStartMeasurement((VL53L1_Dev_t*)dev));
+
+    return VL53L1X_OK;
+}
+#endif /* MODULE_VL53L1X_ST_API */
+
+int vl53l1x_read_data (vl53l1x_t *dev, vl53l1x_data_t *data)
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+{
+    ASSERT_PARAM(dev != NULL);
+    ASSERT_PARAM(data != NULL);
+    DEBUG_DEV("data=%p", dev, data);
+
+    #if MODULE_VL53L1X_ST_API
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    VL53L1_RangingMeasurementData_t details;
+    EXEC_RET(vl53l1x_read_details(dev, &details));
+
+    data->status       = details.RangeStatus;
+    data->distance     = details.RangeMilliMeter;
+    data->signal_rate  = details.SignalRateRtnMegaCps;
+    data->ambient_rate = details.AmbientRateRtnMegaCps;
+
+    #else /* MODULE_VL53L1X_ST_API */
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1292 @@
...
+
+    /* read the results into data structure of type _vl53l1x_raw_data_t */
+    _vl53l1x_raw_data_t raw;
+    EXEC_RET(_read(dev, VL53L1X_RESULT__RANGE_STATUS,
+                        (uint8_t*)&raw, sizeof(_vl53l1x_raw_data_t)));
+
+    /* Copyrighted POLOLU / ST API code begin */
+
+    #if !MODULE_VL53L1X_BASIC
```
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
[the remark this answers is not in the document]
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with commit https://github.com/RIOT-OS/RIOT/pull/10363/commits/71fd9adb7b92b9a63b3b84f731b7cdec408065c3.

drivers/vl53l1x/include/vl53l1x_params.h
```diff
@@ -0,0 +1,113 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @name    Set default configuration parameters
+ * @{
+ */
+#ifndef VL53L1X_PARAM_DEV
+/** device is I2C_DEV(0) */
```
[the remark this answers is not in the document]
gschorcht: A single line in C style is a valid doxygen comment block for brief description. Is it really useful to bloat every very small comment to three lines to document a one-line definition? Postfix style is often not possible with the 100 character limit or the preferred 80 characters.

Usually I would prefer not to document these definitions, but doxygen insists on them

drivers/include/vl53l1x.h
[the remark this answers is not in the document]
gschorcht: Fixed with https://github.com/RIOT-OS/RIOT/pull/10363/commits/7c96913c76cd0c89e4b2bcaf66e337e79e7fe63d.

drivers/vl53l1x/vl53l1x.c
gschorcht: Fixed with https://github.com/RIOT-OS/RIOT/pull/10363/commits/7c96913c76cd0c89e4b2bcaf66e337e79e7fe63d.

drivers/vl53l1x/vl53l1x.c
gschorcht: Hm, `dev` is const in most functions which isn't a problem when calling ST API function where `dev` isn't const by casting it.

drivers/vl53l1x/vl53l1x.c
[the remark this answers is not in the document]
gschorcht: The driver is quite old and was written before we used these macros.

maribu:
I hope that I can resume review on Tuesday. Feel free to squash before then, I will do a fresh review anyway.

gschorcht:
@maribu The rework has been completed.

aabadie:
Do you think that driver would work (with some little modifications) on the vl53l**5** that is available on the ST b-u585i-iot02a board ?

gschorcht:
> Do you think that driver would work (with some little modifications I guess) on the vl53l**5** that is available on the ST b-u585i-iot02a board ?

I don't know, let me take a look.

gschorcht:
> > Do you think that driver would work (with some little modifications I guess) on the vl53l**5** that is available on the ST b-u585i-iot02a board ?
> 
> I don't know, let me take a look.

From the first look to the drivers, I would say that the sensor seems to be different. Since there is almost no other information about the internal register structure than the drivers, I would try to use the vendor driver as package.

gschorcht:
> I hope that I can resume review on Tuesday. Feel free to squash before then, I will do a fresh review anyway.

@maribu Is there progress in reviewing :smile: ?

maribu:
I'm still on vacation and reviewing larger PRs from my smartphone is cumbersome, also I cannot do testing. But it has a prominent position in my todo list ;)

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1316 @@
...
+int vl53l1x_set_timing_budget(vl53l1x_t *dev, uint32_t budget_us)
+{
+    ASSERT_PARAM(dev != NULL);
+    DEBUG_DEV("budget_us=%u", dev, budget_us);
+
+#if IS_USED(MODULE_VL53L1X_ST_API)
+
+    EXEC_RET(VL53L1_SetMeasurementTimingBudgetMicroSeconds(&dev->dev,
+                                                           budget_us));
+#else
```
maribu, line 527: Is there a chance that this would work as C conditionals? That way we can be sure that both ST-API and pololu variant can be compiled without having to provide two test applications. (More instances above and below.)

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1316 @@
...
+    unsigned timeout = VL53L1X_BOOT_TIMEOUT;
+    uint8_t res;
+    while (timeout--) {
+        if ((_read_byte(dev, VL53L1X_FIRMWARE__SYSTEM_STATUS, &res) == VL53L1X_OK) &&
+            (res & 0x01) == 1) {
+            break;
+        }
+        ztimer_sleep(ZTIMER_MSEC, 1);
+    }
+    if (timeout == 0) {
```
maribu, line 989:
```suggestion
    if (timeout == UINT_MAX) {
```

The postdecrement will still kick in, see:

```
~ $ cat test.c
#include <stdio.h>

int main(void) {
    unsigned foo = 3;
    while (foo--) { }
    printf("foo = %u\n", foo);

    return 0;
}
~ $ gcc -o test test.c
~ $ ./test
foo = 4294967295
```

But IMO it would be more readable to just go for

```C
while (timeout) {
    timeout--; 
    ...
```

instead.

maribu (commented):
I only did a quick look, but code looks good so far.

drivers/vl53l1x/vl53l1x.c
gschorcht: Hm, there are a lot of defines and symbol mappings dependent on whether the ST API is used or not that I can't imaging that compilation works if we use C conditionals. I can try it, but I'm not optimistic.

drivers/vl53l1x/vl53l1x.c
gschorcht: > The postdecrement will still kick in, see:

Yeah of course. Good catch. Fixed.

drivers/vl53l1x/vl53l1x.c
gschorcht: I think there is no chance in this case. The problem is that the ST API is used as a package, so of course the header files with a lot of defines and declarations required during compilation are not available if the package is not used by the driver:
```python
drivers/vl53l1x/vl53l1x.c:525:14: error: implicit declaration of function 'VL53L1_SetMeasurementTimingBudgetMicroSeconds' [-Werror=implicit-function-declaration]
     EXEC_RET(VL53L1_SetMeasurementTimingBudgetMicroSeconds(&dev->dev,
              ^
```
To solve this problem we would have to pull the package in any case and insert the headers always, even if they are not used by the driver. IMHO, that would be the wrong approach just to use C Conditionals.

Even if we would do it in that way, there would raise another compilation problem because the driver defines some mappings to be able to use same code independent on whether the ST API is used or not. These mappings would result in a conflict if the ST API is used.

I fully agree that the CI coverage is an argument to use C conditionals. But in this case I would prefer to keep the preprocessor conditionals and to provide a second test application. Would it be ok for you?

gschorcht:
Rebased another time to fix the conflicts and updated to be compatible with the `tests/drivers` directory structure.

drivers/include/vl53l1x.h
```diff
@@ -0,0 +1,599 @@
...
+ * information on how to use the API to calibrate the sensor.
+ *
+ * @note    Since the ST STSW-IMG007 VL53L1X API package is quite complex and
+ *          very large, this driver variant should only be used when memory
+ *          requirements are not an issue.
+ *
+ * @warning The ST STSW-IMG007 VL53L1X API package functions use a significant
+ *          amount of memory in the stack. If you have crashes due to memory
+ *          access errors, try increasing the default thread stack size
+ *          \ref THREAD_STACKSIZE_DEFAULT.
```
crasbe, line 60: Maybe you can give an estimation on how big the stack size should be.

drivers/include/vl53l1x.h
```diff
@@ -0,0 +1,599 @@
...
+                                                                    around check has not been
+                                                                    done */
+#define VL53L1X_RANGESTATUS_WRAP_TARGET_FAIL               7   /**< wrapped target, no matching
+                                                                    phase in other VCSEL period
+                                                                    timing */
+#define VL53L1X_RANGESTATUS_XTALK_SIGNAL_FAIL              9   /**< specific to lite ranging */
+#define VL53L1X_RANGESTATUS_SYNCRONISATION_INT             10  /**< first interrupt when starting
+                                                                    ranging in back to back mode */
+#define VL53L1X_RANGESTATUS_MIN_RANGE_FAIL                 13  /**< user ROI input is not valid */
+#define VL53L1X_RANGESTATUS_NONE                           255 /**< No Update. */
```
crasbe, line 157: You can also put the documentation in the previous line, that would save some line breaks I guess.

drivers/include/vl53l1x.h
```diff
@@ -0,0 +1,599 @@
...
+    VL53L1X_THRESH_IN   = 3, /**< when value exceeds the lower threshold or
+                                  falls below the higher threshold (the value
+                                  enters the threshold window) */
+} vl53l1x_thresh_mode_t;
+
+/**
+ * @brief   Interrupt mode
+ *
+ * It defines when an interrupt is triggered. Interrupts can be triggered
+ * either on new data in each measurement cycly or when the threshold condition
```
crasbe, line 228:
```suggestion
 * either on new data in each measurement cycle or when the threshold condition
```

drivers/include/vl53l1x.h
```diff
@@ -0,0 +1,599 @@
...
+                             #GPIO_UNDEF if not used */
+#if !IS_USED(MODULE_VL53L1X_BASIC)
+    gpio_t pin_shutdown;/**< Shutdown pin low active:
+                             #GPIO_UNDEF if not used */
+    uint32_t budget;    /**< Timing budget given to the sensor to perform range
+                             measurements in us [20000 ... 1000000],
+                             default: 50 ms */
+    uint32_t period;    /**< Inter-measurement period in ms, has to be
+                             vl53l1x_params_t::t_budget + 4ms,
+                             default: 100 ms */
```
crasbe, line 283: But the budget is 50ms, and 50+4ms != 100ms 🤔

drivers/include/vl53l1x.h
```diff
@@ -0,0 +1,599 @@
...
+ * Please refer the [VL53L1X API user manual]
+ * (https://www.st.com/resource/en/user_manual/dm00474730.pdf)
+ * section 2.6 for detailed information about the meaning of detailed
+ * ranging data.
+ *
+ * @note
+ * - Since reading any data resets the interrupt as well as the data-ready
+ *   flag, either function #vl53l1x_read_mm, function #vl53l1x_read_data, or
+ *   function #vl53l1x_read_details can be used in one measurement cycle.
+ * - This function is ONLY available when module \ref vl53l1x_basic is used.
```
crasbe, line 425:
```suggestion
 * - This function is ONLY available when module \ref vl53l1x_st_api is used.
```

However this part of the note is redundant, because in the normal description block, the same information is already given. So you can also remove the sentence I guess.

drivers/saul/init_devs/auto_init_vl53l1x.c
```diff
@@ -0,0 +1,72 @@
...
+
+        saul_entries[i].dev = &(vl53l1x_devs[i]);
+        saul_entries[i].name = vl53l1x_saul_info[i].name;
+        saul_entries[i].driver = &vl53l1x_saul_driver;
+        saul_reg_add(&(saul_entries[i]));
+    }
+}
+
+#else
+typedef int dont_be_pedantic;
```
crasbe, line 71: What is this about? 🤔

drivers/vl53l1x/Kconfig
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2020 HAW Hamburg
+#
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+#
```
crasbe, line 6: Copy&Paste Leftover?

drivers/vl53l1x/include/vl53l1x_params.h
```diff
@@ -0,0 +1,148 @@
...
+ * @name    Default hardware configuration parameters
+ * @{
+ */
+
+/**
+ * @brief    Default device is I2C_DEV(0)
+ */
+#ifndef VL53L1X_PARAM_DEV
+#define VL53L1X_PARAM_DEV           I2C_DEV(0)
+#endif
```
crasbe, line 34:
```suggestion
#ifndef VL53L1X_PARAM_DEV
#  define VL53L1X_PARAM_DEV         I2C_DEV(0)
#endif
```

Also for the rest of the file.

drivers/vl53l1x/vl53l1x.c
```diff
@@ -0,0 +1,1316 @@
...
+ *
+ * For more information, see
+ *
+ * https://www.pololu.com/
+ * https://forum.pololu.com/
+ *
+ * ---------------------------------------------------------------------------
+ *
+ * Copyright (c) 2017, STMicroelectronics
+ * Copyright (c) 2018-2022, Pololu Corporation
```
crasbe, line 30: Would it make sense to add that to the SPDX header?

tests/drivers/vl53l1x_st_api/main.c
```diff
@@ -0,0 +1,111 @@
...
+#include <stdio.h>
+
+#include "mutex.h"
+#include "thread.h"
+#include "ztimer.h"
+
+#include "vl53l1x.h"
+#include "vl53l1x_params.h"
+
+static void isr (void *arg)
```
crasbe, line 38:
```suggestion
static void isr(void *arg)
```

tests/drivers/vl53l1x/main.c
```diff
@@ -0,0 +1,159 @@
...
+#include <stdio.h>
+
+#include "mutex.h"
+#include "thread.h"
+#include "ztimer.h"
+
+#include "vl53l1x.h"
+#include "vl53l1x_params.h"
+
+static void isr (void *arg)
```
crasbe, line 54:
```suggestion
static void isr(void *arg)
```

tests/drivers/vl53l1x/main.c
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * If the configuration parameter VL53L1X_PARAM_PIN_INT is defined, interrupts
+ * are used to get data instead of polling for new data. In the case of driver
+ * variant vl53l1x_st_api, threshold interrupts are configured. Otherwise
+ * only data interrupts are used.
+ *
+ * In all cases, the sensor is configured with following default parameters:
+ *
+ * - timing budget of 50 ms
+ * - intermeasurement period of 100 ms
```
crasbe, line 42: Not sure if it makes sense to replicate the content of the README.md here?

tests/drivers/vl53l1x/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+| Driver Module    | Description                                       |
+|------------------|---------------------------------------------------|
+| `vl53l1x`        | Standard driver with most functionality           |
+| `vl53l1x_basic`  | Basic driver with only very basic functionality   |
+| `vl53l1x_st_api` | ST VL53L1X API driver with complete functionality |
+
+## Usage
+
+The driver variant used is defined by my variable `DRIVER`, which is set to
```
crasbe, line 16: IMO the "defined by ***my*** variable" sounds a bit weird.

drivers/vl53l1x/vl53l1x_saul.c
```diff
@@ -0,0 +1,33 @@
+/*
+ * SPDX-FileCopyrightText: 20215 Gunar Schorcht
```
crasbe, line 2:
```suggestion
 * SPDX-FileCopyrightText: 2025 Gunar Schorcht
```

drivers/vl53l1x/include/vl53l1x_register_map.h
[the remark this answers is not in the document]
crasbe: I support this idea, that would instantly reduce the Lines of Code from 14k to 3k.

tests/drivers/vl53l1x/main.c
crasbe: I think this is also the case for other .c files.

gschorcht:
> I guess this is still a bit WIP at the moment, but here are some comments. There are still some old Copyright notices, I saw that you are/were in the progress of migrating them to the new format.

Oh sorry, I forgot to set the WIP label. I picked up this PR after it was not touched for 4 years and just started with rebasing it. Now, I'm going through step by step to see what have to be done do update this PR. It#s definitly too early to review it for the moment.

gschorcht:
Closed because of a complete rework in PR #21648.

PR #10372, opened 2018-11-12
WIP: pm_layered: print block/unblock

toonst:
can be used to debug or as a basis for a pm test. 

### Contribution description
This still has a problem: `LOG_INFO` is called from within `pm_block` in the `uart_init`, making the test app block. Suggestions are welcome.

### Testing procedure

Run tests/periph_pm.

### DIFF
--- a/sys/include/pm_layered.h
+++ b/sys/include/pm_layered.h
@@ -54,14 +54,24 @@ extern "C" {
  *
  * @param[in]   mode      power mode to block
  */
+#ifdef PM_PRINT
+void _pm_block(unsigned mode, const char *caller);
+#define pm_block(x) _pm_block((x), __func__)
+#else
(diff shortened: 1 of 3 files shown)


kaspar030:
Hm. I can see that this is very useful, but it feels very macro-hacky, and also like a generically useful mechanism is put into pm_layered.

Could you live with call-site addresses (instead of names)? Maybe ```LOG_DEBUG("%s: called from %p, ...\n", __func__, __builtin_return_address(), ...);``` at the top of pm_block/unblock serves the same purpose, with less hackery. The LOG_DEBUG level is below default (e.g., silent by default). It should be possible to enable that level just for pm_layered by adding "CFLAGS += -DLOG_LEVEL=LOG_DEBUG" to ```sys/pm_layered/Makefile```.

Issue #8107, opened 2017-11-22
crypto/ccm: bugs in the implementation of CCM mode


cladmi:
This issue summarize all the issues I have found when trying to add tests for #6706 for `sys/crypto/modes/ccm.c`

Issues shown in the PR:

* Upper bound for `auth_data_len` should be `0xFEFF`. https://tools.ietf.org/html/rfc3610 2.2 page 
* Wrong check for input plaintext message length against len encoding.
* buffer overflow if auth_data_len is larger than 26

Problems found when adding tests:
* In tests 'len_encoding' and 'mac_length' are hardwritten.
* Tests do not support `input_len` >= 256
* ccm code will break if length_encoding >= 4 (even with PR fix because of overflow).
* Code is not saving enough bytes when saving plaintext_length (off by one limit)
* CCM is broken with `input_len` >= 256 (so broken even when fixing previous point)

- [x] Refactor test to allow running other tests (https://github.com/RIOT-OS/RIOT/pull/8109)
- [x] Fix and add test for input_len check (https://github.com/RIOT-OS/RIOT/pull/8112)
- [x] Fix and add test for auth_data_len upper bound (https://github.com/RIOT-OS/RIOT/pull/8114)
- [x] Fix and add test for auth_data_len > 26
- [ ] Fix and add test for auth_data_len = 0
- [x] Fix and test for input_len >= 256
- [x] Test working on 8bit/16bit platforms…
- [x] Add all RFC/NIST tests https://github.com/RIOT-OS/RIOT/pull/12038
    - [x] Make the api take const inputs https://github.com/RIOT-OS/RIOT/pull/10376
    - split `tests-crypto` out of `unittests` (part of #12038)

This is to track and fix all what was done in https://github.com/RIOT-OS/RIOT/pull/7968.
The PRs will use and replace https://github.com/RIOT-OS/RIOT/pull/6706

fjmolinas:
@aabadie can we check the 8bit/16bit point since #12645 is in?

PR #7968, opened 2017-11-08
[FIXME][testbranch] crypto: RFC3610 tests not all passing

cladmi:
I wanted to test the following pull request https://github.com/RIOT-OS/RIOT/pull/6706 and could not get same output for one personal test vector I wrote compared to pycrypto (master branch).

I tried using all rfc3610 tests and discovered that nor the pr branch, nor the master version pass all of them.

This branch adds all these tests from the RFC, (vector i is called TEST i + 10) and adds some debug output. Its not meant to be merged as is but more as a test branch.
I also included https://github.com/RIOT-OS/RIOT/pull/6706 commit to test its version.

From what I see, the errors are located in the final part of the result so calculated from the output of ccm_compute_cbc_mac.

To get the tests files I used https://github.com/dlitz/pycrypto/blob/master/lib/Crypto/SelfTest/Cipher/test_AES.py#L1489 inputs. I included the script to generate the tables.
I only checked that their test for "Packet Vector 7" correctly matches the RFC because it is the first broken test on our suite. So I just expect their tests to be valid.

### Related issue

Tracking Issue https://github.com/RIOT-OS/RIOT/issues/8107

PR #10376, opened 2018-11-12
crypto/modes/ccm: update api to const input buffers

cladmi:
### Contribution description

This changes the `cipher_encrypt_ccm` and `cipher_decrypt_ccm` functions to declare input buffers as `const`, and also the required dependencies to do it.

The tests have been updated to use `const` buffers too.

It is a API change but it is compatible as it is just adding more compatibility except if you use the function signature as the tests did.
Not sure if it requires an "API change" label in that case, I let reviewers remove it if they think it is not required.

### Testing procedure

Run the tests on a board that was working before and also compare the size difference

```
BOARD=samr21-xpro RIOT_VERSION_OVERRIDE=buildtest make  -C tests/unittests/ tests-crypto  flash test
```

The size difference I got: It removes `200` bytes ram and only adds `64` rom.

```
# master
  29292     948    2696   32936    80a8 /home/harter/work/git/RIOT/tests/unittests/bin/samr21-xpro/tests_unittests.elf

# const inputs
  29356     748    2696   32800    8020 /home/harter/work/git/RIOT/tests/unittests/bin/samr21-xpro/tests_unittests.elf
```

### Issues/PRs references

Found in https://github.com/RIOT-OS/RIOT/pull/7968#issuecomment-437962046 which is part of https://github.com/RIOT-OS/RIOT/issues/8107 resolution.

### DIFF
--- a/sys/crypto/modes/ccm.c
+++ b/sys/crypto/modes/ccm.c
@@ -34,8 +34,8 @@ static inline int min(int a, int b)
     }
 }
 
-int ccm_compute_cbc_mac(cipher_t* cipher, uint8_t iv[16],
-                        uint8_t* input, size_t length, uint8_t* mac)
+int ccm_compute_cbc_mac(cipher_t* cipher, const uint8_t iv[16],
+                        const uint8_t* input, size_t length, uint8_t* mac)
 {
     uint8_t offset, block_size, mac_enc[16] = {0};
 
@@ -64,7 +64,7 @@ int ccm_compute_cbc_mac(cipher_t* cipher, uint8_t iv[16],
 
 
 int ccm_create_mac_iv(cipher_t* cipher, uint8_t auth_data_len, uint8_t M,
-                      uint8_t L, uint8_t* nonce, uint8_t nonce_len,
+                      uint8_t L, const uint8_t* nonce, uint8_t nonce_len,
                       size_t plaintext_len, uint8_t X1[16])
 {
     uint8_t M_, L_;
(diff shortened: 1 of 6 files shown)


sys/crypto/modes/ccm.c
```diff
@@ -35,7 +35,7 @@ static inline int min(int a, int b)
 }
 
 int ccm_compute_cbc_mac(cipher_t* cipher, uint8_t iv[16],
```
bergzand: The `iv[16]` here can also be `const`. Do you mind piggy backing that or do you think that it is unnecessary optimization here?

cladmi: I only put the minimal ones that were required to make my tests pass, I did not even tried to understand :p
But sure, I will add it.

bergzand: I didn't really check if it makes sense here, I just noticed that it wasn't modified by the function :)

cladmi: Just to be sure, can it be an issue, in general case, for the called function that `iv` and `mac` can be the same pointer when `iv` is said const ?

Here as it is an internal function, it could even be replaced by only one "output" pointer that should contain the "iv", but that is another topic.

bergzand:
Ack, please squash.

While technically an API change, this allows a superset of the current functionality, so I don't consider it problematic.

cladmi:
Rebased on master, and squashed.

Since https://github.com/RIOT-OS/RIOT/pull/10406 was merged, the reference size was reduced a lot, but the diff stays the same.

With the PR
```
  21556     748    2696   25000    61a8 /home/harter/work/git/RIOT/tests/unittests/bin/samr21-xpro/tests_unittests.elf
```
In master

```
  21492     948    2696   25136    6230 /home/harter/work/git/RIOT/tests/unittests/bin/samr21-xpro/tests_unittests.elf
```

I could successfully execute the same test commands with `samr21-xpro` after rebase.

cladmi:
Thank you both for the review.

PR #8114, opened 2017-11-22
crypto/ccm: fix auth_data_len check

cladmi:
This PR fixes crypto/ccm checking `auth_data_len` against wrong value and add a test to it.

> RFC3610 page 2
> If 0 < l(a) < (2^16 - 2^8), then the length field is encoded as two octets.

To allow testing, I also needed to make `cipher_encrypt_ccm` and `cipher_decrypt_ccm` check for `ccm_compute_adata_mac` return value which was ignored.
I only added a test for encrypt not decrypt.

It is part of solving issue #8107 and includes/replaces a part of #6706

It depends on pull request ~~#8109~~ and ~~#8112~~ (to fix input_len check for len_encoding)

PR #8112, opened 2017-11-22
crypto/ccm: fix input_len checking with length_encoding

cladmi:
This PR fixes crypto/ccm wrongly checking `input_len` with `length_encoding` and adds tests for it.

Copied from https://github.com/RIOT-OS/RIOT/pull/6706
> The first bug (appeared in both cipher_encrypt_ccm and cipher_decrypt_ccm functions) incorrectly
> checks the size of the input plaintext messages against the length encoding L. As per RFC3610
> Section 2.1, the requirement on the input size is that 0 <= l(m) < 2^(8L), where l(m) represents
> the size of the plaintext message. This patch computes 2^(8L) correctly and fixes the logic for checking input_len.

I also found that checking input_len was broken if `length_encoding >= 4` so I added a fix for that.

It is part of solving issue #8107 and includes/replaces a part of https://github.com/RIOT-OS/RIOT/pull/6706

It depends on pull request ~~https://github.com/RIOT-OS/RIOT/pull/8109~~

PR #8109, opened 2017-11-22
tests-crypto-modes-ccm: Refactoring to support more evolved tests

cladmi:
This PR makes tests-crypto-modes-ccm tests more configurable and able to support other cases:

* Different `mac_length`
* Different `nonce_len`
* Tests with `input/adata` bigger than 256

Simple test refactoring to make the next PRs easier to review.

It is part of solving issue https://github.com/RIOT-OS/RIOT/issues/8107

PR #6706, opened 2017-03-07
crypto: fix bugs in the implementation of CCM mode

wentaoshang:
This PR fixes two bugs in the implementation of CCM authenticated encryption in the crypto library. 

The first bug (appeared in both `cipher_encrypt_ccm` and `cipher_decrypt_ccm` functions) incorrectly checks the size of the input plaintext messages against the length encoding L. As per RFC3610 Section 2.1, the requirement on the input size is that `0 <= l(m) < 2^(8L)`, where `l(m)` represents the size of the plaintext message. This patch computes `2^(8L)` correctly and fixes the logic for checking `input_len`.

The second bug in `ccm_compute_adata_mac` line 121 will cause buffer overflow if `auth_data_len` is larger than 26. This patch fixes the issue by separating the MAC computation over `auth_data` into two steps: the first step computes the MAC over the initial block of `auth_data` prepended with length encoding; the second step computes the MAC over the rest of `auth_data`.

PR #10377, opened 2018-11-12
crypto/ccm: fix auth_data_len check [backport 2018.10]

cladmi:
# Backport of #8114

This PR fixes crypto/ccm checking `auth_data_len` against wrong value and add a test to it.

> RFC3610 page 2
> If 0 < l(a) < (2^16 - 2^8), then the length field is encoded as two octets.

To allow testing, I also needed to make `cipher_encrypt_ccm` and `cipher_decrypt_ccm` check for `ccm_compute_adata_mac` return value which was ignored.
I only added a test for encrypt not decrypt.

It is part of solving issue #8107 and includes/replaces a part of #6706

It depends on pull request ~~#8109~~ and ~~#8112~~ (to fix input_len check for len_encoding)

### DIFF
--- a/sys/crypto/modes/ccm.c
+++ b/sys/crypto/modes/ccm.c
@@ -167,7 +167,10 @@ int cipher_encrypt_ccm(cipher_t* cipher, uint8_t* auth_data, uint32_t auth_data_
     }
 
     /* MAC calulation (T) with additional data and plaintext */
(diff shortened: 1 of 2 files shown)


cladmi:
I will quickly retest the rebased branch.

jia200x:
I also confirm it works

PR #10385, opened 2018-11-14
tests/puf_sram: add input args to automation script

PeterKietzmann:
### Contribution description

This PR adds optional input arguments to the automation script for estimating the minimum entropy of the SRAM based random seed (*tests/puf_sram*). It is needed to change the serial port or baudrate when changing to platforms with different serial parameters.

### Testing procedure

- Calling the script without parameters should lead to the same behavior as before
- Calling script help (`example_test.py -h`) should show information about input parameters
```
usage: example_test.py [-h] [-n NUMBER] [-t OFF_TIME] [-p PORT] [-b BAUDRATE]
                       [-d]

optional arguments:
  -h, --help            show this help message and exit
  -n NUMBER, --number NUMBER
                        Number of iterations, default: 500
  -t OFF_TIME, --off_time OFF_TIME
                        Off time, default: 1 [s]
  -p PORT, --port PORT  Serial port, default: /dev/ttyUSB0
  -b BAUDRATE, --baudrate BAUDRATE
                        Baudrate of the serial port, default: 115200
  -d, --disable_output  Disable verbose output
```
- `example_test.py -n 3 -t 10` should reduce the number of iterations to 3 and the power-off time before restart to 10 seconds
- - `example_test.py -d` disables intermediate status reports but the board reboots anyway (status LEDs turn of repeatedly)
- Changing the `-p` or `-b` parameters doesn't make sense right now, as these parameters are fixed for boards

### Issues/PRs references
n/a

PR #10390, opened 2018-11-14
cpu/atmega_common: add mega2560 puf_sram feature 

PeterKietzmann:
### Contribution description

This PR adds the SRAM based random seed generation feature for mega2560 8-bit platforms.

```
Number of seeds: 500       
Seed length    : 32 Bit   
Abs. Entropy   : 30.04 Bit   
Rel. Entropy   : 93.89 perc. 
```
### Testing procedure
- flash *tests/puf_sram* to your arduino-mega2560
- follow instructions in the respective README.md
- Call the example script like `python tests/example_test.py -b 9600`

### Issues/PRs references
based on #10385, #10386

### DIFF
--- a/boards/arduino-mega2560/Makefile.features
+++ b/boards/arduino-mega2560/Makefile.features
@@ -1,3 +1,4 @@
+FEATURES_PROVIDED += puf_sram
 include $(RIOTBOARD)/common/arduino-atmega/Makefile.features
 
 include $(RIOTCPU)/atmega2560/Makefile.features
--- a/cpu/atmega_common/include/cpu_conf.h
+++ b/cpu/atmega_common/include/cpu_conf.h
@@ -51,6 +51,11 @@ extern "C" {
 #define THREAD_STACKSIZE_IDLE      (128)
 /** @} */
 
+/**
+ * @brief   Attribute for memory sections required by SRAM PUF
+ */
+#define PUF_SRAM_ATTRIBUTES __attribute__((used, section(".noinit")))
+
 #ifdef __cplusplus
 }
 #endif
--- a/cpu/atmega_common/startup.c
+++ b/cpu/atmega_common/startup.c
@@ -25,6 +25,9 @@
 
 /* For Catchall-Loop */
 #include "board.h"
+#ifdef MODULE_PUF_SRAM
+#include "puf_sram.h"
+#endif
 
 /**
  * @brief functions for initializing the board, std-lib and kernel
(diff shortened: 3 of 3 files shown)


ZetaR60:
Interesting method in ```puf_sram```! I have a few concerns however:
- Changes adding support to ATmega2560 are combined with changes to ```cortexm_common``` and to puf_sram and its test, which all affect different things. Is this just because you have added the commits from the dependency PRs to this PR? I usually separate them out on my PRs so that is easier to see which changes go to which PR.
- Is there a reason why only ATmega2560 is supported, and not the other ATmegas? It doesn't look like it depends on any specific hardware.

Also, while this might not be the best place to discuss this, I wonder if support should be added to new platforms without more careful qualification of SRAM properties, since it relies on undocumented properties of the manufacturing process:
1. If the transistors have tight tolerances, and each state is well balanced, then the state of the flip-flip will be based primarily on noise. I assume this is what we want.
2. If the transistors have loose tolerances, then it is more likely that each flip-flop will almost always initialize high or low (with only a few being well balanced). This would be good for a static CPU ID, but would result in low entropy as a random seed.
3. The flip-flop could have a significant bias due to either circuit design or flip-flop design. The flip-flops could have an intentional bias (any compelling reasons to do this?), or the sides could be asymmetrically loaded, or there could be asymmetric EMI due to nearby traces.

The problem is that as written there is no way to determine the actual amount of entropy from rebooting, or from manufacturing tolerances, or the lack of entropy do to some systematic bias. I realize this is experimental, but not tackling this early could lead to an oversight with security implications down the road. I think there should be at least some bigger warnings on what exactly the issues are.

PeterKietzmann:
@ZetaR60 thanks for your feedback! The changes to *cortex* and *tests* are part of the above linked PRs which I wanted to get merged before.
I agree that this feature should not be added without qualification of SRAM properties. That's why I only enabled it for the mega2560 so far, which I tested. In deployment scenarios each device or at least device charge should be evaluated. 
In #9290 we merged support for cortex devices which included a [compile time warning](https://github.com/RIOT-OS/RIOT/blob/master/sys/puf_sram/Makefile). We could improve that?
At the last summit I presented memory analysis in a bit more detail ([slides](http://summit.riot-os.org/2018/wp-content/uploads/sites/10/2018/09/3_2-Peter-Kietzmann-Crypto-Fundamentals.pdf)). The TODOs are still ongoing and a more detailed memory evaluation tool as well as a secure ID (key) generation mechanism are part of it.

PeterKietzmann:
Rebased and re-tested on the arduio-mega2560 with *tests/puf_sram*:
```
python tests/example_test.py -n 500 -t 1 -p /dev/ttyUSB0 -b 9600
```
Results look ok:
```
Number of seeds: 500       
Seed length    : 32 Bit   
Abs. Entropy   : 30.04 Bit   
Rel. Entropy   : 93.88 perc. 
```

PR #10393, opened 2018-11-15
Improve and debug Xtimer

Josar:
This PR is a stack of commits which depends upon each other and are a rework of some things done in https://github.com/RIOT-OS/RIOT/pull/8990.

1. Tries to handle the same problem as https://github.com/RIOT-OS/RIOT/pull/7629 and https://github.com/RIOT-OS/RIOT/pull/9595 .

2. _timer_callback changed to only check once for an overflow. Avoiding early firing of xtimer when _next_period is called 2 times. Encountered when testing with xtimer_drift.

3. xtimer_hang test schould use xtimer_periodic_wakeup instead of sleep. This will ensure that all percentage values are printed.

4. Using multiple pins to distinguish the threads.

5. Ensure that worker_thread is higher prior than slacker_thread.

6. _time_left was not able to calculate the time left for a target which is at the beginning of the next period, but has to be fired in this period because of XTIMER_OVERHEAD.

7. Xtimer where fired to early when running xtimer_drift. Which was caused by a overflowing llcounter. As the reference was set to now. The function _time_left calculated the wrong time to wait.

See the commit messages for more details.

Following Test run with good results on the jiminy rfr2
``` 
xtimer_usleep_short
xtimer_usleep
xtimer_reset
xtimer_remove
xtimer_periodic_wakeup
xtimer_msg_receive_timeout
xtimer_hang
xtimer_drift
```

### DIFF
--- a/sys/xtimer/xtimer.c
+++ b/sys/xtimer/xtimer.c
@@ -73,27 +73,42 @@ void _xtimer_tsleep(uint32_t offset, uint32_t long_offset)
 void _xtimer_periodic_wakeup(uint32_t *last_wakeup, uint32_t period) {
     xtimer_t timer;
     mutex_t mutex = MUTEX_INIT;
-
-    timer.callback = _callback_unlock_mutex;
-    timer.arg = (void*) &mutex;
+    uint32_t mult;
 
     uint32_t target = (*last_wakeup) + period;
     uint32_t now = _xtimer_now();
+
     /* make sure we're not setting a value in the past */
     if (now < (*last_wakeup)) {
-        /* base timer overflowed between last_wakeup and now */
-        if (!((now < target) && (target < (*last_wakeup)))) {
-            /* target time has already passed */
-            goto out;
+        /* last_wakeup < target, now overflowed but target not, target passed.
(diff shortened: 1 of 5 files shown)


cladmi:
As I am really not focused on timer, it is hard for me to concentrate on PRs changing many things as it takes a lot of time to review.

It would simplify my task if you could split small parts, still keep the main PR as a reference, but focus on one small part to fix first. The lower level it is, the better as it solves the ground problems first so we can build on top.

Josar:
@cladmi i split out the test and debug related commits to distinct PRs with the intention to take care of the crucial timer related problems after they have been merged. see: https://github.com/RIOT-OS/RIOT/pull/9595 https://github.com/RIOT-OS/RIOT/pull/10173 https://github.com/RIOT-OS/RIOT/pull/10396 https://github.com/RIOT-OS/RIOT/pull/10398

Josar:
@haukepetersen as posted above, there is still a lot to get into master until we can target the issues the xtimer has and which get observeable on 16bit platforms. Maybe someone will have time to dig into.

PR #10397, opened 2018-11-15
test/xtimer hang: separate debug pins

Josar:
Previously one pin was used for both slacker threads.
For better debugging seperating the threads to use one pin each is better.

### DIFF
--- a/tests/xtimer_hang/main.c
+++ b/tests/xtimer_hang/main.c
@@ -20,6 +20,7 @@
  * @author      Daniel Krebs <github@daniel-krebs.net>
  * @author      Kaspar Schleiser <kaspar@schleiser.de>
  * @author      Sebastian Meiling <s@mlng.net>
+ * @author      Josua Arndt <jarndt@ias.rwth-aachen.de>
  *
  * @}
  */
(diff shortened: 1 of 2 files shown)


tests/xtimer_hang/main.c
```diff
@@ -65,8 +79,8 @@ int main(void)
 #endif
 
     LOG_DEBUG("[INIT]\n");
-    uint32_t sleep_timer1 = 1000;
-    uint32_t sleep_timer2 = 1100;
+    uint32_t sleep_timer1 = TEST_SLEEP_TIME_1;
```
A-Paul: Thumbs up for eliminating literal numbers. :-)

tests/xtimer_hang/main.c
```diff
@@ -38,20 +38,34 @@
...
 #endif
     LOG_DEBUG("run thread %" PRIkernel_pid "\n", thread_getpid());
-    while(1) {
-#if defined(WORKER_THREAD_PIN)
-        gpio_set(worker_pin);
-        gpio_clear(worker_pin);
+
+    while (1) {
+#if defined(WORKER_THREAD_PIN_1) && defined(WORKER_THREAD_PIN_2)
+        if (*(uint32_t *)(arg) == TEST_SLEEP_TIME_1) {
```
A-Paul: I think a *switch case* might be a cleaner approach here.

A-Paul:
@Josar, look at this [Patch](https://gist.github.com/A-Paul/324fb41f7e5b82523d283db4357f1c44#file-468b5a76488f5c72b2dd0721e21e1f753e531bbf-patch) as a suggestion.

Josar:
Nice Idea, but does not work. The case has to be in the while. As we overwrite the pin with the second use of the function.

grafik

Josar:
@A-Paul i pushed an updates Version.

Josar:
@A-Paul any time to have a second look?

Josar:
@A-Paul still np Time? @haukepetersen maybe you?

A-Paul:
Hi @Josar, please squash.

Josar:
@A-Paul sqaushed and rebased.

PR #10398, opened 2018-11-15
tests/xtimer_drift: increase priority of worker

Josar:
If the worker thread and the slacker thread have the same priority they will be handle after each other.
Setting the worker tread to a higher priority ensures that the printed jitter is the best achievable.

### DIFF
--- a/tests/xtimer_drift/main.c
+++ b/tests/xtimer_drift/main.c
@@ -163,7 +163,7 @@ int main(void)
 
     /* create and trigger worker thread */
     kernel_pid_t pid3 = thread_create(worker_stack, sizeof(worker_stack),
-                                      THREAD_PRIORITY_MAIN - 1,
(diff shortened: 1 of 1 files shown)


Josar:
Hello, A-Paul wouldn't sending it back to main and printing it there then be the best solution?

A-Paul:
> Hello, A-Paul wouldn't sending it back to main and printing it there then be the best solution?

When you think the slackers are busy enough, fine. :)

Josar:
@A-Paul thinking further it might be good to keep the printing in the worker, this also ensures that timing problems in xtimer or context switch problems can be detected.
I would agreed that keeping high prior task as short as possible is a good design choice but this tests are meant to provoke errors. And there might be times when higher prior task take longer than faster scheduled low prior task so this is a valid scenario, imho.

PR #10407, opened 2018-11-15
netstats: at86rf2xx: tx_bytes only includes 15.4. header, no payload

cgundogan:
### Contribution description

The values returned by the shell command `ifconfig` display a rather
small value for the actual sent out bytes on L2 level.

This commit moves the `tx_bytes` calculation further down, until `len`
includes the 15.4. header *and* the payload length (no FCS).

### Testing procedure

Observe the output of `ifconfig` using the `netstats_l2` module.


bergzand:
@cgundogan While this is a valid fix, do you mind if I close this one in favour of #9793

cgundogan:
>@cgundogan While this is a valid fix, do you mind if I close this one in favour of #9793



sure

PR #8601, opened 2018-02-20
netopt/drivers: Add phy link status output

bergzand:
### Contribution description

This PR adds netopt support to query the status of the phy link of a network driver, whether a cable is attached and the link is functional. Output for the ifconfig command to report this status is added.

### Issues/PRs references

None

### Todo

Other drivers beside the enc28j60:

- [ ] netdev_tap
- [x] encx24j600
~~w5100~~

PR #10415, opened 2018-11-16
drivers/encx24j600: Implemented missing drop case

maribu:
### Contribution description

The netdev_driver_t::recv implementation of the encx24j600 does not provide the drop feature. This PR adds it.

### Testing procedure

### Issues/PRs references

#10410

### DIFF
--- a/drivers/encx24j600/encx24j600.c
+++ b/drivers/encx24j600/encx24j600.c
@@ -373,7 +373,10 @@ static int _recv(netdev_t *netdev, void *buf, size_t len, void *info)
 #endif
         /* read packet (without 4 bytes checksum) */
         sram_op(dev, ENC_RRXDATA, 0xFFFF, buf, payload_len);
+    }
 
+    /* Frame was retrieved or drop was requested --> remove it from buffer */
+    if (buf || (len > 0)) {
         /* decrement available packet count */
         cmd(dev, ENC_SETPKTDEC);
 


aabadie:
Maybe @kaspar030 has the hardware and could test ? since he is the author of this driver.

jia200x:
I can ACK the fundamentals and code design: 

- buf not NULL => receive, flush and return packet size
- buf NULL && len == 0 => return packet size
- buf NULL && len > 0 => flush and return packet size.

jia200x:
nothing interesting to report on coding style here, and documentation doesn't apply

maribu:
@bergzand, @smlng and @miri64 also have worked on the driver. Maybe someone of them has the hardware and can test it?

maribu:
@bergzand, @smlng, @miri64, @kaspar030: Does someone of you have a device for testing?

Also: If there is so little interest in this driver that security fixes do not get merged due to lack of testing, maybe we should consider to rather delete the driver than to continue shipping the vulnerable code?

smlng:
I only have the enc28j60, sorry

bergzand:
Same here, only the enc28j60 :cry:

maribu:
How about we add a feature to the build system to warn about modules that are in desperate need of love? Maybe even with a warning, something like: "If no one stands up to help or at least speaks up to keep this module for three releases, this module is going to be deleted."

To me, this driver as well as a bunch of seemingly no longer used boards should be marked as such. If there is indeed no longer a user of that stuff, maintaining the code would be wasted effort.

(To me it is important that hose warnings would make it in the release. Maybe an end user never cares to test with the current master, but only with the release. If we would add such a message, we should make sure that any potential user gets the hint and ideally contributes the testing result, or donates a piece of hardware to a willing core developer, or something.)

kaspar030:
I'm testing this now. The driver was basically broken since #8601...

kaspar030:
Driver still works as before with basic testing.

maribu:
Thanks for testing :-)

PR #10416, opened 2018-11-16
drivers/cc2420: Implemented missing drop case

maribu:
### Contribution description

The netdev_driver_t::recv implementation of the cc2420 does not provide the drop feature. This PR adds it.

### Testing procedure

### Issues/PRs references

#10410

### Warning

This PR contains a possible regression: The first commit removes a seemingly duplicated line. If the RX FIFO needs indeed to be flushed twice because of a silicon bug (that was not documented in the code), this commit will introduce a regression. Even though the two commits look completely unrelated on the first view, the implementation of the drop feature actually depends on the first commit. If the RX-FIFO has to be flushed indeed twice, the drop feature would have to do so as well.


drivers/cc2420/cc2420.c
```diff
@@ -172,6 +172,10 @@ int cc2420_rx(cc2420_t *dev, uint8_t *buf, size_t max_len, void *info)
         cc2420_ram_read(dev, CC2420_RAM_RXFIFO, &len, 1);
         len -= 2;   /* subtract RSSI and FCF */
         DEBUG("cc2420: recv: packet of length %i in RX FIFO\n", (int)len);
+        if (max_len != 0) {
+            /* dropping frame was requested */
+            cc2420_strobe(dev, CC2420_STROBE_FLUSHRX);
```
jia200x: AFAIS since the second `cc2420_strobe` is in the `else` block, this function will only get called once. Thus, the FIFO won't be flushed properly.

I guess that could be solved putting both `cc2420_strobe` in the end and adding `return` statement with the length here.

jia200x:
BTW we should indeed document the silicon bug (see [CC2420 datasheet, section 14.3](https://web.stanford.edu/class/cs244e/papers/cc2420.pdf)) in another PR :)

maribu:
> BTW we should indeed document the silicon bug (see [CC2420 datasheet, section 14.3](https://web.stanford.edu/class/cs244e/papers/cc2420.pdf)) in another PR :)

Maybe that should be done first and rebase this PR on it? I believe a static inline function to flush the FIFO will be more readable and less surprising instead of having two times the same command. This static inline function can be documented properly.

jia200x:
>  believe a static inline function to flush the FIFO will be more readable and less surprising instead of having two times the same command

Agreed!

Could you open a PR? I can also do it, let me know ;)

maribu:
Today I'm too busy to attend to it. Feel free to create the PR. Otherwise I can do it tomorrow.

jia200x:
> Today I'm too busy to attend to it. Feel free to create the PR. Otherwise I can do it tomorrow.

Ok, I will then.

maribu:
@jia200x: I rebased and also addressed issue https://github.com/RIOT-OS/RIOT/issues/10413. Please re-review :-)

jia200x:
From docs:
```
If buf == NULL and len == 0, returns the packet size – or an upper bound estimation of the size – without dropping the packet. If buf == NULL and len > 0, drops the packet and returns the packet size.

If called with buf != NULL and len is smaller than the received packet:

    The received packet is dropped
    The content in buf becomes invalid. (The driver may use the memory to implement the dropping - or may not change it.)
    -ENOBUFS is returned
```

With this PR:
- `buf == NULL && len == 0` => return len: https://github.com/RIOT-OS/RIOT/blob/aac69c3a5680178137752fc2ad34ee7696deaba3/drivers/cc2420/cc2420.c#L178
-  `buf == NULL && len > 0` => same as before, but also drops packet: https://github.com/RIOT-OS/RIOT/blob/aac69c3a5680178137752fc2ad34ee7696deaba3/drivers/cc2420/cc2420.c#L183
- `buf != NULL && len > max_len'  => drop packet, return -ENOBUFS: https://github.com/RIOT-OS/RIOT/blob/aac69c3a5680178137752fc2ad34ee7696deaba3/drivers/cc2420/cc2420.c#L193
- `buf != NULL && len <= max_len` => receive packet, flush after received: https://github.com/RIOT-OS/RIOT/blob/aac69c3a5680178137752fc2ad34ee7696deaba3/drivers/cc2420/cc2420.c#L221

So, this fixes the initial issue

PeterKietzmann:
> I don't have hardware to test

@jia200x you do!

jia200x:
oh yes, I do :sweat_smile:

Issue #10164, opened 2018-10-15
drivers: Broadcom APDS-99xx ambient light and proximity sensor driver


gschorcht:
#### Description

I plan to port a driver for Broadcom APDS-99xx sensors to RIOT. Is someone working on it already?

### Useful links

[Datasheed of APDS-9960](https://docs.broadcom.com/docs/AV02-4191EN)

bergzand:
I have one over here, but I never started on writing a driver for it. If you have something I can review and test for you.

obgm:
I just happen to need that. What is the status of your port?

gschorcht:
@obgm No problem. I is just waiting for review.

artafakhari:
Hey Guys
Any Update on APDS-9960?

gschorcht:
Fixed with PR #10420

PR #10420, opened 2018-11-17
drivers: add driver for APDS99XX ambient light and proximity sensors

gschorcht:
### Contribution description

This PR adds a driver for Broadcom APDS99XX ambient light and proximity sensor series. It supports the following sensors:

Sensor | Driver variant  | Proximitiy [counts] | Ambient light [counts] | RGB [counts] | Illuminance [Lux]
:--------:|:-------------------:|:------------------------:|:-----------------------------:|:-----------------:|:--------------------:
APDS9900 | ```apds9900``` | x | x | - | x
APDS9901 | ```apds9901``` | x | x | - | x
APDS9930 | ```apds9930``` | x | x | - | x
APDS9950 | ```apds9950``` | x | x | x | -
APDS9960 | ```apds9960``` | x | x | x | -

The driver variants for the various sensors are defined as pseudomodules. To use the driver, the sensor used must be specified by the application by one of these pseudomodules, e.g., 
```
USEMODULE += apds9900
```

The driver functionality is divided in two parts:

- **Basic driver** ```apds99xx```
  The base driver ```apds99xx``` only supports a basic set of functions and has therefore a smaller size. The procedure for retrieving new data is polling. Ambient light and proximity sensing are supported.

- **Fully functional driver** ```apds99xx_full```
   The fully functional driver ```apds99xx_full``` supports all the features supported by the base driver, as well as all other sensor features, including interrupts and their configuration. Data-ready interrupts can be used to retrieve data. In addition, threshold interrupts can be used and configured.

For each of the two driver variants there is a corresponding test application: ```tests/driver_apds99xx``` and ```tests/driver_apds99xx_full```.

The basic driver ```apds99xx``` is automatically enabled when the driver variant for the used sensor is specicied.

### Testing procedure

To test the driver, the test application can be used without interrupts as following
```
USEMODULE=apds9960 make flash -C ```tests/driver_apds99xx``` BOARD=...
```
or with interrupt
```
USEMODULE=apds9960 make flash -C ```tests/driver_apds99xx``` BOARD=...
```
where the sensor used has be specified accordingly in the ```USEMODULE``` statement.

### Issues/PRs references

Fixes issue #10164

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -29,6 +29,15 @@ ifneq (,$(filter apa102,$(USEMODULE)))
   FEATURES_REQUIRED += periph_gpio
 endif
 
+ifneq (,$(filter apds99%full,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_gpio_irq
+endif
+
+ifneq (,$(filter apds99%,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_i2c
+  USEMODULE += apds99xx
+endif
+
 ifneq (,$(filter at,$(USEMODULE)))
   FEATURES_REQUIRED += periph_uart
   USEMODULE += fmt
--- a/drivers/apds99xx/apds99xx.c
+++ b/drivers/apds99xx/apds99xx.c
@@ -1,0 +1,534 @@
+/*
+ * Copyright (C) 2018 Gunar Schorcht
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_apds99xx
+ * @{
+ * @brief       Device driver for the Broadcom APDS99XX proximity and ambient light sensor
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ * @}
+ */
+
+#include <string.h>
+#include <stdlib.h>
+
+#include "apds99xx_regs.h"
+#include "apds99xx.h"
+
+#include "irq.h"
+#include "log.h"
+#include "xtimer.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+#if ENABLE_DEBUG
+
+#define DEBUG_DEV(f, d, ...) \
+        DEBUG("[apds99xx] %s i2c dev=%d addr=%02x: " f "\n", \
+              __func__, d->params.dev, APDS99XX_I2C_ADDRESS, ## __VA_ARGS__);
+
+#else /* ENABLE_DEBUG */
+
+#define DEBUG_DEV(f, d, ...)
+
+#endif /* ENABLE_DEBUG */
+
+#define ERROR_DEV(f, d, ...) \
+        LOG_ERROR("[apds99xx] %s i2c dev=%d addr=%02x: " f "\n", \
+                  __func__, d->params.dev, APDS99XX_I2C_ADDRESS, ## __VA_ARGS__);
+
+/** Forward declaration of functions for internal use */
+
+static int _is_available(const apds99xx_t *dev);
+static void _set_reg_bit(uint8_t *byte, uint8_t mask, uint8_t bit);
+static int _reg_read(const apds99xx_t *dev, uint8_t reg, uint8_t *data, uint16_t len);
+static int _reg_write(const apds99xx_t *dev, uint8_t reg, uint8_t *data, uint16_t len);
+static int _update_reg(const apds99xx_t *dev, uint8_t reg, uint8_t mask, uint8_t val);
+
+int apds99xx_init(apds99xx_t *dev, const apds99xx_params_t *params)
+{
+    /* some parameter sanity checks */
+    assert(dev != NULL);
+    assert(params != NULL);
+    assert(params->als_steps <= 256);
+    assert(params->wait_steps <= 256);
+#if MODULE_APDS9960
+    assert(params->prx_pulses <= 15);
+#endif
+
+    DEBUG_DEV("params=%p", dev, params);
+
+    /* init sensor data structure */
+    dev->params = *params;
+
+#if MODULE_APDS99XX_FULL
+    dev->isr = NULL;
+    dev->isr_arg = NULL;
+    dev->gpio_init = false;
+#endif
+
+    /*
+     * the sensor should be operational 5.7 ms after power on; try to check
+     * its availability for some time (maximum 500 times/I2C address writes)
+     */
+    int res = 0;
+    int count = 500;
+    while (count--) {
+        res = _is_available(dev);
+        if (res == APDS99XX_OK) {
+            break;
+        }
+    }
+    if (res != APDS99XX_OK) {
+        return res;
+    }
+
+    uint8_t reg;
+
+    /* disable and power down the sensor */
+    reg = 0;
+    if (_reg_write(dev, APDS99XX_REG_ENABLE, &reg, 1) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_I2C;
+    }
+
+    /* write ALS integration time and gain parameter */
+    uint8_t atime = 256 - dev->params.als_steps;
+    if (_reg_write(dev, APDS99XX_REG_ATIME, &atime, 1) ||
+        _update_reg(dev, APDS99XX_REG_CONTROL,
+                         APDS99XX_REG_AGAIN, dev->params.als_gain)) {
+        return -APDS99XX_ERROR_I2C;
+    }
+    /* write PRX LED pulses LED drive strength and gain parameter */
+#if MODULE_APDS9900 || MODULE_APDS9901 || MODULE_APDS9930
+    uint8_t ptime = 0xff; /* PTIME is always 0xff as recommended in datasheet */
+    if (_reg_write(dev, APDS99XX_REG_PTIME, &ptime, 1) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_I2C;
+    }
+#endif
+#if MODULE_APDS9960
+    if (dev->params.prx_pulses > 0 &&
+        _update_reg(dev, APDS99XX_REG_PPCOUNT,
+                         APDS99XX_REG_PPULSE,
+                         dev->params.prx_pulses-1) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_I2C;
+    }
+#else
+    if (_reg_write(dev, APDS99XX_REG_PPCOUNT, &dev->params.prx_pulses, 1) ||
+        _update_reg(dev, APDS99XX_REG_CONTROL, APDS99XX_REG_PDIODE, 2)) {
+        return -APDS99XX_ERROR_I2C;
+    }
+#endif
+    if (_update_reg(dev, APDS99XX_REG_CONTROL, APDS99XX_REG_PDRIVE,
+                                               dev->params.prx_drive) ||
+        _update_reg(dev, APDS99XX_REG_CONTROL, APDS99XX_REG_PGAIN,
+                                               dev->params.prx_gain)) {
+        return -APDS99XX_ERROR_I2C;
+    }
+
+    /* write the waiting time */
+    uint8_t wtime = 256 - dev->params.wait_steps;
+    if (_reg_write(dev, APDS99XX_REG_WTIME, &wtime, 1) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_I2C;
+    }
+
+    reg = 0;
+    _set_reg_bit(&reg, APDS99XX_REG_PON, 1);                           /* power on */
+    _set_reg_bit(&reg, APDS99XX_REG_AEN, dev->params.als_steps != 0);  /* enable ALS */
+    _set_reg_bit(&reg, APDS99XX_REG_PEN, dev->params.prx_pulses != 0); /* enable PRX */
+    _set_reg_bit(&reg, APDS99XX_REG_WEN, dev->params.wait_steps != 0); /* enable Wait */
+    if (_reg_write(dev, APDS99XX_REG_ENABLE, &reg, 1) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_I2C;
+    }
+
+    return APDS99XX_OK;
+}
+
+int apds99xx_data_ready_als (const apds99xx_t *dev)
+{
+    assert(dev != NULL);
+    DEBUG_DEV("", dev);
+
+    uint8_t reg;
+    if (_reg_read(dev, APDS99XX_REG_STATUS, &reg, 1) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_I2C;
+    }
+
+    return (reg & APDS99XX_REG_AVALID) ? APDS99XX_OK : -APDS99XX_ERROR_NO_DATA;
+}
+
+int apds99xx_read_als_raw(const apds99xx_t *dev, uint16_t *raw)
+{
+    assert(dev != NULL);
+    assert(raw != NULL);
+    DEBUG_DEV("raw=%p", dev, raw);
+
+    uint8_t data[2];
+
+    if (_reg_read(dev, APDS99XX_REG_CDATAL, data, 2) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_RAW_DATA;
+    }
+
+    /* data LSB @ lower address */
+    *raw = (data[1] << 8) | data[0];
+
+    return APDS99XX_OK;
+}
+
+#if MODULE_APDS9900 || MODULE_APDS9901 || MODULE_APDS9930
+static uint8_t apds99xx_gains[] = { 1, 8, 16, 120 };
+
+int apds99xx_read_illuminance(const apds99xx_t *dev, uint16_t *lux)
+{
+    assert(dev != NULL);
+    assert(lux != NULL);
+    DEBUG_DEV("lux=%p", dev, lux);
+
+    uint8_t data[4];
+
+    if (_reg_read(dev, APDS99XX_REG_CDATAL, data, 4) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_RAW_DATA;
+    }
+
+    /* data LSB @ lower address */
+    uint16_t ch0 = (data[1] << 8) | data[0];
+    uint16_t ch1 = (data[3] << 8) | data[2];
+
+    /* define some device dependent constants */
+    double df = 52;
+#if MODULE_APDS9900 || MODULE_APDS9901
+    double ga = 0.48;   /* glas or lens attenuation factor */
+    double b = 2.23;
+    double c = 0.7;
+    double d = 1.42;
+#else
+    /* APDS_9930 */
+    double ga = 0.49;   /* glas or lens attenuation factor */
+    double b = 1.862;
+    double c = 0.746;
+    double d = 1.291;
+#endif
+
+    /* algorithm from datasheet */
+    double iac1 = ch0 - b * ch1;
+    double iac2 = c * ch0 - d * ch1;
+
+    /* iac = max(iac1, iac2, 0); */
+    double iac = 0;
+    iac = (iac1 > iac) ? iac1 : iac;
+    iac = (iac2 > iac) ? iac2 : iac;
+
+    double lpc = ga * df / (apds99xx_gains[dev->params.als_gain] *
+                             dev->params.als_steps);
+    double luxd = iac * lpc;
+    *lux = luxd;
+
+    return APDS99XX_OK;
+}
+#endif /* MODULE_APDS9900 || MODULE_APDS9901 || MODULE_APDS9930 */
+
+#if MODULE_APDS9950 || MODULE_APDS9960
+int apds99xx_read_rgb_raw(const apds99xx_t *dev, apds99xx_rgb_t *rgb)
+{
+    assert(dev != NULL);
+    assert(rgb != NULL);
+    DEBUG_DEV("rgb=%p", dev, rgb);
+
+    uint8_t data[6] = { }; /* initialize with 0 */
+
+    if (_reg_read(dev, APDS99XX_REG_RDATAL, data, 6) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_RAW_DATA;
+    }
+
+    /* data LSB @ lower address */
+    rgb->val[0] = (data[1] << 8) | data[0];
+    rgb->val[1] = (data[3] << 8) | data[2];
+    rgb->val[2] = (data[5] << 8) | data[4];
+
+    return APDS99XX_OK;
+}
+#endif /* MODULE_APDS9950 || MODULE_APDS9960 */
+
+int apds99xx_data_ready_prx (const apds99xx_t *dev)
+{
+    assert(dev != NULL);
+    DEBUG_DEV("", dev);
+
+    uint8_t reg;
+
+    if (_reg_read(dev, APDS99XX_REG_STATUS, &reg, 1) != APDS99XX_OK) {
+        return -APDS99XX_ERROR_I2C;
+    }
+    return (reg & APDS99XX_REG_PVALID) ? APDS99XX_OK : -APDS99XX_ERROR_NO_DATA;
+}
+
+int apds99xx_read_prx_raw (const apds99xx_t *dev, uint16_t *prox)
+{
+    assert(dev != NULL);
+    assert(prox != NULL);
+    DEBUG_DEV("prox=%p", dev, prox);
+
+    uint8_t data[2] = { }; /* initialize with 0 */
+
+#if MODULE_APDS9900 || MODULE_APDS9901 || MODULE_APDS9930 || MODULE_APDS9950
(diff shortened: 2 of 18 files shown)


drivers/include/apds99xx.h
```diff
@@ -0,0 +1,684 @@
...
+ * @note In addition to specifying whether the base driver ```apds99xx``` or
+ * the fully featured driver ```apds99xx_full``` should be used, the
+ * application has to declare used sensor by means of various
+ * pseudomodules as follows:
+ *
+ *      ADPS9900:     USEMODULE=apds9900
+ *      ADPS9901:     USEMODULE=apds9901
+ *      ADPS9930:     USEMODULE=apds9930
+ *      ADPS9950:     USEMODULE=apds9950
+ *      ADPS9960:     USEMODULE=apds9900
```
obgm: This line should read `ADPS9960:     USEMODULE=apds9960`, I suppose?

gschorcht: Yes. Thanks for the hint.

drivers/apds99xx/apds99xx.c
```diff
@@ -0,0 +1,522 @@
...
+#include "irq.h"
+#include "log.h"
+#include "xtimer.h"
+
+#define ENABLE_DEBUG    (0)
+#include "debug.h"
+
+#if ENABLE_DEBUG
+
+#define ASSERT_PARAM(cond) \
```
aabadie: I already noticed that you were using all these macros in other of your driver PRs and I don't think this is following our code standards.
Even if I understand their usefulness while debugging the driver, they tend to make the code hard to read (personal opinion here) and, when used in different driver implementations, duplicate the code.

I recommend to not use them and keep the checks simple. See existing drivers.

gschorcht: Thanks for your review.

Which code standards do they break? These macros are not global but only local to the according files. 

Why do think that reading `ASSERT_PARAM(cond)` is harder than reading `assert(cond)`? Also, there is no code duplication since they do not generate code if `ENABLE_DEBUG` is not enabled for the module.

The advantage of using those macros is that there is only one place where changes have to be made, for example, when the format of debug outputs has to be changed.

drivers/apds99xx/apds99xx.c
```diff
@@ -0,0 +1,522 @@
...
+#define DEBUG_DEV(f, d, ...)
+
+#endif /* ENABLE_DEBUG */
+
+#define ERROR_DEV(f, d, ...) \
+        LOG_ERROR("[apds99xx] %s i2c dev=%d addr=%02x: " f "\n", \
+                  __func__, d->params.dev, APDS99XX_I2C_ADDRESS, ## __VA_ARGS__);
+
+#define EXEC_RET(f) { \
+    int _r; \
```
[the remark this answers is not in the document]
gschorcht: Thanks for your review.

> It would be better not to assign the value to _r inside of the if statement.

I'm wondering why? What is the reason.

> Also, is the intermediate variable really needed here and below?

Yes, if we want to print out the error in the debug message.

drivers/apds99xx/apds99xx.c
```diff
@@ -0,0 +1,522 @@
...
+static int _update_reg(const apds99xx_t *dev, uint8_t reg, uint8_t mask, uint8_t val);
+
+int apds99xx_init(apds99xx_t *dev, const apds99xx_params_t *params)
+{
+    /* some parameter sanity checks */
+    ASSERT_PARAM(dev != NULL);
+    ASSERT_PARAM(params != NULL);
+    ASSERT_PARAM(params->als_steps <= 256);
+    ASSERT_PARAM(params->wait_steps <= 256);
+    #if MODULE_APDS9960
```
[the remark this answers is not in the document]
gschorcht: Although I don't like to scramble the source structure with `#if ... #endif`, I can change it if you insist on it, even it is not forbidden to indent preprocessor statements. With intensive use of `#if ... # endif` the source code can become unreadable without indentation.

drivers/apds99xx/apds99xx.c
gschorcht: @aabadie Thanks for your review. I need some more clarification. I just want to understand what the problems.

> I already noticed that you were using all these macros ...

Which of these macros are you talking about exactly? I guess `ASSERT_PARAM`, `DEBUG_DEV` and `ERROR_DEV`. As you said, they are useful for debugging and they help to avoid writing down the same source code again and again.

- `ASSERT_PARAM` simply prints the condition that is not fulfilled before just getting core panic if `ENABLE_DEBUG` is set. Finally, we could also define `DEBUG_ASSERT_VERBOSE` to get the same information. I can replace them by the `assert` function without any problem if you insist on it.
- `DEBUG_DEV` and `ERROR_DEV` have a lot of advantages:
   * They guarantee a uniform format for all debug/error messages in the module including additional information.
  * Changes that need to be made, such as the format of the output, need only to be done in one place.
  * The source code lines where they are used become significantly shorter compared to using `DEBUG` or `LOG_ERROR` directly.

> when used in different driver implementations, duplicate the code.

I don't think so. If `ENABLE_DEBUG` is 0, `DEBUG_DEV` produces no code and `ASSERT_PARAM` produces only the call of the `assert` function. In case, `NDEBUG` is defined, also `ASSERT_PARAM` produces no code. `LOG_ERROR` does not generate more code than it would generate if called directly in the source code. So I see no difference or code duplication.

> I recommend to not use them and keep the checks simple.

IMHO, checking parameters is very important during the development to realize potential problems. If `NDEBUG` is defined, the parameter checks are disabled completely and produce no code and no overhead.

I hope you are not referring to the handling of return codes from I2C-based functions. From my point of view, checking the return codes from such functions is essential for the realization of fault tolerant applications. It is not a good idea to ignore errors in error-prone functions like the buses in systems with competing hardware components. PR #10526, for example, adds a lot of such error handling.

> I don't think this is following our code standards.

Which coding standard do they break? I couldn't find anything related with the exception of _global_ multiline macros. Inline functions should be used instead of global multiline macros, but this is not the use case here.

One additional comment to the `EXEC_*` macros. They simply replace a 4 lines block by 1 line in source code. IMHO, the code becomes much more readable. Using
```
    EXEC(function1(a, b, c));
    EXEC(function2(a, b, c));
    EXEC(function3(a, b, c));
```
is much better readable than 
```
    if ((res = function1(a, b, c)) != APDS99XX_OK) {
        DEBUG("[apds99xx] %s: error code %d\n", __func__, res);
        return;
    }
    if ((res = function2(a, b, c)) != APDS99XX_OK) {
        DEBUG("[apds99xx] %s: error code %d\n", __func__, res);
        return;
    }
    if ((res = function3(a, b, c)) != APDS99XX_OK) {
        DEBUG("[apds99xx] %s: error code %d\n", __func__, res);
        return;
    }
 ```
There is also no code duplication. Even more, these macros could be extended to disable the error handling completely if `NDEBUG` is defined to reduce the code to the function call.

File: drivers/apds99xx/apds99xx.c (hunk 4 above)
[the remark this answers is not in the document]
Comment (leandrolanzieri) [reply drivers/apds99xx/apds99xx.c]: > Although I don't like to scramble the source structure with `#if ... #endif`

Totally agree.

There is not an specific rule on our coding styles I think, it's just more a matter of C conventions. In my opinion it makes the preprocessor directives more visible, but I won't enforce this to remove the block. Feel free to keep it like that.

File: drivers/apds99xx/apds99xx.c (hunk 2 above)
Comment (gschorcht) [reply drivers/apds99xx/apds99xx.c]: Ok, good point and reasonable. Too bad, without such macros, the work becomes more difficult and the code gets worse than better readable :worried: We have used such macros over years especially in embedded systems development since they provide the capability to enable or disable parts of code for optimization without using preprocessor directives extensively everywhere.

File: drivers/include/apds99xx.h [hunk 5]
```diff
@@ -0,0 +1,684 @@
...
+ * following **three steps** in the given order:
+ *
+ * - **Proximity Sensing**
+ *
+ *   The sensor generates a number of IR LED pulses and measures the amount
+ *   of the IR energy reflected by an object to determine its distance. The
+ *   time required for proximity sensing (```t_prx``` ) results from the
+ *   time it takes to generate the IR LED pulses and to accumlate the
+ *   reflected IR energy (```t_prx_acc```) as well as the time for the ADC
+ *   conversion (```t_prx_acc```).
```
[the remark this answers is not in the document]
gschorcht: Yes, thanks.

drivers/apds99xx/apds99xx.c
gschorcht: I have changed all macros except `DEBUG_DEV` and `ERROR_DEV` because they really help to structure the output in a consistent format.

obgm:
FWIW: I am currently running the provided tests with an APDS 9960. Both, the basic (polling) version as well as the int-controlled version seem to work as expected. (Did not check the color values, though.)

drivers/include/apds99xx.h
```diff
@@ -0,0 +1,684 @@
...
+ * time apds99xx_params_t::als_steps and the ALS gain
+ * apds99xx_params_t::als_gain. The maximum count value (cnt_als_max) is:
+ *
+ *      cnt_als_max = APDS99XX_CNTS_P_STEP * als_steps * als_gain
+ *
+ * If there are no new data ready to read, last valid data sample is returned.
+ * Function #apds99xx_data_ready_als could be used to check whether new data
+ * are available before this function is called.
+ *
+ * @param[in]   dev     device descriptor of APDS99XX sensor to be initialized
```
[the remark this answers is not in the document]
gschorcht: Yes, thanks.

drivers/include/apds99xx.h
```diff
@@ -0,0 +1,684 @@
...
+ *
+ * Illuminance in lux is computed from raw ambient light sensing (ALS) data
+ * which are measured in counts. Since the range of ALS data depend on ALS
+ * integration time apds99xx_params_t::als_steps and the ALS gain
+ * apds99xx_params_t::als_gain, these parameters also affect the sensitivity
+ * of the illuminance.
+ *
+ * @note This function is only available for APDS9900, APDS9901 and APD9930.
+ *
+ * @param[in]   dev     device descriptor of APDS99XX sensor to be initialized
```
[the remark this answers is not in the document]
gschorcht: Copy & pasta error :sunglasses: thanks.

drivers/include/apds99xx.h
```diff
@@ -0,0 +1,684 @@
...
+ *
+ *      cnt_rgb_max = APDS99XX_CNTS_P_STEP * als_steps * als_gain
+ *
+ * If there are no data ready to read, last valid data sample is returned.
+ * Function #apds99xx_data_ready_als could be used to check whether new data
+ * are available before this function is called.
+ *
+ * @note This function is only available for APDS9950 and APD9960.
+ *
+ * @param[in]   dev     device descriptor of APDS99XX sensor to be initialized
```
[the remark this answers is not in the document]
gschorcht: :sunglasses:

drivers/include/apds99xx.h
```diff
@@ -0,0 +1,684 @@
...
+ * interface. Depending on the sensor used, it consumes only about 1 to 3 uA
+ * in this mode.
+ *
+ * @param[in]   dev     device descriptor of APDS99XX sensor to be initialized
+ *
+ * @retval  APDS99XX_OK      on success
+ * @retval  APDS99XX_ERROR_* a negative error code on error,
+ *                           see #apds99xx_error_codes_t
+ */
+int apds99xx_power_down(apds99xx_t *dev);
```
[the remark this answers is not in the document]
gschorcht: Yes, it's possible to make them constant.

gschorcht:
@leandrolanzieri Thanks for reviewing.

artafakhari:
Hello Everyone
Does the APDS-9960 work reliably on RIOT-OS?

drivers/apds99xx/apds99xx.c
```diff
@@ -0,0 +1,527 @@
...
+    double ga = 0.48;   /* glas or lens attenuation factor */
+    double b = 2.23;
+    double c = 0.7;
+    double d = 1.42;
+#else
+    /* APDS_9930 */
+    double ga = 0.49;   /* glas or lens attenuation factor */
+    double b = 1.862;
+    double c = 0.746;
+    double d = 1.291;
```
benpicco: If you multiply them all by 1000, you wouldn't need to use `double` here, no?

aabadie:
Please rebase, I have the hardware to test this PR (apds9960) :)

gschorcht:
> Please rebase, I have the hardware to test this PR (apds9960) :)

@aabadie Rebased and fixed some typos identified by codespell.

tests/driver_apds99xx/Makefile
```diff
@@ -0,0 +1,8 @@
+include ../Makefile.tests_common
+
+DRIVER ?= apds9960
+
+USEMODULE += $(DRIVER)
+USEMODULE += xtimer
```
leandrolanzieri: Isn't this pulled by `apds99%` already?

leandrolanzieri:
@gschorcht this needs rebasing again

drivers/apds99xx/apds99xx.c
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    /* data LSB @ lower address */
+    *raw = (data[1] << 8) | data[0];
+
+    return APDS99XX_OK;
+}
+
+#if MODULE_APDS9900 || MODULE_APDS9901 || MODULE_APDS9930
+#include <math.h>
```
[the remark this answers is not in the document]
gschorcht: No, removed it.

tests/driver_apds99xx/Makefile
gschorcht: You are right, removed it.

gschorcht:
With the last fixups, I also added test scripts for automatic testing.

tests/driver_apds99xx_full/main.c
```diff
@@ -0,0 +1,149 @@
...
+    (void)arg;
+    thread_flags_set(t_main, APDS99XX_IRQ_FLAG);
+}
+
+int main(void)
+{
+    apds99xx_t dev;
+
+    /* save the reference to the main thread */
+    t_main = (thread_t*)sched_threads[sched_active_pid];
```
smlng: better use `t_main = thread_getpid();` here?

drivers/apds99xx/apds99xx.c
```diff
@@ -0,0 +1,526 @@
...
+    dev->params = *params;
+
+#if MODULE_APDS99XX_FULL
+    dev->isr = NULL;
+    dev->isr_arg = NULL;
+    dev->gpio_init = false;
+#endif
+
+    /* wait for 6 ms after power on reset */
+    xtimer_usleep(6 * US_PER_MS);
```
smlng: would be nice to get rid of `xtimer`  usage, especially as this is the only one. Isn't there a state register or gpio that we could poll (`while(...) {}` style) instead?

tests/driver_apds99xx_full/tests/01-run.py
```diff
@@ -0,0 +1,25 @@
...
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import sys
+from testrunner import run
+
+def testfunc(child):
+    child.expect('Initializing APDS99XX sensor')
+    i = child.expect(['[OK]','[Failed]'])
+    if i==1:
```
aabadie:
```suggestion
    if i == 1:
```

tests/driver_apds99xx_full/tests/01-run.py
```diff
@@ -0,0 +1,25 @@
...
+# Copyright (C) 2020 Gunar Schorcht <gunar@schorcht.net>
+#
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import sys
+from testrunner import run
+
+def testfunc(child):
```
aabadie: The function must be surrounded by 2 blank lines (only one here).

tests/driver_apds99xx_full/tests/01-run.py
```diff
@@ -0,0 +1,25 @@
...
+    child.expect('Initializing APDS99XX sensor')
+    i = child.expect(['[OK]','[Failed]'])
```
aabadie:
```suggestion
    i = child.expect(['[OK]', '[Failed]'])
```

tests/driver_apds99xx/tests/01-run.py
```diff
@@ -0,0 +1,26 @@
...
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import sys
+from testrunner import run
+
+def testfunc(child):
+    child.expect('Initializing APDS99XX sensor')
+    i = child.expect(['[OK]','[Failed]'])
```
aabadie:
```suggestion
    i = child.expect(['[OK]', '[Failed]'])
```

tests/driver_apds99xx/tests/01-run.py
```diff
@@ -0,0 +1,26 @@
...
+    if i==1:
```
aabadie:
```suggestion
    if i == 1:
```

sys/auto_init/saul/auto_init_apds99xx.c
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_apds99xx
+ * @ingroup     sys_auto_init_saul
+ * @brief       Auto initialization of Boradcom APDS99XX proximity and ambient light sensor
```
aabadie:
```suggestion
 * @brief       Auto initialization of Broadcom APDS99XX proximity and ambient light sensor
```

drivers/include/apds99xx.h
```diff
@@ -0,0 +1,684 @@
...
+ *   APDS9900   | 2.72 ms | ```wait_steps``` * 2.72 ms | ```wait_steps``` * 12 * 2.72 ms
+ *   APDS9901   | 2.72 ms | ```wait_steps``` * 2.72 ms | ```wait_steps``` * 12 * 2.72 ms
+ *   APDS9930   | 2.73 ms | ```wait_steps``` * 2.73 ms | ```wait_steps``` * 12 * 2.73 ms
+ *   APDS9950   | 2.40 ms | ```wait_steps``` * 2.40 ms | ```wait_steps``` * 12 * 2.40 ms
+ *   APDS9960   | 2.78 ms | ```wait_steps``` * 2.78 ms | ```wait_steps``` * 12 * 2.78 ms
+ *   <br>
+ *
+ * - **Ambient Light Sensing (ALS)**
+ *
+ *   The sensor converts the amplified phododiode currents using integrating
```
aabadie:
```suggestion
 *   The sensor converts the amplified photodiode currents using integrating
```

aabadie:
Also: please rebase :)

aabadie:
I could test this PR on the adafruit-clue board (after some rebase and fix of conflicts). This board embeds an apds9960 variant. Here is what I get:
- `tests/driver_apds99xx` works like a charm
- `tests/driver_apds99xx_full` doesn't seem to work: no interrupt is generated when I get my hand close to the proximity sensor. I configured the interrupt pin following the board schematic, so it _should_ work. Any idea what could be the cause @gschorcht ?

gschorcht:
> * `tests/driver_apds99xx_full` doesn't seem to work: no interrupt is generated when I get my hand close to the proximity sensor. I configured the interrupt pin following the board schematic, so it _should_ work. Any idea what could be the cause @gschorcht ?

@aabadie Do you see ALS measurements?
```
ambient = 639 [cnts]
red = 208 [cnts], green = 295 [cnts], blue = 263 [cnts]
+-------------------------------------+
```
If yes, interrupts are working since these measurements use the data ready interrupt. The proximity interrupt is triggered, when the proximity value exceeds 200 counts. This happens at a distance of about 2 cm. In that case, the output also shows the proximity value, for example:
```
proximity = 205 [cnts]
+-------------------------------------+
ambient = 322 [cnts]
red = 95 [cnts], green = 168 [cnts], blue = 135 [cnts]
+-------------------------------------+
```

tests/driver_apds99xx_full/main.c
gschorcht: The reference to the main thread is required in `apds99xx_isr`. Using `thread_get(pid)` doesn't work since it replaces the const attribute.

I changed the interrupt handling to a mutex based approach.

drivers/apds99xx/apds99xx.c
gschorcht: The problem is that this time is specified in the data sheet. It is the time the sensor needs for initialization after power-on to become ready for operation. Unfortunately there is no status flag.

Possible solutions:

1. We simply assume that RIOT takes longer to load than this 6 ms before the device is accessed for the first time.

2. We allow multiple tries of the `_is_available' function and use the  I2C ACK to determine when the device is available. However, at an I2C speed of 400 kbps, each failed attempt on the bus would only take 0.025 ms, which could mean a maximum of 250 failed attempts for 6 ms.

gschorcht:
@aabadie @smlng Thanks for reviewing. I have redesigned the PR and hopefully addressed all your remarks.

aabadie:
> I have redesigned the PR and hopefully addressed all your remarks.

Thanks! I'll have a look and test again tomorrow.

drivers/apds99xx/apds99xx.c
gschorcht: I have removed the dependency on xtimer with commit https://github.com/RIOT-OS/RIOT/pull/10420/commits/d909fad27cd9d2665c11d2b464be2d8df8005be2.

tests/driver_apds99xx_full/main.c
```diff
@@ -0,0 +1,145 @@
...
+    }
+    else {
+        puts("[Failed]");
+        return 1;
+    }
+
+    /* configure the interrupts */
+    apds99xx_int_config_t int_cfg = {
+        .als_int_en = true,     /* ALS interrupt enabled */
+        .als_pers = 0,          /* ALS interrupt in each cylce (data-ready) */
```
aabadie:
```suggestion
        .als_pers = 0,          /* ALS interrupt in each cycle (data-ready) */
```

aabadie:
Tested this PR again this morning on adafruit-clue and I can't have the interrupt based test application to work. If I comment out the `mutex_lock`, messages are displayed as expected: the proximity value is only displayed when I put my finger close to the sensor (which means the interrupt source bit is set).
So there's something wrong with the interrupt pin: I have no idea what could be the problem. I suspect something with the board: there are white leds that are on when waiting for the interrupt and they go down when an interrupt occurs.

I trust you that this test is working anyway.

drivers/Makefile.dep
```diff
@@ -29,6 +29,16 @@ ifneq (,$(filter apa102,$(USEMODULE)))
   FEATURES_REQUIRED += periph_gpio
 endif
 
+ifneq (,$(filter apds99%full,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_gpio_irq
+  USEMODULE += apds99xx
```
aabadie: This is dependency is already set in the `ifneq` block below which should also match the _full case. So:
```suggestion
```

gschorcht: Yep, thanks.

sys/auto_init/saul/auto_init_apds99xx.c
```diff
@@ -0,0 +1,108 @@
+/*
```
aabadie: This file is the duplicate of `drivers/saul/init_devs/auto_init_apds99xx.c`. Which one should be kept ?

gschorcht: The version in `sys/auto_init` is just a leftover after the rebase. I was sure that I had moved it from `sys/auto_init/saul/` to `drivers/saul/init_devs` instead of copying it. Good catch, thanks.

gschorcht:
> Tested this PR again this morning on adafruit-clue and I can't have the interrupt based test application to work. If I comment out the `mutex_lock`, messages are displayed as expected: the proximity value is only displayed when I put my finger close to the sensor (which means the interrupt source bit is set).
> So there's something wrong with the interrupt pin: I have no idea what could be the problem. I suspect something with the board: there are white leds that are on when waiting for the interrupt and they go down when an interrupt occurs.

The Adafruit-Clue seems to be a very interesting board. According to the schematic `P0.09` should be the `APDS_IRQ` pin. `GPIO_IN_PU` mode is supported according to `nrf5240` product specification. Might it be a problem that the pin could also be used as `NFC1` pin, something that has to be considered during initialization? The `WHITE_LED` signal is controlled by `P0.10/NFC2`. Might it be that the pins are mixed up?

aabadie:
> Might it be that the pins are mixed up?

I had the same idea and tried to invert the pins but it didn't work either. I won't block this PR because of this anyway.

aabadie:
I think we are good with this one. @gschorcht, please squash!

gschorcht:
Thank you all for reviewing and testing.

BTW, I hope I can implement the gesture recognition function of APDS9960 in next week. I will provide it as separate PR.

PR #10423, opened 2018-11-17
frdm-k64f: Define BTNx macros for user pushbuttons

gebart:
### Contribution description

The configuration was missing definitions for the BTN0_PIN, BTN0_MODE macros used by some applications, e.g. tests/buttons.

### Testing procedure

- [ ] tests/buttons
  - compile and run `tests/buttons`
  - Press the user buttons SW2, SW3
  - Verify in the terminal output that pressing SW2 generates an interrupt `Pressed BTN0`, SW3 generates an interrupt `Pressed BTN1`

### Issues/PRs references

Same as #10414 but for a different board.

### DIFF
--- a/boards/frdm-k64f/include/board.h
+++ b/boards/frdm-k64f/include/board.h
@@ -54,6 +54,20 @@ extern "C"
 /** @} */
 
 /**
+ * @name    Button pin definitions
+ * @{
+ */
+/* SW2, SW3 will short these pins to ground when pushed. PTA4 has an external
+ * pull-up resistor to VDD, but there is no external pull resistor on PTC6 */
+/* BTN0 is mapped to SW2 */
+#define BTN0_PIN            GPIO_PIN(PORT_C,  6)
+#define BTN0_MODE           GPIO_IN_PU
+/* BTN1 is mapped to SW3 */
+#define BTN1_PIN            GPIO_PIN(PORT_A,  4)
+#define BTN1_MODE           GPIO_IN_PU
+/** @} */
+
+/**
  * @name    FXOS8700CQ 3-axis accelerometer and magnetometer bus configuration
  * @{
  */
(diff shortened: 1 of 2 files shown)


boards/frdm-k64f/include/board.h
```diff
@@ -53,6 +53,20 @@ extern "C"
 #define LED2_TOGGLE        (GPIOB->PTOR = LED2_MASK)
 /** @} */
 
+/**
+ * @name    Button pin definitions
+ * @{
+ */
+/* SW2, SW3 will short these pins to ground when pushed. PTA4 has an external
+ * pull-up resistor to VDD, but there is no external pull resistor on PTC6 */
```
aabadie: From the schematic, there's also a pull resistor on PCT6. Are you sure of you are writing here ? (I don't have this hardware for testing unfortunately).
I would have put GPIO_IN_PU to both button modes.

gebart: One of the pull resistors are DNP (do not populate) in the schematic, probably to avoid affecting the INT1 pin on the accelerometer.
I could change both to GPIO_IN_PU if you prefer.

gebart: @aabadie the link you posted is for the frdm-kw41z, this PR is for the frdm-k64f

gebart: I will change the mode to IN_PU on both

aabadie: ooops, not fully awake..

aabadie: Ok, in k64f schematics, both buttons pull resistors are DNP. So ok for IN_PU.

gebart: I noticed that the R75 resistor is only populated in revision E3 boards or later. (Change log in E4 schematic states: "E3:  Depopulating C55, populating R75"). Using IN_PU should probably be the most robust solution here.

gebart:
squashed

aabadie:
All green, let's go!

PR #10429, opened 2018-11-19
make: Add staging keywords. 

jcarrano:
_**IMPORTANT: this is just a demo, I do not intend it to be merged!**_

### Contribution description

Following the discussion in #9430 , I figured it was a good idea to show a demo of how _staging keywording_ can be implemente (see [this comment](https://github.com/RIOT-OS/RIOT/pull/9430#issuecomment-429912408)). I was a bit uneasy with the amount of changes introduced by that PR.

This commit allows modules and packages to be added to a "staging" list, by using `STAGING_MODULES += module_name` in Makefile.dep.

If any "staging" modules is used, the build will fail unless the used staging modules are explicitly added to the `STAGING_ALLOWED` variable.

My philosophy with regards to this feature is that where the code is located and how it is handled need not be coupled. Note that this is not _alternative_, but rather _complementary_ to #9430: nobody says that the modules keyworded as staging cannot all be placed in a separate directory, enforced by static checks. The point I want to make with this PR is that having the build system behave differently depending on which path code is located introduces more complexity than necessary, while reducing flexibility.

### Testing procedure

There is a disposable commit only to show how the STAGING_ALLOWED variable is used.

The ccn-lite package was arbitrarily chosen.

1. ccn-lite was added to STAGING_MODULES
2. examples/ccn-lite-relays contains `STAGING_ALLOWED = ccn-lite` to allow
   it to be built.
3. To test the feature, remove the STAGING_ALLOWED and try compiling.

### Issues/PRs references

See #9430.

### DIFF
--- a/Makefile.dep
+++ b/Makefile.dep
@@ -818,6 +818,7 @@ USEMODULE += $(filter periph_%,$(FEATURES_USED))
 # recursively catch transitive dependencies
 USEMODULE := $(sort $(USEMODULE))
 USEPKG := $(sort $(USEPKG))
+STAGING_MODULES := $(sort $(STAGING_MODULES))
 ifneq ($(OLD_USEMODULE) $(OLD_USEPKG),$(USEMODULE) $(USEPKG))
   include $(RIOTBASE)/Makefile.dep
 endif
--- a/Makefile.include
+++ b/Makefile.include
@@ -437,9 +437,9 @@ ifeq ($(BUILD_IN_DOCKER),1)
 link: ..in-docker-container
 else
 ifeq (,$(RIOTNOLINK))
-link: ..compiler-check ..build-message $(ELFFILE) $(HEXFILE) print-size
+link: ..compiler-check ..build-message ..staging-check $(ELFFILE) $(HEXFILE) print-size
 else
-link: ..compiler-check ..build-message $(BASELIBS)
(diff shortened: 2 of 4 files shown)


kaspar030:
> I was a bit uneasy with the amount of changes introduced by that PR.

The amount of changes in that PR are <30 lines outside of staging/, mostly "if staging enables: include a file" type of changes. Is that making you uneasy?

The other PR allows the build system to completely ignore a staging module (in fact, the whole tree). It makes disabling of modules transitive. It allows applications to be in staging. It makes "info-boards-supported" not include applications depending on a staging module marked as broken. Those are important. 

All of those facilities don't need to be in staging/staging.mk and could be made available to all modules. Where it makes sense, I think we should do that *at some point*. But also, right now, all of that is easy to opt-out using ENABLE_STAGING=0, which is default.

Actually, this PR's STAGING_MODULES and STAGING_ALLOWED are pretty much the same as the other PR's "STAGING_MODULES" and "STAGING_MODULES - BROKEN", just at different places in the makefiles, and named differently.

@jcarrano Seeing that this is implementing part of #9430 just slightly differently, I'd suggest sticking with the implementation there, so we can try the *concept* first.

jcarrano:
I'm closing this as it is clearly going nowhere. I still think keywording is a good idea and may be used solve a few issues besides staging.

PR #10430, opened 2018-11-19
drivers: add PCF857X I2C I/O expander driver

gschorcht:
### Contribution description

This PR adds a driver for the widespread PCF857x serial I/O expander modules which extend the number of GPIOs over I2C. The following expander modules are supported by the driver:

Expander | Type                    | Pseudomodule
:--------|:------------------------|:------------
PCF8574  |  8-bit I2C I/O expander | `pcf8574`
PCF8574A |  8-bit I2C I/O expander | `pcf8574a`
PCF8575  | 16-bit I2C I/O expander | `pcf8575`

The driver defines the PCF857x expander variants by a pseudomodule for each of supported expander module

The driver interface is kept as compatible as possible with the peripheral GPIO interface. The only difference is that the functions have the prefix `pcf857x_` and require an additional parameter, the pointer to the expansion device of type `pcf857x_t`

Since PCF857x expander pins are only quasi-bidirectional without having a separate direction control signal, they only support active LOW driven outputs and inputs that are pulled up to HIGH. HIGH outputs are nothing else than pulled-up inputs. The driver maps this behavior to the `GPIO_IN_PU` and `GPIO_OD` and additionally to the `GPIO_IN`, `GPIO_OUT` and `GPIO_OD_PU` modes.

Optionally, the driver can use the low-active open-drain INT signal of PCF857x expander modules

- to provide external interrupt capabilities for all expander pins which are compatible to module `periph_gpio_irq`
- to keep internal up-to-state of expander inputs to avoid reading expander input pins via I2C each time a single input value read with `pcf857x_gpio_read`, see the benchmarks.

The driver supports direct GPIO mapping to SAUL which is compatible to the peripheral GPIO SAUL interface. Thus PCF857X I/O expander pins can be used with SAUL in the same way as peripheral GPIOS.

**Update** ~The driver also supports GPIO extension API as proposed in PR #9582 and implemented in PR #9860 and #9958 .~

## Benchmarks

The performance of the driver strongly depends on using interrupt capabilities or not:

- PCF857x driver run-time performance benchmark without interrupt capabilities @ 400 kHz I2C bus speed:
```
                 nop loop:       639us  ---   0.063us per call  ---   15649452 calls per sec
                 gpio_set:    842995us  ---  84.299us per call  ---      11862 calls per sec
               gpio_clear:    840009us  ---  84.000us per call  ---      11904 calls per sec
              gpio_toggle:    842000us  ---  84.200us per call  ---      11876 calls per sec
                gpio_read:    877514us  ---  87.751us per call  ---      11395 calls per sec
               gpio_write:    841430us  ---  84.143us per call  ---      11884 calls per sec
```
- PCF857x driver run-time performance benchmark with interrupt capabilities (module `periph_gpio_irq`)  @ 400 kHz I2C bus speed:
```
                 nop loop:       626us  ---   0.062us per call  ---   15974440 calls per sec
                 gpio_set:   1751632us  ---  175.163us per call  ---       5708 calls per sec
               gpio_clear:   1755016us  ---  175.501us per call  ---       5697 calls per sec
              gpio_toggle:   1755082us  ---  175.508us per call  ---       5697 calls per sec
                gpio_read:      4751us  ---   0.475us per call  ---    2104820 calls per sec
               gpio_write:   1759999us  ---  175.999us per call  ---       5681 calls per sec
```
Reference board was an `esp8266-esp-12x` in I2C fast mode.

### Testing procedure

The test application `tests/driver_pcf857x` can be used to test each PCF857x expander I/O pin with shell commands. Dependent on used  PCF857x variant, it has to be compiled with at least one of the pseudomodules `pcf8574`, `pcf8574a` or `pcf8575`, e.g.:

     USEMODULE=pcf8575 make -C tests/driver_pcf857x BOARD=... flash 

Add module `periph_gpio_irq` and override the MCU interrupt pin by parameter  PCF857X_PARAM_INT_PIN` to use the PCF857x interrupt signal, e.g.,

     CFLAGS='-DPCF857X_PARAM_INT_PIN=\(GPIO_PIN\(0,6\)\)' \
     USEMODULE='pcf8575 pcf857x_irq' make -C tests/driver_pcf857x BOARD=... flash

The number of the first PCF857X expander port used by the test application is defined by the macro `PCF857X_P0`, which is 16 by default.

For testing the driver follow the steps:
1. Compile the test application the for a board of your choice using module `pcf857x_irq` and suitable interrupt pin as shown above.
2. Connect any MCU GPIO pin, e.g. `GPIO_PIN(0,0)` with any pin of the PCF857X expander, e.g. pin 6.
3. Use the following commands to test that the driver including interrupts works correctly
    ```
    > init_out 0 0
    > clear 0 0
    > init_int 16 6 2
    > read 16 6
    PCF857X pin (dev 16, pin 06) is LOW
    > toggle 0 0
    INT: external interrupt from pin 6
    > read 16 6
    PCF857X pin (dev 16, pin 06) is HIGH
    > toggle 0 0
    INT: external interrupt from pin 6
    ```

### Issues/PRs references

**UPDATE** ~The driver is compatible with the GPIO extension API, PR #9582, #9860 and #9958, but it does not depend on it.~ Compatibility with the GPIO extension API, PR #9582, #9860 and #9958 has been removed for the moment due to the new proposal of an extensible GPIO API in PR ~#12877~ #14602. The driver will be adapted later for this GPIO API.


kYc0o:
@ZetaR60 has already made some PRs and opened issues to address this feature. It's not straightforward but I think the expanders should use the underlying GPIO API. Could you take a look into #9860 #9690 #9190 and #9958 ?

gschorcht:
@kYc0o Tanks for the hint, I will take a look.

ZetaR60:
@gschorcht See also #9054 which needs updating but may have some overlap. I will take a more in-depth look at your PR soon.

Regarding compatibility with the extension interface:
If you use function definitions in your driver that are compatible with the callback typedefs in ```drivers/include/extend/gpio.h``` from #9958 then you should be able to populate the ```gpio_ext_driver``` struct by casting your driver's functions inside the definition for that struct. That way ```pcf857x_gpio_extend.c``` only contains a ```const gpio_ext_driver_t``` definition rather than a bunch of function wrappers. It _might_ be okay to leave the type of the device descriptor in your driver as the normal descriptor pointer rather than a void pointer and have the function cast take care of the conversion, but this may cause the compiler to complain.

gschorcht:
@ZetaR60 Good news, the driver is working with your GPIO extension API without any changes :smile: It is as fast as direct access:
```
                 nop loop:       625us  ---   0.062us per call  ---   16000000 calls per sec
                 gpio_set:   1701875us  ---  170.187us per call  ---       5875 calls per sec
               gpio_clear:   1699375us  ---  169.937us per call  ---       5884 calls per sec
              gpio_toggle:   1701251us  ---  170.125us per call  ---       5878 calls per sec
                gpio_read:      9250us  ---   0.925us per call  ---    1081081 calls per sec
               gpio_write:   1701001us  ---  170.100us per call  ---       5878 calls per sec
```
For some suggestions concerning the implementation, I will use your PRs.

I will update my PR. As long as module `extend_gpio` is not enabled it behaves exactly as befor and is compilable without the API. So I could already provide the driver before your changes are merged without any conflict.

ZetaR60:
@gschorcht Great! I am glad it is working out well on real hardware.

I have been looking at the documentation for the ICs supported by this PR and those supported by #9054 and it looks like at least some of the chips are pin-compatible and maybe even register compatible (some of the docs were a bit spotty with the register maps). It seems there is a wide variety of ICs from multiple manufacturers that:
- Use I2C with a pull-down interrupt pin
- Have similar use of address input pins
- Are usually 8 or 16 bits/pins wide
- Have identical pinouts for the same width
- Have similar or identical internal registers

I think that, if possible, we should make a single driver for all of them, with just a few parameters that are changeable depending on the device. I tried to do this with #9054 (see the already long list of devices in ```drivers/include/pca95xx.h```), but it looks like you found even more similar devices.

gschorcht:
@ZetaR60 How deep did you test the interrupt handling? I spent a lot of time to test a lot of different cases with single and multiple expanders and had to realize the following problems:

- If an output port is connected to an input port at the same expander, changing the output does not trigger an interrupt. The workaround for me was to call the ISR when outputs have been written. The ISR reads the inputs and calls users callback function when they have changed. That is, an interrupt is simulated.
- When PCF857x expansion ports of different devices are interconnected, an I2C write access to an output port of one device may cause an interrupt by an input port of another device before the I2C write access to the first device has been finished. The handling of the interrupt of the second device requires I2C read access. If the devices are connected to the same I2C bus, the current I2C write access to the first device avoids the required I2C read access from interrupt handler to the second device. The I2C read access fails with an arbitration lost and the interrupt is lost. Because the interrupt handling occurs in the same thread context as the I2C write access, the mutex-based `i2c_aquire` does not help to prevent this conflict. Therefore, the interrupt of the second device must be delayed until the I2C write access of the first device is completed.  The workaround for the PCF857x driver was to implement an ISR event wait list for each device.

I don't know whether these problems also occur for PCA95xx device, but I would guess yes.

gschorcht:
@ZetaR60 The MCP23x17 expanders PR #7652 are completly different and require their own driver.
@PeterKietzmann BTW, if it is OK for you I would take the driver from #7652 to adopt it to be conformant with the @ZetaR60's GPIO extension API PR #9582 and to support the SPI variant of the driver.

ZetaR60:
> Although the differences seem to be small, the handling of the devices may be too different to support them with only one driver. The biggest difference between TCA99xx and PCF857x are
> - TCA95xx has port configuration registers while PCF857x dos not
> - TCA95xx has bidirectional ports while PCF857x has only quasi-bidirectional ports

I missed the register differences when skimming the PCF857x, since it looked like the documentation was just missing. It's weird that they ICs are almost exactly the same in every other way. I agree that they should be a different driver.

> How deep did you test the interrupt handling?

I did not test for the sort of problems you mentioned. I do not have a proper break-out for the PCA95xx, and it is a bit awkward to wire TSSOP pins together when probing them. I am glad you caught this! I agree with you expectation that the exact same problems probably exist with PCA95xx.

> The workaround for the PCF857x driver was to implement an ISR event wait list for each device.

This seems a bit complex for a workaround. The best solution would be to fix ```i2c_acquire``` so that it is aware that it is in an interrupt. I will take a look at what might be required.

ZetaR60:
@gschorcht Looks like your ```i2c_acquire``` issue is actually a bug in the behavior of ```mutex_lock``` on your system (and probably many others). See #10488

wosym:
What's the current state of this? 
I started writing a very basic driver for this component last week during development of a separate project, and was planning to extend it to upstream it eventually. But luckily I found this PR by googling! Otherwise I would have wasted a lot more time reinventing the wheel :) (before I wrote the first line last week I checked the PR's, but PCF8574 didn't have any matches because this PR is named differently)

Is the current form usable? Or are the still open dependencies hard dependencies?

gschorcht:
It can be used as is. It's interface corresponds to the GPIO peripheral API. Once PR #14602 is merged, it will be extended a bit to use it like MCU GPIOs. But this is out of scope of this PR.

drivers/include/pcf857x.h
```diff
@@ -0,0 +1,542 @@
...
+ *   value of an expander GPIO pin is read with #pcf857x_gpio_read.
+ *
+ * Since interrupts are handled in the context of a separate event thread,
+ * enabling interrupts requires more RAM. Therefore, interrupts have to be
+ * activated explicitly with the module `pcf857x_irq`. Furthermore, the MCU
+ * interrupt pin to which the PCF857X INT signal is connected has to be
+ * defined by the parameter pcf857x_params_t::int_pin either in the parameter
+ * file or at the command line, e.g.
+ *
+ *      CFLAGS="-DPCF857X_PARAM_INT_PIN=\(GPIO\(0,6\)\)" \
```
wosym, line 121: Is it possible you meant GPIO**_PIN**(0,6) here?

drivers/include/pcf857x.h
```diff
@@ -0,0 +1,542 @@
...
+ * placing a file `pcf857x_params.h` in the application directory `$APPDIR`.
+ * For example, the definition of the configuration parameter array for the
+ * two devices above could be:
+ *
+ *      static const pcf857x_params_t pcf857x_params[] = {
+ *          {
+ *              .dev = I2C_DEV(0),
+ *              .addr = 0,
+ *              .exp = PCF857X_EXP_PCF8574A,
+ *              .int_pin = GPIO(0,1),
```
wosym, line 184: Same question here.

drivers/include/pcf857x.h
```diff
@@ -0,0 +1,542 @@
...
+ *          },
+ *          {
+ *              .dev = I2C_DEV(0),
+ *              .addr = 0,
+ *              .exp = PCF857X_EXP_PCF8575,
+ *              .int_pin = GPIO(0,2),
```
wosym, line 190: and here.

drivers/include/pcf857x.h
gschorcht: Indeed, thanks for pointing that. I will fix it, when I'm back from vacation.

wosym:
I've spent the last few hours trying to figure out how to enable interrupt-support, but wasn't able to. When `pcf857x_irq` is added as a module, the `pcf857x_init` function fails. It succeeds without adding the irq module.

After a while, I noticed that the instructions in the .h file differ from the instructions the PR Testing Procedure.
In the description of this PR it says:
```
Add module periph_gpio_irq and override the MCU interrupt pin by parameter PCF857X_PARAM_INT_PIN` to use the PCF857x interrupt signal, e.g.,

 CFLAGS="-DPCF857X_PARAM_INT_PIN=\(GPIO\(0,6\)\)" \
 USEMODULE="pcf8575 periph_gpio_irq" make 
```
In the .h file it says:
```
 *      CFLAGS="-DPCF857X_PARAM_INT_PIN=\(GPIO\(0,6\)\)" \
 *      USEMODULE="pcf8575 pcf857x_irq" make -C tests/driver_pcf857x BOARD=...
```
(Note that different modules are enabled)


Neither of these worked for me, and combining them also did not work. The init keeps failing when interrupts are enabled, but succeeds when they are not. 

Strangely enough, I thought I recalled having used interrupts yesterday without any problems... Or did I dream that? 

Would you mind explaining which method is the correct one to enable interrupts?
(But of course, enjoy your vacation first!)

gschorcht:
> After a while, I noticed that the instructions in the .h file differ from the instructions the PR Testing Procedure.
> In the description of this PR it says:
> 
> ```
>  USEMODULE="pcf8575 periph_gpio_irq" make -C tests/driver_pcf857x BOARD=...
> ```
> 
> In the .h file it says:
> 
> ```
>  *      USEMODULE="pcf8575 pcf857x_irq" make -C tests/driver_pcf857x BOARD=...
> ```
> 
> (Note that different modules are enabled)

Thanks for pointing that out. The command given in `pcf857x.h` is correct. That one in the test is just a leftover. In former versions of this driver, interrupts were used implicitly when module `periph_gpio_irq` was enabled. However, since I2C bus access doesn't work in interrupt context, enabling interrupts requires the use of an additional event thread that handles the interrupts in thread context. Because this event thread requires additional memory, interrupts are only used by the driver if module `pcf857x_irq` is enabled explicitly.

gschorcht:
> Correct me if I'm wrong, but I have a feeling that [here](https://github.com/RIOT-OS/RIOT/pull/10430/files#diff-548885ac7542f3f6280654d4bebfd69dR263) the idea is that it is allowed to mix interrupt-devices, and non-interrupt devices. The conditional will cause the read to be skipped if int_pin is defined (because that read has been done previously, when the interrupt occured), but will execute the read if int_pin was not defined (and thus, the device was configured as not using interrupts).
> 
> This makes sense to me, but... it did not work when I tried it. I got a kernel panic straight away.

Mixing interrupt-devices and non-interrupt devices was never intended. It would cause the code to be bloated because of the handling of the special case that `pcf857x_irq` is defined and the interrupt pin is not defined.
 
> I think this is because on [line 93](https://github.com/RIOT-OS/RIOT/pull/10430/files#diff-548885ac7542f3f6280654d4bebfd69dR93) you assert that the pin is never GPIO_UNDEF.
> 
> Also on [line 134](https://github.com/RIOT-OS/RIOT/pull/10430/files#diff-548885ac7542f3f6280654d4bebfd69dR134) the comment says "initialize **if** defined", however there is no conditional statement there.

This is just a leftover from former versions.

> After removing the assert on the int_pin and adding the extra conditional, everything worked as I think you intended it to work.

Do you think it makes sense to have a mixture of interrupt devices and uninterruptible devices? Not from my point of view. So if you are not against I would remove all comments and the conditional here https://github.com/RIOT-OS/RIOT/blob/12f2ea9252f37c6e8fcc58275617fa2ea9044113/drivers/pcf857x/pcf857x.c#L263 Thus, if `pcf857x_irq` is enabled, it is ensured by the assertion that the interrupt pin is defined and all the other code can rely on that.

wosym:
> Mixing interrupt-devices and non-interrupt devices was never intended. It would cause the code to be bloated because of the handling of the special case that `pcf857x_irq` is defined and the interrupt pin is not defined.

I assumed it was intended and made the changes I thought would be necessary (Only 2 deleted lines if I recall correctly). It doesn't bloat anything, and as far as I have experienced it works perfectly (but it's entirely possible that under certain conditions it will fail. I have not tested it extensively). 



> Do you think it makes sense to have a mixture of interrupt devices and uninterruptible devices? Not from my point of view. So if you are not against I would remove all comments and the conditional here

I'm in no position to be against anything. It's your driver, so your decision :) However... I do think it makes sense. In my current project, I have 2 expanders controlling leds, 1 expander controlling dip switches, and 1 expander controlling push buttons. I only defined the one with the push buttons to be interrupt based. The one with the dipswitches is only read when needed, and the leds... well... I'm curious what an interrupt based LED would look like :) 

So yes, I think it makes sense to make it possible to mix them. (Personally, I probably would continue using your older version instead of the updated version if you decided to remove the functionality after all. )

> https://github.com/RIOT-OS/RIOT/blob/12f2ea9252f37c6e8fcc58275617fa2ea9044113/drivers/pcf857x/pcf857x.c#L263
> Thus, if `pcf857x_irq` is enabled, it is ensured by the assertion that the interrupt pin is defined and all the other code can rely on that.

I removed that assertion. Because you check for GPIO_UNDEF at int_pin for each device seperately, I think you unintentionally provided a very valuable feature in your driver (imo) :)

drivers/pcf857x/include/pcf857x_params.h
```diff
@@ -0,0 +1,123 @@
...
+#elif IS_USED(MODULE_PCF8574)
+#define PCF857X_PARAM_EXP       (PCF857X_EXP_PCF8574)
+#elif IS_USED(MODULE_PCF8574A)
+#define PCF857X_PARAM_EXP       (PCF857X_EXP_PCF8574A)
+#endif
+#endif /* PCF857X_PARAM_EXP */
+
+#ifndef PCF857X_PARAM_INT_PIN
+/** MCU interrupt pin */
+#define PCF857X_PARAM_INT_PIN   (GPIO_PIN(0, 0))
```
benpicco, line 56: Should rather be `GPIO_UNDEF` to avoid surprises

drivers/include/pcf857x.h
```diff
@@ -0,0 +1,557 @@
...
+    unsigned int pin_num;     /**< number of I/O pins, depends on used expander
+                                   variant */
+    pcf857x_data_t modes;     /**< expander pin modes */
+    pcf857x_data_t in;        /**< expander input pin values */
+    pcf857x_data_t out;       /**< expander output pin values */
+
+#if IS_USED(MODULE_PCF857X_IRQ) || DOXYGEN
+    gpio_isr_ctx_t isr[PCF857X_GPIO_PIN_NUM];  /**< ISR with arg for each expander pin */
+    gpio_flank_t flank[PCF857X_GPIO_PIN_NUM];  /**< interrupt flank for each expander pin */
+    bool enabled[PCF857X_GPIO_PIN_NUM];        /**< enabled flag for each expander pin */
```
benpicco, line 389: I guess a bitmap would save a fey bytes here

drivers/include/pcf857x.h
```diff
@@ -0,0 +1,557 @@
...
+
+#endif /* MODULE_PCF857X_IRQ */
+
+/**
+ * @brief   PCF857X device data structure type
+ */
+typedef struct {
+    pcf857x_params_t params;  /**< device initialization parameters */
+
+    unsigned int pin_num;     /**< number of I/O pins, depends on used expander
```
benpicco, line 380:
```suggestion
    uint8_t pin_num;     /**< number of I/O pins, depends on used expander
```
should suffice

benpicco (commented):
Looks good! Just some small nits

drivers/pcf857x/pcf857x.c
```diff
@@ -0,0 +1,470 @@
...
+int pcf857x_gpio_init(pcf857x_t *dev, gpio_t pin, gpio_mode_t mode)
+{
+    /* some parameter sanity checks */
+    assert(dev != NULL);
+    assert(pin < dev->pin_num);
+
+    DEBUG_DEV("pin=%u mode=%u", dev, pin, mode);
+
+    switch (mode) {
+#ifndef MCU_CC2538
```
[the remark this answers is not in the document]
gschorcht: To fix compilation errors due to this enumeration type :worried: https://github.com/RIOT-OS/RIOT/blob/fea525f01921abebb195ec8895d67d528f21aa96/cpu/cc2538/include/periph_cpu.h#L150-L158 where `GPIO_OD` and `GPIO_OD_PU` are declared as undefined. I didn't take a look at the implementation to see if these values are used directly as register values. But it wouldn't make any difference, because the special case would still have to be considered.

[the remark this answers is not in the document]
benpicco: But is `GPIO_OD` the same as `GPIO_IN`? Otherwise you can just let this case be hanled by `default` and not support it.

drivers/pcf857x/pcf857x.c
```diff
@@ -0,0 +1,470 @@
...
+
+    assert(irq_event != NULL);
+   _acquire(irq_event->dev);
+    _update_state(irq_event->dev);
+}
+
+/*
+ * @warning: It is expected that the I2C bus is already acquired when the
+ *           function is called. However, it is released by this function
+ *           before the function returns.
```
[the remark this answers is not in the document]
gschorcht: > Seems like an odd choice

Yeah.

> why not call _release() after calling this function? 

The reason is that this function is also called if interrupts are used and the state of a pin changes. If the application has registered a callback function for this pin by `pcf857x_gpio_init_int`, this callback function is called by `_update` in which the application might require the bus. Therefore, the bus has to be released before calling this callback function.

In the former implementation the bus was released before the `_update` function was called and the `_update` function had to acquire it again. I learned from you that releasing and acquiring the bus between instant successive register accesses is a bad idea :wink:

drivers/include/pcf857x.h
gschorcht: maybe `uint_fast8_t`?

benpicco: Sure, although it seems like it's mostly used in asserts? I guess you can check the code size to see what makes most sense.

drivers/pcf857x/pcf857x.c
```diff
@@ -0,0 +1,470 @@
...
+                 return -PCF857X_ERROR_INV_MODE;
+    }
+
+    int res;
+
+    /* write the mode */
+    pcf857x_data_t data = dev->modes | dev->out;
+    _acquire(dev);
+    if ((res = _write(dev, data)) != PCF857X_OK) {
+        return res;
```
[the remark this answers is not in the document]
gschorcht: It is not missed. The bus is either released in function `_update` which is called here if interrupts are used https://github.com/RIOT-OS/RIOT/blob/046c8cb7c6a36986cf1609cfd9664c7bf67bd1fd/drivers/pcf857x/pcf857x.c#L196 or here if interrupts are not used https://github.com/RIOT-OS/RIOT/blob/046c8cb7c6a36986cf1609cfd9664c7bf67bd1fd/drivers/pcf857x/pcf857x.c#L200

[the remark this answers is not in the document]
benpicco: But you `return` here, so those are not called

drivers/include/pcf857x.h
gschorcht: The following tables show the results. For STM32 it requires about 40 byte more ROM to save 12 or 4 bytes of RAM. However, on AVR it requires about 120 or 170 bytes to save 16 or 4 bytes of RAM. I 'm not sure whether this is a good deal.

16-pin expanders
-----------------------
| STM32     |  text   | data  | bss | dec       |
|:--------------|:---------|:-------|:------|:----------|
| array        | 19716 | 540 | 3772 | 24028   |
| bitmap     | 19752 | 540 | 3760 | 24052   |

| AVR     |  text     | data  | bss  | dec      |
|:----------|:----------|:-------|:-------|:---------|
| array    | 18416 | 2250 | 1682 | 22348 |
| bitmap | 18534 | 2250 | 1666 | 22450  |

8-pin expanders
---------------------
| STM32     |  text   | data  | bss | dec       |
|:--------------|:---------|:-------|:------|:----------|
| array        | 19692 | 540 | 3688 | 23920 |
| bitmap     | 19732 | 540 | 3684 | 23956 |

| AVR     |  text     | data  | bss  | dec      |
|:----------|:----------|:-------|:-------|:---------|
| array    | 18114 | 2250 | 1630 | 21994 |
| bitmap | 18282 | 2250 | 1622 | 22154 |

benpicco: Uh I didn't expect that toggling bits would be so expensive, then keep it the way it is

drivers/pcf857x/pcf857x.c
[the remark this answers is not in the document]
gschorcht: :sunglasses: Ups. Indeed.

drivers/include/pcf857x.h
gschorcht: Used `uint8_t` because it saves some bytes on STM32 without increasing the code.

drivers/pcf857x/pcf857x.c
```diff
@@ -0,0 +1,470 @@
...
+    DEBUG_DEV("pin=%u mode=%u", dev, pin, mode);
+
+    switch (mode) {
+        case GPIO_IN_PD: DEBUG_DEV("gpio mode GPIO_IN_PD not supported", dev, mode);
+                         return -PCF857X_ERROR_INV_MODE;
+        case GPIO_OUT:   dev->modes &= ~(1 << pin); /* set mode bit to 0 */
+                         dev->out   &= ~(1 << pin); /* set output bit to 0 */
+                         break;
+        default:         dev->modes |= (1 << pin); /* set mode bit to 1 */
+                         dev->out   |= (1 << pin); /* set output bit to 1 */
```
benpicco, line 163: Now this also catches `GPIO_IN_PU` and `GPIO_OD`/`GPIO_OD_PU`, is this correct?

drivers/pcf857x/pcf857x.c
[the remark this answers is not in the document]
gschorcht: A bit of reordering to handle everything that is not `GPIO_OUT` as default. The question is whether not supported `GPIO_IN_OD` should lead to an assert or simply a failure.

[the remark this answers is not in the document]
benpicco: As long as `gpio_init` returns an error code, you can do so too

drivers/pcf857x/pcf857x.c
gschorcht: Ups, simultaneous communication.

This I/O expander is a very simple device that doesn't have a direction register. In fact it physically supports only the GPIO modes `GPIO_IN_PU` and `GPIO_OD_PU`:
- Writing a 1 to an expander pin configures the pin as an input, which is pulled up to HIGH by a weak pull-up. When reading the pin, its value then depends on the actual voltage.
- Writing a 0 to an expander pin configures the pin as an output and actively drives the pin to LOW.

All other modes are only considered logically identical and behave only like `GPIO_IN`, `GPIO_OD` and `GPIO_OUT`.

benpicco: It's probably worth adding a comment about that

gschorcht: > Now this also catches GPIO_IN_PU and GPIO_OD/GPIO_OD_PU, is this correct?

This affects only the initial level of the pin. If `GPIO_OUT` is used a 0 is written to the pin which configures the pin as physical output an drives it active LOW. In all other cases a 1 is written to the pin which configures it as physical input an the pin is pulled-up to HIGH by the weak pull-up. It would also be possible to use HIGH as initial level for all modes.

gschorcht: > It's probably worth adding a comment about that - feel free to squash directly

It is documented here https://github.com/RIOT-OS/RIOT/blob/0b9b8c28e5bb9a16ca0b4d88e9ec27dded1b3514/drivers/include/pcf857x.h#L67. But I can add a comment in source code too.

gschorcht: Comment added to source code in commit https://github.com/RIOT-OS/RIOT/pull/10430/commits/d02a87fe0ac10115b8edb977abcb323d5a8be5ac.

gschorcht:
Some small changes / clarifications / enhancements in documentation https://github.com/RIOT-OS/RIOT/pull/10430/commits/2ed6c9a97b989ff4c1716e6de2abc5b598b0bf8f.

benpicco:
Please squash!

benpicco (approved):

gschorcht:
@benpicco Many thanks for reviewing and merging.

Issue #10047, opened 2018-09-26
Make warns to expect errors when disabling optional modules


SemjonKerner:
When testing  #10005 I got a warning from Make to **EXPECT ERRORS!** when compiling with `DISABLE_MODULE=periph_gpio_irq`.
This warning is needless since the feature `periph_gpio_irq` is optional and thus excluding it should not throw any errors. Even more, if it throws errors then there is a bug that should neither be expected nor in the code at all.

Meaning if a feature added with `FEATURES_OPTIONAL` is excluded again with `DISABLE_MODULE`, this warning should not be printed.

cladmi:
The `DISABLE_MODULE` should only disable `DEFAULT_MODULES` according to the documentation: https://github.com/RIOT-OS/RIOT/blob/8e9c28dac3fbbf7aede23d5078f29d722e9daf7f/makefiles/vars.inc.mk#L18

Later, a warning was introduced by https://github.com/RIOT-OS/RIOT/pull/1734 which, against the documentation, stated that `DISABLE_MODULE` could disable any module.

And then, later, in conformity to the warning but still against the documentation, the module was indeed removed by https://github.com/RIOT-OS/RIOT/pull/2951 (for a PR that I think at the end did not use it as it change to use the 'default_' module name pattern https://github.com/RIOT-OS/RIOT/pull/2947#issuecomment-101750001)

For me the BUG is that this variable removes the module. It is even problematic currently as `USEMODULES` is modified after

----

Also, `PERIPH_OPTIONAL` is defined as 'include the feature if provided', not you may not use it even if it is available.
https://github.com/RIOT-OS/RIOT/blob/8e9c28dac3fbbf7aede23d5078f29d722e9daf7f/Makefile.dep#L803

In that case, not including the module should be removing `FEATURES_OPTIONAL`.

cladmi:
FEATURES and MODULES are in different concepts. So no `MODULE` should not touch `PERIPH`.

The build system nicely makes all `periph_` features transformed into a dependency to a module of the same name module but it is a really specific case. There are other features that are not transformed directly as modules.

If there should be something to disable optional only features, there should be a dedicated variable 
`FEATURES_OPTIONAL_DISABLE` and it would make sense. Because the `optional` inclusion could be set by something else than the application makefile and you could, because of `FEATURES_CONFLICT` or anything select to not use it.
However, if anything says `FEATURES_REQUIRED` it would still be included.


[...]

PR #10434, opened 2018-11-20
cpu/sam0_common: provide periph_gpio/gpio_irq features at cpu level

aabadie:
### Contribution description

This PR is moving the `FEATURES_PROVIDED += periph_gpio periph_gpio_irq` from each sam0 based boards level to the `cpu/sam0_common` level.

Since there's nothing to configure at the board level for this CPU feature, I think it makes sense to centralize this at the cpu level.

In the meantime, this PR is also fixing the periph_gpio_irq feature that was missing for samr30-xpro (in fact I thought of this PR while I discovered this).

Note that this is also done in #10243 for nrf51 based boards. There might be other architectures concerned.

I hope I didn't miss other sam0 based boards.

### Testing procedure

If Murdock is passing, all should be good.

### Issues/PRs references

None

### DIFF
--- a/boards/arduino-zero/Makefile.features
+++ b/boards/arduino-zero/Makefile.features
@@ -1,5 +1,4 @@
 # Put defined MCU peripherals here (in alphabetical order)
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += periph_i2c
 FEATURES_PROVIDED += periph_pwm
 FEATURES_PROVIDED += periph_rtc
--- a/boards/common/arduino-mkr/Makefile.features
+++ b/boards/common/arduino-mkr/Makefile.features
@@ -1,6 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
 FEATURES_PROVIDED += periph_adc
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += periph_i2c
 FEATURES_PROVIDED += periph_pwm
 FEATURES_PROVIDED += periph_rtc
--- a/boards/feather-m0/Makefile.features
+++ b/boards/feather-m0/Makefile.features
@@ -1,6 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
 FEATURES_PROVIDED += periph_adc
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += periph_i2c
 FEATURES_PROVIDED += periph_pwm
 FEATURES_PROVIDED += periph_rtc
--- a/boards/samd21-xpro/Makefile.features
+++ b/boards/samd21-xpro/Makefile.features
@@ -1,6 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
 FEATURES_PROVIDED += periph_adc
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += periph_i2c
 FEATURES_PROVIDED += periph_pwm
 FEATURES_PROVIDED += periph_rtc
--- a/boards/saml21-xpro/Makefile.features
+++ b/boards/saml21-xpro/Makefile.features
@@ -1,6 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
 FEATURES_PROVIDED += periph_adc
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += periph_i2c
 FEATURES_PROVIDED += periph_rtc
 FEATURES_PROVIDED += periph_rtt
--- a/boards/samr21-xpro/Makefile.features
+++ b/boards/samr21-xpro/Makefile.features
@@ -1,6 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
 FEATURES_PROVIDED += periph_adc
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += periph_i2c
 FEATURES_PROVIDED += periph_pwm
 FEATURES_PROVIDED += periph_rtc
--- a/boards/samr30-xpro/Makefile.features
+++ b/boards/samr30-xpro/Makefile.features
@@ -1,6 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
 FEATURES_PROVIDED += periph_adc
-FEATURES_PROVIDED += periph_gpio periph_gpio_irq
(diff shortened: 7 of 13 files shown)


jcarrano:
It could happen that a board does not have any GPIO (that can be used as such) connected. Probably not with the boards we have in the RIOT tree, as they are all development boards, but it can happen with "real" hardware.

bergzand:
> It could happen that a board does not have any GPIO (that can be used as such) connected. Probably not with the boards we have in the RIOT tree, as they are all development boards, but it can happen with "real" hardware.

Furthermore, `periph_gpio` and `periph_gpio_irq` are not only used for exposed gpio pins, but also for board "internal" pins connected to peripherals on the board itself. Of course it could always happen that somebody creates a board without any GPIO, but then as @dylad suggests, it could always be disabled again at board level.

jcarrano:
Ok, I didn't know it could be disabled.

aabadie:
It seems the change here makes sense finally. Is there someone to give an ACK ?

cladmi:
> it could always be disabled again at board level.

Hmm not really. For me, doing `filter-out` is not a way of disabling things.

For this specific case, it looks like the `NUMOF_IRQS` has a fixed value in `cpu/sam0/periph/gpio.c`.
So is there any need for `#include "periph_conf.h"` in this file? So the board may have no impact in this one.

And for `drivers/include/periph/gpio.h`, if the board has a role in the configuration and **must** provide configuration for it, then it is the board responsibility to do `FEATURES_PROVIDED`.
That is implementation specific.

Can I get confirmation for this ?

dylad:
> Hmm not really. For me, doing filter-out is not a way of disabling things.

`DISABLE_MODULE += periph_gpio_irq` should do the job ?

cladmi:
> DISABLE_MODULE += periph_gpio_irq should do the job ?

Nope

#### Regarding modules

`DISABLE_MODULE` is designed to and has implementation to only correctly support disabling `DEFAULT_MODULES` that do not include any other modules.
It should only configure the policy of automatically including `DEFAULT_MODULES`.

https://github.com/RIOT-OS/RIOT/blob/78164977d83929c5a225b273c8b73b64f2ed98fc/makefiles/vars.inc.mk#L19

There is a line that, only when calling `all` or `flash` (I added `flash` only two month ago for other reasons), also tries to remove them after many dependencies of the disabled modules could already be included
https://github.com/RIOT-OS/RIOT/blob/78164977d83929c5a225b273c8b73b64f2ed98fc/Makefile.include#L585-L593
If you build with `elffile` then it does not work as you expect.

If you remove `periph_rtc` for a build with the `efm32` cpu, `periph_rtc_series$(EFM32_SERIES)` will still be included.

https://github.com/RIOT-OS/RIOT/blob/78164977d83929c5a225b273c8b73b64f2ed98fc/cpu/efm32/Makefile.dep#L1-L3

In general this could also already have triggered adding `LINKFLAGS` or `CFLAGS` in a `board/BOARD/Makefile.include` or `cpu/CPU/Makefile.include`.

In that case, I think `periph_gpio_irq` does not include other modules or triggers other specific compiler options, so should not harm for this specific one regarding modules, but is wrong in general.

#### Regarding `FEATURES` as that is the topic

If you remove the module, you still have `FEATURES_PROVIDED` that contains `periph_gpio_irq`.
If you do it globally in a board, you then have a board that tells the build system it provides a feature, so is automatically selected when looking for compatible boards, but does not provide an implementation for the features it provides.

If you only do it on the command line, then in the specific case of `periph_gpio_irq` it should work as you expect, but is still wrong in general.


#### Documentation

I am trying to summarize all I gathered and understood over time on FEATURES, and want to put it at some point in a documentation. I am currently still sorting out things in my head and explaining what I understood helps me do this.

Other pointers where I tried to explain what I understood:

https://github.com/RIOT-OS/RIOT/issues/10047
https://github.com/RIOT-OS/RIOT/pull/10062#issuecomment-430322739

Original commit describing features

* https://github.com/lebrush/RIOT/commit/f49bd3e6608e7fdfffb8a196123259ce8acde4c3

cladmi:
> You are right, the include of periph_conf.h in drivers/include/periph/gpio.h and cpu/sam0_common/periph/gpio.c can be removed without breaking anything.

Just thinking, it may break as `periph_conf.h` includes `periph_cpu.h` even if the config does not come from the board, it is the entry point.

> I'm still unsure (even if I'm the author of this PR) if it's best to let the board or the cpu to provide the FEATURE_PROVIDED += periph_gpio periph_gpio_irq

A first good thing, is to update the `samr30-xpro` to provide the `periph_gpio_irq` in a separate PR. The other part is the part to discuss.

If for any reason, a board would not connect one of the `periph_gpio_irq` pin (or any other configuration for this implementation, like pullup pulldown something (no idea what it really means :D)), then it make sense to put it in the board (or any other configuration for this module).
If the board has no role in the configuration, it could go in the cpu. That is a hardware and implementation related question.


A good example @kYc0o told me, because I was also thinking about putting them in the cpu.
Is that some cpu could provide 4 uarts, but none are connected on the board. The hardware is in the cpu, the implementation is in the cpu directory, but the hardware is not available to use on the board. So it makes sense that the configuration goes in the board and also the `FEATURES_PROVIDED`.

aabadie:
> So it makes sense that the configuration goes in the board and also the FEATURES_PROVIDED.

Ok, I opened #10553 to fix the missing provided feature for samr30-xpro, to not block this board for too long.
But I also noticed that [kinetis](https://github.com/RIOT-OS/RIOT/blob/master/cpu/kinetis/Makefile.features#L4) and [nrf51](https://github.com/RIOT-OS/RIOT/blob/master/cpu/nrf51/Makefile.features#L1) are providing this feature at CPU level. Maybe we should also move this statement at board level for these CPUs (not here of course) ?

aabadie:
And feel free to close this PR if you think it's not useful or inline with RIOT designs.

haukepetersen:
no, don't close this! This PR makes absolute sense.

The `FEATURE_PROVIDED += periph_gpio(_irq)` simply expresses, that the CPU/board combination is able to provide the GPIO features. As this feature does not require any board specific configuration, it can automatically always be used on this CPU, independent of how the board actually maps its pins. So it actually should be defined in the CPU configuration.

Also remember: as long as the feature is not actively selected (i.e. `USEMODULE += periph_gpio`), it is not compiled... So if there is some application for one of these boards, that does not use any GPIO, the periph GPIO driver is simply not compiled in... This means also, that there is no need for ever actively disabling this feature on a board level!

leandrolanzieri:
The `sodaq-sara-aff` should be added now as well.

aabadie:
Rebased and also updated sodaq-sara-aff board.

PeterKietzmann:
Missing board was added, PR was ACKed, green lights shine. ACK & go

PR #10342, opened 2018-11-06
makefiles/tools/serial: Check empty port

jcarrano:
___Please note, the github UI shows commits in the wrong order.___

### Contribution description

By ensuring the PORT auto-detection worked, we can give meaningful error messages and fail earlier.

### Remarks

Hopefully this serves to illustrate why gratuitous usage of `export` and `:=` should be avoided in makefiles. 

I'm thinking on putting these kind of helper functions in a separate `utils.mk` modules.

### Testing procedure

Try to `make term` for `BOARD=cc2538dk` without any connected board and see how it fails. You should still be able to `make` with no problems or warnings.

### Issues/PRs references

Parallel to #7695.
Contains some changes by (and made on request of, and with assistance of) @cladmi 
Depends on ~#10439~, ~#10440~

PR #10439, opened 2018-11-20
makefile/utils: Add a function for checking that a string is not empty.

jcarrano:
### Contribution description

A call to `$(call ensure_value,x,y)` will fail with message y if x is empty, and otherwise return x. This can be useto write more compact makefiles, while still producing friendly error messages.

### Testing procedure

I should provide an automated test for this, in the meantime:

```
$ cd makefiles/util
$ make -f test-checks.mk test-ensure_value
test-ensure_value
$ make -f test-checks.mk test-ensure_value-negative
test-checks.mk:7: *** "This should fail".  Stop.
```

### Issues/PRs references

Split from #10342.

### DIFF
--- a/makefiles/utils/checks.mk
+++ b/makefiles/utils/checks.mk
@@ -1,0 +1,11 @@
+# Utilities to produce errors inside Make
+# Use this functions to produce friendlier error messages.
+
+# Produce an error if the value is empty
+#
+# Parameters
+#	value: a string which should not be empty
+#	message: The error message to display.
+# Returns:
+#	the first argument, if it is not empty.
+ensure_value = $(if $(1),$(1),$(error $(2)))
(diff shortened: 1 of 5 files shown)


cladmi:
I added a test and updated the behavior of your test to verify the returned value of `ensure_value`.
The test prints nothing in case of success.

It can be tested by replacing, in the Makefile, the tested target by the other: `{,_negative}`.
Verifying that the check tests `ensure_value` returned value can be done by replacing it in `checks.mk`.


On the naming, I would prefer having the directory named `utils` instead of using the singular form.
I still kept my script with your convention.

jcarrano:
I renamed it to `utils` and squashed.

cladmi:
Nice catch on the typo in the error message.

Codacy wants one line less in the README… you can inline fix it directly.
The commit titles mention `makefile/util` and `makefiles/util`, please update them to `makefiles/utils` at the same time.

I re-ran the tests and verified that they also detected the errors by tweaking the test

jcarrano:
Done. Stupid codacy.

PR #10450, opened 2018-11-21
pkg/nimble: Add `ble_hs_stop.c` to makefile

ccollins476ad:
### Contribution description

This file (`ble_hs_stop.c`) was added to NimBLE in this PR: https://github.com/apache/mynewt-nimble/pull/216.  This PR adds an entry for this file to the `pkg/nimble/nimble.host.mk` Makefile.

Without this change, an attempt to build the nimble package fails with a linker error, as seen here: https://travis-ci.org/apache/mynewt-nimble/jobs/458153788

### Testing procedure

Simply ensure the `nimble_gatt` example builds, e.g.,
```
cd examples/nimble_gatt && make
```

### DIFF
--- a/pkg/nimble/nimble.host.mk
+++ b/pkg/nimble/nimble.host.mk
@@ -26,6 +26,7 @@ SRC += ble_hs_mqueue.c
 SRC += ble_hs_misc.c
 SRC += ble_hs_pvcy.c
 SRC += ble_hs_startup.c
+SRC += ble_hs_stop.c
 SRC += ble_ibeacon.c
 SRC += ble_l2cap.c
 SRC += ble_l2cap_coc.c


kaspar030:
@ccollins476ad welcome to the RIOT community and thanks for your contribution!

The nimble commit currently linked in pkg/nimble/Makefile doesn't include this file. I guess the fix would be necessary when bumping the nimble version?

ccollins476ad:
Thanks, @kaspar030.  I am not especially familiar with RIOT.  It seems I should have done a bit more research before submitting this PR!  The failing Travis jobs in the nimble repo were the motivation for this PR.  I was a bit surprised that a change to a foreign repo (nimble) could break RIOT; it seems I should have followed up on this suspicion.

The Murdock job for this PR is failing because it can't find `ble_hs_stop.c`, which makes sense in light of your comment above.  I think it makes the most sense if I close this PR and investigate the proper way to fix the nimble travis failures.

Thanks again.

PR #10473, opened 2018-11-26
board/*/flash.sh: get file to flash from command line

cladmi:
### Contribution description

For the `flash.sh` flasher, get the file to flash from command line instead of environment variable

This is a prepare step to move to having a `FLASHFILE` variable.

### Testing procedure

Flash with the given boards as it is hard to test only the output.
Or review the changes and check to put debug in the flash scripts.

I self tested `calliope-mini` and `mbed_lpc1768` in https://github.com/RIOT-OS/RIOT/pull/8838#issuecomment-382414904 but it is not an external test.

### Issues/PRs references

Split from https://github.com/RIOT-OS/RIOT/pull/8838


aabadie:
Let's wait for Murdock before merging

cladmi:
Thank you for the review ! :+1:

PR #10476, opened 2018-11-27
boards/nrf51: remove ADC_NUMOF define when not needed

aabadie:
### Contribution description

This PR is removing the ADC define from nrf51 based boards that don't provide this feature (yet).

### Testing procedure

If Murdock passes, it should be all good.

### Issues/PRs references

Found with @haukepetersen's review in #10409 (see https://github.com/RIOT-OS/RIOT/pull/10409#discussion_r236325297).


haukepetersen:
and go.

PR #10478, opened 2018-11-27
frdm: Use configuration for OpenOCD 0.10.0 or newer by default

gebart:
### Contribution description

Changes the default setting for the variable USE_OLD_OPENOCD to 0. This variable should be set to 1 only if running OpenOCD 0.9.0 or older. Most distros have moved on to OpenOCD 0.10.0 now.

### Testing procedure

Requires a frdm-k22f or frdm-k64f board. (The frdm-kw41z already forces USE_OLD_OPENOCD=0 because of missing support for that CPU in older versions.)
Build and flash an application for BOARD=frdm-k22f or BOARD=frdm-k64f using `make flash` without any extra environment settings. 
Verify that flashing succeeds and the application runs.

### DIFF
--- a/boards/common/frdm/Makefile.include
+++ b/boards/common/frdm/Makefile.include
@@ -8,10 +8,10 @@ export DEBUG_ADAPTER ?= dap
 #export DEBUG_ADAPTER ?= jlink
 
 # OpenOCD v0.10.0 and newer have built-in support for disabling the Kinetis
-# watchdog automatically. However, current stable releases of Ubuntu and Debian
-# have only version 0.9.0 and older OpenOCD packages (Ubuntu 17.04, Debian Jessie)
-# Set this to 0 to avoid the extra manual step of disabling the watchdog.
-export USE_OLD_OPENOCD ?= 1
+# watchdog automatically. Some older releases of Ubuntu and Debian have only
+# version 0.9.0 or earlier OpenOCD packages (Ubuntu 17.04, Debian Jessie)
+# Set this to 1 if you are using one of these older releases.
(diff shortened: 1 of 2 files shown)


boards/common/frdm/Makefile.include
```diff
@@ -8,10 +8,10 @@ export DEBUG_ADAPTER ?= dap
...
 
 # OpenOCD v0.10.0 and newer have built-in support for disabling the Kinetis
-# watchdog automatically. However, current stable releases of Ubuntu and Debian
-# have only version 0.9.0 and older OpenOCD packages (Ubuntu 17.04, Debian Jessie)
-# Set this to 0 to avoid the extra manual step of disabling the watchdog.
-export USE_OLD_OPENOCD ?= 1
+# watchdog automatically. Some older releases of Ubuntu and Debian have only
+# version 0.9.0 or earlier OpenOCD packages (Ubuntu 17.04, Debian Jessie)
+# Set this to 1 if you are using one of these older releases.
+export USE_OLD_OPENOCD ?= 0
```
aabadie: Slightly out of the scope of this PR but I think the `export` is not needed, it's only used my make.

It would also be good to remove the equivalent line in the frdm-kw41z `Makefile.include`. Could be done in an extra commit.

gebart: Will do 👍

gebart: addressed

MrKevinWeiss:
Also @gebart please squash :)

gebart:
squashed

PR #10479, opened 2018-11-27
openocd: Use reset halt when starting debug-server

gebart:
### Contribution description

Makes the debug-server start up more robust, especially when combined with `reset_config connect_assert_srst` in the OpenOCD configuration.

### Testing procedure

With any board with an onboard debugger interface and using OpenOCD:
```
make debug-server
```
Verify that the OpenOCD instance can connect to the board and that the application running is halted (e.g. does not respond to UART shell commands anymore, software controlled LEDs stop blinking etc.)

in a different terminal:

```
make debug
```
(This will attempt to start another instance of openocd, but that will fail since the same UDP port is already used by the running debug-server instance)

See that GDB works as expected:
```
continue
```
Verify that the application is running.

Press CTRL+C in GDB, verify that GDB has managed to halt the application.

### DIFF
--- a/dist/tools/openocd/openocd.sh
+++ b/dist/tools/openocd/openocd.sh
@@ -86,10 +86,10 @@
 # Debugger flags, will be passed to sh -c, remember to escape any quotation signs.
 # Use ${DBG_DEFAULT_FLAGS} to insert the default flags anywhere in the string
 : ${DBG_FLAGS:=${DBG_DEFAULT_FLAGS} ${DBG_EXTRA_FLAGS}}
-# Initial target state when using debug, by default a 'halt' request is sent to
-# the target when starting a debug session. 'reset halt' can also be used
+# Initial target state when using debug, by default a 'reset halt' request is
+# sent to the target when starting a debug session. 'halt' can also be used
 # depending on the type of target.
-: ${OPENOCD_DBG_START_CMD:=-c 'halt'}
+: ${OPENOCD_DBG_START_CMD:="-c 'reset halt'"}
 # command used to reset the board
 : ${OPENOCD_CMD_RESET_RUN:="-c 'reset run'"}
 # This is an optional offset to the base address that can be used to flash an
@@ -311,7 +311,7 @@ do_debugserver() {
             -c 'gdb_port ${GDB_PORT}' \
             -c 'init' \
             -c 'targets' \
-            -c 'halt'"
+            ${OPENOCD_DBG_START_CMD}"
 }
 
 do_reset() {
--- a/boards/b-l072z-lrwan1/Makefile.include
+++ b/boards/b-l072z-lrwan1/Makefile.include
@@ -12,8 +12,5 @@ include $(RIOTMAKE)/tools/serial.inc.mk
 # this board has an on-board ST-link adapter
 export DEBUG_ADAPTER ?= stlink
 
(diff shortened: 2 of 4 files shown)


dist/tools/openocd/openocd.sh
```diff
@@ -308,7 +308,7 @@ do_debugserver() {
             -c 'gdb_port ${GDB_PORT}' \
             -c 'init' \
             -c 'targets' \
-            -c 'halt'"
+            -c 'reset halt'"
```
aabadie: Instead of this, maybe we could reuse `${OPENOCD_DBG_START_CMD}` like in `do_debug`. This allows boards requiring this to override start command.

gebart:
Modified to reuse OPENOCD_DBG_START_CMD for both debug and debug-server. Changed the default command to be `reset halt` instead of only `halt`

aabadie:
To move forward on this, I allowed myself to rebase this PR and slightly update it: variable documentation is updated in openocd.sh and the default value used for this variable in a some boards (f1 based + b-l072Z-lrwan1 boards) is removed from the makefile, since now it's using the openocd default value.

@cladmi, are interested in having a look ?

cladmi:
I can try on a few boards indeed.

Is this saying we are indeed going toward putting `connect_assert_srst` by default ?
I have some waiting commits to enable it on some boards https://github.com/RIOT-OS/RIOT/pull/10870/commits
If you have an easy way (even arch specific) to put a board into a non flashable mode without this option, I would gladly add a test for it. I know the `tests/driver_adt7310` test was a good one to break some of the mentioned boards.

kaspar030:
This means running "make debug" would always reset a board? do we want that?

I sometimes realize a board is stuck somewhere, then connect the debugger to see where it got stuck.

aabadie:
> This means running "make debug" would always reset a board? do we want that?

By default yes, but one can still have the non reset behaviour by passing `OPENOCD_DBG_START_CMD="-c 'halt'"` to the command line.

cladmi:
@kaspar030 currently there are many cases where doing `make flash` does not flash because the CPU is in an un-handled state. This breaks all testing possibilities and I would put it with a bigger priority even if our default configuration also resets when debugging.

kaspar030:
> @kaspar030 currently there are many cases where doing `make flash` does not flash because the CPU is in an un-handled state.

Then add the corresponding reset halt to the flash command. I don't understand what that has to do with starting the gdb server.

aabadie:
There is a confusion here. This PR is just about debug. Having reset halt before starting the debug-server fixes gdb connection when `connect_assert_srst` is used by the openocd configuration. This option is useful when the CPU is in sleep mode (otherwise OpenOCD is not able to reflash it without manual interaction)

fjmolinas:
Is this one still relevant? with OPENOCD_RESET_USE_CONNECT_ASSERT_SRST there is no issue when launching `make debug`, @aabadie do you agree?

fjmolinas:
ping @aabadie!

leandrolanzieri:
ping!

benpicco:
I agree with @kaspar030 here. I too will use `make debug` to see where the code got stuck, the default behavior shouldn't be to reset the CPU here.

If you want to start debugging in a defined state, how about adding a `reset-debug` target instead?

PR #10484, opened 2018-11-27
atmega: fix periph timer configuration

smlng:
### Contribution description

- correct number of timers for atmega328p from 2 to 1
- correct number of timer channels for atmega328p from 3 to 2
- adapt atmega periph timer implementation accordingly

look at the data sheet of the atmega328p you'll find that timer1 has only 2 compare registers, while timer1 on the atmega2560 has 3 compare registers -> channels.

### Testing procedure

In theory: run `tests/periph_timer` for Arduino-uno, however that is not possible without modifying the test to make it fit and run on that board. But you could use an mega2560 to verify that is still working too.

### DIFF
--- a/boards/common/arduino-atmega/include/periph_conf.h
+++ b/boards/common/arduino-atmega/include/periph_conf.h
@@ -49,7 +49,8 @@ extern "C" {
  * @{
  */
 #ifdef CPU_ATMEGA328P
-#define TIMER_NUMOF         (2U)
+#define TIMER_NUMOF         (1U)
+#define TIMER_CHANNELS      (2)
 
 #define TIMER_0             MEGA_TIMER1
 #define TIMER_0_MASK        &TIMSK1
@@ -60,6 +61,7 @@ extern "C" {
 
 #ifdef CPU_ATMEGA2560
 #define TIMER_NUMOF         (2U)
+#define TIMER_CHANNELS      (3)
 
 #define TIMER_0             MEGA_TIMER1
 #define TIMER_0_MASK        &TIMSK1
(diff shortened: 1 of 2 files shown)


MrKevinWeiss:
When you say the correct number of timers are you talking about implemented and free on RIOT or available in hardware...  As far as I can see the atmega328p has 3 hardware timers (of varying qualities though, two 8 bit and 1 16 bit).

The atmega2560 isn't happy too
```
2018-11-28 15:00:52,042 - INFO # Testing TIMER_0:
2018-11-28 15:00:52,087 - INFO # TIMER_0: ERROR on initialization - skipping
2018-11-28 15:00:52,088 - INFO # 
2018-11-28 15:00:52,088 - INFO # 
2018-11-28 15:00:52,108 - INFO # Testing TIMER_1:
2018-11-28 15:00:52,153 - INFO # TIMER_1: ERROR on initialization - skipping
2018-11-28 15:00:52,153 - INFO # 
2018-11-28 15:00:52,154 - INFO # 
2018-11-28 15:00:52,165 - INFO # TEST FAILED
```
but I get the same results in master.

Also some examples I think need to be updated as per murdock request.

smlng:
I meant the number of timers configured and available in RIOT, which is 1 for 328p and 2 for mega2560- that's what counts at the moment for RIOT, not how many timers are available or offered by the hardware. 

For testing: you need to set TIMER_SPEED in the test app to `250000` (250kHz) because the prescaler of the atmegas cannot match 1MHz.

MrKevinWeiss:
Yup passes now (and on master).  I will do a quick test to try to see the failure in the master with the uno.  I do think it makes sense though.

smlng:
@MrKevinWeiss may I squash?

smlng:
> If you can make murdock happy I think we are good.

I'm trying see #10503

PR #10487, opened 2018-11-27
board: NXP Kinetis USB-KW41Z IoT USB dongle

gebart:
### Contribution description

Add support for [NXP USB-KW41Z development board](https://www.nxp.com/products/wireless/bluetooth-low-energy-ble/bluetooth-low-energy-ieee-802.15.4-packet-sniffer-usb-dongle:USB-KW41Z).
This is the perfect board for running a border router, it has the form factor of a USB memory stick and features a 802.15.4 radio (#7107 for a netdev driver) and can be debugged and flashed without any external tools.

### Testing procedure

The stick needs to be changed from the factory programmed sniffer application to an OpenSDA application for flashing and debugging. Follow [the Quick Start Guide at NXP.com](https://www.nxp.com/products/wireless/bluetooth-low-energy-ble/bluetooth-low-energy-ieee-802.15.4-packet-sniffer-usb-dongle:USB-KW41Z?&tab=In-Depth_Tab&tid=van/usb-kw41z/startnow), click on "2. Get Software" -> "Choose a Development Path: USB-KW41Z Development Platform", scroll down to "2.3 Load Segger JLink OpenSDA App", follow the instructions.

Build a test application, flash and run.
Test make flash and verify that the application runs.
Test make debug and verify that GDB can control the device.

### Issues/PRs references

Depends on ~#10486~ because this board design uses the NMI pin as a GPIO
Depends on ~#10478~ implicitly, Workaround: Set USE_OLD_OPENOCD=0 in the environment.
#7107 provides support for the radio

### DIFF
--- a/boards/usb-kw41z/Makefile.features
+++ b/boards/usb-kw41z/Makefile.features
@@ -1,0 +1,15 @@
+# Put defined MCU peripherals here (in alphabetical order)
+FEATURES_PROVIDED += periph_adc
+FEATURES_PROVIDED += periph_i2c
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
+FEATURES_PROVIDED += periph_spi
+FEATURES_PROVIDED += periph_timer
+FEATURES_PROVIDED += periph_uart
+
+# The board MPU family (used for grouping by the CI system)
+FEATURES_MCU_GROUP = cortex_m0_2
+
+include $(RIOTCPU)/kinetis/Makefile.features
+# Remove this line after TRNG driver is implemented
+FEATURES_PROVIDED := $(filter-out periph_hwrng,$(FEATURES_PROVIDED))
--- a/boards/usb-kw41z/board.c
+++ b/boards/usb-kw41z/board.c
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2018 Eistec AB
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser General
+ * Public License v2.1. See the file LICENSE in the top level directory for more
+ * details.
+ */
+
+/**
+ * @ingroup     boards_usb-kw41z
+ * @{
+ *
+ * @file
+ * @brief       Board specific initialization for the USB-KW41Z
+ *
+ * @author      Joakim Nohlgård <joakim.nohlgard@eistec.se>
+ *
+ * @}
+ */
+
+#include "board.h"
+#include "periph/gpio.h"
+
+void board_init(void)
+{
+    /* initialize the CPU core */
+    cpu_init();
+
+    /* initialize and turn off LEDs */
+#ifdef LED0_PIN
+    gpio_init(LED0_PIN, GPIO_OUT);
+    gpio_set(LED0_PIN);
+#endif
+#ifdef LED1_PIN
+    gpio_init(LED1_PIN, GPIO_OUT);
+    gpio_set(LED1_PIN);
+#endif
+}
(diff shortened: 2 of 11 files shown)


aabadie:
Nice! I have of couple of them, so I can review and test this PR :)

aabadie:
Tested it and seems to work :)

I had to use `USE_OLD_OPENOCD=0` to flash the board by the way.

boards/usb-kw41z/board.c
```diff
@@ -0,0 +1,38 @@
...
+#include "board.h"
+#include "periph/gpio.h"
+
+void board_init(void)
+{
+    /* initialize the CPU core */
+    cpu_init();
+
+    /* initialize and turn off LEDs */
+#ifdef LED0_PIN
```
aabadie: `LED0_PIN` is always defined so this `#ifdef` is not really required?

boards/usb-kw41z/include/gpio_params.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @brief    LED configuration
+ */
+static const  saul_gpio_params_t saul_gpio_params[] =
+{
+    /* There are three LEDs on board, but only two are software controllable,
+     * and one is driven by the debugger interface unless hardware mods have
+     * been applied. See boards/usb-kw41z/include/board.h */
+#ifdef LED0_PIN
```
aabadie: Same here, about the `#ifdef`

boards/usb-kw41z/Makefile.features
```diff
@@ -0,0 +1,15 @@
...
+FEATURES_PROVIDED += periph_adc
+FEATURES_PROVIDED += periph_i2c
+FEATURES_PROVIDED += periph_rtc
+FEATURES_PROVIDED += periph_rtt
+FEATURES_PROVIDED += periph_spi
+FEATURES_PROVIDED += periph_timer
+FEATURES_PROVIDED += periph_uart
+
+# The board MPU family (used for grouping by the CI system)
+FEATURES_MCU_GROUP = cortex_m0_2
```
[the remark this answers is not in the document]
gebart: opened #10491

gebart:
> Maybe add jlink programmer as an alternative ?

How do you mean? The default config is set to use jlink by default since that is the only OpenSDA application I have found for this board.
That same configuration also works for external JLink debuggers if you set DEBUG_ADAPTER_ID="your Jlink serial number" in the environment, no other change is necessary.

> I compared the peripheral configuration with frdm-kw41z and they are almost the same. It would be great to provide a common part for both boards (for uart, i2c, clock, timer and makefiles)

This is something that is similar for most Kinetis boards and it would be possible to create some kind of generic default configuration for all things which do not depend on external pins, but it is outside the scope of this PR IMO.

gebart:
> - `tests/periph_timer`: failed, Timer1 (LPTMR) cannot be initialized

LPTMR only supports 32768 Hz, you need to modify the test application to use that for timer_init.

aabadie:
> How do you mean? The default config is set to use jlink by default since that is the only OpenSDA application I have found for this board.

I meant checking the PROGRAMMER variable. It will also work with jlink, the default being openocd. This is not mandatory, just wanted to mention this is also possible.

> This is something that is similar for most Kinetis boards and it would be possible to create some kind of generic default configuration for all things which do not depend on external pins, but it is outside the scope of this PR IMO.

Ok. Would be great to factorize common code in follow-up PRs.

aabadie:
@gebart, all dependency PRs are merged so I removed the "Waiting for other PRs" label and activated Murdock. I think you'll have some application Makefiles to edit to make it succeed.

gebart:
Rebased and fixed up a minor typo in a Doxygen group name, added usb-kw41z to Cortex-M boards in the unit test Makefile (amended). From the Murdock output it seemed like this board works with all tests pretty much without any specific Makefile adaptations.

aabadie:
Re-ACK, and go!

Issue #10269, opened 2018-10-26
Undefined i2c behaviour after release


MrKevinWeiss:
Currently the i2c_acquire and release functions are not consistant between boards.

Should we have access to the bus on startup without i2c_acquire?

Should we be able to access the bus after an i2c_acquire and release?

Currently most boards support access to the bus without i2c_acquire first as it gets initialized in the cpu init.
On the nucleo-l073rz (and I assume other nucleo boards) the bus cannot be accessed after the bus has been acquired and released.  This is not true for the arduinomega-2560, nrf52dk, nucleo-f411re, samr21-xpro.  

Maybe it should work, if so how does this affect the pm layer?

gebart:
i2c_acquire, i2c_release are locking functions to ensure exclusive access to the bus when there is more than one device driver using the i2c bus. Therefore, i2c_acquire must _always_ be called before using any of the i2c_transfer_x functions.
If you find a device driver which uses i2c_transfer without first acquiring the bus, then that is a bug.

I don't understand this part:
>On the nucleo-l073rz (and I assume other nucleo boards) the bus cannot be accessed after the bus has been acquired and released. This is not true for the arduinomega-2560, nrf52dk, nucleo-f411re, samr21-xpro.

Does the nucleo-l073rz i2c stop working after the first i2c_release call? That would be a blocker bug IMO, and must be fixed.

MrKevinWeiss:
@gebart Sorry for not responding.  To clarify, 

> Does the nucleo-l073rz i2c stop working after the first i2c_release call?

Yes, as expected.  From what you described, after being released the bus should **not** work.  All other boards continue to work.

@dylad I did see that in some places but it seems that it gets enabled in the cpu init in many boards.

So from what I understand, the correct operation should be as follows
- Startup the i2c should **not** work
- Acquire i2c should work
- Release i2c should **not** work

If we all agree I will add these as critical tests to the HiL testing and we can start enforcing this rule.  I may even start a tracker to ensure this is happening everywhere.

@aabadie also has experience with this and it would be nice to get his feedback :)

gebart:
Thanks for clarifying. I originally understood it as "no more i2c_acquire calls an be made after the first i2c_release call (bug)", but you meant that no i2c transfers can be made after a call to i2c_release _unless_ a call to i2c_acquire is made first (enforcing correct application behavior), is that the correct interpretation?

MrKevinWeiss:
@gebart Correct

[...]

PR #10490, opened 2018-11-28
periph/i2c: Add precondition on i2c_acquire for i2c bus access

gebart:
### Contribution description

It was suggested in https://github.com/RIOT-OS/RIOT/issues/10269#issuecomment-440932744 that the documentation regarding the requirement to call i2c_acquire before any I2C bus access is not visible enough. This PR adds a precondition note to all i2c_read/write functions saying that i2c_acquire must be called first.

### Testing procedure

make doc
verify the docs for the periph I2C interface are OK.

### Issues/PRs references

#10269


MrKevinWeiss:
Awesome, thanks.

PR #10461, opened 2018-11-23
Makefile.include: building in docker should be done after clean 

cladmi:
### Contribution description

When building in parallel, clean is done at the same time as the
container starts building.

So the parallel `clean all` handling should also be done for the
`..in-docker-container` target.

### Testing procedure


I added a test commit to increase the probability of showing the issue and should be removed before merging. The issue is also present and was found without.

Testing should do a `BUILD_IN_DOCKER=1 make -j clean all`. It could be shown reliably with any example that takes more than 5 seconds to build on your machine.

```
BUILD_IN_DOCKER=1 DOCKER="sudo docker" make -C examples/gnrc_networking -j clean all        
```

With this PR, (should be run when `sudo` does not ask for the password anymore).

```
BUILD_IN_DOCKER=1 DOCKER="sudo docker" make -C examples/gnrc_networking -j clean all                                                                                                                                                                                                  
make: Entering directory '/home/harter/work/git/RIOT/examples/gnrc_networking'                                                                                                                                                                                                            
sleep 2                                                                                                                                                                                                                                                                                   
for i in  ; do "make" -C /home/harter/work/git/RIOT/pkg/$i clean ; done                                                                                                                                                                                                                   
rm -rf /home/harter/work/git/RIOT/examples/gnrc_networking/bin/native                                                                                                                                                                                                                     
rm -rf scan-build/native
Launching build container using image "riot/riotbuild:latest".
sudo docker run --rm -t -u "$(id -u)" \
    -v '/home/harter/work/git/RIOT:/data/riotbuild/riotbase' \
    -v '/home/harter/work/git/RIOT/cpu:/data/riotbuild/riotcpu' \
    -v '/home/harter/work/git/RIOT/boards:/data/riotbuild/riotboard' \
    -v '/home/harter/work/git/RIOT/makefiles:/data/riotbuild/riotmake' \
    -v '/home/harter/work/git/RIOT:/data/riotbuild/riotproject' \
    -v /etc/localtime:/etc/localtime:ro \
    -e 'RIOTBASE=/data/riotbuild/riotbase' \
    -e 'CCACHE_BASEDIR=/data/riotbuild/riotbase' \
    -e 'RIOTCPU=/data/riotbuild/riotcpu' \
    -e 'RIOTBOARD=/data/riotbuild/riotboard' \
    -e 'RIOTMAKE=/data/riotbuild/riotmake' \
    -e 'RIOTPROJECT=/data/riotbuild/riotproject' \
[...]

Issue #10459, opened 2018-11-23
make: `make clean all` does not make sense and should be removed


jcarrano:
#### Description

What does `make clean all` mean? How can one create a target and clean it at the same time? The answer is that it does not make sense. This ill-defined command is used extensively in RIOT.

It could seem (if one has only a superficial understanding of make) that `make clean all` means `make clean` followed by `make all`. Well, that is not how make works, and if it behaves like that in certain circumstances it is only accidental.

What `make` does is execute all recipes needed to make the required targets, in any order, as long as dependencies are satisfied. In the case of parallel make ("-j") recipe execution may be simultaneous.

The reason `make clean all` works at all is because of hacks that make `all` depend on `clean`. It is a hack because under normal circumstances `all` does not depend on `clean`.

#### Why is `make clean all` even used

See also: [this comment](https://github.com/RIOT-OS/RIOT/pull/10456#issuecomment-441082055)

Normally, one would type `make clean` followed by `make all`. The problem is that in the current implementation, the overhead of starting make is too big. This is caused mainly by the extensive use of immediate assignments (`:=`) combined with  `$(shell ...)` which means there are lots of commands run on each time make is invoked.

On a big CI build, this overhead adds up and cleaning ends up taking considerable time. For this reason a hack was devised to make it possible to run both targets in a single invocation.

#### Steps to reproduce the issue

This issue is hard to reproduce, as it involves parallelism (and non-deterministic behaviour). For some examples, see

- [x] #10456 Removing a (seemingly) unrelated rule caused builds to start failing.
- [ ] #10344 This is the rule that got removed.
- [ ] #10461 It also complicates docker integration.
- [x] #10492 I thought I fixed it, but it turns out it is necessary to add conditional dependencies on "clean" for some targets.

#### Steps to fix

Removing all the code needed for `make clean all`, and banning it's usage will make our build system simpler and saner.

For this we must attack the root cause: high overhead, caused by **unnecessary evaluations of variables containing costly shell invocations.** The good news is that that overhead _can be fixed_.

On the other hand, in the case where those variables are needed, they should be evaluated only once.

Another cause of unnecessary evaluations is the indiscriminate use of "export" statements (see #10440 for an example).

gebart:
What do we gain from removing the hack which makes all (and flash) depend on clean? 
What is you use case?
[...]

PR #10492, opened 2018-11-28
mcuboot.mk: Add mcuboot targets to BUILDDEPS.

jcarrano:
### _Read this before reviewing_

The first two commits are taken from #10461 and are __not__ part of this PR. They are there just to force an error.

### Contribution description

The mcuboot targets were not being added to BUILDDEPS and so the mcuboot targets was being made simultaneous with the clean. See the failed murdock output from #10461.

### Testing procedure

In `tests/mcuboot` run

~`BUILD_IN_DOCKER=1 DOCKER="sudo docker" make clean all`~

EDIT: the test command is running this merged with #10461 

```
make -C tests/mcuboot clean all -j
```

### Issues/PRs references

This is one of the issues derived from #10459 .

### DIFF
--- a/makefiles/mcuboot.mk
+++ b/makefiles/mcuboot.mk
@@ -12,6 +12,8 @@ MCUBOOT_BIN_MD5 ?= 0c71a0589bd3709fc2d90f07a0035ce7
 
 export IMAGE_HDR_SIZE ?= 512
 
+$(MCUBOOT_KEYFILE) $(MCUBOOT_BIN): $(filter clean, $(MAKECMDGOALS))
+
 mcuboot-create-key: $(MCUBOOT_KEYFILE)
 
 ifeq ($(BINDIR)/key.pem,$(MCUBOOT_KEYFILE))


cladmi:
I added the `needs squashing` label to prevent merging before removing the test commits.

cladmi:
The test is not enough to show the issue with only the reference commits. I used this one instead:

```
make -C tests/mcuboot clean all -j
```

And with it this does not currently fix it.

cladmi:
The issue is that `BUILDDEPS` is used here for `clean`

https://github.com/RIOT-OS/RIOT/blob/90db0bf253c56190cc598953f64463700339ae36/Makefile.include#L486

and `mcuboot.inc.mk` is included here:

https://github.com/RIOT-OS/RIOT/blob/90db0bf253c56190cc598953f64463700339ae36/Makefile.include#L720-L721

The last line should be moved before. I would imagine after processing all modules like after

https://github.com/RIOT-OS/RIOT/blob/90db0bf253c56190cc598953f64463700339ae36/Makefile.include#L362

The problem was already here before.

cladmi:
It really follows the big issue that variables should be defined first and rules after.

jcarrano:
@cladmi push the fix to my branch,

makefiles/mcuboot.mk
```diff
@@ -12,6 +12,8 @@ MCUBOOT_BIN_MD5 ?= 0c71a0589bd3709fc2d90f07a0035ce7
 
 export IMAGE_HDR_SIZE ?= 512
 
+BUILDDEPS += $(MCUBOOT_KEYFILE) $(MCUBOOT_BIN) mcuboot
```
cladmi: However, we discussed it IRL, none of these files should be declared as `BUILDDEPS` they are not build dependencies, they just need to be done after `clean` if it is there.

The reason being that they are in the `BINDIR` directory that can be deleted.

I will propose a pre-first fix to go in the direction of adding this `BINDIR` variable and target.

cladmi:
After some more discussions and verification, putting a directory as target dependency forces rebuild whenever a file is put inside. The correct way is to declare it as order-only target. But order-only targets and clean at the same time cannot work properly, for good reasons.
There could be a different fix with a file dependency, but it is a different task. So now, I just applied the same handling there too. 

I added an updated commit message in the `squash!` commit body.

My tests commands were verifying the creation commands are indeed done after `clean` for both following commands:

```
make -C tests/mcuboot clean all -j
```
and 
```
make -C tests/mcuboot clean mcuboot-flash -j
```

jcarrano:
Squashed and removed the test commits.

cladmi:
The squashed message is wrong as it still references `BUILDDEPS`. I do not have my previous commit however…

jcarrano:
I amended the commit message.

cladmi:
Commit title still references `BUILDDEPS`.

PR #7601, opened 2017-09-14
cpu/atmega: use power.h defines instead of direct register access

roberthartung:
fixes #7422

PR #7542, opened 2017-08-30
Atmega PWM implementation

lebrush:
Tested on Atmega328p (Arduino nano)

PR #10498, opened 2018-11-28
atmega: add periph_pwm (taken from #7542)

smlng:
### Contribution description

This adapts #7542 to latest master and utilises AVR power functions as in #7601.

I tested with atmega2560 and both configured PWMs with 2 channels each did work, though the allowed parameters are limited, i.e. mode must be left aligned and resolution 256.

### Testing procedure

flash `tests/periph_pwm` on an atmega2560 and run

```
init 0 0 100 256
set 0 0 1
set 0 0 10
set 0 0 100
```

assuming you connected a LED to pwm0 channel 0 (PB7 or PWM13 on atmega2560) the LED should shine in different brightness.

You should also be able to turn off the power with `power 0 0` which should turn off the LED.

### Issues/PRs references

taken from #7542, which wasn't moving forward.

### DIFF
--- a/boards/common/arduino-atmega/Makefile.features
+++ b/boards/common/arduino-atmega/Makefile.features
@@ -9,6 +9,7 @@ FEATURES_PROVIDED += periph_uart
 # Various other features (if any)
 ifeq (,$(filter jiminy-mega256rfr2,$(BOARD)))
   FEATURES_PROVIDED += arduino
+  FEATURES_PROVIDED += periph_pwm
 endif
 
 # The board MPU family (used for grouping by the CI system)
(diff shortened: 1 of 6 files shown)


smlng:
Please note, that this finally brings in the missing PWM periph for the atmega-based Arduinos

MrKevinWeiss:
I think making the CI happy would be good.

MrKevinWeiss:
Though it seems like it is more a murdock problem and not the code...

MrKevinWeiss:
I think we can ignore the codacy problems.

MrKevinWeiss:
To clarify I checked the datasheet for the porting of the GPIO pins, I looked through the code and couldn't find anything obvious that wasn't a problem and tested on the mega2560 LED.

MrKevinWeiss:
Should you squash or is it valuable to separate the commits?

smlng:
I would leave as is, for one that's how the original author @lebrush made them, and second the separation seems okay (to me).

PR #10504, opened 2018-11-29
periph: API change of macros ADC_LINE and DAC_LINE

gschorcht:
### Contribution description

This PR changes the definition of the `ADC_LINE` and `DAC_LINE` macros to make them more consistent with

- the description what they are,
- the definition of the `GPIO_PIN` macro, and
- the definition of devices and channels.

The need to change them came up during the development of the extension APIs for ADC and DAC extension API as required by issue #9582 and already realized for GPIO in PR #9860.

According to the documentation of ADC and DAC

<i>An ADC line in this context is a tuple consisting out of a hardware ADC device (an ADC functional unit on the MCU) and an ADC channel connected to pin.
Similar to the ADC driver interface (ADC), the DAC interface uses the concept of lines, corresponds to a tuple of a DAC device and a DAC output channel.</i>

a line refers to a channel `y` of device `x`. Thus, the `*_LINE` macros should be defined with two parameters `x` and `y`, the device and the channel. According to the default definition of `GPIO_PIN`, which uses the same concept, the definitions should look like:
```
#define ADC_LINE(x,y)        ((adc_t)((x & 0) | y))
#define DAC_LINE(x,y)        ((dac_t)((x & 0) | y))
```
The channels of the MCU, if there are, would then be device 0.

Changing the macros in this way would it make possible to define the `ADC_EXT_LINE` and `DAC_EXT_LINE` macros in the extension APIs in a consistent way with GPIO extension API where `x` refers the extension device and `y` the channel.
```
#define ADC_EXT_LINE(x,y)      ....
#define DAC_EXT_LINE(x,y)      ....
```

### Testing procedure

The following test applications have to pass on any board that support ADC and DAC peripherals:
```
make -C tests/periph_adc BOARD=...
make -C tests/periph_dac BOARD=...
make -C tests/driver_mq3 BOARD=...
make -C tests/driver_io1_xplained BOARD=...
```

### Issues/PRs references

#9582 Issue with proposal for peripheral extension API
#9860 and #9958 Implementation of GPIO extension API


gschorcht:
@ZetaR60 
> Also, I think that an equivalent of `GPIO_PIN(x,y)` for `periph/adc` and `periph/dac` could be added, just not in this way.

Any suggestion?

For me, the intent of the `ADC_LINE` and `DAC_LINE` macros become not really clear from the documentation. No CPU implementation overrides the default macro. All of them just use the `x` in `ADC_LINE(x)` as it is. In fact, `x` simply identifies the `x`-th's channel on first device, the MCU, how ever they are mapped to the ports and the pins internally by the implementation of `periph/adc` and `periph/dac`.

So why not to say, device 0 is the MCU and all other devices are extensions.

Maybe, @haukepetersen could clearify that.

ZetaR60:
I think that another pair of macros could be added that are the inverse functions of ```ADC_LINE(x)``` and ```DAC_LINE(x)```, and they would take two arguments (nth internal device and mth line). Maybe something like ```ADC_LINE_INV(x,y)```, though there is probably a better name. The documentation should also be clarified. Then when adding stuff for the extension interface, we can add a pair of macros that are something like ```ADC_EXT_LINE_INV(x,y)``` that maps external devices and their lines to the ```adc_t``` and ```dac_t``` representation.

We should see what @haukepetersen has to say first though.

haukepetersen:
Sorry for only seeing this now (don't have much time to spare for periph stuff lately)...

I don't like this change, as it defeats the 'line' concept: the idea is, to have from a users perspective only a set of logical lines. It is up to the driver implementation and board configuration, to map these lines onto `ADC dev / channel` tuples. 

E.g. look at https://github.com/RIOT-OS/RIOT/pull/7277 (a little outdated, but can't find any other example right now): here some different channels on different ADC devices are mapped onto logical lines like this:
```c
static const adc_conf_t adc_config[] = {
￼     { .pin = GPIO_PIN(PORT_A,  3), .dev = ADC_1, .chan =  3 },
￼     { .pin = GPIO_PIN(PORT_C,  0), .dev = ADC_1, .chan = 10 },
￼     { .pin = GPIO_PIN(PORT_C,  3), .dev = ADC_1, .chan = 13 },
￼     { .pin = GPIO_PIN(PORT_F,  3), .dev = ADC_3, .chan =  9 },
￼     { .pin = GPIO_PIN(PORT_F,  5), .dev = ADC_3, .chan = 15 },
￼     { .pin = GPIO_PIN(PORT_F, 10), .dev = ADC_3, .chan =  8 }
￼ };
```
Remember, that this config is CPU specific to STM32 devices and it is up to the CPU to define a config format that makes sense...

In general I am pretty convinced that the way forward with the `periph_adc` driver is to leave it 'as is', but introduce another, more advanced and complex interface for support of many more features. Same is true e.g. for PWM (as in dma driver PWM), and possibly other periph drivers as well.

gschorcht:
There was a suggestion from @ZetaR60 for dealing with `ADC_LINE` and `DAC_LINE` macros for extension APIs in https://github.com/RIOT-OS/RIOT/issues/9582#issuecomment-441894765 but I'm not really lucky with the `FOO_EXT_LINE` definition since it becomes quite complex and erro-prone for multiple ADC or DAC extension devices.

gschorcht:
@haukepetersen 
> In general I am pretty convinced that the way forward with the periph_adc driver is to leave it 'as is'

One further question about the existing interface I have is: How should the `ADC_RES_16BIT` resolution be handled on a platform where `int` is only 16 bit and `-1` indicates an error? Theoretically, on such a platform only 15 bit resolution is possible, right?

gschorcht:
This PR is closed in favor of PR #13247 which should allow to integrate external ADC devices. For DAC devices another solution is required.

gschorcht:
This PR is closed in favor of PR #13247 which should allow to integrate external ADC devices. For DAC devices another solution is required.

PR #10512, opened 2018-11-29
periph/dac: support for extension API

gschorcht:
### Contribution description

This is an implementation of the extension API as proposed in #9582 for DAC extensions. It corresponds to the implementation of the GPIO extension API in #9860 and #9958. The PR contains the following contributions:

- Low level function prototypes `dac_*_ll` in `drivers/include/periph.h`.
- Changes of `dac_*` functions to redirect a function call either to the low level functions `dac_*_ll` or to the `dac_ext_*` function provided by an DAC extension device driver.
- New function prototype `dac_channels` which returns the number of channels, in case of low level drivers simply the `DAC_NUMOF`. This new function makes it possible to get the information from a certain device how many channels it has and allows to iterate over the all channels of all devices.
- Changes of low level function `dac_*_ll` for all CPUs that support DAC channels.

### Testing procedure

A test case is provided that implements a soft-driver for the DAC extension interface. The test case uses this driver to confirm that interception and redirection of the API call are working properly. This has been tested and is working properly on esp8266-esp-12x.

### Issues/PRs references

Implements #9582 
Corresponds to #9860 and #9958 

The changes in the `makefiles/pseudomodules.mk` and  in `drivers/Makefile.include` made to get it working are the same as in #9958. These changes might be removed once #9958 is merged.

PR #10518, opened 2018-11-29
drivers: support for Microchip MCP47xx DAC devices added

gschorcht:
### Contribution description

This PR adds a driver which supports different Microchip MCP47xx DAC variants. Multiple MCP47xx DAC devices and different variants can be used at the same time.

Expander | Channels | Resolution
:--------|:--------:|:----------:
MCP4706  | 1        |  8-bit
MCP4716  | 1        | 10-bit
MCP4725  | 1        | 12-bit
MCP4726  | 1        | 12-bit
MCP4728  | 4        | 12-bit

The following features of MCP47xx DAC devices are not supported at the moment:
- configuring and reading address bits to/from EEPROM
- writing DAC channel output values to the EEPROM
- setting the UDAC bit and using the LDAC pin for MCP4728

The driver interface is kept as compatible as possible with the peripheral DAC interface. The only differences are that
- functions have the prefix `mcp47xx_` and
- functions require an additional parameter, the pointer to the MCP47xx device of type #mcp47xx_t.

### Testing procedure

Please refer the test application in `tests/driver_mcp47xx` for an example on how to use the driver
```
make -C tests/driver_mcp47xx BOARD=...
```


gschorcht:
Still waiting for another PR.

drivers/include/mcp47xx.h
```diff
@@ -0,0 +1,347 @@
...
+    mcp47xx_pd_mode_t   pd_mode;    /**< Power-down mode selection */
+
+} mcp47xx_params_t;
+
+/**
+ * @brief   MCP47xx device data structure type
+ */
+typedef struct {
+    mcp47xx_params_t params;              /**< device configuration parameters */
+    uint16_t values[MCP47XX_CHN_NUM_MAX]; /**< contains the last values set */
```
benpicco, line 230:
```suggestion
    uint16_t values[MCP47XX_CHN_NUM_MAX]; /**< contains the last values set for persistence when device is powered off  */
```

right?

drivers/include/mcp47xx.h
```diff
@@ -0,0 +1,347 @@
...
+ * The value is always given as 16-bit value and is internally scaled to the
+ * actual resolution that the DAC unit provides, e.g., 12-bit.
+ *
+ * @param[in]   dev     descriptor of the MCP47xx DAC device
+ * @param[in]   chn     channel to set
+ * @param[out]  value   value to set line to
+ *
+ * @retval  none
+ */
+void mcp47xx_dac_get(mcp47xx_t *dev, dac_t chn, uint16_t *value);
```
benpicco, line 302: I'm a bit surprised you implemented this when `periph/dac` API does not provide any `dac_get()`.
Feel free to keep it though :)

drivers/include/mcp47xx.h
```diff
@@ -0,0 +1,347 @@
...
+} mcp47xx_t;
+
+#if IS_USED(MODULE_SAUL) || DOXYGEN
+/**
+ * @brief   MCP47xx configuration structure for mapping DAC channels to SAUL
+ */
+typedef struct {
+    const char *name;   /**< name of the MCP47xx device */
+    unsigned int dev;   /**< index of the MCP47xx device */
+    dac_t channel;      /**< channel of the MCP47xx device */
```
benpicco, line 240: `dac_t` ?

benpicco (commented):
Looks good! Only a few clarifications:

drivers/include/mcp47xx.h
gschorcht: Yes, it is the same as for the `perhiph/dac`. The goal was to keep it as compatible as possible with `periph/dac` and to handle channels in same way when we intended to introduced the DAC extension API with PR #10512.

drivers/include/mcp47xx.h
gschorcht: Yes, wil change it. May I squash directly?

benpicco: Sure, squash

drivers/include/mcp47xx.h
benpicco: But `dac_t` would refer to a unique DAC device, not a channel of a DAC.
Just using `uint8_t` here would be less confusing IMHO (there also is no `saul_dac_params_t` yet)

gschorcht: No, `dac_t` refers a line, that device + channel. It is also used in `periph/dac` to [set a certain channel of a device](https://doc.riot-os.org/group__drivers__periph__dac.html#ga179ae55de04d356cd6fbb63e9e3f51f6). In this sense, it also makes sense to use it here.

gschorcht: When we started to introduce the extension API, we used exactly the same types for extenders to refer devices and channels.

gschorcht: Unfortunatly, we had not much luck with the extension interfaces as suggested in PR #10512 thil now.

gschorcht: However, we can of course also use `uint8_t`, if there is an extension API at some point in future, the type `dac_t` can still be mapped to `dev` + `chn`.

benpicco: We can still change it later if we go forward with the extension API. :wink: 
To an unsuspecting user like me the `dac_t` type here just adds confusion.

benpicco:
Please squash!

benpicco (approved):

Issue #10488, opened 2018-11-28
core/*: many important system functions illegal inside ISRs


ZetaR60:
### Description [New]
In many projects, it is desirable to be able to utilize the breadth of what is available to RIOT in code triggered by interrupts. However, many of the important system functions (such as mutex) are not available for use inside of ISRs. Consequently, many important drivers such as I2C and SPI are also not available. Currently, the netdev drivers implement a separate thread that the ISR hands events off to, but doing this for everything is potentially infeasible. See discussion below.

#### Description
On systems that use an interrupt to do a thread context switch ```mutex_lock(mutex)``` does not block as expected when inside an ISR and the mutex is already locked.

#### Steps to reproduce the issue
On a system such as ```esp8266```, issue an ```i2c_acquire(dev)``` or similar while inside of an ISR and while the same device is already acquired. I expect that this can be reproduced on other systems that use an interrupt to do a thread context switch.

#### Expected results
The system yields to another thread until the mutex is unblocked.

#### Actual results
The thread yield is put off until after the ISR completes, and the ISR's access collides with whatever is already using the bus.

#### Versions
Board was an ```esp8266-esp-12x```. Initial behavior found by @gschorcht in #10430, but I am following up on what is causing the issue.

#### References
#10430 Adds support for an I2C-based GPIO expander, and utilizing this device to generate a pin interrupt that alters the state of another expander pin causes some strange behavior. See @gschorcht  's explanation of the behavior [here](https://github.com/RIOT-OS/RIOT/pull/10430#issuecomment-441556234).

ZetaR60:
Note: I initially thought that this does not affect systems with non-interrupt context switches such as ATmega, but I see that ATmega also only schedules the switch until the end of the ISR.

gschorcht:
In short, in the case described in #10430, the problem was simply that an I2C write access to one device causes an interrupt on another device while the I2C write access is not yet completed. However, the ISR requires I2C read access to the second device. If both devices are on the same I2C bus, I2C read access will either fail o interfere with the I2C write access.

I'm not really sure whether the behavior is really wrong and that it is related to the interrupt based context switch.

As I understand what exactly happens is:

- thread x has the mutex and is interrupted
- the ISR tries to get the mutex by calling `mutex_lock`, however the mutex is locked
- since `mutex_lock` is blocking, the mutex tries to add the calling thread (still thread x) to the mutex queue
[...]

PR #10519, opened 2018-11-29
core/thread: thread block legality check

ZetaR60:
### Contribution description
This PR adds a simple sanity check to mbox, msg, mutex, etc. that triggers a kernel panic when a core function tries to block the current thread while within an ISR. The intention is to produce a clear failure compared to the current behavior which can trigger problems that are harder to diagnose. I believe that this will be helpful because there are a lot of libraries that call these core functions and it may not be immediately obvious to less experienced users which ones are ISR safe and which ones are not.

### Issues/PRs references
#10488

### DIFF
--- a/core/include/panic.h
+++ b/core/include/panic.h
@@ -36,7 +36,6 @@ typedef enum {
     PANIC_SOFT_REBOOT,
     PANIC_HARD_REBOOT,
     PANIC_ASSERT_FAIL,
-    PANIC_ILLEGAL_BLOCK,
 #ifdef MODULE_CORTEXM_COMMON
     PANIC_NMI_HANDLER,       /**< non maskable interrupt */
     PANIC_HARD_FAULT,        /**< hard fault */
--- a/core/include/panic.h
+++ b/core/include/panic.h
@@ -36,6 +36,7 @@ typedef enum {
     PANIC_SOFT_REBOOT,
     PANIC_HARD_REBOOT,
     PANIC_ASSERT_FAIL,
+    PANIC_ILLEGAL_BLOCK,
 #ifdef MODULE_CORTEXM_COMMON
     PANIC_NMI_HANDLER,       /**< non maskable interrupt */
     PANIC_HARD_FAULT,        /**< hard fault */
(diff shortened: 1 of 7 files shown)


jcarrano:
I'm not sure I understand the issue, but I see a pattern of 
    `sched_set_status(blocked or sleeping) -> irq_restore(...) -> thread_yield()`

Could this be turned into a function of its own, that also includes this check?

kaspar030:
This needs to be guarded (e.g., by DEVELHELP). I'd rather not bloat deployed code with runtime checks, in the hottest code paths.

Maybe simple asserts are sufficient?

ZetaR60:
> Maybe simple asserts are sufficient?

From what I understand, assert is usually used to verify that parameters fulfill their documented restrictions, so it might be clearer to have a separate error message. But if you think it is better to use assert, I can change it.

> I see a pattern of
> ```sched_set_status(blocked or sleeping) -> irq_restore(...) -> thread_yield()```
> Could this be turned into a function of its own, that also includes this check?

Not everything fits this pattern exactly, and you would be passing a lot of parameters to the function.

kaspar030:
> From what I understand, assert is usually used to verify that parameters fulfill their documented restrictions, so it might be clearer to have a separate error message. But if you think it is better to use assert, I can change it.

TL;DR Yes, please.

The function panics if called within an ISR, with a proper error message. It is very useful when developing, but not worth the overhead (runtime and flash) in production code. assert does exactly that.

The only difference of the added function to a plain assert is a distinct panic code and and a distinct error message. assert also points to the exact position in the binary.

I don't see any use for the panic code. 
A message can be appended like this: ```assert(!irq_is_in() && "function must not be called from within ISR!");```.

The naming of the function is misleading. "thread_block_legal_check()" panics when called from ISR. I wouldn't assume that when just looking at the function name. "panic_if_called_from_isr()" would be much more to the point. Then "assert(!irq_is_in())" is much more idiomatic and used like that in many other places in RIOT's codebase.

ZetaR60:
@kaspar030 I have changed things to follow your suggestions.

PR #10526, opened 2018-11-30
drivers/adcxx1c: improve error handling

vincent-d:
### Contribution description

ADCxx1C driver was not handling error properly. Most of the i2c error were ignored.
This adds errror checking and make use of LOG_* to print logs. It also return `-ECANCELED` in SAUL wrapper in case of error.

### Testing procedure

Normal flow (without i2c errors) still works, tested with `examples/default` and `tests/driver_adcxx1c`.

### DIFF
--- a/drivers/adcxx1c/adcxx1c.c
+++ b/drivers/adcxx1c/adcxx1c.c
@@ -37,6 +37,7 @@
 int adcxx1c_init(adcxx1c_t *dev, const adcxx1c_params_t *params)
 {
     assert(dev && params);
+    int status;
 
     dev->params = *params;
     dev->cb = NULL;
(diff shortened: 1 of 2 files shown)


drivers/adcxx1c/adcxx1c_saul.c
```diff
@@ -26,7 +26,9 @@
 
 static int read_adc(const void *dev, phydat_t *res)
 {
-    adcxx1c_read_raw((const adcxx1c_t *)dev, res->val);
+    if (adcxx1c_read_raw((const adcxx1c_t *)dev, res->val)) {
```
aabadie: Some people tend to think `&res->val[0]` is more explicit.

drivers/adcxx1c/adcxx1c.c
```diff
@@ -116,23 +130,42 @@ int adcxx1c_set_alert_parameters(const adcxx1c_t *dev, int16_t low_limit,
...
+    }
 
     hysteresis <<= (12 - dev->params.bits);
     buf[0] = hysteresis >> 8;
     buf[1] = hysteresis & 0xFF;
-    i2c_write_regs(I2C, ADDR, ADCXX1C_HYSTERESIS_ADDR, buf, 2, 0);
+    status = i2c_write_regs(I2C, ADDR, ADCXX1C_HYSTERESIS_ADDR, buf, 2, 0);
+    if (status < 0) {
+        i2c_release(I2C);
+        LOG_DEBUG("[adcxx1c] set_alert (hysteresis) - error: unable to communicate "
```
aabadie: What's the benefit of using LOG_DEBUG instead of DEBUG ? Is there a consensus or some guidelines if/when one should be used instead of the other ?

drivers/adcxx1c/adcxx1c_saul.c
vincent-d: will fix

drivers/adcxx1c/adcxx1c.c
vincent-d: It was more convenient for me because we use `LOG_*` everywhere and it's enabled for the whole app (not at a file level like `DEBUG`). I'm not aware of any consensus or guidelines for that, except for saul auto-init code which seems to use ` LOG_*`. I can revert this to `DEBUG` if you prefer.

aabadie: I would prefer using the DEBUG, just for consistency. I haven't checked in details but I think this is what is used in general in driver implementations.

vincent-d: Fixed

drivers/adcxx1c/adcxx1c_saul.c
vincent-d: Done

aabadie:
@vincent-d, I just pushed 36e6a4e with fixes for remaining nits. If you are ok you can squash :)

vincent-d:
squashed

PR #10530, opened 2018-11-30
cpu/esp32: fix of adc_res_t

gschorcht:
### Contribution description

esp32 overrides in `periph_cpu.h` the default declaration of `adc_res_t` from `periph/adc.h` with a definition with only four resolutions
- two new platform specific resolutions and
- two resolutions defined by the default declaration of `adc_res_t`.

This gives compilation errors if an application uses other resolutions. According to the documentation, `adc_sample` should return -1 if the resolution is not supported.

All other CPUs override `adc_res_t` either to add new resolutions or to mark resolutions as unsupported. But they all allow to use them at the interface. 

Therefore, esp32 overrides now the declaration of adc_res_t with
- all resolutions that are defined by the default declaration of adc_res_t and
- two new platform specific resolutions.

It returns -1 if a resolution is used in `adc_sample` that is not supported.

### Testing procedure

Use the test application `periph/adc` and terminal program with any esp32 board to observe the output of defined ADC lines.
```
make flash BOARD=esp32-wroom-32 -C tests/periph_adc/
```

### DIFF
--- a/cpu/esp32/include/periph_cpu.h
+++ b/cpu/esp32/include/periph_cpu.h
@@ -206,10 +206,14 @@ typedef enum {
  */
 #define HAVE_ADC_RES_T
 typedef enum {
-    ADC_RES_9BIT = 0,       /**< ADC resolution: 9 bit */
-    ADC_RES_10BIT,          /**< ADC resolution: 10 bit */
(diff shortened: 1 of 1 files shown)


gschorcht:
@smlng Thanks

PR #10543, opened 2018-12-04
Makefile.include: remove functionality to build with another version.

jcarrano:
### Contribution description

Makefile.include contained logic to clone a different RIOT version and build the application against it. This can be easily achieved by making a git worktree and overriding RIOTBASE, without adding any complexity to our makefiles. Therefore, the feature is being removed.

### Testing procedure

CI will test that nothing got broken.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -146,8 +146,6 @@ ifeq ($(WPEDANTIC),1)
   CFLAGS += -Wpedantic
 endif
 
-ifneq (10,$(if $(RIOT_VERSION),1,0)$(if $(__RIOTBUILD_FLAG),1,0))
-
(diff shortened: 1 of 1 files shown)


cladmi:
I also planned to remove it for some time, it introduces a specific branch in the main Makefile.include for something I would not want included in there. Removing it will also simplify cleaning other things.

Maybe we should put a deprecation error for one full release to see if somebody complains.

cladmi:
We do not maintain compatibility with previous versions, even build with docker cannot work, so I do not like having something making think it is the case.

If `RIOT_VERSION` is now unused, it could even replace the `RIOT_VERSION_OVERRIDE`. But it would be another PR/discussion.

PR #10548, opened 2018-12-05
Makefile.include: flash: do not peek into MAKECMDGOALS.

jcarrano:
### Contribution description

When flash-only was introduced (in #8373), the `flash` rule was made conditionally dependent on `all` by looking for `flash-only` in MAKECMDGOALS. This was done to avoid code duplication.

There's a cleaner way, by using [canned recipes](https://www.gnu.org/software/make/manual/html_node/Canned-Recipes.html).

### Testing procedure

Go to any example and type `make flash-only` and `make flash`. You don't even need to have the board connected.

In the first case it should not attempt to make all and in the second it should.

### Issues/PRs references

Refactor of #8373.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -511,16 +511,19 @@ distclean:
 	-@for i in $(USEPKG) ; do "$(MAKE)" -C $(RIOTPKG)/$$i distclean ; done
 	-@rm -rf $(BINDIRBASE)
 
-# if make target != 'flash-only', add target 'all' to ensure build before flash
-ifeq (,$(filter flash-only, $(MAKECMDGOALS)))
-  BUILD_BEFORE_FLASH = all
-endif
+define flash-recipe
+  $(call check_cmd,$(FLASHER),Flash program)
+  $(FLASHER) $(FFLAGS)
+endef
 
-flash: $(BUILD_BEFORE_FLASH) $(FLASHDEPS)
-	$(call check_cmd,$(FLASHER),Flash program)
-	$(FLASHER) $(FFLAGS)
+# Do not add dependencies to "flash" directly, use FLASHDEPS, as this is shared
+# with flash-only too
 
-flash-only: flash
+flash: all $(FLASHDEPS)
+	$(flash-recipe)
+
+flash-only: $(FLASHDEPS)
+	$(flash-recipe)
 
 preflash: $(BUILD_BEFORE_FLASH)
 	$(PREFLASHER) $(PREFFLAGS)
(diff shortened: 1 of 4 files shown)


cladmi:
I globally agree with the change. When introduced there were not many usages of `define` or usage of canned recipe so was not going in this direction as it would have asked to repeat the check and the `FLASHER` lines.

My only remark, is that I would prefer to use an active verb.

The `check_cmd` does check the command exists.
In the documentation they use `run-yacc` or `frobnicate`, where here `flash-commands` it is more a descriptive name or its content than of what it does.

jcarrano:
What about "do-flash"?

jcarrano:
That sucks! It means we cannot `?=` a canned recipe. OSX sucks so bad it spreads it's suckyness everywhere.

For reference,  [OSX's version of make is __12 years old__](https://savannah.gnu.org/forum/forum.php?forum_id=4380).

cladmi:
It was also in debian `wheezy`, but its LTS support has been ended since May 2018 https://www.debian.org/releases/wheezy/

cladmi:
Shown irl, I found this line by checking something else:

https://github.com/RIOT-OS/RIOT/blob/b6abbad6fbece3004e65deb092263bafefc2108d/boards/pba-d-01-kw2x/Makefile.include#L14

There might be others and they must be updated to `FLASHDEPS` then.

cladmi:
For upcoming steps, as you are also working on cleaning it.
This `flash-recipe` can also be used in `mcuboot.mk` and in `riotboot` PR also.

cladmi:
I tested that the `wdog-disable.o wdog-disable.bin` is indeed created when doing `make flash` and also when doing `make all; make flash-only` for `BOARD=mulle`.

For me you can squash. Please put the `kinetis` commit first as it is required to have a working history.

I would like a test on mac before merging this too @smlng

jcarrano:
@smlng can you take look at this? or someone else with a fruit computer?

smlng:
tested on macOS with `pba-d-01-kw2x` aka PhyNode, does work with `flash` and `flash-only` targets as expected.

jcarrano:
@smlng Thanks!

cladmi:
Just a last nitpick, can you please put in the commit description canned recipe definition that we do no use `define name =` or even `define name ?=` because of make 3.81 limitations. It would simplify understanding for future changes. You can directly amend the commit.

Note for future PRs, this can also be propagated later to `mcuboot` and `riotboot` but will need adaptation on their side so should come in followup PRs.

cladmi:
BTW github still shows the kinetis commit after the other one, but the order is correct when looked locally.

jcarrano:
I updated the commit message and added a comment too.

PR #10551, opened 2018-12-05
eclipse.inc.mk: Split eclipse support into it's own file.

jcarrano:
### Contribution description

To keep Makefile.include clean and to be consistent with other tools, the Eclipse IDE support is put in a separate file.

### Testing procedure

```sh
$ make -C examples/hello-world eclipsesym
$ file examples/hello-world/eclipsesym.xml
examples/hello-world/eclipsesym.xml: XML 1.0 document, ASCII text
```

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -565,15 +565,8 @@ objdump:
 	$(call check_cmd,$(OBJDUMP),Objdump program)
 	$(OBJDUMP) $(OBJDUMPFLAGS) $(ELFFILE) | less
 
-# Generate an XML file containing all macro definitions and include paths for
-# use in Eclipse CDT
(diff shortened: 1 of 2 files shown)


gebart:
Maybe this should go somewhere else entirely? We have seen arguments against some new proposals for similar functionality for other IDEs recently IIRC.

jcarrano:
@gebart I don't know. I don't used IDEs. Here I'm trying to clean-up Makefile.include. If people want to add support for IDEs then I think the separate file approach is the correct one. If people want to remove it, I don't care, I just want to keep this code outside our main Makefile.

Why are the proposals being rejected? I really don't want to get into that discussion, or else my tiny 11-line PR won't ever get merged.

PR #10556, opened 2018-12-06
drivers: support for NXP PCA9685 I2C 16-channel, 12-bit PWM controller

gschorcht:
### Contribution description

This PR adds a driver vor the NXP PCA9685 16-channel, 12-bit PWM controller that is connected via I2C. Although the controller is optimized for LED control, the 12-bit resolution also allows the control of servos with a resolution of 4 us at 60 Hz refresh signal.

The following features of the PCA9685 are supported by the driver:

- 16 channels with 12-bit resolution
- Refresh rates from 24 Hz to 1526 Hz with internal 25 MHz oscillator
- Totem pole outputs with 25 mA as sink and 10 mA as source at 5V
- Software programmable open-drain output selection
- Inverted outputs
- Active LOW Output Enable (OE) input pin
- External clock input with max. 50 MHz

The driver interface is kept as compatible as possible with the peripheral PWM interface. The only differences are that

- functions have the prefix `pca9685_` and
- functions require an additional parameter, the pointer to the PWM device of type #pca9685_t.

**Update:**
~The driver is also compatible with the PWM extension API in PR #10533 so that up to 64 PCA9685 PWM devices can be used as PWM extensions with the standard peripheral PWM interface.~
To get it merged soon, all PWM extension API code was removed.

### Testing procedure

Please refer the test application in `tests/driver_pca9685` for an example on how to use the driver 
```
make -C tests/driver_pca9685 BOARD=... flash term
```
Connect a logic analyzer and execute the following tests:

1. The signals should alter with frequency of 100 Hz and the duty cycles should be left aligned.
   ```
   init 0 0 100 1000
   set 0 0 100
   set 0 1 900
   ```
2. The signals should alter with frequency of 500 Hz and the duty cycles should be right aligned.
   ```
   init 0 1 500 100
   set 0 0 10
   set 0 1 90
   ```
3. The signals should alter with frequency of 1000 Hz and the duty cycles should be center aligned.
   ```
   init 0 2 1000 100
   set 0 0 10
   set 0 1 90
   ```

### Issues/PRs references

~The PR is related to #10533 but does not depend on it. As long as module `extend_pwm` is not enabled, it is working without #10533.~

### DIFF
--- a/drivers/include/pca9685.h
+++ b/drivers/include/pca9685.h
@@ -1,0 +1,322 @@
+/*
+ * Copyright (C) 2018 Gunar Schorcht
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_pca9685 PCA9685 I2C PWM controller
+ * @ingroup     drivers_actuators
+ * @ingroup     drivers_saul
+ * @brief       Device driver for the NXP PCA9685
+ *
+ * ## Overview
+ *
+ * The driver supports the NXP PCA9685 16-channel, 12-bit PWM LED controller
+ * connected to I2C. Although the controller is optimized for LED control,
+ * the 12-bit resolution also allows the control of servos with a resolution
+ * of 4 us at 60 Hz refresh signal.
+ *
+ * The following features of the PCA9685 are supported by the driver:
+ *
+ * - 16 channels with 12-bit resolution
+ * - Refresh rates from 24 Hz to 1526 Hz with internal 25 MHz oscillator
+ * - Totem pole outputs with 25 mA as sink and 10 mA as source at 5V
+ * - Software programmable open-drain output selection
+ * - Inverted outputs
+ * - Active LOW Output Enable (OE) input pin
+ * - External clock input with max. 50 MHz
+ *
+ * ## Usage
+ *
+ * The driver interface is kept as compatible as possible with the peripheral
+ * PWM interface. The only differences are that
+ *
+ * - functions have the prefix `pca9685_` and
+ * - functions require an additional parameter, the pointer to the PWM
+ *   device of type #pca9685_t.
+ *
+ * Please refer the test application in `tests/driver_pca9685` for an example
+ * on how to use the driver.
+ *
+ * ## SAUL Capabilities
+ *
+ * The driver provides SAUL capabilities that are compatible to the SAUL
+ * actuators of type #SAUL_ACT_SERVO.
+ *
+ * Each PCA9685 channel can be mapped directly to SAUL by defining an
+ * according entry in \c PCA9685_SAUL_PWM_PARAMS. Please refer file
+ * `$RIOTBASE/drivers/pca9685/include/pca9685_params.h` for an example.
+ *
+ *      pca9685_saul_pwm_params_t pca9685_saul_pwm_params[] = {
+ *          {
+ *              .name = "PCA9685-0:0",
+ *              .dev = 0,
+ *              .channel = 0,
+ *              .initial = (PCA9685_PARAM_RES >> 1),
+ *           },
+ *          {
+ *              .name = "PCA9685-0:1",
+ *              .dev = 0,
+ *              .channel = 1,
+ *              .initial = (PCA9685_PARAM_RES >> 2),
+ *           },
+ *          {
+ *              .name = "PCA9685-0:2",
+ *              .dev = 0,
+ *              .channel = 2,
+ *              .initial = (PCA9685_PARAM_RES >> 3),
+ *           },
+ *      };
+ *
+ * For each PWM channel that should be used with SAUL, an entry with a name,
+ * the device, the channel, and the initial value has to be defined as shown
+ * above.
+ *
+ * @{
+ *
+ * @author      Gunar Schorcht <gunar@schorcht.net>
+ * @file
+ */
+
+#ifndef PCA9685_H
+#define PCA9685_H
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "periph/gpio.h"
+#include "periph/i2c.h"
+#include "periph/pwm.h"
+
+/**
+ * @name    PCA9685 I2C slave addresses
+ *
+ * PCA9685 offers 64 possible hardware-programmable I2C slave addresses.
+ * Therefore the I2C slave address is defined as an offset in the range
+ * from 0 to 63 to a base address #PCA9685_I2C_BASE_ADDR. PCA9685 I2C
+ * slave addresses are then in the range from #PCA9685_I2C_BASE_ADDR + 0 to
+ * #PCA9685_I2C_BASE_ADDR + 63
+ *
+ * Four I2C slave addresses have special meaning when they are enabled, the
+ * All Call address (enabled by default) and three Sub Call addresses
+ * disabled by default). These addresses can be used to address either all or
+ * sub groups of PCF9695 controllers at the same time.
+ *
+ * @{
+ */
+#define PCA9685_I2C_BASE_ADDR   (0x40)  /**< I2C slave base address */
+#define PCA9685_I2C_ALLCALLADDR (0x70)  /**< Default All Call adress */
+#define PCA9685_I2C_SUBADR1     (0x71)  /**< Default Sub Call adress 1 */
+#define PCA9685_I2C_SUBADR2     (0x72)  /**< Default Sub Call adress 2 */
+#define PCA9685_I2C_SUBADR3     (0x73)  /**< Default Sub Call adress 3 */
+/** @} */
+
+/**
+ * @brief   Number of PWM channels provided by PCA9685
+ */
+#define PCA9685_CHANNEL_NUM     (16U)
+
+/**
+ * @brief   Internal PCA9685 channel resolution is 12-bit
+ */
+#define PCA9685_RESOLUTION      (1 << 12)
+
+/**
+ * @brief   Internal PCA9685 oscilator frequency is 25 MHz
+ */
+#define PCA9685_OSC_FREQ        (25000000)
+
+/**
+ * @brief   Maximum external clock frequency is 50 MHz
+ */
+#define PCA9685_EXT_FERQ_MAX    (50000000)
+
+/**
+ * @brief   PCA9685 driver error codes
+ */
+typedef enum {
+    PCA9685_OK              = 0,    /**< Success */
+    PCA9685_ERROR_I2C       = 1,    /**< I2C communication error */
+} pca9685_error_t;
+
+/**
+ * @brief   PCA9685 output driver mode
+ *
+ * The output driver mode defines how the outputs are configured.
+ */
+typedef enum {
+    PCA9685_OPEN_DRAIN = 0,     /**< Open-drain structure ouptut */
+    PCA9685_TOTEM_POLE = 1,     /**< Totem pole structure output */
+} pca9685_out_drv_t;
+
+/**
+ * @brief   PCA9685 output-not-enabled mode
+ *
+ * The output-not-enabled mode defines how the outputs behave when the
+ * active LOW output enable pin /OE is HIGH.
+ */
+typedef enum {
+    PCA9685_OFF = 0,      /**< If /OE pin is HIGH, outputs are LOW */
+    PCA9685_OUT_DRV = 1,  /**< Outputs depend on the output driver mode
+                               pca9685_params_t::out_drv. If /OE pin is HIGH,
+                               it is high-impedance for PCA9685_OPEN_DRAIN,
+                               and HIGH for PCA9685_TOTEM_POLE */
+    PCA9685_HIHGH_Z = 2,  /**< If /OE pin is HIGH, outputs are high-impedance */
+} pca9685_out_ne_t;
+
+/**
+ * @brief   PCA9685 device initialization parameters
+ */
+typedef struct {
+
+    i2c_t i2c_dev;     /**< I2C device, default I2C_DEV(0) */
+    uint8_t i2c_addr;  /**< I2C slave address */
+
+    pwm_mode_t mode;    /**< PWM mode for all channels: #PWM_LEFT, #PWM_CENTER,
+                             #PWM_RIGHT supported, (default PWM_CENTER) */
+    uint32_t freq;      /**< PWM frequency in Hz (default 100) */
+    uint16_t res;       /**< PWM resolution (default 4096) */
+
+    bool inv;           /**< Invert outputs, e.g., for LEDs (default yes) */
+
+    uint32_t ext_freq;  /**< If not 0, EXTCLK pin is used with this frequency */
+    gpio_t oe_pin;      /**< Active LOW output enable pin /OE. If #GPIO_UNDEF,
+                             the pin is not used. (default #GPIO_UNDEF). */
+    pca9685_out_drv_t out_drv; /**< Output driver mode */
+    pca9685_out_ne_t  out_ne;  /**< Output-not-enabled mode */
+
+} pca9685_params_t;
+
+/**
+ * @brief   PCA9685 PWM device data structure type
+ */
+typedef struct {
+
+    pca9685_params_t params;     /**< Device initialization parameters */
+    bool powered_on;             /**< Devices is powered on if true */
+
+} pca9685_t;
+
+#if MODULE_SAUL || DOXYGEN
+/**
+ * @brief   PCA9685 configuration structure for mapping PWM channels to SAUL
+ */
+typedef struct {
+    const char *name;   /**< name of the PCA9685 device */
+    unsigned int dev;   /**< index of the PCA9685 device */
+    uint8_t channel;    /**< channel of the PCA9685 device */
+    uint16_t initial;   /**< initial duty-cycle value*/
+} pca9685_saul_pwm_params_t;
+#endif
+
+/**
+ * @brief   Initialize the PCA9685 PWM device driver
+ *
+ * The function initializes the driver. After calling this function, the PWM
+ * device is in low-power sleep mode (powered off), all outputs off. Before
+ * the PWM device can be used, it has to be initialized with #pca9685_pwm_init.
+ *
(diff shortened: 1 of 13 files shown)


mrusme:
Can this be merged in a way it becomes partially functional and continue with the fancy stuff as soon as the PR this is depending on are ready?

gschorcht:
> Can this be merged in a way it becomes partially functional and continue with the fancy stuff as soon as the PR this is depending on are ready?

From  my point of view the driver is ready and could be used.

The initial plan was to merge the PWM extension API in PR #10533 first. If you take a look to the test application you can see that PWM extender channels could be used exactly like MCU pwm channels when the module `extend_pwm` is enabled what would be pretty cool. Unfortunately, we still can't find and agreement for extension APIs. But, if module `extend_pwm` is not enabled, it can be used with current interface.

That is, you could checkout the PR and you could use it. I will rebase it later today to the current master. I could also remove the extender parts for the moment, if it helps to merge it. It would be great, if you would test it.

gschorcht:
@mrusme I have rebased it to current master and it should be working. Now, I have to find someone (maybe @benpicco) who review it. Since you have probably the hardware, it would help if you could test it.

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,382 @@
...
+
+    /* prescale can only be set while in sleep mode (powered off) */
+    if (dev->powered_on) {
+        pca9685_pwm_poweroff(dev);
+    }
+
+    /* prescale = round(clk / (PCA9685_RESOLUTION * freq)) - 1; */
+    float clk = (float)(dev->params.ext_freq ? dev->params.ext_freq
+                                             : PCA9685_OSC_FREQ);
+    float prescale = clk / PCA9685_RESOLUTION / freq;
```
benpicco: Do you really need `float` here?
```C
uint32_t div = PCA9685_RESOLUTION * freq;
uint8_t byte = (clk + div/2) / div - 1;
```

Should achieve the same result (untested)

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,382 @@
...
+        off = PCA9685_LED_OFF;
+    }
+    else if (val == dev->params.res) {
+        /* full on */
+        on = PCA9685_LED_ON;
+        off = 0;
+    }
+    else {
+        /* duty = scale * val, with scale = 4096 / resolution; */
+        uint16_t duty = round((float)PCA9685_RESOLUTION / dev->params.res * val);
```
benpicco:
```suggestion
        uint32_t div = dev->params.res * val;
        uint16_t duty = (PCA9685_RESOLUTION + div/2) / div;
```

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,382 @@
...
+    /* release the I2C device */
+    i2c_release(dev->params.i2c_dev);
+
+    return PCA9685_OK;
+}
+
+inline static int _write_word(const pca9685_t *dev, uint8_t reg, uint16_t data)
+{
+    uint8_t bytes[2] = { data & 0xff, (data >> 8) & 0xff };
+    return _write (dev, reg, (uint8_t*)&bytes, 2);
```
benpicco:
```suggestion
    data = htons(data);
    return _write (dev, reg, &data, sizeof(data));
```

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,382 @@
...
+        return -PCA9685_ERROR_I2C;
+    }
+
+    /* release the I2C device */
+    i2c_release(dev->params.i2c_dev);
+
+    return PCA9685_OK;
+}
+
+static int _write(const pca9685_t *dev, uint8_t reg, const uint8_t *data, uint32_t len)
```
benpicco: With `const void *data` you don't have to cast later on.

drivers/include/pca9685.h
```diff
@@ -0,0 +1,322 @@
...
+                               and HIGH for PCA9685_TOTEM_POLE */
+    PCA9685_HIHGH_Z = 2,  /**< If /OE pin is HIGH, outputs are high-impedance */
+} pca9685_out_ne_t;
+
+/**
+ * @brief   PCA9685 device initialization parameters
+ */
+typedef struct {
+
+    unsigned i2c_dev;   /**< I2C device, default I2C_DEV(0) */
```
benpicco:
```suggestion
    i2c_t i2c_dev;   /**< I2C device, default I2C_DEV(0) */
```

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,382 @@
...
+
+    while (!((mask >> shift) & 0x01)) {
+        shift++;
+    }
+
+    /* read current register value */
+    EXEC_RET(_read(dev, reg, &byte, 1));
+
+    /* set masked bits to the given value  */
+    byte = (byte & ~mask) | ((data << shift) & mask);
```
benpicco: Looks like `data` is always either `0` or `1`/ mask has only one bit set.
Then you could simply do
```C
*byte = bit ? (*byte | mask) : (*byte & ~mask);
```

The same is almost true for `_set_reg_bit()` too - `PCA9685_MODE2_OUTNE` is the odd one out where you actually set two bits instead of one.

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,382 @@
...
+    EXEC_RET(_update(dev, PCA9685_REG_MODE1, PCA9685_MODE1_AI, 1));
+
+    /* switch off all channels */
+    EXEC_RET(_write_word(dev, PCA9685_REG_ALL_LED_OFF, PCA9685_ALL_LED_OFF));
+
+    /* set Auto-Increment flag */
+    uint8_t byte = 0;
+    _set_reg_bit(&byte, PCA9685_MODE2_OUTDRV, dev->params.inv);
+    _set_reg_bit(&byte, PCA9685_MODE2_OUTDRV, dev->params.out_drv);
+    _set_reg_bit(&byte, PCA9685_MODE2_OUTNE, dev->params.out_ne);
```
benpicco:
```suggestion
    _set_reg_bit(&byte, PCA9685_MODE2_OUTNE0, dev->params.out_ne & 0x1);
    _set_reg_bit(&byte, PCA9685_MODE2_OUTNE1, dev->params.out_ne & 0x2);
```

Maybe this is too hacky, but it world make the set function much simpler.
Another solution world be setting the already shifted bits.

(I can live with the while loop too, I'd just rather see if we can do without)

gschorcht:
@benpicco Thanks for reviewing. I'm really impressed, so fast and so deep in detail.

drivers/pca9685/pca9685.c
gschorcht: Hm, this executed only once. The while loop is not a big thing. I would prefer to leave it as it is, for the sake of clarity. We could think about not to define only a bit mask but also a bit shift value for each for register field. Another way would be to define structures for all registers.

But, since `_set_reg_bit` is only used during `_init`, there should be no performance problem.

But, I realized a real real problem.
```diff
-  _set_reg_bit(&byte, PCA9685_MODE2_OUTDRV, dev->params.inv);
+  _set_reg_bit(&byte, PCA9685_MODE2_INVERT, dev->params.inv);
```

drivers/pca9685/pca9685.c
gschorcht: Indeed, `uint32` sholud be enough for a maximum frequency of 1524 Hz and 12 bit resolution. I will change.

drivers/pca9685/pca9685.c
gschorcht: Ok.

drivers/pca9685/pca9685.c
gschorcht: Are you sure that `htons` is available on all systems`? Furthermore, I thought network byte order is big endian. Thus, on big endian machines `htons` does nothing while we could use `uint16_t` as is on litle endian machines.

drivers/pca9685/pca9685.c
gschorcht: OK

drivers/include/pca9685.h
gschorcht: Makes sense.

drivers/pca9685/pca9685.c
gschorcht: Where do I have to cast?

drivers/pca9685/pca9685.c
gschorcht: OK

gschorcht: Hm, your suggestion is not correct. `duty = round(PCA9685_RESOLUTION / dev->params.res * val)`, not `duty = round(PCA9685_RESOLUTION / (dev->params.res * val))`

gschorcht: I guess
```
uint32_t duty = PCA9685_RESOLUTION * val / dev->params.res
```
should make the trick. Due to the `uint32_t` the multiplication can't overflow.

drivers/pca9685/pca9685.c
benpicco: OK sorry, got confused about which way around the bytes should be.
You can still drop the `(uint8_t)&` from `bytes `

benpicco: OK sorry, got confused about which way around the bytes should be.
You can still drop the `(uint8_t)&` from `bytes ` I think.

drivers/pca9685/pca9685.c
benpicco: If `_write_word()` uses bytes, you don't.

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,382 @@
...
+    uint8_t bytes[2] = { data & 0xff, (data >> 8) & 0xff };
+    return _write (dev, reg, (uint8_t*)&bytes, 2);
+}
+
+static int _update(const pca9685_t *dev, uint8_t reg, uint8_t mask, uint8_t data)
+{
+    uint8_t byte;
+    uint8_t shift = 0;
+
+    while (!((mask >> shift) & 0x01)) {
```
benpicco: `shift` is unused now so you can drop that loop.
Also maybe change `data` to bool so it's clear that it will only apply the mask, not set the individual `data` bits.

drivers/pca9685/pca9685.c
gschorcht: So, it is OK leave it as `uint8_t`? It's an internal interface.

drivers/pca9685/pca9685.c
gschorcht: Yeah, indeed :sunglasses:

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,382 @@
...
+        shift++;
+    }
+
+    /* read current register value */
+    EXEC_RET(_read(dev, reg, &byte, 1));
+
+    /* set masked bits to the given value  */
+    byte = data ? (byte | mask) : (byte & ~mask);
```
gschorcht: Although it is not used til now, `_update` function should allow to set multi-bit fields of a register. Thus, I had to change it as follows:
```diff
- byte = data ? (byte | mask) : (byte & ~mask);
+ _set_reg_bit(&byte, mask, bit);
```

sys/auto_init/saul/auto_init_pca9685.c
```diff
@@ -0,0 +1,97 @@
...
+#include "saul_reg.h"
+#include "saul/periph.h"
+
+#include "pca9685.h"
+#include "pca9685_params.h"
+
+/**
+ * @brief   Number of configured PCA9685 DAC devices
+ */
+#define PCA9685_NUM    (sizeof(pca9685_params) / sizeof(pca9685_params[0]))
```
benpicco: There is an `ARRAY_SIZE()` macro now.

gschorcht: I know, at that time when I wrote this driver (last year December) there was no macro. I can replace it of course if you insist on it.

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,377 @@
...
+
+    /* write back new register value */
+    EXEC_RET(_write(dev, reg, &byte, 1));
+
+    return PCA9685_OK;
+}
+
+static void _set_reg_bit(uint8_t *byte, uint8_t mask, uint8_t bit)
+{
+    ASSERT_PARAM(byte != NULL);
```
benpicco: Seems a bit excessive, this is an internal function that gets only called on local variables.
Maybe the compiler will even optimise it away.

benpicco:
Overall I think this is good, just some minor nits. There are more efficient ways to set those config bits, but since it only happens on init I don't think it's a problem.

@mrusme can you give it a whirl and see if it works for you?

sys/auto_init/saul/auto_init_pca9685.c
benpicco: Just one small fixup ;)

You can squash already.

drivers/pca9685/pca9685.c
gschorcht: OK, I removed the asserts in internal functions.

sys/auto_init/saul/auto_init_pca9685.c
gschorcht: Replaced them by `ARRAY_SIZE` macro.

benpicco:
I'd squash `drivers: makefile support for PCA9685 PWMs` and `drivers/pca9685: PWM extension API code removed` into `drivers: support for PCA9685 PWMs`.

When the PWM extension code gets finalised you couldn't simply revert 998c862587bd59502739be7bfbef8c65fbbe77a7 but would make a new commit anyway.

gschorcht:
> I'd squash `drivers: makefile support for PCA9685 PWMs` and `drivers/pca9685: PWM extension API code removed` into `drivers: support for PCA9685 PWMs`.
> 
> When the PWM extension code gets finalised you couldn't simply revert [998c862](https://github.com/RIOT-OS/RIOT/commit/998c862587bd59502739be7bfbef8c65fbbe77a7) but would make a new commit anyway.

I will try, but I'm not sure whether it works.

gschorcht:
> > I'd squash `drivers: makefile support for PCA9685 PWMs` and `drivers/pca9685: PWM extension API code removed` into `drivers: support for PCA9685 PWMs`.
> > When the PWM extension code gets finalised you couldn't simply revert [998c862](https://github.com/RIOT-OS/RIOT/commit/998c862587bd59502739be7bfbef8c65fbbe77a7) but would make a new commit anyway.
> 
> I will try, but I'm not sure whether it works.

No, it leads to conflicts. I would like to leave it as is.

gschorcht:
I could squash `drivers: makefile support for PCA9685 PWMs` and `drivers: support for PCA9685 PWMs.` But squashing ` drivers/pca9685: PWM extension API code removed` leads to conflicts since it contains reverting changes of other commits.

gschorcht:
Ok, I solved the conflicts by splitting commit `drivers/pca9685: PWM extension API code removed` into separate fixups of the respective commits and squashed these fixups afterwards.

benpicco:
You can do a `git reset` on the last commit and distribute the hunks among the other three commits.

Not sure what's the RIOT policy on history cleanliness, maybe this is just a pet peeve of mine

mrusme:
Hey there, thanks for the patience and sorry for the delay! I've tested this and all looks good from a general perspective. However, I've experienced a slight issue here which is probably only relevant for me - it does however make this driver unusable to me. Let me explain:

For every PWM action a write is being performed. Write then acquires I2C, sets writes the data to the registers and releases the I2C again. That's basically how most drivers run and it's totally fine. In my case, I need to set multiple channels at once, with either identical or different computed values. Hence the poor-man's-driver I wrote for the PCA [contains a small hack](https://github.com/twostairs/RIOT-drivers/blob/master/pca9685/pca9685.c#L168) which allows me to specify whether I2C should be kept acquired. If I was to release the I2C channel between setting PWM for the multiple channels I could end up with a different thread/device acquiring the I2C and blocking it for a certain amount of time, leaving me with a couple of channels set to the new PWM values, while the rest of the channels are still running on the old values.

So, long story short: This driver can be merged as it looks good. However, unfortunately I won't be able to use it in this form, because of the issue that I just described.

gschorcht:
@mrusme The problem is that the built-in PWM driver API doesn't allow to synchronize the `pwm_set` calls for different channels. For compatibility reason, the extender driver realizes `pwm_set` function does it in the same way. However, beside the requirement to be compatible with the built-in PWM driver API, the extener driver could implement additional features, e.g., to set an option and to define an additional function which allows to synchronize channel settings by aquiring the I2C bus as long as it is released explicitly.

I would really appreciate, if you would provide a proposal for an extension of this PR.

BTW, I already knew the problem of channel synchronisation and was thinking about to propose an extension of the built-in PWM driver API.

gschorcht:
@mrusme

> However, beside the requirement to be compatible with the built-in PWM driver API, the extener driver could implement additional features, e.g., to set an option and to define an additional function which allows to synchronize channel settings by aquiring the I2C bus as long as it is released explicitly.

I could imagine to add a member `bool sync_setting` to struct `pca9685_params_t` which is set to `false` by default. This member can be set in parameters to `true`, if synchronized setting is required. If synchronized setting is required, the first call of `pca9685_pwm_set` aquires the I2C bus. An additional function `pca9685_set_sync` releases the I2C bus.

With this approach, the extender can be used as defined by the built-in PWM API, but the user has the option to use synchronized setting too.

benpicco:
I'd rather prefer to merge a driver that is fit for a real-world use case, but it seems there is no easy solution for that.

So I think it will be easier for everyone to propose changes to synchronized PWM output (atomic PWM setting? :wink:) on top of the mainline tree.

The driver works - changing the PWM api to allow synchronized settings of multiple outputs is for another PR.

drivers/pca9685/pca9685.c
```diff
@@ -0,0 +1,369 @@
...
+    return PCA9685_OK;
+}
+
+static void _set_reg_bit(uint8_t *byte, uint8_t mask, uint8_t bit)
+{
+    uint8_t shift = 0;
+    while (!((mask >> shift) & 0x01)) {
+        shift++;
+    }
+    *byte = ((*byte & ~mask) | ((bit << shift) & mask));
```
benpicco: Turns out the compiler will [optimize this all away](https://godbolt.org/z/t38n6f) - neat!

PR #6630, opened 2017-02-20
cortexm_common: add FPU support for cortex-m4f and cortex-m7

vincent-d:
Fixes #1366.

Not sure if every cases are covered, but it works with nucleo-f401 and nucleo-f413.
I added a test app based on `thread_msg` which uses float.

No hard fault so far and float results seem OK with different threads.

PR #10558, opened 2018-12-06
cpu/cortexm_common: Update ARM CMSIS to V5.4.0

dylad:
### Contribution description
This PR updates ARM CMSIS files from` cpu/cortexm_common` to the lastest official version from [ARM repo](https://github.com/ARM-software/CMSIS_5/tree/5.4.0).

I need this PR because I currently work on porting SAML10 and SAML11 MCUs to RIOT. These MCUs are cortex-m23 based (ARMv8-M baseline) and our current ARM CMSIS doesn't support them.

### Testing procedure
I _think_ CI tests should be enough for this PR. Not sure, if we must run much tests on hardware.

### Issues/PRs references
None.


dylad:
Looks like I must update DSP CMSIS files.
@gebart Should I send you a PR to update it on your repo or should we use sources from this repo https://github.com/ARM-software/CMSIS_5/tree/develop/CMSIS/DSP (and drop the pkg ?)

gebart:
@dylad I think the best way would be to keep cmsis-dsp as a pkg, to make it easy to update, but redirect the git repo URL to the ARM repo instead of mine.
The CMSIS core headers for the CPU support should obviously not be in a pkg, it is impossible to build any Cortex-M specific drivers (periph) and other CPU support code without them..

Background:
I created my cmsis-dsp repo way back when the only way to access the CMSIS code was through a tarball download on arm.com via some kind of registration form (even though the licence that you signed allowed redistribution of the downloaded sources).

dylad:
@gebart, I think keeping the pkg is the right way to go. Indeed ARM CMSIS are mandatory but not DSP CMSIS. I'll update the pkg accordingly in this PR (If I create a separate PR to update DSP CMSIS it'll generate build errors because ARM CMSIS isn't update...)
I'll also change the git repo URL to the ARM repo.
Thanks for your help.

dylad:
I updated pkg/cmsis-dsp too and Murdock is happy :)

aabadie:
I'm +1 for merging this PR. @gebart, do you see any problem ?

pkg/cmsis-dsp/Makefile.cmsis-dsp
```diff
@@ -0,0 +1,43 @@
...
+$(OBJ): | $(CMSIS_BINDIRS)
+
+$(CMSIS_BINDIRS):
+	@mkdir -p $@
+
+# Reset the default goal.
+.DEFAULT_GOAL :=
+
+# Include RIOT settings and recipes
+include $(RIOTBASE)/Makefile.base
```
aabadie: Another thing (before I forget): there is a missing newline here.

dylad:
@aabadie I've addressed your comments.
I'll squash once ACK'ed.

dylad:
Huh, a recent merged PR affects this PR. I'll fix asap.

dylad:
I forced-push the branch to add the new patch from #6630
I also restore the ownership of the two patches.
Should be good but you can re-run some tests if you want.

aabadie:
I re-triggered Murdock with `CI: run tests` label set. All green, let's go!

dylad:
Thanks @aabadie ! Next step is SAML10/SAML11 support.

PR #10561, opened 2018-12-06
boards/frdm-k22f: Add second UART to config

MrKevinWeiss:
### Contribution description

This PR adds a second uart to non-conflicting pins.
Since the first UART is used for shell this allows a usable UART.

### Testing procedure

Hook up to a scope/device/HiL/loopback and see if you can send or receive.

### Issues/PRs references
No issues, just think it would be nice.

### DIFF
--- a/boards/frdm-k22f/include/periph_conf.h
+++ b/boards/frdm-k22f/include/periph_conf.h
@@ -114,9 +114,23 @@ static const uart_conf_t uart_config[] = {
         .mode   = UART_MODE_8N1,
         .type   = KINETIS_UART,
     },
+    {
+        .dev    = UART0,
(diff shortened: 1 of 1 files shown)


MrKevinWeiss:
I also tested on our HiL and was able to get an echo at 115200 with PHiLIP-B.

MrKevinWeiss:
Ya at first I tried to follow the arduino pinout but the SPI was using it.  I guess the UART used for the shell -> connected to the USB, isn't the same as the standard arduino headers.
Either way, thanks for the ACK!
If nobody wants to test on a board I will confirm it works on the HiL board and will accept the responsibility if issues occur in the future because of this.

MrKevinWeiss:
@smlng, such a big shiny green button.

PR #10578, opened 2018-12-10
 drivers/i2c: Fix i2c api issues

MrKevinWeiss:
### Contribution description
This was discovered by the HiL testing.

This PR fixes some issues regarding the i2c API.  It updates documentation to reflect implementation and intent.  It also prevents improper flags from being set in the common i2c driver implementation.

### Testing procedure
Read the documentation and maybe try to i2c_read_regs or i2c_write_regs with a board that uses the common implementation (samr21) and receive ENOTSUPPORTED error message.  

### Issues/PRs references
Fixes ##10559.

### DIFF
--- a/drivers/include/periph/i2c.h
+++ b/drivers/include/periph/i2c.h
@@ -388,8 +388,8 @@ int i2c_write_bytes(i2c_t dev, uint16_t addr, const void *data,
  * @brief   Convenience function for writing one byte to a given
  *          register address
  *
- * @note    This function is using a repeated start sequence for writing to the
- *          specified register address.
+ * @note    This function is using a continuous sequence for writing to the
+ *          specified register address. It first writes the register then data.
  *
  * @pre     i2c_acquire must be called before accessing the bus
  *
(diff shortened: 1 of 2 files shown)


drivers/include/periph/i2c.h
```diff
@@ -388,8 +388,8 @@ int i2c_write_bytes(i2c_t dev, uint16_t addr, const void *data,
  * @brief   Convenience function for writing one byte to a given
  *          register address
  *
- * @note    This function is using a repeated start sequence for writing to the
- *          specified register address.
+ * @note    This function is using a continuous sequence for writing to the
```
smlng: if you change the note on the write helpers, why not on the  reads (see above), too?

MrKevinWeiss: Because almost all devices need a repeated start read (you need to write to set the pointer and need to start again to get it into a read state) where as almost all the devices use a continuous write to enter data.  This note reflects implementation now.

MrKevinWeiss: TD;DR: Because the read is correct...

MrKevinWeiss:
@smlng ping

drivers/include/periph/i2c.h
smlng: okay

PR #10586, opened 2018-12-11
core/clist: add missing function to table in doc

haukepetersen:
### Contribution description
When working with `clist.h`, I noticed that some functions were not included in the overview table in the beginning of the doxygen header. As I like that table as quick reference, I added the (some) missing functions.

### Testing procedure
doc changes only, so successful buildtest is all we need.

### Issues/PRs references
none


kaspar030:
&go.

danpetry:
> doc changes only, so successful buildtest is all we need

Does this refer to a doxygen buildtest? I.e. `make doc`? I'm curious to know how to automatically test for correct Doxygen formatting.

PR #10593, opened 2018-12-12
gcoap/example: add comment on resource order

Rotzbua:
It took me an hour to find this unexpected behavior, when I added my own resources....

### Contribution description

add comment to prevent other ppl to waste time.

### Testing procedure

no

### Issues/PRs references

according to nanocoap/example: fix resource order #9919

### DIFF
--- a/examples/gcoap/gcoap_cli.c
+++ b/examples/gcoap/gcoap_cli.c
@@ -34,7 +34,7 @@ static void _resp_handler(unsigned req_state, coap_pkt_t* pdu,
 static ssize_t _stats_handler(coap_pkt_t* pdu, uint8_t *buf, size_t len, void *ctx);
 static ssize_t _riot_board_handler(coap_pkt_t* pdu, uint8_t *buf, size_t len, void *ctx);
 
-/* CoAP resources */
+/* CoAP resources. Must be sorted by path (ASCII order). */
 static const coap_resource_t _resources[] = {
     { "/cli/stats", COAP_GET | COAP_PUT, _stats_handler, NULL },
     { "/riot/board", COAP_GET, _riot_board_handler, NULL },


examples/gcoap/gcoap_cli.c
```diff
@@ -35,6 +35,7 @@ static ssize_t _stats_handler(coap_pkt_t* pdu, uint8_t *buf, size_t len, void *c
 static ssize_t _riot_board_handler(coap_pkt_t* pdu, uint8_t *buf, size_t len, void *ctx);
 
 /* CoAP resources */
+/* must be sorted by path (ASCII order) */
```
smlng: actually its in the documentation: see http://api.riot-os.org/group__net__gcoap.html under *Server Operation* it says: 

> Create an array of resources (coap_resource_t structs) ordered by the resource path, specifically the ASCII encoding of the path characters (digit and capital precede lower case).

I understand that many start by experimenting with and extending of examples, but to me it doesn't make sense to repeat existing present documentation where ever that part is used. Also the documentation is more specific than the suggested comment.

kb2ma: @smlng, if you don't object too strongly, I'm OK with this addition. I agree that in general we don't want to repeat ourselves because that can lead to inconsistency. However, this sorting requirement has tripped up people before, which is why we refined the documentation in the header, so maybe repeating ourselves is worthwhile in this case. In addition the nanocoap_server example implementation file already includes exactly the same comment added here, and of course the change doesn't affect the executable.

Rotzbua: @kb2ma full agree
@smlng I agree that repeated documentation is not good, but it is not intentional that the array of resources must be sorted. There is no hint by naming `_resources[]`. Would you prefer a rename of the array to `_sorted_resources[]` over an extra documentation?

smlng: I don't strongly object, but then please integrate the comment with other one, i.e. it should fit in one line without getting over 80 chars.

cladmi:
One of my first PRs was to add an assert that detects this at runtime if DEVELHELP is enabled.
https://github.com/RIOT-OS/RIOT/pull/7580 I removed it as noticed this should maybe be done more on `nanocoap` that was not in RIOT at that time.

I could re-open one if there is interest for it.

kaspar030:
> I could re-open one if there is interest for it.

+1 (for a nanocoap version)!

kb2ma:
> I could re-open one if there is interest for it.

Thanks for the offer @cladmi! +1 for gcoap, too. I recall you offered in #9919, also. I had hoped that a development-time solution via comments would work, but this subject has come up too many times now.

I still think the present PR is useful.

Rotzbua:
Thanks. Done.

PR #10606, opened 2018-12-13
tests/posix_semaphore: increase timeout margin

aabadie:
### Contribution description

This PR increases the timeout margin used in the `tests/posix_semaphore` test: on nrf51 boards the timeout can be slightly above 100100us. Using 120us margin seems enough.

### Testing procedure

- Start an experiment on _the_ microbit of IoT-LAB (you can also try on a nrf51dk):
```
$ iotlab-experiment submit -n test_pr -d 60 -l saclay,microbit,1
$ iotlab-experiment wait
```
- Build/flash/test the `tests/posix_semaphore` application on it:
```
$ make BOARD=microbit IOTLAB_NODE=auto-ssh -C tests/posix_semaphore flash test
```

With this PR it should pass, on master it returns a timeout.

### Issues/PRs references

None

### DIFF
--- a/tests/posix_semaphore/main.c
+++ b/tests/posix_semaphore/main.c
@@ -239,6 +239,10 @@ void test3(void)
 #ifdef BOARD_NATIVE
 /* native can sometime take more time to respond as it is not real time */
 #define TEST4_TIMEOUT_EXCEEDED_MARGIN (300)
+#elif CPU_FAM_NRF51
+/* nrf51 based boards needs a slightly higher margin value. Using 105us makes
+ test4 result more reliable. */
+#define TEST4_TIMEOUT_EXCEEDED_MARGIN (105)
 #else
 #define TEST4_TIMEOUT_EXCEEDED_MARGIN (100)
 #endif /* BOARD_NATIVE */


cladmi:
After running 3-4 times in master, I reliably got `102` and `103` extra time over 1 second.
Maybe put it in the commit message/comment.

Would it make sense to use a smaller margin for the moment ?
Or did you have any bigger time than 103 ?

aabadie:
> did you have any bigger time than 103 ?

No this is what I had as well. But I wanted a margin on the margin to avoid false positive. Maybe just using 105 would be ok?

tests/posix_semaphore/main.c
```diff
@@ -239,6 +239,10 @@ void test3(void)
 #ifdef BOARD_NATIVE
 /* native can sometime take more time to respond as it is not real time */
 #define TEST4_TIMEOUT_EXCEEDED_MARGIN (300)
+#elif CPU_FAM_NRF51
+/* nrf51 based boards needs a slightly higher margin value. Using 105us makes
+ test4 result more reliable. */
+#define TEST4_TIMEOUT_EXCEEDED_MARGIN (105)
 #else
 #define TEST4_TIMEOUT_EXCEEDED_MARGIN (100)
```
smlng: can't we just increase this default to say `110` instead of having (another) board specific one? I mean `105` is not that much more than the default of `100`.

aabadie: Makes sense. Using 105 should be enough though. @cladmi reported that the microbit is always between 102 and 103 and I also have these results.

cladmi: Having currently a cpu specific variable shows that this value should come from somewhere else.

It is like the first part of TDD where you only define special cases handling until you reach a point where it can/should be put in common and you refactor to using the value from somewhere else.

Here I would assume it is a value that could even be interesting to have in an application and may be defined in timer area. Having these different cases shows that this value must be "board/cpu" specific and would need special handling.

So I would personally keep the special case.

aabadie: Thanks @cladmi. @smlng does it make sense to you ? if not, how do we proceed ?

PR #10608, opened 2018-12-13
cpu/stm_common: Refactor and fix implementation for i2c_2

MrKevinWeiss:
### Contribution description

* Refactors i2c_2 to match the structure of i2c_1 better.
* Corrects functionality issues (such as nack recovery).
* Allows the common implementation of read_regs and write_regs.
* Documents constraints of hardware.
* Reduces code size (about 180 bytes)
* Matches error messages with API.

### Testing procedure
Test on stm32f1 f2 f4 or l1 board
 *If nucleo-f103rb is used it needs external pullups.
4. Test split frame writes 

### Issues/PRs references
helps with some stm32 based boards for #3366
checks some boxes in #9518

### DIFF
--- a/cpu/stm32_common/periph/i2c_2.c
+++ b/cpu/stm32_common/periph/i2c_2.c
@@ -2,6 +2,7 @@
  * Copyright (C) 2017 Kaspar Schleiser <kaspar@schleiser.de>
  *               2014 FU Berlin
  *               2018 Inria
+ *               2018 HAW Hamburg
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -16,9 +17,9 @@
  * @file
  * @brief       Low-level I2C driver implementation
  *
- * This driver supports the STM32 F4 families.
+ * This driver supports the STM32 F1, F2, L1, and F4 families.
  *
- * @note This implementation only implements the 7-bit addressing mode.
+ * @note This implementation only implements the 7-bit addressing polling mode.
  *
  * @author      Peter Kietzmann <peter.kietzmann@haw-hamburg.de>
  * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
@@ -28,6 +29,7 @@
  * @author      Vincent Dupont <vincent@otakeys.com>
  * @author      Víctor Ariño <victor.arino@triagnosys.com>
  * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
+ * @author      Kevin Weiss <kevin.weiss@haw-hamburg.de>
  *
  * @}
  */
@@ -40,28 +42,29 @@
 #include "mutex.h"
 #include "pm_layered.h"
 
-#include "periph_conf.h"
-#include "periph/gpio.h"
 #include "periph/i2c.h"
+#include "periph/gpio.h"
+#include "periph_conf.h"
 
-#define ENABLE_DEBUG    (0)
+/* Some DEBUG statements may cause delays that alter i2c functionality */
+#define ENABLE_DEBUG        (0)
 #include "debug.h"
 
-#define I2C_IRQ_PRIO            (1)
-#define I2C_FLAG_READ           (I2C_READ)
-#define I2C_FLAG_WRITE          (0)
+#define TICK_TIMEOUT        (0xFFFF)
 
-#define TICK_TIMEOUT            (0xFFFF)
+#define I2C_IRQ_PRIO        (1)
+#define I2C_FLAG_READ       (I2C_READ)
+#define I2C_FLAG_WRITE      (0)
 
-#define ERROR_FLAG              (I2C_SR1_AF | I2C_SR1_ARLO | I2C_SR1_BERR)
+#define ERROR_FLAG          (I2C_SR1_AF | I2C_SR1_ARLO | I2C_SR1_BERR)
 
 /* static function definitions */
 static void _i2c_init(I2C_TypeDef *i2c, uint32_t clk, uint32_t ccr);
-static inline int _start(I2C_TypeDef *dev, uint8_t address, uint8_t rw_flag, uint8_t flags);
-static inline void _clear_addr(I2C_TypeDef *dev);
-static inline int _write(I2C_TypeDef *dev, const uint8_t *data, int length);
-static inline int _stop(I2C_TypeDef *dev);
-static inline int _wait_ready(I2C_TypeDef *dev);
+static int _start(I2C_TypeDef *dev, uint8_t address_byte, uint8_t flags,
+                  size_t length);
+static int _stop(I2C_TypeDef *dev);
+static int _is_sr1_mask_set(I2C_TypeDef *i2c, uint32_t mask, uint8_t flags);
+static inline int _wait_for_bus(I2C_TypeDef *i2c);
 
 /**
  * @brief Array holding one pre-initialized mutex for each I2C device
@@ -72,7 +75,6 @@ void i2c_init(i2c_t dev)
 {
     assert(dev < I2C_NUMOF);
 
-    DEBUG("[i2c] init: initializing device\n");
     mutex_init(&locks[dev]);
 
     I2C_TypeDef *i2c = i2c_config[dev].dev;
@@ -105,7 +107,6 @@ void i2c_init(i2c_t dev)
     NVIC_EnableIRQ(i2c_config[dev].irqn);
 
     /* configure pins */
-    DEBUG("[i2c] init: configuring pins\n");
     gpio_init(i2c_config[dev].scl_pin, GPIO_OD_PU);
     gpio_init(i2c_config[dev].sda_pin, GPIO_OD_PU);
 #ifdef CPU_FAM_STM32F1
@@ -125,13 +126,11 @@ void i2c_init(i2c_t dev)
 #endif
 
     /* configure device */
-    DEBUG("[i2c] init: configuring device\n");
     _i2c_init(i2c, i2c_config[dev].clk, ccr);
 
 #if defined(CPU_FAM_STM32F4)
     /* make sure the analog filters don't hang -> see errata sheet 2.14.7 */
     if (i2c->SR2 & I2C_SR2_BUSY) {
-        DEBUG("[i2c] init: line busy after reset, toggle pins now\n");
         /* disable peripheral */
         i2c->CR1 &= ~I2C_CR1_PE;
         /* toggle both pins to reset analog filter */
@@ -195,9 +194,6 @@ int i2c_release(i2c_t dev)
 {
     assert(dev < I2C_NUMOF);
 
-    uint16_t tick = TICK_TIMEOUT;
-    while ((i2c_config[dev].dev->SR2 & I2C_SR2_BUSY) && tick--) {}
-
     periph_clk_dis(i2c_config[dev].bus, i2c_config[dev].rcc_mask);
 
 #ifdef STM32_PM_STOP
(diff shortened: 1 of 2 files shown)


cpu/stm32_common/periph/i2c_2.c
```diff
@@ -117,13 +119,11 @@ void i2c_init(i2c_t dev)
 #endif
 
     /* configure device */
-    DEBUG("[i2c] init: configuring device\n");
     _i2c_init(i2c, i2c_config[dev].clk, ccr);
 
 #if defined(CPU_FAM_STM32F4)
     /* make sure the analog filters don't hang -> see errata sheet 2.14.7 */
     if (i2c->SR2 & I2C_SR2_BUSY) {
-        DEBUG("[i2c] init: line busy after reset, toggle pins now\n");
```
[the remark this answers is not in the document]
MrKevinWeiss: This is because if you do enable debug it doesn't actually work and seems to screw up the shell, I think it has something to do with the uart being initialized after the i2c but I didn't really look into it.   I wrote something above the enable debug to indicate that.  It may only have been for the nucleo-f103rb though.

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -187,9 +187,6 @@ int i2c_release(i2c_t dev)
 {
     assert(dev < I2C_NUMOF);
 
-    uint16_t tick = TICK_TIMEOUT;
-    while ((i2c_config[dev].dev->SR2 & I2C_SR2_BUSY) && tick--) {}
```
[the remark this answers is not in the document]
MrKevinWeiss: If the bus was still transferring the delay should actually allow it to finish (depending on clock speeds).  However, this implementation ensures that, when an i2c stop is issued, the function will only end if the bus is free.  This removes the requirement to check if the line is busy before doing anything (which is needed for repeated start calls).

So yes, it is essentially not doing anything valuable (assuming the api is being followed correctly) and thus should be removed.

cpu/stm32_common/periph/i2c_2.c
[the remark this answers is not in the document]
aabadie: I think the problem is also there with F4 and the workaround could be, as you noticed, to initialize uart before i2c. That's what I did when debugging this driver some months ago.

MrKevinWeiss:
@maribu Thanks for the review, I did want some feedback on the removing of the busy timeout in the release function.  It is a good clarification.

cpu/stm32_common/periph/i2c_2.c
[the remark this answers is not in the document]
maribu: It would be nice for debugging to have stdio working as soon as possible during boot. But that is unrelated to this PR. Keeping those DEBUG statements will bite the next person trying to debug I2C unless someone would change the boot up process to initialize stdio sooner. So I agree that removing them is the better alternative.

maribu:
I can confirm that the SHT3X driver (using I2C) on the bluepill (STM32F103) just works fine with your I2C implementation :-) The driver does not use every function the I2C does define (e.g. the convenience functions for register access are not used) - so this test does not cover all your code.

cpu/stm32_common/periph/i2c_2.c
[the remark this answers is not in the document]
MrKevinWeiss: Noted, I opened an issue #10614 to document this concern.  I also did write that comment explaining it.

MrKevinWeiss:
Thank you very much for testing, I can also volunteer @leandrolanzieri for more test too :laughing:

MrKevinWeiss:
if the stm32f1 fix gets merged I will have to rebase again on top of that.  Just a note to self.

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -40,28 +42,30 @@
...
 #include "pm_layered.h"
 
-#include "periph_conf.h"
-#include "periph/gpio.h"
 #include "periph/i2c.h"
+#include "periph/gpio.h"
+#include "periph_conf.h"
 
-#define ENABLE_DEBUG    (0)
+/* DEBUG statements in the init causes the code to lock up */
```
leandrolanzieri: This could be removed if #10615 is merged.

MrKevinWeiss: Yup, will do if it is merged before this one, otherwise I will remove it with the other PR.

leandrolanzieri:
Tested on nucleo-f410rb and works as intended:
- [x] Basic read regs with 1,2,3 bytes
- [x] Basic write regs with 1,2,3 bytes
- [x] Test split frame (NOSTOP, NOSTOP|NOSTART, NOSTART) reads
- [x] Test split frame writes
- [x] Repeated start write (write_bytes+NOSTOP, write_bytes)
- [x] Getting an address nack then read a proper register.

MrKevinWeiss:
@aabadie or @maribu, well now that all the testing has been done...

aabadie:
Can you provide the corresponding shell commands of `tests/periph_i2c` for testing:
```
4. Test split frame writes 
```

I'd like to give a try to this one on nucleo-l152re

MrKevinWeiss:
@aabadie The thing is you need a sensor or something to interface to.  If you have a nucleo-103rb you can flash the [PHiLIP firmware](https://github.com/riot-appstore/PHiLIP/blob/master/QUALIFICATION/FW/PHiLIP-NUCLEO-F103RB-REV10009.bin), drag and drop the firmware and hook up the scl and sda pins.

If you want to use a sensor then let me know what it is and I can tell you the commands.

MrKevinWeiss:
We can generalize certain things for PHiLIP <addr> = 85 and  <reg> = 152, for a sensor <addr> is the address of the sensor and <reg> would be a register in the sensor (preferable a few) that is readable and writable.  Then we could go:

```
i2c_acquire 0
i2c_read_regs 0 <addr> <reg> 1 0
i2c_read_regs 0 <addr> <reg> 2 0
i2c_read_regs 0 <addr> <reg> 3 0
i2c_write_regs 0 <addr> <reg> 0 42
i2c_write_regs 0 <addr> <reg>+1 0 41 40
i2c_write_regs 0 <addr> <reg>+3 0 39 38 37
/* To confirm read that the register have changed
i2c_read_regs 0 <addr> <reg> 6 0

i2c_read_bytes 0 <addr> 1 4
i2c_read_bytes 0 <addr> 1 12
i2c_read_bytes 0 <addr> 1 8

i2c_read_bytes 0 <addr> 1 4
i2c_read_bytes 0 <addr> 1 12
i2c_read_bytes 0 <addr> 1 8

i2c_write_bytes 0 <addr> 4 <reg>
i2c_write_bytes 0 <addr> 12 35
i2c_write_bytes 0 <addr>  8 34
/* To confirm read that the register have changed
i2c_read_regs 0 <addr> <reg> 2 0

/* Repeated start write */
i2c_write_bytes 0 <addr> 4 <reg>
i2c_write_bytes 0 <addr> 0 33
/* This shouldn't fail but you may not get the correct information unless the sensor supports it */

i2c_read_regs 0 <addr>+1 <reg> 1 0
/* expect nack */
i2c_read_regs 0 <addr> <reg> 1 0 
/* expect success */
```

That is the basics of it.  There is also a test.py in the folder but to use it you should `pip3 install riot_pal` and there may be some other issues like setting the python path correctly.  Or you can get robot framework pushed though and just call

`BOARD=nucleo-l152re PHILIP_PORT=<port of PHiLIP>  make robot-test -C tests/periph_i2c`

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -40,28 +42,30 @@
...
 
 /* static function definitions */
 static void _i2c_init(I2C_TypeDef *i2c, uint32_t clk, uint32_t ccr);
-static inline int _start(I2C_TypeDef *dev, uint8_t address, uint8_t rw_flag, uint8_t flags);
-static inline void _clear_addr(I2C_TypeDef *dev);
-static inline int _write(I2C_TypeDef *dev, const uint8_t *data, int length);
-static inline int _stop(I2C_TypeDef *dev);
-static inline int _wait_ready(I2C_TypeDef *dev);
+static int _start(I2C_TypeDef *dev, uint8_t address_byte, uint8_t flags,
+    size_t length);
```
aabadie: Code style: this should be aligned with `I2C_TypeDef`.

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -371,147 +292,115 @@ int i2c_write_bytes(i2c_t dev, uint16_t address, const void *data,
...
     }
 
-    return ret;
+    return _wait_for_bus(i2c);
 }
 
-int i2c_write_regs(i2c_t dev, uint16_t address, uint16_t reg, const void *data,
-                   size_t length, uint8_t flags)
+static int _start(I2C_TypeDef *i2c, uint8_t address_byte, uint8_t flags,
+    size_t length)
```
aabadie: Code style: this should be aligned with `I2C_TypeDef`.

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -371,147 +292,115 @@ int i2c_write_bytes(i2c_t dev, uint16_t address, const void *data,
...
 {
-    assert(dev < I2C_NUMOF);
-
-    I2C_TypeDef *i2c = i2c_config[dev].dev;
     assert(i2c != NULL);
+    assert((i2c->SR2 & I2C_SR2_BUSY) || !(flags & I2C_NOSTART));
```
aabadie: I got a failed assertion on this line when brainlessly testing using `tests/periph_i2c` on `nucleo-l152re` with an hts221 sensor plugged. With the following commands list:
```
> i2c_acquire 0
> i2c_read_bytes 0 0x5f 1 12
```
as suggested in https://github.com/RIOT-OS/RIOT/pull/10608#issuecomment-448181648

Would it make more sense to have a return with a dedicated error value instead of crashing the application in this case ?

MrKevinWeiss: Yup I don't mind that, I was trying to keep the binary size down but I can see how that would be annoying when using it (it happened to me a few times)

cpu/stm32_common/periph/i2c_2.c
MrKevinWeiss: Thanks will do

cpu/stm32_common/periph/i2c_2.c
MrKevinWeiss: yup yup.

MrKevinWeiss:
@aabadie All your comments have been addressed!

cpu/stm32_common/periph/i2c_2.c
```diff
@@ -371,147 +292,116 @@ int i2c_write_bytes(i2c_t dev, uint16_t address, const void *data,
...
 
-    if ((flags & I2C_REG16) || (flags & I2C_ADDR10)) {
+    if ((flags & I2C_ADDR10) ||
+        (!(i2c->SR2 & I2C_SR2_BUSY) && (flags & I2C_NOSTART)) {
```
toonst: missing `)` here, gives compilation error

MrKevinWeiss: ...shhhhhh, nothing happened here:sweat_smile:

cpu/stm32_common/periph/i2c_2.c
aabadie: @MrKevinWeiss any chance to remove this today? And then we can merge.

MrKevinWeiss: Thought I forgot something, done!

PR #10609, opened 2018-12-13
core/kernel_defines.h: add ARG_UNUSED macro

jia200x:
### Contribution description
This PR adds the ARG_UNUSED macro, available in several C projects (Zephyr, etc).
This:
```
void function(void *arg)
{
    /* arg variable is not used */
    (void) arg;
    ...
}
```
Becomes this:
```
void function(void *arg)
{
    /* arg variable is not used */
    ARG_UNUSED(arg);
    ...
}
```

This makes the code look more clear IMO.

### Testing procedure
Apply this patch with `git apply file.patch`

file.patch:
```
diff --git a/examples/ipc_pingpong/main.c b/examples/ipc_pingpong/main.c
index 655ff39a7..874bcb0af 100644
--- a/examples/ipc_pingpong/main.c
+++ b/examples/ipc_pingpong/main.c
@@ -26,7 +26,7 @@
 
 void *second_thread(void *arg)
 {
-    (void) arg;
+    ARG_UNUSED(arg);
 
     printf("2nd thread started, pid: %" PRIkernel_pid "\n", thread_getpid());
     msg_t m;
```

Compile the `ipc_pingpong` example with `ipc_pingpong`. Check there are no `unused variable` errors.

### Issues/PRs references
None so far

### DIFF
--- a/core/include/kernel_defines.h
+++ b/core/include/kernel_defines.h
@@ -71,6 +71,13 @@
 #endif
 
 /**
+ * @def   ARG_UNUSED
+ * @brief Tell the compiler that the given function argument is not used
+ *        in the scope of the function.
+ */
+#define ARG_UNUSED(x) (void)(x)
+
+/**
  * @def CONST
  * @brief A function declared as *CONST* is #PURE and also not allowed to
  *        examine global memory. I.e. a *CONST* function cannot even


kaspar030:
> This makes the code look more clear IMO.

I don't agree. ```(void) arg;``` is idiomatic C, it's meaning is clear and in the open.

```ARG_UNUSED()``` a custom macro. It *might* just be only defined to ```#define ARG_UNUSED(x) (void)x```, but I'd have to check when encountering it for the first time.

Using this macro also requires to include "kernel_macros.h" in possibly completely kernel-unrelated code, unnecessarily creating a dependency to a mess of macro code to possibly otherwise sane C.

Macros are bad. Avoid them at all costs.

kaspar030:
Next thing we'll see is someone proposing ```FOREACH(entry, list)```. :wink:

jia200x:
> From my side a clear NACK.

Ok

> Macros are bad. Avoid them at all costs.

Sometimes this feels like a vim vs emacs discussion IMO :). It's not like we don't use macros in RIOT... (LL utils, container_of, etc)

> Next thing we'll see is someone proposing FOREACH(entry, list). wink

I think there were already a couple of attempts

jia200x:
I will just close it. Please re-open if needed

gebart:
@kaspar030 
> Next thing we'll see is someone proposing `FOREACH(entry, list)`. wink

You mean like:
https://github.com/RIOT-OS/RIOT/blob/68dc5b0d6e0422df9c3653c2cb8021fc35974aae/sys/include/utlist.h#L468-L474

I agree though, `(void) x;` is such a common phrase that it should not need to be hidden behind a macro, it should be common knowledge among C developers that this is a way to mark an argument as unused.

PR #10610, opened 2018-12-13
cpu/stm_common: Refactor and cleanup i2c_1

MrKevinWeiss:
### Contribution description

* Refactor to use common read_regs.
* Add error reporting and handling for unsupported low level commands.
* Document hardware constraints.
* Reduces code size around 20 to 50 bytes

### Testing procedure
Test on stm32f0, f3, f7, l0 or l4
4. Test split frame writes

### Issues/PRs references
checks some boxes in #9518
helps with some stm32 based boards for #3366
depends on #10608

### DIFF
--- a/cpu/stm32_common/include/periph_cpu_common.h
+++ b/cpu/stm32_common/include/periph_cpu_common.h
@@ -159,9 +159,9 @@ typedef uint32_t gpio_t;
 #define PERIPH_I2C_NEED_READ_REG
 /** Use write reg function from periph common */
 #define PERIPH_I2C_NEED_WRITE_REG
+#define PERIPH_I2C_NEED_READ_REGS
 #if defined(CPU_FAM_STM32F1) || defined(CPU_FAM_STM32F2) || \
     defined(CPU_FAM_STM32L1) || defined(CPU_FAM_STM32F4)
-#define PERIPH_I2C_NEED_READ_REGS
 #define PERIPH_I2C_NEED_WRITE_REGS
 #endif
 /** @} */
--- a/cpu/stm32_common/periph/i2c_1.c
+++ b/cpu/stm32_common/periph/i2c_1.c
@@ -2,6 +2,7 @@
  * Copyright (C) 2015 Jan Pohlmann <jan-pohlmann@gmx.de>
  *               2017 we-sens.com
  *               2018 Inria
+ *               2018 HAW Hamburg
  *
  * This file is subject to the terms and conditions of the GNU Lesser
  * General Public License v2.1. See the file LICENSE in the top level
@@ -26,6 +27,7 @@
  * @author      Jan Pohlmann <jan-pohlmann@gmx.de>
  * @author      Aurélien Fillau <aurelien.fillau@we-sens.com>
  * @author      Alexandre Abadie <alexandre.abadie@inria.fr>
+ * @author      Kevin Weiss <kevin.weiss@haw-hamburg.de>
  *
  * @}
  */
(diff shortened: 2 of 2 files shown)


aabadie:
This needs a rebase @MrKevinWeiss

MrKevinWeiss:
On it!

cpu/stm32_common/periph/i2c_1.c
```diff
@@ -146,365 +147,260 @@ int i2c_release(i2c_t dev)
...
+    DEBUG("[i2c] write_bytes: Starting\n");
+    return _write(i2c, address, data, length, flags, 0);
+}
 
-    int ret = 0;
+static int _write(I2C_TypeDef *i2c, uint16_t address, const void *data,
+                    size_t length, uint8_t flags, uint32_t cr2_flags)
+{
+    assert(i2c != NULL);
+    assert(length < 256);
```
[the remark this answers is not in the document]
MrKevinWeiss: Why the length limitation or why not combine both in a macro?
If length then it is because I would have to do a check to see if it is greater, if it is set the reload bit and keep track of everything sent.  It adds extra steps and I don't know how valuable it would be (even for something like EEPROM, when you are talking about 256 bytes adding a start and stop frame isn't that much overhead).

cpu/stm32_common/periph/i2c_1.c
```diff
@@ -146,365 +147,260 @@ int i2c_release(i2c_t dev)
...
+        return -EOPNOTSUPP;
     }
-
-    DEBUG("[i2c] write_bytes: write the data (%d bytes)\n", length);
-    /* send out data bytes */
-    ret = _write(i2c, data, length);
+    int ret = _start(i2c,
+        (address << 1) |
+        (length << I2C_CR2_NBYTES_Pos) |
+        /* I2C_FLAG_WRITE | */
```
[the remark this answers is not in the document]
MrKevinWeiss: OK will do

MrKevinWeiss:
I think I did what you want?

cpu/stm32_common/periph/i2c_1.c
```diff
@@ -146,365 +147,248 @@ int i2c_release(i2c_t dev)
...
+    return _write(i2c, addr, data, len, I2C_NOSTART, 0);
 }
 
-int i2c_read_regs(i2c_t dev, uint16_t address, uint16_t reg, void *data,
-                  size_t length, uint8_t flags)
+int i2c_read_bytes(i2c_t dev, uint16_t address, void *data,
+                   size_t length, uint8_t flags)
 {
-    assert(dev < I2C_NUMOF);
+    assert(dev < I2C_NUMOF && length < 256);
```
aabadie: I was thinking of having a define for 256. To avoid hard coded values in the code and to have it defined once.

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -159,11 +159,12 @@ typedef uint32_t gpio_t;
 #define PERIPH_I2C_NEED_READ_REG
 /** Use write reg function from periph common */
 #define PERIPH_I2C_NEED_WRITE_REG
+#define PERIPH_I2C_NEED_READ_REGS
 #if defined(CPU_FAM_STM32F1) || defined(CPU_FAM_STM32F2) || \
     defined(CPU_FAM_STM32L1) || defined(CPU_FAM_STM32F4)
-#define PERIPH_I2C_NEED_READ_REGS
 #define PERIPH_I2C_NEED_WRITE_REGS
 #endif
+
```
aabadie: This blankline could be removed

cpu/stm32_common/periph/i2c_1.c
MrKevinWeiss: Yup, makes sense, on it.

MrKevinWeiss:
@aabadie All done!

MrKevinWeiss:
It is possible but I am not sure it is worth the extra code size at the moment.  If anyone requests it in the future I can add it.

PR #10612, opened 2018-12-13
 tools/mcuboot/imgtool: Fix broken ECDSA signing.

jcarrano:
### Contribution description

A couple of typos in the ECDSA implementation meant that it was not possible to use `-t ecdsa-p256` in the `imgtool`.

### Testing procedure

I included a couple of test commits that should be removed before merging. I did it in an unusual way by putting the test commit at the end (with a revert) because I have another branch based on this one and I'd like to avoid rebasing that one if I can: 

- The first test commit enables ECDSA in mcuboot and show it is fixed.
- The second test commit reverts the fix, while keeping ECDSA enabled to show it was broken in the first place.

Checkout each of the test commits  and run:

```
make -C tests/mcuboot clean
make -C tests/mcuboot
```

It is easy to use `git diff` to show that the third commit effectively disables the first one.

```
git diff HEAD^^^
```

### DIFF
--- a/makefiles/mcuboot.mk
+++ b/makefiles/mcuboot.mk
@@ -19,7 +19,7 @@ mcuboot-create-key: $(MCUBOOT_KEYFILE)
 ifeq ($(BINDIR)/key.pem,$(MCUBOOT_KEYFILE))
 $(MCUBOOT_KEYFILE):
 	$(Q)mkdir -p $(BINDIR)
-	$(Q)$(IMGTOOL) keygen -k $@ -t rsa-2048
+	$(Q)$(IMGTOOL) keygen -k $@ -t ecdsa-p256
 endif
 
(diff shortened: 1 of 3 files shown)


makefiles/mcuboot.mk
```diff
@@ -19,7 +19,7 @@ mcuboot-create-key: $(MCUBOOT_KEYFILE)
 ifeq ($(BINDIR)/key.pem,$(MCUBOOT_KEYFILE))
 $(MCUBOOT_KEYFILE):
 	$(Q)mkdir -p $(BINDIR)
-	$(Q)$(IMGTOOL) keygen -k $@ -t rsa-2048
```
jcarrano: __NOTE!__ This is NOT the PR. _See the first commit!_

PR #10615, opened 2018-12-14
cpu/stm32_common/uart: Prevent uart from sending if not initialized

### LINKED ISSUE #10614: DEBUG stops shell if used in i2c init
#### Description
For stm32f4 and stm32f1 based cpu, having debug enabled and writing something in the init seems to prevent from interacting.

Enable debug and print in the i2c init of a stm32f4 or f1 board.

#### Expected results
See the output of the debug print
#### Actual results
Nothing gets printed and trying to write to the shell yields no result.

Mentioned in #10608

MrKevinWeiss:
### Contribution description
Due to the stdio getting called after periph_init the uart may send before initialized.
This adds a simple check so the uart does not get into a locked-up state.

### Testing procedure

Use an stm32F1, F2, F4, or L0
Enable debug in tests/periph_i2c/main.c
Write a DEBUG message in the init.
`BOARD=<selected board from above> make flash term -C tests/periph_i2c/`
`help` 
you should see a result, on master you won't and it will get locked up.

### Issues/PRs references
fixes #10614
discussed in #10608

### DIFF
--- a/cpu/stm32_common/periph/uart.c
+++ b/cpu/stm32_common/periph/uart.c
@@ -145,7 +145,12 @@ static inline void wait_for_tx_complete(uart_t uart)
 void uart_write(uart_t uart, const uint8_t *data, size_t len)
 {
     assert(uart < UART_NUMOF);
-
+#if DEVELHELP
+    /* If tx is not enabled don't try to send */
+    if (!(dev(uart)->CR1 & USART_CR1_TE)) {
+        return;
+    }
+#endif
 #ifdef MODULE_PERIPH_DMA
     if (!len) {
         return;


MrKevinWeiss:
Low impact fix that seems like it could help with other issues.  It adds 12 bytes or so...

cpu/stm32_common/periph/uart.c
```diff
@@ -145,7 +145,10 @@ static inline void wait_for_tx_complete(uart_t uart)
 void uart_write(uart_t uart, const uint8_t *data, size_t len)
 {
     assert(uart < UART_NUMOF);
-
+    if (!(dev(uart)->CR1 & USART_CR1_TE)) {
+        /* If tx is not enabled don't try to send */
```
smlng: nit-picky: I would put the comment *above* the `if` not within. Otherwise looks good to me, how do other platforms handle this, or do they need a similar update?

MrKevinWeiss: Done

maribu:
Tested on a blue pill. Works as expected :-) Please wait 10 more minutes to let me also check on an F4, just to be sure.

MrKevinWeiss:
I will also note that if I do some sort of refactor in the future I will attempt to reduce the byte size for the uart.

maribu:
I can confirm that it also solves the issue on the MSB-IoT (stm32f415rg) :-)

cpu/stm32_common/periph/uart.c
```diff
@@ -145,7 +145,10 @@ static inline void wait_for_tx_complete(uart_t uart)
 void uart_write(uart_t uart, const uint8_t *data, size_t len)
 {
     assert(uart < UART_NUMOF);
-
+    /* If tx is not enabled don't try to send */
+    if (!(dev(uart)->CR1 & USART_CR1_TE)) {
```
jcarrano: I don't think this should be enabled always, but be an assertion.

maribu: Well, `assert()` uses `stdio` as well. I'm not sure what would happen than. In any case, the user will not be able to "see" that the assertion failed, as no output over UART is available.

jcarrano: mmm, I get the point. The problem is that having the UART fail silently will be very surprising for the user.

maribu: I think the best user experience would be to make sure `stdio` becomes available as soon as possible. This would allow us using `DEBUG()` even in the initialization code of other stuff.

Maybe a good compromise would be to perform this check only when `DEVELHELP` is enabled. And also set some global flag (only present with `DEVELHELP`) when this occurs. At the very end of the initialization of `stdio` a check for that flag could be added (again, only with `DEVELHELP`) that would print a warning about lost `stdio` output.

MrKevinWeiss: Then the current implementation would crash every-time you try to debug a periph.

MrKevinWeiss: It currently calls uart_write before init_uart as pointed out by the issue.

MrKevinWeiss: That's why I like error codes everywhere.  As it stands the periphs get initialized before the uart does and attempting to write on certain boards locks it up in an infinite loop waiting for a flag.  An assert would fail if added.  Maybe that is the behavior we desire but then all periphs that want to be debugged and have something in the init would crash.

jcarrano: > I think the best user experience would be to make sure stdio becomes available as soon as possible. 

100%. This PR is kind of a work-around for me. The fix would be to either to split board initialization:
https://github.com/RIOT-OS/RIOT/blob/b709e63581a85a504b1020a025ef6f0e73f80b1d/cpu/cortexm_common/vectors_cortexm.c#L124-L133

into two functions that get called before and after the libc init, or to make board init responsible for initializing the C library.

In any case it is not an easy task, that's why I'm posting this as a comment and not a review, I don't want to block a PR that would prevent a lockup.

kaspar030: > I think the best user experience would be to make sure `stdio` becomes available as soon as possible.

+1

maribu:
I think that performing that check only when `DEVELHELP` is enabled is a good compromise. In production using `stdio` this soon during boot up is unlikely to happen. But in development it is nice for debugging. It would also be helpful to get a hint that UART output was discarded during development. Otherwise people soon will start to open issue that RIOT something silently discards `stdio` output. And in a separate PR we could try to enable `stdio` sooner in the boot process. (I could have a look into that.)

MrKevinWeiss:
@maribu agreed, done!

cpu/stm32_common/periph/uart.c
MrKevinWeiss: Sorry, done now.  Maybe I didn't force push.

MrKevinWeiss: @smlng ping

cpu/stm32_common/periph/uart.c
```diff
@@ -145,7 +145,12 @@ static inline void wait_for_tx_complete(uart_t uart)
 void uart_write(uart_t uart, const uint8_t *data, size_t len)
 {
     assert(uart < UART_NUMOF);
-
+    #if defined(DEVELHELP)
```
maribu: I'm personally in favor of not having any white space in front of preprocessor directives, because:
 1. C and C preprocessor macros are two completely distinct languages and keeping the indent of the C code surrounding it cannot be applied consistently. See example below
 2. Most C code does this. I'm in favor for keeping conventions, unless there is a reason for not doing so
 3. The rest of this file does not use white space in front of preprocessor directives
 4. Preprocessor directives become more visible, if the have a zero indent compared to the C code. And you really don't want to overlook them :-)

Here an example with unclear level of indent.
```` C
int foo(enum bar, int blah)
{
    switch(bar) {
        #ifdef MODULE_FOO /* <-- two levels of indent */
        case some_enum_value:
            some_function(blah);
            break;
            #endif /* <-- two or three levels of indent? */
    }
}
````

However, in the RIOT code base there are both files that *do* *not* use whitespace before preprocessor directives, and files that *do* *use* whitespace. So there is apparently no rule on it. So treat this comment as my personal opinion and not as a review.

Comment (maribu) [reply cpu/stm32_common/periph/uart.c]: You could also use `#ifdef DEVELHELP` here...

Comment (MrKevinWeiss) [reply cpu/stm32_common/periph/uart.c]: Ya, I should have checked that.  I also made it just #if DEVELHELP, though I am not sure if that is preferred.

**Comment by maribu** [comment]
+1 on merging

File: cpu/stm32_common/periph/uart.c (hunk 3 above)
Comment (jcarrano) [reply cpu/stm32_common/periph/uart.c]: > I'm personally in favor of not having any white space in front of preprocessor directives, because:

I'm in favour of avoiding preprocessor directives if possible:

```c
/* somewhere define a macro that is always defined */
#ifdef DEVELHELP
    #define DEVELHELP_ON 1
#else
    #define DEVELHELP_ON 0
#endif /* DEVELHELP */

/* then in the code you use a "real" if */

void uart_write(uart_t uart, const uint8_t *data, size_t len)
{
    assert(uart < UART_NUMOF);

    if (DEVELHELP_ON) {
        /* If tx is not enabled don't try to send */
         if (!(dev(uart)->CR1 & USART_CR1_TE)) {
             return;
         }
    }

/* rest of the function */
}
```

Isn't that much more readable? Also, you get the benefit of having the compiler SEE what is inside the if block always (even if it then gets thrown away.)

maribu: > I'm in favour of avoiding preprocessor directives if possible

Let me point out that your code contains 5 preprocessor directives, which are 3 more than the original code. (It also adds 5 lines of code.)

I personally think the original code was more readable.

MrKevinWeiss: I also agree with @maribu, and for the sake of uniformity.  I a bit think we are getting into a "lets delay this bugfix so we can be nit picky" territory.
...On that note I was thinking if it is a good idea to have something that would work if DEVELHELP==1 and crash if DEVELHELP==0.  Can we assume if DEVELHELP==0 the uart will always be initialized first?

maribu: Just to make sure I did understand you correct: By "DEVELHELP==1" you mean `DEVELHELP` is defined, and "DEVELHELP==0" means `DEVELHELP` is not defined, right?

The only reason for a valid (`assert(uart < UART_NUMOF)` did not trigger) not to be initialized seems to be it is used early in the boot up process. The only use case to me seems debug output via `stdio`. In production there should be no debug output of the early boot process. (E.g. `sys/auto_init` will run only after the periph buses and `stdio` is available, so there is not to much code running before `stdio` is available.)

So I believe it is safe to assume that when `DEVELHELP` is not defined that UARTs are not used before they are initialized.

MrKevinWeiss:
@leandrolanzieri Thanks for merging, now I can update the comment in the other PR!

PR #10619, opened 2018-12-15
drivers/ads101x: Changes for Common ADC API

gschorcht:
### Contribution description

This PR provides
- the changes of ADS101x/ADS111x driver for compatibility with the ADC extension API in PR #13247 which makes it possible to use ADS101x/ADS111x devices as ADC extensions.
- a shell based test application which demonstrates the use of the ADS101x/ADS111x devices
as ADC extension

This PR serves as a proof-of-concept for the Common ADC API in PR #13247 using an example of an existing ADC driver.

### Testing procedure

The test application can be used with up to two ADS101x/ADS111x devices. It is recommended to use two ADS1115 device where the the second ADS1115 is used as ADS1013 to see the differences between the two ADC extension devices. Compile the test application for any board that supports I2C.
```
CFLAGS="-DADS101X_PARAM_ADDR_2=\(0x49\) -DADS101X_PARAM_DEVICE_2=\(ADS101X_DEV_ADS1013\)" \
make flash -C tests/driver_ads101x_ext_api BOARD=...
```

### Issues/PRs references

Depends on PR #13247

### DIFF
--- a/tests/driver_ads101x_adc_ng/Makefile
+++ b/tests/driver_ads101x_adc_ng/Makefile
@@ -1,0 +1,13 @@
+BOARD ?= arduino-mega2560
+include ../Makefile.tests_common
+
+FEATURES_PROVIDED += adc_ng
+
+USEMODULE += ads101x
+USEMODULE += adc_ng
+USEMODULE += adc_ng_util
+USEMODULE += fmt
+USEMODULE += saul_default
+USEMODULE += shell
+
+include $(RIOTBASE)/Makefile.include
--- a/tests/driver_ads101x_adc_ng/README.md
+++ b/tests/driver_ads101x_adc_ng/README.md
@@ -1,0 +1,33 @@
+# Test for TI ADS101x/ADS111x as ADC extension
+
+## Overview
+
+This application can be used for testing the ADS101x/ADS111x device driver
+together with the Common ADC API. It uses the default parameter definition
+from the file `$(RIOTBASE)/drivers/ads101x/include/ads101x_params.h`.
+
+## Usage
+
+The shell commands `init`, `single`, `burst`, ... allow the initialization and
+reading of ADS101x/ADS111x channels with a certain full scale range.
+
+The ADS101x/ADS111x variant as well as the I2C bus and the address
+are defined by the standard parameter definition from the file
+`$(RIOTBASE)/drivers/ads101x/include/ads101x_params.h`. The parameter
+`ADS101X_PARAM_MUX_GAIN` is not used for the test application.
+
+By default, an ADS1115 is used as test device. To use another ADS101x/ADS111x
+variant, the parameter `DADS101X_PARAM_DEVICE` has to be overriden:
+```
+CFLAGS='-DADS101X_PARAM_DEVICE=ADS101X_DEV_ADS1013' \
+make -C tests/driver_ads101x_adc_ng BOARD=...
+```
+
+Parameters like `ADS101X_PARAM_I2C` and `ADS101X_PARAM_ADDR` can also be
+overridden from the make command line:
+```
+CFLAGS='-ADS101X_PARAM_I2C=I2C_DEV\(1\) -DADS101X_PARAM_ADDR=\(0x49\)'
+make -C tests/driver_ads101x_adc_ng BOARD=...
+```
+
+**Please note:** ADS1115 can also be used to test all other device variants.
--- a/tests/driver_ads101x_adc_ng/main.c
+++ b/tests/driver_ads101x_adc_ng/main.c
@@ -1,0 +1,524 @@
+/*
+ * Copyright (C) 2020 Otto-von-Guericke-Universität Magdeburg
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     tests
+ * @{
+ *
+ * @file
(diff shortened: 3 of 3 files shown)


gschorcht:
Requires some work to make compatible with PR #13247.

drivers/ads101x/ads101x.c
```diff
@@ -216,3 +216,70 @@ int ads101x_set_alert_parameters(const ads101x_alert_t *dev,
...
+    else {
+        mux_gain = (6 - ref) << 1;
+    }
+
+    switch (chn) {
+        case 0: mux_gain |= ADS101X_AIN0_SINGM; break;
+        case 1: mux_gain |= ADS101X_AIN1_SINGM; break;
+        case 2: mux_gain |= ADS101X_AIN2_SINGM; break;
+        case 3: mux_gain |= ADS101X_AIN3_SINGM; break;
+        default: return -1;
```
maribu:
```suggestion
        default: return -ENXIO;
```

drivers/ads101x/ads101x.c
```diff
@@ -216,3 +216,70 @@ int ads101x_set_alert_parameters(const ads101x_alert_t *dev,
...
+{
+    ads101x_t *dev = (ads101x_t *)handle;
+    DEBUG("%s: dev %p, dest %p\n", __func__, dev, dest);
+
+    assert(dev != NULL);
+
+    int16_t raw;
+    /* execute one conversion for the selected channel */
+    if (ads101x_read_raw(dev, &raw) != 0) {
+        return -1;
```
maribu:
```suggestion
        return -EIO;
```

Maybe that return value would be more informative during debugging?

drivers/ads101x/ads101x.c
```diff
@@ -216,3 +216,70 @@ int ads101x_set_alert_parameters(const ads101x_alert_t *dev,
...
+    }
+
+    /**
+     * since ADS101X/111X ADC returns the data as two's complement and
+     * ADC peripheral/extension API only support values in the range from
+     * 0 to 2^n-1, we have to transform the results to unsigned int values
+     * in this range.
+     *
+     * The two's complement to unsigned int in range of 0 ... 2^n-1
+     */
```
maribu: This comment should no longer be needed

gschorcht:
> Some comments inline.
> 
> I wonder if we couldn't get the test be driver independet. E.g. every non-peripheral driver could contain a check like this:
> 
> ```c
> #if (ADC_NG_NUMOF == 1) && defined(MODULE_ADS101X)
> const adc_ng_backend_t adc_ng_backends[ADC_NG_NUMOF] = {
>     {
>         .driver = &ads101x_adc_ng_driver,
>         .handle = &ads101x_devs[0]
>     }
> };
> #endif
> ```
> 
> Combined with auto-init this should just work without any support of either board or application, as long as exactly one ADC-NG driver (with only one entry in params) is used. I bet that this is also the common case. And only applications/boards actually having multiple ADCs would have go through the pain of providing the backends themselves.
> 
> Once the cross file arrays land in RIOT, this would also work without board/application support for any number of drivers :-)

:thumbs: Seems to be a good idea. I suppose you would place that code fragment in the drivers code.

gschorcht:
>  Seems to be a good idea. I suppose you would place that code fragment in the drivers code.

To place this in drivers code requires to make `ads101x_devs` visible, i.e. , to remove keyword `static` which is already the case for some other drivers.

gschorcht:
> > Seems to be a good idea. I suppose you would place that code fragment in the drivers code.
> 
> To place this in drivers code requires to make `ads101x_devs` visible, i.e. , to remove keyword `static` which is already the case for some other drivers. Furthermore, auto-init only works when `saul` is enabled.

@maribu It works indeed without any board definition if module `saul_default` is used and the driver code reuses the devices that are auto-initialized in `sys/aut-init/saul`. In fact, the only difference to `tests/adc_ng` is that module `adc_ng_default` is no used and modules `ads101x` and `saul_default` are used. `main.c` is exactly the same.

The question is how we can test each ADC driver that supports `adc_ng` without duplicating the test. It would be great if RIOT's test architecure would support variants of the same test application.

maribu:
> if module `saul_default`

I'm working on a PR to split auto-initialization of the drivers and SAUL, so that sensors/actuators can also be auto-initialized in case SAUL is not used.

> It would be great if RIOT's test architecure would support variants of the same test application.

Some tests can be controlled using environment parameters. E.g. we could introduce a variable like

``` Makefile
BACKEND ?= periph_adc_ng
```

and if someone would run with `BACKEND=ads101x make` instead, the `ads101x` backend would be chosen instead. That might not be a solution for hardware in the loop testing, but for manual testing this should be fine.

gschorcht:
> > if module `saul_default`
> 
> I'm working on a PR to split auto-initialization of the drivers and SAUL, so that sensors/actuators can also be auto-initialized in case SAUL is not used.

:+1:
 
> Some tests can be controlled using environment parameters. E.g. we could introduce a variable like
> 
> ```makefile
> BACKEND ?= periph_adc_ng
> ```
> 
> and if someone would run with `BACKEND=ads101x make` instead, the `ads101x` backend would be chosen instead. That might not be a solution for hardware in the loop testing, but for manual testing this should be fine.

Hm, that still can't be automated. What I meant is to have a script that automatically compiles and tests variants of the same test applications, for example for different `BACKEND` modules.

maribu:
Not stale, I'll resume work on the ADC-NG API hopefully soon. Then this can get in as well :-)

PR #10621, opened 2018-12-17
cpu/nrfx: rework periph_uart driver to allow use of multiple UARTs with nrf52840

aabadie:
### Contribution description

The initial goal of this PR was to update the UART configuration of the nrf52840dk board with a second UART, since there are 2 available on this CPU. I need this for #10563 where external radios devices (for WiFi, NB_IoT and maybe GSM) uses the second UART with AT commands.

When working on this, I realized the second UART needs a different struct defined in the CMSIS (NRF_UART**E**_TYPE) so it couldn't be added as-is. But the good thing is that the first UART can also use this other struct. Then the only missing thing was adapting the `periph_uart` driver to this new struct: it requires to use EasyDMA for transfering data.

So, to summarize, this PR provides several things:
- adapts the `periph_uart` to be compatible with NRF_UARTE_TYPE, only used with nrf52840xxx cpus
- provides hardware flow control with nrf52840 cpus and configure STDIO to use this on nrf52840dk
- use the array style configuration for UART config with nrf52840dk board (or nrf52840 based boards in general)
- configure the second UART available on the Arduino pinout of this board.

This PR was tested with success using the `tests/periph_uart` application.

There is also a small increase of ROM size (+36 bytes) for nrf51/nrf52 based boards. The ROM size increase is more important with nrf52840dk, (+190 bytes), probably because there's a second UART and also some parts are now very different from the initial driver.
Note that I'm not an expert in ROM size optimization, so any help or advice is welcome.

Also note that this PR is based on #10279 for simplicity (because common files are modified/moved by both PRs).

### Testing procedure

If you don't have all the hardware, you can use the boards provided by IoT-LAB to verify everything is still working:

- Start an experiment on IoT-LAB
```
$ iotlab-experiment submit -n pr -d 60 -l 1,site=saclay+archi=nrf51dk:ble -l 1,site=saclay+archi=nrf52dk:ble -l 1,site=saclay+archi=nrf52840dk:multi
```
- Flash any application using the shell on the boards:
```
$ make BOARD=nrf51dk IOTLAB_NODE=auto-ssh -C examples/saul flash term
$ make BOARD=nrf52dk IOTLAB_NODE=auto-ssh -C examples/saul flash term
$ make BOARD=nrf52840dk IOTLAB_NODE=auto-ssh -C examples/saul flash term
```
- Verify the shell works with all of them. This will validate that the existing driver (nrf51dk and nrf52dk) still works but also that the UART with EasyDMA is working as well (nrf52840dk).

Regarding the test of the second UART configured on the nrf52840dk, there are 2 possibilities:
1. Physically grab on this board, flash the `tests/periph_uart` application on it, shortcut RX/TX (D0/D1) pins, initialize the second UART and send data:
```
$ make BOARD=nrf52840dk -C tests/periph_uart flash term
> init 1 115200
> send test
```
(You can also try different baudrates if you want)
2. Trust me that it also works ;)

### Issues/PRs references

This PR is based on #10279 and will be required by #9833 and #10563 very soon.


MrKevinWeiss:
It looks pretty good.  I don't know a way around the byte size increase, it is an increase that allows more than one uart, which is pretty acceptable I think.  Even with only one UART configured for the nrf52840 it increases 112 bytes but I still think it is necessary.

aabadie:
@haukepetersen, any chance that you could have a look at this PR. It will be useful for #10563 and #9833 is also relying on it now.

cpu/nrf5x_common/periph/uart.c
```diff
@@ -31,39 +33,68 @@
...
+#ifndef UART_HWFLOWCTRL
+#define UART_HWFLOWCTRL  0
+#endif
 #endif
 
 /**
  * @brief Allocate memory for the interrupt context
  */
-static uart_isr_ctx_t uart_config;
+static uart_isr_ctx_t isr_ctx;
```
haukepetersen: just a single ISR context for (potentially) more than 1 configured UART interface? This seems broken to me. Probably this should be 
```c
static uart_isr_ctx_t isr_ctx[UART_NUMOF];
```
right?

cpu/nrf5x_common/periph/uart.c
```diff
@@ -76,128 +107,185 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
-    gpio_init(UART_PIN_CTS, GPIO_IN);
-    /* configure RTS and CTS pins to use */
-    PSEL_RTS = UART_PIN_RTS;
-    PSEL_CTS = UART_PIN_CTS;
-    NRF_UART0->CONFIG |= UART_CONFIG_HWFC_Msk;  /* enable HW flow control */
-#else
-    PSEL_RTS = 0xffffffff;            /* pin disconnected */
-    PSEL_CTS = 0xffffffff;            /* pin disconnected */
-#endif
+    if (UART_HWFLOWCTRL) {
```
haukepetersen: if I see it correctly, this would disallow for using hw flow control for all non nrf52840 platforms, right? But we want to use that feature also with the nrf51 etc...

cpu/nrf5x_common/periph/uart.c
aabadie: oops

cpu/nrf5x_common/periph/uart.c
aabadie: In the case of nrf52840 family, this MACRO is defined [above](https://github.com/RIOT-OS/RIOT/pull/10621/files#diff-09415752bfc5dc7483589c86f7a1f0a5R45) and depends on the fact that `rts_pin` and `cts_pin` are defined to something different from `GPIO_UNDEF`.

cpu/nrf5x_common/periph/uart.c
aabadie: Fixed using preprocessor macros...

haukepetersen:
finally I got to look into this in more detail. I don't like the fact, that this PR introduces additional code size for platforms that are not effected (e.g. `nrf51dk`, `nrf52dk`, ...). Will propose some fixes to your PR that will prevent this, just seems that I broke the `nrf52840dk`s UART right now. Will fix it and then PR against your branch.

haukepetersen:
done: https://github.com/aabadie/RIOT/pull/10

This PR had actually some serious issues, but should be at least working with the proposed fixes.

aabadie:
Thanks @haukepetersen, I merged your proposition. Let me know when I can squash.

haukepetersen:
feel free to squash (including my commit...).

aabadie:
> feel free to squash (including my commit...).

it won't be easy, I'll try to keep it (without the fixup)

PR #10624, opened 2018-12-18
doc/rdm: RDM on a common api for testing

MrKevinWeiss:
### Contribution description

This is the initial commit of Guidelines for Write Firmware to Expose RIOT APIs RDM.
It attempts to standardize some semantics for firmware written in tests.  This should allow easier parsing, sorting and filtering of useful information for tests and experiments. without negatively effecting other parts of RIOT.  It's aim is to provide answers to simple questions of something like, what should I expect at the end of a test, "success", "Success", "SUCCESS", "PASS", "FAIL", "ERROR", "-1", "0".  Something we can all agree upon and, if we adhere to it utilize the standardization for other things.

### DIFF
--- a/doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
+++ b/doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
@@ -1,0 +1,141 @@
+- rdm-draft-weiss-exposing-api-firmware-guidelines.md
+- Title: Guidelines for Write Firmware to Expose RIOT APIs
+- Author: Kevin Weiss
+- Status: In-progress
+- Type: Process
+- Date: December 2018
+
+## Abstract
+
+In order to improve the testing and interface tools for RIOT a standard way of
+exposing API calls to a higher level should be defined. This document describes
+a set of guidelines indicating semantics and basic design on the firmware level
+that allow interfaces and tools to be build upon.
+
+## Status
+
+This document is currently under open discussion. Its content is licensed with
+a Creative Commons CC-BY-SA license.
+
+## Terminology
+This memo uses the RFC2119 terminology and the following acronyms:
+API: Application Programming Interface
+cmd: Command
+msg: Message
+
+# 1. Introduction
+The guidelines are used indicate the semantics and design when writing
+firmware to expose APIs through the shell.
+
+_NOTE: It may not be applicable for every API call_
+
+ # 2. Test Firmware Design
+
+The following guidelines are used when writing the C program to be flashed to
+the board.
+
+ ## 2.1. Test Firmware Declarations
+
+- Tests should interact through a shell
+- Commands should be synchronous
+- Tests should have a `get_metadata` command
+- The `get_metadata` command should contain at least board name, and
+application name
+- Tests should have a `echo <random 8 hex digits>` command
+- Test commands should provide similar API/function calls
+- Tests that require `test_<name of test command>` command only if interaction
+with the shell causes inaccuracies
+- Commands that use `test_<name of test command>` should give as much data
+and information as possible to reproduce command and allow higher level
+functionality
+- Asynchronous actions should be hidden with the `test_<name of test command>`
+format
+- There should be a standard interface for every command sent
+[cmd, data, msg, result]
+- [cmd] should only be provided once for each command
+- [cmd] should allow the user to reproduce the step, this may be the
+API/function call with the variables or a descriptive name for a set of calls
+- [msg] should be optional
+- [msg] if used, may be provided several times
+- [msg] should not be actionable and only contain debug/log
+information
+- [data] should be optional
+- [data] if used, may be provided several times
+- [data] should be actionable and adhere to pythons
+ast.literal_eval formatting
+- [result] should only be provided once for each command
+- [result] value should only be "SUCCESS" or "ERROR"
+- [result] should be the last value in the command response.
+
+ ## 2.2. Expected Response Format From Commands In Firmware
+
+- [cmd] -> `printf("{'cmd': 'api_call(%i, %i, &input_data)'}\n", arg1, arg2)`
+- [msg] -> `printf("This is useful debug info, anything that has no
+structure will be saved as a [msg]\n")`
+- [data] -> `printf("{'data': %i}\n", data)`
+- [data] -> `printf("{'data': '%s'}\n", actionable_string)`
+- [result] -> `printf("{'result': 'SUCCESS'}\n")` or `printf("{'result':
+'ERROR'}\n")
+
+## 2.3.  Basic Real Example
+This example show how an real api call should be exposed.
+```c
+int cmd_i2c_read_reg(int argc, char **argv)
+{
+    // Do parsing
+    ...
+    printf("{'cmd': 'i2c_read_reg(%i, 0x%02X, %02X, &data, 0x%02X)'}\n",
+           dev, addr, reg, flags);
+    ret = i2c_read_reg(dev, addr, reg, &data, flags);
+    if (ret == I2C_ACK) {
+        printf("Successfully read from register\n");
+        printf("{'data': 0x%02X}\n", data);
+        printf("{'result': 'SUCCESS'}\n");
+    }
+    else {
+        printf("Errorcode: %s\n", _get_error_code_string(ret));
+        printf("{'data': %i}\n", ret);
+        printf("{'result': 'ERROR'}\n");
+    }
+    return ret;
+}
+```
+
+## 2.4. Multiple Data Example
+
+This example shows how large amounts of data can be sent with one call.
+It shows multiple data responses.
+It also shows a [cmd] field that doesn't exactly match the api call but
+gives enough information to be able to reproduce.
+
+ ```c
+ int cmd_get_lots_of_data(int argc, char **argv)
+ {
+     // Do parsing
+     ...
+     printf("{'cmd': 'get_data_api_command(get_data_arg=[0 to %i])'}\n",
+            amount_of_api_calls)
+     printf("Getting several values for the api command\n")
+     for (i = 0; i < amount_of_api_calls; i++) {
+         data = get_data_api_command(i);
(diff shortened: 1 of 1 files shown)


doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
```diff
@@ -0,0 +1,141 @@
...
+- [cmd] should only be provided once for each command
+- [cmd] should allow the user to reproduce the step, this may be the
+API/function call with the variables or a descriptive name for a set of calls
+- [msg] should be optional
+- [msg] if used, may be provided several times
+- [msg] should not be actionable and only contain debug/log
+information
+- [data] should be optional
+- [data] if used, may be provided several times
+- [data] should be actionable and adhere to pythons
```
jcarrano: Why not use something compatible with JSON?

a. There is code to generate that already
b. You don't tie the receiving end to python.

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
```diff
@@ -0,0 +1,141 @@
...
+format
+- There should be a standard interface for every command sent
+[cmd, data, msg, result]
+- [cmd] should only be provided once for each command
+- [cmd] should allow the user to reproduce the step, this may be the
+API/function call with the variables or a descriptive name for a set of calls
+- [msg] should be optional
+- [msg] if used, may be provided several times
+- [msg] should not be actionable and only contain debug/log
+information
```
jcarrano: I think that "[msg] should not start with {" and data should. That way it is unambiguous.

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
```diff
@@ -0,0 +1,141 @@
...
+- [data] should be optional
+- [data] if used, may be provided several times
+- [data] should be actionable and adhere to pythons
+ast.literal_eval formatting
+- [result] should only be provided once for each command
```
jcarrano: "At most once" or "exacly once"?

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
```diff
@@ -0,0 +1,141 @@
...
+and information as possible to reproduce command and allow higher level
+functionality
+- Asynchronous actions should be hidden with the `test_<name of test command>`
+format
+- There should be a standard interface for every command sent
+[cmd, data, msg, result]
+- [cmd] should only be provided once for each command
+- [cmd] should allow the user to reproduce the step, this may be the
+API/function call with the variables or a descriptive name for a set of calls
+- [msg] should be optional
```
jcarrano: Instead of writing so much text, you could roughly specify the grammar of the response:

```
response := message* data* result

message := MESSAGE_CONTENTS EOL

data := <json> EOL

result := "{" "'result'" ":" RESULT_VALUE "}" EOL

EOL := '\n'
MESSAGE_CONTENTS := ([^{].+)?
RESULT_VALUE := (SUCCESS) | (ERROR)
```

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
MrKevinWeiss: The reason for this is a mechanism to catch anything that shouldn't be formatted.  This way not data can be missed.

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
MrKevinWeiss: I like the idea, if I use standard json do I need to know the full frame?  I would worry more about the coordination the messages need.

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
jcarrano: Ok, but how do you differentiate between data and message. Another option is to say that anything that cannot be successfully parsed as data is a message.

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
MrKevinWeiss: Exactly, I will update!  It is used to end the message so we don't have to wait for timeouts.

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
jcarrano: Here I'm assuming you are going to put a single [data] element per line. If that is not the case, it is not a problem either, because json can be parsed from a stream (meaning you could pretty format the [data] into multiple lines.)

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
MrKevinWeiss: Can I have both? I will add that in.

MrKevinWeiss:
Thanks @jcarrano, that is a very useful review!

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
MrKevinWeiss: Ok, can we say a either json compatible line or generic string which defaults to the [msg] form?  I will update the examples.

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
MrKevinWeiss: Agreed that's what I will do!

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
```diff
@@ -0,0 +1,155 @@
...
+- Title: Guidelines for Write Firmware to Expose RIOT APIs
+- Author: Kevin Weiss
+- Status: In-progress
+- Type: Process
+- Date: December 2018
+
+## Abstract
+
+In order to improve the testing and interface tools for RIOT a standard way of
+exposing API calls to a higher level should be defined. This document describes
```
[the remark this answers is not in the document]
tcschmidt: YES: It will be really helpful to make the context and subject of consideration crisply clear so that the reader understands right away what to expect from this memo.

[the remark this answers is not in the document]
danpetry: Ditto, NB that "exposing API calls" refers to a module exposing functions for other modules to use. Not to a human

danpetry:
My initial thoughts are that the way these rules are specified would make applying them quite onerous for developers: they would have to go through all bullet points one by one. Perhaps a format such as a commented template (or maybe condensing and structuring them somewhat) would make it easier for people to use (and more likely for it to get used therefore).

danpetry:
Also, is there any scope for this document to take into account other aspects of testing? For example, when things should be unit tested using embunit or have CLI tests written using testrunner, and the kind of test coverage that's expected? Or outlining what our eventual/ideal CI strategy would be? The testing in general is very varied seemingly without strong common logic in what is tested and how, not just in the area of how tests are exposed through CLI output

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
```diff
@@ -0,0 +1,155 @@
...
+the board.
+
+## 2.1. Test Firmware Declarations
+
+- Tests should interact through a shell
+- Commands should be synchronous
+- Tests should have a `get_metadata` command
+- The `get_metadata` command should contain at least board name, and
+application name
+- Tests should have a `echo <random 8 hex digits>` command
```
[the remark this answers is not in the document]
MrKevinWeiss: Yup, I will try to implement something, I suppose it would be like the ps/reboot/ect. where you can just include it so there will be no code duplication.

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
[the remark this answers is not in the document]
MrKevinWeiss: OK, I will try to rephrase it in a clearer way.

MrKevinWeiss:
> Perhaps a format such as a commented template (or maybe condensing and structuring them somewhat)

Can you show me what you mean, maybe give an example?

danpetry:
>> Perhaps a format such as a commented template (or maybe condensing and structuring them somewhat)

> Can you show me what you mean, maybe give an example?

Maybe presenting the details of [msg data result] etc would look better in a table. On second look the spec is probably too generic to provide a single template although the code examples help. In general some other way of presenting than a long list of bullet points would be good. Potentially breaking up the points into subsections, or enumerating them, or using tables might help break it up/give it some high level structure to make the points more easy to understand/digest/contextualise

danpetry:
> that was how it started but was suggested to reduce the scope
ok, understood

doc/memos/rdm-draft-weiss-exposing-api-firmware-guidelines.md
```diff
@@ -0,0 +1,155 @@
...
+ int cmd_get_lots_of_data(int argc, char **argv)
+ {
+     // Do parsing
+     ...
+     printf("{\"cmd\": \"get_data_api_command(get_data_arg=[0 to %i])\"}\n",
+            amount_of_api_calls)
+     printf("Getting several values for the api command\n")
+     for (i = 0; i < amount_of_api_calls; i++) {
+         data = get_data_api_command(i);
+         prqintf("{\"data\": 0x%X}\n", data);
```
MrKevinWeiss: oops...

MrKevinWeiss:
I am currently working on implementation of this.  It will require an update but I don't think it will be forgotten!

MrKevinWeiss:
I think this is out of date and somehow a part of the turo + riotctrl.

PR #9171, opened 2018-05-23
cpu/stm32_common: add DMA support for all stm32

vincent-d:
### Contribution description

This adds DMA support for all stm32.
STDIO UART works fine on nucleo-f091 and on b-l072z-lrwan1. But SPI doesn't work on b-l072z-lrwan1, not tested on nucleo-f091.
I'm not sure I'll have a lot of time to continue this soon.

### Issues/PRs references

Follow-up #7658

PR #10633, opened 2018-12-18
drivers: stm32 eth peripheral driver

crest42:
### Contribution description
Hi All,

as discussed in #5643 this is a new PR based on @lebrush's work.
The original PR implemented an ethernet peripheral driver for stm32f4 boards. 

Changes:

1. Rebase to current master branch
1.1. Rework driver to the new iolist netdev interface
1.2. Rework driver to use new vendor headers
2. Use dynamic MAC Addresses based on board uid
3. Extend driver to work with stm32f7 boards as well
4. Add GNRC Testcase

Open Tasks:
1. Adapt to DMA API?
2. Fix static buffer used as workaround for iolist
3. ~~Maybe use network interface as default interface for boards with on board eth?~~ Implemented
4. Fix a few details (mentioned in in-line comments)

### Testing procedure
1. Go to tests/stm32_eth_lwip or stm32_eth_gnrc, build an flash.
2. Use ICMP or UDP on port 12345 to test the echo server.

I testes with an Nucleo-F767ZI. Original work was tested with STM32F4Discovery. IHMO This should work with every F4 and F7 Board with ethernet peripheral. Would be great if somebody with access to any F4 or F7 hardware could test as well.

### Issues/PRs references
see #5643 
All Credits to the great work of @lebrush who saved my thesis.
Also involved: @MrKevinWeiss  @aabadie @smlng


boards/common/nucleo/Makefile.dep
```diff
@@ -1,3 +1,7 @@
 ifneq (,$(filter saul_default,$(USEMODULE)))
   USEMODULE += saul_gpio
 endif
+
+ifneq (,$(filter periph_stm32_eth,$(FEATURES_USED)))
+  USEMODULE += luid
```
crest42: I am certain this is not the right way to do. I need the luid module for mac address generation. Any better ideas?

boards/nucleo-f767zi/include/periph_conf.h
```diff
@@ -126,6 +126,38 @@ static const uart_conf_t uart_config[] = {
...
+ #define ETH_DMA_ISR        isr_dma2_stream0
+ static const eth_conf_t eth_config = {
+    .mode = RMII,
+    .mac = { 0 },
+    .speed = ETH_SPEED_100TX_FD,
+    .dma_chan = 0,
+    .dma_stream = 8,
+    .phy_addr = 0x01,
+    .pins = {
+        GPIO_PIN(PORT_G, 13),
```
crest42: New Pinout for F767ZI. I would assume this is common for all F7 devices. I will check. If my assumption is right this should move to f7 common, right?

boards/stm32f4discovery/Makefile.features
```diff
@@ -2,10 +2,14 @@
 FEATURES_PROVIDED += periph_adc
 FEATURES_PROVIDED += periph_dac
 FEATURES_PROVIDED += periph_gpio periph_gpio_irq
+FEATURES_PROVIDED += periph_eth
+FEATURES_PROVIDED += periph_gpio
```
crest42: This shouldnt be here. I will check

cpu/stm32_common/periph/Makefile
```diff
@@ -15,4 +15,8 @@ ifneq (,$(filter periph_flashpage periph_eeprom,$(USEMODULE)))
   SRC += flash_common.c
 endif
 
+ifneq (,$(filter periph_stm32_eth,$(FEATURES_PROVIDED)))
+  SRC += eth.c
```
crest42: How to include the pheriph driver automatically when FEATURES_REQUIRED is set? I dont this this is the correct way

cpu/stm32_common/periph/eth.c
```diff
@@ -0,0 +1,546 @@
...
+#include "net/netdev/eth.h"
+#include "net/eui64.h"
+#include "net/ethernet.h"
+#include "net/netstats.h"
+#include "net/phy.h"
+
+#define ENABLE_DEBUG (0)
+#include "debug.h"
+
+char send_buf[ETH_TX_BUFFER_SIZE];
```
crest42: This is a work around for using iolist. I will try to fix this soon.

gebart:
I suggest that this driver is moved to drivers/stm32_eth or similar.

boards/stm32f4discovery/include/periph_conf.h
```diff
@@ -211,6 +211,41 @@ static const i2c_conf_t i2c_config[] = {
 #define I2C_NUMOF           (sizeof(i2c_config) / sizeof(i2c_config[0]))
 /** @} */
 
+/**
```
aabadie: I know it was done this way in the initial PR, but I'm against having an eth peripheral configured by default on a board that doesn't provide an Ethernet interface.
I would prefer to have it on the nucleo-f429zi for F4 CPUs.

cpu/stm32f4/include/periph_cpu.h
```diff
@@ -84,6 +84,122 @@ typedef struct {
...
+    uint8_t dma_stream;                 /**< DMA stream used for TX */
+    uint8_t dma_chan;                   /**< DMA channel used for TX */
+    char mac[6];                        /**< Et hernet MAC address */
+    char phy_addr;                      /**< PHY address */
+    gpio_t pins[];                      /**< Pins to use. MII requires 18 pins,
+                                             RMII 9 and SMI 9. Not all speeds are
+                                             supported by all modes. */
+} eth_conf_t;
+
+/**
```
aabadie: all DMA stuff shouldn't be there. There is a `periph_dma` already available (I have to re-check this). You can also have a look at #9171 (and maybe also base this PR on it ?)

boards/stm32f4discovery/include/periph_conf.h
crest42: Maybe I should also check if there other boards from the 4/7 family which got Ethernet. Should work for other boards from the same family as well.

crest42:
Thanks all for the feedback. I plan to continue to work on this in Jan/Feb and implement your suggestions.

crest42:
I just pushed a new version with some smaller changes:

1. I split the driver in a netdev part located under driver/stm32_eth and the original driver still located in stm32_common/periph/eth.c
2. Implement auto init for the netdev driver
3. Moved the eth config from the stm32f4discovery to  nucleo-f429zi and added nucleo-f746zg

Next milestone is the dma stuff.

boards/nucleo-f767zi/include/periph_conf.h
crest42: Same for all f7 devices but I will let it in place. I think it is unnecessary overhead for all non ethernet boards.

boards/stm32f4discovery/include/periph_conf.h
crest42: Removed by now since the DMA interface changed and I've got no test board for the f4 family. If somebody have a board to test i will add this again in the correct file

crest42:
A Short update with pretty good news so far. I was able to migrate the driver to the stm32 DMA API and run a successful test for an f767zi board. Thanks for the great work @vincent-d and @aabadie.

I had to remove the configuration for the f429zi since the DMA configuration is missing and I have no test board available. 

If some of you have the time it would be great if you could review this PR again :)

@aabadie iirc you have a f429zi available? Would you mind to add the configuration again after a successful test? The following should work:

```
/**
 * @name ETH configuration
 * @{
 */
#define ETH_NUMOF           (1)
#define ETH_RX_BUFFER_COUNT (4)
#define ETH_TX_BUFFER_COUNT (4)

#define ETH_RX_BUFFER_SIZE (1524)
#define ETH_TX_BUFFER_SIZE (1524)

static const eth_conf_t eth_config = {
    .mode = RMII,
    .mac = { 0 },
    .speed = ETH_SPEED_100TX_FD,
    .dma = 0,
    .dma_chan = 0,
    .phy_addr = 0x01,
    .pins = {
        GPIO_PIN(PORT_G, 13),
        GPIO_PIN(PORT_B, 13),
        GPIO_PIN(PORT_G, 11),
        GPIO_PIN(PORT_C, 4),
        GPIO_PIN(PORT_C, 5),
        GPIO_PIN(PORT_A, 7),
        GPIO_PIN(PORT_C, 1),
        GPIO_PIN(PORT_A, 2),
        GPIO_PIN(PORT_A, 1),
    }
};
/** @} */

/**
 * @name    DMA streams configuration
 * @{
 */
#ifdef MODULE_PERIPH_DMA
static const dma_conf_t dma_config[] = {
    { .stream = 8 },    /* DMA2 Stream 8 - ETH_TX    */
};

#define DMA_0_ISR  isr_dma2_stream0

#define DMA_NUMOF           (sizeof(dma_config) / sizeof(dma_config[0]))
#endif
/** @} */
```

aabadie:
Thanks for the work @crest42, that's promising.

> you have a f429zi available? Would you mind to add the configuration again after a successful test?

Yes and yes but not before next week.

crest42:
Take your Time. I still need to get rid of the static buffer anyways. I hope to fix this next week so we can finish this PR :)

boards/stm32f4discovery/Makefile.features
crest42: Removed

cpu/stm32_common/periph/Makefile
crest42: Removed

cpu/stm32_common/periph/eth.c
crest42: fixed. will push the update soon

crest42: Fixed

crest42:
Ok, I just get rid of the static buffer. For this, I simply assume that an iolist only hold a single Ethernet frame at a time. This makes things a lot easier and maintainable.

I appreciate any feedback :)

kaspar030:
I get these (I'm trying gnrc_networking):

```
stm32_eth.c: In function '_recv':
stm32_eth.c:79:15: error: 'netdev_t' {aka 'struct netdev'} has no member named 'stats'                                                       
         netdev->stats.rx_count++;
               ^~
stm32_eth.c:80:15: error: 'netdev_t' {aka 'struct netdev'} has no member named 'stats'                                                       
         netdev->stats.rx_bytes += len;
               ^~
stm32_eth.c: In function '_send':
stm32_eth.c:104:11: error: 'netdev_t' {aka 'struct netdev'} has no member named 'stats'                                                      
     netdev->stats.tx_bytes += ret;
           ^~
```

Haven't netstats been moved into gnrc completely?

Other than that, I've tried ping with link-local addresses on ethernet, which works fine.

crest42:
Thanks for the feedback. I never tried with The netstats Module enabled. I will fix this Monday.

kaspar030:
I've been thinking about the namespacing.

> Also, the netdev adaption is done in the cpu implementation, and I think it should be somewhere in `drivers/`. The CPU will only provide `eth_init`, `eth_read` and `eth_write`.
> Is it possible to have multiple eth interfaces on the same board ?
> 
> I have other small comments below. But I think the global design needs more work. Maybe one day, we will want to provide a `periph_eth` for the FRDM_KW64F (Kinetis). If everything is already in place, it will be easier to add.

@aabadie  I'm not sure this makes sense. It would basically mean that "periph_eth" is the interface that is to be implemented. But ethernet is fundamentally different from uart, gpio, timers as in *no* application will directly use the API. They will use netdev. So IMO there's not much need to unify below that. I mean, just think of netdev as "periph_netdev"...

What might cause trouble is using ```eth_*()``` as function namespace...

Thus I suggest:

- keep stm32_eth as feature (for conditional compilation, only provide that for boards that have the periperal, the wiring and the configuration)
- drop periph_eth and periph_stm32_eth features (no need to use either without netdev)
- move all of the implementation into ```cpu/stm32_common/stm32_eth```
- rename all (global) functions to ```stm32_eth_*()```

Looking forward to get this in!

crest42:
> Haven't netstats been moved into gnrc completely?
> 
> Other than that, I've tried ping with link-local addresses on ethernet, which works fine. _edit_ on nucleo-f767zi.

Yep, l2_netstats are implemented in grnc_netif. I Removed the code and implemented the appropriate calls to the netdev event callback.  l2_Netstats should work now.

crest42:
> * keep stm32_eth as feature (for conditional compilation, only provide that for boards that have the periperal, the wiring and the configuration)
> * drop periph_eth and periph_stm32_eth features (no need to use either without netdev)
> * move all of the implementation into `cpu/stm32_common/stm32_eth`
> * rename all (global) functions to `stm32_eth_*()`

I'm fine with whatever you two agrees on. Maybe i would keep the discrimination between the netdev wrapper in drivers/ and the peripheral implementation under cpu/stm32_common. Logically this are separate parts and can be exchanged independently of each other.

crest42:
Already got some feedback for me regarding tests and the layout of the code/files?

crest42:
Force pushed after rebase against current master and resolved conflict. Test for stm32f767zi still working without problems.

aabadie:
Let's give an update on this one:
I tested this PR on nucleo-f767zi and was able to ping and exchange udp packets with it from my laptop which is already awesome :) Thanks @crest42, this is good job.

I tried your test applications, based on gnrc and lwip, it worked (IPv6 and v4) but I noticed some stability issues after reflashing the board. Sometimes I have trouble in getting the ethernet back in a usable state: I have to start/stop my eth interface on Linux, reboot the board to get things back to normal. Maybe there's something wrong with the initialization code or something (I haven't looked in detail) ?

To move forward fast, I allowed myself to change a few minor things:
- rebased the branch on latest master
- fixed a few rebase minor issues
- did some tiny style cleanup in the implementation eth.c file
- dropped some remaining occurrences of periph_stm32_eth
- added nucleo-f767zi in the list of boards that netif in `examples/default` Makefile

More generally, I think the test applications are not needed if the board is added to `examples/default` since there's an auto_init support so the CI will at least build the eth peripheral driver in this case.
If people wants to test higher level feature, they can try gcoap or gnrc_networking applications, it should also work (thanks to auto_init).

@kaspar030, it would be great if you could have a look at the code organisation and tell if you are happy with the current state.

crest42:
Thanks for the testing and your changes. I also was able to observe the problem you described. Only the yellow led on the eth port is on sometimes. Normally it works after a reset. I will have a look into it.

crest42:
@aabadie sadly I am not able to reproduce the problem observed before. I Did ~30 flashes and ethernet came up every time. Are you still able to reproduce the problem?

MrKevinWeiss:
Hmmm.  The soft feature freeze is tomorrow... is there any change of finishing it then? Otherwise I think it would be best to push this later.

crest42:
IMHO everything works and little to no existing infrastructure is modified. So if @kaspar030 or someone else can do a code review in such a short time it should be fine. The chances of breaking anything are pretty low.

The Problem (sometimes) observed by @aabadie could also be fixed later on. Atm I can not reproduce the problem on my machine and we could fix it via a new issue.

MrKevinWeiss:
Maybe we can also ping @smlng to help.

tests/stm32_eth_gnrc/Makefile
```diff
@@ -0,0 +1,13 @@
+APPLICATION = stm32_eth_gnrc
```
kaspar030: isn't this application basically part of examples/default?

tests/stm32_eth_lwip/README.md
```diff
@@ -0,0 +1,21 @@
+Tests for stm32 Ethernet periph driver
```
kaspar030: same here, can we reuse another lwip test application, or make this a generic "lwip_udp_echo" application that pulls in stm32_eth using ```FEATURES_OPTIONAL += stm32_eth```?

kaspar030:
I talked offline with @aabadie, he remembers the code is actually working, so do I, so ACK after resolving the name issue and dropping the test applications. Please directly squash after fixing!

crest42:
About to do the requested changes. Not sure what you mean with squashing. Only squash the new changes with the naming fixes or the complete commit history of this PR?

kaspar030:
>  The soft feature freeze is tomorrow...

Well, the "soft feature freeze" is supposed to be the last date for high-impact features (meaning, code that touches a lot of other code). A driver can be added until the last minute before release testing starts.
Anyhow, this should be in soon. ;)

kaspar030:
> Not sure what you mean with squashing. Only squash the new changes with the naming fixes or the complete commit history of this PR?
> 
> Further, squash everything into a single commit?

Your commit messages are all well formatted. Usually they're not. Anyhow, we usually squash the history of fixes very small, like:

1. a commit adding the driver code in ```drivers/stm32_eth``` (and maybe the ```cpu/stm32.../``` updates)
2. a commit adding support for the driver to the f767
3. a commit adding a test application (or adding the driver to an existing one)

You get the idea.

If you think the individual commits are very important, we can also keep them, but most RIOT'ers prefer combining them.

crest42:
Sure. I think the easy way is to combine commits by the author. If something is wrong it is probably reverted all together anyway.

I will have access to my testbed in the evening. After another successful test with the current master and after the renaming of the functions I going to drop the test application and send a (hopefully) ready-to-merge update.

MrKevinWeiss:
> Well, the "soft feature freeze" is supposed to be the last date for high-impact features (meaning, code that touches a lot of other code)

True...
I did want this to be included in the soft feature freeze but it would be pretty cool to have this in and it seems like it won't be long so lets do it.  I will push the testing of this to just before the hard feature freeze then.

Good luck guys!

crest42:
Good news everybody. I forced pushed a new version with the following changes:

- Rename of eth_phy_[read|write] to stm32_eth_phy_[read|write]
- Remove of the test applications
- Rebase to the current master branch
- Squashed the commits into 3 big commits

I was able to make a successful test using tests/gnrc_udp (icmp echo & udp receive).

Did i miss anything?

Regards,
Robin

kaspar030:
ACK. Nice!

kaspar030:
The ethernet configuration in periph_conf.h seems to need a guard.

crest42:
> The ethernet configuration in periph_conf.h seems to need a guard.

Fixed

boards/nucleo-f767zi/include/periph_conf.h
```diff
@@ -158,6 +160,42 @@ static const spi_conf_t spi_config[] = {
 #define SPI_NUMOF           (sizeof(spi_config) / sizeof(spi_config[0]))
 /** @} */
 
+#ifdef MODULE_STM32_ETH
```
aabadie: I don't think this is needed. Did you get an error with the CI because of this ?

boards/nucleo-f767zi/include/periph_conf.h
```diff
@@ -158,6 +160,42 @@ static const spi_conf_t spi_config[] = {
...
+/**
+ * @name ETH configuration
+ * @{
+ */
+#define ETH_NUMOF           (1)
```
aabadie: This is not used anywhere, so should be removed.

boards/nucleo-f767zi/include/periph_conf.h
```diff
@@ -158,6 +160,42 @@ static const spi_conf_t spi_config[] = {
...
+#define ETH_RX_BUFFER_COUNT (4)
+#define ETH_TX_BUFFER_COUNT (4)
+
+#define ETH_RX_BUFFER_SIZE (1524)
+#define ETH_TX_BUFFER_SIZE (1524)
+
+#define ETH_DMA_ISR        isr_dma2_stream0
```
aabadie: I think all defines should go under the struct.

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -708,6 +708,168 @@ int dma_configure(dma_t dma, int chan, const volatile void *src, volatile void *
 #include "candev_stm32.h"
 #endif
 
+#ifdef MODULE_STM32_ETH
```
aabadie: This should go before the eth functions declaration, e.g. [here](https://github.com/RIOT-OS/RIOT/pull/10633/files#diff-569047d5a7bc79b2ac9e91264ac20eb0R851).

boards/nucleo-f767zi/include/periph_conf.h
crest42: yes. The Definition of the struct is wrapped in a guard block. As an alternative we could remove the guard from both, the definition and the implementation, but this would add a few byte unused memory to the binary.

cpu/stm32_common/include/periph_cpu_common.h
crest42: Works just fine. Thanks :)

boards/nucleo-f767zi/include/periph_conf.h
crest42: Done

boards/nucleo-f767zi/include/periph_conf.h
crest42: Implemented as suggested

aabadie:
It seems the `eth_conf_t` struct has problems with c++. I also noticed the attributes were not in the same order in the struct definition and in `periph_conf.h` and this is important with c++.
In `cpu/stm32_common/periph_cpu_common.h`, how about defining enums and struct the following way (the order is also fixed, I kept the same one use in periph_conf.h):
```c
/**
 * @brief   Ethernet configuration mode
 */
typedef enum {
    MII = 18,                       /**< Configuration for MII */
    RMII = 9,                       /**< Configuration for RMII */
    SMI = 2,                        /**< Configuration for SMI */
} eth_mode_t;

/**
 * @brief   Speed selection
 */
typedef enum {
    ETH_SPEED_10T_HD = 0x0000,
    ETH_SPEED_10T_FD = 0x0100,
    ETH_SPEED_100TX_HD = 0x2000,
    ETH_SPEED_100TX_FD = 0x2100,
} eth_speed_t;

/**
 * @brief   Ethernet Peripheral configuration
 */
typedef struct {
    eth_mode_t mode;                    /**< Select configuration mode */
    char mac[6];                        /**< Ethernet MAC address */
    eth_speed_t speed;                  /**< Speed selection */
    uint8_t dma;                        /**< Locical CMA Descriptor used for TX */
    uint8_t dma_chan;                   /**< DMA channel used for TX */
    char phy_addr;                      /**< PHY address */
    gpio_t pins[];                      /**< Pins to use. MII requires 18 pins,
                                             RMII 9 and SMI 9. Not all speeds are
                                             supported by all modes. */
} eth_conf_t;
```

There might be better alternatives but this one builds.

crest42:
Looks good to me. I implemented the changes as suggested by @aabadie and just make the naming a little bit more specific. (eth_speed_t -> stm32_eth_speed_t)

Should I also rename eth_conf_t to stm32_eth_conf_t?

aabadie:
> Should I also rename eth_conf_t to stm32_eth_conf_t?

I don't think it's needed (same for eth_speed_t, etc). @kaspar030, do you confirm ?

drivers/stm32_eth/stm32_eth.c
```diff
@@ -0,0 +1,157 @@
...
+    .isr = _isr,
+    .get = _get,
+    .set = _set,
+};
+
+void stm32_eth_netdev_setup(netdev_t *netdev)
+{
+    _netdev = netdev;
+    netdev->driver = &netdev_driver_stm32f4eth;
+}
```
aabadie: nit: missing new line here.

crest42:
Both are fine for me. I reverted the naming changes and add the missing newline. Let's just leave it like this until somebody may have a better idea.

crest42:
Rebased to current master to resolve merge conflict

crest42:
> Also I allowed my self to squash the commits to remove mentions to my GitHub login to prevent notifications to be sent each time the commit is pushed to github (I think @vincent-d was also affected ;)).

Damn, I wasn't aware this is a thing. Sorry for the spam ;)

crest42:
Everything is green. Let's merge this! Thanks a lot for all the helpful feedback and support to everyone :)

aabadie:
@kaspar030: ok to hit the green button ?

kaspar030:
Yes, thanks!

Issue #10634, opened 2018-12-18
pyterm: weird behaviour with long lines.


jcarrano:
#### Description

Pyterm fails to handle long input lines (more than 64 characters,)

#### Steps to reproduce the issue

Go to `tests/shell`. I used a samr21:

```
$ cd tests/shell
$ BOARD=samr21-xpro make flash term
```

In the terminal, type a long line. No output will show up. Then type something else (say, `help`) a gardbage output follows:

```
$ BOARD=samr21-xpro make term
/home/jcarrano/source/vanillaRIOT/dist/tools/pyterm/pyterm -p "/dev/ttyACM0" -b "115200"
/home/jcarrano/source/vanillaRIOT/dist/tools/pyterm/pyterm:58: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  logging.getLogger("").warn("Twisted not available, please install "
Twisted not available, please install it if you want to use pyterm's JSON capabilities
/home/jcarrano/source/vanillaRIOT/dist/tools/pyterm/pyterm:490: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
  self.config = configparser.SafeConfigParser()
2018-12-18 18:38:11,397 - INFO # Connect to serial port /dev/ttyACM0
Welcome to pyterm!
Type '/exit' to exit.
2018-12-18 18:38:20,745 - INFO # main(): This is RIOT! (Version: 2018.07-devel-885-g77b91-luxemburg-optparse)
2018-12-18 18:38:20,746 - INFO # test_shell.
> sdfwdjewoijoiejdioewdjeoiwdjiowedjeoiwjdewoidjewoidjweoijdioewjdoweidjwoeidjweoijdwoeidjweoidjiwoed
help
2018-12-18 18:38:32,852 - INFO #  sdfwdjewoijoiejdioewdjeoiwdjiowedjeoiwjdewoidjewoidjweoijdioewjdM.X��,l��G�Y% �{�z�'���d�Bhelp
2018-12-18 18:38:32,861 - INFO # shell: command not found: sdfwdjewoijoiejdioewdjeoiwdjiowedjeoiwjdewoidjewoidjweoijdioewjd
> 
```

Notice the garbage characters.

#### Expected results

Using another term (I use miniterm.py) gives the expected result:

```
$ miniterm.py --raw /dev/ttyACM0 115200
[...]

PR #10635, opened 2018-12-18
 sys/shell: correctly detect and handle long lines

jcarrano:
### Contribution description

The numeric value for EOF is (-1). This caused the shell to return the same code when EOF was encountered and when the line lenght was exceeded. Additionally, if the line length is exceeded, the correct behaviour is to consume the remaining characters until the end of the line, to prevent the following line from containing (potentially dangerous) garbage.

### Testing procedure

The first commit adds a test to `tests/shell`. It adds a command to retrieve buffer size and then forces an extremely long line. This is failing right now because of a bug in the shell.

Murdock should be checking this, but you can do it locally by checkout both commits and running

```
BOARD=some_board make -C tests/shell flash test
```
In each one. The first one fails and the second succeeds.

I can switch the order of the commits if keeping a working history is important.

### Issues/PRs references

Testing was made specially difficult because of #10634 .
Part of  #12105

### DIFF
--- a/sys/shell/shell.c
+++ b/sys/shell/shell.c
@@ -29,6 +29,8 @@
 #include <stdio.h>
 #include <stdint.h>
 #include <stdlib.h>
+#include <stdbool.h>
+#include <assert.h>
 #include "shell.h"
 #include "shell_commands.h"
 
@@ -51,6 +53,12 @@ static void flush_if_needed(void)
     fflush(stdout);
 #endif
 }
+/**
+ * Code indicating that the line buffer size was exceeded.
+ *
+ * This definition ensures there's no collision with EOF.
+ */
+#define READLINE_TOOLONG (-EOF + 1)
 
 static shell_command_handler_t find_handler(const shell_command_t *command_list, char *command)
 {
@@ -227,14 +235,36 @@ static void handle_input_line(const shell_command_t *command_list, char *line)
     }
 }
 
+/**
+ * Read a single line from standard input into a buffer.
+ *
+ * In addition to copying characters, this routine echoes the line back to
+ * stdout and also supports primitive line editing.
+ *
+ * If the input line is too long, the input will still be consumed until the end
+ * to prevent the next line from containing garbage.
+ *
+ * @param   buf     Buffer where the input will be placed.
+ * @param   size    Size of the buffer. The maximum line length will be one less
+ *                  than size, to accommodate for the null terminator.
+ *                  The minimum buffer size is 1.
+ *
+ * @return  length of the read line, excluding the terminator, if reading was
+ *          successful.
+ * @return  EOF, if the if the end of the input stream is reached.
+ * @return  -READLINE_TOOLONG if the buffer size was exceeded
+ */
 static int readline(char *buf, size_t size)
 {
-    char *line_buf_ptr = buf;
+    int curr_pos = 0;
+    bool length_exceeded = false;
+
+    assert((size_t)size > 0);
 
     while (1) {
-        if ((line_buf_ptr - buf) >= ((int) size) - 1) {
-            return -1;
-        }
+        /* At the start of the loop, cur_pos should point inside of
+         * buf. This ensures the terminator can always fit. */
+        assert((size_t)curr_pos < size);
 
         int c = getchar();
         if (c < 0) {
@@ -246,23 +276,27 @@ static int readline(char *buf, size_t size)
         /* DOS newlines are handled like hitting enter twice, but empty lines are ignored. */
         /* Ctrl-C cancels the current line. */
         if (c == '\r' || c == '\n' || c == ETX) {
-            *line_buf_ptr = '\0';
+            buf[curr_pos] = '\0';
 #ifndef SHELL_NO_ECHO
             _putchar('\r');
             _putchar('\n');
 #endif
 
-            /* return 1 if line is empty, 0 otherwise */
-            return c == ETX || line_buf_ptr == buf;
+            return (length_exceeded)? -READLINE_TOOLONG : curr_pos;
         }
+
         /* QEMU uses 0x7f (DEL) as backspace, while 0x08 (BS) is for most terminals */
-        else if (c == 0x08 || c == 0x7f) {
-            if (line_buf_ptr == buf) {
+        if (c == 0x08 || c == 0x7f) {
+            if (curr_pos == 0) {
                 /* The line is empty. */
                 continue;
             }
 
-            *--line_buf_ptr = '\0';
+            /* after we dropped characters do not edit the line, yet keep the
+             * visual effects */
+            if (!length_exceeded) {
+                buf[--curr_pos] = '\0';
+            }
             /* white-tape the character */
 #ifndef SHELL_NO_ECHO
             _putchar('\b');
@@ -271,7 +305,13 @@ static int readline(char *buf, size_t size)
 #endif
         }
         else {
-            *line_buf_ptr++ = c;
+            /* Always consume characters, but do not not always store them */
+            if ((size_t)curr_pos < size - 1) {
+                buf[curr_pos++] = c;
+            }
+            else {
+                length_exceeded = true;
+            }
 #ifndef SHELL_NO_ECHO
             _putchar(c);
 #endif
@@ -290,6 +330,8 @@ static inline void print_prompt(void)
     flush_if_needed();
 }
 
+#define TOOLONG_MESSAGE "shell: maximum line length exceeded"
+
 void shell_run(const shell_command_t *shell_commands, char *line_buf, int len)
 {
     print_prompt();
@@ -297,14 +339,22 @@ void shell_run(const shell_command_t *shell_commands, char *line_buf, int len)
     while (1) {
         int res = readline(line_buf, len);
 
-        if (res == EOF) {
-            break;
-        }
-
-        if (!res) {
-            handle_input_line(shell_commands, line_buf);
+        switch (res) {
+            case EOF:
+                goto shell_run_exit;
+            case -READLINE_TOOLONG:
+                puts(TOOLONG_MESSAGE);
+                break;
+            case 0:
+                break;
+            default:
+                handle_input_line(shell_commands, line_buf);
+                break;
         }
 
         print_prompt();
     }
+
+shell_run_exit:
+    return;
 }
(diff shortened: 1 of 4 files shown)


sys/shell/shell.c
```diff
@@ -217,14 +225,34 @@ static void handle_input_line(const shell_command_t *command_list, char *line)
...
+ *
+ * If the input line is too long, the input will still be consumed until the end
+ * to prevent the next line from containing garbage.
+ *
+ * @param   buf     Buffer where the input will be placed.
+ * @param   size    Size of the buffer. The maximum line length will be one less
+ *                  than size, to accommodate for the null terminator.
+ *                  The minimum buffer size is 1.
+ *
+ * @return  READLINE_HASTEXT if the line was not empty
```
kaspar030: I don't think we should do some non-standard error reporting here. -1 on EOF, 0 ... 2 for other errors? EOF defined in a different header, the rest in a local enum? That's at least confusing.

How about we change the signature of the function to "size_t".
Empty line returns "0". Line with text returns it's length.
For the error cases, we define an enum, or re-use something from errno. ENOBUFS, EBADFD  come to mind.

sys/shell/shell.c
```diff
@@ -241,17 +269,22 @@ static int readline(char *buf, size_t size)
             _putchar('\n');
 #endif
 
-            /* return 1 if line is empty, 0 otherwise */
-            return line_buf_ptr == buf;
+            return (line_buf_ptr == buf)? READLINE_EMPTY
```
kaspar030: While I love the ternary operator, I find two in one line too much.

sys/shell/shell.c
```diff
@@ -260,9 +293,16 @@ static int readline(char *buf, size_t size)
 #endif
         }
         else {
-            *line_buf_ptr++ = c;
+            /* Always consume characters, but do not not always store them */
+            if (line_buf_ptr < buf + size - 1) {
```
kaspar030: Would you mind defining a "buf_end" ptr? ```char *buf_end = buf + size```, and use it here?

sys/shell/shell.c
```diff
@@ -280,21 +320,31 @@ static inline void print_prompt(void)
 #endif
 }
 
+static const char *TOOLONG_MESSAGE = "shell: maximum line length exceeded";
```
kaspar030: It's fine to use a define for a static string literal.

sys/shell/shell.c
```diff
@@ -280,21 +320,31 @@ static inline void print_prompt(void)
...
 
-        if (res == EOF) {
-            break;
-        }
-
-        if (!res) {
-            handle_input_line(shell_commands, line_buf);
+        switch (res) {
+            case EOF:
+                goto shell_run_exit;
```
kaspar030: this is probably a style question, but I think goto is controversal enough to just ```return``` here.

jcarrano: I'd rather have the procedure return at only one point. Goto is controversial for those who do not know how to use it correctly:  forwards and outwards jumps are OK.

jcarrano:
I rebased (due to #10630 being merged) and squashed the fixup.

jcarrano:
I would be a good idea to merge this considering people are wasting time in fixing the same bugs all over again (#11054 )

jcarrano:
I rebased on top of (the rebased) rawterm. Note that rawterm is not needed for this fix, but it makes testing easier.

tests/shell/Makefile
```diff
@@ -20,4 +20,7 @@ TEST_ON_CI_WHITELIST += all
 
 APP_SHELL_FMT ?= NONE
 
+# needed to correctly test ling lines
```
smlng: /ling/long/ ?

tests/shell/main.c
```diff
@@ -56,7 +56,17 @@ static int print_echo(int argc, char **argv)
...
+{
+    (void) argc;
+    (void) argv;
+    printf("%d\n", SHELL_DEFAULT_BUFSIZE);
+
+    return 0;
+}
+
 static const shell_command_t shell_commands[] = {
+    { "bufsize", "Get the shell's buffer size", print_shell_bufsize },
```
smlng: maybe plain and simple: "Get shell buffer size"?

sys/shell/shell.c
smlng: there are (better) places where `goto` has its (controversial) benefits - here it looks rather cosmetic (to me) and I would go for a simple `return;` as @kaspar030 suggested.

sys/shell/shell.c
```diff
@@ -286,21 +326,31 @@ static inline void print_prompt(void)
...
     while (1) {
         int res = readline(line_buf, len);
 
-        if (res == EOF) {
-            break;
-        }
-
-        if (!res) {
-            handle_input_line(shell_commands, line_buf);
+        switch (res) {
```
smlng: as there is no other use of `res` could be replace with `switch (readline(line_buf, len)) {`

sys/shell/shell.c
```diff
@@ -286,21 +326,31 @@ static inline void print_prompt(void)
...
+                break;
+            default:
+                handle_input_line(shell_commands, line_buf);
+                break;
         }
 
         print_prompt();
     }
+
+shell_run_exit:
```
smlng: removing `goto` make these 2 lines obsolete

tests/shell/tests/01-run.py
```diff
@@ -50,7 +50,18 @@ def check_cmd(child, cmd, expected):
...
-    child.expect('test_shell.')
+    child.expect('test_shell.\r')
+
+    child.sendline('bufsize')
+    child.expect('([0-9]+)')
+
+    bufsize = int(child.match[1])
+
+    # check a long line
+    child.sendline("_"*bufsize + "whatever")
```
smlng: maybe use `foobar` or `42` instead 🤓

sys/shell/shell.c
jcarrano: I'm _not_ going to `return` from a inside a `case` inside a `while`.

sys/shell/shell.c
jcarrano: Se comment above. I refuse to `return` in the middle of a function inside 2 nesting levels.

jcarrano:
Ouch! I discovered I broke control-c handling when I rebased this and also that testing via expect would miss that bug. I'm working on fixing it.

jcarrano:
Sorry for so many force-pushing- pexpect and the shell are conspiring against me.

sys/shell/shell.c
```diff
@@ -246,23 +276,32 @@ static int readline(char *buf, size_t size)
         /* DOS newlines are handled like hitting enter twice, but empty lines are ignored. */
         /* Ctrl-C cancels the current line. */
         if (c == '\r' || c == '\n' || c == ETX) {
-            *line_buf_ptr = '\0';
+            if (c == ETX) {
+                curr_pos = 0;
+                length_exceeded = 0;
```
smlng: /0/false/

sys/shell/shell.c
```diff
@@ -227,14 +235,36 @@ static void handle_input_line(const shell_command_t *command_list, char *line)
...
+ *
+ * @return  length of the read line, excluding the terminator, if reading was
+ *          successful.
+ * @return  EOF, if the if the end of the input stream is reached.
+ * @return  -READLINE_TOOLONG if the buffer size was exceeded
+ */
 static int readline(char *buf, size_t size)
 {
-    char *line_buf_ptr = buf;
+    int curr_pos = 0;
```
smlng: maybe use `unsigned` here? 

Also I don't like this half-abbreviated variable name, if you want it short why not simply use `pos` or `bufpos` to indicate the what and where.

sys/shell/shell.c
```diff
@@ -246,23 +276,32 @@ static int readline(char *buf, size_t size)
...
+        if (c == 0x08 || c == 0x7f) {
+            if (curr_pos == 0) {
                 /* The line is empty. */
                 continue;
             }
 
-            *--line_buf_ptr = '\0';
+            /* after we dropped characters do not edit the line, yet keep the
+             * visual effects */
+            if (!length_exceeded) {
```
smlng: could add `&& (curr_pos > 0)` to be safe.

sys/shell/shell.c
```diff
@@ -271,7 +310,13 @@ static int readline(char *buf, size_t size)
...
         }
         else {
-            *line_buf_ptr++ = c;
+            /* Always consume characters, but do not not always store them */
+            if ((size_t)curr_pos < size - 1) {
+                buf[curr_pos++] = c;
+            }
+            else {
+                length_exceeded = true;
+            }
```
smlng: missing `return` with cases for `length_exceeded` or not?

jcarrano:
@smlng much of this code will disappear with  #12099

sys/shell/shell.c
jcarrano: this flag will be gone in  #12099

sys/shell/shell.c
jcarrano: curr_pos can never go negative.

sys/shell/shell.c
jcarrano: No, the function only returns on enter or ctrl-c or EOF. Otherwise it makes it very hard to write robust scripts that interact with the shell as the remaining characters after the length is exceeded would go on to form another line.

sys/shell/shell.c
jcarrano: The current position is used as return value, so it I use an unsigned int there I need to do an explicit cast in line 290.

cladmi: The file is already using abbreviated names `buf`, `line_buf_ptr`. This does not change the style.

jcarrano:
@smlng I'm doing what I can trying to repair a code that is of the lowest standard. At the end I decided it needs more or complete rewrite (thus my "refactor" PRs) but first it is necessary to fix the bugs in the old code.

There is no point trying to make this perfect or even good, because I'm building on bad foundations.

sys/shell/shell.c
smlng: ah okay

sys/shell/shell.c
smlng: fine, leave as is

sys/shell/shell.c
smlng: agreed

sys/shell/shell.c
smlng: that might be, but the flag is introduces in this PR as `bool` and below it is set to `true` so why not use the (more) appropriate `false` here?

HendrikVE:
Can be closed due to #13195

smlng:
closing as needs rebase and is superseded by #13195

Issue #9191, opened 2018-05-25
Logic errors in nanocoap_get()


kb2ma:
#### Description
Two issues in nanocoap_get():

1. miscounts retries for a confirmable message
1. includes unnecessary gotos

The easiest way to reproduce the retry problem is to disable the server you're testing with. You should see nanocoap_get() attempt only three retries, not the expected four.

The retry loop is problematic -- it counts up, but stops one retry short. The loop counts up, but also includes a test for remaining tries equal to zero. Not gonna happen. :-) Also, the gotos presently in the while loop are unnecessary; a break will suffice.

smlng:
someone working on this, otherwise I could?

kb2ma:
Go for it, @smlng! I'd be happy to review.

PR #10640, opened 2018-12-19
tests: add nanocoap CLI app for client and server testing

kb2ma:
### Contribution description
Adds a CLI test app for nanocoap client and server testing, like the gnrc_udp test. As a CLI, it allows interactive or scripted testing. It also allows us to override/augment nanocoap as needed. For example, I have coded up a fix for #9191, and need an app to test it. So, in a follow-on PR I plan to extend the server side of the test app to ignore a configurable number of retries for a confirmable request.

### Testing procedure
See [README.md](https://github.com/kb2ma/RIOT/tree/tests/nanocoap_infrastructure/tests/coap_nanocoap) for the top level commands. You can run separate instances interactively as the client and server endpoints using the tapsetup tool.

### Issues/PRs references
Depends on #10642.


smlng:
> This PR also includes a fix for a simple include issue uncovered by writing the test app.

can you please separate the fix out into its own PR, I guess I could merge that in a second. But this one with the test might take a bit longer

kb2ma:
> can you please separate the fix out into its own PR

Done! Thanks for reviewing, @smlng.

smlng:
please rebase as #10692 is merged

tests/coap_nanocoap/request_handlers.c
```diff
@@ -0,0 +1,70 @@
...
+        break;
+    case COAP_PUT:
+    case COAP_POST:
+    {
+        /* convert the payload to an integer and update the internal value */
+        char payload[16] = { 0 };
+        memcpy(payload, (char*)pkt->payload, pkt->payload_len);
+        internal_value = strtol(payload, NULL, 10);
+        code = COAP_CODE_CHANGED;
+    }
```
smlng: This should have a `default:` setting `code` to `4.00` or `4.05

tests/coap_nanocoap/request_handlers.c
```diff
@@ -0,0 +1,70 @@
...
+        /* write the response buffer with the internal value */
+        p += fmt_u32_dec(rsp, internal_value);
+        code = COAP_CODE_205;
+        break;
+    case COAP_PUT:
+    case COAP_POST:
+    {
+        /* convert the payload to an integer and update the internal value */
+        char payload[16] = { 0 };
+        memcpy(payload, (char*)pkt->payload, pkt->payload_len);
```
smlng: maybe first check that `pkt->payload_len < 16`, what if I send something large with a different client?

kb2ma:
Rebased.

> I would suggest a different name, something like `nanocoap_cli` or `shell_nanocoap`

Do you mean for the directory under tests, which now is `coap_nanocoap`, or for the test definitions, which now are in `nanotest_(client|server).c`. I'm guessing the latter, so I would choose `nanocoap_cli_(client|server).c` for consistency with the `gcoap_cli.c` example.

smlng:
no, actually I meant the directory and with that test name, the `.c` files are fine ...

kb2ma:
> I would suggest a different name, something like `nanocoap_cli` or `shell_nanocoap`.

Renamed to `nanocoap_cli` in 5909aff. 

> I like this shell-based test app, I would love to script it

I was thinking the [same thing](https://github.com/kb2ma/riot-coap-pytest/commits/master). :-)

tests/coap_nanocoap/README.md
```diff
@@ -0,0 +1,18 @@
...
+`/value`<br>
+read/write an unsigned 8-bit integer
+
+`/.well-known/core`<br>
+Read the list of resources. Expects a block2 based request or else returns the
+first 16 bytes of the list.
+
+## *client*
+Provides get/put/post to an address, with an optional payload. Enter the command
+with no other text (just *client*) to see command syntax. Options are similar to
```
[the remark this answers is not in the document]
kb2ma: Removed reference to gcoap example in e90e665. 

Must be confirmable because that is a requirement of the `nanocoap_request()` function. It should be a simple change to allow for non-confirmable requests. I will add this task as a follow-up to my upcoming PR for #9191 because that issue fixes some of the handling around retries. After that, we can add a parameter to this test to specify confirmable or non-confirmable.

kb2ma:
Added b0b2c89 to display inet6 addresses configured for the node. I took the minimum amount of code from `sc_gnrc_netif.c` When using a two-port tap testing setup, I use this command as a simple way to determine the address of the server. Let me know if there is a better way.

tests/coap_nanocoap/request_handlers.c
kb2ma: Defaults to 4.05 in acbeb8c.

tests/coap_nanocoap/request_handlers.c
kb2ma: Good catch; fixed in acbeb8c.

kb2ma:
I have addressed all comments. Let me know of any other concerns.

smlng:
I tested and it works but when I try to get `/.well-known/core` from another node I only get the first few bytes:

```
client get  fe80::d1c1:6d42:ab00:1336 5683 /.well-known/core
2019-01-08 10:14:07,913 - INFO # nanocli: sending msg ID 1, 23 bytes
2019-01-08 10:14:07,925 - INFO # nanocli: response Success, code 2.05, 16 bytes
2019-01-08 10:14:07,927 - INFO # </.well-known/co
```

I use 2 PhyNodes (pba-d-01-kw2x) could be related to that...

smlng:
otherwise this looks good, I'll check with Linux native

smlng:
nope problem is also with native, sniffer shows only one package in response which is a BLOCK but no further:

```
Capturing on 'tapbr0'
    1 0.000000000 fe80::34ae:a5ff:fea2:ab69 ? fe80::88e:cff:fe12:1dc3 CoAP 85 CON, MID:1, GET, TKN:da ec, /.well-known/core
    2 0.000438913 fe80::88e:cff:fe12:1dc3 ? fe80::34ae:a5ff:fea2:ab69 CoAP 89 ACK, MID:1, 2.05 Content, TKN:da ec, Block #0, /.well-known/core
```

smlng:
though I guess this is a nanocoap problem with handling BLOCK transfer?

kb2ma:
The root problem is a bug that a (nano|g)coap client does not recognize when they receive a message with a critical option that they do not understand. RFC 7252, sec. 5.4.1 says that a client in this case must silently ignore the ACK. So, a fix would be for nanocoap_request() to return some error value. I will file a bug. There is an equivalent issue with server request handling.

On the brighter side, I am just about to post an issue and a PR that starts to add client side Block handling. See the [wiki page]([issue](https://github.com/kb2ma/RIOT/wiki/Completing-CoAP-block)).

I don't think this bug should block this PR though. At least we have a way convenient way to test a nanocoap client now.

smlng:
@kb2ma agreed, the issue is unrelated to this PR.

Hence, please rebase and squash as need so we get this merged.

kb2ma:
Rebased and squashed.

PR #10642, opened 2018-12-20
net/nanocoap: fix includes for nanocoap sock

kb2ma:
### Contribution description
This PR is a simple fix for a missing includes in nanocoap sock -- the header and implementation files. `nanocoap_sock.h` neglected to include `nanocoap.h`, which clearly is necessary. This issue was masked because the only use of `nanocoap_sock.h` was in the nanocoap_server example, and it explicitly included `nanocoap.h`. which is unnecessary.

This PR also fixes nanocoap's `sock.c` to include `nanocoap_sock.h` as well, as it implements the declarations in the header.

### Testing procedure
It should be enough to compile the nanocoap_server example. You could also try to build #10640, which now depends on this PR and includes `nanocoap_sock.h`.

### Issues/PRs references
none

### DIFF
--- a/examples/nanocoap_server/main.c
+++ b/examples/nanocoap_server/main.c
@@ -19,9 +19,7 @@
 
 #include <stdio.h>
 
-#include "net/nanocoap.h"
 #include "net/nanocoap_sock.h"
-
 #include "xtimer.h"
 
 #define COAP_INBUF_SIZE (256U)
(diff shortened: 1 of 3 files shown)


kb2ma:
Sorry for the force push, but just noticed that nanocoa's `sock.c` was including `nanocoap.h` rather than `nanocoap_sock.h`. Hopefully no one had started to look at this yet.

PR #10643, opened 2018-12-20
boards/esp32: UART configuration fix

gschorcht:
### Contribution description

Fixes the UART pin configuration for UART_DEV(0).

### Testing procedure

Since this configuration is defined for documentation purposes only, the change has no impact on the code. There is nothing to test.


gschorcht:
@MrKevinWeiss and this configuration fix too. This shouldn't be a big thing since ithe configuration for UART_DEV(0) is fix and is only defined for doxygen.

gschorcht:
Thanks

PR #10646, opened 2018-12-20
boards/nucleo-f072rb: configure I2C using common stm32 configuration

aabadie:
### Contribution description

While testing #10610 on a nucleo-f072, I realized that the i2c configuration was missing. This PR is adding one for this board.

To make it work, since some macro used by RIOT were missing, I had to update the CMSIS. This is also done in this PR.

### Testing procedure

Build and flash any application that requires I2C, it should work. And it will work even better when #10610 will be merged.

### Issues/PRs references

I needed it for testing #10610


leandrolanzieri:
I tested on b-l072z-lrwan1 and can confirm that it is working, so vendor headers are ok.

MrKevinWeiss:
Sorry, my bad.  We don't have this board to test currently.

Since it currently passes all murdock build tests I think that would be sufficient for me.  Just to be safe I will give it some time if anyone else wants to test the board.

aabadie:
Just in case, this PR is now based on #10690 and it is still usable: the I2C available on the Arduino pinout works.

MrKevinWeiss:
Heh, I was expecting that!  Any chance you can tell me where you got the updated vendor files from?  I have tried to verify but I can only find the old files (around 5000 lines not 10000 lines).

aabadie:
> Any chance you can tell me where you got the updated vendor files from?

I took it from the STM32 CubeMX application: generate a project and the files can be found somewhere in the project workspace.

MrKevinWeiss:
Ok, I double checked (it looks close enough).  I am happy with the vendor header update and we can just get the other PR in before finishing this one!

aabadie:
Interesting to see Murdock failing without any apparent error...

aabadie:
Thank you for reviewing @MrKevinWeiss !

PR #10647, opened 2018-12-20
boards/nucleo-l433rc: add I2C configuration

aabadie:
### Contribution description

This PR configure one I2C peripheral for the nucleo-l433rc. I needed it for testing #10610 on L4 on this board and was too lazy for looking for an other one already configured.

### Testing procedure

Build and flash any application that requires I2C, it should work. The I2C is configured to use SDA/SCL pins from the Arduino pinout.

### Issues/PRs references

I needed it for testing #10610

### DIFF
--- a/boards/nucleo-l433rc/Makefile.features
+++ b/boards/nucleo-l433rc/Makefile.features
@@ -1,4 +1,5 @@
 # Put defined MCU peripherals here (in alphabetical order)
+FEATURES_PROVIDED += periph_i2c
 FEATURES_PROVIDED += periph_lpuart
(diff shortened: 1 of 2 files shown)


tcschmidt:
> It appears we cannot test. I am still confident everything is OK. I will give an untested ACK!
@MrKevinWeiss can you assign labels according to the (partial) review you performed?

aabadie:
> It appears we cannot test. I am still confident everything is OK

I could test my own PR and confirm it is still working after the rebase of this morning. Let's merge this. Thank you for reviewing @MrKevinWeiss !

PR #10648, opened 2018-12-20
Improved docs on module name uniqueness

enzingerm:
I ran into `undefined reference to` errors and it was hard for me to track these down to two modules overwriting each others files. Maybe I can save someone else some time with this additional hint.

### DIFF
--- a/doc/doxygen/src/creating-modules.md
+++ b/doc/doxygen/src/creating-modules.md
@@ -37,7 +37,8 @@ application's Makefile.
 ### Pitfalls ###
 
 The `MODULE` name should be unique or build breaks as modules overwrite the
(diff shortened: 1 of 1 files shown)


cladmi:
I agree to indeed improve the comment. I wrote it in the first place when it was not respected everywhere and did not thought about writing some problems.

It means the documentation is read, thank you for improving it :)

Please prepend your commit message with `doc: ` and maybe describe the change more than "improve". Something like `doc: example of problems of non uniq module name`.

enzingerm:
Thanks for the appreciation, I followed your request and changed the commit message.

cladmi:
Thanks for the contribution :)

PR #10649, opened 2018-12-21
ifconfig: Add manual CCA operation

gebart:
### Contribution description

Adds a new `cca` subcommand to the `ifconfig` shell command for performing a clear channel assessment (CCA) operation.

This is mainly useful for debugging and for giving an indication of the noise level in the environment.

From `ifconfig help`
```
usage: ifconfig <if_id> cca
       perform a clear channel assessment and print the result
```

### Testing procedure

Build and flash one of the network examples which have the ifconfig shell command (gnrc_networking, gnrc_border_router)
in the shell, type
ifconfig x cca
where x is the interface number which can be found by running ifconfig with no arguments.

Example output from gnrc_border_router:
```
> ifconfig 7 cca
ifconfig 7 cca
  7: CCA 1, ED=-83
> ifconfig 6 cca
ifconfig 6 cca
  6: CCA not supported
> ifconfig 3 cca
ifconfig 3 cca
error: invalid interface given
```

### DIFF
--- a/sys/shell/commands/sc_gnrc_netif.c
+++ b/sys/shell/commands/sc_gnrc_netif.c
@@ -980,6 +980,12 @@ static void _l2filter_usage(const char *cmd)
 }
 #endif
 
+static void _cca_usage(const char *cmd)
+{
+    printf("usage: %s <if_id> cca\n", cmd);
+    puts("       perform a clear channel assessment and print the result\n");
+}
+
 static void _usage(char *cmd)
 {
     printf("usage: %s\n", cmd);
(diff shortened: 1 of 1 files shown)


sys/shell/commands/sc_gnrc_netif.c
```diff
@@ -980,13 +980,20 @@ static void _l2filter_usage(const char *cmd)
 }
 #endif
 
+static void _cca_usage(const char *cmd)
+{
+    printf("usage: %s <if_id> cca\n", cmd);
+    puts("       perform a clear channel assessment and print the result\n");
```
cgundogan: The line feed at the end is IMO unnecessary. When calling `ifconfig help` it produces an empty line between the `cca` and `stat reset` usages. I have no strong opinion about this, but for my taste I would prefer no line feed at all, since the usage text is already indented.

PR #10570, opened 2018-12-07
dist/tools/edbg: update to lastest version

dylad:
### Contribution description
This PR updates edbg tool to the lastest available version.
This edbg version introduces support for SAML10/SAML11 MCUs. I'm currently porting them to RIOT so I need to update edbg to flash them first.

### Testing procedure
rebuild edbg using the lastest version, check if edbg still flash your SAM0 MCUs.

### Issues/PRs references
None.

Issue #10523, opened 2018-11-30
saml21 system time vs rtc


fedepell:
#### Description
System time on my saml21-xpro seems to be quite innacurate, both compared to the RTC and to an external source. In about 1 hour I can see a drift of about 15 seconds which seems quite a lot. 
Is this to be expected or could something be lying behind that could be analyzed better? (I can try to give a deeper look if this is not expected)

#### Steps to reproduce the issue
I did a simple application that has a loop that every minute reads the system time, reads the RTC time and prints them out. Something more or less like:

```
    time_t rawtime;
    struct tm *info;
    struct tm curtime;

    char buffer[80];
    char buffer2[80];
    int contatore=0;

    while (1) {
      time( &rawtime );
      info = localtime( &rawtime );      
      strftime(buffer,80,"%x %H:%M:%S", info);

      rtc_get_time(&curtime);
      strftime(buffer2,80,"%x %H:%M:%S", &curtime);
      
      printf("[CNT: %8d][SYS: %s][RTC: %s]\r\n",contatore,buffer,buffer2);

      contatore++;
      xtimer_sleep(60);
    }
```
RTC is initialized at some known value at the beginning. 

#### Expected results
I would expect them to run roughly at the same speed.

#### Actual results
I see a drift of about 15 seconds in an hour which sounds quite something to me. Compared to an external source (my PC) the time read from the RTC is the correct one, as that one is aligned (at least after a couple of hours and to a precision of 1 sec) to the PC.
What I see running the test is (I omit some iterations) for example:

```
RTC Start time is 18 November 2018 at 08:00:00
[CNT:        0][SYS: 01/01/70 00:00:00][RTC: 11/18/18 08:00:00]
[CNT:        1][SYS: 01/01/70 00:01:00][RTC: 11/18/18 08:01:00]
[CNT:        2][SYS: 01/01/70 00:02:00][RTC: 11/18/18 08:02:00]
[CNT:        3][SYS: 01/01/70 00:03:00][RTC: 11/18/18 08:03:00]
[CNT:        4][SYS: 01/01/70 00:04:00][RTC: 11/18/18 08:04:00]
[CNT:        5][SYS: 01/01/70 00:05:00][RTC: 11/18/18 08:05:01]
[CNT:        6][SYS: 01/01/70 00:06:00][RTC: 11/18/18 08:06:01]
[CNT:        7][SYS: 01/01/70 00:07:00][RTC: 11/18/18 08:07:01]
[...]

PR #10653, opened 2018-12-24
cpu/saml1x: add support for SAML10 and SAML11 MCUs (Cortex-M23)

dylad:
### Contribution description

This PR adds a very basic support for our first Cortex-M23 MCUs, SAML10 and SAML11 so people can start to play with it. These chips are the same except that SAML11 add security stuff (ARM TrustZone, non-secure+secure registers,...)

~~In the current state, only GPIOs, timer and UART work.~~

SAML10/SAML11 IP peripherals are really close to SAML21. But I don't know if these MCUs should re-use stuff within sam0_common since they are cortex-m23 based...So right now, I copied GPIOs, timer , uart SAML21 drivers and make the required changes to use them but this can be optimized. (now or later). In fact, we can re-use all sam0 drivers with these MCUs but it will take much more time to fully test it. We can stay as it now and optimize later or we can do it now but I was hoping to see this PR merged for the incoming release.

Regarding MPU support, I intentionally prevent from using this feature on Cortex-M23 because armv8 introduces several changes and I need more time to rework and test this feature. I (or someone with more MPU knowledge) will do it later.

### Testing procedure

run make BOARD=saml10-xpro -C [test_you_want]
or  make BOARD=saml11-xpro -C [test_you_want]

EDBG will be used to flash the board.

### Issues/PRs references
This PR is based on top of #10558
see also #10570 for EDBG update which supports these MCUs.

### DIFF
--- a/boards/common/saml1x/Makefile.dep
+++ b/boards/common/saml1x/Makefile.dep
@@ -1,0 +1,3 @@
+ifneq (,$(filter saul_default,$(USEMODULE)))
+  USEMODULE += saul_gpio
+endif
--- a/boards/common/saml1x/include/periph_conf.h
+++ b/boards/common/saml1x/include/periph_conf.h
@@ -1,0 +1,155 @@
+/*
+ * Copyright (C) 2019 Mesotic SAS
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     boards_common_saml1x
+ * @{
+ *
+ * @file
+ * @brief       Peripheral MCU configuration for the Microchip
+ *              SAML10 & SAML11 Xplained Pro board
+ *
+ * @author      Dylan Laduranty <dylan.laduranty@mesotic.com>
+ */
+
+#ifndef PERIPH_CONF_H
+#define PERIPH_CONF_H
+
+#include "periph_cpu.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief   GCLK reference speed
+ */
+#define CLOCK_CORECLOCK     (16000000U)
+
+/**
+ * @name    Timer peripheral configuration
+ * @{
+ */
+#define TIMER_NUMOF        (1U)
+#define TIMER_0_EN         1
+
+/* Timer 0 configuration */
+#define TIMER_0_DEV        TC0->COUNT32
+#define TIMER_0_CHANNELS   1
+#define TIMER_0_MAX_VALUE  (0xffffffff)
+#define TIMER_0_ISR        isr_tc0
+/** @} */
+
+/**
+ * @name    UART configuration
+ * @{
+ */
+static const uart_conf_t uart_config[] = {
+    {    /* Virtual COM Port */
+        .dev      = &SERCOM2->USART,
+        .rx_pin   = GPIO_PIN(PA,25),
+        .tx_pin   = GPIO_PIN(PA,24),
+        .mux      = GPIO_MUX_D,
+        .rx_pad   = UART_PAD_RX_3,
+        .tx_pad   = UART_PAD_TX_2,
+        .flags    = UART_FLAG_NONE,
+        .gclk_src = GCLK_PCHCTRL_GEN_GCLK0
+    }
+};
+
+/* interrupt function name mapping */
+#define UART_0_ISR          isr_sercom2_2
+
+#define UART_NUMOF          (sizeof(uart_config) / sizeof(uart_config[0]))
+/** @} */
+
+/**
+ * @name    SPI configuration
+ * @{
+ */
+static const spi_conf_t spi_config[] = {
+    {
+        .dev      = &(SERCOM0->SPI),
+        .miso_pin = GPIO_PIN(PA,  4),
+        .mosi_pin = GPIO_PIN(PA, 14),
+        .clk_pin  = GPIO_PIN(PA, 15),
+        .miso_mux = GPIO_MUX_D,
+        .mosi_mux = GPIO_MUX_D,
+        .clk_mux  = GPIO_MUX_D,
+        .miso_pad = SPI_PAD_MISO_0,
+        .mosi_pad = SPI_PAD_MOSI_2_SCK_3
+
+    }
+};
+
+#define SPI_NUMOF           (sizeof(spi_config) / sizeof(spi_config[0]))
+/** @} */
+
+/**
+ * @name    I2C configuration
+ * @{
+ */
+static const i2c_conf_t i2c_config[] = {
+    {
+        .dev      = &(SERCOM1->I2CM),
+        .speed    = I2C_SPEED_NORMAL,
+        .scl_pin  = GPIO_PIN(PA, 17),
+        .sda_pin  = GPIO_PIN(PA, 16),
+        .mux      = GPIO_MUX_C,
+        .gclk_src = GCLK_PCHCTRL_GEN_GCLK0,
+        .flags    = I2C_FLAG_NONE
+    }
+};
+
+#define I2C_NUMOF           (sizeof(i2c_config) / sizeof(i2c_config[0]))
+/** @} */
+
+/**
+ * @name    RTC configuration
+ * @{
+ */
+#define RTC_NUMOF           (1)
+#define EXTERNAL_OSC32_SOURCE                    1
+#define INTERNAL_OSC32_SOURCE                    0
+#define ULTRA_LOW_POWER_INTERNAL_OSC_SOURCE      0
+/** @} */
+
+/**
+ * @name    RTT configuration
+ * @{
+ */
+#define RTT_FREQUENCY       (32768U)
+#define RTT_MAX_VALUE       (0xffffffffU)
+#define RTT_NUMOF           (1)
+/** @} */
+
+/**
+ * @name ADC Configuration
+ * @{
+ */
+#define ADC_NUMOF                          (1U)
+
+/* ADC 0 Default values */
+#define ADC_0_CLK_SOURCE                   0 /* GCLK_GENERATOR_0 */
+#define ADC_0_PRESCALER                    ADC_CTRLB_PRESCALER_DIV256
+
+static const adc_conf_chan_t adc_channels[] = {
+    /* port, pin, muxpos */
+    {GPIO_PIN(PA, 10), ADC_INPUTCTRL_MUXPOS(ADC_INPUTCTRL_MUXPOS_AIN8)},
+};
+
+#define ADC_0_NEG_INPUT                    ADC_INPUTCTRL_MUXNEG(0x18u)
+#define ADC_0_REF_DEFAULT                  ADC_REFCTRL_REFSEL_INTVCC2
+/** @} */
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* PERIPH_CONF_H */
+/** @} */
--- a/boards/saml10-xpro/Makefile.features
+++ b/boards/saml10-xpro/Makefile.features
@@ -1,0 +1,1 @@
+include $(RIOTBOARD)/common/saml1x/Makefile.features
--- a/cpu/sam0_common/periph/gpio.c
+++ b/cpu/sam0_common/periph/gpio.c
@@ -37,10 +37,24 @@
 #define MODE_PINCFG_MASK            (0x06)
 
 #ifdef MODULE_PERIPH_GPIO_IRQ
+
 /**
  * @brief   Number of external interrupt lines
  */
+#ifdef CPU_SAML1X
+#define NUMOF_IRQS                  (8U)
+#else
 #define NUMOF_IRQS                  (16U)
+#endif
+
+/**
+ * @brief   External Interrupts Controller selection macros
+ */
+#ifdef CPU_FAM_SAML11
+#define _EIC EIC_SEC
+#else
+#define _EIC EIC
+#endif
 
 static gpio_isr_ctx_t gpio_config[NUMOF_IRQS];
 #endif /* MODULE_PERIPH_GPIO_IRQ */
@@ -181,26 +195,31 @@ int gpio_init_int(gpio_t pin, gpio_mode_t mode, gpio_flank_t flank,
     MCLK->APBAMASK.reg |= MCLK_APBAMASK_EIC;
     GCLK->PCHCTRL[EIC_GCLK_ID].reg = GCLK_PCHCTRL_CHEN | GCLK_PCHCTRL_GEN_GCLK0;
     /* disable the EIC module*/
-    EIC->CTRLA.reg = 0;
-    while (EIC->SYNCBUSY.reg & EIC_SYNCBUSY_ENABLE) {}
+    _EIC->CTRLA.reg = 0;
+    while (_EIC->SYNCBUSY.reg & EIC_SYNCBUSY_ENABLE) {}
 #endif
     /* configure the active flank */
-    EIC->CONFIG[exti >> 3].reg &= ~(0xf << ((exti & 0x7) * 4));
-    EIC->CONFIG[exti >> 3].reg |=  (flank << ((exti & 0x7) * 4));
+    _EIC->CONFIG[exti >> 3].reg &= ~(0xf << ((exti & 0x7) * 4));
+    _EIC->CONFIG[exti >> 3].reg |=  (flank << ((exti & 0x7) * 4));
     /* enable the global EIC interrupt */
+#ifdef CPU_SAML1X
+    /* EXTI[4..7] are binded to EIC_OTHER_IRQn */
+    NVIC_EnableIRQ((exti > 3 )? EIC_OTHER_IRQn : (EIC_0_IRQn + exti));
+#else
     NVIC_EnableIRQ(EIC_IRQn);
+#endif
     /* clear interrupt flag and enable the interrupt line and line wakeup */
-    EIC->INTFLAG.reg = (1 << exti);
-    EIC->INTENSET.reg = (1 << exti);
+    _EIC->INTFLAG.reg = (1 << exti);
+    _EIC->INTENSET.reg = (1 << exti);
 #ifdef CPU_FAM_SAMD21
-    EIC->WAKEUP.reg |= (1 << exti);
+    _EIC->WAKEUP.reg |= (1 << exti);
     /* enable the EIC module*/
-    EIC->CTRL.reg = EIC_CTRL_ENABLE;
-    while (EIC->STATUS.reg & EIC_STATUS_SYNCBUSY) {}
+    _EIC->CTRL.reg = EIC_CTRL_ENABLE;
+    while (_EIC->STATUS.reg & EIC_STATUS_SYNCBUSY) {}
 #else /* CPU_FAM_SAML21 */
     /* enable the EIC module*/
-    EIC->CTRLA.reg = EIC_CTRLA_ENABLE;
-    while (EIC->SYNCBUSY.reg & EIC_SYNCBUSY_ENABLE) {}
+    _EIC->CTRLA.reg = EIC_CTRLA_ENABLE;
+    while (_EIC->SYNCBUSY.reg & EIC_SYNCBUSY_ENABLE) {}
 #endif
     return 0;
 }
@@ -211,7 +230,7 @@ void gpio_irq_enable(gpio_t pin)
     if (exti == -1) {
         return;
     }
-    EIC->INTENSET.reg = (1 << exti);
+    _EIC->INTENSET.reg = (1 << exti);
 }
 
 void gpio_irq_disable(gpio_t pin)
@@ -220,17 +239,45 @@ void gpio_irq_disable(gpio_t pin)
     if (exti == -1) {
         return;
     }
-    EIC->INTENCLR.reg = (1 << exti);
+    _EIC->INTENCLR.reg = (1 << exti);
 }
 
 void isr_eic(void)
 {
     for (unsigned i = 0; i < NUMOF_IRQS; i++) {
-        if (EIC->INTFLAG.reg & (1 << i)) {
-            EIC->INTFLAG.reg = (1 << i);
+        if (_EIC->INTFLAG.reg & (1 << i)) {
+            _EIC->INTFLAG.reg = (1 << i);
             gpio_config[i].cb(gpio_config[i].arg);
         }
     }
     cortexm_isr_end();
 }
+
+#ifdef CPU_SAML1X
+void isr_eic0(void)
+{
+    isr_eic();
+}
+
+void isr_eic1(void)
+{
+    isr_eic();
+}
+
+void isr_eic2(void)
+{
+    isr_eic();
+}
+
+void isr_eic3(void)
+{
+    isr_eic();
+}
+
+void isr_eic_other(void)
+{
+    isr_eic();
+}
+#endif /* CPU_SAML1X */
+
 #endif /* MODULE_PERIPH_GPIO_IRQ */
--- a/cpu/sam0_common/periph/i2c.c
+++ b/cpu/sam0_common/periph/i2c.c
@@ -44,7 +44,7 @@
 #define BUSSTATE_OWNER SERCOM_I2CM_STATUS_BUSSTATE(2)
 #define BUSSTATE_BUSY SERCOM_I2CM_STATUS_BUSSTATE(3)
 
-#if defined(CPU_FAM_SAML21) || defined(CPU_FAM_SAMR30)
+#if defined(CPU_SAML21) || defined(CPU_SAML1X)
 #define SERCOM_I2CM_CTRLA_MODE_I2C_MASTER SERCOM_I2CM_CTRLA_MODE(5)
 #endif
 
@@ -98,6 +98,10 @@ void i2c_init(i2c_t dev)
                   SERCOM0_GCLK_ID_SLOW : SERCOM5_GCLK_ID_SLOW)].reg =
     (GCLK_PCHCTRL_CHEN | i2c_config[dev].gclk_src  );
     while (GCLK->SYNCBUSY.bit.GENCTRL) {}
+#elif defined (CPU_SAML1X)
+    GCLK->PCHCTRL[SERCOM0_GCLK_ID_SLOW].reg = (GCLK_PCHCTRL_CHEN |
+                                              i2c_config[dev].gclk_src  );
+     while (GCLK->SYNCBUSY.bit.GENCTRL0) {}
 #else
     /* GCLK_SERCOMx_SLOW is shared for all sercom */
     GCLK->CLKCTRL.reg = (GCLK_CLKCTRL_CLKEN |
--- a/cpu/sam0_common/periph/uart.c
+++ b/cpu/sam0_common/periph/uart.c
@@ -99,7 +99,11 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
     if ((rx_cb) && (uart_config[uart].rx_pin != GPIO_UNDEF)) {
         uart_ctx[uart].rx_cb = rx_cb;
         uart_ctx[uart].arg = arg;
+#if defined (CPU_SAML1X)
+        NVIC_EnableIRQ(SERCOM0_2_IRQn + (sercom_id(dev(uart)) * 4));
+#else
         NVIC_EnableIRQ(SERCOM0_IRQn + sercom_id(dev(uart)));
+#endif
         dev(uart)->CTRLB.reg |= SERCOM_USART_CTRLB_RXEN;
         dev(uart)->INTENSET.reg |= SERCOM_USART_INTENSET_RXC;
         /* set wakeup receive from sleep if enabled */
--- a/pkg/qDSA/Makefile.include
+++ b/pkg/qDSA/Makefile.include
@@ -1,4 +1,4 @@
-ifneq (,$(filter cortex-m0%,$(CPU_ARCH)))
+ifneq (,$(filter cortex-m23 cortex-m0%,$(CPU_ARCH)))
   QDSA_IMPL ?= arm
 else
 ifneq (,$(filter atmega_common,$(USEMODULE)))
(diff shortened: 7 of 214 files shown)


aabadie:
Very nice! Thanks.

I'll review and test it in 2019 :)

boards/saml10-xpro/Makefile.features
```diff
@@ -0,0 +1,9 @@
+# Put defined MCU peripherals here (in alphabetical order)
+FEATURES_PROVIDED += periph_gpio periph_gpio_irq
```
aabadie: This is not needed and can be provided at CPU levels (several PRs were recently merged in that direction)

dylad: I'll remove it. This is a leftover because this PR was started some times before this change was merged.

dylad:
@aabadie Thanks for the review.

>  I would suggest to directly rename sam0_common to samx_common

In this case, I'll rework this PR but it'll take some time. 

> I can help if you want.

Since you're pretty busy with the release management I don't think you should afford such effort. I'll rework all SAM0 to support SAML1x MCUs.

dylad:
@aabadie I've forced-push the branch.
SAML10/SAML11 now use SAM0 driver. I was able to test ADC and I2C on SAML10.
This PR is ready for another round !

dylad:
> the gpio driver is broken with interrupts

On which board ? L10 or L11 ?
> I found the saml21 and saml1x peripheral implementations very close. I won't ask you to factorize in this PR though, but I see room for improvement here.

Yes, this should be factorize (later). I had not idea how to reuse SAML21 drivers without adding another dependency. IMHO, the best way to do is merge timer, rtc and rtt driver to sam0_common. I'll work on this later I swear.
> tests/periph_rtc and tests/periph_rtt doesn't work:

I'll fix.
> tests/periph_flashpage doesn't build, it needs to be adapted.

I probably miss this driver.
> gpio_params.h is missing but saul_gpio is included with the saul_default module

I'll fix.

boards/saml10-xpro/Makefile.features
```diff
@@ -0,0 +1,9 @@
+FEATURES_PROVIDED += periph_adc
```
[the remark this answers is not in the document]
dylad: Sure.

cpu/sam0_common/periph/uart.c
```diff
@@ -99,7 +99,11 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
     if ((rx_cb) && (uart_config[uart].rx_pin != GPIO_UNDEF)) {
         uart_ctx[uart].rx_cb = rx_cb;
         uart_ctx[uart].arg = arg;
+#if defined (CPU_FAM_SAML10) || defined (CPU_FAM_SAML11)
```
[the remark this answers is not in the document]
dylad: Good catch.

boards/common/saml1x/Makefile.dep
```diff
@@ -0,0 +1,3 @@
+ifneq (,$(filter saul_default,$(USEMODULE)))
+  USEMODULE += saul_gpio
```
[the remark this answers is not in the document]
dylad: I'll find the way.

dylad:
@aabadie I'm sorry I had to rebased the branch to the current master to include some fix to flashpage for SAM0 but It does work yet on SAML10/SAML11. I'll investigate further but I'm not an expert.
Good news are : 

- GPIO IRQ are working (I missed some commits from my private repo) 
- RTC/RTT are working
- SAUL is working
- minor comments have been addressed

I didn't squash, so you can have a look at the changes introduce by the new commits.

aabadie:
@dylad, I tested your last commit and it fixes the `tests/buttons` application and saul LED/button, good job!

Just in case, for rebase/squash ease, you may want to split the commit in 2, one for saul, one for periph gpio.

dylad:
@aabadie Thanks for the SAUL tips, it is now fixed.

> Also the tests/buttons doesn't work. Maybe the exti array needs an update (haven't checked the datasheet). Or there is something else with the peripheral driver ?

I forget something important for this ~~damned~~ SAML11 MCU in the GPIO driver, the EXTI array should be fine, I looked at the datasheet.
tests/buttons should now works for SAML11 (it was working but for SAML10 only).

dylad:
> Just in case, for rebase/squash ease, you may want to split the commit in 2, one for saul, one for periph gpio.

Sure.
I'll take another look tonight to see if I didn't miss something else. If everything is good I'll rebase/squash tomorrow.

aabadie:
Now that #10558, you need to rebase this PR.

I'll do a full pass with all test applications to verify saml11-xpro before merging this PR.

dylad:
> I'll do a full pass with all test applications to verify saml11-xpro before merging this PR.

Is there a way to easily disable flashpage driver for SAML1X (which depend on sam0_common) ?
I don't think I'll be able to fix this driver in time for the release.

aabadie:
> Is there a way to easily disable flashpage driver for SAML1X

I tested it and it was working, so there's no need to disable it.

dylad:
> I tested it and it was working, so there's no need to disable it.

Really ? I'll triple-check that tonight.

dylad:
Removed #10558 commits and rebased.

aabadie:
> I can confirm that driver_flashpage is broken right now for SAML11 only

This is strange, I tried `flash test` on a saml11-xpro and it succeeded. Are you manually testing ?

dylad:
> This is strange, I tried flash test on a saml11-xpro and it succeeded. Are you manually testing ?

Yes, the test_last command, ran from shell, fails.

aabadie:
Ok, I'm not with the board right now so I can't re-test, will do tomorrow. I checked the automatic test code and it calls also `test_last`, there should be no difference with what you do :confused:

dylad:
@aabadie I also tested `make flash test` and it failed on my hardware.
We will see tomorrow then !

dylad:
@aabadie we're both right in fact.
If you erase the whole memory (`edbg -t mchp_cm23 -e`), run `make flash test` on SAML11. It works.
If you re-run the same test right after, it will fail.

aabadie:
> If you erase the whole memory (edbg -t mchp_cm23 -e), run make flash test on SAML11. It works.
If you re-run the same test right after, it will fail.

That's it, I re-tested this morning and could reproduce this behaviour. So how do we proceed here, should we add the `-e` option to the flasher or enforce a page erase before write in the driver ? The latter would be my preferred solution.

Otherwise, I ran the full test suite on the saml11-xpro during the night and had a few tests failing:
- [tests/cb_mux_bench](tests/cb_mux_bench/test.failed)
- [tests/periph_flashpage](tests/periph_flashpage/test.failed)
- [tests/pkg_hacl](tests/pkg_hacl/test.failed)
- [tests/pkg_monocypher](tests/pkg_monocypher/test.failed)
- [tests/pkg_qdsa](tests/pkg_qdsa/test.failed)
- [tests/pthread_condition_variable](tests/pthread_condition_variable/test.failed)
- [tests/pthread_cooperation](tests/pthread_cooperation/test.failed)
- [tests/pthread_rwlock](tests/pthread_rwlock/test.failed)
- [tests/thread_flags](tests/thread_flags/test.failed)
- [tests/xtimer_periodic_wakeup](tests/xtimer_periodic_wakeup/test.failed)
- [tests/xtimer_usleep](tests/xtimer_usleep/test.failed)
- [tests/xtimer_usleep_short](tests/xtimer_usleep_short/test.failed)

I could fix the easy ones locally (pkg_hacl, pkg_monocypher) by increasing the timeout value, pkg_qdsa I'm not sure out to fix it, we know the story with periph_flashpage, others are related to a problem with xtimer.

dylad:
>  others are related to a problem with xtimer.

Does that mean a bug in the timer driver ?

>So how do we proceed here, should we add the -e option to the flasher or enforce a page erase before write in the driver ?

First solution isn't enough, you cannot do two write in a row at the same page.
Second solution is broken, the erase command doesn't erase anything from the memory... (tested manually with tests/periph_flashpage).
I still have no idea of what's going on here. My suggestion for now is to use `DISABLE_MODULE=periph_flashpage` or to postpone this PR until we fix it.

aabadie:
> Does that mean a bug in the timer driver ?

It could be, but honestly I have no idea.

> My suggestion for now is to use DISABLE_MODULE=periph_flashpage

Or maybe something something like
```mk
ifeq (,$(filter saml1%-xpro,$(BOARD)))
  FEATURES_PROVIDED += periph_flashpage
  FEATURES_PROVIDED += periph_flashpage_raw
endif
```
in `cpu/sam0_common/Makefile.features`

dylad:
I must say that I like your solution better :)
How much time did I have to try to fix the flashpage driver ? This is not as important as the timer driver for a "basic support" so we can afford to drop this driver for now.
I must fix the timer driver ASAP since this is much more important.

dylad:
Lastest commit fixes almost all the issue you reported.

aabadie:
Have a look at https://github.com/aabadie/RIOT/commit/e81991c2d8e2d47d6606b1373b97f06e119c351f where I adapted the pkg_hacl and pkg_monocypher automatic tests. For pkg_qdsa I have no idea other than rewrite the test main.c (without sizeof and using strncmp to compare the strings)

aabadie:
For `pthread_cooperation` it could be a default stacksize issue. I tried with `CFLAGS += -DTHREAD_STACKSIZE_DEFAULT=512` in the application Makefile and it works.

aabadie:
I rerun the failing tests and now have those ones still failing (see above for potential fixes for some of them):
```
- [tests/periph_flashpage](tests/periph_flashpage/test.failed)
- [tests/pkg_hacl](tests/pkg_hacl/test.failed)
- [tests/pkg_monocypher](tests/pkg_monocypher/test.failed)
- [tests/pkg_qdsa](tests/pkg_qdsa/test.failed)
- [tests/pthread_cooperation](tests/pthread_cooperation/test.failed)
```

It's getting closer :)

dylad:
Ok i can confirm [tests/pkg_hacl] and [tests/pkg_monocypher] are working with the timeout (btw SAML21 also needs this timeout otherwise the test fails)
Regarding tests/pthread_cooperation, it is working with 'CFLAGS += -DTHREAD_STACKSIZE_DEFAULT=512' but I have a bad feeling... Why should we increase the stack for this MCU ? This is weird.

dylad:
More info about tests/pthread_cooperation: It also fails on SAML21 (if I hacked the RAM_LEN params to 16K) so at least this bug is not mine :) Not sure If it should be fix here. @aabadie could you try on another MCU arch with 16K RAM to see if you can reproduce this issue ? I don't have such hardware.
Finally about, tests/pkg_qdsa I found the issue. I'm working on a clean patch.

pkg/qDSA/Makefile.include
```diff
@@ -1,4 +1,4 @@
-ifneq (,$(filter cortex-m0%,$(CPU_ARCH)))
+ifneq (,$(filter cortex-m23,$(CPU_ARCH)))
```
aabadie: Why removing `cortex-m0%` ?

aabadie:
> could you try on another MCU arch with 16K RAM to see if you can reproduce this issue ?

I think this test is working on microbit (16k RAM)

> Why should we increase the stack for this MCU ?

The new stacksize was not increased but divided by 2. The default is 1024 on cortexm.

pkg/qDSA/Makefile.include
dylad: I missed a ctrl-s before commit, see lastest commit.

dylad:
> The new stacksize was not increased but divided by 2. The default is 1024 on cortexm.

12 threads with 1024 bytes per thread ? that's 12kB + all the stuff used by RIOT. Do we have enough RAM ?

aabadie:
> Do we have enough RAM ?

That's why decreasing THREAD_STACKSIZE_DEFAULT helps. I'm pretty sure this test also fails on smaller platforms. There are a few nucleos with 8kB RAM and they are not blacklisted. We could just blacklist saml1x-xpro for now.

aabadie:
And propose the stack reduce change in another PR.

larissazech:
Hi,

I am completely new to RIOT and the Cortex-M23 and as a first start, I tried to run the peripheral tests. I am using a Microchip SAML10 Xplained Pro without extensions. 

You seem to make rapid progress with this PR and I am a bit overwhelmed. Just to make sure that I am working with the current codebase: the latest commit in my git log is the fixup qdsa. And to make sure that I run the tests correctly: I execute `BOARD=saml10-xpro make all flash term` in the directory of the test I would like to run. 

—
My test results were as follows.

At first, I ran `test/periph_adc`. As of now, I have not been able to connect the ADC pins to any known voltage levels as requested in the test’s `README`. What exact voltage levels/range of voltage levels do you use for this test? A hint on where to find appropriate values in the docs from Microchip would be great as well! 

When running `test/periph_cpuid`, the output consists of 16 hex numbers. I will try to find a hint on how to validate this number within the Cortex-M23 Technical Reference Manual.

`test/periph_flashpage` provides me with the flash start address and the page size of 256 and page number of 256. According to the test’s `README`, I don’t want to mess with my board’s flash memory. I would greatly appreciate if you could give me some description on what test sequence you are using. A reference to the docs would also be sufficient! According to the board features listed on the Microchip webshop, there is 64KB flash memory. Nevertheless, I was unable to find a memory map. 

Can you provide me with an explanation or doc reference for the `test/periph_gpio`? There is no `README` in this directory and I could not find a wiki page, tutorial or any other RIOT documentation for the tests. The description I receive when executing the test does not help me either, I am very sorry.

Since I do not have any extensions or sensors, I am not sure if there is a way for me to sufficiently perform `test/periph_i2c`. Executing `i2c_acquire 0` and `i2c_release 0` both return `Success`. According to [1], page 7, the i2c connects the Microchip Embedded Debugger to the Control MCU of the Analog Module and offers a data line and a clock line. I do not see how this is related to the test’s read/write of registers/I2C devices and what input parameters are useful.

`test/periph_pm` outputs the test description. Just a side note: Is "Layered support is not unavailable for this CPU.“ correct? „Not unavailable“? As of now, I don’t have an external power measurement device.

`test/periph_rtc` outputs 4 alarms, each 2 seconds apart. Side note: the `README`s of `test/periph_rtc` and `test/periph_rtt` both claim that an alarm is coming every 10 seconds.

For `test/periph_rtt`, I waited for 10 hellos. All of them came 10 seconds apart.

What setup configuration do you use for `test/periph_spi`? According to the aforementioned user guide [1], „this is a one-way data transfer channel from the XAM to the EDBG“. Therefore I assume that it only makes sense to initialize the board as master and try to receive some current measurements from the XAM? I don’t see how `send - transfer string to slave` (output of `help`) is supposed to work for this board and what the benchmarks are going to do. 

`test/periph_timer` outputs a successful test of `TIMER_0`. 

`test/periph_uart` outputs 1 available device `UART_DEV(0)`. I will try to connect RX and TX tomorrow.

The following tests were outputted as unspecified features: `test/periph_dac`, `test/periph_eeprom`, `test/periph_hwrng`, `test/periph_pwm`, `test/periph_qdec`.
`test/periph_dma` was also outputted as unspecified feature. Nevertheless, it is executable and prints `DMA dummy test app`. 

—

In addition to the questions above, I have one last question concerning the tests as a whole. In this PR’s conversation, @aabadie published results of running „the full test suite“. Can you tell me where to find its documentation? Or explain to me how to run this test suite? 

I am aware that this is a very long list of open questions and I understand that you are concerned with fixing this pull request in time for the next release. Still, it would be great if you were able to give me any kind of feedback or help!

Thank you very much in advance for any comments!

Best,
Larissa


[1]  „SAM L10/L11 Xplained Pro User Guide“, published by Microchip, revision b

dylad:
@larissazech, thanks for playing with this PR. I'll try to answer most of your questions.

 > At first, I ran test/periph_adc. As of now, I have not been able to connect the ADC pins to any known voltage levels as requested in the test’s README. What exact voltage levels/range of voltage levels do you use for this test? A hint on where to find appropriate values in the docs from Microchip would be great as well!

IIRC, the SAM0 adc driver uses a 3,3V internal voltage as reference. For testing this feature, I connected a potentiometer to the pin declared in boards/common/saml1x/include/periph_conf.h. 

> test/periph_uart outputs 1 available device UART_DEV(0). I will try to connect RX and TX tomorrow.

The only UART interface declared is binded to the USB DBG port through the embedded debugger.

> test/periph_gpio

I suggest you to use the command make BOARD=saml10-xpro flash term. You will be able to access all the test command through the RIOT  shell. Then, just play around with the available commands (type '?' then the enter key, it will output the list of available command)

In general, you can manually test every tests/xxxx app using `make flash term`. It is a very good way to understand what is happening (from my point of view).

> test/periph_pm outputs the test description. Just a side note: Is "Layered support is not unavailable for this CPU.“ correct? „Not unavailable“? As of now, I don’t have an external power measurement device.

I'll probably add support for PM later. But I must rework some peripherals drivers first. For now, consider it as not supported.

> test/periph_flashpage

Not an expert on this driver, and it is broken for both SAML10/SAML11 right now. But the memory map of the device is available in its [datasheet.](http://ww1.microchip.com/downloads/en/DeviceDoc/SAM-L10L11%20Family-DataSheet%20-%20DS60001513B.pdf) See chapter 9.

Regarding what you're trying to do with I2C/SPI, I think it's unfeasible. XAM is connected to the control MCU which is not the SAML10 but either the AVR chip or the SAMD20 (in the bottom side of the dev board). So I don't think you can access it through the SAML10. I'll have a look at the schematics if I find some time.

aabadie:
> test/periph_pm outputs the test description. Just a side note: Is "Layered support is not unavailable for this CPU.“ correct? „Not unavailable“?

This is certainly a typo in the application's code.

> test/periph_rtc outputs 4 alarms, each 2 seconds apart. Side note: the READMEs of test/periph_rtc and test/periph_rtt both claim that an alarm is coming every 10 seconds.

Again this is a typo in `periph_rtc`'s README. In the code, the delay between each Alarm is 2s.

> test/periph_dma was also outputted as unspecified feature. Nevertheless, it is executable and prints DMA dummy test app.

This is because `periph_dma` cannot be used but the test simply consists in printing a message. If you have dma but it doesn't work, nothing will be printed. If you don't have DMA, something will be printed in any case. Maybe the application could make this more explicit using some macro.

> published results of running „the full test suite“. Can you tell me where to find its documentation? Or explain to me how to run this test suite?

This is not documented in the RIOT main repository. I was running task 99 from the release specs. Since it's a new board, it should full-fill all requirements of a release, see https://github.com/RIOT-OS/Release-Specs/blob/master/99-compile-and-test-one-board/99-compile-and-test-one-board.md
This the commands that I used:
```
$ export RIOTBASE=<path to RIOT containing this PR>
$ export BOARD=saml11-xpro
$ ./compile_and_test_for_board.py $RIOTBASE $BOARD
```
This takes quite some time, maybe one hour and a half.

dylad:
I think I have found the issue for the periph_flashpage : 
the erase command is broken for SAM0.
Procedure to test
use test_last to write the page, read to check if it's ok, now use erase (lastpage-2), re-read the page
Page is not blank at the end but still fill with test_last data.
I was able to reproduce this on SAML11 AND SAML21.
This seems to cause us pain for SAML10/SAML11 but SAML21 just doesn't care.
@fedepell Sorry to bother you but could you confirm it on your side that the erase command doesn't work as expected on your SAML21 ?

If @fedepell confirms it is a bug, I think we should open another PR to fix it.

Of course I have a patch for this :)
`NVMCTRL->ADDR.reg = (((uint32_t)page_addr) >> 1);`
must become
`NVMCTRL->ADDR.reg = (((uint32_t)page_addr) >> 1) << 1;`

dylad:
BTW This bug should be also presents on SAMD21, SAMR21. In fact all SAM0 MCUS.

aabadie:
Regarding tests with `pkg_hacl` and `pkg_monocypher`, I opened #10813 because I think saml21 is also affected.

larissazech:
@dylad @aabadie Many thanks to both of you for the very fast response and the very helpful explanations!

The SAML10/11-family data sheet is extremely helpful. I will study it carefully.

> Regarding what you're trying to do with I2C/SPI, I think it's unfeasible. [...] I'll have a look at the schematics if I find some time.

Never mind, this question came out of total confusion. Of course, these tests aim for the MCUs communication interfaces and not the XAM interface. I'm sorry.

boards/common/saml1x/include/periph_conf.h
```diff
@@ -0,0 +1,155 @@
...
+ */
+#define ADC_NUMOF                          (1U)
+
+/* ADC 0 Default values */
+#define ADC_0_CLK_SOURCE                   0 /* GCLK_GENERATOR_0 */
+#define ADC_0_PRESCALER                    ADC_CTRLB_PRESCALER_DIV256
+
+static const adc_conf_chan_t adc_channels[] = {
+    /* port, pin, muxpos */
+    {GPIO_PIN(PA, 10), ADC_INPUTCTRL_MUXPOS(ADC_INPUTCTRL_MUXPOS_AIN8)},
```
larissazech: Could you explain to me why you chose the GPIO_PIN PA10 here? According to the user guide, PA02 and PA03 are the ADC(+) and ADC(-) pins (see table 4-1. "Extension Header EXT1" on p. 13)?

dylad: There is not specific reason, It was a random choice. I just wanted to test one ADC line and it was the first pin I saw ;)
This can be changed if it's puzzled you.

aabadie:
> Lastest commit fix the flashpage driver. I re-run three times my test and I cannot reproduce anymore

Great thanks! I merged #10777 this afternoon. Can you rebase your branch on top of latest master ? This way I'm sure it will be tested with all new things (`git pull --rebase <upstream remote> master`)

dylad:
>  Can you rebase your branch on top of latest master ?

Done.

aabadie:
Ok thanks, I triggered Murdock in order to have an overview of the insufficient memory failures with this boards.

aabadie:
So I re-ran the magic `compile_and_test_for_board.py` and most of the problems are gone. There build failures due to insufficient memory:
```
Failures during compilation:
- [examples/asymcute_mqttsn](examples/asymcute_mqttsn/compilation.failed)
- [examples/dtls-echo](examples/dtls-echo/compilation.failed)
- [examples/gnrc_border_router](examples/gnrc_border_router/compilation.failed)
- [examples/gnrc_networking](examples/gnrc_networking/compilation.failed)
- [examples/gnrc_tftp](examples/gnrc_tftp/compilation.failed)
- [examples/javascript](examples/javascript/compilation.failed)
- [examples/lua_REPL](examples/lua_REPL/compilation.failed)
- [examples/lua_basic](examples/lua_basic/compilation.failed)
- [tests/conn_can](tests/conn_can/compilation.failed)
- [tests/gnrc_netif](tests/gnrc_netif/compilation.failed)
- [tests/gnrc_rpl_srh](tests/gnrc_rpl_srh/compilation.failed)
- [tests/gnrc_sixlowpan](tests/gnrc_sixlowpan/compilation.failed)
- [tests/gnrc_tcp_client](tests/gnrc_tcp_client/compilation.failed)
- [tests/gnrc_tcp_server](tests/gnrc_tcp_server/compilation.failed)
- [tests/lwip_sock_tcp](tests/lwip_sock_tcp/compilation.failed)
- [tests/unittests](tests/unittests/compilation.failed)
```

Only 3 tests are now failing:
```
Failures during test:
- [tests/gnrc_ipv6_ext](tests/gnrc_ipv6_ext/test.failed)
- [tests/ps_schedstatistics](tests/ps_schedstatistics/test.failed)
- [tests/pthread_cooperation](tests/pthread_cooperation/test.failed)
```

- `tests/gnrc_ipv6_ext` is failing for an unrelated reason: it needs root privilege to succeed...
- `tests/pthread_cooperation` it failing because of too much stack usage. It's a problem with the test itself, not by this PR
- `tests/ps_schedstatistics` is failing and it's new. When one uses the `schedstatistics` module, the board is crashing (not response on UART). Do you have any idea @dylad ?

So when `tests/ps_schedstatistics` test and build failures are fixed, we can merge this PR. Good job @dylad !

dylad:
>  Do you have any idea @dylad ?

I'll check if I don't throw those boards away...

dylad:
Add saml10-xpro and saml11-xpro to BLACKLIST for unitests because there is weird errors...
I don't know if it is the right solution... Any tips ?

fedepell:
@dylad : i tested now and as you already mentioned SAML21 seems not to be affected by the flash problem you found. 
I tried with master and with your branch and get same results, ie. erase works as expected!

dylad:
Thanks a lot for testing @fedepell !
the timer issue is not related to #10523. But for now, the workaround would be enough.

dylad:
Huh, I have a problem with unitests... I remove the BLACKLIST label but even if I put saml10-xpro and saml11-xpro in ARM_CORTEX_M_BOARDS and BOARD_INSUFFICIENT_MEMORY, the test is compiled and it failed with an internal compiler error.
I have no idea how to deal with it...

dylad:
Not proud of it but I found a possible workaround for my weird unittests bug. see lastest commit.
This PR is ready for one last round.

cpu/sam0_common/periph/gpio.c
```diff
@@ -220,17 +239,70 @@ void gpio_irq_disable(gpio_t pin)
     if (exti == -1) {
         return;
     }
-    EIC->INTENCLR.reg = (1 << exti);
+    _EIC->INTENCLR.reg = (1 << exti);
+}
+
+#ifdef CPU_SAML1X
+void isr_eic0(void)
```
aabadie: I would do this differently do avoid all this code duplication:

```c
#ifdef CPU_SAML1X
void isr_eic0(void)
{
    isr_eic()
}
...
void isr_eic1(void)
{
    isr_eic()
}

void isr_eic_other(void)
{
    isr_eic()
}
#endif /* SAML1X */

void isr_eic(void)
{
    for (unsigned i = 0; i < NUMOF_IRQS; i++) {
        if (_EIC->INTFLAG.reg & (1 << i)) {
            _EIC->INTFLAG.reg = (1 << i);
            gpio_config[i].cb(gpio_config[i].arg);
        }
    }
    cortexm_isr_end();
}
```

This is untested but I think it should work.

cpu/sam0_common/periph/i2c.c
```diff
@@ -92,19 +92,6 @@ void i2c_init(i2c_t dev)
 
     /* I2C using CLK GEN 0 */
     sercom_set_gen(bus(dev),i2c_config[dev].gclk_src);
-#if defined(CPU_FAM_SAML21) || defined(CPU_FAM_SAMR30)
```
aabadie: This seems unrelated. Is there a good reason to remove this part ?

dylad: What sorcery is this ?

aabadie:
I have 2 remaining code comments, and I think they will be my last ones.

I also re-ran the magic Python script and got no failure this time (except the 2 ones that we already know to be unrelated to this PR).

cpu/sam0_common/periph/i2c.c
dylad: Oh right I remember now.
In fact I removed this part on purpose because this is pointless. Sercom slow clock is used only by some SMBus features like timeout. But we don't use them at all, so no need for this clock (and it saves some power).

cpu/sam0_common/periph/gpio.c
dylad: I'll give it a try.

dylad:
comments answered/addressed.

cpu/sam0_common/periph/i2c.c
aabadie: It is unclear to me what this code is about. Are you sure it's strictly not required ? What if someone wants to use the slow clock ?

dylad: This code setup the SERCOM_GCLK_SLOW clock for the SERCOM your are using. (which is completely unrelated to the SERCOM_GCLK)
According to, SAML21 datasheet, the SERCOM_GLCK_SLOW is used for : 
```
Two generic clocks ared used by SERCOM, GCLK_SERCOMx_CORE and 
GCLK_SERCOM_SLOW.
The core clock (GCLK_SERCOMx_CORE) can clock the I 2 C when working as a master. The slow clock
(GCLK_SERCOM_SLOW) is required only for certain functions, e.g. SMBus timing. These clocks must
be configured and enabled in the Generic Clock Controller (GCLK) before using the I 2 C.

```
Those SMBus features can be active in the CTRLA register for each SERCOM. We're not using them in RIOT codebase.

See chapter 34.5.3 of [SAML21 datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/60001477A.pdf).

The SERCOM_SLOW clock is NOT used by UART nor SPI.

aabadie: I know it's annoying but I think this change should be in it's own PR because it's not only related to saml1x. In any case, there is no need for this PR to depend on it IIUC.

dylad: I reapplied this piece of code. (And I will remove it as soon as this PR is merged ;) )

aabadie:
The branch was squashed and the CI is now all green. Let's go.

Thanks again for the good work and your perseverance @dylad !

dylad:
> Thanks again for the good work and your perseverance @dylad !
Thanks for your help and your tests @aabadie ! I'm happy this one is finally merged :)

emmanuelsearch:
great!

keestux:
What I'm missing in the merge is a precise description where the vendor files came from. Is a copy from ASF? Then I want to see which version ASF and whether it was a verbatim copy, and if not, then describe what was changed.

Now that the merge is still fresh we should add the information before it is too late.

The place to document is `cpu/sam0_common/include/vendor/README.md`. There it is written that ASF 3.35 was used.

Notice the same can be said of #9437. In that PR the file `cpu/sam0_common/include/vendor/samr30/README.md` was added with a note that ASF 3.34.2 was used. Huh? Why?

keestux:
@dylad OK, I see. In that case we need to write down the version of these atpacks. I'll take a look and will come up with a (small) PR. Thanks for the URL, that helped. That said, URLs of Atmel are moving targets (to put it mildly).

keestux:
See #10914

PR #10656, opened 2018-12-26
cpu/esp8266: change of ETS task handling 

gschorcht:
### Contribution description

This PR changes the ETS task handling approach when using the Espressif SDK. This new ETS task handling approach is required for the upcoming ESP-NOW netdev driver. In addition, the default thread stack size can be reduced by half to save a lot of memory.

**Update** The PR is also prerequsit for refactoring in upcoming PR that reduces the code duplication by using some parts of the vendor code for ESP32 and ESP8266.

#### Background

To use the WiFi interface or software timers with the `esp_sw_timers` module, functions of the Espressif SDK have be used. For handling hardware interrupts, the SDK internally uses its own priority-based multitasking system, the so-called ETS:
- Instead of handling hardware interrupts directly in the interrupt service routines, there is an ETS task for each hardware component.
- The interrupt service routines use the `ets_post` ROM function to send an event to the corresponding ETS tasks, which then process the interrupts asynchronously.
- The ETS uses the ROM function `ets_run` to periodically execute all ETS tasks with pending interrupt events in an endless loop. Context switches are not possible in interrupt service routines.

If the SDK is used, the ETS tasks with the pending interrupt events must also be processed by the RIOT port. For that purpose, the RIOT port overwrites the ROM functions `ets_run` and `ets_post`:

-  The `ets_post` function is overridden to receive a call back to the RIOT system in interrupt service routines.
- The `ets_run` function is replaced by function `ets_task_run`. This function performs the execution of all ETS tasks with pending interrupt events exactly once.

Before this PR, function `ets_task_run` was called
- before every context switch and
- before and after going to sleep mode.

This approach had the disadvantage that context switches based on RIOT priorities were mixed with handling interrupt events in prioritized ETS tasks. This resulted in some stability issues with the WiFi interface and required a lot of memory for the thread stacks.

#### Changes

The approach introduced with this PR assumes that ETS tasks are only used to handle hardware interrupts and their handling requires therefore always the highest priority. It

- introduces a flag which indicates that there are ETS tasks with pending interrupt events,
- changes the `ets_post` function which is called at the end of an ETS interrupt service routine to set the  the flag, and
- introduces a separate thread `ets_task_func` with the highest possible priority which
  - waits for the flag and
  - calls the `ets_task_run` function to handle all ETS tasks with pending interrupt events once the flag is set.

Further changes with this PR are:
- default thread stack size is downsized to the half
- number of priority levels increased due to the additional thread
- context switches based on software-interrupts have to be used by default
- some cleanups of the makefile

### Testing procedure

Compile `tests/shell`  with enabled `ets_sw_timer` and flash any ESP8266 board, e.g.,
```
USEMODULE=ets_sw_timer make -C tests/shell BOARD=esp8266-eps-13x
```

### Issues/PRs references

This PR is prerequisite for PRs #9917 and #10792 as well ass refactoring according to issue #10658.

### DIFF
--- a/cpu/esp8266/Makefile.include
+++ b/cpu/esp8266/Makefile.include
@@ -24,6 +24,16 @@ endif
 
 ifeq ($(ENABLE_GDBSTUB), 1)
     USEMODULE += esp_gdbstub
+endif
+
+# SPECIAL module dependencies
+# cannot be done in Makefile.dep since Makefile.dep is included too late
+
+ifneq (, $(filter esp_sw_timer, $(USEMODULE)))
+    USEMODULE += esp_sdk
+endif
+
+ifneq (, $(filter esp_gdbstub, $(USEMODULE)))
     USEMODULE += esp_gdb
 endif
 
@@ -39,7 +49,17 @@ PSEUDOMODULES += esp_sdk_int_handling
 PSEUDOMODULES += esp_sw_timer
 PSEUDOMODULES += esp_spiffs
 
+USEMODULE += esp
+USEMODULE += mtd
+USEMODULE += periph
+USEMODULE += periph_common
+USEMODULE += ps
+USEMODULE += random
+USEMODULE += sdk
+USEMODULE += xtensa
+
 ifneq (, $(filter pthread, $(USEMODULE)))
+    # has to be included before $(ESP8266_NEWLIB_DIR)
     INCLUDES += -I$(RIOTBASE)/sys/posix/pthread/include
 endif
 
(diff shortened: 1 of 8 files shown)


gschorcht:
@aabadie @kaspar030 Is it possible to ask someone to review this PR because it is essential for further changes?

cpu/esp8266/Makefile.include
```diff
@@ -143,5 +147,5 @@ else
     export FFLAGS += -p $(PORT) -b $(PROGRAMMER_SPEED) write_flash
     export FFLAGS += -fm $(FLASH_MODE)
     export FFLAGS += 0 $(ELFFILE)-0x00000.bin
-    export FFLAGS += 0x10000 $(ELFFILE)-0x10000.bin
+    export FFLAGS += 0x10000 $(ELFFILE)-0x10000.bin; esptool.py -p $(PORT) run
```
smlng: this looks a bit weird, can you explain. To me this looks like the latter is a necessary POSTFLASH command?

gschorcht: Yep, it is the comment to reset after flashing. I couldn't figure out any other way to define something like that. RESET variable is not doing that since it requires to call `make reset` explicitly.

smlng: okay, then leave as is for now. But we should keep this in mind and clean it up later on - meaning: this should be handled properly by the build system.

smlng:
tested with board `esp8266-esp-12x` and `examples/default` (and others), all still working and also the esp_wifi is working now.

smlng:
please squash

gschorcht:
@smlng Many thanks for reviewing and testing.

Rebased and squashed.

PR #10662, opened 2018-12-27
cpu/esp32: GPIO macro cleanup in periph_cpu.h

gschorcht:
### Contribution description

The default macros GPIO_PIN and GPIO_UNDEF do not have to be overridden. The GPIO_PIN macro definition was even wrong for 40 GPIOs without splitting into ports, even if that did not lead to erroneous behavior.

### Testing procedure

Compile and flash `tests/periph_gpio` to an ESP32 board, e.g.,
```
make -C tests/periph_gpio BOARD=esp32-wroom-32 flash test
```

### DIFF
--- a/cpu/esp32/esp-eth/esp_eth_netdev.c
+++ b/cpu/esp32/esp-eth/esp_eth_netdev.c
@@ -109,10 +109,13 @@ static esp_err_t IRAM_ATTR _eth_input_callback(void *buffer, uint16_t len, void 
     return ESP_OK;
 }
 
-#if EMAC_PHY_POWER_PIN != GPIO_UNDEF
 static void _esp_eth_phy_power_enable_gpio(bool enable)
 {
     assert(EMAC_ETHERNET_PHY_CONFIG.phy_power_enable);
+
+    if (EMAC_PHY_POWER_PIN == GPIO_UNDEF) {
+        return;
+    }
 
     if (!enable) {
         EMAC_ETHERNET_PHY_CONFIG.phy_power_enable(false);
@@ -127,7 +130,6 @@ static void _esp_eth_phy_power_enable_gpio(bool enable)
         EMAC_ETHERNET_PHY_CONFIG.phy_power_enable(true);
     }
 }
-#endif
 
 static int _esp_eth_init(netdev_t *netdev)
 {
@@ -146,13 +148,13 @@ static int _esp_eth_init(netdev_t *netdev)
     config.gpio_config = _eth_gpio_config_rmii;
     config.tcpip_input = _eth_input_callback;
 
-    #if EMAC_PHY_POWER_PIN != GPIO_UNDEF
     /*
      * Replace the default 'power enable' function with an example-specific
      * one that toggles a power GPIO.
      */
-    config.phy_power_enable = phy_device_power_enable_via_gpio;
-    #endif
+    if (EMAC_PHY_POWER_PIN != GPIO_UNDEF) {
+        config.phy_power_enable = _esp_eth_phy_power_enable_gpio;
+    }
 
     /* initialize EMAC with config */
     if (ret == ESP_OK && (ret = esp_eth_init(&config)) != ESP_OK) {
--- a/cpu/esp32/include/periph_cpu.h
+++ b/cpu/esp32/include/periph_cpu.h
@@ -47,16 +47,6 @@ extern "C" {
 /** @} */
 
 /**
- * @brief   Definition of a fitting UNDEF value
- */
-#define GPIO_UNDEF (0xff)
-
-/**
- * @brief   Define CPU specific GPIO pin generator macro
- */
-#define GPIO_PIN(x, y)  ((x << 4) | y)
-
-/**
  * @brief   Define CPU specific number of GPIO pins
  * @{
  */


cpu/esp32/esp-eth/esp_eth_netdev.c
```diff
@@ -109,11 +109,14 @@ static esp_err_t IRAM_ATTR _eth_input_callback(void *buffer, uint16_t len, void
     return ESP_OK;
 }
 
-#if EMAC_PHY_POWER_PIN != GPIO_UNDEF
 static void _esp_eth_phy_power_enable_gpio(bool enable)
 {
     assert(EMAC_ETHERNET_PHY_CONFIG.phy_power_enable);
 
+    if (EMAC_PHY_POWER_PIN == GPIO_UNDEF) {
```
aabadie: This seems unrelated to this PR, or maybe I'm missing something ?

aabadie:
Regarding the testing procedure, i tried `make -C tests/periph_gpio BOARD=esp32-wroom-32 flash test` but it doesn't work because the `test` tries to reset the board while a terminal is already opened. This seems to cause problem with `esptool.py`.
There is a similar issue with arduino and firefly boards because their flasher uses the UART to flash/reset the boards, so you can't keep a terminal opened while resetting/flashing them.

Is there any other way to test this PR using the `tests/periph_gpio` application ?

cpu/esp32/esp-eth/esp_eth_netdev.c
gschorcht: It is related since I had to change with the removing the the GPIO_UNDEF macro
```
#define GPIO_UNDEF (0xff)
```
and using the default definition from `periph/gpio.h`.
```
#define GPIO_UNDEF          ((gpio_t)(UINT_MAX))
```
Since the latter one uses a type, I could not use `GPIO_UNDEF` anymore to disable the whole function. Now, the function exists but returns.
```diff
- #if EMAC_PHY_POWER_PIN != GPIO_UNDEF	
  static void _esp_eth_phy_power_enable_gpio(bool enable)
  {
      assert(EMAC_ETHERNET_PHY_CONFIG.phy_power_enable);
+     if (EMAC_PHY_POWER_PIN == GPIO_UNDEF) {
+        return;
+    }
      ...
  }
- #endif

**Comment by gschorcht** [comment]
> Regarding the testing procedure, i tried make -C tests/periph_gpio BOARD=esp32-wroom-32 flash test but it doesn't work because the test tries to reset the board while a terminal is already opened. This seems to cause problem with esptool.py.

@MrKevinWeiss also observed this problem.

> There is a similar issue with arduino and firefly boards because their flasher uses the UART to flash/reset the boards, so you can't keep a terminal opened while resetting/flashing them.

> Is there any other way to test this PR using the tests/periph_gpio application ?

Yes, just press the `RST` (on some boards also called `EN`) button once you see 
```
Welcome to pyterm!
Type '/exit' to exit.
```
in terminal window.

**Comment by smlng** [comment]
@gschorcht is right, that the `GPIO_PIN` macro cannot be used for ESP right now? When I look at the datasheet `GPIOxy` does not match to pin number `xy`.

**Comment by gschorcht** [comment]
@smlng 
> is right, that the GPIO_PIN macro cannot be used for ESP right now?

Why not? With the default definition of `GPIO_PIN` in `periph/gpio.h`, `GPIO_PIN(x, y) simply refers ESP32 GPIOy. That is, there is only one port with 40 pins. The `GPIOn` macro definition simply map to `GPIO_PIN(0,n)`.

**Comment by smlng** [comment]
I looked into this [data sheet](https://www.espressif.com/sites/default/files/documentation/esp32_datasheet_en.pdf), and I was wondering because in `Table 1: Pin Description` for instance `GPIO25` is on pin `14`, `GPIO2` on `22` and `GPIO0` on `23`.

Using the default `GPIO_PIN` macro would map `GPIO_PIN(0, 25)` to `25` but according to the above it should actually be `14`, or what did I miss here?

**Comment by gschorcht** [comment]
Yes, but this is probably the case for most CPUs. For example:
```
$ find cpu -name '*.h' -exec grep '#define GPIO_PIN(' {} \; -print
#define GPIO_PIN(port, pin) (port<<5 | pin)
cpu/lpc2387/include/periph_cpu.h
#define GPIO_PIN(x, y)      (((uint32_t)PIOA + (x << 9)) | y)
cpu/sam3/include/periph_cpu.h
#define GPIO_PIN(x, y)      ((gpio_t) ((x << 4) | y))
cpu/efm32/include/periph_cpu.h
...
#define GPIO_PIN(x, y)      ((x << 4) | y)
cpu/ezr32wg/include/periph_cpu.h
#define GPIO_PIN(x, y)      ((gpio_t)(((x & 0xff) << 8) | (1 << (y & 0x07))))
cpu/msp430fxyz/include/periph_cpu.h
#define GPIO_PIN(x,y) ((gpio_t)((x<<4) | y))
cpu/lm4f120/include/periph_cpu.h
#define GPIO_PIN(x,y)       ((x << 4) | (y & 0xf))
cpu/mips_pic32_common/include/periph_cpu_common.h
#define GPIO_PIN(port, pin)     (gpio_t)((port << 5) | pin)
cpu/lpc1768/include/periph_cpu.h
#define GPIO_PIN(x, y)  ((x << 4) | y)
cpu/esp8266/include/periph_cpu.h
#define GPIO_PIN(x, y)          ((x << 4) | y)
cpu/atmega1284p/include/periph_cpu.h
...
```
All of them generate some kind of an ordinal which doesn't correspond to the pin at the chip. Rather, this ordinal is used in `gpio_*` function to map it on the right pin of the chip. From my point of view, `GPIO_PIN(x,y)` does not mean the pin of the chip but the y-th pin at x-th port.

smlng:
> All of them generate some kind of an ordinal which doesn't correspond to the pin at the chip. Rather, this ordinal is used in gpio_* function to map it on the right pin of the chip. From my point of view, GPIO_PIN(x,y) does not mean the pin of the chip but the y-th pin at x-th port.

that's only partly right, I think - because for most other CPUs there is at least some relation chip and order of the GPIO PIN number and the corresponding pin on the CPU, i.e. GPIO0 < GPIO1 < GPIO2 ... 

But for the ESPs it seems to be all over the place 😄 Anyway, my initial fear was that the `GPIO_PIN` macro might not work here as expected, but I now see there are further macros and code that do the correct mapping.

gschorcht:
> that's only partly right, I think - because for most other CPUs there is at least some relationship and order of the GPIO PIN number and the corresponding pin on the CPU, i.e. GPIO0 < GPIO1 < GPIO2 ...

Even though they are in ascending order, the ordinals are not the pins of the chips. For some CPUs, e.g. Atmega328, the actual pinout even depends on the chip package.

> Anyway, my initial fear was that the GPIO_PIN macro might not work here as expected, but I now see there are further macros and code that do the correct mapping.

For all boards I know, the GPIOs on the boards are labeled with something like `IO0...IO39` or simply `0...39`. As long as the `GPIO_PIN(0,0) ... GPIO_PIN(0, 39)` is used, everything should work as expected.

smlng:
please squash

gschorcht:
Rebased and squashed

gschorcht:
@smlng can we merge?

smlng:
yep

gschorcht:
@smlng Thanks for reviewing and testing :smile:

PR #10671, opened 2018-12-28
net/nanocoap: fix confirmable retry countdown

### LINKED ISSUE #9191: Logic errors in nanocoap_get()

kb2ma:
### Contribution description
The nanocoap sock module includes a `nanocoap_request()` function, which sends the request confirmably. As described in #9191, the function does not perform the expected count of retries when a send fails. This PR fixes the problem with a rework of the countdown logic.

### Testing procedure
This PR extends the `nanocoap_cli` test in #10640 with a server that provides an option (-i) to ignore a count of requests. So, you can setup a tap bridge with both endpoints running this test. One of them runs as a server and the other as a client.

Try setting up the server to ignore 4 requests. The client will timeout on the RIOT master branch, but this PR should succeed on the final retry. When the server is set up to ignore 5 requests, the client should timeout on this PR, as expected. I created an unofficial [automated test](https://github.com/kb2ma/riot-coap-pytest/commits/nanocoap_retries) that uses a similar approach.

### Issues/PRs references
Fixes #9191. Depends on #10640.

### DIFF
--- a/sys/net/application_layer/nanocoap/sock.c
+++ b/sys/net/application_layer/nanocoap/sock.c
@@ -52,7 +52,7 @@ ssize_t nanocoap_request(coap_pkt_t *pkt, sock_udp_ep_t *local, sock_udp_ep_t *r
 
         res = sock_udp_send(&sock, buf, pdu_len, NULL);
         if (res <= 0) {
-            DEBUG("nanocoap: error sending coap request\n");
+            DEBUG("nanocoap: error sending coap request, %d\n", (int)res);
             break;
         }
 
@@ -68,7 +68,7 @@ ssize_t nanocoap_request(coap_pkt_t *pkt, sock_udp_ep_t *local, sock_udp_ep_t *r
                 }
                 continue;
             }
-            DEBUG("nanocoap: error receiving coap request\n");
+            DEBUG("nanocoap: error receiving coap response, %d\n", (int)res);
             break;
         }
         else {
--- a/sys/net/application_layer/nanocoap/sock.c
+++ b/sys/net/application_layer/nanocoap/sock.c
@@ -46,19 +46,14 @@ ssize_t nanocoap_request(coap_pkt_t *pkt, sock_udp_ep_t *local, sock_udp_ep_t *r
 
     /* TODO: timeout random between between ACK_TIMEOUT and (ACK_TIMEOUT *
      * ACK_RANDOM_FACTOR) */
-    uint32_t timeout = COAP_ACK_TIMEOUT * (1000000U);
-    int tries = 0;
-    while (tries++ < COAP_MAX_RETRANSMIT) {
-        if (!tries) {
-            DEBUG("nanocoap: maximum retries reached.\n");
-            res = -ETIMEDOUT;
-            goto out;
-        }
+    uint32_t timeout = COAP_ACK_TIMEOUT * US_PER_SEC;
+    unsigned tries_left = COAP_MAX_RETRANSMIT + 1;  /* add 1 for initial transmit */
+    while (tries_left) {
 
         res = sock_udp_send(&sock, buf, pdu_len, NULL);
         if (res <= 0) {
             DEBUG("nanocoap: error sending coap request\n");
-            goto out;
+            break;
         }
 
         res = sock_udp_recv(&sock, buf, len, timeout, NULL);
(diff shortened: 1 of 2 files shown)


sys/net/application_layer/nanocoap/sock.c
```diff
@@ -46,19 +46,14 @@ ssize_t nanocoap_request(coap_pkt_t *pkt, sock_udp_ep_t *local, sock_udp_ep_t *r
...
      * ACK_RANDOM_FACTOR) */
-    uint32_t timeout = COAP_ACK_TIMEOUT * (1000000U);
-    int tries = 0;
-    while (tries++ < COAP_MAX_RETRANSMIT) {
-        if (!tries) {
-            DEBUG("nanocoap: maximum retries reached.\n");
-            res = -ETIMEDOUT;
-            goto out;
-        }
+    uint32_t timeout = COAP_ACK_TIMEOUT * (1 * US_PER_SEC);
```
smlng: I guess you can skip the `1 *` and simplify to `COAP_ACK_TIMEOUT * US_PER_SEC`, or not?

sys/net/application_layer/nanocoap/sock.c
```diff
@@ -46,19 +46,14 @@ ssize_t nanocoap_request(coap_pkt_t *pkt, sock_udp_ep_t *local, sock_udp_ep_t *r
...
+    unsigned tries_left = COAP_MAX_RETRANSMIT + 1;  /* add 1 for initial transmit */
+    while (tries_left) {
```
smlng: I guess the *correct* (more C'ish) (?) way would be to use `do { ... } while ( ... );`, like:

```
unsigned tries_left = COAP_MAX_RETRANSMIT;
do {
[...]
} while (tries_left--);
```

This should also work for `COAP_MAX_RETRANSMIT = 0`, that is: send once and no retries.
though I also guess this will not impact code size, as the compiler will optimise both the same way.

sys/net/application_layer/nanocoap/sock.c
kb2ma: Yeah, that was silly. Fixed in 7e8cc3b.

sys/net/application_layer/nanocoap/sock.c
kb2ma: In my experience, use of the decrement operator on an unsigned value is error prone. In this case, how do I display the "maximum retries reached" debug message when max retransmit is zero? I tested, and it doesn't work. The decrement operator rolled it back over.

I'm OK with converting tries_left to a signed value, but then I would want to use an explicit inequality in the do test "tries_left-- > 0". If I don't the logic would look strange against the test for debug below.

```
    } while (tries_left-- > 0);

    if (tries_left <= 0) {
        DEBUG("nanocoap: maximum retries reached\n");
    }
```

I still like this solution better than my original because I think the overall flow is easier to follow. Let me know what you think.

smlng: yeah, agreed - leave as is

smlng:
please rebase and squash

kb2ma:
Squashed and rebased.

kb2ma:
ping @smlng

kb2ma:
Murdock is complaining about the type of a printf argument in nanocoap_server and nanocoap_cli on mips platforms:
```
sock.c:55:13: error: format '%d' expects argument of type 'int', but argument 2 has type 'ssize_t' [-Werror=format=]
             DEBUG("nanocoap: error sending coap request, %d\n", res);
```
I plan to cast to ssize_t, squash, and force push.

kb2ma:
Updated the 'net/nanocoap: improve debug messages' commit. As planned, the only change is the cast to int for the arg in those two debug messages. Let's see if Murdock is happy now.

kb2ma:
Looks good to me, @smlng. I'll leave it for you to confirm.

PR #10687, opened 2018-12-31
net/nanocoap: remove obsolete macro

kb2ma:
### Contribution description
Removes definition of NANOCOAP_URL_MAX macro. This macro was useful prior to #9156, when the resource path for a gcoap message was stored in a coap_pkt_t struct.

### Testing procedure
Do a text search for the macro to confirm it is unused.

### Issues/PRs references
-none-


kb2ma:
Thanks for stepping in, @kaspar030!

PR #7652, opened 2017-09-28
drivers: add driver mcp23017

dnahm:
This PR includes the driver for MCP23017 - 16-Bit I/O Expander with Serial Interface.
The aconno board ACD52832 includes this IC. #7501 and #7557 are required for testing.

PR #10688, opened 2019-01-02
drivers: add MCP23x17 I/O expander support

gschorcht:
### Contribution description

This PR adds a driver for Microchip MCP23x17 I/O expanders which can be used to extend the number of GPIOs over I2C or SPI. It is a complete rework of the driver that was provided in PR #7652 and rewritten from scratch.

The following expander devices are supported:

 Expander | Type                | Interface | Pseudomodule
 :--------|:--------------------|:----------|:------------
 MCP23017 | 16-bit I/O expander | I2C       | `mcp23017` or `mcp23x17_i2c`
 MCP23S17 | 16-bit I/O expander | SPI       | `mcp23s17` or `mcp23x17_spi`

The main features of the driver are:
 - Its interface is compatible with the peripheral GPIO interface. The only difference is that the functions have the prefix `mcp23x17_` and require an additional parameter, the pointer to the expander device of type `mcp23x17_`
- ~It supports the GPIO extension API as defined in PR #9582 and provided in PR #9860 and #9958.~
- It supports the following GPIO modes
  - `GPIO_IN`, `GPIO_IN_PU` and `GPIO_OUT` in hardware
  - `GPIO_OD` and `GPIO_OD_PU` are emulated.
- Optionally, the a low-active push-pull INT signal of MCP23x17 expanders can be used to provide external interrupt capabilities for all expander pins. The driver interface is compatible to module `periph_gpio_irq`
- It supports direct GPIO mapping to SAUL which is compatible to the peripheral GPIO SAUL interface. Thus MCP23x17 I/O expander pins can be used with SAUL in the same way as peripheral GPIOS.
- Multiple MCP23x17 expander modules and different variants can be used simultaneously.

## Benchmarks

MCP23x17 driver run-time performance benchmark over SPI using 10 MHz clock speed:
```
                 nop loop:       634us  ---   0.063us per call  ---   15772870 calls per sec
                 gpio_set:    540009us  ---  54.000us per call  ---      18518 calls per sec
               gpio_clear:    540000us  ---  54.000us per call  ---      18518 calls per sec
              gpio_toggle:    811005us  ---  81.100us per call  ---      12330 calls per sec
                gpio_read:    271001us  ---  27.100us per call  ---      36900 calls per sec
               gpio_write:    539005us  ---  53.900us per call  ---      18552 calls per sec
```
MCP23x17 driver run-time performance benchmark over SPI using 1 MHz clock speed:
```
                 nop loop:       634us  ---   0.063us per call  ---   15772870 calls per sec
                 gpio_set:   1429044us  ---  142.904us per call  ---       6997 calls per sec
               gpio_clear:   1429005us  ---  142.900us per call  ---       6997 calls per sec
              gpio_toggle:   2144009us  ---  214.400us per call  ---       4664 calls per sec
                gpio_read:    715000us  ---  71.500us per call  ---      13986 calls per sec
               gpio_write:   1428004us  ---  142.800us per call  ---       7002 calls per sec
```
MCP23x17 driver run-time performance benchmark over I2C using 400 kHz clock speed:
```
                 nop loop:       639us  ---   0.063us per call  ---   15649452 calls per sec
                 gpio_set:   1965040us  ---  196.504us per call  ---       5088 calls per sec
               gpio_clear:   1960008us  ---  196.000us per call  ---       5102 calls per sec
              gpio_toggle:   3088493us  ---  308.849us per call  ---       3237 calls per sec
                gpio_read:   1127503us  ---  112.750us per call  ---       8869 calls per sec
               gpio_write:   1960004us  ---  196.000us per call  ---       5102 calls per sec
```

Reference board was an `esp8266-esp-12x`.

### Testing procedure

The test application `tests/driver_mcp23x17` can be used to test each MCP23x17 I/O expander pin with shell commands.

1. Compile the application. Dependent on the MCP23x17 expander module, compile it with at least one of the pseudomodules `mcp23x17_i2c` or `mcp23x17_spi`, e.g., for the I2C version MCP23017:
    ```
    USEMODULE=mcp23x17_i2c make flash -C tests/driver_mcp23x17 BOARD=...
    ```
2. Add the interrupt functionality by enabling module `mcp23x17_irq` and override the default GPIO pin for the interrupt signal, e.g. for the I2C version MCP23017:
    ```
    CFLAGS="-DMCP23X17_PARAM_I2C_INT=\(GPIO\(0,6\)\)" \
    USEMODULE="mcp23x17_i2c mcp23x17_irq"  flash -C tests/driver_mcp23x17 BOARD=...
    ```

3. Connect the MCP23x17 I/O expander module using the interface and the GPIO for the interrupt signal as defined during compilation.
4. Connect a GPIO pin of the MCU, e.g. `GPIO_PIN(0,4)`, and a MCP23x17 I/O expander pin, e.g. Port A (port 16 in test application) and Pin 6.
5. Execute some or all of the following commands (results in comment):
   ```
    > init_in 0 4
    > init_od_pu 16 6

    > set 0 4
    > read 0 4          # GPIO pin (port 0, pin 4)) is HIGH

    > clear 16 6
    > read 0 4          # GPIO pin (port 0, pin 4)) is LOW

    > init_od 16 6
    > init_in_pu 0 4

    > set 16 6
    > read 0 4          # GPIO pin (port 0, pin 4)) is HIGH

    > clear 16 6
    > read 0 4          # GPIO pin (port 0, pin 4)) is LOW

    > init_in 16 6
    > read 16 6         # MCP23x17 pin (dev 16, pin 6)) is HIGH
    > init_out 0 4 
    > read 16 6         # MCP23x17 pin (dev 16, pin 6)) is LOW

    > set 0 4 
    > read 16 6         # MCP23x17 pin (dev 16, pin 6)) is HIGH

    > init_int 16 6 0
    > clear 0 4         # INT: external interrupt from MCP23x17 pin 6
    > set 0 4

    > init_int 16 6 1
    > clear 0 4
    > set 0 4           # INT: external interrupt from MCP23x17 pin 6

    > init_int 16 6 2
    > clear 0 4         # INT: external interrupt from MCP23x17 pin 6
    > set 0 4           # INT: external interrupt from MCP23x17 pin 6 

    > disable_int 16 6
    > clear 0 4
    > enable_int 16 6
    > set 0 4           INT: external interrupt from MCP23x17 pin 6
    > clear 0 4         INT: external interrupt from MCP23x17 pin 6

    > init_int 16 5 2
    > disable_int 16 5
    > set 0 4           INT: external interrupt from MCP23x17 pin 6
   ```

### Issues/PRs references

~The driver is compatible with the GPIO extension API, PRs #9582, #9860 and #9958, but it does not depend on it.~


drivers/mcp23x17/mcp23x17.c
```diff
@@ -0,0 +1,578 @@
...
+    if ((res = _write(dev, MCP23X17_REG_IOCONA, &iocon, 1)) ||
+        (res = _write(dev, MCP23X17_REG_IOCONB, &iocon, 1))) {
+        DEBUG_DEV("error writing IOCON registers", dev);
+        return res;
+    }
+
+    return MCP23X17_OK;
+}
+
+int mcp23x17_gpio_init(mcp23x17_t *dev, gpio_t pin, gpio_mode_t mode)
```
jeandudey, line 226: I'm bit unsure about using `gpio_t` in here as some CPUs could define it as an structure, I think it may be more suitable to add a proper `mc23x17_gpio_t` type, same could apply for mode as CPUs could define their own and probably the IO expander has it's modes defined independently of the host.

jeandudey (commented):
This might be a little bit old but left some comments for a first pass, I hope it gets a rebase in the near future as the driver seems very complete.

drivers/mcp23x17/mcp23x17.c
gschorcht: `gpio_t` and `gpio_mode_t` are used here intentionally. As a result of the review of the first driver provided in PR #7652, the goal of this PR was to implement an API that is compatible with the GPIO API in order to integrate the GPIO expanders transparently.

At the moment, gpio_t must be a scalar for all GPIO implementations because there are hundreds of comparisons in RIOT that use the == operator.

In PR #12877 and later in PR #14602 I tried to introduce a GPIO API using a structured `gpio_t` type more than a year ago, but I didn't succeed until now. So there is no reason not to use `gpio_t` at the moment. Once, the structured GPIO is introduced, the driver could easily be changed, as the [example of the low-level GPIO API implementation of an expander driver](https://github.com/RIOT-OS/RIOT/pull/14602/files#diff-f5b86696beed1342e9731fb44e4aa41f0c744de7d25c9ee5933731675734fe0c) in PR #14602 shows.

drivers/mcp23x17/Kconfig
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2021 Gunar Schorcht
```
gschorcht, line 1: @leandrolanzieri How can I test the config for a new driver. Everytime I use `menuconfig` as make target, I only see `LM75A/TMP1075` in `Sensor Device Drivers`. When I use `[A] Toggle show-all mode`, I see all variables in red, but I can't change anything. Furthermore, I get the following message in red.
```
*** !! ERROR: There are conflicting modules active !! ***
```

leandrolanzieri: @gschorcht to test the dependency modelling with Kconfig you should set a variable: `env TEST_KCONFIG=1 make menuconfig`. 

>Furthermore, I get the following message in red.
>`*** !! ERROR: There are conflicting modules active !! ***`

Everything that is in red is not visible, so that message is actually hidden if I understood you correctly.

For this modelling I'd suggest the following change:

```diff
diff --git a/drivers/mcp23x17/Kconfig b/drivers/mcp23x17/Kconfig
index ca79fcc382..705780263b 100644
--- a/drivers/mcp23x17/Kconfig
+++ b/drivers/mcp23x17/Kconfig
@@ -36,6 +36,8 @@ config MODULE_MCP23X17_IRQ
     depends on HAS_PERIPH_GPIO_IRQ
     depends on MODULE_MCP23X17
     select MODULE_PERIPH_GPIO_IRQ
-    select MODULE_EVENT_THREAD_MEDIUM
+    depends on MODULE_EVENT_THREAD_MEDIUM
+    help
+        To use the IRQs the MODULE_EVENT_THREAD symbol should be set.
 
 endif # MODULE_MCP23X17
diff --git a/tests/driver_mcp23x17/app.config.test b/tests/driver_mcp23x17/app.config.test
index 181cba5f04..9d5911c6bf 100644
--- a/tests/driver_mcp23x17/app.config.test
+++ b/tests/driver_mcp23x17/app.config.test
@@ -2,6 +2,10 @@
 # application configuration. This is only needed during migration.
 CONFIG_MODULE_MCP23X17=y
 
+# enable event thread to process IRQs
+CONFIG_MODULE_EVENT=y
+CONFIG_MODULE_EVENT_THREAD=y
+CONFIG_MODULE_EVENT_THREAD_MEDIUM=y
 # enable interrupt support by default
 CONFIG_MODULE_MCP23X17_IRQ=y
```

You may also want to expose which thread priority to use on the driver, similar to the `at` driver (see [here](https://github.com/RIOT-OS/RIOT/blob/9256970517798bd844f4ee1dc4cddd9f61cef6e1/drivers/at/Kconfig#L22-L48)).

gschorcht: @leandrolanzieri Either I still don't understand how Kconfig is used correctly or there is something wrong with ESP8266 Kconfig. After making the changes according to your suggestions and running Kconfig, I get the following output for `USEMODULE` variable:
```
TEST_KCONFIG=1 make -C tests/driver_mcp23x17/ BOARD=esp8266-esp-12x menuconfig
TEST_KCONFIG=1 make -C tests/driver_mcp23x17/ BOARD=esp8266-esp-12x info-debug-variable-USEMODULE
```suggestion
benchmark tsrb core_init periph_init cpu periph_init_uart newlib stdin mcp23x17_irq event core_panic mcp23x17 periph_i2c periph_init_gpio periph_timer mcp23x17_i2c core periph_uart periph_init_pm event_thread periph_common event_thread_medium stdio_uart_rx shell periph_init_timer core_msg periph_gpio_irq newlib_syscalls_default auto_init core_thread_flags periph_gpio periph_pm div sys periph_init_i2c auto_init_xtimer stdio_uart isrpipe board xtimer core_idle_thread periph_init_gpio_irq mcp23x17_irq_medium
```
When I set `USEMODULE` as environment variable, I get the following output for `USEMODULE` variable:
```
USEMODULE='mcp23x17_i2c mcp23x17_irq' make -C tests/driver_mcp23x17 BOARD=esp8266-esp-12x info-debug-variable-USEMODULE

auto_init auto_init_random auto_init_xtimer benchmark board boards_common_esp8266 core core_idle_thread core_init core_msg core_panic core_thread_flags cpp cpu div esp_common esp_common_periph esp_freertos esp_freertos_common esp_i2c_sw esp_idf esp_idf_esp8266 esp_idf_nvs_flash esp_idf_spi_flash esp_idf_util esp_idf_wpa_supplicant_crypto esp_sdk event event_thread event_thread_medium isrpipe log luid mcp23017 mcp23x17 mcp23x17_i2c mcp23x17_irq mtd newlib newlib_syscalls_default periph periph_common periph_cpuid periph_flash periph_gpio periph_gpio_irq periph_hwrng periph_i2c periph_i2c_sw periph_init periph_init_cpuid periph_init_flash periph_init_gpio periph_init_gpio_irq periph_init_hwrng periph_init_i2c periph_init_i2c_sw periph_init_pm periph_init_timer periph_init_uart periph_pm periph_timer periph_uart prng prng_tinymt32 ps random shell stdin stdio_uart stdio_uart_rx sys tinymt32 tsrb xtensa xtimer
```
That is, a number of specific `esp_*` modules are missing in Kconfig configuration for ESP8266 boards. The same happens for ESP32 boards.

leandrolanzieri: We still did not make it to the esp for the Kconfig modelling of dependencies, that's why it's not tested yet in the CI https://github.com/RIOT-OS/RIOT/blob/9256970517798bd844f4ee1dc4cddd9f61cef6e1/.murdock#L10-L18 (help doing so is really welcome :). It should be working fine with the platforms that already have their modules in.

gschorcht: @leandrolanzieri Is there a more elegant way to map a possible choice to an enum than the following?

`driver.h`
```.c
typedef enum {
    DRIVER_MODE_1 = 0x10,
    DRIVER_MODE_2 = 0x20,
    DRIVER_MODE_3 = 0x30,
} driver_mode_t;
```
`driver_params.h`
```.c
#ifdef CONFIG_DRIVER_MODE_1
#define CONFIG_DRIVER_MODE    (DRIVER_MODE_1)
#elif CONFIG_DRIVER_MODE_2
#define CONFIG_DRIVER_MODE    (DRIVER_MODE_2)
#elif CONFIG_DRIVER_MODE_3
#define CONFIG_DRIVER_MODE    (DRIVER_MODE_3)
#endif
```

Kconfig:
```
choice
    bool "Select right mode
    default DRIVER_MODE_1

config DRIVER_MODE_1
    bool "Mode 1"

config DRIVER_MODE_2
    bool "Mode 2"

config DRIVER_MODE_3
    bool "Mode 3"

endchoice
```
When migrating default configuration parameters to `Kconfig`, a further question arose. What are the rules when a define in `driver_params.h` should be called `CONFIG_DRIVER_PARAM_<NAME>` and when a parameter should simply be called `DRIVER_PARAM_<NAME>`? I have seen a mixture which doesn't seem to be consistent.

Would be a mapping
```
#define DRIVER_PARAM_<NAME> (CONFIG_DRIVER_PARAM_<NAME>`
```
OK to keep the compatibility with existing documentation where the default parameters could be changed by overriding them by `CFLAGS` at the make command line?

leandrolanzieri: > @leandrolanzieri Is there a more elegant way to map a possible choice to an enum than the following?

What about the following:

```Kconfig
choice
    bool "Select right mode"
    default DRIVER_MODE_1

config DRIVER_MODE_1
    bool "Mode 1"

config DRIVER_MODE_2
    bool "Mode 2"

config DRIVER_MODE_3
    bool "Mode 3"

endchoice

config DRIVER_MODE
    hex
    default DRIVER_MODE_1_VAL if DRIVER_MODE_1
    default DRIVER_MODE_2_VAL if DRIVER_MODE_2
    default DRIVER_MODE_3_VAL if DRIVER_MODE_3

config DRIVER_MODE_1_VAL
    hex
    default 0x10

config DRIVER_MODE_2_VAL
    hex
    default 0x20

config DRIVER_MODE_3_VAL
    hex
    default 0x30
```
Then you can also have:
```C
typedef enum {
    DRIVER_MODE_1 = CONFIG_DRIVER_MODE_1_VAL, // not actually configurable, just not to duplicate
    DRIVER_MODE_2 = CONFIG_DRIVER_MODE_2_VAL,
    DRIVER_MODE_3 = CONFIG_DRIVER_MODE_3_VAL,
} driver_mode_t;
```
Note that this last part would only work in the case Kconfig is actually run.

---

> What are the rules when a define in `driver_params.h` should be called `CONFIG_DRIVER_PARAM_<NAME>` and when a parameter should simply be called `DRIVER_PARAM_<NAME>`? I have seen a mixture which doesn't seem to be consistent.

Personally I think many of those parameters are more a hardware configuration rather than a configuration of the driver itself, and would be better off left for other tools (e.g. the current `driver_params.h` or Device Tree). Nevertheless there are some drivers that expose these default values for their first instance as Kconfig symbols.


> Would be a mapping
> 
> ```
> #define DRIVER_PARAM_<NAME> (CONFIG_DRIVER_PARAM_<NAME>`
> ```
> 
> OK to keep the compatibility with existing documentation where the default parameters could be changed by overriding them by `CFLAGS` at the make command line?

When Kconfig is not used (it does not run by default) the `CONFIG_DRIVER_PARAM_<NAME>` value can also be injected via `CFLAGS`, so I think just a documentation update is enough, no mapping should be needed if I understood you correctly.

gschorcht: > What about the following:
> ```
> ...
> config DRIVER_MODE_1_VAL
>    hex
>    default 0x10
> ...
> ```

Hm, that would require coding enumerations again, which I think is error prone. In my approach I can use the named constants as they are already defined by the driver.

> Personally I think many of those parameters are more a hardware configuration rather than a configuration of the driver itself, and would be better off left for other tools (e.g. the current `driver_params.h` or Device Tree).

I'm only talking about selections the user has to make. A typical example would be a sensor for which the desired output data rate has to be defined. The output data rate is configured using a register and all possible values are defined as an enumeration. For efficiency, the enumeration directly encodes the register values.

gschorcht:
Since the review of the PR hasn't really started yet and I've changed a lot, I felt free to restructure the PR and force-push it.

mguetschow:
Does anyone of the active @RIOT-OS/maintainers have hardware and time to test this? Seems like a worthwhile PR to me.

chrysn:
Is there any supported board on which this is present?

crasbe:
> Does anyone of the active @RIOT-OS/maintainers have hardware and time to test this? Seems like a worthwhile PR to me.

I am not a maintainer, but I can test it in the near future (once the Mouser order arrives). Mikro-E makes breakout boards for both chips that are covered by this PR.
https://www.mouser.de/ProductDetail/932-MIKROE-951
https://www.mouser.de/ProductDetail/932-MIKROE-1838

Given that this is a driver, the potential impact on the rest of RIOT should be low.
This IO expander should work well in my binary clock. Using RIOT for it wasn't the plan, but why not 🤣

tests/driver_mcp23x17/Makefile
```diff
@@ -0,0 +1,10 @@
+include ../Makefile.tests_common
```
crasbe, line 1:
```suggestion
include ../Makefile.drivers_common
```

The test have been restructured in the meantime and now this is the common Makefile for all tests of drivers.

crasbe:
So... I'm not sure if we can reach @gschorcht, but the main thing that conflicts with a rebase to the current master branch is the location of the test in `tests/drivers_mcp23x17` instead of the current folder structure `tests/drivers/mcp23x17`. (Well, the rebase doesn't care, but it doesn't make sense to merge it like that).

The `drivers/Kconfig` and `drivers/Makefile.dep` files couldn't be merged automatically, but that is just because some stuff from the `Kconfig` file was deleted in the meanwhile and git incorrectly detected the place where to put the additional lines in the `Makefile.dep`.
But the Kconfig stuff is obsolete now anyways, so it could be deleted entirely.

The only change that has to be made to compile the code is in the suggestion above.

<hr/>
So far I did some very basic tests with both chips and they generally seem to work. I can run the benchmark tests, configure and read/write the GPIOs. I'll take a deeper look at everything and give it a more thorough test, but for now it's good that it can be rebased and compiles with the current master.

Teufelchen1:
Nice, CI is almost happy. 
Are you happy with your testing @crasbe? If so, do we have a chance of merging it this year? Super awesome to push through with a five year old PR! ❤️

crasbe:
> Are you happy with your testing @crasbe? If so, do we have a chance of merging it this year? Super awesome to push through with a five year old PR! ❤️

As soon as I have access to my hardware again, I'll give it a test (next week).
We were very busy putting our hardware (with RIOT btw) into space 🚀

drivers/mcp23x17/mcp23x17.c
```diff
@@ -0,0 +1,620 @@
...
+
+#if IS_USED(MODULE_MCP23X17_IRQ)
+#include "event/thread.h"
+#endif
+
+#define ENABLE_DEBUG 0
+#include "debug.h"
+
+#if ENABLE_DEBUG
+
```
crasbe, line 33:
```suggestion
#include <string.h>
```

The function `strerror` that is used in some of the DEBUG statements needs the `string.h` header.

crasbe:
I tested the I2C variant successfully with a NUCLEO-L073RZ board. Pin 0, 4 is A2 on the Nucleo and Pin 0,6 is D12.
In the original test instructions, there are two small errors (which are probably from the older age of the original desription, some folder structure changed in the meantime):

Instead of "\(GPIO\(0,6\)\)" it should be "\GPIO_PIN\(0,6\)\)" and instead of "tests/driver_mcp23x17" it should be "tests/drivers/mcp23x17"
```
CFLAGS="-DMCP23X17_PARAM_I2C_INT=\(GPIO_PIN\(0,6\)\)" \
USEMODULE="mcp23x17_i2c mcp23x17_irq"  flash -C tests/drivers/mcp23x17 BOARD=...
```

But since that's not part of the official documentation, it probably doesn't have to be fixed.

<hr>

For SPI, I used Pin 0,9 which is D8 on the Nucleo (D12 is MISO, so it's occupied) which worked as well (after I discovered I had to set the DRIVER variable, which is not mentioned in the Testing Procedure 😅 ).

<hr>

Even though it is not for me to decide, I'd say this is good to be merged. One thing I noticed is that in the DEBUG_DEV statements, that access the `pin` variable, the format string is unhappy. This is due to the fact that `gpio_t` has wildly different data types depending on the architecture (from uint8_t on AVRs to uint32_t on STM32). I'm not sure how to get around that without a lot of `#ifdef` statements or similar. But that's not in the scope of this PR.

Teufelchen1:
Nice! Thanks a lot! @gschorcht would you address the nitpicks and squash?

gschorcht:
> In the original test instructions, there are two small errors (which are probably from the older age of the original desription, some folder structure changed in the meantime):
> ...
> which worked as well (after I discovered I had to set the DRIVER variable, which is not mentioned in the Testing Procedure 😅 ).

@crasbe Sorry for the inconvinience but the PR is alsmost 6 years old and never changed the testing instructions.

crasbe:
@gschorcht It's all good, I should've just read the README, which is up to date. I am very thankful that you're still pushing this through (especially considering it's that old) ❤️

gschorcht:
@crasbe Should I squash it?

crasbe:
Yes, you can. 👍

Gunar Schorcht ***@***.***> schrieb am Fr., 22. Nov. 2024,
06:31:

crasbe (approved):

drivers/mcp23x17/mcp23x17.c
```diff
@@ -0,0 +1,622 @@
...
+#if IS_USED(MODULE_MCP23X17_IRQ)
+
+#if IS_USED(MODULE_MCP23X17_IRQ_MEDIUM)
+#define  MCP23X17_EVENT_PRIO    EVENT_PRIO_MEDIUM
+#elif IS_USED(MODULE_MCP23X17_IRQ_HIGHEST)
+#define  MCP23X17_EVENT_PRIO    EVENT_PRIO_HIGHEST
+#endif
+
+/* interrutp service routine for IRQs */
```
crasbe, line 66:
```suggestion
/* interrupt service routine for IRQs */
```

This is the typo the static tests were complaining about. Odd that the workflow didn't catch it before?

crasbe:
So basically there are three issues that let the CI fail:
1) The `cc2538` CPU not supporting `GPIO_OD` and `GPIO_OD_PU`:
https://github.com/RIOT-OS/RIOT/blob/04a169867e530e0a1113540c3db321c1b8d95002/cpu/cc2538/include/periph_cpu.h#L149-L160

In this case, `GPIO_OD` and `GPIO_OD_PU` are both `0xff`, which leads to the issue in the switch case.
Some other CPUs don't support `GPIO_OD_PU`, but that is not an issue because the mode is never `0xff`.

I'm not sure yet what's the best way to solve this. The `gpio_init` function checks for `0xff`. It would probably be possible to set `GPIO_OD` to a second value like `0xfe` and change the condition to `if (mode >= MODE_NOTSUP)` with `#define MODE_NOTSUP (0xf0)`
https://github.com/RIOT-OS/RIOT/blob/04a169867e530e0a1113540c3db321c1b8d95002/cpu/cc2538/periph/gpio.c#L100-L105

`0xf0` is still very far away from the valid configuration values:
https://github.com/RIOT-OS/RIOT/blob/04a169867e530e0a1113540c3db321c1b8d95002/cpu/cc2538/include/cc2538_gpio.h#L304-L313

<hr/>

2) The atmega8 doesn't have enough memory. This is easily solved by adding a `tests/drivers/mcp23x17/Makefile.ci` with the following content:
```makefile
BOARD_INSUFFICIENT_MEMORY := \
    atmega8 \
    #
```

<hr/>

3) The `kintis` CPU used by the `mulle` board and others include the `unistd.h` file, which is taken from `newlib` and implements the function `read`. This gives a conflict with the built-in `read` function.

The easiest solution would probably be to rename the internal `read` function to `_read` in line 340 and 485.

<hr/>

I tested the changes locally and they work.
@gschorcht if you want to, I can create a PR for the first point.

Enoch247:
FYI, you can use `make -C tests/drivers/mcp23x17 generate-Makefile.ci` to auto generate the `Makefile.ci`.

tests/drivers/mcp23x17/Makefile.ci
```diff
@@ -0,0 +1,8 @@
+BOARD_INSUFFICIENT_MEMORY := \
+    arduino-duemilanove \
+    arduino-leonardo \
+    arduino-nano \
+    arduino-uno \
+    atmega328p \
+    atmega328p-xplained-mini \
+    #
```
crasbe, line 8:
```suggestion
BOARD_INSUFFICIENT_MEMORY := \
    arduino-duemilanove \
    arduino-nano \
    arduino-uno \
    atmega328p \
    atmega328p-xplained-mini \
    atmega8 \
    #
```

drivers/mcp23x17/mcp23x17.c
crasbe: This typo is still here 👀

crasbe (approved):
Please address the one "interrutp" typo. You can squash the changes directly.

If CI is happy, this can be merged :)

drivers/mcp23x17/mcp23x17.c
gschorcht: :sunglasses:

gschorcht:
@crasbe Thanks for reviewing and merging.

PR #10689, opened 2019-01-02
boards/stm32: remove unnecessary ADC defines in board config

aabadie:
### Contribution description

This PR cleans up the ST based boards configuration. While having another look at #8939, I found that there were still a lot of them used while this is not necessary since the features is already not provided by the board.

### Testing procedure

If Murdock is green, all should be ok.

### Issues/PRs references

None


aabadie:
Thanks for reviewing and merging @MrKevinWeiss !

PR #10691, opened 2019-01-02
boards/stm32: remove useless empty conf defines

aabadie:
### Contribution description

Another cleanup of useless defines in STM32 based boards configuration files. Those ones are leftover from #10689 but also applied to RTC, DAC, etc. Found while working on the different factorizations of I2C config.

### Testing procedure

Murdock's happiness should be enough.

### Issues/PRs references

Follow-up of #10689

### DIFF
--- a/boards/lobaro-lorabox/include/periph_conf.h
+++ b/boards/lobaro-lorabox/include/periph_conf.h
@@ -210,20 +210,6 @@ static const i2c_conf_t i2c_config[] = {
 #define I2C_NUMOF           (sizeof(i2c_config) / sizeof(i2c_config[0]))
 /** @} */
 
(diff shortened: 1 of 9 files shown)


MrKevinWeiss:
Just doing a bit of thinking here... I don't think it makes a difference for the code size and this way it doesn't explicitly state anything in the documentation.  Do you think that is the right way to go (less lines of code vs doc saying there are 0 adcs or 0 rtcs)?

MrKevinWeiss:
After talking with @jia200x I think the parameters should be removed since they would not be in the features provided anyways.

MrKevinWeiss:
Just want to rerun murdock.

PR #10698, opened 2019-01-03
tests/periph_uart: fix baudrate truncation

### LINKED ISSUE #10517: Arduino UART cannot set 16 bit baudrate
#### Description
The HiL CI caught a bug in the arduino-mega2560 but I imagine it is everywhere.  The baudrate seems to get truncated if it is set higher than 32767.  I imagine it is a type problem and will look into it later.
Initialize the baudrate on the arduino-mega2560 to anything greater that 32767 and measure.

#### Expected results
It works or throws an error saying it cannot be set to that speed.

#### Actual results
It initialized and tries to send at the incorrect baud rate (115200 was measured at 250)

MrKevinWeiss:
### Contribution description
Since some boards an int is 16 bits the atoi truncates values.
This commit using a long instead of an int.

### Testing procedure
On any board where an int is 16 bits (such as the arduino-mega2560
run `BOARD=arduino-mega2560 make flash term -C tests/periph_uart/`
initialize with 115200 and try sending to something else with that same baudrate (loopback shouldn't work) or scoping it.

### Issues/PRs references
Fixes #10517

### DIFF
--- a/tests/periph_uart/main.c
+++ b/tests/periph_uart/main.c
@@ -133,7 +133,7 @@ static int cmd_init(int argc, char **argv)
     if (dev < 0) {
         return 1;
     }
-    baud = atoi(argv[2]);
+    baud = strtol(argv[2], NULL, 0);
 
     /* initialize UART */
     res = uart_init(UART_DEV(dev), baud, rx_cb, (void *)dev);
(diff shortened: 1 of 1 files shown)


aabadie:
Tested and the correct baudrate is set. Maybe [this line](https://github.com/RIOT-OS/RIOT/pull/10698/files#diff-17d73fb4c775ea0c8dbf4a19bedba419R148) could also be changed like this:
```c
printf("Success: Successfully initialized UART_DEV(%i) at BAUDRATE %"PRIu32"\n",
       dev, baud);
```
Using this makes the problem more obvious.

MrKevinWeiss:
@aabadie done with slight alteration so it fits on one line!

PR #10702, opened 2019-01-05
net/nanocoap: write struct-based block1 option

kb2ma:
### Contribution description
Adds a function to write a block1 option using the struct-based nanocoap API.

### Testing procedure
See the [gcoap-block](https://github.com/kb2ma/applications/tree/kb2ma-master/gcoap-block) server app. It is equivalent to the nanocoap_server example, and includes an `/sha256` resource. The handler writes a block1 option in the responses. Note that the RIOTBASE variable in `Makefile` does not use the RIOT submodule.

For a more complete test, see pytest-based [block1_test](https://github.com/kb2ma/riot-coap-pytest/blob/master/block1_test.py).

### Issues/PRs references
Part of the implementation of #9309.

### DIFF
--- a/sys/include/net/nanocoap.h
+++ b/sys/include/net/nanocoap.h
@@ -645,6 +645,20 @@ ssize_t coap_opt_finish(coap_pkt_t *pkt, uint16_t flags);
 size_t coap_opt_put_block2(uint8_t *buf, uint16_t lastonum, coap_block_slicer_t *slicer, bool more);
 
 /**
+ * @brief   Add the given block1 option into pkt
+ *
+ * @post pkt.payload advanced to first byte after option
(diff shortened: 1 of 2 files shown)


kb2ma:
This PR unfortunately was premature. My original intention was to allow writing a block1 option in control usage using the struct-based API, similar to coap_put_block1_ok(). It seemed an easy thing to do, and it was.

However, my larger goal is to complete the block implementation, which means addition of the client side for block1 and block2. I just started to look at this after I posted this PR. In this case, I think that the function coap_opt_add_block1() added in this PR instead will be used to write a message with the content, which uses a coap_block_slicer_t.

So, closing for now. My next step is to open an issue to describe the function names for the complete block implementation. Then I can revisit/rework this PR in that larger context.

PR #10704, opened 2019-01-05
drivers/bmx055: eliminate some casts

keestux:
First, every read function converts the void pointer into a const bmx055_t

Second, it is cleaner (more obvious) to pass phydat_t.val[3] instead of
type casting to int16_t* and relying that .val[3] is the first struct
member of phydat_t.

Notice that it wasn't tested but I'm fairly sure that the code behavior hasn't changed.

### DIFF
--- a/drivers/bmx055/bmx055_saul.c
+++ b/drivers/bmx055/bmx055_saul.c
@@ -24,7 +24,9 @@
 
 static int read_mag(const void *dev, phydat_t *res)
 {
-    if (bmx055_mag_read((bmx055_t *)dev, (int16_t *)res) != BMX055_OK) {
+    const bmx055_t *mydev = (const bmx055_t *)dev;
+
+    if (bmx055_mag_read(mydev, res->val) != BMX055_OK) {
         return 0;
     }
     res->unit = UNIT_GS;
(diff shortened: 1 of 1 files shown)


drivers/bmx055/bmx055_saul.c
```diff
@@ -24,7 +24,9 @@
 
 static int read_mag(const void *dev, phydat_t *res)
 {
-    if (bmx055_mag_read((bmx055_t *)dev, (int16_t *)res) != BMX055_OK) {
+    const bmx055_t *mydev = (const bmx055_t *)dev;
+
+    if (bmx055_mag_read(mydev, res->val) != BMX055_OK) {
```
smlng: to me it would be even better to use `&res->val[0]` as you did in https://github.com/RIOT-OS/RIOT/pull/10711, that makes it explicit we want a single value, compare to gyros and mags that actually expect a triplet of `int16_t`s

keestux:
Ok, I'll verify that and update the commitOn Jan 7, 2019 09:55, Sebastian Meiling <notifications@github.com> wrote:@smlng requested changes on this pull request.

minor, otherwise (as before) ACK

In drivers/bmx055/bmx055_saul.c:
> @@ -24,7 +24,9 @@
 
 static int read_mag(const void *dev, phydat_t *res)
 {
-    if (bmx055_mag_read((bmx055_t *)dev, (int16_t *)res) != BMX055_OK) {
+    const bmx055_t *mydev = (const bmx055_t *)dev;
+
+    if (bmx055_mag_read(mydev, res->val) != BMX055_OK) {

to me it would be even better to use &res->val[0] as you did in #10711, that makes it explicit we want a single value, compare to gyros and mags that actually expect a triplet of int16_ts

drivers/bmx055/bmx055_saul.c
keestux: In this case `bmx055_mag_read` actually fills in all three elements of `res->val`
So let's keep it that way.

smlng: ah damn, right, leave as is ... my bad (eyes)

smlng: sorry for the noise

PR #10709, opened 2019-01-06
drivers/ds18: refactor functions with const dev pointer

keestux:
All DS18 functions have a dev argument. All except the init function use
it as an IN parameter, so we can prototype it as const ds18_t*.

As a consequence we can drop the cast in read_temperature() in ds18_saul.c
which was the primary trigger for the changes.

The commit also follows the preferred convention that "params" is a field
in the device struct. Only the init function needs to write it.

Notice, this has not been tested because of lack of hardware for it. Also the SAUL
part needs to be tested.

### DIFF
--- a/drivers/include/ds18.h
+++ b/drivers/include/ds18.h
@@ -54,21 +54,20 @@ extern "C" {
 
 
 /**
- * @brief   Device descriptor for a ds18 device
- */
-typedef struct {
-    gpio_t pin;             /**< Pin the sensor is connected to */
-    gpio_mode_t out_mode;  /**< Pin output mode */
-    gpio_mode_t in_mode;    /**< Pin input mode */
-} ds18_t;
-
-/**
  * @brief Device initialization parameters
  */
 typedef struct {
     gpio_t pin;             /**< Pin the sensor is connected to */
-    gpio_mode_t out_mode;    /**< Pin output mode */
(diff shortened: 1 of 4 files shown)


drivers/include/ds18.h
```diff
@@ -54,21 +54,20 @@ extern "C" {
 
 
 /**
- * @brief   Device descriptor for a ds18 device
+ * @brief Device initialization parameters
```
smlng: minor nit (I know it was wrong before, too): typically we have tab-alignment/spacing between `@brief` and the description/heading (see below for the `ds18_t`).

Please amend that directly, THX!

danpetry:
> The commit also follows the preferred convention that "params" is a field
in the device struct. Only the init function needs to write it.

Just a note - copying params into dev during init saves a derefencing step during runtime. I.e. it's a small optimization

keestux:
@danpetry You underestimate what compilers can do these days. And in this, there is no rocket science. There is no difference, because there is no extra pointer involved.

drivers/include/ds18.h
keestux: @smlng Do you want me to do the other @brief's in this file as well?

smlng: ah, I see lots of inconsistencies there ... looks like we need a beautifier tool for that, too.

So leave as is, I'll withdraw my change commit - this should be done automatically by some tool

PR #10710, opened 2019-01-06
drivers: add Differentially Operated Serial Ethernet (DOSE) driver

jue89:
### Contribution description

This contribution adds a new driver for communicating Ethernet over (electrical) CAN. The goal is to create an easy and cheap way to interconnect several RIOT nodes using a shared bus wire.

Technically, a CAN transceiver chip is connected to the UART (!) of a RIOT device. Ethernet frames are sent onto the CAN bus using `uart_write()` while observing the received echo from the bus. This way collisions are detected (received echo != transmitted octet) and retransmissions are scheduled. The frames are appended with a CRC16 to protect the system from transmission errors.

Additional to the two UART pins, another GPIO is connected to the RX pin to sense start bits on the bus and detect that another node just started sending an Ethernet frame. Just waiting for the UART interrupt would add a delay, since started transmissions are detected after the first octet has been fully received. This would increase the probability of collisions.

Since this driver is utilising the UART and GPIO driver to interface with the underlaying hardware, every 
board offering a UART and GPIO driver should be capable of utilising this driver.

Please let me know if this kind of driver is welcome in the RIOT codebase. If the maintainers like the idea, I am willing to create some benchmarks how good this driver performs.

I created a finite state machine diagram that hopefully helps to understand what's going on:

ethocan-state

 * `BLOCKED`: The driver waits for incoming frames but blocks send requests of the underlaying netif.
 * `IDLE`: The driver waits for incoming frames and is also able to send frames to the bus.
 * `RECV`: The driver receives a frame from a remote node.
 * `SEND`: The driver sends a frame to the bus.

### Testing procedure

I used a slightly modified version of examples/gnrc_networking and patched its Makefile:
```diff
--- a/examples/gnrc_networking/Makefile
+++ b/examples/gnrc_networking/Makefile
@@ -18,7 +18,9 @@ BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-mega2560 arduino-uno \
 
 # Include packages that pull up and auto-init the link layer.
 # NOTE: 6LoWPAN will be included if IEEE802.15.4 devices are present
-USEMODULE += gnrc_netdev_default
+USEMODULE += dose
+CFLAGS += '-DDOSE_PARAM_UART=UART_DEV(2)'
+CFLAGS += '-DDOSE_PARAM_SENSE_PIN=GPIO_PIN(PORT_B, 1)'
 USEMODULE += auto_init_gnrc_netif
 # Activate ICMPv6 error messages
 USEMODULE += gnrc_icmpv6_error
```

I used the board "Bluepill" to test everything: `make BOARD=bluepill-128kib all`

testbed

(Any other 3.3V CAN transceiver should do the job, as well.)

After flashing, connect the Bluepill using a RS232 adapter to UART 2 (Pin A3 and A2) and execute: `ping6 ff02::1`.

### DIFF
--- a/drivers/Makefile.dep
+++ b/drivers/Makefile.dep
@@ -193,6 +193,15 @@ ifneq (,$(filter ethos,$(USEMODULE)))
   USEMODULE += tsrb
 endif
 
+ifneq (,$(filter dose,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_uart
+  USEMODULE += iolist
+  USEMODULE += netdev_eth
+  USEMODULE += random
+  USEMODULE += xtimer
+  FEATURES_REQUIRED += periph_gpio_irq
+endif
+
 ifneq (,$(filter feetech,$(USEMODULE)))
   USEMODULE += uart_half_duplex
 endif
--- a/drivers/dose/dose.c
+++ b/drivers/dose/dose.c
@@ -1,0 +1,567 @@
+/*
+ * Copyright (C) 2019 Juergen Fitschen <me@jue.yt>
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     drivers_dose
+ * @{
+ *
+ * @file
+ * @brief       Implementation of the Differentially Operated Serial Ethernet driver
+ *
+ * @author      Juergen Fitschen <me@jue.yt>
+ *
+ * @}
+ */
+
+#include <string.h>
+
+#include "dose.h"
+#include "luid.h"
+#include "random.h"
+#include "irq.h"
+
+#include "net/netdev/eth.h"
+#include "net/eui64.h"
+
+#define ENABLE_DEBUG (0)
+#include "debug.h"
+
+static uint16_t crc16_update(uint16_t crc, uint8_t octet);
+static dose_signal_t state_transit_blocked(dose_t *ctx, dose_signal_t signal);
+static dose_signal_t state_transit_idle(dose_t *ctx, dose_signal_t signal);
+static dose_signal_t state_transit_recv(dose_t *ctx, dose_signal_t signal);
+static dose_signal_t state_transit_send(dose_t *ctx, dose_signal_t signal);
+static void state(dose_t *ctx, uint8_t src);
+static void _isr_uart(void *arg, uint8_t c);
+static void _isr_gpio(void *arg);
+static void _isr_xtimer(void *arg);
+static void clear_recv_buf(dose_t *ctx);
+static void _isr(netdev_t *netdev);
+static int _recv(netdev_t *dev, void *buf, size_t len, void *info);
+static uint8_t wait_for_state(dose_t *ctx, uint8_t state);
+static int send_octet(dose_t *ctx, uint8_t c);
+static int _send(netdev_t *dev, const iolist_t *iolist);
+static int _get(netdev_t *dev, netopt_t opt, void *value, size_t max_len);
+static int _set(netdev_t *dev, netopt_t opt, const void *value, size_t len);
+static int _init(netdev_t *dev);
+void dose_setup(dose_t *ctx, const dose_params_t *params);
+
+static uint16_t crc16_update(uint16_t crc, uint8_t octet)
+{
+    crc = (uint8_t)(crc >> 8) | (crc << 8);
+    crc ^= octet;
+    crc ^= (uint8_t)(crc & 0xff) >> 4;
+    crc ^= (crc << 8) << 4;
+    crc ^= ((crc & 0xff) << 4) << 1;
+    return crc;
+}
+
+static dose_signal_t state_transit_blocked(dose_t *ctx, dose_signal_t signal)
+{
+    uint32_t backoff;
+    (void) signal;
+
+    if (ctx->state == DOSE_STATE_RECV) {
+        /* We got here from RECV state. The driver's thread has to look
+         * if this frame should be processed. By queuing NETDEV_EVENT_ISR,
+         * the netif thread will call _isr at some time. */
+        SETBIT(ctx->flags, DOSE_FLAG_RECV_BUF_DIRTY);
+        ctx->netdev.event_callback((netdev_t *) ctx, NETDEV_EVENT_ISR);
+    }
+
+    /* Enable GPIO interrupt for start bit sensing */
+    gpio_irq_enable(ctx->sense_pin);
+
+    /* The timeout will bring us back into IDLE state by a random time.
+     * If we entered this state from RECV state, the random time lays
+     * in the interval [1 * timeout, 2 * timeout]. If we came from
+     * SEND state, a time in the interval [2 * timeout, 3 * timeout]
+     * will be picked. This ensures that responding nodes get preferred
+     * bus access and sending nodes do not overwhelm listening nodes. */
+    if (ctx->state == DOSE_STATE_SEND) {
+        backoff = random_uint32_range(2 * ctx->timeout_base, 3 * ctx->timeout_base);
+    }
+    else {
+        backoff = random_uint32_range(1 * ctx->timeout_base, 2 * ctx->timeout_base);
+    }
+    xtimer_set(&ctx->timeout, backoff);
+
+    return DOSE_SIGNAL_NONE;
+}
+
+static dose_signal_t state_transit_idle(dose_t *ctx, dose_signal_t signal)
+{
+    (void) ctx;
+    (void) signal;
+
+    return DOSE_SIGNAL_NONE;
+}
+
+static dose_signal_t state_transit_recv(dose_t *ctx, dose_signal_t signal)
+{
+    dose_signal_t rc = DOSE_SIGNAL_NONE;
+
+    if (ctx->state != DOSE_STATE_RECV) {
+        /* We freshly entered this state. Thus, no start bit sensing is required
+         * anymore. Disable GPIO IRQs during the transmission. */
+        gpio_irq_disable(ctx->sense_pin);
+    }
+
+    if (signal == DOSE_SIGNAL_UART) {
+        /* We received a new octet */
+        int esc = (ctx->flags & DOSE_FLAG_ESC_RECEIVED);
+        if (!esc && ctx->uart_octet == DOSE_OCTECT_ESC) {
+            SETBIT(ctx->flags, DOSE_FLAG_ESC_RECEIVED);
+        }
+        else if (!esc && ctx->uart_octet == DOSE_OCTECT_END) {
+            SETBIT(ctx->flags, DOSE_FLAG_END_RECEIVED);
+            rc = DOSE_SIGNAL_END;
+        }
+        else {
+            if (esc) {
+                CLRBIT(ctx->flags, DOSE_FLAG_ESC_RECEIVED);
+            }
+            /* Since the dirty flag is set after the RECV state is left,
+             * it indicates that the receive buffer contains unprocessed data
+             * from a previously received frame. Thus, we just ignore new data. */
+            if (!(ctx->flags & DOSE_FLAG_RECV_BUF_DIRTY)
+                && ctx->recv_buf_ptr < DOSE_FRAME_LEN) {
+                ctx->recv_buf[ctx->recv_buf_ptr++] = ctx->uart_octet;
+            }
+        }
+    }
+
+    if (rc == DOSE_SIGNAL_NONE) {
+        /* No signal is returned. We stay in the RECV state. */
+        xtimer_set(&ctx->timeout, ctx->timeout_base);
+    }
+
+    return rc;
+}
+
+static dose_signal_t state_transit_send(dose_t *ctx, dose_signal_t signal)
+{
+    (void) signal;
+
+    if (ctx->state != DOSE_STATE_SEND) {
+        /* Disable GPIO IRQs during the transmission. */
+        gpio_irq_disable(ctx->sense_pin);
+    }
+
+    /* Don't trace any END octets ... the timeout or the END signal
+     * will bring us back to the BLOCKED state after _send has emitted
+     * its last octet. */
+
+    xtimer_set(&ctx->timeout, ctx->timeout_base);
+
+    return DOSE_SIGNAL_NONE;
+}
+
+static void state(dose_t *ctx, dose_signal_t signal)
+{
+    /* Make sure no other thread or ISR interrupts state transitions */
+    unsigned irq_state = irq_disable();
+
+    do {
+        /* The edges of the finite state machine can be identified by
+         * the current state and the signal that caused a state transition.
+         * Since the state only occupies the first 4 bits and the signal the
+         * last 4 bits of a uint8_t, they can be added together and hence
+         * be checked together. */
+        switch (ctx->state + signal) {
+            case DOSE_STATE_INIT + DOSE_SIGNAL_INIT:
+            case DOSE_STATE_RECV + DOSE_SIGNAL_END:
+            case DOSE_STATE_RECV + DOSE_SIGNAL_XTIMER:
+            case DOSE_STATE_SEND + DOSE_SIGNAL_END:
+            case DOSE_STATE_SEND + DOSE_SIGNAL_XTIMER:
+                signal = state_transit_blocked(ctx, signal);
+                ctx->state = DOSE_STATE_BLOCKED;
+                break;
+
+            case DOSE_STATE_BLOCKED + DOSE_SIGNAL_XTIMER:
+                signal = state_transit_idle(ctx, signal);
+                ctx->state = DOSE_STATE_IDLE;
+                break;
+
+            case DOSE_STATE_IDLE + DOSE_SIGNAL_GPIO:
+            case DOSE_STATE_IDLE + DOSE_SIGNAL_UART:
+            case DOSE_STATE_BLOCKED + DOSE_SIGNAL_GPIO:
+            case DOSE_STATE_BLOCKED + DOSE_SIGNAL_UART:
+            case DOSE_STATE_RECV + DOSE_SIGNAL_UART:
+                signal = state_transit_recv(ctx, signal);
+                ctx->state = DOSE_STATE_RECV;
+                break;
+
+            case DOSE_STATE_IDLE + DOSE_SIGNAL_SEND:
+            case DOSE_STATE_SEND + DOSE_SIGNAL_UART:
+                signal = state_transit_send(ctx, signal);
+                ctx->state = DOSE_STATE_SEND;
+                break;
+
+            default:
+                DEBUG("dose state(): unexpected state transition (STATE=0x%02d SIGNAL=0x%02d)\n", ctx->state, signal);
+                signal = DOSE_SIGNAL_NONE;
+        }
+    } while (signal != DOSE_SIGNAL_NONE);
+
+    /* Indicate state change by unlocking state mutex */
+    mutex_unlock(&ctx->state_mtx);
+    irq_restore(irq_state);
+}
+
+static void _isr_uart(void *arg, uint8_t c)
+{
+    dose_t *dev = (dose_t *) arg;
+
+    dev->uart_octet = c;
+    state(dev, DOSE_SIGNAL_UART);
+}
+
+static void _isr_gpio(void *arg)
+{
+    dose_t *dev = (dose_t *) arg;
+
+    state(dev, DOSE_SIGNAL_GPIO);
+}
+
+static void _isr_xtimer(void *arg)
+{
+    dose_t *dev = (dose_t *) arg;
+
+    state(dev, DOSE_SIGNAL_XTIMER);
+}
+
+static void clear_recv_buf(dose_t *ctx)
+{
+    unsigned irq_state = irq_disable();
+
+    ctx->recv_buf_ptr = 0;
+    CLRBIT(ctx->flags, DOSE_FLAG_RECV_BUF_DIRTY);
+    CLRBIT(ctx->flags, DOSE_FLAG_END_RECEIVED);
+    CLRBIT(ctx->flags, DOSE_FLAG_ESC_RECEIVED);
+    irq_restore(irq_state);
+}
+
+static void _isr(netdev_t *netdev)
+{
+    dose_t *ctx = (dose_t *) netdev;
+    unsigned irq_state;
+    int dirty, end;
+
+    /* Get current flags atomically */
+    irq_state = irq_disable();
+    dirty = (ctx->flags & DOSE_FLAG_RECV_BUF_DIRTY);
+    end = (ctx->flags & DOSE_FLAG_END_RECEIVED);
+    irq_restore(irq_state);
+
+    /* If the receive buffer does not contain any data just abort ... */
+    if (!dirty) {
+        DEBUG("dose _isr(): no frame -> drop\n");
+        return;
+    }
+
+    /* If we haven't received a valid END octet just drop the incomplete frame. */
+    if (!end) {
+        DEBUG("dose _isr(): incomplete frame -> drop\n");
+        clear_recv_buf(ctx);
+        return;
+    }
+
+    /* The set dirty flag prevents recv_buf or recv_buf_ptr from being
+     * touched in ISR context. Thus, it is safe to work with them without
+     * IRQs being disabled or mutexes being locked. */
+
+    /* Check for minimum length of an Ethernet packet */
+    if (ctx->recv_buf_ptr < sizeof(ethernet_hdr_t) + DOSE_FRAME_CRC_LEN) {
+        DEBUG("dose _isr(): frame too short -> drop\n");
+        clear_recv_buf(ctx);
+        return;
+    }
+
+    /* Check the dst mac addr if the iface is not in promiscuous mode */
+    if (!(ctx->opts & DOSE_OPT_PROMISCUOUS)) {
+        ethernet_hdr_t *hdr = (ethernet_hdr_t *) ctx->recv_buf;
+        if ((hdr->dst[0] & 0x1) == 0 && memcmp(hdr->dst, ctx->mac_addr.uint8, ETHERNET_ADDR_LEN) != 0) {
+            DEBUG("dose _isr(): dst mac not matching -> drop\n");
+            clear_recv_buf(ctx);
+            return;
+        }
+    }
+
+    /* Check the CRC */
+    uint16_t crc = 0xffff;
+    for (size_t i = 0; i < ctx->recv_buf_ptr; i++) {
+        crc = crc16_update(crc, ctx->recv_buf[i]);
+    }
+    if (crc != 0x0000) {
+        DEBUG("dose _isr(): wrong crc 0x%04x -> drop\n", crc);
+        clear_recv_buf(ctx);
+        return;
+    }
+
+    /* Finally schedule a _recv method call */
+    DEBUG("dose _isr(): NETDEV_EVENT_RX_COMPLETE\n");
+    ctx->netdev.event_callback((netdev_t *) ctx, NETDEV_EVENT_RX_COMPLETE);
+}
+
+static int _recv(netdev_t *dev, void *buf, size_t len, void *info)
+{
+    dose_t *ctx = (dose_t *) dev;
+
+    (void)info;
+
+    size_t pktlen = ctx->recv_buf_ptr - DOSE_FRAME_CRC_LEN;
+    if (!buf && !len) {
+        /* Return the amount of received bytes */
+        return pktlen;
+    }
+    else if (!buf && len) {
+        /* The user drops the packet */
+        clear_recv_buf(ctx);
+        return pktlen;
+    }
+    else if (len < pktlen) {
+        /* The provided buffer is too small! */
+        DEBUG("dose _recv(): receive buffer too small\n");
+        clear_recv_buf(ctx);
+        return -1;
+    }
+    else {
+        /* Copy the packet to the provided buffer. */
+        memcpy(buf, ctx->recv_buf, pktlen);
+        clear_recv_buf(ctx);
+        return pktlen;
+    }
+}
+
+static uint8_t wait_for_state(dose_t *ctx, uint8_t state)
+{
+    do {
+        /* This mutex is unlocked by the state machine
+         * after every state transition */
+        mutex_lock(&ctx->state_mtx);
+    } while (state != DOSE_STATE_ANY && ctx->state != state);
+    return ctx->state;
+}
+
+static int send_octet(dose_t *ctx, uint8_t c)
+{
+    uart_write(ctx->uart, (uint8_t *) &c, sizeof(c));
+
+    /* Wait for a state transition */
+    uint8_t state = wait_for_state(ctx, DOSE_STATE_ANY);
+    if (state != DOSE_STATE_SEND) {
+        /* Timeout */
+        DEBUG("dose send_octet(): timeout\n");
+        return -2;
+    }
+    else if (ctx->uart_octet != c) {
+        /* Mismatch */
+        DEBUG("dose send_octet(): mismatch\n");
+        return -1;
+    }
+
+    return 0;
+}
+
+static int send_data_octet(dose_t *ctx, uint8_t c)
+{
+    int rc;
+
+    /* Escape special octets */
+    if (c == DOSE_OCTECT_ESC || c == DOSE_OCTECT_END) {
+        rc = send_octet(ctx, DOSE_OCTECT_ESC);
+        if (rc) {
+            return rc;
+        }
+    }
+
+    /* Send data octet */
+    rc = send_octet(ctx, c);
+
+    return rc;
+}
+
+static int _send(netdev_t *dev, const iolist_t *iolist)
+{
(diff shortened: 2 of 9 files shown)


jue89:
(I'm sorry ... I was too quick and forgot the labels. I would suggest: `Area: network`, `Discussion: RFC`, `State: WIP`, 'Type: new feature`)

smlng:
this looks (very) interesting, nice work!

jue89:
As you can see above, I already fixed some bugs:
1) The transition from state `BLOCKED` to `IDLE` after a random timeout does not occure, if the selected timeout is close to zero. So I changed the interval from [0, MAX_TIMEOUT] to [0.1*MAX_TIMEOUT, MAX_TIMEOUT].
1) If the frame's CRC happens to be 0xFE or 0xFF - our special octets for ESC and END - they weren't escaped. Fixed that ...
1) When we enter SEND state, there is a small chance that we get interrupted by a start bit of an incoming frame. Thus, we now check if we actually entered SEND and if we didn't, we will try again.
1) I think, I revealed the source of my auto initialization code ;)

jue89:
> this looks (very) interesting, nice work!

Thank you!

---

I just did some ping tests:

```
> ping6 10000 ff02::1 0 0
8 bytes from fe80::89:eff:fe7d:66cd: id=85 seq=1 hop limit=64 time = 16.534 ms
8 bytes from fe80::89:eff:fe7d:66cd: id=85 seq=2 hop limit=64 time = 16.553 ms
8 bytes from fe80::89:eff:fe7d:66cd: id=85 seq=3 hop limit=64 time = 16.884 ms
8 bytes from fe80::89:eff:fe7d:66cd: id=85 seq=4 hop limit=64 time = 16.711 ms
8 bytes from fe80::89:eff:fe7d:66cd: id=85 seq=5 hop limit=64 time = 16.408 ms
--- snip ---
8 bytes from fe80::89:eff:fe7d:66cd: id=84 seq=9998 hop limit=64 time = 17.075 ms
8 bytes from fe80::89:eff:fe7d:66cd: id=84 seq=9999 hop limit=64 time = 16.588 ms
8 bytes from fe80::89:eff:fe7d:66cd: id=84 seq=10000 hop limit=64 time = 16.944 ms
--- ff02::1 ping statistics ---
10000 packets transmitted, 10000 received, 0% packet loss, time 234.06317704 s
rtt min/avg/max = 16.313/39959.250/4294918.971 ms
```

No package loss :)

smlng:
@jue89 please stay tuned ...

We are currently in the process of doing another RIOT release, which involves a lot of testing and fixing existing stuff. And I haven't had time to take a closer look at the coding part not to mention running any test on this. 

I'll try to do a first review in the next days, if you want to add fixes here: do not hesitate to do so.

smlng:
for starters look at what Travis complains about and fix that, ignore the errors regarding your "fixup" commits, these can be squashed later. But documentation and header guards etc. can be fixed.

jue89:
Thank you for your response, @smlng!

I pushed further commits into the branch:
1) 6ad457a Remove the CRC from the eth-o-can frames before handing them over to the upper layer.
1) 185e49d Documentation
1) 70dc556 Different back off time before transitioning from BLOCKED to IDLE. If a node was sending a frame onto the bus, the back off time will be significant larger than the time if it was just receiving a frame. Therefore, nodes that received a request are prioritized to respond before the request sending node can send another request and may overwhelm the receiving node.
1) ee6c577 Implementation of NETOPT_PROMISCUOUSMODE.
1) a95716a Optimisation of the state machine's main switch-case statement.

miri64:
To be clear, I only had a quick look at this PR and I don't know much about CAN, but is there any reason you decided against using the existing [CAN API](http://doc.riot-os.org/group__sys__can.html) instead of directly accessing the device over UART. This would have the added benefit that one could test this on `native`, since bindings to `libsocketcan` exist.

jue89:
Hi @miri64,

*I tried to explain why I did this design decision. But in the end it became the story behind this PR - so it got a little bit longer. I hope you don't mind ...*

My initial intention was to bring IPv6 to connected nodes using a bus wire for my home automation. IPv6 because it’s the protocol of the Internet. It is routable and nodes can connect devices outside the local network segment without any protocol conversion. It also supports multicast … I think you know the benefits of using IPv6.

The bus wire just travels through my flat on the heating pipes. I don’t need much bandwidth and want to save energy. Hence, solutions with wired Ethernet and a MAC/PHY (e.g. ENC28J60) aren’t a solution to my problem. Especially the star architecture creates hassle when you connect many devices.

In the first place I wanted to use a RS485 bus to connect all nodes. But this is not really multi master capable. (Of course I could pass around some tokens and every node is the master at some time – but I throw that idea away. Too much complexity. Energy consumption even if no data needs to be transmitted. High delays due to the token system …)

So I stumbled over CAN. I tried to map IPv6 on the CAN frames but faced the problem quite fast: The payload holds a payload of only 8 bytes. [1] I thought about fragmenting IPv6 like 6LoWPAN does. But the complexity of that task and the overhead (40% of the sent payload) kept me from doing that. And if you are using the built-in CAN periphery you have to stick to that frame format.

Finally, I came to the idea of utilising the UART periphery instead of the CAN periphery, as proposed on this PR.

[1] https://en.wikipedia.org/wiki/CAN_bus#Data_frame

miri64:
> Finally, I came to the idea of utilising the UART periphery instead of the CAN periphery, as proposed on this PR.

Okay, but why then not use just SLIP? Or if you really need Ethernet PPP?

miri64:
Or, if you like hacks and it really boils down to Ethernet-over-UART why not [`ethos`](http://doc.riot-os.org/group__drivers__ethos.html)? Don't get me wrong. If your solution works better for you, we are happy to merge this, but I just want to make sure if you know about the alternatives.

jue89:
*No worries. It's completely ok to me to discuss about the solution. Maybe we can find a better approach? I would be fine with that!*

SLIP, PPP and ethos are just point to point. I want to connect several nodes to the bus.

miri64:
> SLIP, PPP and ethos are just point to point. I want to connect several nodes to the bus.

Sorry, I completely forgot the *bus* part m(. Nevermind then. I rest my case :D

vincent-d:
> Finally, I came to the idea of utilising the UART periphery instead of the CAN periphery, as proposed on this PR.

Indeed, this has nothing to do with CAN as it's normally used. You're only using the physical layer. This is clever though :)

drivers/ethocan/ethocan.c
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    /* The set dirty flag prevents recv_buf or recv_buf_ptr from being
+     * touched in ISR context. Thus, it is safe to work with them without
+     * IRQs being disabled or mutexes being locked. */
+
+    /* Check the dst mac addr if the iface is not in promiscous mode */
+    if (!opt_isset(ctx, ETHOCAN_OPT_PROMISCUOUS)) {
+        ethernet_hdr_t *hdr = (ethernet_hdr_t *) ctx->recv_buf;
+        if ((hdr->dst[0] & 0x1) == 0 && memcmp(hdr->dst, ctx->mac_addr, ETHERNET_ADDR_LEN) != 0) {
```
benpicco: Is this bit used to indicate broadcast?
Why not check for the broadcast address?

benpicco:
Just an idea from an interested bystander: Many MCUs implement an 'address detect' feature in 9-bit mode (parity bit is used as an additional data bit) where they will only generate an interrupt if the 9th bit is set.

So you have the address at the start of your frame and only when it matches you turn off the address detect mode to get an interrupt for every byte, so all other nodes are not bothered with interrupts for packets that are not intended for them.

Also an 8-bit address (`ETHERNET_ADDR_LEN = 1`) is probably enough for such a bus - each node will add impedance and degrade the signal. CAN is specified for 128 nodes, maybe you can do more than that but I doubt you need 48-bit addresses ;)

drivers/ethocan/ethocan.c
jue89: This bit identifies multicast and broadcast addresses. If I understood the Ethernet spec correctly, this bit is only zero if it is a unicast frame.

benpicco: Ah, I did not know that!

jue89:
Thank you for diving into this PR, @benpicco :)

> Just an idea from an interested bystander: Many MCUs implement an 'address detect' feature in 9-bit mode (parity bit is used as an additional data bit) where they will only generate an interrupt if the 9th bit is set.
> 
> So you have the address at the start of your frame with the 9th bit set and only when it matches you turn off address detect mode to get an interrupt for every byte. This way all other nodes are not bothered with interrupts from packets that are not intended for them.
> 
> Also an 8-bit address (`ETHERNET_ADDR_LEN = 1`) is probably enough for such a bus - each node will add impedance and degrade the signal. CAN is specified for 128 nodes, maybe you can do more than that but I doubt you need 48-bit addresses ;)

That would be a good optimization - indeed! A lot of interrupts are generated during transmissions. And a lot of overhead is carried around to transfer small bits of payload. But I think there are a few points against this optimization:
 * The generic UART driver inside of RIOT is not capable of using this 9th address bit, since it is boiled down to the feature set that every MCU offers. I really would like to sick to that driver because it enables this driver to run on almost every MCU.
 * A reduced address space creates a higher likelihood of address collisions. I do not set the MAC address by hand. Instead I use the CPU ID to generate a unique address. I like the idea not to have to fiddle around with manual address assignment.

jue89:
Slightly off-topic: Is someone at Embedded World next week? I'll be around on Wednesday and Thursday.

benpicco:
I'll be there at Wednesday too.

miri64:
@benpicco @jue89 I'm personally not there but we have an exhibition booth where some RIOTers will be.

jue89:
> @benpicco @jue89 I'm personally not there but we have an exhibition booth where some RIOTers will be.
I know - its definitely on my to-vist list :)

@benpicco I wrote you an email. I hope the address that can be found in your recent commits is the right one ;)

sys/auto_init/netif/auto_init_ethocan.c
```diff
@@ -0,0 +1,71 @@
...
+ * @brief   Stacks for the MAC layer threads
+ */
+static char _netdev_eth_stack[ETHOCAN_MAC_STACKSIZE];
+
+void auto_init_ethocan(void)
+{
+    LOG_DEBUG("[auto_init_netif] initializing ethocan #0.\n");
+
+    /* setup netdev device */
+    ethocan_params_t p;
```
benpicco:
```suggestion
    for (unsigned i = 0; i < ARRAY_SIZE(ethocan_params); i++) {
        ethocan_setup(&ethocan[i], &ethocan_params[i]);
        gnrc_netif_ethernet_create(_netdev_eth_stack[i], ETHOCAN_MAC_STACKSIZE,
                                   ETHOCAN_MAC_PRIO, "ethocan", (netdev_t *)&ethocan[i]);
    }
```

sys/auto_init/netif/auto_init_ethocan.c
```diff
@@ -0,0 +1,71 @@
...
+#include "debug.h"
+#include "ethocan.h"
+#include "periph/uart.h"
+#include "periph/gpio.h"
+#include "net/gnrc/netif/ethernet.h"
+
+/**
+ * @brief global ethocan object
+ */
+ethocan_t ethocan;
```
benpicco:
```suggestion
static ethocan_t ethocan[ARRAY_SIZE(ethocan_params)];
```

sys/auto_init/netif/auto_init_ethocan.c
```diff
@@ -0,0 +1,71 @@
...
+#endif
+
+#ifndef ETHOCAN_BAUDRATE
+#define ETHOCAN_BAUDRATE      (115200)
+#endif
+
+/**
+ * @brief   Stacks for the MAC layer threads
+ */
+static char _netdev_eth_stack[ETHOCAN_MAC_STACKSIZE];
```
benpicco:
```suggestion
static char _netdev_eth_stack[ARRAY_SIZE(ethocan_params)][ETHOCAN_MAC_STACKSIZE];
```

sys/auto_init/netif/auto_init_ethocan.c
```diff
@@ -0,0 +1,71 @@
...
+
+void auto_init_ethocan(void)
+{
+    LOG_DEBUG("[auto_init_netif] initializing ethocan #0.\n");
+
+    /* setup netdev device */
+    ethocan_params_t p;
+    p.uart = ETHOCAN_UART;
+    p.baudrate = ETHOCAN_BAUDRATE;
+    p.sense_pin = ETHOCAN_SENSE_PIN;
```
benpicco: Move these to `ethocan_params.h` like it is done in [`at86rf2xx_params.h`](https://github.com/RIOT-OS/RIOT/blob/master/drivers/at86rf2xx/include/at86rf2xx_params.h#L53).

jue89:
Thank you for your review, @benpicco!

I'll try to find some time within the next week to address your good suggestions. :)

miri64:
[This discussion](https://mailarchive.ietf.org/arch/msg/6lo/ReCq7JGvh_j3TU4shH_7LsobO6U) might also be of interest here.

jue89:
> [This discussion](https://mailarchive.ietf.org/arch/msg/6lo/U4BI68en6pjxwbd6JBq72JF9knM) might also be of interest here.

That's an interesting proposal! Thank you for the hint :)

wysman:
Thanks you for your proposal,
I am interested to interconnect multiple devices on a DIN rail.

benpicco:
I have a couple of CAN transceivers now, will set up a test network when I get around to it :)

Maybe the name is misleading, this is only using the CAN transceivers to provide a robust differential signal, on the MCU side it's all UART.
How about ***B**us **U**sing **E**thernet over **C**an **H**ardware with **S**erial **E**ndpoints* or
***D**ifferentially **O**perated **S**erial **E**thernet* to make clear it's totally not CAN 😉

jue89:
> Maybe the name is misleading, this is only using the CAN transceivers to provide a robust differential signal, on the MCU side it's all UART.
> 
> How about _**B**us **U**sing **E**thernet over **C**an **H**ardware with **S**erial **E**ndpoints_ or
> _**D**ifferentially **O**perated **S**erial **E**thernet_ to make clear it's totally not CAN 😉

Yeah, I am pretty bad at naming things. I really have to improve this skill before having children :D

I really like your proposed names! Both are well pronounceable for Germans. The last one also for English speakers. And I see what you did there. I am a huge fan of the reference to tin can telephones :)

The only advantage of sticking with "ethocan" is that it can be found underneath the ethos driver in the API description.

And changing the name before merge should be a problem?! I am open to other opinions and willing to change the name if we find a good alternative.

jue89:
BTW a small glimpse into what's going on in my little lab:

I am currently working an a Devkit. It features:
 * A CAN transceiver hooked up to RJ10 jacks with some blinky LEDs for RX/TX indication.
 * 16 GPIOs (with ADC, DAC, SPI, I2C and UART as secondary functions) that are exposed to a mini perfboard section.

I want to use it for testing this driver in a real-world application and fast-prototyping of nodes. The PCB already arrived. ATM I am waiting for the parts. If this layout proves to be okay, I'll publish the KiCAD files here on Github.

*(If someone wants to support this PR by testing but doesn't have the right hardware, I would mail some boards at cost price to interested persons. But I cannot make promises, yet.)*

PCB

Schematic

jue89:
New commits! Finally! Thank you for being patient :)

 1. 168d93d Some clean-up. And a minor change in the state machine code: Checking if the current state is valid (what's always the case if the code is considered *bug free*) is now ensured by `assert()` calls. Thus, this is only checked with `DEVELHELP=1` and doesn't cost valuable CPU cycles in production.
 2. f5a5492 Implementation of @benpicco good suggestions.

With the second commit, the path for the *gnrc_network* example changed slightly:
``` diff
diff --git a/examples/gnrc_networking/Makefile b/examples/gnrc_networking/Makefile
index 7aac1ae..4e78940 100644
--- a/examples/gnrc_networking/Makefile
+++ b/examples/gnrc_networking/Makefile
@@ -18,7 +18,10 @@ BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-mega2560 arduino-uno \
 
 # Include packages that pull up and auto-init the link layer.
 # NOTE: 6LoWPAN will be included if IEEE802.15.4 devices are present
-USEMODULE += gnrc_netdev_default
+USEMODULE += ethocan
+CFLAGS += '-DETHOCAN_PARAM_UART=UART_DEV(2)'
+CFLAGS += '-DETHOCAN_PARAM_SENSE_PIN=GPIO_PIN(PORT_B, 1)'
 USEMODULE += auto_init_gnrc_netif
 # Activate ICMPv6 error messages
 USEMODULE += gnrc_icmpv6_error
```

Left ToDos:
 * Instead of implementing an own CRC16, maybe use the shipped CRC16-CCITT implementation.
 * Further documentation.
 * Rename *ethocan*?!

jue89:
Further progress:
 * 5509b63: Documentation. Basically I addressed @benpicco suggestion to include more details in the header file and put the wiring diagram into the Doxygen docu.
 * 0aa88b5: Instead of implementing own bit arithmetic, I used the one I found in the RIOT core.
 * 5509b63: Fixed typos Travis pointed out ...

I am still not sure whether to use the CRC16-CCITT that comes with RIOT or stick with the one I implemented. What do you think, @benpicco @miri64? It will required more ROM but would be significantly faster compared to the current one.

benpicco:
This is looking good!

As for CRC, I think the best would be to have it compile-time configurable. But this is out of the scope of this PR and should be it's own PR.

I can imagine stations where speed is not as important but you don't want to spend 512 bytes on CRC16.

jue89:
I would like to address the naming thing before this PR gets merged. After several nights of sleeping on it, I would prefer the name DOSE. It is pronounceable for English and German speakers. Differentially Operated Serial Ethernet is descriptive. Nobody would imply that any CAN is involved. Any opinions an that?

Travis is complaining about a typo in auto_init.c that is actually an include. How do I “fix” this?

benpicco:
I'm fine with that :laughing: 
A less whimsical name could also be ethos-bus, if that's not confusing for not sharing any code with ethos.

The typos are no Errors, just warnings - they won't block a PR.

If you decide for a name, you can squash directly. Makes it easier for you and us reviewing the patch.

jue89:
Squashed. I hope it is okay that everything went into one commit.

> A less whimsical name could also be ethos-bus, if that's not confusing for not sharing any code with ethos.

I have a soft spot for whimsical names ;) Thank you for that good idea! I really like it :)

drivers/dose/dose.c
```diff
@@ -0,0 +1,566 @@
...
+    .isr = _isr,
+    .get = _get,
+    .set = _set
+};
+
+void dose_setup(dose_t *ctx, const dose_params_t *params)
+{
+    ctx->netdev.driver = &netdev_driver_dose;
+
+    ctx->state = DOSE_STATE_UNDEF;
```
jue89: This is set twice. `ctx->state` is also reset in the netdev driver's `init()` method.

drivers/dose/dose.c
```diff
@@ -0,0 +1,566 @@
...
+        return;
+    }
+
+    /* The set dirty flag prevents recv_buf or recv_buf_ptr from being
+     * touched in ISR context. Thus, it is safe to work with them without
+     * IRQs being disabled or mutexes being locked. */
+
+    /* Check the dst mac addr if the iface is not in promiscuous mode */
+    if (!(ctx->opts & DOSE_OPT_PROMISCUOUS)) {
+        ethernet_hdr_t *hdr = (ethernet_hdr_t *) ctx->recv_buf;
```
jue89: We haven't checked if enough bytes have been received. We need at least `sizeof(ethernet_hdr_t) + DOSE_FRAME_CRC_LEN` bytes.

drivers/dose/dose.c
```diff
@@ -0,0 +1,572 @@
...
+            break;
+        case DOSE_STATE_SEND + DOSE_SIGNAL_XTIMER:
+            new_state = DOSE_STATE_BLOCKED;
+            break;
+        case DOSE_STATE_SEND + DOSE_SIGNAL_END:
+            new_state = DOSE_STATE_BLOCKED;
+            break;
+    }
+
+    assert(new_state > DOSE_STATE_UNDEF && new_state < DOSE_STATE_INVALID);
```
jue89: In rare cases, this assertions actually raises kernel panics! This happens if sense pin interrupts are lost or being overtaken by uart interrupts. I need to rethink the state machine to handle these cases!

jue89:
I pushed further fixup commits. Basically, I reworked the whole state machine. It is much cleaner now - in my humble opinion ;) It does not get stuck if IRQs are firing unexpectedly. Furthermore, it would also function without the sense pin! Once PR is merged, I am planning to open another PR that makes the sense pin optional. (With to cost of a higher probability of collisions ...)

I've tested these changes for 18 hours and none of the MCUs raised a kernel panic :)

I updated the finite state machine diagram. It should support the review process: Every transition is a `case` statement inside the `state()` function. Every arrow head stands for a `state_transit_x()` call with `x` being the state the arrow head is pointing at. Please note that the `state_transit_x()` function is called just before the state is entered.

DOSE FSM

Is anyone already in the review process (@benpicco)? If not, I would squash again.

benpicco:
Feel free to squash, I haven't gotten around to look at it again yet - maybe this weekend :)

drivers/dose/dose.c
```diff
@@ -0,0 +1,557 @@
...
+    state(ctx, DOSE_SIGNAL_END);
+
+    return pktlen;
+
+collision:
+    DEBUG("dose _send(): collision!\n");
+    if (--retries < 0) {
+        return 0;
+    }
+    goto send;
```
jue89: Call `ctx->netdev.event_callback((netdev_t *) ctx, NETDEV_EVENT_TX_MEDIUM_BUSY);` if we weren't able to send the packet. This will increase the `tx_failed` counter for the `ifconfig` command.

drivers/dose/include/dose_params.h
```diff
@@ -0,0 +1,62 @@
...
+ * @{
+ */
+#ifndef DOSE_PARAM_UART
+#define DOSE_PARAM_UART         (UART_DEV(0))
+#endif
+#ifndef DOSE_PARAM_BAUDRATE
+#define DOSE_PARAM_BAUDRATE     (115200)
+#endif
+#ifndef DOSE_PARAM_SENSE_PIN
+#define DOSE_PARAM_SENSE_PIN    (GPIO_PIN(0, 0))
```
benpicco:
```suggestion
#define DOSE_PARAM_SENSE_PIN    GPIO_UNDEF
```
If you want to signal that pin can be unused by default.
Beware that calling gpio functions on a GPIO with the value `GPIO_UNDEF` will lead to crashes on some platforms. (`gpio_set()` and friends are not expected to do sanity checking of parameters for the sake of responsiveness)

sys/auto_init/netif/auto_init_dose.c
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @brief   Define stack parameters for the MAC layer thread
+ * @{
+ */
+#define DOSE_MAC_STACKSIZE (THREAD_STACKSIZE_DEFAULT + DEBUG_EXTRA_STACKSIZE)
+#ifndef DOSE_MAC_PRIO
+#define DOSE_MAC_PRIO      (GNRC_NETIF_PRIO)
+#endif
+
+#define DOSE_NUM (sizeof(dose_params) / sizeof(dose_params[0]))
```
benpicco:
```suggestion
#define DOSE_NUM ARRAY_SIZE(dose_params)
```

drivers/Makefile.dep
```diff
@@ -178,6 +178,15 @@ ifneq (,$(filter ethos,$(USEMODULE)))
   USEMODULE += tsrb
 endif
 
+ifneq (,$(filter dose,$(USEMODULE)))
+  FEATURES_REQUIRED += periph_uart
+  USEMODULE += iolist
+  USEMODULE += netdev_eth
+  USEMODULE += random
+  USEMODULE += xtimer
+  USEMODULE += periph_gpio_irq
```
benpicco:
```suggestion
  FEATURES_REQUIRED += periph_gpio_irq
```

drivers/dose/dose.c
```diff
@@ -0,0 +1,557 @@
...
+
+    do {
+        /* The edges of the finite state machine can be identified by
+         * the current state and the signal that caused a state transition.
+         * Since the state only occupies the first 4 bits and the signal the
+         * last 4 bits of a uint8_t, they can be added together and hance
+         * be checked together. */
+        switch (ctx->state + signal) {
+            case DOSE_STATE_INIT + DOSE_SIGNAL_INIT:
+            case DOSE_STATE_RECV + DOSE_SIGNAL_END:
```
benpicco: Neat idea!

benpicco:
When running on ATmega, I get `dose state(): unexpected state transition (STATE=0x00 SIGNAL=0x64)`

jue89:
Thank you very much for testing out and reproducing my PR :)

> When running on ATmega, I get `dose state(): unexpected state transition (STATE=0x00 SIGNAL=0x64)` and then it seems to be getting stuck in the [first loop](https://github.com/RIOT-OS/RIOT/blob/2266234de9e410a37a6f1a3567f3920b0635cfad/drivers/dose/dose.c#L401) of `_send()`.

This is weird ... there exists no signal `0x64` ... I'll try to investigate on that. What board are you using?

benpicco:
> What board are you using?

`arduino-mega2560` (*Mega 2560 PRO*). I have to use `gnrc_minimal` as `gnrc_networking` does not fit.

drivers/dose/dose.c
```diff
@@ -0,0 +1,557 @@
...
+static void state(dose_t *ctx, dose_signal_t signal)
+{
+    /* Make sure no other thread or ISR interrupts state transitions */
+    int irq_state = irq_disable();
+
+    do {
+        /* The edges of the finite state machine can be identified by
+         * the current state and the signal that caused a state transition.
+         * Since the state only occupies the first 4 bits and the signal the
+         * last 4 bits of a uint8_t, they can be added together and hance
```
kaspar030:
```suggestion
         * last 4 bits of a uint8_t, they can be added together and hence
```

jue89:
@kaspar030 Thank you for your review and the warm words about the coding style :)

@benpicco I went through the code twice. I cannot find a single path where `SIGNAL=0x64` could occur ... a friend of mine owns a *Arduino Mega 2560 PRO*. He will give a me in a few days ... maybe I can reproduce this behaviour. But I also think this is some timer-related problem.

I think I find the time to address your suggestions this weekend. Thank you both again!

drivers/dose/include/dose_params.h
jue89: I would suggest to make the `SENSE_PIN` optional in a follow-up PR. Otherwise the number of tests that must be performed nearly doubles. But I'll keep that in mind ;)

drivers/dose/dose.c
jue89: Added with commit 5959328

drivers/dose/dose.c
```diff
@@ -0,0 +1,557 @@
...
+    ctx->uart = params->uart;
+    uart_init(ctx->uart, params->baudrate, _isr_uart, (void *) ctx);
+
+    ctx->sense_pin = params->sense_pin;
+    gpio_init_int(ctx->sense_pin, GPIO_IN, GPIO_FALLING, _isr_gpio, (void *) ctx);
+    gpio_irq_disable(ctx->sense_pin);
+
+    luid_get(ctx->mac_addr, ETHERNET_ADDR_LEN);
+    ctx->mac_addr[0] &= (0x2);
+    ctx->mac_addr[0] &= ~(0x1);
```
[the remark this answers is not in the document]
jue89: I push commit 1ad4787 which includes your suggested changes.

Do I use the new `eui48_t` type correctly? I'm now doing a lot of `memcpy` and `memcmp` with the included `uint8` struct member. Or are there any helper functions to achieve this a little bit more elegant?

From a technical point of view this should work perfectly. I'm ensuring `ETHERNET_ADDR_LEN` has the same length as the `eui48_t.uint8` array.

drivers/dose/dose.c
```diff
@@ -0,0 +1,557 @@
...
+    CLRBIT(ctx->flags, DOSE_FLAG_RECV_BUF_DIRTY);
+    CLRBIT(ctx->flags, DOSE_FLAG_END_RECEIVED);
+    CLRBIT(ctx->flags, DOSE_FLAG_ESC_RECEIVED);
+    irq_restore(irq_state);
+}
+
+static void _isr(netdev_t *netdev)
+{
+    dose_t *ctx = (dose_t *) netdev;
+    int irq_state, dirty, end;
```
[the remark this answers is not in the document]
jue89: Commit 3e23083 should include all your requested changes :)

drivers/dose/dose.c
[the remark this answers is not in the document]
benpicco: I don't think there are any helper functions for that.
The main purpose of that type is to signal that the bytes are in the correct order which turns out can be not so obvious when you just have a byte array.

benpicco:
Btw: I figured out the issue on ATmega:

 - you are using xtimer_set() with ticks, but it expects µs
 - you are assuming you can disable the timer ISR by using `irq_disable()`, but if the offset is smaller than `XTIMER_BACKOFF` no interrupt will be generated but the callback will be executed immediately.

A simple solution would be to ensure you set the timer always to at least `XTIMER_BACKOFF`.

```patch
diff --git a/drivers/dose/dose.c b/drivers/dose/dose.c
index 2d477cd23..de52d3dec 100644
--- a/drivers/dose/dose.c
+++ b/drivers/dose/dose.c
@@ -51,6 +51,18 @@ static int _set(netdev_t *dev, netopt_t opt, const void *value, size_t len);
 static int _init(netdev_t *dev);
 void dose_setup(dose_t *ctx, const dose_params_t *params);
 
+static void xtimer_set_safe(xtimer_t *timer, uint32_t offset)
+{
+    xtimer_ticks32_t t;
+
+    if (offset <= XTIMER_BACKOFF) {
+        offset = XTIMER_BACKOFF + 1;
+    }
+
+    t.ticks32 = offset;
+    xtimer_set(timer, xtimer_usec_from_ticks(t));
+}
+
 static uint16_t crc16_update(uint16_t crc, uint8_t octet)
 {
     crc = (uint8_t)(crc >> 8) | (crc << 8);
@@ -89,7 +101,7 @@ static dose_signal_t state_transit_blocked(dose_t *ctx, dose_signal_t signal)
     else {
         backoff = random_uint32_range(ctx->timeout_ticks / 10, ctx->timeout_ticks);
     }
-    xtimer_set(&ctx->timeout, backoff);
+    xtimer_set_safe(&ctx->timeout, backoff);
 
     return DOSE_SIGNAL_NONE;
 }
@@ -138,7 +150,7 @@ static dose_signal_t state_transit_recv(dose_t *ctx, dose_signal_t signal)
 
     if (rc == DOSE_SIGNAL_NONE) {
         /* No signal is returned. We stay in the RECV state. */
-        xtimer_set(&ctx->timeout, ctx->timeout_ticks);
+        xtimer_set_safe(&ctx->timeout, ctx->timeout_ticks);
     }
 
     return rc;
@@ -157,7 +169,7 @@ static dose_signal_t state_transit_send(dose_t *ctx, dose_signal_t signal)
      * will bring us back to the BLOCKED state after _send has emitted
      * its last octet. */
 
-    xtimer_set(&ctx->timeout, ctx->timeout_ticks);
+    xtimer_set_safe(&ctx->timeout, ctx->timeout_ticks);
 
     return DOSE_SIGNAL_NONE;
 }
```
It's not pretty, but with this `gnrc_minimal` is working on my `arduino-mega2560` and I can ping it.
Maybe @kaspar030 has a better idea :wink:

jue89:
> Btw: I figured out the issue on ATmega:
> 
> * you are using `xtimer_set()` with ticks, but it expects µs
> * you are assuming you can disable the timer IRQ by using `irq_disable()`, but if the offset is smaller than `XTIMER_BACKOFF` (in ticks!) no interrupt will be generated but the callback will be executed immediately.

Thank you for pointing this out! I'm a little bit confused about the xtimer API, to be honest ;)

But in this case I would store the timeout in µs inside the `dose_t` context not in ticks. So we can get rid of the back conversion. Instead I would check if the desired timeout is above a minimal timeout. I think the compiler should be able to calculate this minimal timeout during compile time ...

drivers/dose/dose.c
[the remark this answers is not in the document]
jue89: Are you are happy with my implementation? Or would you have done something differently?

jue89:
I wasn't able to sleep - so I head back to my PC :D

I just pushed 1127afd.

The basic idea: Instead of always checking if the selecting timeout is long enough, we only check this once during `dose_setup()`.

Good night! I really appreciate your effort put into this PR :)

benpicco:
I can confirm that with that patch the driver is working on ATmega too.

jue89:
@benpicco Thank you! Have you had a look into the documentation generated by Doxygen? The `Reviewed: Documentation` label hasn't been set, yet.

kaspar030:
I'll take a quick look at the docs.

kaspar030:
> I'll take a quick look at the docs.

LGTM.

kaspar030:
pls squash!

jue89:
Thank you all for your great support! :)

PR #10714, opened 2019-01-06
drivers/ads101x: eliminate const to non-const pointer cast

keestux:
A helper variable is introduced to make it clear how the void pointer
argument is intended to be used. After that the other casts are not
needed anymore.

This is not yet tested, but I have good hopes that no code will break because of this change.


smlng:
needs rebase

smlng:
ah well somehow #10707 already did most of that, so duplicate?

keestux:
Sorry about that. It's a duplicate indeed.

PR #10717, opened 2019-01-06
sys/evtimer: refactor evtimer add()/del()

keestux:
I did a few brief tests (native `evtimer_msg` and `evtimer_underflow`). They operate as before. Hopefully we have plenty of other tests that will tell us if the change is correct.

The main reason for doing the refactoring was to get rid of the casting hack.

    /* XXX this function is intentionally non-static, since the optimizer can't
     * handle the pointer hack in this function */
    void evtimer_add_event_to_list(evtimer_t *evtimer, evtimer_event_t *event)
    {
        /* we want list->next to point to the first list element. thus we take the
         * *address* of evtimer->events, then cast it from (evtimer_event_t **) to
         * (evtimer_event_t*). After that, list->next actually equals
         * evtimer->events. */
        evtimer_event_t *list = (evtimer_event_t *)&evtimer->events;

As far as I can tell it is "just" a singly linked list and we want to add and delete entries from it. (And
updating the offset field of the new (add) or next (del) event.) It should be a schoolbook exercise, but
perhaps I'm missing some important detail.

### DIFF
--- a/sys/evtimer/evtimer.c
+++ b/sys/evtimer/evtimer.c
@@ -31,64 +31,63 @@
 #define ENABLE_DEBUG (0)
 #include "debug.h"
 
-/* XXX this function is intentionally non-static, since the optimizer can't
- * handle the pointer hack in this function */
-void evtimer_add_event_to_list(evtimer_t *evtimer, evtimer_event_t *event)
+static void _add_event_to_list(evtimer_t *evtimer, evtimer_event_t *event)
 {
     DEBUG("evtimer: new event offset %" PRIu32 " ms\n", event->offset);
-    /* we want list->next to point to the first list element. thus we take the
-     * *address* of evtimer->events, then cast it from (evtimer_event_t **) to
-     * (evtimer_event_t*). After that, list->next actually equals
-     * evtimer->events. */
-    evtimer_event_t *list = (evtimer_event_t *)&evtimer->events;
+    evtimer_event_t **list = &evtimer->events;
 
-    while (list->next) {
-        evtimer_event_t *list_entry = list->next;
+    while (*list) {
         /* Stop when new event time is nearer then next */
-        if (event->offset < list_entry->offset) {
+        if (event->offset < (*list)->offset) {
             DEBUG("evtimer: next %" PRIu32 " < %" PRIu32 " ms\n",
-                  event->offset, list_entry->offset);
+                  event->offset, (*list)->offset);
             break;
         }
         /* Set event offset relative to previous event */
-        event->offset -= list_entry->offset;
-        list = list->next;
+        event->offset -= (*list)->offset;
+        list = &(*list)->next;
     }
 
     DEBUG("evtimer: new event relativ offset %" PRIu32 " ms\n", event->offset);
 
     /* Set found next bigger event after new event */
-    event->next = list->next;
-    if (list->next) {
+    event->next = *list;
+    if (*list) {
         /* Set offset following event relative to new event */
-        evtimer_event_t *next_entry = list->next;
         DEBUG("evtimer: recalculate offset for %" PRIu32 " ms\n",
-              next_entry->offset);
+              (*list)->offset);
 
-        next_entry->offset -= event->offset;
+        (*list)->offset -= event->offset;
 
         DEBUG("evtimer: resulting new event offset %" PRIu32 " ms\n",
-              next_entry->offset);
+              (*list)->offset);
     }
 
-    list->next = event;
+    *list = event;
 }
 
 static void _del_event_from_list(evtimer_t *evtimer, evtimer_event_t *event)
 {
-    evtimer_event_t *list = (evtimer_event_t *) &evtimer->events;
+    evtimer_event_t **list = &evtimer->events;
 
-    while (list->next) {
-        evtimer_event_t *list_entry = list->next;
-        if (list_entry == event) {
-            list->next = event->next;
-            if (list->next) {
-                list_entry = list->next;
-                list_entry->offset += event->offset;
-            }
+    /* Find the entry to delete from the list */
+    while (*list) {
+        if (*list == event) {
             break;
         }
-        list = list->next;
+        list = &(*list)->next;
+    }
+
+    if (*list) {
+        /* This is the entry we want to remove */
+        *list = (*list)->next;
+        /* If the deleted entry was _not_ the last one
+         * then update 'offset' of the entry that
+         * followed.
+         */
+        if (*list) {
+            (*list)->offset += event->offset;
+        }
     }
 }
 
@@ -145,7 +144,7 @@ void evtimer_add(evtimer_t *evtimer, evtimer_event_t *event)
     DEBUG("evtimer_add(): adding event with offset %" PRIu32 "\n", event->offset);
 
     _update_head_offset(evtimer);
-    evtimer_add_event_to_list(evtimer, event);
+    _add_event_to_list(evtimer, event);
     if (evtimer->events == event) {
         _set_timer(&evtimer->timer, event->offset);
     }
--- a/sys/evtimer/evtimer.c
+++ b/sys/evtimer/evtimer.c
@@ -209,10 +209,11 @@ void evtimer_init(evtimer_t *evtimer, evtimer_callback_t handler)
 void evtimer_print(const evtimer_t *evtimer)
 {
     evtimer_event_t *list = evtimer->events;
+    int nr = 0;
 
-    while (list->next) {
-        evtimer_event_t *list_entry = list->next;
-        printf("ev offset=%u\n", (unsigned)list_entry->offset);
+    while (list) {
+        nr++;
+        printf("ev #%d offset=%u\n", nr, (unsigned)list->offset);
         list = list->next;
     }
 }
--- a/tests/evtimer_msg/main.c
+++ b/tests/evtimer_msg/main.c
@@ -27,14 +27,20 @@
 
 static char worker_stack[THREAD_STACKSIZE_MAIN];
 static evtimer_t evtimer;
-static evtimer_msg_event_t events[] = {
-    { .event = { .offset = 1000 }, .msg = { .content = { .ptr = "supposed to be 1000" } } },
-    { .event = { .offset = 1500 }, .msg = { .content = { .ptr = "supposed to be 1500" } } },
-    { .event = { .offset = 659 }, .msg = { .content = { .ptr = "supposed to be 659" } } },
-    { .event = { .offset = 3954 }, .msg = { .content = { .ptr = "supposed to be 3954" } } },
+#define NEVENTS (unsigned)(4)
+/*
+ * The events (.offset) are modified by evtimer_add.
+ * This list of offsets is used to populate the events
+ * before adding (or re-adding).
+ */
+static uint32_t offsets[NEVENTS] = {
+        1000,
+        1500,
+        659,
+        3954,
 };
-
-#define NEVENTS ((unsigned)(sizeof(events) / sizeof(evtimer_msg_event_t)))
+static evtimer_msg_event_t events[NEVENTS];
+static char texts[NEVENTS][40];
 
 /* This thread will print the drift to stdout once per second */
 void *worker_thread(void *arg)
@@ -56,7 +62,7 @@ void *worker_thread(void *arg)
 
 int main(void)
 {
-    uint32_t now = xtimer_now_usec() / US_PER_MS;
+    uint32_t now;
 
     evtimer_init_msg(&evtimer);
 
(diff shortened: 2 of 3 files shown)


keestux:
@kaspar030 would you have a chance to look at this?

keestux:
@maribu can I ask you to have a look at this?

sys/evtimer/evtimer.c
```diff
@@ -31,64 +31,53 @@
...
     DEBUG("evtimer: new event offset %" PRIu32 " ms\n", event->offset);
-    /* we want list->next to point to the first list element. thus we take the
-     * *address* of evtimer->events, then cast it from (evtimer_event_t **) to
-     * (evtimer_event_t*). After that, list->next actually equals
-     * evtimer->events. */
-    evtimer_event_t *list = (evtimer_event_t *)&evtimer->events;
-
-    while (list->next) {
-        evtimer_event_t *list_entry = list->next;
+    evtimer_event_t **list = &evtimer->events;
```
maribu: Why not `evtimer_event_t *list = evtimer->events;` and replace `*list` by `list` in the following lines?

keestux: Well, it is quite essential to have `evtimer_event_t **list = &evtimer->events` !!
Take a look at the last line in the function
```
    *list = event;
```

maribu: Why not using `evtime_event_t *list` and stopping when `list->next` `event` is to be inserted between `list` and `list->next`? Than you would update `event->next` to point to `list->next` and update `list->next` to point to `event`. This would make the whole code much more readable.

keestux: :-) I like you're persistence to want `evtime_event_t *list`.

Have you thought about inserting the first event, when `evtimer->events` is NULL? I think not.

Again I want to emphasize that having the pointer pointer is there on purpose. First it is used to find the place where the list ends (including the situation of an empty list). And when it finds it it already points to the place where the element must be inserted.

maribu: Let me explain why I think it is beneficial to avoid an extra layer of indirection if this can be done without pain: As part of my job I teach low level programming (among other stuff) to computer science students. From my experience at most every second student is able to understand the concept of pointers, as long as you remain at one level of indirection (pointer to foo). When you add one more layer (pointer to pointer to foo), at most one out of ten students will be able to understand the code.

While this is basically anecdotal evidence, I'm pretty sure that using 2 levels of indirection (pointer to pointer to foo) will give a lot of people a hard time trying to understand the code. There clearly are places where using pointers to pointers can make the code much shorter, more performant, or may just not be practical to implement with 1 level of indirection. But when there is no "pain" in using pointer to foo instead of pointer to pointer to foo, this will make a lot of peoples life easier ;-)

Let me sketch how this could be done with 1 level of indirection. (Untested code without having spend more than a second of thought.):

``` C
static void evtimer_add_event_to_list(evtimer_t *evtimer, evtimer_event_t *event)
{
    if (!evtimer->events || (event->offset < evtimer->events->offset)) {
        /* insert as first element */
        event->next = evtimer->events;
        evtimer->events = event;
    }
    else {
        evtimer_event_t *i;
        for (i = evtimer->events; i->next; i = i->next) {
            /* Set event offset relative to previous event */
            event->offset -= i->offset;

            if (event->offset < i->next->offset) {
                break;
            }
        }
        /* insert event between i and i->next */
        event->next = i->next;
        i->next = event;
    }

    if (event->next) {
        /* Updating offset of the follow-up to the newly inserted event to
           be relative to the newly inserted one */
        event->next->offset -= event->offset;
    }
}
```

keestux: I'm in favor of easy to understand code. But I still prefer my version. What else can I do to convince you? Shall we measure code complexity (McCabe?), or shall we measure code size? Do we need more comments (it's basically a simple while-loop), explaining the iteration. ASCII art?

Leaving out DEBUG, it comes down to the following.
```
static void evtimer_add_event_to_list(evtimer_t *evtimer, evtimer_event_t *event)
{
    evtimer_event_t **list = &evtimer->events;

    while (*list) {
        /* Stop when new event time is nearer then next */
        if (event->offset < (*list)->offset) {
            break;
        }
        /* Set event offset relative to previous event */
        event->offset -= (*list)->offset;
        list = &(*list)->next;
    }

    /* Set found next bigger event after new event */
    event->next = *list;
    if (*list) {
        /* Set offset following event relative to new event */
        (*list)->offset -= event->offset;
    }
    *list = event;
}
```

maribu: Note that I deliberately did not mark this comment as "changes requested".

But out of curiosity I did compare the assembly emitted from GCC and clang. Your code and the current master are translated to the exact same assembly with clang8 on x86 with my suggestion being a bit longer. With GCC 8.3.0 on ARM your code is translated to the shortest assembly, followed by the master code and my code being the longest. With GCC 8.3.0 on x86 your code translates again to the shortest assembly, followed by my code with the current master being the longest.

sys/evtimer/evtimer.c
```diff
@@ -210,10 +199,11 @@ void evtimer_init(evtimer_t *evtimer, evtimer_callback_t handler)
 void evtimer_print(const evtimer_t *evtimer)
 {
     evtimer_event_t *list = evtimer->events;
+    int nr = 0;
 
-    while (list->next) {
-        evtimer_event_t *list_entry = list->next;
-        printf("ev offset=%u\n", (unsigned)list_entry->offset);
+    while (list) {
```
maribu: Am I wrong, or did the old code fail to print the first entry? Anyway, this code does what it should.

maribu:
I'll test tomorrow (hopefully)

sys/evtimer/evtimer.c
keestux: Yes, and that is why I have a second commit for that

maribu:
Tested with `tests/evtimer_msg`.

With this PR:
```
2019-07-05 20:15:34,896 - INFO # main(): This is RIOT! (Version: 2019.07-devel-808-gdc399-keestux)
2019-07-05 20:15:34,900 - INFO # Testing generic evtimer (start time = 5 ms)
2019-07-05 20:15:34,906 - INFO # Are the reception times of all 4 msgs close to the supposed values?
2019-07-05 20:15:34,909 - INFO # If yes, the tests were successful
2019-07-05 20:15:35,563 - INFO # At    668 ms received msg 0: "supposed to be 659"
2019-07-05 20:15:35,904 - INFO # At   1009 ms received msg 1: "supposed to be 1000"
2019-07-05 20:15:36,404 - INFO # At   1509 ms received msg 2: "supposed to be 1500"
2019-07-05 20:15:38,858 - INFO # At   3963 ms received msg 3: "supposed to be 3954"
```

with master prior to applying the commits of this pr:
```
2019-07-05 20:20:10,776 - INFO # main(): This is RIOT! (Version: 2019.07-devel-806-gb4079-keestux)
2019-07-05 20:20:10,780 - INFO # Testing generic evtimer (start time = 5 ms)
2019-07-05 20:20:10,785 - INFO # Are the reception times of all 4 msgs close to the supposed values?
2019-07-05 20:20:10,788 - INFO # If yes, the tests were successful
2019-07-05 20:20:11,444 - INFO # At    668 ms received msg 0: "supposed to be 659"
2019-07-05 20:20:11,784 - INFO # At   1009 ms received msg 1: "supposed to be 1000"
2019-07-05 20:20:12,284 - INFO # At   1509 ms received msg 2: "supposed to be 1500"
2019-07-05 20:20:14,738 - INFO # At   3963 ms received msg 3: "supposed to be 3954"
```

maribu:
Hmm, seemingly unrelated stuff now segfaults. I already restarted Murdock to check whether those tests are failing reproducible, which seems to be true... @keestux: Do you have an idea?

keestux:
@maribu I'll have to investigate

keestux:
Ah, my fault. I didn't look carefully enough to the delete function. I've pushed an update, and added a comment in the delete function.

sys/evtimer/evtimer.c
```diff
@@ -31,64 +31,53 @@
 #define ENABLE_DEBUG (0)
 #include "debug.h"
 
-/* XXX this function is intentionally non-static, since the optimizer can't
- * handle the pointer hack in this function */
-void evtimer_add_event_to_list(evtimer_t *evtimer, evtimer_event_t *event)
+static void evtimer_add_event_to_list(evtimer_t *evtimer, evtimer_event_t *event)
```
kaspar030:
```suggestion
static void _add_event_to_list(evtimer_t *evtimer, evtimer_event_t *event)
```

keestux: Good idea

keestux:
@maribu can you please restart Murdock. I expect that all tests will pass now.

kaspar030:
> @maribu can you please restart Murdock. I expect that all tests will pass now.

It should re-build if you push any changes.

keestux:
Argh. I didn't pay attention when pushing. I pushed into RIOT-OS instead of my own. Hopefully I am allowed to removed it here. Sorry.

maribu:
@keestux: I guess in a new branch into RIOT-OS? If so, I guess you can just remove that branch.

maribu:
Btw: It would be very nice to either have more evtimer tests or to extend the two tests there. I was able to run both tests successfully. Ideally, at least one test should have triggered the bug.

maribu:
Feel free to squash

keestux:
@maribu I was thinking the same, about extending the tests
First I'll squash, and then think about tests

maribu:
@keestux: It would be super nice, if you would add the test :-) Do you want to add them here, or as independent PR? I think there are good points for both.

keestux:
I was thinking to add it to this PR

keestux:
@maribu I've pushed a commit with changes to evtimer_msg. This test was the easiest to expand.

keestux:
@maribu What do you think about the modifications of test `tests/evtimer_msg`? Is this enough, for now?

maribu:
Sorry for the long stall. Codewise the test looks fine to me. It is especially nice that you took the start up delay into account, so that the expected times and the actual times are now matching.

I found one typo in a comment (squash right away)

tests/evtimer_msg/main.c
```diff
@@ -65,11 +71,47 @@ int main(void)
...
+    /* Add the all events */
     for (unsigned i = 0; i < NEVENTS; i++) {
+        events[i].event.offset = offsets[i];
+        now = xtimer_now_usec() / US_PER_MS;
+        snprintf(texts[i], sizeof(texts[i]) - 1, "#%u supposed to be %" PRIu32, i, now + events[i].event.offset);
+        events[i].msg.content.ptr = texts[i];
         evtimer_add_msg(&evtimer, &events[i], pid);
     }
+
+    /* Delete the all events */
```
maribu: should be "Delete all the events"

tests/evtimer_msg/main.c
```diff
@@ -65,11 +71,47 @@ int main(void)
                                      THREAD_PRIORITY_MAIN - 1,
                                      THREAD_CREATE_STACKTEST,
                                      worker_thread, NULL, "worker");
-    printf("Testing generic evtimer (start time = %" PRIu32 " ms)\n", now);
+
+    printf("Testing generic evtimer\n");
+
+    /* Add the all events */
```
maribu: Should be "Add all the events"

tests/evtimer_msg/main.c
```diff
@@ -65,11 +71,47 @@ int main(void)
...
+    evtimer_del(&evtimer, &events[2].event);
+    printf("This should list %u items\n", NEVENTS - 2);
+    evtimer_print(&evtimer);
+
+    /* Delete the remaining entries */
+    for (unsigned i = 0; i < NEVENTS; i++) {
+        evtimer_del(&evtimer, &events[i].event);
+    }
+
+    /* Add the all events, again */
```
maribu: Should be "Add all the events, again"

maribu:
Forgot to point out: I tested the test against the `evtimer` implementation in this PR and against RIOT's current master. Both are passing the test, which indicates that this PR indeed fixes code style without changing the behaviour, as intended.

tests/evtimer_msg/main.c
keestux: Done

tests/evtimer_msg/main.c
keestux: Done

tests/evtimer_msg/main.c
keestux: Done. Thanks for the review

maribu:
Sorry, a bit of nit-picking: I'd prefer if the commit message of the last commit would start with `tests:` instead of `evtimer:`, as that commit does not change the `evtimer` code but the tests for the evtimer code. Apart from that, I'd say merge :-)

Thank you very much for cleaning up the code style and extending and improving the unit test!

keestux:
No problem. I'll change the commit message

keestux:
The first line of the commit message is now changed to:
```
tests: adapt evtimer tests
```

Alternatively the first line could be as follows. This prefix is used in the past. 
```
tests/evtimer_msg: modify test to also delete entries
```

keestux:
@maribu besides the commit message, what is your opinion about rebasing to today's master and to do one more test run?

maribu:
> @maribu besides the commit message, what is your opinion about rebasing to today's master and to do one more test run?

If you rebase, I do another round of tests. And then let's finally merge

> tests/evtimer_msg: modify test to also delete entries

I'd personally would prefer that one :-)

keestux:
@maribu rebased and a change in the commit message (as discussed above).

I think we're now ready for a merge.

keestux:
Thanks @maribu

maribu:
You're very welcome :-) Thanks for addressing this issue!

PR #10721, opened 2019-01-07
asymcute: check for minimum packet length early

nmeum:
### Contribution description

Without this patch `_len_get` reads one byte beyond `con->rxbuf` if the incoming packet consists only of the byte `0x01`. Fortunately, the `rxbuf` is initialized with zeros in `asymcute_listener_run` so this doesn't access uninitialized memory. I would still consider this an error though.

### Testing procedure

1. Start the application `examples/asymcute_mqttsn`
2. Send a packet containing only the byte `0x01` to the sock ephemeral port.

Step 2. can be performed using netcat, e.g.:

```
printf '\x01' | nc -u '[link-local-addr%tap0]' 49152
```

### Issues/PRs references

None.

### DIFF
--- a/sys/net/application_layer/asymcute/asymcute.c
+++ b/sys/net/application_layer/asymcute/asymcute.c
@@ -516,6 +516,10 @@ static void _on_unsuback(asymcute_con_t *con, const uint8_t *data, size_t len)
 
 static void _on_data(asymcute_con_t *con, size_t pkt_len, sock_udp_ep_t *remote)
 {
+    if (pkt_len < 2) {
+        return;
+    }
+
     size_t len;
     size_t pos = _len_get(con->rxbuf, &len);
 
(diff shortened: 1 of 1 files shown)


sys/net/application_layer/asymcute/asymcute.c
```diff
@@ -517,15 +517,20 @@ static void _on_unsuback(asymcute_con_t *con, const uint8_t *data, size_t len)
 static void _on_data(asymcute_con_t *con, size_t pkt_len, sock_udp_ep_t *remote)
 {
     size_t len;
-    size_t pos = _len_get(con->rxbuf, &len);
+    size_t pos;
```
smlng: why not add the check ( `if (pkt_len < 2) ...` ) right before the variable declarations? RIOT is using C99 and above so no need to declare all vars up front.

smlng:
why closed?

nmeum:
Oops, that was a mistake on my side.

sys/net/application_layer/asymcute/asymcute.c
```diff
@@ -517,15 +517,20 @@ static void _on_unsuback(asymcute_con_t *con, const uint8_t *data, size_t len)
...
+
+    if (pkt_len < 2) {
```
smlng: I'm (still) in favour of moving this check to the beginning of the function, and thus leave the declaration and assignment of `pos` as it was.

Otherwise this is good to go

smlng: please amend (or squash) directly, so we get this through quickly

PR #10725, opened 2019-01-07
drivers/cc2420: move flush sequence to inline function 

jia200x:
### Contribution description
See https://github.com/RIOT-OS/RIOT/pull/10416#issuecomment-451980544

### Testing procedure
Check it compiles properly. Try to receive packets and check they are received correctly (if FIFO is not flushed, new packets won't be received)

### Issues/PRs references
#10416

### DIFF
--- a/drivers/cc2420/cc2420.c
+++ b/drivers/cc2420/cc2420.c
@@ -158,6 +158,12 @@ void cc2420_tx_exec(cc2420_t *dev)
     }
 }
 
+static inline void _flush_rx_fifo(cc2420_t *dev)
+{
+    cc2420_strobe(dev, CC2420_STROBE_FLUSHRX);
+    cc2420_strobe(dev, CC2420_STROBE_FLUSHRX);
+}
+
 int cc2420_rx(cc2420_t *dev, uint8_t *buf, size_t max_len, void *info)
 {
     (void)info;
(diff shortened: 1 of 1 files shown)


smlng:
I guess its safe to squash already

maribu:
> I guess its safe to squash already

+1

jia200x:
done!

maribu:
If Travis and Murdock don't find any issues, I would just merge this PR untested. I really cannot see how this could break anything - I would bet that the compiler would generate the exact same machine code before and after this PR.

maribu:
@smlng, @PeterKietzmann: Are you fine with untested merge of this PR?

maribu:
OK. I did just compile it with and without the PR and disassembled cc2420.o. The result is the same - so indeed the same machine code is generated. I think this should be as good as testing. (I cannot test because I'm missing the hardware.)

maribu:
Huh! A moment ago all three checks were green - now Codacy is running again... Strange. (But I guess that test will pass again, as no code was change since it was green and I hit merge.)

maribu:
For the record Codacy [says the PR is up to standards](https://app.codacy.com/app/danpetry/RIOT/pullRequest?prid=2834530) now. Next time I'll hit Ctrl+F5 instead of F5 before clicking on merge. Maybe the browser chache served me ill this time.

PR #10726, opened 2019-01-07
nanocoap: fix sock_udp return value checks

nmeum:
### Contribution description

Both `sock_udp_create` and `sock_udp_recv` return negative `errno(3)` values on error.  Thus errors weren't detected properly.

### Testing procedure

Read the doxygen documentation of `sock_udp_create` and `sock_udp_recv`.

### Issues/PRs references

None.

### DIFF
--- a/sys/net/application_layer/nanocoap/sock.c
+++ b/sys/net/application_layer/nanocoap/sock.c
@@ -128,17 +128,17 @@ int nanocoap_server(sock_udp_ep_t *local, uint8_t *buf, size_t bufsize)
     }
 
     ssize_t res = sock_udp_create(&sock, local, NULL, 0);
-    if (res == -1) {
+    if (res != 0) {
         return -1;
     }
 
     while (1) {
         res = sock_udp_recv(&sock, buf, bufsize, -1, &remote);
-        if (res == -1) {
+        if (res < 0) {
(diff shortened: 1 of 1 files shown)


nmeum:
Accidentally closed this.

sys/net/application_layer/nanocoap/sock.c
```diff
@@ -128,13 +128,13 @@ int nanocoap_server(sock_udp_ep_t *local, uint8_t *buf, size_t bufsize)
...
     ssize_t res = sock_udp_create(&sock, local, NULL, 0);
-    if (res == -1) {
+    if (res != 0) {
         return -1;
     }
 
     while (1) {
         res = sock_udp_recv(&sock, buf, bufsize, -1, &remote);
-        if (res == -1) {
+        if (res < 0) {
```
smlng: while you are at it: the else below could be changed into a `else if (res > 0) {` bc if `res == 0` everything is good but nothing was received so no need to call `coap_parse`.

nmeum: I initially didn't want to change the handling of the zero return value. I just changed this nonetheless.

smlng: thx

PR #10737, opened 2019-01-09
boards/esp32-wroom-32: configure saul button as inverted

aabadie:
### Contribution description

This PR configures the saul gpio BOOT button of the esp32-wroom-32 board with the inverted flag.

### Testing procedure

1. Build and flash the `examples/default` on the board:
```
$ make BOARD=esp32-wroom-32 -C examples/default flash term
```
2. Check the value returned by saul read is correct:
```
> saul read 0  (BOOT not pressed)
Reading from #0 (BOOT|SENSE_BTN)
Data:	          0
> saul read 0  (BOOT pressed)
Reading from #0 (BOOT|SENSE_BTN)
Data:	          1
```

### Issues/PRs references

None

### DIFF
--- a/boards/esp32-olimex-evb/include/gpio_params.h
+++ b/boards/esp32-olimex-evb/include/gpio_params.h
@@ -41,7 +41,7 @@ static const  saul_gpio_params_t saul_gpio_params[] =
         .name = "BUT1",
         .pin = BUTTON0_PIN,
         .mode = GPIO_IN,
-        .flags = 0
+        .flags = SAUL_GPIO_INVERTED
     },
 };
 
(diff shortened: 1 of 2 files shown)


gschorcht:
Thanks. Was not clear to me that the INVERT flag is also applied to buttons, but is seems to be logical that pressing a button should result into logical 1. It is the same for https://github.com/RIOT-OS/RIOT/blob/484572f881b170ce4a23c6509aec5cdffca2e1ff/boards/esp32-olimex-evb/include/gpio_params.h#L40-L45. Would you change it with your PR too? I can test it then.

aabadie:
> Would you change it with your PR too?

I'll do that.

aabadie:
@gschorcht, I also updated the `esp32-olimex-evb` board. I don't have it for testing but I'm trusting you here.

aabadie:
@gschorcht, so were you able to test ?

gschorcht:
@aabadie I'm testing ...

PR #8610, opened 2018-02-21
[WIP] examples: add Wakaama/LwM2M example client

moenoel:
## Contribution description

This pull request is a followup to #8148 and contains the following changes:

* Bumps pkg/wakaama version
* Fixes include path in pkg/wakaama/Makefile.include
* Adds patches to fix warnings in pkg/wakaama example code
* Makefile copies some code from pkg/wakaama client example
* Ported wakaama/examples/client to RIOT/~~GNRC~~ sock API
* Adds a patch to fix a memory alignment problem to pkg/wakaama
* Comes with support for sensors that return float or bool values
    * Sensors can be automatically added from the SAUL registry

The big change is obviously the port of the example client, which supports bootstrapping, DTLS (depends on TinyDTLS) and comes with the mandatory LwM2M objects/resources, as defined in the LwM2M specification (*server*, *security* and *access control* and *device*).

I tested this with with the [Eclipse Leshan](https://github.com/eclipse/leshan#test-leshan-demos-locally) demo server and demo bootstrap server with and without DTLS (see readme file for instructions). Currently only PSK is supported/tested for DTLS.

~~As I currently don't have a board to test with, I only ran it on native so far.~~

Tested on native, stm32f4discovery, pba-d-01-kw2x and kw41z-mini.

I've reused `BOARD_BLACKLIST` and `BOARD_INSUFFICIENT_MEMORY` from `examples/dtls-echo` since I'm not certain how to (easily) determine which boards have enough resources to run this (~~especially since Wakaama makes heavy use of `malloc()`~~ this is not an issue anymore, since moving to memarray (#7651). Though TinyDTLS still uses `malloc()`).

---

## Known issues

* Objects that are "too big" can't be requested whole, as the buffers are not large enough. Increasing `DTLS_MAX_BUF` in the Makefile or `RCVD_SIZE` (depending on whether DTLS is enabled) in 
`lwm2mclient.c` can fix this.

* When a NAT timeout is set (`DTLS_NAT_TIMEOUT > 0` in `connection.h`), the re-handshake will sometimes fail.

* When `LWM2M_WITH_LOGS` is defined, some boards (e.g. iotlab-m3) die from a stack overflow.

* [Possible timer issues](https://github.com/RIOT-OS/RIOT/pull/8610#issuecomment-393534813) resulting in DTLS handshake failures.

---

Possible next steps for this are:
* ~~Add ECC support for DTLS~~
    * Done but untested as of yet
[...]

PR #10738, opened 2019-01-09
pkg/wakaama: bump version

jia200x:
### Contribution description
This PR is a part of a rework of #8610 by @moenoel (see https://github.com/RIOT-OS/RIOT/pull/8610#issuecomment-451982208). 
This PR:

- Bumps pkg/wakaama version (LWM2M)
- Fixes include path in pkg/wakaama/Makefile.include
- Adds patches to fix warnings in pkg/wakaama example code
- Makefile copies some code from pkg/wakaama client example
- Adds a patch to fix a memory alignment problem to pkg/wakaama
**Update**
- Adds a patch to fix path issues
- Adds a patch to fix access control object

### Testing procedure
Hard to test without an example application, which will be introduced with the rework of #8610).
However it's possible to check it compile with no errors. E.g
```
USEPKG=wakaama make all
```

### Issues/PRs references
#8610

### DIFF
--- a/pkg/wakaama/patches/0014-Use-lwm2m_strdup-instead-of-strdup.patch
+++ b/pkg/wakaama/patches/0014-Use-lwm2m_strdup-instead-of-strdup.patch
@@ -1,0 +1,25 @@
+From 47d78e80ceb80a06e0cb9b0b7f7c959d04bd2c22 Mon Sep 17 00:00:00 2001
+From: Christian Manal <moenoel@informatik.uni-bremen.de>
+Date: Mon, 5 Feb 2018 15:52:10 +0100
+Subject: [PATCH] Use lwm2m_strdup() instead of strdup()
+
+---
+ examples/client/object_security.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/examples/client/object_security.c b/examples/client/object_security.c
+index bfb5f85..d91f6e5 100644
+--- a/examples/client/object_security.c
++++ b/examples/client/object_security.c
+@@ -540,7 +540,7 @@ lwm2m_object_t * get_security_object(int serverId,
+             targetP->securityMode = LWM2M_SECURITY_MODE_PRE_SHARED_KEY;
+             if (bsPskId)
+             {
+-                targetP->publicIdentity = strdup(bsPskId);
++                targetP->publicIdentity = lwm2m_strdup(bsPskId);
+                 targetP->publicIdLen = strlen(bsPskId);
+             }
+             if (pskLen > 0)
+-- 
+2.1.4
+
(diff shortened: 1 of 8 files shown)


pkg/wakaama/patches/0014-Use-lwm2m_strdup-instead-of-strdup.patch
```diff
@@ -0,0 +1,25 @@
...
+
+diff --git a/examples/client/object_security.c b/examples/client/object_security.c
+index bfb5f85..d91f6e5 100644
+--- a/examples/client/object_security.c
++++ b/examples/client/object_security.c
+@@ -540,7 +540,7 @@ lwm2m_object_t * get_security_object(int serverId,
+             targetP->securityMode = LWM2M_SECURITY_MODE_PRE_SHARED_KEY;
+             if (bsPskId)
+             {
+-                targetP->publicIdentity = strdup(bsPskId);
```
jia200x: @moenoel do you remember what was the reason for using `lwm2m_strdup` instead of `strdup`?

pkg/wakaama/patches/0015-fixed-alignment-problem-in-lwm2m_data_t.patch
```diff
@@ -0,0 +1,32 @@
...
+
+diff --git a/core/liblwm2m.h b/core/liblwm2m.h
+index 3f16a6d..53616bf 100644
+--- a/core/liblwm2m.h
++++ b/core/liblwm2m.h
+@@ -299,7 +299,6 @@ typedef struct _lwm2m_data_t lwm2m_data_t;
+ 
+ struct _lwm2m_data_t
+ {
+-    lwm2m_data_type_t type;
```
jia200x: @moenoel is this safe? E.g wakaama code could eventually cast `_lwm2m_data_t` to `lwm2m_data_type_t` because address_of(_lwm2m_data_t) == addess_of(_lwm2m_data_t->type). With this change, address_of(_lwm2m_data_t) == address_of(_lwm2m_data_t->id).

Do you remember the reason behind this?

moenoel: Not entirely sure if it's safe, though I didn't encounter any obvious problems running it. The reason for that change was to get rid of a memory alignment issue on the ARM boards we used. 

IIRC `lwm2m_data_type_t` is an enum and apparently enums can have strange widths, so all the following struct members got misaligned, leading to hard faults. Making the enum member the last one fixed that for us.

pkg/wakaama/patches/0014-Use-lwm2m_strdup-instead-of-strdup.patch
moenoel: One of the reasons was consistency, as Wakaama defines prototypes for a few functions that it expects the specific implementation to fill in (e.g. `lwm2m_malloc()` and `lwm2m_free()`) and it just seemed to make sense to use those wrappers throughout the Wakaama related code.

There was at least one other reason, but I can't for the life of me remember what that was specifically.

leandrolanzieri:
I added two fixups that bump to the latest release and add a fix on the access control object that [hasn't yet been merged to upstream](https://github.com/eclipse/wakaama/pull/427)

PeterKietzmann:
I can confirm that any application builds with `USEPKG=wakaama` for native and samr21-xpro. The pkg is cloned and patches were applied without conflicts. @leandrolanzieri don't you have a test application at hand?

PR #10742, opened 2019-01-10
gnrc_netif_*: use the `gnrc_netif_t::pid` member instead of the pid of the current thread in netif header

gschorcht:
### Contribution description

Due to the interrupt context problem, the receive sequence of defined as follows.

gnrc_netif_1

It is not an UML compliant message sequence chart by intention. Activities show the active context.

In `netif::_recv`, the `thread_getpid` function is used to set the `netif` pid in `gnrc_netif_hdr`. This works since `_recv` is executed in the `netif` thread context.

However, if the `netdev` polls for new data using its own thread context, data are not fetched in interrupt context and there is no necessity to send an `NETDEV_EVENT_ISR` event to the `netif` first. Rather, `netdev` can call the event callback of `netif` with NETDEV_EVENT_RX_COMPLETE directly.

gnrc_netif_2

This has two advantages:
- it is faster
- there is no need for an additional buffer in `netdev`, data can be copied directly from hardware to the `netif` buffer.

However, since `netif::_recv` is executed in the `netdev` thread context, the `thread_getpid` function returns the wrong pid.

Since the `netif` has a member `pid` it should be used here.  Both approaches are possible with the change in this PR

### Testing procedure

Use any board with Ethernet interface or IEEE802.15.4 interfaces and use `example/gnrc_networking` to ping them.


gschorcht:
@miri64 Thanks

PR #10743, opened 2019-01-10
RFC: UART enhanced settings

yegorich:
### Contribution description

As suggested in PR #5899 add a routine uart_mode() that will setup databits, stopbits and parity at runtime. 

`uart.h` provides a set of enums defining these settings and each platform will override them to specify values corresponding to its configuration registers.

The idea behind the enums is to specify default settings i.e. 8N1 through the 0 value item. Invoking `uart_mode(uart, 0, 0, 0)` will setup 8N1 mode.

9-bit mode was excluded as UART API is working with uin8_t data units.

As a proof of concept uart_mode was implemented for STM32 UART and tested on Nucleo-F411re.

### Testing procedure

periph_uart test was extended with `mode` command that will be invoked directly after `init` command.  Below some examples:

- `mode 1 8 e 1` - configures 8E1 mode, i.e. 8 databits, even parity and 1 stopbit
- `mode 1 7 o 2` - configures 7O2 mode, i.e. 7 databits, odd parity and 2 stopbits

### Issues/PRs references

This PR reimplements PR #5899 for general use.

### DIFF
--- a/cpu/stm32_common/include/periph_cpu_common.h
+++ b/cpu/stm32_common/include/periph_cpu_common.h
@@ -371,6 +371,56 @@ typedef enum {
 } uart_type_t;
 
 /**
+ * @brief   Invalid UART mode mask
+ *
+ * This mask is also used to force data_bits_t to be uint32_t type
+ * since it may be assigned a uint32_t variable in uart_mode
+ */
+#define UART_INVALID_MODE   (0x8000000)
+
+/**
+ * @brief   Override parity values
+ * @{
+ */
+#define HAVE_UART_PARITY_T
+typedef enum {
+   UART_PARITY_NONE = 0,                               /**< no parity */
+   UART_PARITY_EVEN = USART_CR1_PCE,                   /**< even parity */
+   UART_PARITY_ODD = (USART_CR1_PCE | USART_CR1_PS),   /**< odd parity */
+   UART_PARITY_MARK = UART_INVALID_MODE | 4,           /**< not supported */
+   UART_PARITY_SPACE = UART_INVALID_MODE  | 5          /**< not supported */
+} uart_parity_t;
+/** @} */
+
+/**
+ * @brief   Override data bits length values
+ * @{
+ */
+#define HAVE_UART_DATA_BITS_T
+typedef enum {
+    UART_DATA_BITS_5 = UART_INVALID_MODE | 1,   /**< not supported */
+    UART_DATA_BITS_6 = UART_INVALID_MODE | 2,   /**< not supported unless parity is set */
+#if defined(USART_CR1_M1)
+    UART_DATA_BITS_7 = USART_CR1_M1,            /**< 7 data bits */
+#else
+    UART_DATA_BITS_7 = UART_INVALID_MODE | 3,   /**< not supported unless parity is set */
+#endif
+    UART_DATA_BITS_8 = 0,                       /**< 8 data bits */
+} uart_data_bits_t;
+/** @} */
+
+/**
+ * @brief   Override stop bits length values
+ * @{
+ */
+#define HAVE_UART_STOP_BITS_T
+typedef enum {
+   UART_STOP_BITS_1 = 0,                  /**< 1 stop bit */
+   UART_STOP_BITS_2 = USART_CR2_STOP_1,   /**< 2 stop bits */
+} uart_stop_bits_t;
+/** @} */
+
+/**
  * @brief   Structure for UART configuration data
  */
 typedef struct {
--- a/cpu/stm32_common/periph/uart.c
+++ b/cpu/stm32_common/periph/uart.c
@@ -53,8 +53,15 @@
 
 /**
  * @brief   Allocate memory to store the callback functions
+ *
+ * Extend standard uart_isr_ctx_t with data_mask field. This is needed
+ * in order to mask parity bit.
  */
-static uart_isr_ctx_t isr_ctx[UART_NUMOF];
+static struct {
+    uart_rx_cb_t rx_cb;   /**< data received interrupt callback */
+    void *arg;            /**< argument to both callback routines */
+    uint8_t data_mask;    /**< mask applied to the data register */
+} isr_ctx[UART_NUMOF];
 
 static inline USART_TypeDef *dev(uart_t uart)
 {
@@ -105,8 +112,9 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
     assert(uart < UART_NUMOF);
 
     /* save ISR context */
-    isr_ctx[uart].rx_cb = rx_cb;
-    isr_ctx[uart].arg   = arg;
+    isr_ctx[uart].rx_cb     = rx_cb;
+    isr_ctx[uart].arg       = arg;
+    isr_ctx[uart].data_mask = 0xFF;
 
     uart_init_pins(uart, rx_cb);
 
@@ -153,6 +161,53 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
 
     return UART_OK;
 }
+
+#ifdef MODULE_PERIPH_UART_MODECFG
+int uart_mode(uart_t uart, uart_data_bits_t data_bits, uart_parity_t parity,
+              uart_stop_bits_t stop_bits)
+{
+    assert(uart < UART_NUMOF);
+
+    isr_ctx[uart].data_mask = 0xFF;
+
+    if (parity) {
+        switch (data_bits) {
+            case UART_DATA_BITS_6:
+                data_bits = UART_DATA_BITS_7;
+                isr_ctx[uart].data_mask = 0x3F;
+                break;
+            case UART_DATA_BITS_7:
+                data_bits = UART_DATA_BITS_8;
+                isr_ctx[uart].data_mask = 0x7F;
+                break;
+            case UART_DATA_BITS_8:
+                data_bits = USART_CR1_M;
+                break;
+            default:
+                return UART_NOMODE;
+        }
+    }
+    if ((data_bits & UART_INVALID_MODE) || (parity & UART_INVALID_MODE)) {
+        return UART_NOMODE;
+    }
+
+#ifdef USART_CR1_M1
+    if (!(dev(uart)->ISR & USART_ISR_TC)) {
+        return UART_INTERR;
+    }
+    dev(uart)->CR1 &= ~(USART_CR1_UE | USART_CR1_TE);
+    dev(uart)->CR1 &= ~USART_CR1_M1;
+#endif
+
+    dev(uart)->CR2 &= ~USART_CR2_STOP;
+    dev(uart)->CR1 &= ~(USART_CR1_PS | USART_CR1_PCE | USART_CR1_M);
+
+    dev(uart)->CR2 |= stop_bits;
+    dev(uart)->CR1 |= (USART_CR1_UE | USART_CR1_TE | data_bits | parity);
+
+    return UART_OK;
+}
+#endif /* MODULE_PERIPH_UART_MODECFG */
 
 static inline void uart_init_usart(uart_t uart, uint32_t baudrate)
 {
@@ -301,7 +356,8 @@ static inline void irq_handler(uart_t uart)
     uint32_t status = dev(uart)->ISR;
 
     if (status & USART_ISR_RXNE) {
-        isr_ctx[uart].rx_cb(isr_ctx[uart].arg, (uint8_t)dev(uart)->RDR);
+        isr_ctx[uart].rx_cb(isr_ctx[uart].arg,
+                            (uint8_t)dev(uart)->RDR & isr_ctx[uart].data_mask);
     }
     if (status & USART_ISR_ORE) {
         dev(uart)->ICR |= USART_ICR_ORECF;    /* simply clear flag on overrun */
@@ -312,7 +368,8 @@ static inline void irq_handler(uart_t uart)
     uint32_t status = dev(uart)->SR;
 
     if (status & USART_SR_RXNE) {
-        isr_ctx[uart].rx_cb(isr_ctx[uart].arg, (uint8_t)dev(uart)->DR);
+        isr_ctx[uart].rx_cb(isr_ctx[uart].arg,
+                            (uint8_t)dev(uart)->DR & isr_ctx[uart].data_mask);
     }
     if (status & USART_SR_ORE) {
         /* ORE is cleared by reading SR and DR sequentially */
--- a/drivers/include/periph/uart.h
+++ b/drivers/include/periph/uart.h
@@ -112,6 +112,41 @@ enum {
 };
 
 /**
+ * @brief   Definition of possible parity modes
+ */
+#ifndef HAVE_UART_PARITY_T
+typedef enum {
+   UART_PARITY_NONE,   /**< no parity */
+   UART_PARITY_EVEN,   /**< even parity */
+   UART_PARITY_ODD,    /**< odd parity */
+   UART_PARITY_MARK,   /**< mark parity */
+   UART_PARITY_SPACE   /**< space parity */
+} uart_parity_t;
+#endif
+
+/**
+ * @brief   Definition of possible data bits lengths in a UART frame
+ */
+#ifndef HAVE_UART_DATA_BITS_T
+typedef enum {
+    UART_DATA_BITS_5,   /**< 5 data bits */
+    UART_DATA_BITS_6,   /**< 6 data bits */
+    UART_DATA_BITS_7,   /**< 7 data bits */
+    UART_DATA_BITS_8,   /**< 8 data bits */
+} uart_data_bits_t;
+#endif
+
+/**
+ * @brief   Definition of possible stop bits lengths in a UART frame
+ */
+#ifndef HAVE_UART_STOP_BITS_T
+typedef enum {
+   UART_STOP_BITS_1,   /**< 1 stop bit */
+   UART_STOP_BITS_2,   /**< 2 stop bits */
+} uart_stop_bits_t;
+#endif
+
+/**
  * @brief   Initialize a given UART device
  *
  * The UART device will be initialized with the following configuration:
@@ -136,6 +171,20 @@ enum {
  * @return                  UART_INTERR on other errors
  */
 int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg);
+
+/**
+ * @brief   Setup parity, data and stop bits for a given UART device
+ *
+ * @param[in] uart          UART device to initialize
+ * @param[in] data_bits     number of data bits in a UART frame
+ * @param[in] parity        parity mode
+ * @param[in] stop_bits     number of stop bits in a UART frame
+ *
+ * @return                  UART_OK on success
+ * @return                  UART_NOMODE on other errors
+ */
+int uart_mode(uart_t uart, uart_data_bits_t data_bits, uart_parity_t parity,
+              uart_stop_bits_t stop_bits);
 
 /**
  * @brief   Write data from the given buffer to the specified UART device
--- a/tests/periph_uart/Makefile
+++ b/tests/periph_uart/Makefile
@@ -2,8 +2,9 @@ include ../Makefile.tests_common
 
 BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-uno nucleo-f031k6
 
-FEATURES_REQUIRED = periph_uart
-FEATURES_OPTIONAL = periph_lpuart  # STM32 L0 and L4 provides lpuart support
+FEATURES_REQUIRED += periph_uart
+FEATURES_OPTIONAL += periph_lpuart  # STM32 L0 and L4 provides lpuart support
+FEATURES_OPTIONAL += periph_uart_modecfg
 
 USEMODULE += shell
 USEMODULE += xtimer
--- a/tests/periph_uart/main.c
+++ b/tests/periph_uart/main.c
@@ -53,6 +53,15 @@ static uart_ctx_t ctx[UART_NUMOF];
 static kernel_pid_t printer_pid;
 static char printer_stack[THREAD_STACKSIZE_MAIN];
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+static uart_data_bits_t data_bits_lut[] = { UART_DATA_BITS_5, UART_DATA_BITS_6,
+                                            UART_DATA_BITS_7, UART_DATA_BITS_8 };
+static int data_bits_lut_len = sizeof(data_bits_lut)/sizeof(data_bits_lut[0]);
+
+static uart_stop_bits_t stop_bits_lut[] = { UART_STOP_BITS_1, UART_STOP_BITS_2 };
+static int stop_bits_lut_len = sizeof(stop_bits_lut)/sizeof(stop_bits_lut[0]);
+#endif
+
 static int parse_dev(char *arg)
 {
     unsigned dev = atoi(arg);
@@ -142,7 +151,7 @@ static int cmd_init(int argc, char **argv)
         return 1;
     }
     else if (res != UART_OK) {
-        puts("Error: Unable to initialize UART device\n");
+        puts("Error: Unable to initialize UART device");
         return 1;
     }
     printf("Success: Initialized UART_DEV(%i) at BAUD %"PRIu32"\n", dev, baud);
@@ -153,6 +162,74 @@ static int cmd_init(int argc, char **argv)
 
     return 0;
 }
+
+#ifdef MODULE_PERIPH_UART_MODECFG
+static int cmd_mode(int argc, char **argv)
+{
+    int dev, data_bits_arg, stop_bits_arg;
+    uart_data_bits_t data_bits;
+    uart_parity_t  parity;
+    uart_stop_bits_t  stop_bits;
+
+    if (argc < 5) {
+        printf("usage: %s <dev> <data bits> <parity> <stop bits>\n", argv[0]);
+        return 1;
+    }
+
+    dev = parse_dev(argv[1]);
+    if (dev < 0) {
+        return 1;
+    }
+
+    data_bits_arg = atoi(argv[2]) - 5;
+    if (data_bits_arg >= 0 && data_bits_arg < data_bits_lut_len) {
+        data_bits = data_bits_lut[data_bits_arg];
+    }
+    else {
+        printf("Error: Invalid number of data_bits (%i).\n", data_bits_arg + 5);
+        return 1;
+    }
+
+    argv[3][0] &= ~0x20;
+    switch (argv[3][0]) {
+        case 'N':
+            parity = UART_PARITY_NONE;
+            break;
+        case 'E':
+            parity = UART_PARITY_EVEN;
+            break;
+        case 'O':
+            parity = UART_PARITY_ODD;
+            break;
+        case 'M':
+            parity = UART_PARITY_MARK;
+            break;
+        case 'S':
+            parity = UART_PARITY_SPACE;
+            break;
+        default:
+            printf("Error: Invalid parity (%c).\n", argv[3][0]);
+            return 1;
+    }
+
+    stop_bits_arg = atoi(argv[4]) - 1;
+    if (stop_bits_arg >= 0 && stop_bits_arg < stop_bits_lut_len) {
+        stop_bits = stop_bits_lut[stop_bits_arg];
+    }
+    else {
+        printf("Error: Invalid number of stop bits (%i).\n", stop_bits_arg + 1);
+        return 1;
+    }
+
+    if (uart_mode(UART_DEV(dev), data_bits, parity, stop_bits) != UART_OK) {
+        puts("Error: Unable to apply UART settings");
+        return 1;
+    }
+    printf("Success: Successfully applied UART_DEV(%i) settings\n", dev);
+
+    return 0;
+}
+#endif /* MODULE_PERIPH_UART_MODECFG */
 
 static int cmd_send(int argc, char **argv)
 {
@@ -177,6 +254,9 @@ static int cmd_send(int argc, char **argv)
 
 static const shell_command_t shell_commands[] = {
     { "init", "Initialize a UART device with a given baudrate", cmd_init },
+#ifdef MODULE_PERIPH_UART_MODECFG
+    { "mode", "Setup data bits, stop bits and parity for a given UART device", cmd_mode },
+#endif
     { "send", "Send a string through given UART device", cmd_send },
     { NULL, NULL, NULL }
 };
--- a/cpu/stm32_common/Makefile.features
+++ b/cpu/stm32_common/Makefile.features
@@ -1,6 +1,7 @@
 FEATURES_PROVIDED += periph_cpuid
 FEATURES_PROVIDED += periph_gpio periph_gpio_irq
 FEATURES_PROVIDED += puf_sram
+FEATURES_PROVIDED += periph_uart_modecfg
 
 ifneq (,$(filter $(BOARDS_WITHOUT_HWRNG),$(BOARD)))
   FEATURES_PROVIDED := $(filter-out periph_hwrng,$(FEATURES_PROVIDED))
--- a/tests/periph_uart/tests/periph_uart_if.py
+++ b/tests/periph_uart/tests/periph_uart_if.py
@@ -19,6 +19,11 @@ class PeriphUartIf(DutShell):
         """Initialize DUT's UART."""
         return self.send_cmd("init {} {}".format(dev, baud))
 
+    def uart_mode(self, dev, data_bits, parity, stop_bits):
+        """Setup databits, parity and stopbits."""
+        return self.send_cmd(
+            "mode {} {} {} {}".format(dev, data_bits, parity, stop_bits))
+
     def uart_send_string(self, dev, test_string):
         """Send data via DUT's UART."""
         return self.send_cmd("send {} {}".format(dev, test_string))
--- a/tests/periph_uart/tests/test.py
+++ b/tests/periph_uart/tests/test.py
@@ -17,7 +17,7 @@ import random
 import string
 
 from periph_uart_if import PeriphUartIf
-from if_lib import PhilipIf
+from riot_pal import LLMemMapIf, PHILIP_MEM_MAP_PATH
 
 
 def kwexpect(val1, val2, level="WARN"):
@@ -120,7 +120,7 @@ def setup_test(bpt, t):
     bpt.reset_mcu()
     t.manual_test("Reset BPT")
 
-    bpt.set_uart_baud(115200)
+    bpt.write_reg('uart.baud', 115200)
     bpt.execute_changes()
     t.manual_test("BPT: setup default baudrate")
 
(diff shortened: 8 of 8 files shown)


cpu/stm32_common/periph/uart.c
```diff
@@ -139,6 +141,47 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+    cr1 = dev(uart)->CR1;
+    cr2 = dev(uart)->CR2;
+
+    if (parity) {
+	/* enable parity */
+        cr1 |= USART_CR1_PCE;
+        /* enable 9-bit mode for 8 databits or
+           mask the MSB for 7 databits mode */
+        if (databits == UART_DATABITS_8)
+            cr1 |= USART_CR1_M;
```
MrKevinWeiss: It seems there are some differences between the uart registers similar to the i2c registers the stm32l1xxx and stm32l0xxx are good examples.
In the l1 version the M register cannot be modified while transmitting (maybe add some protection for that) and the l0 has both a M0 and M1 bit (allowing data of 7,8,9).

Also the L0 says 'This bit can only be written when the USART is disabled (UE=0)' so maybe just shut off to change the mode.

[L1 example](https://www.st.com/content/ccc/resource/technical/document/reference_manual/cc/f9/93/b2/f0/82/42/57/CD00240193.pdf/files/CD00240193.pdf/jcr:content/translations/en.CD00240193.pdf) page 740
[L0 example](https://www.st.com/content/ccc/resource/technical/document/reference_manual/2f/b9/c6/34/28/29/42/d2/DM00095744.pdf/files/DM00095744.pdf/jcr:content/translations/en.DM00095744.pdf) page 801

cpu/stm32_common/periph/uart.c
```diff
@@ -294,7 +337,7 @@ static inline void irq_handler(uart_t uart)
     uint32_t status = dev(uart)->ISR;
 
     if (status & USART_ISR_RXNE) {
-        isr_ctx[uart].rx_cb(isr_ctx[uart].arg, (uint8_t)dev(uart)->RDR);
+        isr_ctx[uart].rx_cb(isr_ctx[uart].arg, (uint8_t)dev(uart)->RDR & data_mask);
```
MrKevinWeiss: Does the 8th bit remain in the buffer if the config says only 7?

kaspar030:
Did you consider a combined value for all options?

E.g.,
```
typedef uint16_t uart_mode_t
#define UART_PARITY_OFFSET 0
#define UART_PARITY_MASK 0x3
#define UART_DATABITS_OFFSET 3
#define UART_DATABITS_MASK 0x3

//...
int parity = mode & UART_PARITY_MASK >> UART_PARITY_OFFSET
int databits = mode & UART_DATABITS_MASK >> UART_DATABITS_OFFSET
```

In addition, define the most used ones as speaking defines:
```
#define UART_MODE_8N1 (UART_DATABITS_8 << UART_DATABITS_OFFSET | UART_PARITY_N << UART_PARITY_OFFSET ...)
```

My main reason would be that this way, uart users actually using modes don't need all the fields:

```
typedef struct  {
 //...
 uint8_t uart_parity;
 uint8_t uart_databits;
 uint8_t uart_stopbits;
} random_driver_params_t;

// vs
typedef struct  {
 //...
 uart_mode_t uart_mode;
} random_driver_params_t;
```

cpu/stm32_common/periph/uart.c
```diff
@@ -139,6 +141,47 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+            || (databits == UART_DATABITS_9 && parity))
+        return UART_NOMODE;
+
+    if (parity == 0x02)
+        return UART_NOMODE;
+
+    cr1 = dev(uart)->CR1;
+    cr2 = dev(uart)->CR2;
+
+    if (parity) {
```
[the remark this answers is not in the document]
yegorich: No need to clear anything as we only change the default values.

cpu/stm32_common/periph/uart.c
yegorich: Yes. Parity bit is the MSB and have to be masked to get the raw data.

yegorich:
@kaspar030 we already had this discussion [here](https://github.com/RIOT-OS/RIOT/pull/5899#issuecomment-432622619) and so I've made this PR according to it.

cpu/stm32_common/periph/uart.c
MrKevinWeiss: Should you #ifdef them around or does everything work out?

cpu/stm32_common/periph/uart.c
MrKevinWeiss: Can you read the register to see if it needs to be masked?  I think that is better than a static global (silly hardware though).

yegorich: I can make this.

yegorich:
v2 changes:

- add support for partial 6-bit and full 7-bit modes to the devices defining `USART_CR1_M1`
- add tests for even and odd parity against PHiLIP hardware

yegorich:
v3 changes:

- convert `uart_mode` to a feature `periph_uart_modecfg`
- move tests to a separate commit

yegorich:
@bluephoton could you test this PR with your setup and report the results here? Thanks.

bluephoton:
@yegorich , I gave it a try but it didn't work! I know that I need to invert RX for my scenario so I added the following code (at the very end before you enable uart):
```
MODIFY_REG(dev(uart)->CR2, USART_CR2_RXINV_Msk, USART_CR2_RXINV);
```
but this didn't help!

I'll look bit more tomorrow, in case its something with my environment.

Here is how I initialize (working case):
```
uart_init_ex(uart, baudrate, UartReceiveCallback, this, false);
uart_configure(uart, UART_9BITS, UART_PARITY_EVEN, UART_STOPBITS_TWO, UART_RX_INVERT);
uart_enable(uart);
```

Here is with new uart_mode (not working case):
```
uart_init(uart, baudrate, UartReceiveCallback, this);
uart_mode(uart, UART_DATABITS_8, UART_PARITY_EVEN, UART_STOPBITS_2);
```

Couple notes on implementation:
- I'm not an expert in this, but not a fan of starting up device, stopping it to configure, then starting it again.
- Not sure why we have UART_DATABITS_9 if we are setting this internally (code could use some comment around this as it's not intuitive)

bluephoton:
@yegorich I found the culprit! Two issues:
- uart_stop() line 171 & 178 should be
```
cr1 &= ~USART_CR1_UE;   // note the ~ should be there in both cases
```
- stopbits seems to be not set correctly, when I change it as follow it works:

```
// cr2 |= stopbits;
cr2 |= USART_CR2_STOP_1;
```
- And of course inverting RX bin was required and you should add while you are at it.

I also recommend replacing all numeric constants with actual macro as defined in CMSIS header file. Also not use _M (mask) and use actual bit value (even if it happen to be the same) like:

```
if (databits == UART_DATABITS_8) {
            //cr1 |= USART_CR1_M;
            cr1 |= USART_CR1_M0;
```

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -370,6 +370,51 @@ typedef enum {
...
+/** @} */
+
+/**
+ * @brief   Override stopbits length values
+ * @{
+ */
+#define HAVE_UART_STOPBITS_T
+typedef enum {
+   UART_STOPBITS_1 = 0,              /**< 1 stopbit */
+   UART_STOPBITS_2 = USART_CR2_STOP, /**< 2 stopbits */
```
bluephoton:
```suggestion
   UART_STOPBITS_2 = USART_CR2_STOP_1, /**< 2 stopbits */
```

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +159,96 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+
+    /* disbale transmit */
+    cr1 &= USART_CR1_TE;
+    dev(uart)->CR1 = cr1;
+
+    /* wait for transmission completion */
+    while (!(dev(uart)->ISR & USART_ISR_TC)) {}
+
+    /* disable UART */
+    cr1 &= USART_CR1_UE;
```
bluephoton:
```suggestion
    cr1 &= ~USART_CR1_UE;
```

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +159,96 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+#ifdef MODULE_PERIPH_UART_MODECFG
+#if defined(USART_CR1_M1)
+static inline void uart_stop(uart_t uart)
+{
+    uint32_t cr1;
+
+    cr1 = dev(uart)->CR1;
+
+    /* disbale transmit */
+    cr1 &= USART_CR1_TE;
```
bluephoton:
```suggestion
    cr1 &= ~USART_CR1_TE;
```

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +159,96 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+    cr1 = dev(uart)->CR1;
+    cr2 = dev(uart)->CR2;
+
+    if (parity) {
+        /* enable parity */
+        cr1 |= USART_CR1_PCE;
+
+        /* setup parity, databits and databits masking */
+        if (databits == UART_DATABITS_8) {
+            cr1 |= USART_CR1_M;
```
bluephoton:
```suggestion
            cr1 |= USART_CR1_M0;
```

yegorich:
v4 changes:

- fix UE and TE bit configuration
- fix 2 stopbits macro

yegorich:
@bluephoton thanks for the testing and bug finding :-) I had only 8-bit capable CPUs so couldn't really test the code with UE and TC bit changes.

`UART_DATABITS_9` mode is for future use. For now, RIOT's UART stack is working with 8-bit arrays. But some UARTs including STM32 allow to write 9 bits into the data register and all these bits will be transferred. The same is true for the reception. Besides STM32 is special because it saves parity bit in the data register and hence requires 9-bit mode for 8E1.

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -370,6 +370,51 @@ typedef enum {
...
+typedef enum {
+    UART_DATABITS_8 = 0,        /**< 8 databits */
+    UART_DATABITS_5 = 0x02,     /**< not supported */
+#if defined(USART_CR1_M1)
+    UART_DATABITS_6 = 3,        /**< 6 databits */
+#else
+    UART_DATABITS_6 = 0x02,     /**< not supported */
+#endif
+    UART_DATABITS_7 = 1,        /**< 7 databits */
+    UART_DATABITS_9 = 0x02      /**< not supported */
```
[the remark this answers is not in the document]
MrKevinWeiss: maybe instead of setting the not supported bits to 0x02 you set it to a macro like ENOTSUPPORTED then when you do you check you can check for that?

cpu/stm32_common/periph/uart.c
bluephoton: Good to know! thanks for the info!

bluephoton:
@yegorich You are welcome :) Thanks for details on 9 bits choice!

yegorich:
v5 changes:

- replace 0x02 with `UART_INVALID_MODE` macro
- add comment about 9-bit mode

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -370,6 +370,52 @@ typedef enum {
...
+#ifdef MODULE_PERIPH_UART_MODECFG
+#define UART_INVALID_MODE 0xFF
+/**
+ * @brief   Override parity values
+ * @{
+ */
+#define HAVE_UART_PARITY_T
+typedef enum {
+   UART_PARITY_NONE = 0,                   /**< no parity */
+   UART_PARITY_EVEN = 1,                   /**< even parity */
```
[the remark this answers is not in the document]
yegorich: ACK

cpu/stm32_common/periph/uart.c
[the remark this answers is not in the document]
MrKevinWeiss: Yup, I suppose that what the community accepts.

cpu/stm32_common/periph/uart.c
MrKevinWeiss: I think you have a handle on it, it is actually nice to use the real macro then the #if STM32F0 | STM32F1 ...

cpu/stm32_common/periph/uart.c
MrKevinWeiss: Nice!

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -370,6 +370,52 @@ typedef enum {
...
+/** @} */
+
+/**
+ * @brief   Override stopbits length values
+ * @{
+ */
+#define HAVE_UART_STOPBITS_T
+typedef enum {
+   UART_STOPBITS_1 = 0,                  /**< 1 stopbit */
+   UART_STOPBITS_2 = USART_CR2_STOP_1,   /**< 2 stopbits */
```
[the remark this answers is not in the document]
MrKevinWeiss: IRL we talked and I think we won't have a 1.5 stop bit.

yegorich:
@bluephoton pin inversion seems to be out of scope for this PR. Could you create an issue with this particular feature request?

bluephoton:
@yegorich a simpler option for supporting inversion (simpler than passing structure) is to add couple functions:
uart_invert_rx(), uart_invert_tx() and define if macros exist.

Would this something you can consider in this release?

yegorich:
@bluephoton do you need to invert the pins at runtime or is this a permanent configuration?

bluephoton:
@yegorich This is permanent, once at initialization and it stays as is.

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +154,43 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
     return UART_OK;
 }
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+int uart_mode(uart_t uart, uart_data_bits_t data_bits, uart_parity_t parity, uart_stop_bits_t stop_bits)
+{
+    assert(uart < UART_NUMOF);
+
+    uint32_t data_bits_holder = data_bits;
```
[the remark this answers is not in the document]
yegorich: Because of this error:

```
cpu/stm32f4/include/vendor/stm32f411xe.h:6695:39: error: large integer implicitly truncated to unsigned type [-Werror=overflow]
 #define USART_CR1_M_Msk               (0x1U << USART_CR1_M_Pos)                /*!< 0x00001000 */
                                       ^
cpu/stm32f4/include/vendor/stm32f411xe.h:6696:39: note: in expansion of macro 'USART_CR1_M_Msk'
 #define USART_CR1_M                   USART_CR1_M_Msk                          /*!<Word length                            */
                                       ^~~~~~~~~~~~~~~
cpu/stm32_common/periph/uart.c:163:17: note: in expansion of macro 'USART_CR1_M'
     data_bits = USART_CR1_M;
                 ^~~~~~~~~~~

```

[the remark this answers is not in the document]
yegorich: Another way to override would be to use `#define UART_INVALID_MODE 0xffffffff`. I'll test this variant.

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -370,6 +370,52 @@ typedef enum {
     STM32_LPUART,           /**< STM32 Low-power UART (LPUART) module type */
 } uart_type_t;
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+#define UART_INVALID_MODE 0xFF
```
MrKevinWeiss: If you just make it a bit this will force the uint32_t for the enum 
`#define UART_INVALID_MODE 0x8000000`
(please add a comment why)

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -370,6 +370,52 @@ typedef enum {
...
+/**
+ * @brief   Override parity values
+ * @{
+ */
+#define HAVE_UART_PARITY_T
+typedef enum {
+   UART_PARITY_NONE = 0,                               /**< no parity */
+   UART_PARITY_EVEN = USART_CR1_PCE,                   /**< even parity */
+   UART_PARITY_ODD = (USART_CR1_PCE | USART_CR1_PS),   /**< odd parity */
+   UART_PARITY_MARK = UART_INVALID_MODE,               /**< not supported */
```
MrKevinWeiss: To be able to differentiate you should set the invalid bit and a unique number.
If you don't have this then when 6e1 and 7e1 appear the same (since the compare is always UART_INVALID_MODE)
```c
#define HAVE_UART_PARITY_T
typedef enum {
   UART_PARITY_NONE = 0,                               /**< no parity */
   UART_PARITY_EVEN = USART_CR1_PCE,                   /**< even parity */
   UART_PARITY_ODD = (USART_CR1_PCE | USART_CR1_PS),   /**< odd parity */
   UART_PARITY_MARK = UART_INVALID_MODE | 4,               /**< not supported */
   UART_PARITY_SPACE = UART_INVALID_MODE  | 5               /**< not supported */
} uart_parity_t;
/** @} */

/**
 * @brief   Override data bits length values
 * @{
 */
#define HAVE_UART_DATA_BITS_T
typedef enum {
   UART_DATA_BITS_8 = 0,                    /**< 8 data bits */
    UART_DATA_BITS_5 = UART_INVALID_MODE | 1,   /**< not supported */
    UART_DATA_BITS_6 = UART_INVALID_MODE | 2,   /**< not supported unless parity is set */
#if defined(USART_CR1_M1)
    UART_DATA_BITS_7 = USART_CR1_M1,        /**< 7 data bits */
#else
    UART_DATA_BITS_7 = UART_INVALID_MODE | 3,   /**< not supported unless parity is set */
#endif
    UART_DATA_BITS_9 = UART_INVALID_MODE | 4    /**< not supported */
} uart_data_bits_t;
/** @} */
```

tests/periph_uart/main.c
```diff
@@ -67,6 +67,87 @@ static int parse_dev(char *arg)
     return dev;
 }
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+static int parse_data_bits(char *arg)
+{
+    unsigned data_bits = atoi(arg);
+    int res;
```
MrKevinWeiss: return type should be data_bits_t

tests/periph_uart/main.c
```diff
@@ -67,6 +67,87 @@ static int parse_dev(char *arg)
...
+            res = -1;
+    }
+
+    return res;
+}
+
+static int parse_stop_bits(char *arg)
+{
+    unsigned stop_bits = atoi(arg);
+    int res;
```
MrKevinWeiss: return type should be stop_bits_t

tests/periph_uart/main.c
```diff
@@ -67,6 +67,87 @@ static int parse_dev(char *arg)
...
+            printf("Error: Invalid number of stop bits (%u).\n", stop_bits);
+            res = -1;
+    }
+
+    return res;
+}
+
+static int parse_parity(char *arg)
+{
+    int res;
```
MrKevinWeiss: I think you see where I am going with this.

tests/periph_uart/main.c
```diff
@@ -67,6 +67,87 @@ static int parse_dev(char *arg)
...
+            printf("Error: Invalid number of data_bits (%u).\n", data_bits);
+            res = -1;
+    }
+
+    return res;
+}
+
+static int parse_stop_bits(char *arg)
+{
+    unsigned stop_bits = atoi(arg);
```
MrKevinWeiss: please use int not unsigned

tests/periph_uart/main.c
```diff
@@ -67,6 +67,87 @@ static int parse_dev(char *arg)
...
+
+    switch (stop_bits) {
+        case 1:
+            res = UART_STOP_BITS_1;
+            break;
+        case 2:
+            res = UART_STOP_BITS_2;
+            break;
+        default:
+            printf("Error: Invalid number of stop bits (%u).\n", stop_bits);
```
MrKevinWeiss: switch this to int

tests/periph_uart/main.c
```diff
@@ -67,6 +67,87 @@ static int parse_dev(char *arg)
     return dev;
 }
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+static int parse_data_bits(char *arg)
+{
+    unsigned data_bits = atoi(arg);
```
MrKevinWeiss: use int because that is what atoi does

tests/periph_uart/main.c
```diff
@@ -67,6 +67,87 @@ static int parse_dev(char *arg)
...
+            res = UART_DATA_BITS_6;
+            break;
+        case 7:
+            res = UART_DATA_BITS_7;
+            break;
+        case 8:
+            res = UART_DATA_BITS_8;
+            break;
+        default:
+            printf("Error: Invalid number of data_bits (%u).\n", data_bits);
```
MrKevinWeiss: update to %i

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +154,43 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+        }
+        else if (data_bits_holder == UART_DATA_BITS_7) {
+            data_bits_holder = UART_DATA_BITS_8;
+        }
+        else if (data_bits_holder == UART_DATA_BITS_6) {
+            data_bits_holder = UART_DATA_BITS_7;
+        }
+    }
+
+    if (data_bits_holder == UART_INVALID_MODE || parity == UART_INVALID_MODE) {
```
MrKevinWeiss: just mask for the invalid bit
`if ((data_bits & UART_INVALID_MODE) || (parity & UART_INVALID_MODE)) `
or even
`if ((data_bits | parity)  & UART_INVALID_MODE)) `

MrKevinWeiss:
And it simplifies the test a bit

tests/periph_uart/main.c
```diff
@@ -154,6 +235,32 @@ static int cmd_init(int argc, char **argv)
...
+    }
+
+    dev = parse_dev(argv[1]);
+    if (dev < 0) {
+        return 1;
+    }
+    res = uart_mode(dev, parse_data_bits(argv[2]), parse_parity(argv[3]),
+                    parse_stop_bits(argv[4]));
+    if (res != UART_OK) {
+        puts("Error: Unable to apply UART settings\n");
```
MrKevinWeiss: no need for the \n when you use puts

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +154,41 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+        else if (data_bits == UART_DATA_BITS_6) {
+            data_bits = UART_DATA_BITS_7;
+        }
+    }
+    if ((data_bits & UART_INVALID_MODE) || (parity & UART_INVALID_MODE)) {
+        return UART_NOMODE;
+    }
+
+#ifdef USART_CR1_M1
+    if (dev(uart)->ISR & USART_ISR_TC) {
```
MrKevinWeiss: This should actually be `if (!(dedev(uart)->ISR & USART_ISR_TC)) {

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +154,41 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+        return UART_INTERR;
+    }
+    dev(uart)->CR1 &= ~(USART_CR1_UE | USART_CR1_TE);
+#endif
+
+    dev(uart)->CR1 |= (USART_CR1_UE | USART_CR1_TE | data_bits | parity);
+    dev(uart)->CR2 |= stop_bits;
```
MrKevinWeiss: Maybe move this above the CR1 so that it can be set when the uart is disabled

MrKevinWeiss:
@aabadie @haukepetersen, this is pretty close to getting merged.  Now is the time to speak up since I am pretty happy after the changes are applied.

cpu/stm32_common/periph/uart.c
yegorich: ACK

cpu/stm32_common/periph/uart.c
yegorich: ACK

tests/periph_uart/main.c
yegorich: Removed

MrKevinWeiss:
ACK, looks and works well now.  I will see if anyone else want to say anything before I merge but I think we can get this in for this release!

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +154,41 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+
+#ifdef USART_CR1_M1
+    if (!(dev(uart)->ISR & USART_ISR_TC)) {
+        return UART_INTERR;
+    }
+    dev(uart)->CR1 &= ~(USART_CR1_UE | USART_CR1_TE);
+#endif
+
+    dev(uart)->CR2 |= stop_bits;
+    dev(uart)->CR1 |= (USART_CR1_UE | USART_CR1_TE | data_bits | parity);
```
haukepetersen: why tough the `UE` and `TE` bits here? Also as stated above, the data length and parity bits should be reset before setting them...

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +154,41 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
     return UART_OK;
 }
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+int uart_mode(uart_t uart, uart_data_bits_t data_bits, uart_parity_t parity,
+              uart_stop_bits_t stop_bits)
+{
+    assert(uart < UART_NUMOF);
+
```
haukepetersen: don't remember to a 100% anymore: but don't we need to (temporarily) disable the UART peripheral before being able to change its mode paramters?

cpu/stm32_common/periph/uart.c
```diff
@@ -292,6 +327,19 @@ void uart_poweroff(uart_t uart)
 #endif
 }
 
+uint8_t _data_mask(uart_t uart)
```
haukepetersen: needs to be declared `static`

haukepetersen: further: I think it would be better to calculate these masks during configuration and save them in the context struct -> doing this evaluation in the **hot interrupt code path** seems not very nice to me.

drivers/include/periph/uart.h
```diff
@@ -111,6 +111,44 @@ enum {
...
+/**
+ * @brief   Definition of possible data bits lengths in a UART frame
+ */
+#ifndef HAVE_UART_DATA_BITS_T
+typedef enum {
+    UART_DATA_BITS_8,   /**< 8 data bits */
+    UART_DATA_BITS_5,   /**< 5 data bits */
+    UART_DATA_BITS_6,   /**< 6 data bits */
+    UART_DATA_BITS_7,   /**< 7 data bits */
+    UART_DATA_BITS_9    /**< 9 data bits. For now, there is no interface
```
haukepetersen: remove this option. Once/if we ever want to enable our interface to send more then 8 bit, we can add this value to the enum then!

drivers/include/periph/uart.h
```diff
@@ -111,6 +111,44 @@ enum {
...
+   UART_PARITY_SPACE   /**< space parity */
+} uart_parity_t;
+#endif
+
+/**
+ * @brief   Definition of possible data bits lengths in a UART frame
+ */
+#ifndef HAVE_UART_DATA_BITS_T
+typedef enum {
+    UART_DATA_BITS_8,   /**< 8 data bits */
```
haukepetersen: why not put these in some coherent order? Don't care if incremental or decremental, but some consistent order :-)

drivers/include/periph/uart.h
```diff
@@ -137,6 +175,20 @@ enum {
  */
 int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg);
 
+/**
+ * @brief   Setup parity, data and stop bits for a given UART device
+ *
+ * @param[in] uart          UART device to initialize
+ * @param[in] data_bits      number of data bits in a UART frame
```
haukepetersen: indention broken

drivers/include/periph/uart.h
```diff
@@ -137,6 +175,20 @@ enum {
...
+ * @param[in] parity        parity mode
+ * @param[in] stop_bits      number of stop bits in a UART frame
+ *
+ * @return                  UART_OK on success
+ * @return                  UART_NODEV on invalid UART device
+ * @return                  UART_NOMODE on other errors
+ */
+int uart_mode(uart_t uart, uart_data_bits_t data_bits, uart_parity_t parity, uart_stop_bits_t stop_bits);
```
haukepetersen: line length broken

tests/periph_uart/Makefile
```diff
@@ -4,6 +4,7 @@ BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-uno nucleo-f031k6
 
 FEATURES_REQUIRED = periph_uart
 FEATURES_OPTIONAL = periph_lpuart  # STM32 L0 and L4 provides lpuart support
```
haukepetersen: not really part of this PR, but we might want to fix the above lines to use `+=` while we are add it...

tests/periph_uart/main.c
```diff
@@ -57,16 +57,97 @@ static int parse_dev(char *arg)
 {
     unsigned dev = atoi(arg);
     if (dev >= UART_NUMOF) {
-        printf("Error: Invalid UART_DEV device specified (%u).\n", dev);
+        printf("msg: Invalid UART_DEV device specified (%u).\n", dev);
```
haukepetersen: unrelated? And inconsistent with the rest of the file?

tests/periph_uart/main.c
```diff
@@ -57,16 +57,97 @@ static int parse_dev(char *arg)
...
         return -1;
     }
     else if (UART_DEV(dev) == STDIO_UART_DEV) {
-        printf("Error: The selected UART_DEV(%u) is used for the shell!\n", dev);
+        printf("msg: The selected UART_DEV(%u) used for the shell!\n", dev);
```
haukepetersen: same here?!

tests/periph_uart/main.c
```diff
@@ -57,16 +57,97 @@ static int parse_dev(char *arg)
...
     return dev;
 }
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+static uart_data_bits_t parse_data_bits(char *arg)
+{
+    int data_bits = atoi(arg);
+    uart_data_bits_t res;
+
+    switch (data_bits) {
```
haukepetersen: here and in the following: how about a lookup tables?!

tests/periph_uart/main.c
```diff
@@ -57,16 +57,97 @@ static int parse_dev(char *arg)
...
+
+    return res;
+}
+
+static uart_parity_t parse_parity(char *arg)
+{
+    uart_parity_t res;
+
+    switch (arg[0]) {
+        case 'n':
```
haukepetersen: ne need for the 'double cases', simply substract the ASCII offeset before hand:
```c
if (arg[0] > 'Z') {
  arg[0] -= 32;
}

switch (arg[0]) {
   case 'N': ...
```

tests/periph_uart/main.c
```diff
@@ -154,6 +235,32 @@ static int cmd_init(int argc, char **argv)
...
+    if (argc < 5) {
+        printf("usage: %s <dev> <data bits> <parity> <stop bits>\n", argv[0]);
+        return 1;
+    }
+
+    dev = parse_dev(argv[1]);
+    if (dev < 0) {
+        return 1;
+    }
+    res = uart_mode(dev, parse_data_bits(argv[2]), parse_parity(argv[3]),
```
haukepetersen: this does potentiall break pretty nicely :-) 

-> **must** use `UART_DEV(dev)` as first parameter

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -370,6 +370,53 @@ typedef enum {
     STM32_LPUART,           /**< STM32 Low-power UART (LPUART) module type */
 } uart_type_t;
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+/* Invalid mask forces enums to be 32 bits */
+#define UART_INVALID_MODE 0x8000000
```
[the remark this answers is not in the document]
MrKevinWeiss: > why would you want to force the enum to be 32 bits?

This is because data_bits_t is used as an argument, if a certain type cpu is used (that doesn't contain the M1 bit) like the stm32f103rb then the size is 8 bits (more likely compiler dependent).  Since the data_bits variable is being used and altered when the mode is set depending on parity it must be a uint32_t type for the `data_bits = USART_CR1_M;` call.
The other way that was implemented was having a holder but I think this is a bit cleaner

Documentation should be updated though!

cpu/stm32_common/periph/uart.c
MrKevinWeiss: My bad, I suggested that.  I also tested that, I am not sure if before can also me at the same time assuming the bits were cleared initially.

MrKevinWeiss: It can be done at the same time.
The reason for this is the [stm32l0xxx](https://www.st.com/content/ccc/resource/technical/document/reference_manual/2f/b9/c6/34/28/29/42/d2/DM00095744.pdf/files/DM00095744.pdf/jcr:content/translations/en.DM00095744.pdf) page 802 states

> This bit can only be written when the USART is disabled (UE=0).

cpu/stm32_common/periph/uart.c
MrKevinWeiss: Most don't but some do L0 for example, you could ifdef I suppose.

cpu/stm32_common/periph/uart.c
MrKevinWeiss: That can be done as well, I mean it isn't doing that much then you don't need to increase the size of the struct.

tests/periph_uart/main.c
MrKevinWeiss: It is just so it can adhere to the riot_pal interface and RDM.

tests/periph_uart/main.c
haukepetersen: even better :-)

cpu/stm32_common/include/periph_cpu_common.h
```diff
@@ -370,6 +370,53 @@ typedef enum {
     STM32_LPUART,           /**< STM32 Low-power UART (LPUART) module type */
 } uart_type_t;
 
+#ifdef MODULE_PERIPH_UART_MODECFG
```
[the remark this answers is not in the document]
yegorich: I'll remove the `#ifdef`.

drivers/include/periph/uart.h
yegorich: ACK

drivers/include/periph/uart.h
yegorich: ACK

drivers/include/periph/uart.h
yegorich: Fixed.

drivers/include/periph/uart.h
yegorich: Fixed.

tests/periph_uart/main.c
yegorich: Done

tests/periph_uart/main.c
yegorich: Done

tests/periph_uart/main.c
yegorich: Reworked to check return values of parse_data_bits&Co before calling uart_mode

cpu/stm32_common/periph/uart.c
```diff
@@ -154,6 +154,41 @@ int uart_init(uart_t uart, uint32_t baudrate, uart_rx_cb_t rx_cb, void *arg)
...
+
+    if (parity) {
+        if (data_bits == UART_DATA_BITS_8) {
+            data_bits = USART_CR1_M;
+        }
+        else if (data_bits == UART_DATA_BITS_7) {
+            data_bits = UART_DATA_BITS_8;
+        }
+        else if (data_bits == UART_DATA_BITS_6) {
+            data_bits = UART_DATA_BITS_7;
```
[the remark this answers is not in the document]
yegorich: Done

cpu/stm32_common/periph/uart.c
yegorich: Reworked `uart_mode` to first clear data and stop bits as also parity settings.

cpu/stm32_common/include/periph_cpu_common.h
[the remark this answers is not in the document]
yegorich: Done

tests/periph_uart/Makefile
yegorich: Fixed

cpu/stm32_common/periph/uart.c
```diff
@@ -292,6 +329,19 @@ void uart_poweroff(uart_t uart)
 #endif
 }
 
+static uint8_t _data_mask(uart_t uart)
```
haukepetersen: I am still strongly in favor of spending one byte RAM per UART device for pre-calculating this mask instead of having these ~10 instructions overhead in each ISR call!

tests/periph_uart/main.c
```diff
@@ -67,6 +76,59 @@ static int parse_dev(char *arg)
     return dev;
 }
 
+#ifdef MODULE_PERIPH_UART_MODECFG
+static uart_data_bits_t parse_data_bits(char *arg)
+{
+    int data_bits = atoi(arg) - 5;
+
+    if (data_bits >= 0 && data_bits < data_bits_lut_len)
+        return data_bits_lut[data_bits];
```
haukepetersen: parenthesis missing

tests/periph_uart/main.c
```diff
@@ -67,6 +76,59 @@ static int parse_dev(char *arg)
...
+    printf("Error: Invalid number of data_bits (%i).\n", data_bits + 5);
+    return -1;
+}
+
+static uart_stop_bits_t parse_stop_bits(char *arg)
+{
+    int stop_bits = atoi(arg) - 1;
+
+    if (stop_bits >= 0 && stop_bits < stop_bits_lut_len)
+        return stop_bits_lut[stop_bits];
```
haukepetersen: parents missing here as well

tests/periph_uart/main.c
```diff
@@ -154,6 +216,45 @@ static int cmd_init(int argc, char **argv)
...
+    if (argc < 5) {
+        printf("usage: %s <dev> <data bits> <parity> <stop bits>\n", argv[0]);
+        return 1;
+    }
+
+    dev = parse_dev(argv[1]);
+    if (dev < 0) {
+        return 1;
+    }
+    data_bits = parse_data_bits(argv[2]);
```
haukepetersen: this does potentially break: there is no guarantee what the actual value returned by this function might look like (it might be defined freely on a per-CPU basis). So the check in the line below might not work reliable for any platform...

I suggest simply to remove the static function and put the data bit parsing code here directly!

tests/periph_uart/main.c
```diff
@@ -154,6 +216,45 @@ static int cmd_init(int argc, char **argv)
...
+    if (data_bits < 0) {
+        return 1;
+    }
+    parity = parse_parity(argv[3]);
```
haukepetersen: same here

tests/periph_uart/main.c
yegorich: I have moved all parsing actions to `cmd_mode`.

tests/periph_uart/main.c
yegorich: Fixed.

tests/periph_uart/main.c
yegorich: Fixed.

cpu/stm32_common/periph/uart.c
MrKevinWeiss: Yup fair enough.

drivers/include/periph/uart.h
```diff
@@ -97,6 +97,7 @@ typedef void(*uart_rx_cb_t)(void *arg, uint8_t data);
 typedef struct {
     uart_rx_cb_t rx_cb;     /**< data received interrupt callback */
     void *arg;              /**< argument to both callback routines */
+    uint8_t data_mask;      /**< mask applied to the data register */
```
haukepetersen: Nooo, this is bad. Do not change the **global** interface to integrate some STM specific value!

Instead, simply define a custom type inside the `stm32_common/periph/uart.c`, e.g.
```c
typedef struct {
    uart_isr_ctx_t base;
    uint8_t data_mask;
} isrctx_t; // name does not matter, as it is only local to this file...

static isrctx_t _isr_ctx;
```
or maybe even better, define an inline struct in the uart.c file
```c
static struct {
    uart_rx_cb_t rx_cb;
    void *arg;
    uint8_t data_mask;
} _isr_ctx;
```

yegorich: You're right. I've created a local structure with `data_mask` field.

drivers/include/periph/uart.h
```diff
@@ -137,6 +172,21 @@ enum {
...
+/**
+ * @brief   Setup parity, data and stop bits for a given UART device
+ *
+ * @param[in] uart          UART device to initialize
+ * @param[in] data_bits     number of data bits in a UART frame
+ * @param[in] parity        parity mode
+ * @param[in] stop_bits     number of stop bits in a UART frame
+ *
+ * @return                  UART_OK on success
+ * @return                  UART_NODEV on invalid UART device
```
haukepetersen: just wanted to ACK, but noticed one more thing (sorry...): the interface description does not match with the implementation above: the `uart_mode()` function should not return the `UART_NODEV` error code, but should use `assert()` internally to check the given device (as done in the implementation above). So please remove the line above.

haukepetersen:
@MrKevinWeiss could you give this a final test-run? Thx. If you succeed, feel free to merge!

MrKevinWeiss:
Ok hold for retesting

MrKevinWeiss:
Seems good to me.  GO!
Nice job @yegorich!

PR #10744, opened 2019-01-10
make: handle disabled default modules before dependencies

kaspar030:
### Contribution description

Currently, default modules can be disabled by adding the name to ```DISABLE_MODULE```.
Unfortunately, this gets parsed *after* resolving the dependencies.

This makes it impossible to disable e.g., stdio_uart and implicitly periph/uart. The dependency parsing would select periph/uart (as it is dependend on stdio_uart), but later stdio_uart would be dropped, leaking the dependency.

This PR moves ```DISABLE_MODULE``` before dependency resolution.

### Testing procedure

Correct compilation of all applications should suffice.

### Issues/PRs references

#10741

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -286,10 +286,11 @@ ifneq ($(GNRC_NETIF_NUMOF),1)
   CFLAGS += -DGNRC_NETIF_NUMOF=$(GNRC_NETIF_NUMOF)
 endif
 
+# handle removal of default modules
+USEMODULE += $(filter-out $(DISABLE_MODULE), $(DEFAULT_MODULE))
+
 # process dependencies
 include $(RIOTBASE)/Makefile.dep
-
-USEMODULE += $(filter-out $(DISABLE_MODULE), $(DEFAULT_MODULE))
 
 ifeq ($(strip $(MCU)),)
 	MCU = $(CPU)


basilfx:
Maybe I'm using it wrong, but it isn't working for me.

I have a board that includes `stdio_rtt` in its `Makefile.dep`. Now I want to compile it with `stdio_rtt` disabled (and default to `stdio_uart` since it's the default).

I run `DISABLE_MODULE=stdio_rtt BOARD=basilfx-taster make -j8` and I get this error:
```
DISABLE_MODULE="stdio_uart" BOARD=basilfx-taster make -j8
Required modules were disabled using DISABLE_MODULE: stdio_uart

EXPECT ERRORS!

examples/gnrc_knx_taster/../../Makefile.include:737: examples/gnrc_knx_taster/bin/basilfx-taster/stdio_uart.a
examples/gnrc_knx_taster/../../Makefile.include:739: *** BASELIBS value changed.  Stop.
```

kaspar030:
> Maybe I'm using/understanding it wrong, but it isn't working for me.

Unfortunately, despite it's naming, DISABLE_MODULE only works for DEFAULT_MODULES.

This PR is the first, there are more to come that allow disabling stdio_uart...

cladmi:
The only definitions of `DEFAULT_MODULE` were in `makefiles/defaultmodules.inc.mk` and one in `cpu/msp430_common/Makefile.include` for some reason, so has no bad impact now when looking at which modules are inside.

However, I do not really understood the issue that it will solve. The values in `DEFAULT_MODULE` are rather generic mechanisms that are included in a normal firmware, and modules you are supposed to find there. 

I also mix a bit the order with depends on is dependent on so cannot really see the goal right now.

If A depends on B, you disable B, but still request A. I would maybe expect it to fail, not silently add B because it is a dependency. I do not think that you should able to say add and remove something at the same time.
Even more right now when it is for generic things like there is in the `DEFAULT_MODULE` list.

But we may indeed be missing a mechanism for your need.

PR #10746, opened 2019-01-10
cpu/lpc2387: remove newlib dependencies

kaspar030:
### Contribution description

Currently, the boot code on lpc2387 hard-codes the use of ```printf()``` and uses ```exit()```, both inducing a dependency to libc.

This PR replaces ```printf()``` with ```LOG_ERROR()``` and ```exit(1)``` with an endless loop.

### Testing procedure

Run basic tests on lpc2387.


kaspar030:
Thx @maribu! I was a little bit afraid that noone would look at this...

PR #10754, opened 2019-01-11
nanocoap: fix server-side option_count overflow

### LINKED ISSUE #10753: nanocoap: options buffer overflow
#### Description

nanocoap contains a buffer overflow which has been introduced with
commit dee793d29f09bafa83bb2c9fe293eb7b16fc009a. The bug allows an
attacker to overflow the `options` buffer in the `coap_pkt_t` supplied
to `coap_parse`.

The relevant code part is the following:

```C
if (option_delta) {
        optpos->opt_num = option_nr;
        optpos->offset = (uintptr_t)option_start - (uintptr_t)hdr;
        DEBUG("optpos option_nr=%u %u\n", (unsigned)option_nr, (unsigned)optpos->offset);
        optpos++;
        option_count++;
}
```

`optpos` is a pointer to the `options` buffer in the `coap_pkt_t`. This
pointer is incremented without checking if it exceeds
`NANOCOAP_NOPTS_MAX` (the size of the buffer). It also used to write a
new option at the current position in the `options` buffer.

To trigger this buffer overflow an attacker can send a CoAP request
containing more than `NANOCOAP_NOPTS_MAX` options to a nanocoap server.
This may allow an attacker to crash the nanocoap server and cause a
denial of service. Additionally, a clever attacker might also be able to
use this for a remote code execution.

A malicious CoAP packet triggering this vulnerability can be send to a RIOT
node in order to crash it. I created a malicious packet containing 42 which
should successfully crash native.

To test this build `examples/nanocoap_server` and send the crafted
malicious packet to it. For example:

```
$ echo QAEJJgEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFxEXERcRFw== > crash
$ while sleep 1; do busybox nc -u '[ip-address]' 5683 -e base64 -d crash; done
```

---

I personally prefer full disclosure for reporting such issues. Especially since #10749 hasn't been resolved yet. I also believe that doing full disclosure including documentation of an exploit worked quite well to get #10739 fixed in time. Besides, I didn't want to sit on this any longer and full disclosure is way better than no disclosure at all.

kaspar030:
### Contribution description

nanocoap limits the number of options it parses, but didn't check if that limit was reached when parsing a packet. (See report in #10753).

This PR provides a fix. It also adds a unittest that exposes the flaw.

Fixes #10753.

### Testing procedure

Run unittests without uppermost commit. The nanocoap test should crash on native. Run again with topmost commit. Tests should succeed.

### Issues/PRs references

See #10753.

### DIFF
--- a/sys/net/application_layer/nanocoap/nanocoap.c
+++ b/sys/net/application_layer/nanocoap/nanocoap.c
@@ -111,6 +111,9 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
                 DEBUG("optpos option_nr=%u %u\n", (unsigned)option_nr, (unsigned)optpos->offset);
                 optpos++;
                 option_count++;
+                if (option_count >= NANOCOAP_NOPTS_MAX) {
+                    return -ENOMEM;
+                }
             }
 
             pkt_pos += option_len;
--- a/tests/unittests/tests-nanocoap/tests-nanocoap.c
+++ b/tests/unittests/tests-nanocoap/tests-nanocoap.c
@@ -15,6 +15,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <string.h>
+#include <stdio.h>
 
 #include "embUnit.h"
 
@@ -433,6 +434,31 @@ static void test_nanocoap__server_reply_simple_con(void)
     TEST_ASSERT_EQUAL_INT(COAP_TYPE_ACK, coap_get_type(&pkt));
 }
 
+static void test_nanocoap__server_option_count_overflow_check(void)
+{
+    static uint8_t pkt_data[] = {
+        0x40, 0x01, 0x09, 0x26, 0x01, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17,
+        0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17,
+        0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17,
+        0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17,
+        0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17,
+        0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17,
+        0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17, 0x11, 0x17,
+        0x11, 0x17, 0x11, 0x17 };
+
+    coap_pkt_t pkt;
+    volatile uint32_t guard_value = 0x12345678;
+
+    /* print the pointer values to ensure the data is kept on the stack (not in
+     * registers) */
+    printf("%p %p\n", (void*)&pkt, (void*)&guard_value);
+
+    int res = coap_parse(&pkt, pkt_data, sizeof(pkt_data));
+
+    TEST_ASSERT_EQUAL_INT(-ENOMEM, res);
+    TEST_ASSERT_EQUAL_INT(0x12345678, guard_value);
+}
+
 Test *tests_nanocoap_tests(void)
 {
     EMB_UNIT_TESTFIXTURES(fixtures) {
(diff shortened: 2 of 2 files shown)


kb2ma:
Just took a first look. The fix looks appropriate. Need to do a little more testing.

In the unit test, how do I know that the printed pointer values indicate a good test? Just that they are close to each other? Is it possible to be more specific? It's odd to print values as part of a unit test. I would think it makes more sense to compare the values within the test, and fail the test if the difference is outside of some range.

kb2ma:
Sorry, I have no insights. In keeping with the current approach, you could put a guard value on both sides to be safe.

Another approach is to set the maximum number of options (NANOCOAP_NOPTS_MAX) to a small value and try a functional test. That's what I'm working on.

kb2ma:
You also could craft an array with NANOCOAP_NOPTS_MAX options, parse it, and expect success. Then add another option to it and expect -ENOMEM. That should be sufficient for this test. No need to get into the low level overflow detection.

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -111,6 +111,9 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
                 DEBUG("optpos option_nr=%u %u\n", (unsigned)option_nr, (unsigned)optpos->offset);
                 optpos++;
                 option_count++;
+                if (option_count >= NANOCOAP_NOPTS_MAX) {
```
kb2ma: It should be OK to include NANOCOAP_NOPTS_MAX options. This test should be moved, above addition to this array, probably line 109. The test then should be:

`if (option_count == NANOCOAP_NOPTS_MAX) {`

kaspar030: Yes. The new test seems to be working, it exposed that using "==" here caused to write one behind the options array. So I moved the check to the beginning of the if case.

kaspar030:
> Then add another option to it and expect -ENOMEM. That should be sufficient for this test. No need to get into the low level overflow detection.

I think I found a way. At least on native it works as expected.

sys/net/application_layer/nanocoap/nanocoap.c
```diff
@@ -106,6 +106,11 @@ int coap_parse(coap_pkt_t *pkt, uint8_t *buf, size_t len)
             DEBUG("option count=%u nr=%u len=%i\n", option_count, option_nr, option_len);
 
             if (option_delta) {
+                if (option_count >= NANOCOAP_NOPTS_MAX) {
+                    DEBUG("nanocoap: max nr of options exceeded");
```
kb2ma: I agree debug statement is helpful. Needs newline.

tests/unittests/tests-nanocoap/tests-nanocoap.c
```diff
@@ -433,6 +434,40 @@ static void test_nanocoap__server_reply_simple_con(void)
     TEST_ASSERT_EQUAL_INT(COAP_TYPE_ACK, coap_get_type(&pkt));
 }
 
+static void test_nanocoap__server_option_count_overflow_check(void)
```
kb2ma: I kind of understand what you are doing to test the overflow. Why are there 42 options in guard data? How does the unit test work? I think it's important for others to understand this without a lot of thought/investigation.

sys/net/application_layer/nanocoap/nanocoap.c
kaspar030: > Needs newline.

fixed

tests/unittests/tests-nanocoap/tests-nanocoap.c
kaspar030: I've added some comments.

kb2ma:
I agree it's OK to include both tests since they verify from different directions. These tests run OK for me on native and samr21-xpro. I verified that allowing coap_parse() to write an extra option makes the tests fail.

Thanks for adding the comments in the test. The only remaining question for me is why are there 42 options in guard_data instead of only one or two. When I allowed the parse to overwrite the guard data, I found the overwrite occurred in the first few bytes.

kaspar030:
> The only remaining question for me is why are there 42 options in guard_data instead of only one or two. When I allowed the parse to overwrite the guard data, I found the overwrite occurred in the first few bytes.

@nmeum provided this packet in his original report, and I was lazy enough to take it as-is. The number just needs to be higher than ```NANOCOAP_NOPT_MAX```. (42 seemed to be the right answer after thinking for 7.5 million years.) I've added a note stating that 42 is basically randomly chosen.

kb2ma:
OK, good enough. ACK, please squash.

jia200x:
@kaspar030 could you provide a backport to the 2018.10 branch?

PR #10755, opened 2019-01-11
drivers/ltc4150: (Re-)implemented driver for the LTC4150 coulomb counter

maribu:
### Contribution description

This PR implements a driver for the LTC4150 coulomb counter. A driver for that device was previously part for RIOT, but was dropped because of quality shortcomings. This driver is written from scratch.

### Testing procedure

The LTC4150 is integrated in the MSB-A2 board, thus using that board will be most convenient to use for testing. However, this chip is also cheaply available as breakout board from various vendors, which all seem to be based on the same open hardware PCB.

Testing:

1. Build, flash & run `tests/driver_ltc4150` on the MSB-A2. I intended the program to be self-explaining - so I will not provide more information here to see if I indeed reached that goal
2. Build, flash & run `example/default`. Try to read the data using the `saul` shell command

### Issues/PRs references

This PR competes with PR https://github.com/RIOT-OS/RIOT/pull/9653 also opened by me. I personally favor this PR strongly over the old approach.

### DIFF
--- a/drivers/include/ltc4150.h
+++ b/drivers/include/ltc4150.h
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2019 Otto-von-Guericke-Universität Magdeburg
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @defgroup    drivers_ltc4150 LTC4150 coulomb counter
+ * @ingroup     drivers_sensors
+ * @brief       Driver for the Linear Tech LTC4150 Coulomb Counter
+ *              (a.k.a. battery gauge sensor or power consumption sensor)
+ *
+ * # Wiring the LTC4150
+ * Hint: M Grusin thankfully created an
+ * [open hardware breakout board](https://cdn.sparkfun.com/datasheets/BreakoutBoards/LTC4150_BOB_v10.pdf).
+ * As a result, virtually all LTC4150 breakout boards are using this schematic.
+ * Whenever this documentation refers to a breakout board, this open hardware
+ * board is meant. Of course, this driver works with the "bare" LTC4150 as well.
+ *
+ * Please note that this driver works interrupt driven and does not clear the
+ * signal. Thus, the /CLR and /INT pins on the LTC4150 need to be connected
+ * (in case of the breakout board: close solder jumper SJ1), so that the signal
+ * is automatically cleared.
+ *
+ * Hint: The breakout board uses external pull up resistors on /INT, POL and
+ * /SHDN. Therefore /SHDN can be left unconnected and no internal pull ups are
+ * required for /INT and POL. In case your board uses 3.3V logic the solder
+ * jumpers SJ2 and SJ3 have to be closed, in case of 5V they have to remain
+ * open. Connect the VIO pin to the logic level, GND to ground, IN+ and IN- to
+ * the power supply and use OUT+ and OUT- to power your board.
+ *
+ * In the easiest case only the /INT pin needs to be connected to a GPIO,
+ * and (in case of external pull ups) /SHDN and POL can be left unconnected.
+ * The GPIO /INT is connected to support for interrupts, /SHDN and POL
+ * (if connected) do not require interrupt support.
+ *
+ * In case a battery is used the POL pin connected to another GPIO. This allows
+ * to distinguish between charge drawn from the battery and charge transferred
+ * into the battery (used to load it).
+ *
+ * In case support to power off the LTC4150 is desired, the /SHDN pin needs to
+ * be connected to a third GPIO.
+ *
+ * # Things to keep in mind
+ * The LTC4150 creates pulses with a frequency depending on the current drawn.
+ * Thus, more interrupts need to be handled when more current is drawn, which
+ * in turn increases system load (and power consumption). The interrupt service
+ * routing is quite short and even when used outside of specification less than
+ * 20 ticks per second will occur. Hence, this effect should hopefully be
+ * negligible.
+ *
+ * @{
+ *
+ * @file
+ * @brief       LTC4150 coulomb counter
+ *
+ * @author      Marian Buschsieweke <marian.buschsieweke@ovgu.de>
+ */
+
+#ifndef LTC4150_H
+#define LTC4150_H
+
+#include <stdint.h>
+
+#include "mutex.h"
+#include "periph/gpio.h"
+#include "xtimer.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Configuration flags of the LTC4150 coulomb counter
+ */
+enum {
+    /**
+     * @brief External pull on the /INT pin is present
+     */
+    LTC4150_INT_EXT_PULL_UP = 0x01,
+    /**
+     * @brief External pull on the /POL pin is present
+     */
+    LTC4150_POL_EXT_PULL_UP = 0x02,
+    /**
+     * @brief External pull on the /INT *and* the /POL pin is present
+     */
+    LTC4150_EXT_PULL_UP = LTC4150_INT_EXT_PULL_UP | LTC4150_POL_EXT_PULL_UP,
+};
+
+/**
+ * @brief Enumeration of directions in which the charge can be transferred
+ */
+typedef enum {
+    LTC4150_CHARGE,             /**< The battery is charged */
+    LTC4150_DISCHARGE,          /**< Charge is drawn from the battery */
+} ltc4150_dir_t;
+
+/**
+ * @brief LTC4150 coulomb counter
+ */
+typedef struct ltc4150_dev ltc4150_dev_t;
+
+/**
+ * @brief Interface to allow recording of the drawn current in a user defined
+ *        resolution
+ *
+ * @note  Keep in mind that the data recording may be performed by the CPU of
+ *        the system to monitor - thus keep power consumption for the recording
+ *        low!
+ *
+ * The LTC4150 driver will only track total charge transferred (separately for
+ * charging in discharging direction). However, there are use cases that
+ * required more precise data recording, e.g. a rolling average of the last
+ * minute. This interface allows application developers to implement the ideal
+ * trade-off between RAM, ROM and runtime overhead for the data recording and
+ * the level of information they require.
+ */
+typedef struct {
+    /**
+     * @brief   Function to call on every pulse received from the LTC4150
+     * @warning This function is called in interrupt context
+     *
+     * @param[in] dev       The device the pulse was received from
+     * @param[in] dir       Direction in which the charge is transferred
+     * @param[in] now_usec  The system time the pulse was received in µs
+     * @param[in] arg       (Optional) argument for this callback
+     */
+    void (*pulse)(ltc4150_dev_t *dev, ltc4150_dir_t dir, uint64_t now_usec, void *arg);
+    /**
+     * @brief   Function to call upon driver initialization or reset
+     *
+     * @see ltc4150_init
+     * @see ltc4150_reset_counters
+     *
+     * @param[in] dev       The LTC4150 device to monitor
+     * @param[in] now_usec  The system time the pulse was received in µs
+     * @param[in] arg       (Optional) argument for this callback
+     */
+    void (*reset)(ltc4150_dev_t *dev, uint64_t now_usec, void *arg);
+} ltc4150_recorder_t;
+
+/**
+ * @brief Parameters required to set up the LTC4150 coulomb counter
+ */
+typedef struct {
+    /**
+     * @brief Pin going LOW every time a specific charge is drawn, labeled INT
+     */
+    gpio_t interrupt;
+    /**
+     * @brief Pin indicating (dis-)charging, labeled POL
+     *
+     * Set this pin to `GPIO_UNDEF` to tread every pulse as discharging. This
+     * pin is pulled low by the LTC4150 in case the battery is discharging.
+     */
+    gpio_t polarity;
+    /**
+     * @brief Pin to power off the LTC4150 coulomb counter, labeled SHDN
+     *
+     * Set this pin to `GPIO_UNDEF` if the SHDN pin is not connected to the MCU
+     */
+    gpio_t shutdown;
+    /**
+     * @brief Pulse per ampere hour of charge
+     *
+     * pulses = 3600 * 32.55 * R
+     *
+     * Where R is the resistance (in Ohm) between the SENSE+ and SENSE- pins.
+     * E.g. the MSBA2 has 0.390 Ohm (==> 45700 pulses), while most breakout
+     * boards for the LTC4150 have 0.050 Ohm (==> 5859 pulses).
+     */
+    uint16_t pulses_per_ah;
+    /**
+     * @brief Configuration flags controlling if inter pull ups are required
+     *
+     * Most [breakout boards](https://cdn.sparkfun.com/datasheets/BreakoutBoards/LTC4150_BOB_v10.pdf)
+     * and the MSBA2 board use external pull up resistors, so no internal pull
+     * ups are required. Clear the flags to use internal pull ups instead.
+     */
+    uint16_t flags;
+    /**
+     * @brief `NULL` or a `NULL`-terminated array of data recorders
+     * @pre   If not `NULL`, the last element of the array must be `NULL`
+     */
+    ltc4150_recorder_t **recorders;
+    /**
+     * @brief `NULL` or an array of the user defined data for each recorder
+     * @pre   If @see ltc4150_params_t::recorders is not `NULL`, this must point
+     *        to an array of `void`-Pointers of the same length.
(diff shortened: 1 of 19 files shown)


maribu:
Before you start to review: Let me fix the missing doc Travis complained about and squash

maribu:
I'm aware that commit "tests/driver_ltc4150: Workarround for msp430" could be squashed into commit "tests: Added test for ltc4150 driver". But I would actually prefer to keep them separated: That commit did not fix a bug in the test, but was a workarround for a bug in the msp430 toolchain (the missing fputs). Once the c lib of the msp430 toolchain gains support for fputs, it can be safely reverted to make the test source code shorter and more readable. For that reason I would like to keep that commit separate.

maribu:
@MichelRottleuthner: Could you kindly do the testing? I think @smlng does not have a MSB-A2 board at hand (but I'm not sure).

miri64:
I believe @MichelRottleuthner doesn't have one either, but I have 8 ;-). Once the fundamentals and and code design are cleared (e.g. by @MichelRottleuthner or @smlng) I'm happy to test this PR.

drivers/include/ltc4150.h
```diff
@@ -0,0 +1,347 @@
...
+    LTC4150_EXT_PULL_UP = LTC4150_INT_EXT_PULL_UP | LTC4150_POL_EXT_PULL_UP,
+};
+
+/**
+ * @brief Enumeration of directions in which the charge can be transferred
+ */
+typedef enum {
+    LTC4150_CHARGE,             /**< The battery is charged */
+    LTC4150_DISCHARGE,          /**< Charge is drawn from the battery */
+} ltc4150_dir_t;
```
smlng: so its either `charge` or `discharge`, never both, right? So why not use a simple `bool charge` with `true` and `false` in the implementation. To me that would be more readable, e.g. by omitting `switch () case`.

smlng: plus it saves 16 Bytes, when using `bool` instead of the `enum` ...

but I leave it to you, that's a non-blocking suggestion

smlng: @MichelRottleuthner what do you think?

drivers/ltc4150/include/ltc4150_params.h
```diff
@@ -0,0 +1,96 @@
...
+#define LTC4150_PARAM_INT             (GPIO_PIN(0, 4))
+#endif
+#ifndef LTC4150_PARAM_POL
+#define LTC4150_PARAM_POL             (GPIO_UNDEF)
+#endif
+#ifndef LTC4150_PARAM_SHUTDOWN
+#define LTC4150_PARAM_SHUTDOWN        (GPIO_PIN(0, 5))
+#endif
+#ifndef LTC4150_PARAM_PULSES
+#define LTC4150_PARAM_PULSES          (45700)
```
smlng: didn't we agree on adding (at least) a `U` suffix in the previous PR?

drivers/ltc4150/ltc4150.c
```diff
@@ -0,0 +1,201 @@
...
+#include "debug.h"
+
+static void pulse_cb(void *_dev)
+{
+    uint64_t now;
+    ltc4150_dir_t dir;
+    ltc4150_dev_t *dev = _dev;
+
+    if (
+        (dev->params.polarity == GPIO_UNDEF) ||
```
smlng: personally I would not put this on a new line, as the indention stays the same, hence I think its fine to write this as:

```
if ((dev->params.polarity == GPIO_UNDEF) ||
    (!gpio_read(dev->params.polarity))) {
```

tests/saul/Makefile
```diff
@@ -5,4 +5,7 @@ USEMODULE += saul_default
 
 USEMODULE += xtimer
 
+# Too little flash:
+BOARD_INSUFFICIENT_MEMORY := arduino-duemilanove arduino-uno
```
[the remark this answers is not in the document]
maribu: The flash increase comes from these two commits: https://github.com/RIOT-OS/RIOT/pull/10755/commits/45511d4a8bf250decdb99e6ba1aeb7a0b6c36968 https://github.com/RIOT-OS/RIOT/pull/10755/commits/30b350b19cc2f19bf875a4a0e7f9bba331fff17b

These boards will not have `USEMODULE += ltc4150` in `tests/saul` enabled, so this is not the drivers fault. It's really only the strings of the units and sensors type that increase the ROM.

drivers/include/ltc4150.h
miri64: Sidenote from afar: I like descriptive value names better.

drivers/ltc4150/ltc4150_last_minute.c
```diff
@@ -0,0 +1,94 @@
...
+    .reset = init_or_reset,
+    .pulse = pulse,
+};
+
+static void init_or_reset(ltc4150_dev_t *dev, uint64_t now_usec, void *arg)
+{
+    (void)dev;
+    ltc4150_last_minute_data_t *data = arg;
+
+    memset(data, 0, sizeof(ltc4150_last_minute_data_t));
```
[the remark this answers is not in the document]
smlng: agreed

drivers/ltc4150/ltc4150.c
```diff
@@ -0,0 +1,201 @@
...
+    dev->last_update_sec = now / US_PER_SEC;
+}
+
+int ltc4150_init(ltc4150_dev_t *dev, const ltc4150_params_t *params)
+{
+    if (!dev || !params) {
+        return -EINVAL;
+    }
+
+    memset(dev, 0x00, sizeof(ltc4150_dev_t));
```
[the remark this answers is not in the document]
smlng: yep, true that ... leave as is.

However, please be consistent and style/usage, i.e. here you write `memset(..., 0x00 ...)` and below `memset(..., 0, ...)` ... I think the latter is nicer.

tests/driver_ltc4150/main.c
```diff
@@ -0,0 +1,320 @@
...
+/**
+ * @brief Thread that will put three levels of CPU load on the MCU
+ */
+static void *busy_thread(void *arg)
+{
+    (void)arg;
+    while (1) {
+        /* one minute of ~0% CPU usage */
+        LED0_OFF;
+        LED1_OFF;
```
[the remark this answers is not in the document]
smlng: yeah I know we have the dummy macros, I just wanted to point out that on some boards you won't see a second LED (or GPIO) changing.

tests/saul/Makefile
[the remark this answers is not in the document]
smlng: the above was compiled and run with your changes on an Arduino-Uno, here build output:

```
BOARD=arduino-uno make -C tests/saul/
Building application "tests_saul" for "arduino-uno" with MCU "atmega328p".

"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/boards/arduino-uno
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/boards/common/arduino-atmega
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/boards/common/atmega
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/core
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/atmega328p
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/atmega_common
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/cpu/atmega_common/periph
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/drivers
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/drivers/periph_common
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/drivers/saul
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/auto_init
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/auto_init/saul
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/div
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/fmt
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/isrpipe
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/phydat
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/saul_reg
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/stdio_uart
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/tsrb
"/Library/Developer/CommandLineTools/usr/bin/make" -C /Volumes/devel/github/smlng/RIOT/sys/xtimer
  10382	    906	    981	  12269	   2fed	/Volumes/devel/github/smlng/RIOT/tests/saul/bin/arduino-uno/tests_saul.elf

```

though is done on macOS with following versions:

```

Operating System Environment
-----------------------------
       Operating System: Mac OS X 10.14.2
                 Kernel: Darwin 18.2.0 x86_64 i386

Installed compiler toolchains
-----------------------------
             native gcc: gcc (Homebrew GCC 8.2.0) 8.2.0
                avr-gcc: avr-gcc (GCC) 8.2.0
       mips-mti-elf-gcc: missing
             msp430-gcc: missing
   riscv-none-embed-gcc: missing
   xtensa-esp32-elf-gcc: missing
   xtensa-lx106-elf-gcc: missing
                  clang: Apple LLVM version 10.0.0 (clang-1000.10.44.4)

Installed compiler libs
-----------------------
   arm-none-eabi-newlib: "2.5.0"
    mips-mti-elf-newlib: missing
xtensa-esp32-elf-newlib: missing
xtensa-lx106-elf-newlib: missing

Installed development tools
---------------------------
                  cmake: cmake version 3.13.2
               cppcheck: Cppcheck 1.86
                doxygen: 1.8.15
                 flake8: 3.6.0 (mccabe: 0.6.1, pycodestyle: 2.4.0, pyflakes: 2.0.0) CPython 3.7.2 on Darwin
                    git: git version 2.20.1
                   make: GNU Make 3.81
                 python: Python 2.7.15
                python2: Python 2.7.15
                python3: Python 3.7.2
             coccinelle: spatch version 1.0.6 compiled with OCaml version 4.05.0
```

Could be that it does not fit using some other compiler version on Linux. However, if it is necessary to pass Murdock leave it in - I don't like these BOARDS_XXX list anyway and such should be handled differently by the build system ... but that's another story (or PR).

drivers/include/ltc4150.h
MichelRottleuthner: I also prefer the named values

tests/saul/Makefile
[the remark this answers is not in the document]
MichelRottleuthner: I can confirm that the build is too big for arduino-uno and arduino-duemilanove on my local setup, and alow with BUILD_IN_DOCKER=1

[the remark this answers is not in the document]
maribu: I added this because of a Murdock build fail in the previous PR. As the strings of the units and the sensors have not been changed, it should still be necessary.

> I don't like these BOARDS_XXX list anyway and such should be handled differently by the build system 

+1 on that. E.g. if unrelated code gets improved for ROM size previously blacklisted boards will likely remain blacklisted, even if that would no longer be possible. (But I personally do not have an idea how a better solution would look like...)

tests/driver_ltc4150/main.c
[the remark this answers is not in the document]
maribu: Ahh! Your point is that the description above claiming LEDs will blink is misleading, if no LEDs are present. That is indeed true, I should fix that.

drivers/ltc4150/ltc4150.c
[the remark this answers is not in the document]
maribu: > However, please be consistent and style/usage, i.e. here you write memset(..., 0x00 ...) and below memset(..., 0, ...) ... I think the latter is nicer.

+1

drivers/ltc4150/include/ltc4150_params.h
maribu: Indeed, sorry

drivers/include/ltc4150.h
maribu: > plus it saves 16 Bytes, when using `bool` instead of the `enum` ...

This is a good point. I personally consider enums more readable than `bool`, but if someone wants to use that driver and her/his ROM is overflown by only a couple of bytes, this user would clearly favor the `bool` over the `enum`.

Honestly, I am totally surprised that there is a difference in code size by using a two-value `enum` over a `bool`. I'll have to look into this to understand that.

@miri64, @MichelRottleuthner, @smlng:

Would 

``` C
typedef bool ltc4150_dir_t;

#define LTC4150_CHARGE      true
#define LTC4150_DISCHARGE   false
```

be a compromise? (Sure, with the `enum` gdb would be able to show `LTC4150_CHARGE` instead of `1`...)

smlng: I think the size difference comes from avoiding `switch ... case` when using bool

smlng: but maybe, leave as if for the first integration of that driver and refine later? At least I wouldn't block if you stick with the enum for now.

tests/saul/Makefile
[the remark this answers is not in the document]
smlng: oki, than leave the boards in there

miri64:
@maribu just a remark: if you use the commit summary `git commit --fixup=<commit>` provides you with, squashing becomes easier due to `git rebase -i <base> --autosquash` ;-).

drivers/include/ltc4150.h
maribu: > I think the size difference comes from avoiding `switch ... case` when using bool

Yes, obviously. I should have seen that. I now optimized it by making the switch binary (by making one of the two `case`s the `default`), which reduced the ROM size by `28` bytes on my machine. With that, `enum` and `bool` should now indeed be the same regarding ROM size.

maribu:
> Tests work as expected, though I was confused due to text speaking about drawn charges, while the "Charging" column was raising constantly.

Ohh! Let me check

maribu:
@miri64: Fixed :-) I put the values in the wrong columns

maribu:
@miri64: I added the line between load levels and tried to make the description clearer.

tests/driver_ltc4150/main.c
```diff
@@ -45,6 +45,7 @@ static const ltc4150_recorder_t test_recorder = {
 static kernel_pid_t target_pid;
 static char busy_thread_stack[THREAD_STACKSIZE_DEFAULT];
 static ltc4150_dev_t ltc4150;
+static int change_of_load_level;
```
miri64: I know BSS is initialized with 0, but this rings all the "possible uninitialized value" bells in my head with the access in the main. Could you please add a `= 0` to calm those bells ;-)?

maribu:
@miri64: Addressed :-)

maribu:
May I squash?

Do we all agree that this PR is better than the competing PR at https://github.com/RIOT-OS/RIOT/pull/9653? If so, I would close https://github.com/RIOT-OS/RIOT/pull/9653

MichelRottleuthner:
I prefer this PR over the old one. Regarding the squash @miri64 and @smlng should decide because they requested the changes. From my side it's good to go!

miri64:
I hadn't looked at the code as much, just tested and my comments were addressed. So I'm fine with squashing as well.

miri64:
(you should wait for @smlng's go though)

maribu:
@smlng: May I squash?

smlng:
sure, please squash

maribu:
Hmm, the "force-pushed" link is not clever enough to filter out changes due to rebasing against current master. I still have a back up here https://github.com/maribu/RIOT/tree/ltc4150-new-prior-squash for reference, which is neither squashed nor rebased (in other words: Contains the state prior force push).

miri64:
> Hmm, the "force-pushed" link is not clever enough to filter out changes due to rebasing against current master.

That's why I usually just squash (i.e. make a rebase against the merge-base with master) if it is not absolutely necessary. It's not about cleverness. The link is just an HTML-ized version of `git diff <pre>..<post>`.

miri64:
Welp, that's for @github to resolve ;-)

miri64:
Re-run the tests. Still working.

maribu:
@miri64, @MichelRottleuthner, @smlng: Thank you very much for your valuable feedback :-)

PR #10763, opened 2019-01-14
tests/periph_rtc: test sub-second precision

benemorius:
### Contribution description
This expands `tests/periph_rtc` to verify the microsecond precision of RTC timers. This is needed to test that an RTC alarm fires at the desired time and not with an arbitrary offset up to +/-1 second. It also tests that `rtc_set_time()` behaves equivalently correctly, which is what I needed to test when I wrote this.

### Testing procedure
Run `tests/periph_rtc` and observe new output (`error -177429 microseconds`).

Currently it fails for `kinetis/kw41z` (perhaps all `kinetis`) and that looks like this:
```
main(): This is RIOT! (Version: 2018.10-RC1-679-ga296b-telnet-shell)

RIOT RTC low-level driver test
This test will display 'Alarm!' every 2 seconds for 4 times
  Setting clock to   2011-12-13 14:15:15
Clock value is now   2011-12-13 14:15:15
  Setting alarm to   2011-12-13 14:15:17
   Alarm is set to   2011-12-13 14:15:17

[30303] First alarm expected at 2018615 microseconds
[1841186] Alarm! after 1822571 microseconds (error -177429 microseconds)
[3841186] Alarm! after 3822571 microseconds (error -177429 microseconds)
[5841186] Alarm! after 5822571 microseconds (error -177429 microseconds)
[7841186] Alarm! after 7822571 microseconds (error -177429 microseconds)
```
In the case of `kw41z` the error is always a random value (abs(error) < 1 second) depending on what value is randomly in the RTC subsecond register which is currently ignored by `kinetis/periph/rtt_set_counter()`.

A successful test should have output similar to:
```
main(): This is RIOT! (Version: 2018.10-RC1-680-g83348-telnet-shell)

RIOT RTC low-level driver test
This test will display 'Alarm!' every 2 seconds for 4 times
  Setting clock to   2011-12-13 14:15:15
Clock value is now   2011-12-13 14:15:15
  Setting alarm to   2011-12-13 14:15:17
   Alarm is set to   2011-12-13 14:15:17

[30303] First alarm expected at 2018585 microseconds
[2018615] Alarm! after 2000030 microseconds (error 30 microseconds)
[4018615] Alarm! after 4000030 microseconds (error 30 microseconds)
[6018615] Alarm! after 6000030 microseconds (error 30 microseconds)
[8018615] Alarm! after 8000030 microseconds (error 30 microseconds)
```
The error precision is limited by the frequency of `xtimer`'s timer, hence the 30us error for `kw41z` with a 16bit 32kHz `xtimer` where 30us is only 1 clock tick. Platforms with a faster `xtimer` should have a smaller error. Perhaps no platform should have an error over 100us.

### Comments
Must test other platforms.
Perhaps I should have updated the readme.
Why doesn't newlib declare `settimeofday()`?

### Issues/PRs references
#10764 fixes `kinetis/periph_rtt` to give the successful test output shown above.

### DIFF
--- a/tests/periph_rtc/main.c
+++ b/tests/periph_rtc/main.c
@@ -34,6 +34,8 @@
 #define TM_YEAR_OFFSET      (1900)
 
 static unsigned cnt = 0;
+static uint32_t start_usecs = 0;
+static uint32_t alarm_expected_usecs = 0;
 
 static void print_time(const char *label, const struct tm *time)
 {
@@ -58,13 +60,18 @@ static void cb(void *arg)
 {
     (void)arg;
 
-    puts("Alarm!");
+    uint32_t now_usecs = xtimer_now_usec();
+    printf("[%" PRIu32 "] Alarm! after %" PRIu32 " microseconds (error %" PRId32 " microseconds)\n",
+           now_usecs,
+           now_usecs - start_usecs,
+           (int32_t)now_usecs - alarm_expected_usecs);
 
     if (++cnt < REPEAT) {
         struct tm time;
         rtc_get_alarm(&time);
         inc_secs(&time, PERIOD);
         rtc_set_alarm(&time, cb, NULL);
+        alarm_expected_usecs += US_PER_SEC * PERIOD;
     }
 }
 
@@ -86,20 +93,27 @@ int main(void)
     /* set RTC */
     print_time("  Setting clock to ", &time);
     rtc_set_time(&time);
+    start_usecs = xtimer_now_usec();
 
     /* read RTC to confirm value */
-    rtc_get_time(&time);
-    print_time("Clock value is now ", &time);
+    struct tm time_now;
+    rtc_get_time(&time_now);
+    print_time("Clock value is now ", &time_now);
 
     /* set initial alarm */
     inc_secs(&time, PERIOD);
     print_time("  Setting alarm to ", &time);
     rtc_set_alarm(&time, cb, NULL);
+    alarm_expected_usecs = start_usecs + US_PER_SEC * PERIOD;
 
     /* verify alarm */
     rtc_get_alarm(&time);
     print_time("   Alarm is set to ", &time);
     puts("");
+
+    printf("[%" PRIu32 "] First alarm expected at %" PRIu32 " microseconds\n",
+           xtimer_now_usec(),
+           alarm_expected_usecs);
 
     return 0;
 }


jcarrano:
I think this test is useful and necessary, although I'm concerned about the case where the RTC and Xtimer are powered by the same clock. Wouldn't the test stop working in that case?

benemorius:
> I think this test is useful and necessary, although I'm concerned about the case where the RTC and Xtimer are powered by the same clock. Wouldn't the test stop working in that case?

It works for me on kw41z where both xtimer and periph_rtc use the same low power clock source. I think most low power platforms will be this way. I can't think of a reason it should be a problem.

PeterKietzmann:
I think this change is useful. It already helped in #10764.

> I'm concerned about the case where the RTC and Xtimer are powered by the same clock.

@jcarrano I'm no convinced about that but also lacking knowledge about xtimers internals so I'd rather run some tests on exemplary devices. Do you know of any where your concerns apply?

aabadie:
> I think this change is useful.

I fully agree.
The Python test automation script should be updated, otherwise it's broken in the current state.

benemorius:
> The Python test automation script should be updated, otherwise it's broken in the current state.

I didn't see those scripts until now. Wow that's a very nice system.

Should we check that the error is less than some predetermined value? I didn't expect to see any platform with an error over 100us but @PeterKietzmann found in #10764 that it can be higher. Although it's possible that that's due to another problem.

PeterKietzmann:
> Should we check that the error is less than some predetermined value?

I think this would be a nice improvement. However, this requires certain board specific assumptions about maximum deviations. @MrKevinWeiss came to the great idea to look up worst-case parameters of clock sources at DigiKey which lead to a total number around 500ppm. With that we would at least find misconfigured timers

PeterKietzmann:
BTW: @MrKevinWeiss would you like to take this PR over?

MrKevinWeiss:
Only for +-250ppm for frequency stability and +-250ppm for frequency tolerance (not necessarily compounded but we can say that as a dumb worst case) for crystals...  Not to mention internal RC oscillators.
@PeterKietzmann Like is a strong word, but if we need it to be finished and added in to be runnable with the compile_and_test_for_board.py tool, then I will, it will only be after the uart test PR is merged though, FIFO style.

MrKevinWeiss:
Oops.  I need to find time to take this over!

benpicco:
That's pretty useful (also for RTT) - want to give it a rebase?

MrKevinWeiss:
I have #13483 to replace this.  It takes care of the conflict and adapts the python test.

PR #10764, opened 2019-01-14
cpu/kinetis/rtt: clear TPR when writing TSR

benemorius:
### Contribution description
Currently the `periph_rtt` implementation for `kinetis` ignores the subsecond RTC register (`RTC->TPR`). As a result, when `rtc_set_time()` is called, the RTC is left with a known second count but an unknown subsecond count.

This means that setting the RTC incurs an error of +/-1 second.

This PR clears the subsecond register `RTC->TPR` while setting the second register `RTC->TSR` to the requested value so that the RTC is left unambiguously set to the requested time.

### Testing procedure
Using #10763, run `tests/periph_rtc` with and without this PR on a `kw41z` or other `kinetis` platform.

The RTC alarm error output from the test should be always minimal (<100us) with this PR. Without it, the error will vary +/-1 second and will depend on how and when the last reset occurred. In particular, a power-on reset will synchronize the RTC clock domain and cause good error offsets even without this PR until the next non-POR reset.

### DIFF
--- a/cpu/kinetis/periph/rtt.c
+++ b/cpu/kinetis/periph/rtt.c
@@ -110,6 +110,8 @@ void rtt_set_counter(uint32_t counter)
 {
     /* Disable time counter before writing to the timestamp register */
     bit_clear32(&RTC->SR, RTC_SR_TCE_SHIFT);
+    RTC->TPR = 0;
+    /* write TSR after TPR, as clearing TPR bit 14 will increment TSR */
     RTC->TSR = counter;
     /* Enable when done */
     bit_set32(&RTC->SR, RTC_SR_TCE_SHIFT);


PeterKietzmann:
For reference I paste  the following from the MKW2xD Reference Manual:

**47.2.2 RTC Time Prescaler Register (RTC_TPR)**

*When the time counter is enabled, the TPR is read only and increments every 32.768 kHz clock cycle. The
time counter will read as zero when SR[TOF] or SR[TIF] are set. When the time counter is disabled, the
TPR can be read or written. The TSR[TSR] increments when bit 14 of the TPR transitions from a logic one
to a logic zero.*

PeterKietzmann:
I've tested *tests/periph_rtc* with #10763 repeatedly on a phytec pba-d-01-kw2x always with similar outcome:

**w/o this fix:**
```
# [28888] First alarm expected at 2018102 microseconds
# [1256147] Alarm! after 1238045 microseconds (error -761955 microseconds)
# [3256052] Alarm! after 3237950 microseconds (error -762050 microseconds)
# [5255957] Alarm! after 5237855 microseconds (error -762145 microseconds)
# [7255861] Alarm! after 7237759 microseconds (error -762241 microseconds)
```

**w/ fix:**
```
# [2018020] Alarm! after 1999919 microseconds (error -81 microseconds)
# [4017924] Alarm! after 3999823 microseconds (error -177 microseconds)
# [6017828] Alarm! after 5999727 microseconds (error -273 microseconds)
# [8017732] Alarm! after 7999631 microseconds (error -369 microseconds)
```

**without** this fix there is a big (erroneous) offset which varied from 10 - 80 ms (not us!). With this PR the offset was always around 100us.

benemorius:
> ```
> # [2018020] Alarm! after 1999919 microseconds (error -81 microseconds)
> # [4017924] Alarm! after 3999823 microseconds (error -177 microseconds)
> # [6017828] Alarm! after 5999727 microseconds (error -273 microseconds)
> # [8017732] Alarm! after 7999631 microseconds (error -369 microseconds)
> ```

It's interesting that the error is incrementing each time, and not by a small amount. I never saw that happen on kw41z. I wonder if the error would continue growing with additional iterations. Was this a coincidence or does it increment every time you test it?

It's unrelated to this PR, but there could be another problem there.

benemorius:
> It's interesting that the error is incrementing each time, and not by a small amount.

Actually I just realized that this must be because pba-d-01-kw2x uses different clock sources for xtimer and periph_rtc. So in this case any drift between the two clock sources will add to the error shown in the test.

PeterKietzmann:
@benemorius yes this seems like drift between clock sources. With 80*10^-7s/2s=40ppm the magnitude seems reasonable. To double check I ran the same application on a frdm-kw41z which uses same clock sources and did not show increasing errors.

PR #10773, opened 2019-01-15
cpu/stm_common/i2c: set fast mode flag

toonst:
### Contribution description

Set fast mode flag in CCR register. At the moment fast mode seems to be working fine even though it is not set, but it's cleaner to follow the datasheet.

### Testing procedure
Use I2C in fast mode.

### DIFF
--- a/cpu/stm32_common/periph/i2c_2.c
+++ b/cpu/stm32_common/periph/i2c_2.c
@@ -92,10 +92,12 @@ void i2c_init(i2c_t dev)
         case I2C_SPEED_NORMAL:
             /* 100Kbit/s */
             ccr = i2c_config[dev].clk / 200000;
+            ccr &= ~I2C_CCR_FS;
             break;
 
         case I2C_SPEED_FAST:
             ccr = i2c_config[dev].clk / 800000;
+            ccr |= I2C_CCR_FS;
             break;
 
         default:


toonst:
@MrKevinWeiss could you have a look?

MrKevinWeiss:
I can put it on the list to do first thing tomorrow!

MrKevinWeiss:
Well so far this is what I have found out...
First some stats (BOARD=nucleo-f401re, DUT=PHiLIP, 470 ohm external pullups used/needed)
According to the datasheet, setting that bit changes the Tlow
```
Sm mode or SMBus:
Thigh = CCR * TPCLK1
Tlow = CCR * TPCLK1
Fm mode:
If DUTY = 0:
Thigh = CCR * TPCLK1
Tlow = 2 * CCR * TPCLK1
If DUTY = 1: (to reach 400 kHz)
Thigh = 9 * CCR * TPCLK1
Tlow = 16 * CCR * TPCLK1
```
This is reflected in the traces below (white traces are with fast mode bit set, yellow without).  
fastmodetest1
fastmodetest2

This however messes up the frequency setting (with the new fast mode the 400 kHz signal becomes around 250 kHz).

Accoding to the [specs](https://www.nxp.com/docs/en/user-guide/UM10204.pdf) it states:

> A Fast-mode I2C-bus device can be used in a Standard-mode I2C-bus system, but the requirement tSU;DAT 250 ns must then be met. This will automatically be the case if the
device does not stretch the LOW period of the SCL signal. If such a device does stretch the LOW period of the SCL signal, it must output the next data bit to the SDA line
tr(max) + tSU;DAT = 1000 + 250 = 1250 ns (according to the Standard-mode I2C-bus specification) before the SCL line is released. Also the acknowledge timing must meet this
set-up time.

Which may be why it was working fine.

It also shows that standard mode needs a `Tlow=4.7us` `Thigh=4.0` about 55% duty cycle for the min requirements.  For fast mode it needs  `Tlow=1.3us` `Thigh=0.6` about 70% duty cycle for the min time.

This would probably by why the chip has a different duty cycle in the modes.

Now with this information we must also decide what is better.  If we think that a weaker pullup resistor will reduce the effective Thigh time (because it takes a while to charge) then we can use less power for the speed.  If we keep it at 50% duty cycle and we use very strong pullups then the Tlow would be 50% of 400kHz = 1.25us (out of spec).

Let me know what you think after digesting this info (power/speed vs safety decision)

PR #10780, opened 2019-01-16
cpu/stm32f0: add missing definition for M1 bit

yegorich:
### Contribution description

USART_CR1_M1 bit enables 7-bit mode for stm32f0x0 UARTs. All other headers for this device family already provide this definition. So this PR just adds missing ones.

### Issues/PRs references

PR #10743 depends on it.

### DIFF
--- a/cpu/stm32f0/include/vendor/stm32f030x8.h
+++ b/cpu/stm32f0/include/vendor/stm32f030x8.h
@@ -4757,7 +4757,7 @@ typedef struct
 #define USART_CR1_WAKE                USART_CR1_WAKE_Msk                       /*!< Receiver Wakeup method */
 #define USART_CR1_M_Pos               (12U)
 #define USART_CR1_M_Msk               (0x1U << USART_CR1_M_Pos)                /*!< 0x00001000 */
-#define USART_CR1_M                   USART_CR1_M_Msk                          /*!< Word Length */
+#define USART_CR1_M                   USART_CR1_M_Msk                          /*!< Word Length - Bit 0 */
 #define USART_CR1_MME_Pos             (13U)
 #define USART_CR1_MME_Msk             (0x1U << USART_CR1_MME_Pos)              /*!< 0x00002000 */
 #define USART_CR1_MME                 USART_CR1_MME_Msk                        /*!< Mute Mode Enable */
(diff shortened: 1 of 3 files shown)


cpu/stm32f0/include/vendor/stm32f030x8.h
```diff
@@ -4789,6 +4789,9 @@ typedef struct
 #define USART_CR1_EOBIE_Pos           (27U)
 #define USART_CR1_EOBIE_Msk           (0x1U << USART_CR1_EOBIE_Pos)            /*!< 0x08000000 */
 #define USART_CR1_EOBIE               USART_CR1_EOBIE_Msk                      /*!< End of Block interrupt enable */
+#define USART_CR1_M1_Pos              (28U)
```
aabadie: 7 bit mode is not supported by stm32f030x4/6/8, see reference manual DM00091010, table 85, page 597.

So I would not add the `USART_CR1_M1` definitions here.

cpu/stm32f0/include/vendor/stm32f031x6.h
```diff
@@ -4967,6 +4967,9 @@ typedef struct
 #define USART_CR1_EOBIE_Pos           (27U)
 #define USART_CR1_EOBIE_Msk           (0x1U << USART_CR1_EOBIE_Pos)            /*!< 0x08000000 */
 #define USART_CR1_EOBIE               USART_CR1_EOBIE_Msk                      /*!< End of Block interrupt enable */
+#define USART_CR1_M1_Pos              (28U)
```
aabadie: I haven't checked in details using the reference manual (it takes some time!), but if it is not in the CMSIS already, there is a high chance that 7 bit is not supported for this platform.

cpu/stm32f0/include/vendor/stm32f051x8.h
```diff
@@ -3259,7 +3259,7 @@ typedef struct
 #define  USART_CR1_PS                        ((uint32_t)0x00000200)            /*!< Parity Selection */
 #define  USART_CR1_PCE                       ((uint32_t)0x00000400)            /*!< Parity Control Enable */
 #define  USART_CR1_WAKE                      ((uint32_t)0x00000800)            /*!< Receiver Wakeup method */
-#define  USART_CR1_M0                        ((uint32_t)0x00001000)            /*!< Word length bit 0 */
+#define  USART_CR1_M0                        ((uint32_t)0x00001000)            /*!< Word length - Bit 0 */
```
aabadie: This CMSIS file could just be updated to the latest version. It seems a bit outdated. In my CubeMX version, this definition no longer exists.

cpu/stm32f0/include/vendor/stm32f051x8.h
```diff
@@ -3278,6 +3278,7 @@ typedef struct
 #define  USART_CR1_DEAT_4                    ((uint32_t)0x02000000)            /*!< Bit 4 */
 #define  USART_CR1_RTOIE                     ((uint32_t)0x04000000)            /*!< Receive Time Out interrupt enable */
 #define  USART_CR1_EOBIE                     ((uint32_t)0x08000000)            /*!< End of Block interrupt enable */
+#define  USART_CR1_M1                        ((uint32_t)0x10000000)            /*!< Word length - Bit 1 */
```
aabadie: Same here about the fact that 7 bit more might just not be supported by this platform, there is no definition in my CubeMX CMSIS file.

yegorich:
@aabadie you're right. Unfortunately the datasheets  don't refer to this table in the register description.

So the detection of 7-bit capability will be more difficult. I'll then introduce like `STM32_UART_7_BIT` macro.

PR #10789, opened 2019-01-16
pkg/jerryscript: namespace BUILD_DIR variable

cladmi:
### Contribution description

I forgot to git grep when merging `BUILD_DIR` and I found issues about it.

BUILD_DIR was introduced but the variable was already used by
`pkg/jerryscript` so there is a name collision.
Namespace it to prevent issues.

### Testing procedure

Build `examples/javascript` and the `build` directory should not be created, and the archives will be generated in `examples/javascript/bin/native/jerryscript/lib/`
When building in docker, `build` will be created but empty.

There is no more other usages now in the repository:

```
git grep '(BUILD_DIR)'
Makefile.include:override BUILD_DIR      := $(abspath $(BUILD_DIR))
makefiles/docker.inc.mk:        $(Q)mkdir -p $(BUILD_DIR)
makefiles/docker.inc.mk:            -v '$(BUILD_DIR):$(DOCKER_BUILD_ROOT)/build' \
```

### Issues/PRs references

Original PR https://github.com/RIOT-OS/RIOT/pull/10038

### DIFF
--- a/pkg/jerryscript/Makefile.jerryscript
+++ b/pkg/jerryscript/Makefile.jerryscript
@@ -1,4 +1,4 @@
-BUILD_DIR  ?= $(BINDIR)/jerryscript
+JERRYSCRIPT_BUILD_DIR  ?= $(BINDIR)/jerryscript
 
 JERRYHEAP  ?= 16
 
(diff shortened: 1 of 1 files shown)


cladmi:
Thanks for the review.

PR #10695, opened 2019-01-03
drivers/lpsxxx: refactor lps331ap and add support for lps25hb + lps22hb

aabadie:
### Contribution description

This PR is a rework of the initial ST lps331ap atmospheric pressure and temperature sensor. It does several things:
- move the original driver name from lps331ap to lpsxxx
- add support for lps25hb and lps22hb sensors
- use pseudo-modules for specific names
- add a check on the device type using the WHO_AM_I register
- change the default address from `0x44` to `0x5D` (why was it `0x44` initially, I don't know). The default address depends on the SDO/SA0 pad: when connected to ground it is `0x5D`, when connected to power supply, it is `0x5C`. On ST `x-nucleo-iks01a2` (lps22hb) and `x-nucleo-iks01a1` (lps25hb) extensions, the pad is connected to ground, thus using `0x5D` by default, makes more sense (to me)
- The driver API has changed to use return codes (specific to this driver) instead of returning the measurement values. The SAUL adaption was adapted accordingly
- A default configuration for lps22hb is provided for the ST B-l475E-IOT01A board since it embeds this sensor.

For simplicity, only basic polling mode with continuous measurements is provided by the driver. It should be possible to implement the other modes: One shot, Interrupt (with data ready) or FIFO in an extend version. Maybe a good candidate for a follow-up ? (Note that it might become a bit messy because of `#ifdef`s).

### Testing procedure

This PR can be fully tested using the IoT-LAB testbed, because some boards provides the 3 types of sensors:

1. Start an experiment on IoT-LAB with boards providing the different sensor: iotlab-m3 (lps331ap), st-iotnode (b-l475e-iot01a with lps22hb), arduino-zero (lps25hb on x-nucleo-iks01a1 extension) and nrf51dk (lps22hb on x-nucleo-iks01a2 extension):
```
$ iotlab-experiment submit -n test_pr -d 120 -l 1,archi=arduino-zero:xbee+site=saclay -l 1,archi=st-iotnode:multi+site=saclay -l 1,archi=m3:at86rf231+site=saclay -l 1,archi=nrf51dk:ble+site=saclay
```
2. Build and flash the `tests/driver_lpsxxx` test application on the different nodes and verify that it works:
```
$ make BOARD=iotlab-m3 IOTLAB_NODE=auto-ssh -C tests/driver_lpsxxx flash term
$ DRIVER=lps25hb make BOARD=arduino-zero IOTLAB_NODE=auto-ssh -C tests/driver_lpsxxx flash term
[...]

PR #10793, opened 2019-01-17
boards/thingy52: add dependency for on-board hts221/lps22hb sensors

aabadie:
### Contribution description

This PR provides a default I2C configuration for the nrf52 Thingy52 and add hts221 sensor as dependency when saul is used. lps22hb could also be added when #10695 will be merged. ccs811 requires support for the GPIO expander SX1509.

This PR is untested, I don't have this hardware. Maybe @haukepetersen will be able to test this.

### Testing procedure

- Build and flash `tests/driver_hts221` and verify the output is correct:
```
    $ make BOARD=thingy52 -C tests/driver_hts221 flash term
```
- Build and flash `examples/saul` and check the hts221 and lps22hb values:
```
    $ make BOARD=thingy52 -C examples/saul flash term
```

### Issues/PRs references

None

### DIFF
--- a/boards/thingy52/Makefile.dep
+++ b/boards/thingy52/Makefile.dep
@@ -1,6 +1,7 @@
 ifneq (,$(filter saul_default,$(USEMODULE)))
   USEMODULE += lis2dh12
   USEMODULE += hts221
+  USEMODULE += lps22hb
 endif
 
 include $(RIOTBOARD)/common/nrf52/Makefile.dep
--- a/boards/thingy52/Makefile.dep
+++ b/boards/thingy52/Makefile.dep
@@ -1,5 +1,6 @@
 ifneq (,$(filter saul_default,$(USEMODULE)))
   USEMODULE += lis2dh12
+  USEMODULE += hts221
 endif
 
 include $(RIOTBOARD)/common/nrf52/Makefile.dep
--- a/boards/thingy52/board.c
+++ b/boards/thingy52/board.c
@@ -21,8 +21,19 @@
 #include "cpu.h"
 #include "board.h"
 
+#if defined(MODULE_LPS22HB) || defined(MODULE_HTS221)
+#include "periph/gpio.h"
+#endif
+
 void board_init(void)
 {
+    /* LP22HB and HTS221 can only be powered with VREG when VDD_PWR_CTRL_PIN
+       is set. */
+#if defined(MODULE_LPS22HB) || defined(MODULE_HTS221)
+    gpio_init(VDD_PWR_CTRL_PIN, GPIO_OUT);
+    gpio_set(VDD_PWR_CTRL_PIN);
+#endif
+
     /* initialize the CPU */
     cpu_init();
 }
--- a/boards/thingy52/include/board.h
+++ b/boards/thingy52/include/board.h
@@ -34,6 +34,14 @@ extern "C" {
 /** @} */
 
 /**
+ * @name    Power switch pins definitions
+ * @{
+ */
+#define SPK_PWR_CTRL_PIN    GPIO_PIN(0, 29) /**< Speaker power switch */
+#define VDD_PWR_CTRL_PIN    GPIO_PIN(0, 30) /**< VDD power switch */
+/** @} */
+
+/**
  * @name    LIS2DH12 low power accelerometer configuration
  * @{
  */
--- a/boards/thingy52/include/board.h
+++ b/boards/thingy52/include/board.h
@@ -40,6 +40,12 @@ extern "C" {
 #define LIS2DH12_PARAM_I2C  I2C_DEV(1)
 /** @} */
 
+/**
+ * @name    LPS22HB device address
+ * @{
+ */
+#define LPSXXX_PARAM_ADDR   (0x5c)
+/** @} */
 
 /**
  * @brief   Initialize board specific hardware


maribu:
I will test this tomorrow

aabadie:
Maybe @haukepetersen could test this PR on its Thingy52 ?

PeterKietzmann:
> I will test this tomorrow

@maribu did you?

maribu:
Sadly not. I wanted to borrow my colleague's Thingy:52 for testing another PR and this one, but the Thingy:52 went missing. I only put this information in the other PR and forgot about this one.

Citrullin:
We have a lot of these Thingy52. If somebody needs one, I probably can give a few to you. I just have to ask my boss first. So, ping me, if you need one :)

Citrullin:
The hts221 should automaticlly be found when I use saul_default, right? 
So when I use saul in the shell the sensor should be there, right?

aabadie:
Thanks for having a look @Citrullin !

> So when I use saul in the shell the sensor should be there, right?

Yes, that's the idea. The `tests/driver_hts221` should also work.

MrKevinWeiss:
I cannot get an ack from the hts221, I tried 0x5F address and it doesn't work... The 0x19 address works though which is weird since it should be connected to I2C_DEV 1 not 0...
I believe it has something to do with the power settings.  It appears the VDD is actually controlled by this `VDD_PWD_CTRL` pin (p0.30).  It seems a high turns the power on and low turns the power off.

aabadie:
Thanks for testing @MrKevinWeiss! I'll have another look when time permits and will try to update.

MrKevinWeiss:
I can also return to it (when time permits as well), I tried a bit but couldn't get it going.

aabadie:
Now that #12290 was merged (with the same I2C configuration provided by this PR...), I rebased and also added support for lps22hb (that was merged in the meantime as well).

I also found why @MrKevinWeiss couldn't make it work. Will open a PR now. The bug is pretty awesome.

MrKevinWeiss:
Ya it still seems like the I2C itself is having some problems.  I cannot even use the i2c_scan command...

aabadie:
There might be other problems then. I already fixed another one with lps22hb: the address was not correct.

MrKevinWeiss:
Hmm ya.  With the two PRs I get lockup on i2c dev 0 and can read and scan on address 0x19 on i2c dev 1.  With master both 1 and 0 behave like one (which is incorrect).  I don't know why the terminal locks up on i2c dev 0 though...

MrKevinWeiss:
I would use `make BOARD=thingy52 -C tests/periph_i2c/ flash term` and `USEMODULE=i2c_scan make BOARD=thingy52 -C tests/shell/ flash term` to show what I am talking about.

aabadie:
> It appears the VDD is actually controlled by this VDD_PWD_CTRL pin (p0.30)

Can you try to initialize and set this pin high in board init ?

MrKevinWeiss:
Yup it seems to have fix it.

So adding
```c
void board_init(void)
{
    gpio_init(GPIO_PIN(0, 30), GPIO_OUT);
    gpio_set(GPIO_PIN(0, 30));
    /* initialize the CPU */
    cpu_init();
}
```

to both the #12372 and this PR then using the i2c_scan command on i2c_dev 0, no longer freezes the terminal but gives the following output.

```
> i2c_scan 0
2019-10-04 15:01:41,023 #  i2c_scan 0
2019-10-04 15:01:41,023 # Scanning I2C device 0...
2019-10-04 15:01:41,024 # addr not ack'ed = "-", addr ack'ed = "X", addr reserved = "R", error = "E"
2019-10-04 15:01:41,024 #      0 1 2 3 4 5 6 7 8 9 a b c d e f
2019-10-04 15:01:41,024 # 0x00 R R R R R R R R R R R R R R - -
2019-10-04 15:01:41,035 # 0x10 - - - - - - - - - - - - - - - -
2019-10-04 15:01:41,035 # 0x20 - - - - - - - - - - - - - - - -
2019-10-04 15:01:41,035 # 0x30 - - - - - - - - X - - - - - X -
2019-10-04 15:01:41,035 # 0x40 - - - - - - - - - - - - - - - -
2019-10-04 15:01:41,036 # 0x50 - - - - - - - - - - - - X - - X
2019-10-04 15:01:41,036 # 0x60 - - - - - - - - - - - - - - - -
2019-10-04 15:01:41,036 # 0x70 - - - - - - - - R R R R R R R R
```

I still don't understand why the terminal is frozen.  Maybe it could be that the i2c pins are actually held down and it is waiting on a flag or something. If the sda and scl pins were high impedance I would expect them to just return an error code.

aabadie:
@MrKevinWeiss, I rebased on latest master and added the management of VDD pin in `board_init` (note that this might have other low-power problems, according to the user manual). Can you give this another look ?

MrKevinWeiss:
I am out of office until Monday.  Maybe ping @leandrolanzieri or @jia200x.

MrKevinWeiss:
Well it seems to work now however the temperatures and humidity seem off...

```
2019-10-14 10:11:05,685 # Init HTS221 on I2C_DEV(0)
2019-10-14 10:11:05,685 # H: 0.0%, T: 8.9°C
2019-10-14 10:11:05,685 # H: 0.0%, T: 8.2°C
2019-10-14 10:11:07,657 # H: 0.0%, T: 8.2°C

```

also on the saul example
```
> saul read 1
2019-10-14 10:14:33,392 #  saul read 1
2019-10-14 10:14:33,392 # Reading from #1 (hts221|SENSE_TEMP)
2019-10-14 10:14:33,392 # Data:	     9.3°C
```

I don't know if getting the correct values is out of scope of this PR or not...  It isn't that cold in the office.

aabadie:
Does the lps22hb driver work ?

MrKevinWeiss:
Yup, that seems OK

```
> saul read 4
2019-10-14 10:48:14,685 #  saul read 4
2019-10-14 10:48:14,686 # Reading from #4 (lps22hb|SENSE_PRESS)
2019-10-14 10:48:14,686 # Data:	       1013 hPa
> saul read 5
2019-10-14 10:48:18,090 #  saul read 5
2019-10-14 10:48:18,091 # Reading from #5 (lps22hb|SENSE_TEMP)
2019-10-14 10:48:18,091 # Data:	     24.98°C
```

MrKevinWeiss:
If you want this in I would just raise an issue for the HTS221.  I or @JannesVolkens can look into it later.

MrKevinWeiss:
Looking at the datasheet it may also be improperly calibrated...  I would need further investigation but the more I look into this the more I think this is out of scope with the PR.  Any objections @aabadie ?

aabadie:
> Any objections @aabadie ?

Please ACK :)

PR #10796, opened 2019-01-17
sys/base64: Add size estimation functions

leandrolanzieri:
### Contribution description
This moves the size estimation calculation of encoding and decoding to inline functions so it can be used by other modules that may need to estimate buffer sizes.

### Testing procedure
Run the base64 unit test:
```
cd tests/unittests
make tests-base64 test
```

### Issues/PRs references
None

### DIFF
--- a/sys/include/base64.h
+++ b/sys/include/base64.h
@@ -33,6 +33,30 @@ extern "C" {
 #define BASE64_ERROR_DATA_IN_SIZE     (-4) /**< error value for invalid input buffer size      */
 
 /**
+ * @brief Estimates the amount of bytes needed for decoding @p base64_in_size
+ * characters from base64.
+ *
+ * @param[in] base64_in_size Size of the string to be decoded
+ * @return Amount of bytes estimated to be used after decoding
+ */
+static inline size_t base64_estimate_decode_size(size_t base64_in_size)
+{
+    return ((base64_in_size / 4) * 3);
+}
+
+/**
+ * @brief Estimates the length of the resulting string after encoding
+ * @p data_in_size bytes into base64.
+ *
+ * @param[in] data_in_size Amount of bytes to be encoded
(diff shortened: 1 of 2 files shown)


kaspar030:
Factoring that out is a good idea. But macros are bad. Please change them to static inline functions.

leandrolanzieri:
> Factoring that out is a good idea. But macros are bad. Please change them to static inline functions.

You are right. Done

sys/include/base64.h
```diff
@@ -32,6 +32,30 @@ extern "C" {
...
+/**
+ * @brief Estimates the length of the resulting string after encoding
+ * @p data_in_size bytes into base64.
+ *
+ * @param[in] data_in_size Amount of bytes to be encoded
+ * @return Amount of characters the output string is estimated to have
+ */
+static inline size_t base64_estimate_encode_size(size_t data_in_size)
+{
+    return (4 * ((data_in_size + 2) / 3));
```
danpetry: [CCN-LITE](https://github.com/cn-uofbasel/ccn-lite/blob/master/src/ccnl-utils/src/base64.c) adds a +1 to the end of this. Which would make the following check in `base64_encode()` more tight, so it's worth knowing why. Do you have an idea why they have this?

leandrolanzieri: I think it may be due to the fact that their 'output' is a NULL-terminated array of characters, so they add that extra byte at the end.

leandrolanzieri:
@danpetry squashed

sys/include/base64.h
danpetry: ok thanks!

danpetry:
Tests still pass, ACK

kaspar030:
If anyone cares, they could add a note regarding whether the trailing zero is part of the count.

PR #10804, opened 2019-01-17
boards/hifive1: fix file permissions of flasher.sh

pyropeter:
### Contribution description

* Permissions were wrong. I fixed them. Yay.

### Testing procedure

    make BOARD=hifive1 -C examples/default flash


pyropeter:
Git only tracks the execute bit, actually.

aabadie:
and go

PR #10805, opened 2019-01-17
boards/hifive1: fix CLOCK_CORECLOCK

pyropeter:
### Contribution description

* The `CLOCK_CORECLOCK` was completely wrong; fixed it

### Testing procedure

    make BOARD=hifive1 -C tests/leds/

### DIFF
--- a/boards/hifive1/include/periph_conf.h
+++ b/boards/hifive1/include/periph_conf.h
@@ -27,12 +27,8 @@ extern "C" {
  * @name    Core Clock configuration
  * @{
  */
-#define CLOCK_CORECLOCK             (1600000ul)
-/*
- * #define CLOCK_CORECLOCK           (20000000ul)
- * #define CLOCK_CORECLOCK           (27000000ul)
- * #define CLOCK_CORECLOCK           (38400000ul)
- */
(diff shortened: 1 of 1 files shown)


aabadie:
I don't have the hardware for testing on hifive1, it would be great if someone could test before merging.

pyropeter:
@maribu you seem to be involved with FE310 at the moment... Could you test this?

aabadie:
Maybe @kenrabold could test?

kenrabold:
Yes, it is fine. That #define is not used anywhere except in the new PR to support SPI. I don't have anything to test that the SPI code works, but the change above is OK. The board.c for the HiFive1 inits the main clock  in board_init_clock() and it uses a different #define for the desired clock rate. The UART driver then checks what the board freq is dynamically to compute the clock divider for the selected baud rate. That way the UART doesn't have to rely on getting the clock rate based on a set #define. Perhaps the SPI code can do something similar.

pyropeter:
`CLOCK_CORECLOCK` is also used in `tests/leds`. Thanks for the heads-up regarding the UART driver. I will look into that for SPI.

Issue #10645, opened 2018-12-20
stdio_init happens late during boot


gebart:
This applies to newlib platforms at least, possibly others.

### Actual behavior

stdio_init is called from _init which is run after board_init. This has the downside that it is not possible to use printf or DEBUG from board_init or from the init function of periph drivers which are initialized by periph_init, e.g. SPI, since those run before _init. 

### Desired behavior

stdio_init should be called as early as possible, for stdio_uart this would likely be immediately after clocking has been set up, (which is a prerequisite for generating the proper baud rate) 

### References

https://github.com/RIOT-OS/RIOT/pull/10615#discussion_r241721679

kaspar030:
How about moving it into kernel_init()?

dylad:
> How about moving it into kernel_init()?


Moving it to `kernel_init()` will not help since it is called even after `__libc_init_array() ` (for cortexm_common).
Maybe we should moving it to `cpu_init()` or `board_init()` ?

gebart:
board init might be the best location since the board developer should be aware of which hardware modules need to be initialized before stdio can work.

dylad:
As I see, most (almost all ?) `periph_init()` are called within `cpu_init()` of each platform. Moreover, `cpu_init()` is called from `board_init()`. But we cannot call `periph_init()` before our clocks are setup which happen during `cpu_init()` BUT to solve our issue we must call `stdio_init()` BEFORE the end `cpu_init()`. This also means we cannot call `stdio_init()` from `board_init()`.

It gives us two options here : 
- call `stdio_init` within` cpu_init()` but before `periph_init()`
- call `stdio_init` at the very beginning of `periph_init()`

I guess the later will reduce the number of modified files.

aabadie:
> I guess the latter will reduce the number of modified files.

This is what I did when working STM32s I2C unification, especially because of this.

maribu:
@aabadie, @dylad: Have you already created a PR for

> - call `stdio_init` at the very beginning of `periph_init()`

?

If not, I might just do that ;-)

Such a PR might result in some serious bug fixing. E.g. the MSB-A2 did not run `periph_init()` at all until recently. Other bugs like that on the older and less commonly used boards might be exposed by that. (Which will look like a bad thing at first, but in retrospect, when all the hard labor of bug-fixing is done, it will likely look like a good thing ;-))

maribu:
Fixed in https://github.com/RIOT-OS/RIOT/pull/11367

PR #10806, opened 2019-01-17
boot: Moved stdio_init() into periph_init()

maribu:
### Contribution description

Prior to this commit `stdio` is initialized at different stages during boot depending on the board used. This commit moves the initialization of `stdio` to the beginning of `periph_init()`. This commit makes `stdio` available at a predictable and hardware independent stage during boot, which makes the use of `stdio` (e.g. via `DEBUG(...)`) during boot more robust.

### Testing procedure

If `examples/hello_world` still compiles and prints, this would be a good indicator that `stdio` is still initialized properly.

### Issues/PRs references
Addresses issue https://github.com/RIOT-OS/RIOT/issues/10645
Related to PR https://github.com/RIOT-OS/RIOT/pull/10615

#### ARM familiy
- [x] MSB-A2 (@maribu)
- [x] Blue Pill (@maribu)
- [x] Arduino Due (@maribu)
- [x] nucleo-f103rb (@MrKevinWeiss)
- [x] nucleo-f303re (@MrKevinWeiss)
- [x] nucleo-l476lg (@MrKevinWeiss)

#### ESP familiy
- [x] Some ESP8266 board
- [x] Some ESP32 board

#### RISC-V family
- [x] HiFive1 (@kenrabold)

#### Native Family
- [x] "native" board (@maribu, maybe some else can confirm?)

#### ATmega Family
- [x] Arduino Uno (@maribu)
- [x] Arduino Mega2560 (@maribu)

#### MSP430 Family
 - [x] WSN430 Version 1.3b (Using FIT/IoT Lab @maribu)

#### MIPS Familiy
 - [x] pic32-wifire (@francois-berder)
 - [x] pic32-clicker (@francois-berder)

------
Update 1: Beatified markdown output
Update 2: Added tracker for testing
Update 3: Added testing results for Nulceo-F{103RB,303RE,401RE}
Update 4: Un-checked all tests because of force push
Update 5: Added new testing results for MSB-A2, Blue Pill, Arduino Due, Arduino Mega 2560
Update 6: Added test case for `stdio_rtt` to todo list
Update 7: Added test results for WiFire and Clicker board
Update 8: Removed `stdio_rtt` testing (which is not working in master anyway). Testing for ARM should be enough now.
Update 9: Added RISC-V test result
Update 10: Added MSP430 test result
Update 11: Added Arduino Uno test result

### DIFF
--- a/drivers/periph_common/init.c
+++ b/drivers/periph_common/init.c
@@ -55,7 +55,7 @@ void periph_init(void)
 #ifdef MODULE_AVR_LIBC_EXTRA
     atmega_stdio_init();
 #endif /* MODULE_AVR_LIBC_EXTRA */
-#else
+#elif defined(MODULE_STDIO_UART) || defined(MODULE_STDIO_RTT)
     stdio_init();
 #ifdef CPU_FE310
     setvbuf(stdout, NULL, _IONBF, 0);
--- a/drivers/periph_common/init.c
+++ b/drivers/periph_common/init.c
@@ -20,13 +20,9 @@
  * @}
  */
 
-#ifdef MODULE_NEWLIB_SYSCALLS_FE310
-#include <stdio.h>
-#endif
-
-#ifdef MODULE_ATMEGA_COMMON
 #include "cpu.h"
-#else
+
+#ifndef HAVE_ARCH_STDIO_INIT
 #include "stdio_base.h"
 #endif
 #ifdef MODULE_PERIPH_I2C
@@ -45,22 +41,26 @@
 #include "periph/hwrng.h"
 #endif
 
+#ifndef HAVE_ARCH_STDIO_INIT
+/**
+ * @brief Initializes stdio
+ *
+ * @details This function can be overwritten by `#define`ing
+ * `HAVE_ARCH_STDIO_INIT` in cpu.h and declaring a function with the same
+ * signature there. This allows handling architecture specific stuff if needed.
+ */
+static inline void arch_stdio_init(void)
+{
+    stdio_init();
+}
+#endif /* HAVE_ARCH_STDIO_INIT */
 
 void periph_init(void)
 {
     /* initialize stdio first to allow DEBUG() during later stages */
-#ifdef CPU_NATIVE
-    /* nothing to do for native */
-#elif defined(MODULE_ATMEGA_COMMON)
-#ifdef MODULE_AVR_LIBC_EXTRA
-    atmega_stdio_init();
-#endif /* MODULE_AVR_LIBC_EXTRA */
-#elif defined(MODULE_STDIO_UART) || defined(MODULE_STDIO_RTT)
-    stdio_init();
-#ifdef CPU_FE310
-    setvbuf(stdout, NULL, _IONBF, 0);
-#endif /* CPU_FE310 */
-#endif /* CPU_NATIVE */
+#if defined(MODULE_STDIO_UART) || defined(MODULE_STDIO_RTT)
+    arch_stdio_init();
+#endif
 
     /* initialize configured I2C devices */
 #ifdef MODULE_PERIPH_I2C
--- a/drivers/periph_common/init.c
+++ b/drivers/periph_common/init.c
@@ -20,6 +20,15 @@
  * @}
  */
 
+#ifdef MODULE_NEWLIB_SYSCALLS_FE310
+#include <stdio.h>
+#endif
+
+#ifdef MODULE_ATMEGA_COMMON
+#include "cpu.h"
+#else
+#include "stdio_base.h"
+#endif
 #ifdef MODULE_PERIPH_I2C
 #include "periph/i2c.h"
 #endif
@@ -36,8 +45,23 @@
 #include "periph/hwrng.h"
 #endif
 
+
 void periph_init(void)
 {
+    /* initialize stdio first to allow DEBUG() during later stages */
+#ifdef CPU_NATIVE
+    /* nothing to do for native */
+#elif defined(MODULE_ATMEGA_COMMON)
+#ifdef MODULE_AVR_LIBC_EXTRA
+    atmega_stdio_init();
+#endif /* MODULE_AVR_LIBC_EXTRA */
+#else
+    stdio_init();
+#ifdef CPU_FE310
+    setvbuf(stdout, NULL, _IONBF, 0);
+#endif /* CPU_FE310 */
+#endif /* CPU_NATIVE */
+
     /* initialize configured I2C devices */
 #ifdef MODULE_PERIPH_I2C
     for (unsigned i = 0; i < I2C_NUMOF; i++) {
--- a/boards/common/atmega/board.c
+++ b/boards/common/atmega/board.c
@@ -33,11 +33,6 @@ void board_init(void)
 {
 
     atmega_set_prescaler(CPU_ATMEGA_CLK_SCALE_INIT);
-
-#ifdef MODULE_AVR_LIBC_EXTRA
-    atmega_stdio_init();
-#endif
-
     cpu_init();
     led_init();
     irq_enable();
--- a/boards/common/msb-430/board_init.c
+++ b/boards/common/msb-430/board_init.c
@@ -21,7 +21,6 @@
 #include "cpu.h"
 #include "irq.h"
 #include "board.h"
-#include "stdio_base.h"
 #include "periph_conf.h"
 #include "msp430.h"
 #include "debug.h"
@@ -198,7 +197,4 @@ void board_init(void)
     msb_ports_init();
 
     msp430_set_cpu_speed(CLOCK_CORECLOCK);
-
-    /* finally initialize STDIO */
-    stdio_init();
 }
--- a/boards/common/wsn430/board_init.c
+++ b/boards/common/wsn430/board_init.c
@@ -23,7 +23,6 @@
 #include "cpu.h"
 #include "irq.h"
 #include "assert.h"
-#include "stdio_uart.h"
 #include "periph_conf.h"
 
 enum {
@@ -96,7 +95,4 @@ void board_init(void)
     unsigned state = irq_disable();
     msp430_init_cpuclk(CLOCK_MODE);
     irq_restore(state);
-
-    /* initialize STDIO over UART */
-    stdio_init();
 }
--- a/boards/hifive1/board.c
+++ b/boards/hifive1/board.c
@@ -114,8 +114,8 @@ void board_init_clock(void)
 void board_init(void)
 {
     /* Initialize CPU and clocks */
-    cpu_init();
     board_init_clock();
+    cpu_init();
 
     /* Configure pin muxing for UART0 */
     GPIO_REG(GPIO_OUTPUT_VAL) |= IOF0_UART0_MASK;
--- a/boards/pic32-clicker/clicker.c
+++ b/boards/pic32-clicker/clicker.c
@@ -10,7 +10,6 @@
 #include <stdio.h>
 #include <stdint.h>
 #include "periph/gpio.h"
-#include "periph/uart.h"
 #include "bitarithm.h"
 #include "board.h"
 #include "cpu.h"
@@ -25,11 +24,6 @@ void board_init(void)
      */
     U3RXR =    0x2;          /*connect pin RPF5 to UART3 RX*/
     RPF4R =    0x1;          /*connect pin RPF4 to UART3 TX*/
-
-    /* intialise UART used for debug (printf) */
-#ifdef DEBUG_VIA_UART
-    uart_init(DEBUG_VIA_UART, DEBUG_UART_BAUD, NULL, 0);
-#endif
 
     /* Turn off all LED's */
     gpio_init(LED1_PIN, GPIO_OUT);
--- a/boards/pic32-clicker/include/board.h
+++ b/boards/pic32-clicker/include/board.h
@@ -65,6 +65,12 @@ extern "C" {
 /** @} */
 
 /**
+  * @brief We route STDIO via UART3 on this board,
+  *        this is the UART connected to the MikroBUS
+  */
+#define STDIO_UART_DEV      UART_DEV(3)
+
+/**
  * @brief   Board level initialization
  */
 void board_init(void);
--- a/boards/pic32-clicker/include/periph_conf.h
+++ b/boards/pic32-clicker/include/periph_conf.h
@@ -46,15 +46,11 @@ extern "C" {
 /**
   * @name    UART Definitions
   *          There are 4 UARTS available on this CPU.
-  *          We route debug via UART3 on this board,
-  *          this is the UART connected to the MikroBUS
   *
   *          Note Microchip number the UARTS 1->4
   * @{
   */
 #define UART_NUMOF          (4)
-#define DEBUG_VIA_UART      (3)
-#define DEBUG_UART_BAUD     (9600)
 /** @} */
 
 #ifdef __cplusplus
--- a/boards/pic32-wifire/include/board.h
+++ b/boards/pic32-wifire/include/board.h
@@ -77,6 +77,12 @@ extern "C" {
 /** @} */
 
 /**
+ * @brief   Use the 4th UART for STDIO on this board
+ *          this is the UART connected to the FTDI USB <-> UART device.
+ */
+#define STDIO_UART_DEV      UART_DEV(4)
+
+/**
  * @brief   Board level initialization
  */
 void board_init(void);
--- a/boards/pic32-wifire/include/periph_conf.h
+++ b/boards/pic32-wifire/include/periph_conf.h
@@ -43,15 +43,11 @@ extern "C" {
 /**
   * @name    UART Definitions
   *          There are 6 UARTS available on this CPU.
-  *          We route debug via UART4 on this board,
-  *          this is the UART connected to the FTDI USB <-> UART device.
   *
-  *          Note Microchip number the UARTS 1->4.
+  *          Note Microchip number the UARTS 1->6.
   * @{
   */
 #define UART_NUMOF          (6)
-#define DEBUG_VIA_UART      (4)
-#define DEBUG_UART_BAUD     (9600)
 /** @} */
 
 #ifdef __cplusplus
--- a/boards/pic32-wifire/wifire.c
+++ b/boards/pic32-wifire/wifire.c
@@ -12,7 +12,6 @@
 #include <stdint.h>
 #include "periph/gpio.h"
 #include "periph/hwrng.h"
-#include "periph/uart.h"
 #include "bitarithm.h"
 #include "board.h"
 #include "cpu.h"
@@ -27,11 +26,6 @@ void board_init(void)
      */
     U4RXR =   0xb;            /* connect pin RPF2 to UART 4 RX */
     RPF8R =   0x2;            /* connect pin RPF8 to UART 4 TX */
-
-    /* intialise UART used for debug (printf) */
-#ifdef DEBUG_VIA_UART
-    uart_init(DEBUG_VIA_UART, DEBUG_UART_BAUD, NULL, 0);
-#endif
 
     /* Turn off all LED's */
     gpio_init(LED1_PIN, GPIO_OUT);
--- a/boards/telosb/board.c
+++ b/boards/telosb/board.c
@@ -9,7 +9,6 @@
 
 #include "cpu.h"
 #include "board.h"
-#include "stdio_uart.h"
 
 void uart_init(void);
 
@@ -121,9 +120,6 @@ void board_init(void)
 
     telosb_ports_init();
     msp430_init_dco();
-
-    /* initialize STDIO */
-    stdio_init();
 
     /* enable interrupts */
     __bis_SR_register(GIE);
--- a/boards/waspmote-pro/board.c
+++ b/boards/waspmote-pro/board.c
@@ -58,10 +58,6 @@ void board_init(void)
 #endif
 #endif
 
-#ifdef MODULE_AVR_LIBC_EXTRA
-    atmega_stdio_init();
-#endif
-
     cpu_init();
     led_init();
     irq_enable();
--- a/boards/z1/board.c
+++ b/boards/z1/board.c
@@ -24,7 +24,6 @@
 
 #include "cpu.h"
 #include "board.h"
-#include "stdio_uart.h"
 
 static void z1_ports_init(void)
 {
@@ -214,9 +213,6 @@ void board_init(void)
 
     /* initializes DCO */
     msp430_init_dco();
-
-    /* initialize STDIO */
-    stdio_init();
 
     /* enable interrupts */
     __bis_SR_register(GIE);
--- a/cpu/atmega_common/avr_libc_extra/atmega_stdio.c
+++ b/cpu/atmega_common/avr_libc_extra/atmega_stdio.c
@@ -41,7 +41,7 @@ static int _uart_getchar(FILE *stream)
     return (int)c;
 }
 
-void atmega_stdio_init(void)
+void arch_stdio_init(void)
 {
     stdio_init();
 
--- a/cpu/atmega_common/include/cpu.h
+++ b/cpu/atmega_common/include/cpu.h
@@ -59,6 +59,11 @@ extern "C"
 /** @} */
 
 /**
+ * @brief  Declare that a custom arch_stdio_init is provided
+ */
+#define HAVE_ARCH_STDIO_INIT
+
+/**
  * @brief global in-ISR state variable
  */
 extern volatile uint8_t __in_isr;
@@ -150,7 +155,7 @@ static inline void atmega_set_prescaler(uint8_t clk_scale)
 /**
  * @brief   Initializes avrlibc stdio
  */
-void atmega_stdio_init(void);
+void arch_stdio_init(void);
 
 #ifdef __cplusplus
 }
--- a/cpu/esp32/startup.c
+++ b/cpu/esp32/startup.c
@@ -295,14 +295,9 @@ static NORETURN void IRAM system_init (void)
     srand(hwrand());
 
     #if defined(MODULE_NEWLIB_SYSCALLS_DEFAULT)
-    /*
-     * initialization as it should be called from newlibc (includes the
-     * execution of stdio_init)
-    */
+    /* initialization as it should be called from newlibc */
     extern void _init(void);
     _init();
-    #elif defined(MODULE_STDIO_UART)
-    stdio_init();
     #endif
 
     /* add SPI RAM to heap if enabled */
--- a/cpu/fe310/include/cpu.h
+++ b/cpu/fe310/include/cpu.h
@@ -25,9 +25,26 @@
 #ifndef CPU_H
 #define CPU_H
 
+#include <stdio.h>
+#include "stdio_base.h"
+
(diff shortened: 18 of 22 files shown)


maribu:
**WARNING**: This PR can easily break stuff... So the more testing the better. But testing is easy and fast :-)

maribu:
@kenrabold: Mind to test this on the `hifive1`?
@gschorcht: Mind to test this on the ESP family?
@kYc0o: Mind to test this for some ATmega boards? (The ATmega boards had consistent `stdio` initialization, so if it works for one it should *theoretically* work for all...)
@MrKevinWeiss / @dylad / @kaspar030: Mind to test for your favorite ARM board?

maribu:
Oh, the msp430 family also needs testing, of course...

maribu:
Being honest, the impact is actually minor, but the chance to break stuff is. So I still felt to better set the "Impact: major" label in order to ring some alarm bells :-)

MrKevinWeiss:
Can do!

maribu:
Done! Compilation errors fixed and squashed. Reviewing & testing can start now :-)

MrKevinWeiss:
I ran all tests (release specs compile_and_test_for_board.py) on the nucleo-f103rb/f401re/f303re.  The only tests that failed are likely not due to this PR

```
nucleo-f103rb
Failures during test:
- [tests/driver_grove_ledbar](tests/driver_grove_ledbar/test.failed)
- [tests/driver_my9221](tests/driver_my9221/test.failed)
- [tests/gnrc_ipv6_ext](tests/gnrc_ipv6_ext/test.failed)
- [tests/gnrc_rpl_srh](tests/gnrc_rpl_srh/test.failed)
- [tests/pkg_fatfs_vfs](tests/pkg_fatfs_vfs/test.failed)
- [tests/xtimer_periodic_wakeup](tests/xtimer_periodic_wakeup/test.failed)
- [tests/xtimer_usleep_short](tests/xtimer_usleep_short/test.failed)
```

```
nucleo-f401re
Failures during test:
- [tests/driver_grove_ledbar](tests/driver_grove_ledbar/test.failed)
- [tests/driver_my9221](tests/driver_my9221/test.failed)
- [tests/gnrc_ipv6_ext](tests/gnrc_ipv6_ext/test.failed)
- [tests/gnrc_rpl_srh](tests/gnrc_rpl_srh/test.failed)
- [tests/pkg_fatfs_vfs](tests/pkg_fatfs_vfs/test.failed)
```

```
nucleo-f303re
Failures during test:
- [tests/driver_grove_ledbar](tests/driver_grove_ledbar/test.failed)
- [tests/driver_my9221](tests/driver_my9221/test.failed)
- [tests/gnrc_ipv6_ext](tests/gnrc_ipv6_ext/test.failed)
- [tests/gnrc_rpl_srh](tests/gnrc_rpl_srh/test.failed)
- [tests/pkg_fatfs_vfs](tests/pkg_fatfs_vfs/test.failed)
```

maribu:
@gschorcht @MrKevinWeiss: Thanks for the testing! :-)

kenrabold:
The PR as is breaks the HiFive1 board. The code for the board was never setup to call periph_init(), so the modifications stop all stdio output. Easy fix however. Similar to the msba2 mods, you can add the periph_init() call in board.c for HiFive1:
 
```
    /* Initialize newlib-nano library stubs */
    nanostubs_init();

    /* Initialize peripherals */
    periph_init();
}

```
(and the #include "periph/init.h" )

You will also need to update the Makefile.include in boards\hifive1 to include the periph modules:


```
# Export the peripheral drivers to be linked into the final binary:
export USEMODULE += periph

# include common periph code
export USEMODULE += periph_common

```

maribu:
Oh, you noticed that ugly hack in the MSB-A2. If I remember correctly, `periph_init()` is to be called by `cpu_init()`. The lack of `cpu_init()` for MSB-A2 lead to the quick and dirty fix for the MSB-A2 which hopefully gets replaced be a proper fix...

Maybe we can try a proper fix for the HiFive1 from the beginning, as it is not a legacy board.

Am 18. Januar 2019 23:01:34 MEZ schrieb kenrabold <notifications@github.com>:
>The PR as is breaks the HiFive1 board. The code for the board was never
>setup to call periph_init(), so the modifications stop all stdio
>output. Easy fix however. Similar to the msba2 mods, you can add the
>periph_init() call in board.c for HiFive1:
> 
>```
>    /* Initialize newlib-nano library stubs */
>    nanostubs_init();
>
>    /* Initialize peripherals */
>    periph_init();
>}
>
>```
>(and the #include "periph/init.h" )
>
>You will also need to update the Makefile.include in boards\hifive1 to
>include the periph modules:
>
>
>```
># Export the peripheral drivers to be linked into the final binary:
>export USEMODULE += periph
>
># include common periph code
>export USEMODULE += periph_common
>
>```
>
>
>
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/RIOT-OS/RIOT/pull/10806#issuecomment-455702016

-- 
Via mobile phone

maribu:
With https://github.com/RIOT-OS/RIOT/pull/10825 merged, this PR now works also on the HiFive1.

May I rebase against current master to fix the conflicts. (This would also make re-testing on the HiFive1 easier, as it than should work without needing to cherry-pick https://github.com/RIOT-OS/RIOT/pull/10825)

maribu:
I will just interpret the absence of rejections as an implicit agreement that I may rebase to solve the conflicts :-)

maribu:
I did the rebase. Please test again. (Hopefully now the HiFive1 will also work :-))

MrKevinWeiss:
```
- [tests/driver_grove_ledbar](tests/driver_grove_ledbar/test.failed)
- [tests/driver_hd44780](tests/driver_hd44780/test.failed)
- [tests/driver_my9221](tests/driver_my9221/test.failed)
- [tests/gnrc_ipv6_ext](tests/gnrc_ipv6_ext/test.failed)
- [tests/gnrc_rpl_srh](tests/gnrc_rpl_srh/test.failed)
- [tests/pkg_fatfs_vfs](tests/pkg_fatfs_vfs/test.failed)
- [tests/pthread_rwlock](tests/pthread_rwlock/test.failed)
- [tests/xtimer_periodic_wakeup](tests/xtimer_periodic_wakeup/test.failed)
- [tests/xtimer_usleep_short](tests/xtimer_usleep_short/test.failed)
```
The failing tests on the nucleo-f103rb, still don't think any problems caused by this device.

maribu:
@francois-berder, @aabadie: Could one of you test this on one (or better two) MIPS board?

@rousselk, @aabadie, @haukepetersen, @OlegHahm : Could one of you test this on one (or better two) MSP430 based board?

@kYc0o: Could you test this on the Waspmote Pro and the Arduino Uni? (I could test this myself on the Uno, but I would feel more confident when at least two boards per family are tested independently from me.)

MrKevinWeiss:
```
ERROR:nucleo-l476rg:Tests failed: 6
Failures during test:
- [tests/driver_grove_ledbar](tests/driver_grove_ledbar/test.failed)
- [tests/driver_my9221](tests/driver_my9221/test.failed)
- [tests/gnrc_ipv6_ext](tests/gnrc_ipv6_ext/test.failed)
- [tests/gnrc_rpl_srh](tests/gnrc_rpl_srh/test.failed)
- [tests/periph_rtt](tests/periph_rtt/test.failed)
- [tests/pkg_fatfs_vfs](tests/pkg_fatfs_vfs/test.failed)
```
Seems OK for the nucleo-l476rg

MrKevinWeiss:
Hmmm I have some issue with the ESP32.  If I run something with a shell it works but hello_world doesn't output anything, even after reset.  It works on master though so I think it is something with the PR...

it seems to actually not work, I can't get it to output past the following (even for the tests/leds)
```
2019-02-07 14:45:35,916 - INFO # Starting ESP32 with ID: 7d240ac40dbeb4
2019-02-07 14:45:35,916 - INFO # 
2019-02-07 14:45:35,918 - INFO # Current clocks in Hz: CPU=80000000 APB=80000000 XTAL=40000000 SLOW=150000
2019-02-07 14:45:35,920 - INFO # PRO cpu is up (single core mode, only PRO cpu is used)
2019-02-07 14:45:35,921 - INFO # PRO cpu starts user code
```

maribu:
The above lines are printed using `ets_printf()` from `cpu/esp32/startup.c`, not using RIOT's stdio. Thus they are not affected.

Also, for the ESP32 this PR actually moves the stdio initialization to a later point in time during boot. One of the goal for writing this PR was to make it available earlier...

@gschorcht: Could the later initialization of stdio be the issue?

@MrKevinWeiss: Could you try it after rebasing it against the master state it worked for you, just to rule out it is an issue unrelated to this PR?

gschorcht:
@MrKevinWeiss @maribu From my point of view, the PR should not affect ESP32. `ets_printf` as well as `stdio_uart` are using `UART_DEV(0)` which is already initialized by the bootloader. The system initalization function only sets the baudrate to configured `STDIO_UART_BAUDRATE`: https://github.com/RIOT-OS/RIOT/blob/6037fa2536b1b2267f175b614e3074baeb35cc30/cpu/esp32/startup.c#L257-L261

Function `uart_init` which is called during `stdio_init` has no affect for `UART_DEV(0)`. The `uart_init` function only initializes the UART interfaces `UART_DEV(1)` or `UART_DEV(2)`.

gschorcht:
@maribu I thought I tested `examples/hello_world` for ESPs already some weeks ago :thinking:

gschorcht:
@maribu I have to revise my statement, ESP8266 as well as ESP32 are not working with the changes in this PR. They are both working with current master.

gschorcht:
@maribu @MrKevinWeiss `tests/shell` which uses module `stdio_uart` is working properly with ESP32 and this PR. That is, the changes in this PR seem to work. I'm not sure, but one reason that `examples/hello_world` seems to be not working might be that it exits before all characters have been written to the console. ESP32 has a FIFO buffer of 128 characters. Maybe, some synchronisation have to be added to the `uart_init` function.

maribu:
A, I see. There was is some workaround for the RISC-V disabling buffering on stdio which I cargo culted (= copy pasted without understanding). That seems like a solution to the problem with the ESP32, but might have a performance impact

gschorcht:
@maribu At the moment, I'm not really understanding what exactly happens. It was just a guess that it might be related to some synchronization issue. However, adding `uart_tx_wait_idle(uart)` to `uart_init` didn't help. At least one thing is clear, if `uart_init` is called from `stdio_init` the TX queue is reset.

gschorcht:
@maribu Changing `UART_FIFO_MAX` from 127 to 1 which corresponds to 1 place in FIFO and thus no FIFO also didn't help.

gschorcht:
@maribu I found the problem. ESP8266 and ESP32 simply crashes, when `periph_init` calls `stdio_init` but module `stdio_init` is not enabled. You get undefined references during the linking.

Maybe `periph_init` should only call `stdio_init` if there is some `stdio_*` module enabled. Alternatively, ESP8266 and ESP32 could always enable `stdio_uart` module.

gschorcht:
@maribu If you don't want to make the call of `stdio_init` dependent on the existence of any `stdio_*` module, these would be the changes for ESP8266 and ESP32.
```diff
diff --git a/cpu/esp32/Makefile.dep b/cpu/esp32/Makefile.dep
index c5d8e5803..59f319042 100644
--- a/cpu/esp32/Makefile.dep
+++ b/cpu/esp32/Makefile.dep
@@ -91,10 +91,6 @@ ifneq (,$(filter xtimer,$(USEMODULE)))
     USEMODULE += newlib_syscalls_default
 endif
 
-ifneq (,$(filter newlib_syscalls_default,$(USEMODULE)))
-    USEMODULE += stdio_uart
-endif
-
 # if SPI RAM is enabled, ESP-IDF heap and quot flash mode have to be used
 ifneq (,$(filter esp_spi_ram,$(USEMODULE)))
     USEMODULE += esp_idf_heap
diff --git a/cpu/esp32/Makefile.include b/cpu/esp32/Makefile.include
index 738fcbd71..81c20338e 100644
--- a/cpu/esp32/Makefile.include
+++ b/cpu/esp32/Makefile.include
@@ -69,6 +69,7 @@ USEMODULE += log
 USEMODULE += periph
 USEMODULE += periph_common
 USEMODULE += random
+USEMODULE += stdio_uart
 USEMODULE += xtensa
 
 INCLUDES += -I$(RIOTCPU)/$(CPU)/vendor/
diff --git a/cpu/esp8266/Makefile.dep b/cpu/esp8266/Makefile.dep
index cc2422d6f..167b4231c 100644
--- a/cpu/esp8266/Makefile.dep
+++ b/cpu/esp8266/Makefile.dep
@@ -48,6 +48,10 @@ ifneq (, $(filter shell, $(USEMODULE)))
     USEMODULE += xtimer
 endif
 
+ifneq (, $(filter stdio_uart, $(USEMODULE)))
+    USEMODULE += xtimer
+endif
+
 ifneq (, $(filter xtimer, $(USEMODULE)))
     USEMODULE += newlib_syscalls_default
 endif
@@ -56,7 +60,3 @@ ifneq (, $(filter vfs, $(USEMODULE)))
     USEMODULE += newlib_syscalls_default
     USEMODULE += xtimer
 endif
-
-ifneq (, $(filter newlib_syscalls_default, $(USEMODULE)))
-    USEMODULE += stdio_uart
-endif
diff --git a/cpu/esp8266/Makefile.include b/cpu/esp8266/Makefile.include
index 4658d192a..369a8be21 100644
--- a/cpu/esp8266/Makefile.include
+++ b/cpu/esp8266/Makefile.include
@@ -62,6 +62,7 @@ USEMODULE += periph_common
 USEMODULE += ps
 USEMODULE += random
 USEMODULE += sdk
+USEMODULE += stdio_uart
 USEMODULE += xtensa
 
 ifneq (, $(filter pthread, $(USEMODULE)))
```

maribu:
> You get undefined references during linking in that case.

Strange, I thought the build would abort on linking errors?

> Maybe periph_init should only call stdio_init if there is some stdio_* module enabled.

I hope the fixup commit above does this. Mind to test again?

maribu:
@haukepetersen: Would you mind testing on the Thingy:52 or RuuviTag (or do you know someone having access to one of theses boards)?

gschorcht:
> Strange, I thought the build would abort on linking errors?

Normally yes, but I had to change `LINKFLAGS += -Wl,--warn-unresolved-symbols` for ESP* to get all tests compiled with murdock. I don't remember exactly for which tests or which packages I had to do this.

gschorcht:
@maribu Tested. Work as expected for me.

maribu:
> Tested. Work as expected for me.

Perfect. It is still a pity that it ends up moving `stdio` initialization to a later point in time for the ESP devices. May [this GSoC project](https://github.com/RIOT-OS/RIOT/wiki/GSOC-Idea-List#project-s3-unify-riots-startup-code-across-architectures) will allow use to find a better place to move `stdio_init()` to.

gschorcht:
@maribu
> It is still a pity that it ends up moving `stdio` initialization to a later point in time for the ESP devices.

I don't think that it is a big problem since UART is already working before `stdio_init` without any problems.

francois-berder:
@maribu I tried on a WiFire board and it gave me a panic message while it worked on master. I will investigate later.

maribu:
@francois-berder: Thanks for pointing that out. Looks like a failed [assertion](http://api.riot-os.org/assert_8h.html#a3153a272f18d0f805028fce7e4337b53). Maybe you can recompile with `CFLAGS += DEBUG_ASSERT_VERBOSE` added to your `Makefile` and test again to get more output?

francois-berder:
So I finally found why it was not working on the WiFire board (and Clicker as well). By default, `STDIO_UART_DEV` is set to `UART_DEV(0)`. Since `UART_DEV` is not redefined, this means that `STDIO_UART_DEV` has value 0 and the call to `uart_init(STDIO_UART_DEV, ...)` triggers the assertion in `cpu/mips_pic32_common/periph/uart.c:51` because, on PIC32MX and PIC32MZ, UART peripherals indexing start from 1 (as I think all other peripherals).

Here are the patches for pic32-wifire and pic32-clicker boards (I have tested this PR on both boards): [mips-patches.zip](https://github.com/RIOT-OS/RIOT/files/2891511/mips-patches.zip)

maribu:
@francois-berder: Thanks for identifying and fixing the problem. I added your patches to this PR

maribu:
@kenrabold or @pyropeter: Could one of you test again with the HiFive1 and confirm that it is now working?

@OlegHahm: I heard you have some msp430 laying around? Would you mind to give this PR a test?

Would someone mind to confirm that it works still on AVR? (I tested successfully on the Arduino Uno and Arduino Mega2560, but I'm afraid of the "Works on My Machine Syndrome".)

maribu:
Btw: @haukepetersen noticed that `stdio_rtt` no longer works in master. So testing on the RuuviTag or the Thingy:52 is not possible anyway.

kenrabold:
@maribu For this PR to work on HiFive1, this change is needed in the HiFive1 board_init() function:

```
void board_init(void)
{
    /* Initialize CPU and clocks */
    board_init_clock();
    cpu_init();

```
The board_init_clock() call must happen before the cpu_init().  Because periph_init() is now called by cpu_init(), and that periph_init() will call down into stdio_init() which then calls uart_init(), the UART on the HiFive1 is configured based on the power on clock, and then that gets changed by board_init_clock(). That results in the divisor for the UART not being correct and the text going to serial output is all messed up.

So, just update the board.c for the HiFive1 to be like the code above and this PR will work.

maribu:
@kenrabold: Thanks for pointing that out. Can you test again to confirm it now works?

kenrabold:
@maribu - Yes, that works

maribu:
OK, I tested on the MSP430 platform myself using the wonderful [FIT/IoT-Lab](https://www.iot-lab.info). It seems to work on every supported platform. May I squash?

drivers/periph_common/init.c
```diff
@@ -36,8 +45,23 @@
 #include "periph/hwrng.h"
 #endif
 
+
 void periph_init(void)
 {
+    /* initialize stdio first to allow DEBUG() during later stages */
+#ifdef CPU_NATIVE
```
haukepetersen: This 'ifdef' hell should not be here. After all, `periph_common` is a platform independent module and adding all this platform dependencies here does destroy this whole concept!

The better way I see is to move these calls to the common platform modules, calling these functions from right before `periph_init` is called (or not in e.g. case of native). This way there is no need for any hard to read ifdefs anywhere, just a `#ifdef MODULE_STDIO` in for each supported architecture...

maribu: One general remark: It is very rarely seen that someone jumps in at the very end of a process and after a lot of work has been done and asks for dropping core concepts of this process. For good reason. Also a lot of people [tend to get very upset](https://mailarchive.ietf.org/arch/msg/tls/CzjJB1g0uFypY8UDdr6P9SCQBqA) about such things.

That said, the idea of this PR is to assure that stdio is available when the periphs are initialized, so that using `DEBUG()` there is a save thing to do. This solves a real [issue](https://github.com/RIOT-OS/RIOT/issues/10645), as people starting to `ENABLE_DEBUG` in periph initialization would like to debug that periph code and don't want to think about why the debug output does not appear in the console or even crashes the board.

By having the stdio initialization here centralized, it is enforced that `DEBUG()` is safe to do in periph initialization. If it is up to the board/cpu, inconsistent behavior can easily sneak back it. So this idea about making RIOT *less* hardware dependent.

How about this instead:

``` C
#ifndef HAVE_ARCH_STDIO_INIT
static inline void arch_stdio_init(void)
{
    stdio_init();
}
#endif

void periph_init(void)
{
#if defined(MODULE_STDIO_UART) || defined(MODULE_STDIO_RTT)
    arch_stdio_init();
#endif
    ....
}
```

This would allow to provide a architecture specific `void arch_stdio_init()` if needed and handle the hardware specific stuff there.

haukepetersen: > One general remark: It is very rarely seen that someone jumps in at the very end of a process

I am aware I and I honestly don't feel good about my behavior. I simply havn't seen this PR before (as I am not spending too much time on tracking PRs lately), and I stumbled upon this today - but when I see something like this I would simply feel bad about staying quit. So please accept my apologies.

On the topic at hand: I still disagree with putting this initialization into `periph_common`. `periph_common` is about peripheral, and STDIO is clearly not part of this. So just using `periph_common/init.c` as init trigger for some (critical, yes)) system module breaks with a global concept and is ergo always a bad idea!

So what would be needed is some generic 'pre-init' function reliably called by all platforms before `periph_common/init`, that can be used to bootstrap this kind of system modules. I don't see a large issue with doing such a thing, right?

Process wise I would be ok with merging this PR first (as it is already nicely tested), if someone promises to address this issue right afterwards.

maribu:
@haukepetersen: This [latest commit](https://github.com/RIOT-OS/RIOT/pull/10806/commits/a776e9470ac94ee8d84458f54ea98a2b314e5f52) implements what I suggested as alternative to your feedback

drivers/periph_common/init.c
maribu: Maybe it is best to split off the mips32 changes, as any PR trying to reduce the hardware specific differences will depend on that. Maybe @francois-berder could open an PR for that, as he authored those commits anyway?

And maybe it would be a good idea to add get a general picture how the boot up process in an *ideal* RIOT board look like. Maybe a sequence diagram like [this one](https://api.riot-os.org/riot-netdev-rx.svg) in the [doc of the Network Device Driver API](https://api.riot-os.org/group__drivers__netdev__api.html) could help. In that diagram the hardware specific part could be color coded, as those will likely still differ in various ways due to specific hardware dependent requirements.

Here is the sequence diagram I linked above (sadly with broken style):

seq_diagram

Such a diagram might help to a lot to at least have new added boards/architectures behave in a similar way and might be a good starting point to refactor and work out the differences of the existing boards.

haukepetersen:
I see, but it does not solve the issue for abusing the `periph_comon/init.c` for non-peripheral implemenations.

maribu:
> I see, but it does not solve the issue for abusing the periph_comon/init.c for non-peripheral implemenations.

Maybe one could view stdio as an peripheral device in a broader sense, as it is an upper layer for interacting with peripheral devices such as keyboard/terminal/serial port on a PC and UART/RTT in the context of RIOT. Those devices are definitely peripheral devices and stdio is RIOT's default interface to UART when there is a human on the other side of the UART port.

Looking into the code of `stdio_init()` for the UART case, it basically initializes the UART device, thus, a peripheral device. So this argument is not as weak as it appears on the first view.

kaspar030:
> Looking into the code of `stdio_init()` for the UART case, it basically initializes the UART device, thus, a peripheral device. So this argument is not as weak as it appears on the first view.

IMO that proves only that stdio_uart has periph/uart as a dependency. Thus initializing stdio_uart before the uart periph subsystem got a chance to be initialized (through periph_init()) is at least dangerous (or well, will lead to crashes).

I'd assume that stdio_rtt would have a similar issue. It depends on xtimer, which depends on periph_timer, which is also not initialized at this point.

haukepetersen:
I agree with @kaspar030, that just because `stdio` is using a periph driver, it is not a periph module on its own.

But one important point: the `periph_common/init` is there for initializing peripherals, that can be statically configured without taking any 'using module's' needs into account (e.g. i2c, spi, ...). 'timer' and 'uart' are a little different, as they are under full control if their users (e.g. xtimer, stdio, ...), and so these periph drivers are initialized by those modules directly, and are therefore independent from the `periph_common` module.

So again, we all agree, that `stdio` needs some extra care when it comes to initialization order. So I don't see a reason not to include its initialization in the platform specific code, especially considering the potential stdlib differences.

Side note (revisiting the discussion above): the `segger_rtt` mapping was not broken, it was my local python that prevented it from working. So now I can test those platforms in question again (at least when back in the office).

maribu:
@kaspar030, @haukepetersen:
OK, so the bottom line is that `stdio` initialization should be moved into `cpu_init()` just before the call to `periph_init()`, right?

How about in addition to that we add a test to `periph_init()` to verify that `stdio` is initialized and working during automated tests? That way it could be kind of enforced that `DEBUG()` is safe to call at least at that point in time during boot.

haukepetersen:
> OK, so the bottom line is that stdio initialization should be moved into cpu_init() just before the call to periph_init(), right?

yes from my side.

> How about in addition to that we add a test to periph_init() to verify that stdio is initialized and working during automated tests? That way it could be kind of enforced that DEBUG() is safe to call at least at that point in time during boot.

Personally, I think this is overkill. Of course we need to make sure, that `DEBUG` does not trigger any harmful behavior (as in hardfaults etc). So IMHO DEBUG, or more precise the STDIO, should just do nothing in case STDIO is not (yet) initialized or even selected for a build. 

What we might want to think about in parallel, is to add some specific, general documentation on the boot order and the 'special cases' like `stdio init` and similar. Also a general rule about where it is save to use DEBUG and STDIO in the boot sequence could be added.

maribu:
Honestly, I'm not fully sure if a panic on `DEBUG()` before stdio is available is better than not having any indication that something went wrong. `DEBUG()` is not supposed to happen in production code. And a developer that added `DEBUG()` to some code and gets no output might have to look at a lot of things before tracing the issue down to not having `stdio` available.

But more generally: @MrKevinWeiss already implemented a safeguard in https://github.com/RIOT-OS/RIOT/pull/10615 that drops UART output when used prior to initialization. But that change affects `DEBUG()` and production code alike and could potentially result in `DEBUG()` output getting lost. This PR is intended to be a kind-of follow up to make sure that this never happens.

> Personally, I think this is overkill

Let me point out again that the test I suggested to add to `periph_init()` for automated testing is not supposed to be compiled in by default (as that is only useful for automated testing). It could be as simple as:

```C
#ifdef TEST_STDIO_OUTPUT_ON_PERIPH_INIT
    puts("Initializing peripheral devices...");
#endif
```

maribu:
I remembered https://github.com/RIOT-OS/RIOT/pull/10615 incorrectly. It will only drop output with `DEVELHELP` to fix boards not booting with `ENABLE_DEBUG` in some periph code.

haukepetersen:
I'd like to take a step back and focus on the core problem at hand: a clear definition of the point in time in the boot sequence, from which it is safe to use STDIO. We all agree, that this should as soon as possible, to allow for using STDIO early on during boot. If this is 'standardized', IMO there is no need for additional checking code if STDIO is available or not.

So what are the challenges:
- (a) different `stdio` implementations have diverging dependencies on `periph` and other libraries (`stdio_uart` uses `periph_uart`, while `stdio_rtt` uses `xtimer` and with this `periph_timer`). So this is not something we can easily generalize.
- (b) depending on the used platform, there are potential some prerequisites, that have to be done before stdio can be initialized
- (c) ?

(b) is easy to solve by moving the init code into the platform specific init before `periph_init()`. 
For (a) it is not quite as trivial -> using `stdio_uart` one can use DEBUG in the timer driver, if using `stdio_rtt` one can clearly not. Calling it from `periph_init` does also not really solve this, right?!

maribu:
I'm closing this PR now in favor of creating individual PRs per platform that move `stdio_init()` prior to `periph_init()`, ideally in `cpu_init()` just before `periph_init()`.

@francois-berder: Would you mind to create such a PR for the mips platform? This PR would greatly depend on the code you contributed anyway.

PR #10808, opened 2019-01-17
doc: include *_params.h files in documentation

### LINKED ISSUE #10427: doxygen: *_params.h files are excluded
#### Description

```*_params.h``` files which usually contain

- default configuration parameters for drivers or
- board specific configurations of GPIOs and ADCs that are mapped to SAUL

are excluded from doxygen documentation by the doxygen file.

What is the reason? Isn't it worth it to have these parameters and configurations documented in doxygen? IMHO, it is. It avoids looking into the corresponding files. Furthermore, at the moment, it is quite difficult to reference definitions, parameters and configuration from other parts of documentation.

I have seen, that excluding ```_params.h``` was added to ```EXCLUDE_PATTERN``` with PR #6688,  commit 28c335e8e7000f31283c0fe6b7a1590fe2ca0057 by @OlegHahm.

### PR DESCRIPTION
**Pull request by gschorcht**
### Contribution description

This PR tries to fix the problem described in issue #10427. Due to some warnings like
```
drivers/sdcard_spi/include/sdcard_spi_params.h:13: warning: the name 'drivers/sdcard_spi/include/sdcard_spi_params.h' supplied as the second argument in the \file statement matches the following input files:
   drivers/sdcard_spi/include/sdcard_spi_params.h
Please use a more specific name by including a (larger) part of the path!
```
`*_params.h` were excluded from documentation :worried:

After some investigations it became clear that `*_params.h` files can be included in the documentation without any warning if `STRIP_FROM_INC_PATH` variable is empty in `riot.doxyfile`. The generated documentation looks as before including the include paths of header files in documentation. The only differences are
- `*_params.h` files are now included in documentation as expected and
- dependency graphs include all files from which they depend and not only the direct includes.

This PR depends on PR #10810.

### Testing procedure

- Generate the docuementation with `make doc` in your RIOT directory
  ```
   cd <your_RIOT_dir>
   make doc
   ```
- Open the documentation in browser using URL `https:///<your_RIOT_dir>/doc/doxygen/html/index.html`.
- Open complex pages like
   - `Networking » Generic (GNRC) network stack » IPv6 » ICMPv6` and
   - `Networking » Generic (GNRC) network stack » IPv6 » ICMPv6 » gnrc/icmpv6.h`
- Check for
   - included header files
   - dependency graph
   - navigation tree
   ...
- Search for parameters, for example `sdcard_spi_params`
  - In the official docs the only thing that show up is `sdcard_spi_params_t ` (the definition of the structure).
  - With this PR the instantiations of `sdcard_spi_params` (define in "*_pararam.h" files) also appear.

### Issues/PRs references

Fixes #10427. Depends on PR #10810.

### DIFF
--- a/doc/doxygen/riot.doxyfile
+++ b/doc/doxygen/riot.doxyfile
@@ -140,7 +140,7 @@ INLINE_INHERITED_MEMB  = NO
 # shortest path that makes the file name unique will be used
 # The default value is: YES.
 
-FULL_PATH_NAMES        = NO
+FULL_PATH_NAMES        = YES
 
 # The STRIP_FROM_PATH tag can be used to strip a user-defined part of the path.
 # Stripping is only done if one of the specified strings matches the left-hand
--- a/doc/doxygen/riot.doxyfile
+++ b/doc/doxygen/riot.doxyfile
@@ -161,7 +161,7 @@ STRIP_FROM_PATH        =
 # specify the list of include paths that are normally passed to the compiler
 # using the -I flag.
 
-STRIP_FROM_INC_PATH    = $(STRIP_FROM_INC_PATH_LIST) # Exported from Makefile.
+STRIP_FROM_INC_PATH    = # $(STRIP_FROM_INC_PATH_LIST) # Exported from Makefile.
 
 # If the SHORT_NAMES tag is set to YES, doxygen will generate much shorter (but
 # less readable) file names. This can be useful is your file systems doesn't
--- a/doc/doxygen/riot.doxyfile
+++ b/doc/doxygen/riot.doxyfile
@@ -152,7 +152,7 @@ FULL_PATH_NAMES        = YES
 # will be relative from the directory where doxygen is started.
 # This tag requires that the tag FULL_PATH_NAMES is set to YES.
 
-STRIP_FROM_PATH        =
+STRIP_FROM_PATH        = ../../
 
 # The STRIP_FROM_INC_PATH tag can be used to strip a user-defined part of the
 # path mentioned in the documentation of a class, which tells the reader which
--- a/doc/doxygen/riot.doxyfile
+++ b/doc/doxygen/riot.doxyfile
@@ -821,7 +821,6 @@ EXCLUDE_SYMLINKS       = NO
 # exclude all test directories for example use the pattern */test/*
 
 EXCLUDE_PATTERNS       = */board/*/tools/* \
-                         *_params.h \
                          */cpu/atmega_common/include/sys/*.h \
                          */cpu/msp430_common/include/stdatomic.h \
                          */cpu/msp430_common/include/sys/*.h \

### CODE REVIEW

**Comment by miri64** [comment]
Edited OP to allow for automatic closure of #10427 ;-) (see https://help.github.com/articles/closing-issues-using-keywords/.)

**Comment by jcarrano** [comment]
I updated the description so that people can see the problem being fixed.

**Comment by jcarrano** [comment]
In my case, I'm getting a __ton__ of "xxx is not documented" warnings. Other than that the docs seem to build fine.

<details><summary>My system</summary>
<p>

```
Operating System Environment
-----------------------------
                 Kernel: Linux 4.20.1-arch1-1-ARCH x86_64 unknown

Installed compiler toolchains
-----------------------------
             native gcc: gcc (GCC) 8.2.1 20181127
                avr-gcc: missing
       mips-mti-elf-gcc: missing
             msp430-gcc: missing
   riscv-none-embed-gcc: missing
   xtensa-esp32-elf-gcc: missing
   xtensa-lx106-elf-gcc: missing
                  clang: clang version 7.0.1 (tags/RELEASE_701/final)

Installed compiler libs
-----------------------
   arm-none-eabi-newlib: "3.0.0"
    mips-mti-elf-newlib: missing
xtensa-esp32-elf-newlib: missing
xtensa-lx106-elf-newlib: missing
               avr-libc: missing (missing)

Installed development tools
---------------------------
                  cmake: cmake version 3.13.2
               cppcheck: Cppcheck 1.86
                doxygen: 1.8.15
                 flake8: 3.6.0 (mccabe: 0.6.1, pycodestyle: 2.4.0, pyflakes: 2.0.0) CPython 3.7.2 on Linux
                    git: git version 2.20.1
                   make: GNU Make 4.2.1
                 python: Python 3.7.2
                python2: Python 2.7.15
                python3: Python 3.7.2
             coccinelle: missing
```
</p>
</details>

**Comment by gschorcht** [comment]
@jcarrano Ah, you are using doxygen 1.8.15. I saw a lot of warnings too, when I tried to use the latest release version of doxygen. The question is, which version do we have to support for the moment. riotdocker contains 1.8.11.

**Comment by gschorcht** [comment]
@jcarrano BTW, you will get the same warnings with doxygen 1.8.15 with the master branch. So it seems that it has nothing to do with this PR.

**Comment by gschorcht** [comment]
Travis also gives green light. That is, it also worked without warnings or errors.

**Comment by gschorcht** [comment]
@miri64 Could you take a look to the comment I added in https://github.com/RIOT-OS/RIOT/pull/10808/commits/c01a666f911bf3fd034bf74ebf434bcdcf8256b whether it is ok for you?

**Comment by gschorcht** [comment]
@miri64 
> Edited OP to allow for automatic closure of #10427 ;-) (see https://help.github.com/articles/closing-issues-using-keywords/)

Thanks, Good to know, I thought the issue keyword is also accepted.

File: cpu/native/include/socket_zep_params.h [hunk 1]
```diff
@@ -25,7 +25,7 @@ extern "C" {
 #endif
 
 /**
- * @brief   Number of allocated parameters at @ref socket_zep_params
+ * @brief   Number of allocated parameters at #socket_zep_params
```
miri64: I don't think this is portable to XML and LaTeX.

miri64: For me (also doxygen 1.8.15 at the moment) keeping it as it was doesn't pose a problem.

gschorcht: Ok, I thought to remember that \ref explicitly requires the definition of an \anchor. However, this only seems to be only the case when you try to refer to a section in another document. I will change it back.

gschorcht:
> Regarding updating the doc: regardless of how we fix the errors, this should not happen in this PR. Since Murdock and Travis are at the moment happy, we could either just merge it as is or fix the found errors in separate PRs and then merge this PR. It's definitely a nice newcomer task ;-)

I considered this PR more like "make `*_params.h` visible and fix the remaining warnings in `*_params.h`". I can split them.

gschorcht:
@miri64
> For me (also doxygen 1.8.15 at the moment) keeping it as it was doesn't pose a problem.

You have not seen the warnings that @jcarrano and I have seen with doxygen 1.8.15? I'm wondering what the difference is.

gschorcht:
@miri64 PR splitted into #10808 and #10810.

gschorcht:
One question remains: Should we set `FULL_PATH_NAMES = yes`?
The only visible difference is that header files are referenced only by their file names and not by their relative paths in the documentation. But from my point of view, these relative paths are more confusing than helpful. For example, in [Module » CPU » Native](https://riot-os.org/api/group__cpu__native.html)

screen_shot_2019-01-18 08-01-40

all files are coming from `cpu/native/include` but the paths seem to be arbitrary. And because the paths are sorted alphabetically by filenames, it is less readable than if only file names were shown.

cpu/native/include/socket_zep_params.h
miri64: With other documents is not correct either. `@anchor` you only need if the thing you want to reference is not auto-anchored (e.g. groups created with `@name`).

miri64:
> @miri64
> 
> > For me (also doxygen 1.8.15 at the moment) keeping it as it was doesn't pose a problem.
> 
> You have not seen the warnings that @jcarrano and I have seen with doxygen 1.8.15? I'm wondering what the difference is.

Oh I did, lots of them, just the one you were trying to fix with the `s/@ref/#/` I did not see ;-).

cpu/native/include/socket_zep_params.h
gschorcht: I just remember the case where I had to add an anchor. I had a document that was defining sections like markdown with `#`.  Then, I had to add the `\anchor` statement explicitly.

gschorcht:
@miri64 
> $ make doc 2>&1 | wc -l
13073
>
> make doc 2>&1 | wc -l
12704

I guess the different numbers result from the fact that with PR there are more files that are included.

gschorcht:
@jcarrano @miri64 

I investigated the `multiple use of section label` in doxygen 1.8.15 a bit more since we should expect that it will become the standard version in upcoming Linux distros.

It seems to be a bug, maybe related to https://github.com/doxygen/doxygen/issues/6605. But he is claiming in his bug report that the problem was not present in version 1.8.6. And, indeed there was a [problem in 1.8.7 ](https://stackoverflow.com/questions/31118585/doxygen-multiple-use-of-section-label-bug) which was fixed in 1.8.8.

I have used the very small test file together with doxygen-1.8.15 and the standard Doxyfile:
```md
Main Page                                                           {#mainpage}
=========

Section 1                                                           {#section1}
=========
Test text for Section 1 

Section 2                                                           {#section2}
=========
Test text for Section 2
```
It throws exactly the same warnings.

gschorcht:
@jcarrano @miri64 I have opened a tracking issue #10815 for the doxygen 1.8.15 problems. It is a more general problem that has definitely nothing to do with this PR.

gschorcht:
I think we should make a decision regarding my question https://github.com/RIOT-OS/RIOT/pull/10808#issuecomment-455448277 to make some progress. IMHO, it would be worth to get `*_params.h` documented and maybe to backport it to `2019.01`.

The warnings mentioned by @jcarrano are not related to this PR but a general problem to which we will run into with doxygen 1.8.15.

miri64:
> IMHO, it would be worth to get `*_params.h` documented and maybe to backport it to `2019.01`.

a) Feature freeze is in effect, so only bug fixes get backported now,
b) The rendered doc @ doc.riot-os.org is only rendered from master so this wouldn't even make sense

gschorcht:
> a) Feature freeze is in effect, so only bug fixes get backported now,

Ok. The PR is probably rather a fix than a feature :wink: Excluding `*_params.h` from documentation was more a hack than a solution.

> b) The rendered doc @ doc.riot-os.org is only rendered from master so this wouldn't even make sense

Ah, I see. Then, indeed it doesn't matter.

Anyway, I think it's worth making some progress.

gschorcht:
@jcarrano

> In my case, I'm getting a **ton** of "xxx is not documented" warnings. Other than that the docs seem to build fine.

Referring to your comments regarding the tons of warnings in https://github.com/RIOT-OS/RIOT/issues/10815#issuecomment-461368336 and https://github.com/RIOT-OS/RIOT/pull/10808#issuecomment-455252346, all these warnings are not related to this PR but caused by doxygen 1.8.15, also with the current master.

IMHO, we should merge this PR sometime to document also `* _params.h` files. The only remaining question was whether to use `FULL_PATH_NAMES = yes` or not, see my comment https://github.com/RIOT-OS/RIOT/pull/10808#issuecomment-455448277.

jcarrano:
@gschorcht 

> One question remains: Should we set `FULL_PATH_NAMES = yes`?

I think yes: the paths generated will be much more useful. In that case, I think it's good to also set `STRIP_FROM_PATH=../..` as we discussed in #10427.

gschorcht:
@jcarrano 
> I think yes: the paths generated will be much more useful. In that case, I think it's good to also set `STRIP_FROM_PATH=../..` as we discussed in #10427.

I added `FULL_PATH` and `STRIP_FROM_PATH` changes to this PR. Furthermore, I rebased it to see whether it still generates the current doc without warnings.

jcarrano:
This looks good for me. I wonder if it makes sense to squash commits 1, 3 and 4 together.

gschorcht:
@jcarrano 
> This looks good for me. I wonder if it makes sense to squash commits 1, 3 and 4 together.

I could do this. Commit message of commit 1 has to be fixed anyway :worried: Should I squash?

jcarrano:
Go ahead.

gschorcht:
Finally, I decided to leave the changes as separate commits to be able to understand the changes later.

gschorcht:
@jcarrano Thanks

PR #10812, opened 2019-01-18
tests/periph_rtc: add automatic python test script

aabadie:
### Contribution description

This PR is just adding a test script for the `periph_rtc` application. There is some specific checks for the native plateform because some functions used in the application are not implemented.

### Testing procedure

This PR can be tested on native and on a board providing the RTC feature. I tested with b-l072z-lrwan1 but others could be used.
- On native:
```
$ make -C tests/periph_rtc all test
```
- On b-l072z-lrwan1:
```
$ make BOARD=b-l072z-lrwan1 -C tests/periph_rtc flash test
```

### Issues/PRs references

None

### DIFF
--- a/tests/periph_rtc/tests/01-run.py
+++ b/tests/periph_rtc/tests/01-run.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+
+# Copyright (C) 2019 Inria
+#
+# This file is subject to the terms and conditions of the GNU Lesser
+# General Public License v2.1. See the file LICENSE in the top level
+# directory for more details.
+
+import os
+import sys
+from testrunner import run
+
+
+BOARD = os.getenv('BOARD', 'native')
+DATE_PATTERN = r'\d{4}\-\d{2}\-\d{2} \d{2}\:\d{2}\:\d{2}'
+
+
+def testfunc(child):
+    child.expect(r'This test will display \'Alarm\!\' every 2 seconds '
+                 r'for (\d{1}) times')
+    alarm_count = int(child.match.group(1))
+    child.expect(r'  Setting clock to   ({})'.format(DATE_PATTERN))
+    clock_set = child.match.group(1)
+    if BOARD == 'native':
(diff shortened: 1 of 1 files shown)


tests/periph_rtc/tests/01-run.py
```diff
@@ -0,0 +1,49 @@
...
+        child.expect(r'.*rtc_set_alarm: not implemented')
+    child.expect(r'   Alarm is set to   ({})'.format(DATE_PATTERN))
+    alarm_value = child.match.group(1)
+    if BOARD != 'native':
+        # Set alarm is not implemented for native board so no need to compare
+        # alarm values
+        assert alarm_set == alarm_value
+
+    if BOARD != 'native':
+        for _ in range(4):
```
leandrolanzieri: Would it make sense to extract the number of iterations from the first message?

aabadie: Yes definitely, I updated the script accordingly and amended directly

tests/periph_rtc/tests/01-run.py
```diff
@@ -0,0 +1,50 @@
...
+        child.expect(r'.*rtc_set_alarm: not implemented')
+    child.expect(r'   Alarm is set to   ({})'.format(DATE_PATTERN))
+    alarm_value = child.match.group(1)
+    if BOARD != 'native':
+        # Set alarm is not implemented for native board so no need to compare
+        # alarm values
+        assert alarm_set == alarm_value
+
+    if BOARD != 'native':
+        for _ in range(alarm_count):
```
leandrolanzieri:
```suggestion
        for _ in range(int(alarm_count)):
```
Here should be int

aabadie: You are right, did that in a rush on native where it's not used... I opted for changing this on line 21 because the variable is supposed to contain a number.

aabadie:
Thanks for reviewing and testign @leandrolanzieri !

PR #10818, opened 2019-01-18
dist/tools/compile_and_test_for_board: add compile and test script

cladmi:
### Contribution description

Move the compile and test script from Release-Specs.
https://github.com/RIOT-OS/Release-Specs/blob/271dc8/02-tests/compile_and_test_for_board.py

By default it should be run as

     ./compile_and_test_for_board.py path_to_riot_directory board_name [results]

The script is migrated as-is so has not been changed to automatically use the current repository.

I added some TODOs in the README for a better integration/documentation that can be done in further PRs without blocking this one.

### Testing procedure

Run the script for one board as 

     ./compile_and_test_for_board.py path_to_riot_directory board_name

it will compile everything and execute the automated tests for this board.

### Issues/PRs references

It was discussed IRL why the script was not in RIOT.
As it was not in RIOT for the previous release, it was easier to have it outside to use it for the Releases tests.

### DIFF
--- a/dist/tools/compile_and_test_for_board/README.md
+++ b/dist/tools/compile_and_test_for_board/README.md
@@ -19,3 +19,18 @@ They can be checked with:
 
     find results/ -name '*.failed'
     find results/ -name 'test.success'
+
+
+Implementation TODO
+-------------------
+
+The script was developed for being used outside of RIOT so does not currently
+simply adapt to be used from within RIOT.
+
+* Provide another script for new interface?
+* Provide a RIOT/Makefile integration
+* Save output files into `${BUILD_DIR}/output/compile_and_test` directory by
+  default
+* tox file to run `doctests/pylint/flake8`. Add tests.
+* Implement the `board_is_supported`/`board_has_enough_memory`/`has_tests` to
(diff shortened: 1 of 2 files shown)


cladmi:
If could be the good moment to update the script name if wanted.

MrKevinWeiss:
Ya, maybe simplify to test_all?  How would it be invoked, keep it just a python or integrate to make (if so then how can you call make without targeting a specific test)?

MrKevinWeiss:
> It does not test_all but compile all supported/enough memory and test when there are tests

For the sake of simplicity it is running what should be all valid tests.

>  And only for one board so not really all.

But you would specify the board.  I could also go for run_board_tests or just run_tests.


> I would integrate it in the mainRIOT/Makefile like we have for static-test, not the application/Makefile.

Good good

cladmi:
Ohh, maybe I misunderstood. I agree that a short name when doing `make run_the_thing` is good.

I was talking more about the name for the script in `dist/tools` where I find an explicit descriptive name better than a short one.

For the short name/make integration, I would do another PR though.

aabadie:
Let's merge this PR.

